PR #12600, opened 2023-05-09
Add Icon storybook docs

b12f:
This commit adds a documentation page for icons. To achieve this, we extract a list of all icons in the icon font generation script and write it out to `frontend/src/app/spot/icon-font/icons.json`.

This commit also moves all stories to the `frontend/src/stories` directory, so component stories are not in a different place anymore.


oliverguenther (approved):
code-wise, LGTM :+1: I'm not sure I'm on the newest page regarding where the icons should reside now.

PR #12609, opened 2023-05-10
Send crowdin source file for seeders translations

cbliard:
https://community.openproject.org/wp/48016


app/seeders/source/filter_translatables.rb
```diff
@@ -0,0 +1,72 @@
...
+    else
+      object if translatable
+    end
+  end
+
+  def filter_translatables_in_hash(hash)
+    hash
+      .to_h do |key, value|
+        translatable = key.start_with?('t_')
+        [key.sub(/^t_/, ''), filter_translatables_in_object(value, translatable:)]
```
ulferts, line 53: The `t_` has by now spread across a number of files (`Source::FilterTranslatables` and `Source::Translate`) and is becoming somewhat magical one has to know. Extracting it into a constant might be a bit burdensome but should increase stability.

ulferts (approved):
Works as advertised. I only have one minor comment. Feel free to merge this PR, @cbliard, in case you want to address the comment separately.

app/seeders/source/filter_translatables.rb
cbliard: You're right, I'll do that. Thanks.

PR #12610, opened 2023-05-10
[#48167] Fix storages.provider_fields serialization issue.

ba1ash:
See https://community.openproject.org/work_packages/48167 description.


wielinde (changes requested):
Hey @ba1ash, We still need to migrate storage models that were created with the wrong default value for provider_fields. All the rest looks really good.

wielinde (approved):
Great! I checked it, and it works!

PR #12611, opened 2023-05-10
[46803] Implement show changes icons on cards

bsatarnejad:
https://community.openproject.org/projects/openproject/work_packages/46803/activity


oliverguenther (changes requested):
This PR addds duplication on how the icon is created. Let's extract this into a helper method

oliverguenther (approved):

PR #12466, opened 2023-04-20
Make seed data strings translatable

cbliard:
https://community.openproject.org/wp/47353

The idea is to generate the translation files from the seed data files `standard.yml` and `bim.yml`, get it communicated to crowdin to have actual translations, and translate the seed data. Any text in the seed data files having a `t_` prefix in its key is a text to be translated. The corresponding i18n key is generated from its path in the seed data yaml.

* [x] Translate translatable keys when loading the seed data file.
* [x] Identify translatable seed text in `standard.yml` and `bim.yml`.
* [x] Stop looking up data by its English name. It would not work anymore with another language. Implement a reference mechanism instead.
* ~~Generate `en.seed.yml` files from the translatable keys of the seed data files and send it to crowdin as part of the crowdin GitHub action.~~ Moved to work package 48016

PR #12617, opened 2023-05-11
Address seed data translation review comments

cbliard:
Address review comments from 2 other PRs:
- [x] [Do not use magic string](https://github.com/opf/openproject/pull/12609#discussion_r1189701708)
- [x] [Move type and status translations into seed files](https://github.com/opf/openproject/pull/12466#discussion_r1187207937) 
- [x] [Do not store work package instance in hash attributes](https://github.com/opf/openproject/pull/12466#discussion_r1187214062)

This PR blocks [OP#47355 Use default language setting in docker installation for seeding](https://community.openproject.org/projects/openproject/work_packages/47355)


cbliard:
Once merged, some translation strings like `default_status_new` will be removed from the `config/locales/en.yml` file, and the equivalent will be added in the generated `config/locales/en.seeders.yml` file and uploaded to crowdin.

This would normally mean that translators in crowdin have to translate them all again.
This can be avoided.

Crowdin is able to automatically reuse a known translation if the text to translate has already been translated once. To do so with the seed files, we have to:
1. generate the `en.seeders.yml` files and upload them to crowdin by triggering the crowdin GitHub action
2. ask crowdin to reuse existing translations by running the following command (uses the crowdin cli)
    ```
    crowdin pre-translate --translate-untranslated-only --branch dev --method tm
    ```

app/seeders/basic_data/role_seeder.rb
```diff
@@ -26,176 +26,83 @@
...
-        roles.each do |attributes|
-          Role.create!(attributes)
-        end
-
-        global_roles.each do |attributes|
-          GlobalRole.create!(attributes)
-        end
+  class RoleSeeder < ModelSeeder
+    self.model_class = Role
+    self.seed_data_model_key = 'roles'
```
ulferts, line 31: You COULD provide defaults for `model_class` and `seed_data_model_key` so as that they are deduced from the name of the seeder class. That way, one wouldn't have to specify it time and again in every base class. On the other hand, doing it like it is done here is more explicit.

ulferts (changes requested):
I like the changes in this PR quite a lot. To me, it makes the code easier to read and it also looks to be easier to maintain.

Some small bugs seem to have crept in, though:

The following screenshots shows a comparison of two sql dump files where the dev state is on the left and the PR state is on the right side.

* [x] Some queries lack the type filter :

image

* [x] The name of some document categories change:

image

* [x] The number of permissions when seeding dev vs seeding with the PR differ vastly. It appears that the `dev` state is at fault here, though seeding permissions for the same role multiple times. The pr state seeds an additional permission for the 'Project admin'  which is 'share_calendars' which I think is ok. In total, the PR seems to seed the permissions correctly but it makes sense to double check that again:

image

cbliard:
Thanks for the review and for taking the time to make a diff.

For the queries lacking the type filter, that's something that I noticed and fixed in a subsequent PR: 2ee0b65ab4a6166d312f140e47d00d8491d1017b in #12732. I'll move the commit inside this PR where it belongs.

For the 2 others, I'll have a second look and fix them as well.

app/seeders/basic_data/role_seeder.rb
cbliard: You are right. Yet I prefer the explicit approach here, even if it can feel a bit repetitive. 

I'm always a bit hesitant about this kind of topics, as deducing names by convention feels more "the Rails way", but showing a bit of the inners helps about understanding how the class works.

cbliard:
@ulferts I did not find such differences between the permissions when seeding from `dev` and this pr.

I used the unit test `spec/seeders/root_seeder_standard_edition_spec.rb` to test the permissions with `RolePermission.joins(:role).group('roles.name').count`.

dev:
```
[1] pry(#<RSpec::ExampleGroups::RootSeederStandardEdition::DemoData>)> RolePermission.joins(:role).group('roles.name').count
=> {"Reader"=>24, "Staff and projects manager"=>3, "Anonymous"=>5, "Non member"=>8, "Project admin"=>118, "Member"=>76}
```

this pr:
```
[1] pry(#<RSpec::ExampleGroups::RootSeederStandardEdition::DemoData>)> RolePermission.joins(:role).group('roles.name').count
=> {"Reader"=>23, "Staff and projects manager"=>3, "Anonymous"=>5, "Non member"=>8, "Project admin"=>118, "Member"=>74}
```

So Member has 2 less permissions, and Reader one less. I'll check which ones are missing.
But it's not as much as you found out. I'll investigate more tomorrow.

cbliard:
@ulferts I found that boards permissions were duplicated in the common part and in the module part, which lead to additional permissions created in `dev`, and not on this branch because there is a call to `.uniq` before processing them, but apart from that I found nothing else.

note to self: I used the following snippet: `pp RolePermission.joins(:role).pluck('roles.name', 'permission').sort.group_by(&:first).transform_values { |v| v.flat_map(&:last) }`

Can you do a dump of your seeded database from `dev` branch so I can have a closer look at it?

cbliard:
The failing test `support/pages/work_packages/work_packages_table.rb:122` is passing locally. I did not manage to make it fail locally, so maybe it is order-dependent.

PR #12625, opened 2023-05-11
Improve installation instructions for Ubuntu 22.04

DanielGibson:
The apt-key deprecation warning can be avoided by saving the key (with .asc file ending because it's in ascii format) in /etc/apt/trusted.gpg.d/ instead of feeding it to apt-key add

See also https://manpages.ubuntu.com/manpages/jammy/man8/apt-key.8.html#deprecation


as-op:
Thank you!

DanielGibson:
Thanks for merging!
As it also affects the current release, you probably also want this in the stable branch or whatever it takes to show up at https://www.openproject.org/docs/installation-and-operations/installation/packaged/#ubuntu-2204 ?

PR #12626, opened 2023-05-11
[#47993] Support Project status explanation changes when rendering project activity

aaron-contreras:
Adds rendering ability for project status explanation changes on activity pages

Work package: https://community.openproject.org/work_packages/47993


config/locales/en.yml
```diff
@@ -551,6 +551,15 @@ en:
           true: "public"
           false: "private"
         queries: "Queries"
+        status_code: 'Status'
+        status_explanation: 'Project status description'
```
aaron-contreras, line 555: Until the required PR is merged into `dev`, the only real diff is the line I'm commenting on.

app/controllers/journals_controller.rb
```diff
@@ -101,9 +101,8 @@ def field_param
     @field_param ||= params[:field].parameterize.underscore
   end
 
-  # Is this a valid field for diff'ing?
-  def valid_field?
-    field_param == 'description'
+  def valid_field_for_diffing?
```
cbliard, line 104: I appreciate that you removed the comment and updated the method name to make the intention clearer :+1:

spec/features/activities/project_attributes_activity_spec.rb
```diff
@@ -87,6 +87,7 @@
 
       # own fields
       expect(page).to have_selector('li', text: "Name changed from #{previous_project_attributes['name']} to #{project.name}")
+      expect(page).to have_selector('li', text: 'Project status description set (Details)')
```
cbliard: Can you move this assertion under the `expect(page).to have_selector('li', text: 'Project marked as template')` so that the order is the same for assignments and assertions?

cbliard (approved):
Thanks. I tested it and it works as expected.

I only have a nitty comment about assertions order and then it will be good to merge.
Once fixed feel free to merge it and update the work package status.

spec/features/activities/project_attributes_activity_spec.rb
aaron-contreras: I sure can!

PR #12633, opened 2023-05-12
Feature: docker multi arch support

machisuji:
Adds dockerization for arm64/v8 and ppc64le on top of amd64. Doesn't push right now. Once this has been approved via review we can adjust the workflow to actually push.

Todo once approved and **BEFORE** merging:

- [x] Remove branch trigger for this feature branch
- [x] Enable push in last step of docker workflow
- [x] change docker project from community-test to community

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -6,8 +6,6 @@ on:
     - cron: '20 2 * * *' # Daily at 02:20
 
   push:
-    branches:
-      - feature/docker-multi-arch-support
     tags:
       - v*
 
@@ -15,7 +13,7 @@ permissions:
   contents: read # to fetch code (actions/checkout)
 
 env:
-  REGISTRY_IMAGE: openproject/community-test
+  REGISTRY_IMAGE: openproject/community
 
 jobs:
   build:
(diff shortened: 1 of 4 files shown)


.github/workflows/docker.yml
```diff
@@ -1,40 +1,75 @@
...
+        uses: docker/setup-buildx-action@v1
+      - name: Login to Docker Hub
+        uses: docker/login-action@v2
         with:
-          name: openproject/community
           username: ${{ secrets.DOCKER_USERNAME }}
           password: ${{ secrets.DOCKER_PASSWORD }}
-          tag_semver: true
-          cache: ${{ github.event_name != 'schedule' }}
-      - name: Test
```
oliverguenther, line 35: You removed the test step from the steps, can we enable this for all builds now?

oliverguenther, line 35: You removed the test step from the steps, can we enable this for all builds now?

oliverguenther (changes requested):

.github/workflows/docker.yml
machisuji: Good point. I found a way to test things BEFORE pushing the image, too. Will amend that!

machisuji:
That's now working. The job builds the docker image for amd64, ppc64le and arm64/v8 and pushes everything under one tag:

image

Since this is a feature branch, there was no version detected and the tag is `sha-...`. For the `dev` branch the tag will be `dev` and for actual releases it will be the respective version(s) (major, major.minor, major.minor.patch).

machisuji:
@oliverguenther This is good to merge now IMHO.

oliverguenther (approved):

PR #12643, opened 2023-05-15
[#47992] Support Project status changes when rendering project activity

aaron-contreras:
Adds rendering ability for project status code changes on activity pages

Work package: https://community.openproject.org/work_packages/47992

### DIFF
--- a/spec/features/activities/project_attributes_activity_spec.rb
+++ b/spec/features/activities/project_attributes_activity_spec.rb
@@ -57,6 +57,7 @@ describe 'Project attributes activities' do
     new_project_attributes = {
       name: 'a new project name',
       description: 'a new project description',
+      status_code: 'on_track',
       public: true,
       parent: nil,
       identifier: 'a-new-project-name',
@@ -88,6 +89,7 @@ describe 'Project attributes activities' do
       # own fields
       expect(page).to have_selector('li', text: "Name changed from #{previous_project_attributes['name']} to #{project.name}")
       expect(page).to have_selector('li', text: 'Description set (Details)')
+      expect(page).to have_selector('li', text: "Project status set to On track")
       expect(page).to have_selector('li', text: 'Visibility set to public')
       expect(page).to have_selector('li', text: "No longer subproject of #{parent_project.name}")
       expect(page).to have_selector('li', text: 'Project unarchived')
--- a/spec/lib/journal_formatter/project_status_code_spec.rb
+++ b/spec/lib/journal_formatter/project_status_code_spec.rb
@@ -1,0 +1,81 @@
+# frozen_string_literal: true
+
+#-- copyright
+# OpenProject is an open source project management software.
+# Copyright (C) 2012-2022 the OpenProject GmbH
+#
+# This program is free software; you can redistribute it and/or
+# modify it under the terms of the GNU General Public License version 3.
+#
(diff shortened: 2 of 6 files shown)


spec/features/activities/project_attributes_activity_spec.rb
```diff
@@ -88,7 +89,7 @@
       # own fields
       expect(page).to have_selector('li', text: "Name changed from #{previous_project_attributes['name']} to #{project.name}")
       expect(page).to have_selector('li', text: 'Description set (Details)')
-      expect(page).to have_selector('li', text: 'Visibility set to public')
```
cbliard, line 91: That's nice, when inserting the new assertion, you respected the same order in the assignments and the assertions.

It looks like the visibility change text assertion was removed. Can you restore it please?

spec/lib/journal_formatter/project_status_code_spec.rb
```diff
@@ -0,0 +1,81 @@
+# frozen_string_literal: true
```
cbliard, line 1: Nice :)

spec/lib/journal_formatter/project_status_code_spec.rb
```diff
@@ -0,0 +1,81 @@
...
+    context 'when setting a status' do
+      let(:old_value) { nil }
+      let(:new_value) { status_code_as_integer('off_track') }
+      let(:expected) do
+        I18n.t(:text_journal_set_to,
+               label: "<strong>Project status</strong>",
+               value: "<i>Off track</i>")
+      end
+
+      it { expect(instance.render(status_code_key, [old_value, new_value])).to eq(expected) }
```
cbliard, line 52: I was worried that the oneliner syntax would yield some hard-to-read test descriptions, but in fact it's quite the opposite. Here is what I get when running `rspec --format documentation <this-file>`:

```
OpenProject::JournalFormatter::ProjectStatusCode
  #render
    when deleting a status
      is expected to eq "<strong>Project status</strong> deleted (<strike><i>Discontinued</i></strike>)"
    when modifying a status
      is expected to eq "<strong>Project status</strong> changed from <i>Off track</i> to <i>On track</i>"
    when setting a status
      is expected to eq "<strong>Project status</strong> set to <i>Off track</i>"
```

It reads nicely I think. Good job.

cbliard (approved):
Good job :+1: 

There is only one assertion that got deleted and which you should restore before merging. Once done, feel free to merge it.

spec/features/activities/project_attributes_activity_spec.rb
aaron-contreras: Fat-fingered it, my bad.

spec/lib/journal_formatter/project_status_code_spec.rb
aaron-contreras: I saw this pattern in the specs for other formatters. I was curious what it would read out but it seems to be quite... explicit. Which in my mind is perhaps a really good way of actually documenting what is expected. I had trouble coming up with a description for the individual examples that can beat what "no description" generates. Do you have an idea for something better?

cbliard: I agree with you. It is sometimes troublesome to find an appropriate description, and sometimes it feels like duplicating the code.
Here the autogenerated description is really good. I would not change it.

PR #12655, opened 2023-05-17
[#47010] Replace Members Cells by ViewComponent

dombesz:
The second part of https://community.openproject.org/work_packages/47010 . This replaces all the remaining Member cells with components. Also introduces the `render_collection` method instead of rendering each row separately.


app/components/members/row_component.rb
```diff
@@ -89,13 +89,12 @@ def roles_label
       label
     end
 
-    def role_form_cell
-      Members::RoleFormCell.new(
+    def role_form_component
+      render Members::RoleFormComponent.new(
```
cbliard: nitpicking a bit: if it's returning the rendered component, it should not be named `role_form_component` but maybe `role_form`.

cbliard (approved):
Thanks :pray: 
You ported the Cells to Components. I leads to a little bit of duplication but I understand this is only transient and this duplication will disappear by itself once everything is migrated.
I have a minor remark that you are free to ignore. Feel free to merge whenever you want.
Nice work.

PR #12659, opened 2023-05-17
Add bin/dev command from Rails

klaustopher:
- Adds the `bin/dev` command
- Updates the documentation to use `bin/dev` instead of `foreman start`


aaron-contreras (commented):
Not in "Tech writers" but big thumbs up! 👍

oliverguenther (approved):

PR #12666, opened 2023-05-19
 [#48259] Fix missing wiki details

mrmir:
Fixes https://community.openproject.org/work_packages/48259

The line was removed when wiki_content was removed but as the tests didn't check the detail being displayed explicitly it was not caught earlier. I have reinstated the custom journal formatter in wiki_page this time and also updated the spec file to check for this case in the future.

### DIFF
--- a/app/models/wiki_page.rb
+++ b/app/models/wiki_page.rb
@@ -54,6 +54,8 @@ class WikiPage < ApplicationRecord
 
   acts_as_journalized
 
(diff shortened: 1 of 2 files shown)


cbliard (approved):
Thanks for the fix

cbliard (approved):
Thanks for the fix

cbliard:
Merging as the failing unit test is unrelated and originates from an issue in `dev` (which is now fixed).

PR #12671, opened 2023-05-22
[#46331] Project members experience a fully automatically managed project folder

ba1ash:
https://community.openproject.org/work_packages/46331
https://community.openproject.org/projects/work_packages/48137


modules/storages/app/common/storages/peripherals/storage_interaction/nextcloud/add_user_to_group_command.rb
```diff
@@ -0,0 +1,81 @@
...
+      )
+
+      case response
+      when Net::HTTPSuccess
+        statuscode = Nokogiri::XML(response.body).xpath('/ocs/meta/statuscode').text
+        case statuscode
+        when "100"
+          ServiceResult.success(message: "User has been added successfully")
+        when "101"
+          ServiceResult.failure(message: "No group specified")
```
apfohl: In failure cases we should return the `:error` symbol as well. Maybe use the `Util.error` method for it.

modules/storages/app/common/storages/peripherals/storage_interaction/nextcloud/file_query.rb
```diff
@@ -27,55 +27,58 @@
 #++
 
 module Storages::Peripherals::StorageInteraction::Nextcloud
-  class FileQuery < Storages::Peripherals::StorageInteraction::StorageQuery
+  class FileQuery
     using Storages::Peripherals::ServiceResultRefinements
 
+    # TODO: check if this path is present on nextcloud instance withou pretty urls enabled
```
apfohl: `without`

modules/storages/app/common/storages/peripherals/storage_interaction/nextcloud/legacy_upload_link_query.rb
```diff
@@ -27,21 +27,19 @@
 #++
 
 module Storages::Peripherals::StorageInteraction::Nextcloud
-  class LegacyUploadLinkQuery < Storages::Peripherals::StorageInteraction::StorageQuery
-    using Storages::Peripherals::ServiceResultRefinements # use '>>' (bind) operator for ServiceResult
+  class LegacyUploadLinkQuery
```
apfohl: Can be removed.

modules/storages/app/common/storages/peripherals/storage_interaction/nextcloud/propfind_query.rb
```diff
@@ -0,0 +1,115 @@
...
+      # ]
+
+      body = Nokogiri::XML::Builder.new do |xml|
+        xml['d'].propfind(
+          'xmlns:d' => 'DAV:',
+          'xmlns:oc' => 'http://owncloud.org/ns',
+          'xmlns:nc' => 'http://nextcloud.org/ns'
+        ) do
+          xml['d'].prop do
+            # xml['nc'].send('acl-list')
```
apfohl: Could you add a comment, why this is commented out?

modules/storages/app/common/storages/peripherals/storage_interaction/nextcloud/remove_user_from_group_command.rb
```diff
@@ -0,0 +1,81 @@
...
+      )
+
+      case response
+      when Net::HTTPSuccess
+        statuscode = Nokogiri::XML(response.body).xpath('/ocs/meta/statuscode').text
+        case statuscode
+        when "100"
+          ServiceResult.success(message: "User has been removed from group")
+        when "101"
+          ServiceResult.failure(message: "No group specified")
```
apfohl: Also think about using the `Util.error` method here.

modules/storages/app/common/storages/peripherals/storage_interaction/nextcloud/set_permissions_command.rb
```diff
@@ -27,116 +27,77 @@
...
-          xml['nc'].send('acl-mapping-id', username)
-          xml['nc'].send('acl-mask', '31')
-          xml['nc'].send('acl-permissions', assignable_permission)
+      case response
+      when Net::HTTPSuccess
+        doc = Nokogiri::XML(response.body)
+        if doc.xpath("/d:multistatus/d:response/d:propstat[d:status[text() = 'HTTP/1.1 200 OK']]/d:prop/nc:acl-list").present?
+          ServiceResult.success
+        else
+          ServiceResult.failure
```
apfohl: Return a specific error in this case.

modules/storages/app/common/storages/peripherals/storage_requests.rb
```diff
@@ -28,61 +28,50 @@
...
+      define_method(request) do
+        result(clazz(@storage, request))
+      end
     end
 
-    def create_folder_command
-      storage_commands
-        .create_folder_command
-        .map { |command| command.method(:execute).to_proc }
+    def upload_link_query
```
apfohl: Can be removed, because the legacy upload line query is removed.

apfohl (changes requested):

modules/storages/app/common/storages/peripherals/storage_interaction/nextcloud/propfind_query.rb
```diff
@@ -38,44 +66,18 @@ def initialize(storage)
...
-      #   oc:checksums
-      #   oc:size
-      #   nc:has-preview
-      #   nc:rich-workspace
-      #   nc:contained-folder-count
-      #   nc:contained-file-count
-      #   nc:acl-list
-      # ]
-
+    def call(depth:, path:, props: nil)
```
apfohl: `nil` is not a good choice for `props` here, as if no parameter is passed, it would fail in line 77 with `undefined method `each' for nil:NilClass`. I would suggest having no default value for that parameter or having `[]` as the default. I'd prefer having no default though.

modules/storages/app/workers/manage_nextcloud_integration_job.rb
```diff
@@ -29,141 +29,192 @@
...
+    end
+  end
 
-      requests = Storages::Peripherals::StorageRequests.new(storage:)
+  class NextcloudStorageManager
+    def initialize(storage)
+      @storage = storage
+      @nextcloud_system_user = storage.username
+      @group = storage.group
+      @groupfolder = storage.groupfolder
```
apfohl: `@group_folder`

modules/storages/app/workers/manage_nextcloud_integration_job.rb
```diff
@@ -29,141 +29,192 @@
...
+      @nextcloud_usernames_used_in_openproject = Set.new
+      @project_folder_ids_used_in_openproject = Set.new
+      @requests = Storages::Peripherals::StorageRequests.new(storage:)
+    end
 
-      group_users = requests.group_users_query.call(group:).on_failure do |r|
-        raise "group_users_query failed: #{r.inspect}, group: #{group}"
-      end.result
+    def call
+      set_groupfolder_permissions
```
apfohl: `set_group_folder_permissions`

modules/storages/app/workers/manage_nextcloud_integration_job.rb
```diff
@@ -29,141 +29,192 @@
...
+    end
 
-      (group_users - nextcloud_usernames_used_in_openprojects.to_a).each do |user|
-        requests.remove_user_from_group_command.call(user:)
-                .on_failure { |r| raise "remove_user_from_group_command failed: #{r}, user: #{user}" }
+    def file_ids
+      @file_ids ||= folders_props.map { |_path, props| props['fileid'] }
+    end
+
+    def folders_props
```
apfohl: `folder_properties`

modules/storages/app/workers/manage_nextcloud_integration_job.rb
```diff
@@ -29,141 +29,192 @@
...
+      @requests.rename_file_command.call(source:, target:).on_failure do |r|
+        raise "rename_file_command failed(source: #{source} target: #{target}) failed: #{r.inspect}"
       end
+    end
 
-      lost_folder_paths = folders_props
-                            .except("#{groupfolder}/")
-                            .each_with_object([]) do |(path, attrs), array|
-        array.push(path) if project_folder_ids_used_in_openproject.exclude?(attrs['fileid'])
+    def create_folder(path:, project_storage:)
```
apfohl, line 147: It would be nice to not have this method have a side effect. My suggestions is extraction the inner request to a separate method `file_id(path:)` and have three statements in line 105:

1. create folder
2. obtain file ID
3. update file ID in `ProjectStorages`

That would make the methods more pure and don't lie from their name.

apfohl: You could do a `>>` (Bind) chain here:

```ruby
create_folder >> obtain_file_id >> update_project_storages
```

maybe. ;)

modules/storages/app/workers/manage_nextcloud_integration_job.rb
```diff
@@ -29,141 +29,192 @@
...
+    Storages::NextcloudStorage
+      .where("provider_fields->>'has_managed_project_folders' = 'true'")
+      .includes(:oauth_client)
+      .each do |storage|
+      NextcloudStorageManager.new(storage).call
+    end
+  end
 
-      requests = Storages::Peripherals::StorageRequests.new(storage:)
+  class NextcloudStorageManager
```
apfohl: I really like that the main logic is now extracted and could be used somewhere else. I'm not fine with the name though. A `...Manager` implies a longer living object that gets hold of something that needs to be managed. This is not the case here, as this class only contain extracted methods. Basically it's still one function that does a specific thing: Fixing and applying settings to the group folder and users/groups in the storage.

I don't want to inflate this here even more, but later we can extract all single methods as `public` method to a utility class for changing things in the storage with regards to project folders. Than have a small `ProjectFolderRepairer` class that make use of the methods from the utility class. This is then executed by the `Job`. What do you think. It would be nice to do this for the single trigger points.

modules/storages/app/workers/manage_nextcloud_integration_job.rb
```diff
@@ -29,141 +29,192 @@
...
+        users: { @nextcloud_system_user.to_sym => ALL_PERMISSIONS },
+        groups: { @group.to_sym => PERMISSIONS_MAP[:read_files] }
+      }
+      @requests
+        .set_permissions_command
+        .call(path: @groupfolder, permissions:)
+        .on_failure { |r| raise "set_permissions_command(path: #{@groupfolder}, permissions: #{permissions}) failed: #{r.inspect}" }
+    end
+
+    def handle_project_folder(project_storage:)
```
apfohl: `handle_project_folder` is very unspecific. I'm okay with grouping coherent methods together. `add_users_to_group(tokens)` shall not belong here for example, because it has not directly something to do with a project folder.

Maybe it could be called `ensure_project_folder` but without `add_users_to_group`?

modules/storages/app/workers/manage_nextcloud_integration_job.rb
```diff
@@ -29,141 +29,192 @@
...
+      @groupfolder = storage.groupfolder
+      @nextcloud_usernames_used_in_openproject = Set.new
+      @project_folder_ids_used_in_openproject = Set.new
+      @requests = Storages::Peripherals::StorageRequests.new(storage:)
+    end
 
-      group_users = requests.group_users_query.call(group:).on_failure do |r|
-        raise "group_users_query failed: #{r.inspect}, group: #{group}"
-      end.result
+    def call
```
apfohl, line 65: What happens, when call gets executed twice from the same instance?

apfohl (changes requested):

modules/storages/app/common/storages/peripherals/storage_interaction/nextcloud/propfind_query.rb
ba1ash: You're right. I'll remove the default value.

modules/storages/app/workers/manage_nextcloud_integration_job.rb
ba1ash: Then probably it makes sense to rename `Storage#groupfolder` to be `Storage#group_folder` as well. Wdyt?

modules/storages/app/workers/manage_nextcloud_integration_job.rb
ba1ash: The data structure returned by this method looks like:
```ruby
{
   "OpenProject/" => { "fileid" => "349" },
   "OpenProject/Project #2/" => { "fileid" => "381" },
   "OpenProject/Project#1/" => { "fileid" => "773" },
   "OpenProject/Project#2/" => { "fileid" => "398" },
}
```
Let it be `folders_properties` due to a multiple folders inside?

modules/storages/app/workers/manage_nextcloud_integration_job.rb
apfohl: That's true.

modules/storages/app/workers/manage_nextcloud_integration_job.rb
apfohl: That makes sense.

modules/storages/app/workers/manage_nextcloud_integration_job.rb
ba1ash: Nice suggestion! I'll do it this way.

modules/storages/app/workers/manage_nextcloud_integration_job.rb
ba1ash: > I really like that the main logic is now extracted and could be used somewhere else. I'm not fine with the name though. A `...Manager` implies a longer living object that gets hold of something that needs to be managed. This is not the case here, as this class only contain extracted methods. Basically it's still one function that does a specific thing: Fixing and applying settings to the group folder and users/groups in the storage

I do not like the name as well. I did not find a better one and decided to stick with it for a time being. Maybe, `NextcloudStorageSynchonizer` or `NetxcouldStorageRepairer` would work. Wdyt?

> I don't want to inflate this here even more, but later we can extract all single methods as `public` method to a utility class for changing things in the storage with regards to project folders. Than have a small `ProjectFolderRepairer` class that make use of the methods from the utility class. This is then executed by the `Job`. What do you think. It would be nice to do this for the single trigger points.

I am not sure that I completely get the idea. I would stick to a single big class/module and have things in one place till the moment we really "fix" things sparingly as a reaction on a particular event in OP (user has been added to a project => we only trigger appropriate updates in NC).

apfohl: > I am not sure that I completely get the idea. I would stick to a single big class/module and have things in one place till the moment we really "fix" things sparingly as a reaction on a particular event in OP (user has been added to a project => we only trigger appropriate updates in NC).

That's what I meant. Not now, but later for the events. Good.

> I do not like the name as well. I did not find a better one and decided to stick with it for a time being. Maybe, `NextcloudStorageSynchonizer` or `NetxcouldStorageRepairer` would work. Wdyt?

I would go with something like `ProjectFolderPropertiesSynchronization`. A bit long but maybe there is a shorter one.

modules/storages/app/workers/manage_nextcloud_integration_job.rb
```diff
@@ -0,0 +1,220 @@
...
+# along with this program; if not, write to the Free Software
+# Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+#
+# See COPYRIGHT and LICENSE files for more details.
+#++
+
+class ManageNextcloudIntegrationJob < Cron::CronJob
+  using ::Storages::Peripherals::ServiceResultRefinements
+
+  PERMISSIONS_MAP = {
```
wielinde: @ba1ash These constants probably should go into the class below

modules/storages/app/workers/manage_nextcloud_integration_job.rb
```diff
@@ -0,0 +1,220 @@
...
+      @groupfolder = storage.groupfolder
+      @nextcloud_usernames_used_in_openproject = Set.new
+      @project_folder_ids_used_in_openproject = Set.new
+      @requests = Storages::Peripherals::StorageRequests.new(storage:)
+    end
+
+    def call
+      set_groupfolder_permissions
+
+      @storage.projects_storages
```
wielinde, line 75: Maybe we could create a scope for it?

modules/storages/app/workers/manage_nextcloud_integration_job.rb
wielinde: `set_group_folder_root_permissions`

modules/storages/app/workers/manage_nextcloud_integration_job.rb
```diff
@@ -0,0 +1,220 @@
...
+              .where(project_folder_mode: 'automatic')
+              .includes(project: %i[users enabled_modules])
+              .each do |project_storage|
+        handle_project_folder(project_storage:)
+      end
+
+      (group_users - @nextcloud_usernames_used_in_openproject.to_a).each do |user|
```
wielinde: extract to `remove_inactive_users_from_group`

modules/storages/app/workers/manage_nextcloud_integration_job.rb
```diff
@@ -0,0 +1,220 @@
...
+        remove_user_from_group(user)
+      end
+
+      inactive_project_folder_paths.each { |folder| hide_folder(folder) }
```
wielinde: improve the already good readability `hide_inactive_project_folders`

modules/storages/app/workers/manage_nextcloud_integration_job.rb
ba1ash: True, the name is poor. `ensure_project_folder` sounds better for me.
And it makes sense to have 3 steps:
1. Ensure project folder
2. Set project folder permissions
3. Add users to the group. Maybe this step could be done later for all used usernames.

modules/storages/app/workers/manage_nextcloud_integration_job.rb
ba1ash: It will do the same actions for the same storage. So, 'fix' will happen twice sequentially.

modules/storages/app/workers/manage_nextcloud_integration_job.rb
ba1ash: I will try, but naming it could be not easy.

modules/storages/app/workers/manage_nextcloud_integration_job.rb
apfohl: > `set_group_folder_root_permissions`

I don't want to blow this up, but isn't the group folder some kind of root? It's is already very specific imo.

modules/storages/app/workers/manage_nextcloud_integration_job.rb
apfohl: I'm asking about the state (sets) in the class.

modules/storages/app/workers/manage_nextcloud_integration_job.rb
wielinde: I fell into that trap already twice. `set_group_folder_permissions` could also mean to set the permission on all its subfolders, which this method is not doing. At least I read it like it when I saw it as the first line of the `call` action: "Yes, that is everything I want that Job to do." and then I see "Ah, wait, that the subfolder stuff is handled further below".

modules/storages/app/workers/manage_nextcloud_integration_job.rb
apfohl: > I'm asking about the state (sets) in the class.

They are only reset from the initializer. A second `call` would start with a different state.

modules/storages/app/workers/manage_nextcloud_integration_job.rb
ba1ash: > That's what I meant. Not now, but later for the events. Good.

:+1: 


> I would go with something like `ProjectFolderPropertiesSynchronization`. A bit long but maybe there is a shorter one.

I would expect `ProjectFolderPropertiesSynchronization` to operate on one project folder and set its properties, but current class operates in scope of `group_folder` and its one level children:
1. sets permissions for them
2. create/rename/hide project_folders
3. Add/remove users to/from `group`

What about the following signature: `GroupFolderSynchronization.new(storage).call`?

modules/storages/app/workers/manage_nextcloud_integration_job.rb
ba1ash: I get it. You're right it is inconsistent. I'll handle this case. Good catch, thanks!

modules/storages/app/workers/manage_nextcloud_integration_job.rb
apfohl: > What about the following signature: `GroupFolderSynchronization.new(storage).call`?

I would assume that it's syncing files in a group folder.

I would like to keep the "Properties" part. And to that, "group folders" are only one way of implementing the solution. And only for next cloud. The class should operate independently from the type of storage. Therefor it is Nextcloud only at the moment.

But we should get done now. `GroupFolderPropertiesSynchronization`?

ba1ash: Agreed, let it be `GroupFolderPropertiesSynchronization`

modules/storages/app/workers/manage_nextcloud_integration_job.rb
```diff
@@ -131,40 +144,27 @@ def rename_folder(source:, target:)
...
-        .remove_user_from_group_command
-        .call(user:)
-        .on_failure { |r| raise "remove_user_from_group_command(user: #{user}) failed: #{r.inspect}" }
-    end
-
-    def inactive_project_folder_paths
-      folders_props.except("#{@groupfolder}/").each_with_object([]) do |(path, attrs), paths|
-        paths.push(path) if @project_folder_ids_used_in_openproject.exclude?(attrs['fileid'])
+    def save_file_id
+      ->((project_storage, file_id)) do
```
apfohl, line 156: Really good choice to return lambdas here. Good.

modules/storages/app/workers/manage_nextcloud_integration_job.rb
```diff
@@ -48,78 +38,101 @@ def perform
...
 
-      if file_ids.include?(project_folder_id)
-        source = folders_props.find { |_k, v| v['fileid'] == project_folder_id }.first
-        rename_folder(source:, target:) if source != target
+      if project_folder_id.present? && file_ids.include?(project_folder_id)
+        source = folders_properties.find { |_k, v| v['fileid'] == project_folder_id }.first
+        rename_folder(source:, target: project_folder_path) if source != project_folder_path
       else
-        create_folder(path: target, project_storage:)
+        create_folder(path: project_folder_path, project_storage:) >> obtain_file_id >> save_file_id
```
apfohl, line 120: Yeah 🤘

apfohl (approved):
Good job. I have nothing to add.

apfohl:
Last thing is, to fix somehow the rubocop issues.

PR #12673, opened 2023-05-22
Change icons to spot-icon part 2

b12f:
Switch icons to spot-icon in the following places:

* admin pages WP type form configuration
* invite user modal
* work package comment add
* work package details toolbar (split screen on the bottom, watch/unwatch, "more", mark as read)
* work package full view toolbar (on the top, with zen mode, watch/unwatch, "more"

There are also some a11y changes in here especially in the wp type form configuration. A lot of click event listeners were set on `span` or `div` instead of a `button`.

On top, there's small style changes & fixes; 
* a dynamic form without fields would still render with margins, causing weird spacings. It will now be hidden
* WP Type form configuration icons are different sizes to align with the hierarchical nature, and some spacings have improved

Part of https://community.openproject.org/work_packages/47363/activity

### DIFF
--- a/frontend/src/app/features/admin/types/query-group.component.html
+++ b/frontend/src/app/features/admin/types/query-group.component.html
@@ -10,8 +10,8 @@
     </div>
     <div class="type-form-query">
         <span class="type-form-query-group--edit-button" (click)="editQuery.emit()">
-          <op-icon icon-classes="button--icon icon-edit"></op-icon>
-            {{ text.edit_query }}
+          <span class="spot-icon spot-icon_edit"></span>
+          <span>{{ text.edit_query }}</span>
         </span>
     </div>
 </div> <!-- END query group -->
--- a/frontend/src/app/features/admin/types/query-group.component.html
+++ b/frontend/src/app/features/admin/types/query-group.component.html
@@ -1,17 +1,27 @@
 <div class="type-form-conf-group type-form-query-group">
-    <div class="group-head">
-        <span class="group-handle icon-drag-handle"></span>
-        <op-group-edit-in-place
-                [name]="group.name"
-                (onValueChange)="rename($event)"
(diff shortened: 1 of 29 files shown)


marcalcobe (changes requested):
Hi @b12f,

I've detected a couple of places where the icons have problems:
- In the admin pages WP type form configuration the enterprise banner seams to have the icons wrong positioned (I assume this is a separate PR). Also the handle icon for the Inactive fields seems to be bigger than the active ones.
Screenshot 2023-05-23 at 14 43 57
Screenshot 2023-05-23 at 14 43 45

- In the invite user modal mI only see a back icon in the button, is there any other icon in the process?

- In the work package full view toolbar (on the top, with zen mode, watch/unwatch, "more") I see two different sizes of dropdown arrows (create vs others buttons), the filter, table and more icons seem to still not be spot.icon and the info and zen mode buttons are not square anymore as there is an extra left/right padding.
Screenshot 2023-05-23 at 14 54 21

frontend/src/app/features/admin/types/query-group.component.html
```diff
@@ -1,17 +1,27 @@
...
+      <span class="spot-icon spot-icon_1 spot-icon_close"></span>
+    </button>
+  </div>
+  <div class="type-form-query">
+    <button
+      class="clear-button type-form-query-group--edit-button"
+      type="button"
+      (click)="editQuery.emit()"
+    >
+      <span class="spot-icon spot-icon_1_25 spot-icon_edit spot-icon_inline"></span>
```
Kharonus, line 23: so, you use `spot-icon_inline` for buttons, that are not `display: flex`, correct?

b12f: Specifically if we're in an inline context

PR #12674, opened 2023-05-22
[#48250] implement global index page for boards

klaustopher:
Implements the global index page for boards.

See https://community.openproject.org/projects/openproject/work_packages/48250

### DIFF
--- a/modules/boards/app/views/boards/boards/overview.html.erb
+++ b/modules/boards/app/views/boards/boards/overview.html.erb
@@ -29,21 +29,6 @@ See COPYRIGHT and LICENSE files for more details.
 
 <% html_title(t('boards.label_boards')) -%>
 
-<%= toolbar title: t('boards.label_boards') do %>
-  <% if authorize_for(:boards, :new) %>
-    <li class="toolbar-item">
-      <a href="<%= "foo" %>"
-         id="add-board-button"
-         title="<%= I18n.t(:label_meeting_new) %>"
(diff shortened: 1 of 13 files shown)


modules/boards/app/views/boards/boards/overview.html.erb
```diff
@@ -0,0 +1,51 @@
...
+         id="add-board-button"
+         title="<%= I18n.t(:label_meeting_new) %>"
+         arial-label="<%= I18n.t(:label_meeting_new) %>"
+         class="button -alt-highlight">
+        <%= op_icon('button--icon icon-add') %>
+        <span class="button--text"><%= t(:label_meeting) %></span>
+      </a>
+    </li>
+  <% end %>
+<% end %>
```
klaustopher: The modal for creating a new board currently does not include any selection of the project. Should we add a WP (like we have for global meetings create) to implement a new modal view that will include a product selection?

//cc @oliverguenther

klaustopher: Never mind, I saw that https://community.openproject.org/projects/openproject/work_packages/47989/activity exists

oliverguenther (changes requested):
Only a small thing I noticed: The global page is missing a heading. Apart from that, LGTM :+1:

klaustopher:
Nice catch. Title added

oliverguenther (approved):

PR #12683, opened 2023-05-23
Fix migration for non-writable blank settings

cbliard:
https://community.openproject.org/wp/48125

### DIFF
--- a/db/migrate/20220428071221_restore_defaults_on_empty_settings.rb
+++ b/db/migrate/20220428071221_restore_defaults_on_empty_settings.rb
@@ -12,7 +12,11 @@ class RestoreDefaultsOnEmptySettings < ActiveRecord::Migration[6.1]
 
       next if definition.value == ''
 
-      setting.update_attribute(:value, definition.value)
+      if definition.writable?
+        setting.update_attribute(:value, definition.value)
+      else
+        setting.destroy
+      end
     end
   end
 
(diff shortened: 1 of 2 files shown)


cbliard:
Once again I forgot to make the PR against 12.5. I'll rebase it.

cbliard:
About the implementation, I hesitated to just delete the setting from the db so that the default value is picked (instead of trying to save the default value in the database, which is the current implementation).

machisuji:
> About the implementation, I hesitated to just delete the setting from the db so that the default value is picked (instead of trying to save the default value in the database, which is the current implementation).

I would vote for deleting the setting. Since they are not writable they are not supposed to be in the DB to begin with. We could just check in the definition if it's writable, and if it isn't delete the setting rather than writing the default value.

cbliard:
> > About the implementation, I hesitated to just delete the setting from the db so that the default value is picked (instead of trying to save the default value in the database, which is the current implementation).
> 
> I would vote for deleting the setting. Since they are not writable they are not supposed to be in the DB to begin with. We could just check in the definition if it's writable, and if it isn't delete the setting rather than writing the default value.

I'll do that then. Thanks @machisuji

machisuji (approved):
👍

machisuji:
Thanks for the fix. Looks good to me!

PR #12695, opened 2023-05-24
Update the Timestamp duration? logic to handle all valid duration string

dombesz:
An  amendment to the PR #12686 to fix the `Timestamp#duration?` method by allowing all valid duration string.

### DIFF
--- a/app/models/timestamp.rb
+++ b/app/models/timestamp.rb
@@ -34,6 +34,8 @@ class Timestamp
   class Exception < StandardError; end
 
   class TimestampParser
+    DURATION_REGEX = /[+-]?P/ # ISO8601 "Period"
+
     DATE_KEYWORD_REGEX =
       %r{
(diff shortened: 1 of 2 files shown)


app/models/timestamp.rb
```diff
@@ -34,6 +34,8 @@ class Timestamp
   class Exception < StandardError; end
 
   class TimestampParser
+    DURATION_REGEX = /[+-]?P/ # ISO8601 "Period"
```
aaron-contreras, line 37: I feel this constant could make the comment obsolete if it's named something along the lines of

- `ISO8601_DURATION_REGEX`
or
- `ISO8601_PERIOD_REGEX`

It'd read nicely to me :)

aaron-contreras (approved):
Running the specs with `--format documentation` reads beautifully.

PR #12699, opened 2023-05-24
[#48248] Work package in query without timestamps have cached timestamps

dombesz:
https://community.openproject.org/work_packages/48248


ulferts (approved):
:+1:

PR #12704, opened 2023-05-25
[#45745] Group timezones with same identifier

klaustopher:
WP: https://community.openproject.org/notifications/details/45745/activity

---

Removes the special code that only allows selecting the canonical identifier for timezones. For zones that are linked to another (i.e. `Europe/Copenhagen` -> `Europe/Berlin`), this would cause the user selecting Copenhagen, but on the next page view, they see Berlin. Also, the old code caused several options in the select to use the same option value, always leading to a wrongly shown currently selected timezone:

Bildschirmfoto 2023-05-25 um 08 11 14

### DIFF
--- a/app/components/settings/time_zone_setting_component.rb
+++ b/app/components/settings/time_zone_setting_component.rb
@@ -72,9 +72,18 @@ module Settings
     def time_zone_entries
       UserPreferences::UpdateContract
         .assignable_time_zones
-        .map do |tz|
-        [tz.to_s, tz.tzinfo.canonical_identifier]
-      end
+        .group_by { |tz| tz.tzinfo.canonical_zone }
+        .map { |canonical_zone, included_zones| time_zone_option(canonical_zone, included_zones) }
+    end
+
+    private
+
+    def time_zone_option(canonical_zone, zones)
+      zone_names = zones.map(&:name).join(', ')
+      [
+        "(UTC#{ActiveSupport::TimeZone.seconds_to_utc_offset(canonical_zone.base_utc_offset)}) #{zone_names}",
+        canonical_zone.identifier
+      ]
     end
   end
 end
--- a/spec/components/settings/time_zone_setting_component_spec.rb
+++ b/spec/components/settings/time_zone_setting_component_spec.rb
@@ -1,0 +1,44 @@
+# frozen_string_literal: true
+
+#-- copyright
+# OpenProject is an open source project management software.
+# Copyright (C) 2012-2023 the OpenProject GmbH
+#
+# This program is free software; you can redistribute it and/or
+# modify it under the terms of the GNU General Public License version 3.
+#
+# OpenProject is a fork of ChiliProject, which is a fork of Redmine. The copyright follows:
+# Copyright (C) 2006-2013 Jean-Philippe Lang
+# Copyright (C) 2010-2013 the ChiliProject Team
+#
+# This program is free software; you can redistribute it and/or
+# modify it under the terms of the GNU General Public License
+# as published by the Free Software Foundation; either version 2
+# of the License, or (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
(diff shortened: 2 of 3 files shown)


oliverguenther:
Please rebase this against `release/12.5` to target the next patch release

klaustopher:
Will do. Still playing around and looking at the history why this was introduced :)

oliverguenther (changes requested):
The selected option appears to be wrong for existing values:

My user has this time_zone pref:

```ruby
=> #<UserPreference:0x00000001186bf148
 id: 6,
 user_id: 6,
 settings:
  {"time_zone"=>"Asia/Yangon", ... }
```
But with that time zone (canonically mapped from Asia/Rangoon +6:30, the is now shown as unselected (value '')
Bildschirm­foto 2023-05-25 um 21 31 22

klaustopher:
Will check on tuesday. Interesting: The timezone Asia/Yangon was introduced by [renaming `Asia/Rangoon` in 2016](https://mm.icann.org/pipermail/tz-announce/2016-September/000041.html)

It also exists when querying for it:

```ruby
>> ActiveSupport::TimeZone['Asia/Yangon']
=>
#<ActiveSupport::TimeZone:0x000000010857ff90
 @name="Asia/Yangon",
 @tzinfo=#<TZInfo::DataTimezone: Asia/Yangon>,
 @utc_offset=nil>
```

but in `AS::TimeZone.all` it is not included:

```ruby
>> ActiveSupport::TimeZone.all.select { |tz| tz.tzinfo.identifier == 'Asia/Yangon' }
=> []

>> ActiveSupport::TimeZone.all.select { |tz| tz.tzinfo.identifier == 'Asia/Rangoon' }
=>
[#<ActiveSupport::TimeZone:0x0000000108a74640
  @name="Rangoon",
  @tzinfo=#<TZInfo::DataTimezone: Asia/Rangoon>,
  @utc_offset=nil>]
```

That seems very strange ... Will take a look

klaustopher:
In the TZInfo DB it is correctly implemented that Rangoon is an alias for Yangon: https://github.com/tzinfo/tzinfo-data/blob/master/lib/tzinfo/data/definitions/Asia/Rangoon.rb
```ruby
>> TZInfo::Timezone.all.select { |tz| tz.name == "Asia/Yangon" }
=> [#<TZInfo::TimezoneProxy: Asia/Yangon>]
>> TZInfo::Timezone.all.select { |tz| tz.name == "Asia/Rangoon" }
=> [#<TZInfo::TimezoneProxy: Asia/Rangoon>]
```

Maybe it makes sense, we take `TZInfo::Timezone` as the source of truth.

All entries in TZInfo can be thrown into `AS::TimeZone` and they produce the right data:

```ruby
>> TZInfo::Timezone.all.select { |tzinfo| ActiveSupport::TimeZone[tzinfo.name].nil? }
=> []
```

I'll probably file a bug with Rails. There are **many** timezones missing in `AS::TimeZone.all`:

```ruby
TZInfo::Timezone.all.map { |x| x.send(:real_timezone) } - ActiveSupport::TimeZone.all.map(&:tzinfo)
```

returns 462 items

klaustopher:
Seems like rails has intentionally only added 134 out of the ~600 timezone identifiers into `AS::TimeZone` (see comment in https://github.com/rails/rails/blob/main/activesupport/lib/active_support/values/time_zone.rb). That's why `Rangoon` is still in there and and `Yangon` is not.

I think we have 2 options forward:

1. Use `TZInfo::Timezone.all` to list all timezones that are available.
   That would result in a very long list of options and might be confusing for the users. The [TZInfo database itself suggest](https://github.com/tzinfo/tzinfo#time-zone-selection) to not do that, but instead let the user choose a country and then based on that country selection select the timezone. I like that approach as it removes a lot of the complexity (including the long select list). But IMHO this is too much of a refactoring for this bug fix here.
2. Implement a bit more grouping logic, so that the options that link to the same canonical zone get aggregated in the list. This way we would have an option like `UTC+1:00 Berlin, Copenhagen, Brussels, ...`

I'm sort of tending towards the second option.

klaustopher:
Implemented this way now: Options are grouped together when they use the same identifier.

Bildschirmfoto 2023-05-30 um 09 22 14

oliverguenther (approved):

PR #12710, opened 2023-05-25
[#48275] Default work packages PDF export styling

as-op:
https://community.openproject.org/work_packages/48275


as-op:
Hi @klaustopher, may I kindly ask you for a follow-up review? This PR extracts the hardcoded styles into a yml-file and adds schema validation for it.
The default styling will probably stay in the yml, while user created styling will be stored in and retrieved from the db.

app/models/work_package/pdf_export/common.rb
```diff
@@ -111,15 +111,15 @@ def get_subject_column_cell(work_package, value)
...
+    "<color rgb='#{styles.link_color}'><link href='#{href}'>#{caption}</link></color>"
   end
 
   def make_link_anchor(anchor, caption)
     "<link anchor=\"#{anchor}\">#{caption}</link>"
   end
 
-  def align_to_left_position(text, align, style)
-    text_width = pdf.width_of(text, style)
+  def align_to_left_position(text, align, text_style)
```
klaustopher, line 121: 👍 for better naming of variables

app/models/work_package/pdf_export/overview_table.rb
```diff
@@ -143,17 +143,17 @@ def build_nr_cell(work_package, id_wp_meta_map)
   end
 
   def build_subject_cell(content, work_package, id_wp_meta_map)
-    opts = overview_table_cell_padding_style
-    padding_left = opts[:padding_left]
+    opts = styles.overview_table_cell
+    padding_left = opts[:padding_left] || 0
```
klaustopher: :memo: When you want to provide a default value for a hash lookup you can also use 

```suggestion
    padding_left = opts.fetch(:padding_left, 0)
```

To some that looks nicer. No need to change, just pointing it out

app/models/work_package/pdf_export/standard.yml
```diff
@@ -0,0 +1,176 @@
...
+      no_border_top: true
+    image:
+      align: "left"
+    codeblock:
+      background_color: "F5F5F5"
+      color: "880000"
+      padding: "3mm"
+      size: 8
+      margin_top: "2mm"
+      margin_bottom: "2mm"
```
klaustopher: ❓ Is there a reason we are adding units here and no tfor the other margins?

klaustopher (approved):
LGTM!

app/models/work_package/pdf_export/standard.yml
as-op: No good reason other than it's possible to use units. ^^
I unified the usage to have non specified. (PostscriptPoint is the default)

PR #12721, opened 2023-05-29
Implementation/48289 add automatic project folder management settings for storages

akabiru:
https://community.openproject.org/projects/openproject/work_packages/48289

Let OpenProject create folders per project automatically. This is recommended as it ensures that every team member always has the correct access permissions.

- [x] Define UI blueprint
  - [x] Update from checkbox to `spot-switch`
- [x] Redirect from oauth client form (❓ _Feature flag_)
  - [x] _Optional_ scope configuration in nested controller
- [x] _FIXME:_ _update html title_
- [x] Persistence rules
  - [x] Default values: toggle is on, require app password
  - [x] username should always be "OpenProject" if storage provider type is nextcloud
  - [x] Clear `application`_{username,password}` on `automatically_managed` `false`
  - [x] Only show managed project folders form when there were no pre-existing values (during edit of oauth credentials)
- [x] Display errors on form validation failure
- [x] Add managed project folders index & edit views
   -  🚧 Add `applicationPassword` validation ([Separate WP](https://community.openproject.org/projects/openproject/boards/3944/details/48569/overview))
   - [x] On edit, show password filled status as masked placeholder
- [x] Add API
   - [x] Add new storage representer parsing spec
   - [x] Ensure `applicationPassword` is not rendered in storage representer rendering
   - [x] Add request specs
- [x] Move models logic for nextcloud to `Storages::NextcloudStorge`
<details> <summary> Screenshots 🔽 </summary>

Screenshot 2023-06-14 at 11 58 42 AM
Screenshot 2023-06-14 at 11 56 31 AM
Screenshot 2023-06-14 at 11 55 55 AM
Screenshot 2023-06-14 at 11 55 44 AM
Screenshot 2023-06-14 at 11 55 16 AM

</details>


wielinde (changes requested):
@akabiru I haven't yet fully worked my way through the code. However, I already give you feedback on what I have observed:
- On my dev machine, I had storages configured before applying your branch. They do not yet know anything about app passwords and modes etc. When I edit on of those, the button for adding the missing configuration is "+ Create". We didn't specify that case in Figma. I think we should call int "Configure ..." or something link that (@marcalcobe what do you think?)
- Then, after pressing the button, I would have expected that the switch was set to true (as that is the default for not present configurations). However, for me it shows itself as off.

modules/storages/app/views/storages/admin/storages/show_managed_project_folders.html.erb
```diff
@@ -0,0 +1,102 @@
...
+  t("storages.label_managed_project_folders_breadcrumb")
+] %>
+<%= toolbar title: t("storages.page_titles.managed_project_folders.title"),
+            subtitle: t("storages.page_titles.managed_project_folders.subtitle")
+%>
+
+<%= error_messages_for_contract @storage, @errors %>
+
+<%= labelled_tabular_form_for @storage, url: admin_settings_storage_managed_project_folders_path(@storage),
+                                        method: :patch, data: {
```
wielinde: line break behind the comma, right?

modules/storages/app/models/storages/storage.rb
wielinde, line 1: Probably most of the changes are Nextcloud specific and belong to `NextcloudStorage`

modules/storages/spec/lib/api/v3/storages/storages_representer_parsing_spec.rb
```diff
@@ -0,0 +1,103 @@
...
+        }
+      }
+    end
+
+    context 'with basic attributes' do
+      it 'is parsed correctly' do
+        expect(parsed).to have_attributes(name: 'Nextcloud Local', host: storage.host,
+                                          provider_type: "Storages::NextcloudStorage")
+
+        aggregate_failures 'does not parse applicationPassword' do
```
wielinde:
```suggestion
        aggregate_failures 'does not parse hasApplicationPassword' do
```

modules/storages/spec/lib/api/v3/storages/storages_representer_parsing_spec.rb
```diff
@@ -0,0 +1,103 @@
...
+        end
+
+        it 'is parsed as automatic folder management disabled' do
+          expect(parsed).to have_attributes(automatically_managed: false, application_password: nil)
+        end
+      end
+
+      context 'with applicationPassword blank' do
+        let(:parsed_hash) do
+          super().merge(
```
wielinde, line 92:
```suggestion
          super.merge(
```

modules/storages/app/controllers/storages/admin/managed_project_folders_controller.rb
```diff
@@ -0,0 +1,138 @@
...
+class Storages::Admin::ManagedProjectFoldersController < ApplicationController
+  # See https://guides.rubyonrails.org/layouts_and_rendering.html for reference on layout
+  layout 'admin'
+
+  # Before executing any action below: Make sure the current user is an admin
+  # and set the @<controller_name> variable to the object referenced in the URL.
+  before_action :require_admin
+
+  # specify which model #find_model_object should look up
+  model_object Storages::Storage
```
wielinde:
```suggestion
  model_object Storages::NextcloudStorage
```

modules/storages/app/controllers/storages/admin/managed_project_folders_controller.rb
```diff
@@ -0,0 +1,138 @@
...
+    # Set default parameters using a "service".
+    # See also: storages/services/storages/storages/set_attributes_services.rb
+    # That service inherits from ::BaseServices::SetAttributes
+    @storage = ::Storages::Storages::SetProviderFieldsAttributesService
+                .new(user: current_user,
+                     model: @object,
+                     contract_class: ::Storages::Storages::BaseContract)
+                .call
+                .result
+    render '/storages/admin/storages/show_managed_project_folders'
```
wielinde:
```suggestion
    render '/storages/admin/storages/edit_managed_project_folders'
```

modules/storages/app/controllers/storages/admin/managed_project_folders_controller.rb
```diff
@@ -0,0 +1,138 @@
...
+  end
+
+  # Renders an edit page (allowing the user to change automatically_managed bool and application_password).
+  # Used by: The StoragesController#edit, when user wants to update application credentials.
+  # Called by: Global app/config/routes.rb to serve Web page
+  def edit
+    render '/storages/admin/storages/show_managed_project_folders'
```
wielinde:
```suggestion
    render '/storages/admin/storages/edit_managed_project_folders'
```

modules/storages/app/controllers/storages/admin/managed_project_folders_controller.rb
```diff
@@ -0,0 +1,138 @@
...
+                       .new(user: current_user,
+                            model: @storage)
+                       .call(permitted_storage_params_with_defaults)
+
+    if service_result.success?
+      flash[:notice] = I18n.t(:notice_successful_update)
+      redirect_to admin_settings_storage_path(@storage)
+    else
+      @errors = service_result.errors
+      render '/storages/admin/storages/show_managed_project_folders'
```
wielinde:
```suggestion
      render '/storages/admin/storages/edit_managed_project_folders'
```

wielinde (changes requested):
Looks all very good. Find minor naming comments inline.

One slightly bigger topic is moving Nextcloud specifics from the `Storage` model to `NextcloudStorage` model.

modules/storages/spec/lib/api/v3/storages/storages_representer_parsing_spec.rb
akabiru: Just found out implicit argument passing is not supported in meta context 🎓 

```ruby
super.merge(
         "applicationPassword" => ''
       )

     RuntimeError:
       implicit argument passing of super from method defined by define_method() is not supported. Specify all arguments explicitly.
```

modules/storages/app/models/storages/storage.rb
```diff
@@ -94,4 +94,8 @@ def self.shorten_provider_type(provider_type)
   def short_provider_type
     @short_provider_type ||= self.class.shorten_provider_type(provider_type)
   end
+
+  def provider_type_nextcloud?
+    provider_type == ::Storages::Storage::PROVIDER_TYPE_NEXTCLOUD
+  end
```
akabiru, line 100: @wielinde retaining this method in the parent `Storage` model as the use case applies in this context. E.g. in `SetAttributesService` we want to `#set_nextcloud_application_credentials_defaults` on the condition that the storage is a `provider_type_nextcloud?` ; as the attribute service is  agnostic of the different types of storages.

modules/storages/app/models/storages/nextcloud_storage.rb
```diff
@@ -27,34 +27,23 @@
...
+    username: 'OpenProject'
+  }.freeze
 
-  alias_method :has_managed_project_folders?, :has_managed_project_folders
+  store_attribute :provider_fields, :automatically_managed, :boolean
+  store_attribute :provider_fields, :username, :string, default: PROVIDER_FIELDS_DEFAULTS[:username]
+  store_attribute :provider_fields, :password, :string
+  store_attribute :provider_fields, :has_managed_project_folders, :boolean
+  store_attribute :provider_fields, :group, :string, default: PROVIDER_FIELDS_DEFAULTS[:username]
+  store_attribute :provider_fields, :group_folder, :string, default: PROVIDER_FIELDS_DEFAULTS[:username]
```
akabiru: @ba1ash Updated to use stored_attributes- looks really neat! ✨

ba1ash: Yep :+1:

modules/storages/app/controllers/storages/admin/oauth_clients_controller.rb
```diff
@@ -55,14 +55,21 @@ def new
...
-      redirect_to admin_settings_storage_path(@storage)
+      if OpenProject::FeatureDecisions.automatically_managed_project_folders_active?
+        if @storage.automatic_management_unspecified?
+          redirect_to new_admin_settings_storage_automatically_managed_project_folders_path(@storage)
+        else
+          redirect_to admin_settings_storage_path(@storage)
+        end
+      else
+        redirect_to admin_settings_storage_path(@storage)
+      end
```
mereghost: 2 of the branches end up in the same place. Do we need the nested `if`s for some reason or can we simply join then?
```suggestion
      if OpenProject::FeatureDecisions.automatically_managed_project_folders_active? && @storage.automatic_management_unspecified?
          redirect_to new_admin_settings_storage_automatically_managed_project_folders_path(@storage)
      else
        redirect_to admin_settings_storage_path(@storage)
      end
```

modules/storages/app/services/storages/storages/set_attributes_service.rb
```diff
@@ -54,5 +76,9 @@ def derive_default_storage_name
 
       "#{prefix} #{last_id + 1}"
     end
+
+    def nextcloud_storage?
+      storage.provider_type_nextcloud? && storage.is_a?(Storages::NextcloudStorage)
```
mereghost: Noob question: Is there a way that the second can be true if the first isn't? 😕

mereghost (commented):
I don't have enough context on the feature/architecture but code wise seems good.

Just a couple of comments/questions. =)

modules/storages/app/services/storages/storages/set_attributes_service.rb
akabiru: Good eye, I think we can just do with the latter 👍🏾

modules/storages/app/controllers/storages/admin/oauth_clients_controller.rb
akabiru: ~Given one branch is wrapped in a feature flag, that is bound to be removed- I found it more readable in it's current form and simpler to understand- albeit minimal duplication.~

Scratch that it's actually more compact- good looking out @mereghost ! 👍🏾

frontend/src/app/spot/components/switch/switch.component.html
```diff
@@ -5,6 +5,7 @@
   [disabled]="disabled"
   [tabindex]="tabindex"
   [ngModel]="checked"
+  [value]="checked"
```
Kharonus, line 8: why is that needed? `[ngModel]` is already bound to `checked`.

modules/storages/app/views/storages/admin/storages/automatically_managed_project_folders/edit.html.erb
```diff
@@ -0,0 +1,102 @@
...
+                                          'storages--automatically-managed-project-folders-form-is-automatically-managed-value': @storage.automatically_managed?,
+                                        } do |f| -%>
+  <div class="form--field -required">
+    <%= styled_label_tag :automatically_managed, class: "-bold -flex" do %>
+      <%= t(:"storages.label_managed_project_folders.automatically_managed_folders") %>
+    <% end %>
+
+    <div class="form--field-container -horizontal">
+      <label class="form--label -flex" data-storages--automatically-managed-project-folders-form-target="automaticallyManagedSwitch">
+        <%= angular_component_tag 'spot-switch',
```
Kharonus, line 55: We should put the stimulus target on the `spot-switch` itself, not the label wrapping it. Then we can simplify the stimulus controller.

modules/storages/lib/api/v3/storages/storage_representer.rb
```diff
@@ -81,6 +81,28 @@ def initialize(model, current_user:, embed_links: nil)
...
+               else
+                 represented.automatically_managed = false
+               end
+             }
+
+    property :hasApplicationPassword,
+             skip_parse: true,
+             skip_render: ->(represented:, **) { !represented.provider_type_nextcloud? },
+             getter: ->(represented:, **) {
+               break unless represented.provider_type_nextcloud?
```
Kharonus, line 100: is this break needed, if you already define `skip_render` with the same condition?

modules/storages/lib/open_project/storages/engine.rb
```diff
@@ -44,6 +44,7 @@ class Engine < ::Rails::Engine
       OpenProject::FeatureDecisions.add :storage_file_picking_select_all
       OpenProject::FeatureDecisions.add :storage_project_folders
       OpenProject::FeatureDecisions.add :managed_project_folders
+      OpenProject::FeatureDecisions.add :automatically_managed_project_folders
```
Kharonus, line 47: this is duplicate, isn't it? `:managed_project_folders` was already defined for the scope of anything that goes for the automatically managed folders.

frontend/src/stimulus/controllers/dynamic/storages/automatically-managed-project-folders-form.controller.ts
```diff
@@ -0,0 +1,88 @@
...
+  };
+
+  declare readonly automaticallyManagedSwitchTarget:HTMLElement;
+  declare readonly applicationPasswordInputTarget:HTMLElement;
+  declare readonly hasApplicationPasswordInputTarget:boolean;
+
+  declare isAutomaticallyManagedValue:boolean;
+
+  spotSwitchComponent!:HTMLElement;
+  boundUpdateDisplay!:(evt:InputEvent) => (void | boolean);
```
Kharonus: I'd like to suggest some structural changes in the stimulus controller, that hopefully simplifies the code here:

- change signature `updateDisplay` to 

```typescript
public updateDisplay(_evt:InputEvent):void
```

- use it directly via an stimulus action on the spot switch component (if possible, needs testing)

if you want to check this together, I would gladly have a look at it. We also can do this in a follow up step.

But overall I think, this controller can be reduced by more then half.

Kharonus (changes requested):
I have some small remarks regarding the stimulus parts of the code.

modules/storages/app/views/storages/admin/storages/automatically_managed_project_folders/edit.html.erb
akabiru: I'm not sure the `angular_component_tag` helper supports passing data attributes atm

modules/storages/lib/api/v3/storages/storage_representer.rb
akabiru: We need to guard against non-nextcloud models which do not define nextcloud provider_fields methods such as `#automatically_managed?` which is only defined in `Storages::NextcloudStorage`.

AFAICT from the unit tests `skip_render` is not exclusive of the `getter`. It seems both expressions are still evaluated; regardless of `skip_render` evaluating to `true`.

See failing tests after removing the break guard clause

```ruby
getter: ->(represented:, **) {
              #  break unless represented.provider_type_nextcloud?

               represented.automatically_managed?
             },
#
# $ docker compose exec backend-test bundle exec rspec modules/storages/spec/lib/api/v3/storages
  1) API::V3::Storages::StorageRepresenter rendering properties Automatically managed project folders with a non-Nextcloud storage does not include the property hasApplicationPassword
     Failure/Error: subject(:generated) { representer.to_json }

     NoMethodError:
       undefined method `automatically_managed?' for #<Storages::Storage id: nil, provider_type: "unknown", name: "Storage 18", host: "https://host18.example.com", creator_id: nil, created_at: nil, updated_at: nil, provider_fields: {}>
```

modules/storages/lib/open_project/storages/engine.rb
akabiru: IIRC we discussed and agreed to define a dedicated flag as this is in the global administration. `managed_project_folders` is for Project settings folder mode activation. Perhaps we could have named it better to be more clear. cc @wielinde

frontend/src/app/spot/components/switch/switch.component.html
akabiru: In this case, the component is not used in a pure angular context- the data is submitted in the context of a Rails view (HTML) form. In order to post the form data in vanilla HTML, we need the value attribute defined.

modules/storages/app/models/storages/nextcloud_storage.rb
```diff
@@ -27,22 +27,24 @@
...
+    automatically_managed: true,
+    username: 'OpenProject'
+  }.freeze
 
-  alias_method :has_managed_project_folders?, :has_managed_project_folders
+  store_attribute :provider_fields, :automatically_managed, :boolean
+  store_attribute :provider_fields, :username, :string
+  store_attribute :provider_fields, :password, :string
+  store_attribute :provider_fields, :group, :string, default: PROVIDER_FIELDS_DEFAULTS[:username]
+  store_attribute :provider_fields, :group_folder, :string, default: PROVIDER_FIELDS_DEFAULTS[:username]
```
ba1ash: Newly generated accessors behaves differently. A specified default value is returned only when the record is new. So, we have to decide how to handle the case, for instance, with existing records where `group` and `group_folder` have not been saved to the database. And we should make sure that when we create new records using the form `group` and `group_folder` are saved to the database or use something like previous implementation with the following logic: `provider_fields['group'] || ''OpenProject'`

akabiru: Good catch, reinstating a runtime default as you proposed sounds safer. It seems store_attribute is mainly useful for typecasting.

akabiru: Added https://github.com/opf/openproject/pull/12721/commits/a78ca272136d2a6cadadc02a3de6ed47d491f1dd

modules/storages/app/services/storages/storages/set_attributes_service.rb
```diff
@@ -36,12 +36,31 @@ def set_default_attributes(_params)
       storage.name ||= derive_default_storage_name
     end
 
+    def set_default_provider_fields(_params)
+      set_nextcloud_application_credentials_defaults if nextcloud_storage?
+    end
+
```
ba1ash:
```suggestion
```

modules/storages/app/services/storages/storages/set_attributes_service.rb
```diff
@@ -36,12 +36,31 @@ def set_default_attributes(_params)
...
     private
 
+    def set_attributes(params)
+      super(params)
+      set_default_provider_fields(params)
```
ba1ash:
```suggestion
      remove_nextcloud_credentials if nextcloud_storage?
```

modules/storages/app/services/storages/storages/set_attributes_service.rb
```diff
@@ -36,12 +36,31 @@ def set_default_attributes(_params)
...
+    def set_nextcloud_application_credentials_defaults
+      # Do not overwrite if has never been set.
+      # E.g. when setting up a new storage for the first time, passthrough, credentials are set in a later stage.
+      return if storage.automatic_management_unspecified?
+
+      unless storage.automatically_managed?
+        %w[username password].each { |field| storage.provider_fields.delete(field) }
+      end
+    end
+
```
ba1ash:
```suggestion
    def remove_nextcloud_credentials
      # Do not overwrite if has never been set.
      # E.g. when setting up a new storage for the first time, passthrough, credentials are set in a later stage.
      return if storage.automatic_management_unspecified?

      unless storage.automatically_managed?
        %w[username password].each { |field| storage.provider_fields.delete(field) }
      end
    end

```

ba1ash (approved):

PR #12727, opened 2023-05-30
[#48271] Loop through query.name into the QueryExporter

as-op:
https://community.openproject.org/work_packages/48271


ulferts:
@as-op it seemed to me that you had to work around the name validation. Initially I was going to suggest introducing a constant to be able to reference the placeholder name value in various files. But now I simply moved the validation to our contracts so when only filtering, a name value no longer needs to be set. The value assignment in the `retrieve_query` is then removed by my PR (see #12736).

In the end, this should simplify your code as well but I think we should now wait for my PR to be merged and your PR to be adapted afterwards.

ulferts:
@as-op the mentioned PR has been merged so you can go ahead adapting this one here.

as-op:
@ulferts adapted! Thanks!

ulferts (changes requested):
The name is passed into the query object and thus the export should show the correct title. However, the title passed is oftentimes still "Work packages" when it should e.g. be "Overdue". 

I could easily recreate that behaviour by going to the work package list (which will show "All open") and then clicking on one of the menu items to the left.

I don't think that the problem originates in the code changes of this PR. I surmise that the problem existed before. E.g. when having navigated to "Overdue" and then reloading the page, the title will switch to "Work packages".

as-op:
Hi @ulferts, this bug is connected to the PDF export, but also exists for all "persistent queries". The name for those are collected by an angular state service, which is not up-to-date in all cases. E.g. I have "Work packages" also as the title on the page and even in the headline after a page reload, so the export does not have the right title, too. I will talk to the frontend team about this.

PR #12732, opened 2023-05-31
Use default language setting in docker installation for seeding

cbliard:
https://community.openproject.org/wp/47355

### DIFF
--- a/spec/lib/redmine/i18n_spec.rb
+++ b/spec/lib/redmine/i18n_spec.rb
@@ -78,11 +78,11 @@ module OpenProject
     describe 'all_languages' do
       # Those are the two languages we support
       it 'includes en' do
-        expect(all_languages).to include(:en)
+        expect(all_languages).to include('en')
       end
 
       it 'includes de' do
-        expect(all_languages).to include(:de)
+        expect(all_languages).to include('de')
       end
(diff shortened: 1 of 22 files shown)


spec/lib/redmine/i18n_spec.rb
```diff
@@ -91,49 +91,57 @@ module OpenProject
...
-        allow(Setting).to receive(:available_languages).and_return([:en])
+        with_settings(available_languages: ['en'])
 
-        expect(valid_languages).to eql [:en]
+        expect(valid_languages).to eq ['en']
       end
 
       it 'allows only languages that exist' do
-        allow(Setting).to receive(:available_languages).and_return([:'123'])
+        with_settings(available_languages: ['en', 'de', 'klingon'])
```
klaustopher, line 107: Fun fact: Klingon is listed in [ISO 639-2 and has an official language identifier: `tlh`](https://en.wikipedia.org/wiki/Codes_for_constructed_languages)

cbliard: :laughing: Haha. I should have used a dummier string then. Thanks for sharing.

machisuji (approved):
👍

PR #12738, opened 2023-05-31
Enable RSpec zero monkey patching mode

cbliard:
Avoid monkey patching global namespace by RSpec. 

See https://github.com/rspec/rspec-core/blob/main/features/configuration/zero_monkey_patching_mode.feature for details.


klaustopher:
Nice one! Personally I am a fan to also remove the global `create`/`build`/... methods from `FactoryBot` and use `FactoryBot.create` instead.

cbliard:
@klaustopher that's funny, because we did the opposite 1 year ago in https://github.com/opf/openproject/pull/10090

Personally, I prefer without the `FactoryBot.` prefix. This prefix feels just like noise when reading a test, and it makes the lines longer and harder to read.

klaustopher:
Yeah, as I said, just a personal preference. I am fine with either way :)

cbliard:
Unsurprinsigly, rubocop fails, and the github actions is reaching a limit here (`Only 65535 characters are allowed; 66075 were supplied`). 
If the tests are passing, I would suggest to just merge it.

klaustopher (approved):
"Problem" is that it's rescanning every file because you touched everything and there is just a lot of offenses in the test files. If the tests still pass, everything should be fine IMHO

aaron-contreras (approved):
Beautiful.

PR #12740, opened 2023-05-31
[48239] Baseline time zone label not vertically centered

bsatarnejad:
https://community.openproject.org/work_packages/48239/activity

### DIFF
--- a/frontend/src/app/features/work-packages/components/wp-baseline/baseline/baseline.component.sass
+++ b/frontend/src/app/features/work-packages/components/wp-baseline/baseline/baseline.component.sass
@@ -23,7 +23,7 @@
 
     &-help
       display: flex
-      align-items: center
+      align-items: end
       justify-content: center
       background-color: $spot-color-basic-gray-6
       border: 1px solid $spot-color-basic-gray-3


oliverguenther:
Hi @bsatarnejad , I think center is already correct and was fixed by me last week: https://github.com/opf/openproject/commit/0f982e093198c7ac93e1a85ca5e8dae3179f685a Maybe the bug just has the wrong status? using flex-end seems to be a bit off to me

bsatarnejad:
> Hi @bsatarnejad , I think center is already correct and was fixed by me last week: [0f982e0](https://github.com/opf/openproject/commit/0f982e093198c7ac93e1a85ca5e8dae3179f685a) Maybe the bug just has the wrong status? using flex-end seems to be a bit off to me

Hi @oliverguenther 
Yeah I thought the same, center is correct but I found something a little wierd here:
Screenshot from 2023-05-31 13-54-40

When align-items value is center, there is still a tiny space below the word 'UTC+1', but when its value is end, this space is divided between top and bottom of the word 'UTC+1'
Screenshot from 2023-05-31 13-55-14

PR #12751, opened 2023-06-01
[#46841] Filtering with two timestamps for custom field values

dombesz:
https://community.openproject.org/work_packages/46841


cbliard (approved):
Thanks!

PR #12756, opened 2023-06-01
[57911] Implement show changes legend - remove margin-bottom for wp toolbar title

bsatarnejad:
https://community.openproject.org/work_packages/47911/activity#activity-9

There is an extra margin battom of 10px for the toolbar title and its items which is not necessary since the container itself has a margin-bottom of 1 rem.


b12f (commented):
Are you sure this shouldn't be fixed by removing the margin top on the legend instead of removing the marign-bottom on the toolbar-items? Does it still look good when there's no legend?

oliverguenther (approved):

PR #12759, opened 2023-06-01
[#48390] Fix time zone dependency of the `today` query operator

aaron-contreras:
`Date.current` ensures that the comparison against the timestamp is performed in UTC, which prevents causing two different "todays", compared to using `Date.today` which uses the host's time zone.

The `projects_index_spec` already captures the failing behavior without the fix and demonstrates a successful
run with this fix applied regardless of the time of day or host machine that runs it.

Example:
The host machine's time is 23:00 May 29th
The Rails server's time (UTC) is 01:00 May 30th

If you just created a record and you're attempting to filter it with a **created_on today** filter, then
May 30th(`created_at` timestamp's date) != May 29th(`Date.today`'s return value)
so your record wouldn't be returned in the query when it should have been.

See: https://community.openproject.org/work_packages/48390

### DIFF
--- a/app/models/queries/operators/date_range_clauses.rb
+++ b/app/models/queries/operators/date_range_clauses.rb
@@ -32,10 +32,10 @@ module Queries::Operators
     # the end of the day of yesterday + from until the end of today + to.
     def relative_date_range_clause(table, field, from, to)
       if from
-        from_date = Date.today + from
+        from_date = Date.current + from
       end
       if to
-        to_date = Date.today + to
+        to_date = Date.current + to
       end
       date_range_clause(table, field, from_date, to_date)
     end


cbliard (commented):
To reproduce the issue, I set my host clock to May 30th 23:30 with UTC-02 time zone (Nuuk/Greenland), and I ran the test on `dev` branch.

Then when running the test, it fails because the `project_created_on_today` has created_at set to Wed, 31 May 2023 01:33:07 UTC +00:00, and the filter for today displays the projects created on server's date, which on my host is 30 May. The project is created on May 31st because the test uses `Time.zone.now` and Rails `Time.zone` is UTC.

About Rails time zone being UTC: `Time.zone` returns `Time.zone_default` here. This value is set from `Rails.application.config.time_zone`. In OpenProject [it is not overridden](https://github.com/opf/openproject/blob/678d01f0c203b3753c62fe15d8c2d846a272c6f2/config/application.rb#L134) so it keeps the default value of `UTC`, as defined in the railties gem in [`lib/rails/application/configuration.rb:49`](https://github.com/rails/rails/blob/91f512819b1d0797532fbce2d221b51a5e2347ef/railties/lib/rails/application/configuration.rb#L49).

I'm not sure if the fix you proposed is the right one. There is definitely something wrong about the way the current test is written, but I think there is also something wrong about the way "today" currently works. I'll explain. Stepping away from this test, on `dev`, it takes `Date.today` which for community server is UTC date, because all the servers used by our instances have UTC time zone (it may be different for instances managed by our customers). And on this PR, it takes `Time.zone.now` which will also be UTC. On the other hand, when a user selects today, they probably means today in their time zone, so a user in San Francisco with UTC-7 who created a project on May 30th in the morning at 9:00 (UTC-7) will see the project listed with a "today" filter when looking at the list at 12:00, but will not see it anymore after 17:00, because 17:01 UTC-7 is 24:01 UTC so "today" is already tomorrow for server's time.

And continuing in this direction, if the server is installed in San Francisco, and has a time zone of UTC-7, then `Date.today` will return a correct value this time, whereas `Date.current` will return the date in UTC which would yield incorrect results (for people living in UTC-7).

So I'm not sure about what's right to do here. Sure it fixes the spec, but it also changes the current behavior for people having installed their server on a host with a configured time zone different from UTC.
It looks like the people working on baseline comparison had to solve a similar issue with regards to time zones. 

Do you have an opinion on it @ulferts?

aaron-contreras:
> To reproduce the issue, I set my host clock to May 30th 23:30 with UTC-02 time zone (Nuuk/Greenland), and I ran the test on `dev` branch.
> 
> Then when running the test, it fails because the `project_created_on_today` has created_at set to Wed, 31 May 2023 01:33:07 UTC +00:00, and the filter for today displays the projects created on server's date, which on my host is 30 May. The project is created on May 31st because the test uses `Time.zone.now` and Rails `Time.zone` is UTC.
> 
> About Rails time zone being UTC: `Time.zone` returns `Time.zone_default` here. This value is set from `Rails.application.config.time_zone`. In OpenProject [it is not overridden](https://github.com/opf/openproject/blob/678d01f0c203b3753c62fe15d8c2d846a272c6f2/config/application.rb#L134) so it keeps the default value of `UTC`, as defined in the railties gem in [`lib/rails/application/configuration.rb:49`](https://github.com/rails/rails/blob/91f512819b1d0797532fbce2d221b51a5e2347ef/railties/lib/rails/application/configuration.rb#L49).
> 
> I'm not sure if the fix you proposed is the right one. There is definitely something wrong about the way the current test is written, but I think there is also something wrong about the way "today" currently works. I'll explain. Stepping away from this test, on `dev`, it takes `Date.today` which for community server is UTC date, because all the servers used by our instances have UTC time zone (it may be different for instances managed by our customers). And on this PR, it takes `Time.zone.now` which will also be UTC. On the other hand, when a user selects today, they probably means today in their time zone, so a user in San Francisco with UTC-7 who created a project on May 30th in the morning at 9:00 (UTC-7) will see the project listed with a "today" filter when looking at the list at 12:00, but will not see it anymore after 17:00, because 17:01 UTC-7 is 24:01 UTC so "today" is already tomorrow for server's time.
> 
> And continuing in this direction, if the server is installed in San Francisco, and has a time zone of UTC-7, then `Date.today` will return a correct value this time, whereas `Date.current` will return the date in UTC which would yield incorrect results (for people living in UTC-7).
> 
> So I'm not sure about what's right to do here. Sure it fixes the spec, but it also changes the current behavior for people having installed their server on a host with a configured time zone different from UTC. It looks like the people working on baseline comparison had to solve a similar issue with regards to time zones.
> 
> Do you have an opinion on it @ulferts?

@cbliard Yeah, good catch. This all started as a flickering spec but has raised questions about what the source of truth for `today` should actually be. My idea behind using `Date.current` was to normalize comparisons to occur in the same time zone - in this case always the server's, so it cut down the problem (but not really now that I see your example and point of view).

This raises some questions as you mentioned as to whether we normalize comparisons to the time zone set in the user's preference (if defined), but then if there is no time zone preference, what should the fall-back behavior be to? The `today` operator in this scenario is probably no use. A `between-date` would probably be the most accurate providing the range from the browser's perspective.

Thanks for pointing out these gotchas!

PR #12767, opened 2023-06-02
[48328] My Page: My spent time hows incorrect non-working days

bsatarnejad:
- [x] Show Instance-wide non-working days on my-spent-time widget.
- [x] Show non-working days correctly on calendar.

https://community.openproject.org/projects/openproject/work_packages/48328/activity


ulferts (changes requested):
Hi @bsatarnejad ,

your change finally loads the non working days which is an improvement.

The bug reported is not fixed by this, though as the weekdays are off by one, I think. With the configuration of:

image

The non working days displayed in the calendar are:
image

E.g. Mondays are non working days but Friday isn't and the non repeating non working day on the 12th is displayed to happen on the 13th. 

I guess it is just because of a different way of counting the working days (1 based vs 0 based)

I also came across the case where no non working days were displayed at all. That happened every time I loaded the page initially:

image

Upon switching the week, the non working days would pop up.

Adding an automated feature spec would also help to avoid regressions in the future.

bsatarnejad:
Closing this PR because I fixed it on another PR.

PR #12768, opened 2023-06-02
avoid login check on ical controller

ulferts:
The user never logs into the application on iCal requests (User.current is not set)

https://community.openproject.org/wp/48412

### DIFF
--- a/modules/calendar/app/controllers/calendar/ical_controller.rb
+++ b/modules/calendar/app/controllers/calendar/ical_controller.rb
@@ -28,6 +28,9 @@
 
 module ::Calendar
   class ICalController < ApplicationController
+    # Authentication and authorization is handled within the service.
+    skip_before_action :check_if_login_required
+
     def show
       begin
         call = ::Calendar::ICalResponseService.new.call(
@@ -41,7 +44,6 @@ module ::Calendar
 
       if call.present? && call.success?
         send_data call.result, filename: "openproject_calendar_#{DateTime.now.to_i}.ics"
-        # render plain: call.result # TODO: remove this, it's just handy for development debugging
       else
         render_404
       end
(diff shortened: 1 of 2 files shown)


modules/calendar/app/controllers/calendar/ical_controller.rb
```diff
@@ -28,6 +28,9 @@
 
 module ::Calendar
   class ICalController < ApplicationController
+    # Authentication and authorization is handled within the service.
+    skip_before_action :check_if_login_required
```
jonasjabari, line 32: Just out of curiosity: Why was this not an issue on the pull preview environment?

modules/calendar/spec/controllers/ical_controller_spec.rb
```diff
@@ -28,7 +28,7 @@
 
 require 'spec_helper'
 
-RSpec.describe Calendar::ICalController do
+RSpec.describe Calendar::ICalController, with_settings: { login_required: true } do
```
jonasjabari: Shouldn't it be `login_required: false`?

modules/calendar/app/controllers/calendar/ical_controller.rb
ulferts: Because the setting `login_required` is `false` by default but for most installations, `login_required` will later be reset to `true`. Writing this makes we wonder why the default is not the other way around but that is a topic for a different time.

modules/calendar/spec/controllers/ical_controller_spec.rb
ulferts: Turning the setting to `true` will lead to the specs failing, at least those that are supposed to be successful. However, it is probably better to have tests for both cases which is done in 6963cb4

modules/calendar/app/controllers/calendar/ical_controller.rb
jjabari-op: I see!

jjabari-op (approved):

PR #12769, opened 2023-06-02
[#47048] Prepare the database migration

apfohl:
https://community.openproject.org/work_packages/47048

### DIFF
--- a/modules/storages/app/models/storages/last_project_folder.rb
+++ b/modules/storages/app/models/storages/last_project_folder.rb
@@ -26,33 +26,6 @@
 # See COPYRIGHT and LICENSE files for more details.
 #++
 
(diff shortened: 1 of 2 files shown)


ba1ash (commented):
Looks good to me. One question I have: Would you like to add a unique constraint for `projects_storage_id` and `mode` pair to highlight/ensure we store only last values and not the whole history? Does it make sense? Do we need the whole history, by the way?

apfohl:
> Looks good to me. One question I have: Would you like to add a unique constraint for `projects_storage_id` and `mode` pair to highlight/ensure we store only last values and not the whole history? Does it make sense? Do we need the whole history, by the way?

That is actually necessary. Thanks for the hint.

ba1ash (approved):
:+1:

PR #12770, opened 2023-06-02
Fix XSS attack in news author div

cbliard:
Visible in landing page latest news section, when using a name with html entities in it.

Name settings in profile:

image

The xss:

image

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -219,7 +219,7 @@ module ApplicationHelper
     return if author.nil?
 
     I18n.t(:'js.label_added_time_by',
-           author: author.name,
+           author: html_escape(author.name),
            age: created,
            authorLink: user_path(author)).html_safe
   end
--- a/config/locales/js-en.yml
+++ b/config/locales/js-en.yml
@@ -380,7 +380,7 @@ en:
     label_add_row_before: "Add row before"
     label_add_selected_columns: "Add selected columns"
     label_added_by: "added by"
-    label_added_time_by: "Added by <a href=%{authorLink}>%{author}</a> at %{age}"
+    label_added_time_by: "Added by <a href=\"%{authorLink}\">%{author}</a> at %{age}"
     label_ago: "days ago"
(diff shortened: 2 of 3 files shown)


app/helpers/application_helper.rb
```diff
@@ -219,7 +219,7 @@ def authoring_at(created, author)
     return if author.nil?
 
     I18n.t(:'js.label_added_time_by',
-           author: author.name,
+           author: html_escape(author.name),
```
klaustopher, line 222: 💭  It's a good idea to fix this on the display side, but shouldn't we possibly fix this on the saving side as well? For fields like a name, etc there shouldn't be a reason to add any html tags.

klaustopher (approved):

ulferts:
@cbliard I did not check but assume that this should be backported to release/12.5. Can you please do so if applicable? Our CSP should prevent an actual XSS but it is always better to have the root source fixed as well.

app/helpers/application_helper.rb
cbliard: How would fix it on the saving side? By saving it already escaped? Or stripping the unwanted characters?

cbliard:
> @cbliard I did not check but assume that this should be backported to release/12.5. Can you please do so if applicable? Our CSP should prevent an actual XSS but it is always better to have the root source fixed as well.

Sure, I just created #12778 for `release/12.5` branch.

app/helpers/application_helper.rb
klaustopher: Yeah, that would need to be discussed. But HTML tags should in my opinion be stripped from names. There's not really a reason to have `<foo>` in a name. But that's just my opinion. But maybe out of scope for this issue here :)

ulferts:
With #12778 merged into `release/12.5` and `release/12.5` being merged into `dev` afterwards, I would have expected this PR to be empty since the changes look to be the same between the two PRs. 

Because of this, I am closing this PR without merging but please correct me if my assumption is wrong, @cbliard .

PR #12779, opened 2023-06-05
[#47624] reworked docker setup

Kharonus:
- https://community.openproject.org/work_packages/47624
- extracted port exposure from docker compose


docker-compose.yml
```diff
@@ -45,6 +45,9 @@ x-op-backend: &backend
     OPENPROJECT_RAILS__RELATIVE__URL__ROOT: "${OPENPROJECT_RAILS__RELATIVE__URL__ROOT:-}"
     DATABASE_URL: postgresql://${DB_USERNAME:-postgres}:${DB_PASSWORD:-postgres}@${DB_HOST:-db}:${DB_PORT:-5432}/${DB_DATABASE:-openproject}
     OPENPROJECT_EDITION: ${OPENPROJECT_EDITION:-standard}
+    OPENPROJECT_WEB_MAX__THREADS: 1
```
machisuji, line 48: 👍

docs/development/development-environment-docker/README.md
```diff
@@ -12,114 +12,159 @@ And nothing else!
...
 
-Afterwards, set the environment variables to your liking. `DEV_UID` and `DEV_GID` are required to be set so your project
+Afterward, set the environment variables to your liking. `DEV_UID` and `DEV_GID` are required to be set so your project
 directory will not end up with files owned by root.
 
-Both `docker-compose` and `bin/compose` will load the env from this file.
+`docker compose` will load the env from this file.
 
-### 3) Setup database and install dependencies
+You also will want to create a `docker-compose.override.yml` file, which can contain the port exposure of you
```
machisuji: Perhaps you meant "for your containers" here?

docs/development/development-environment-docker/README.md
```diff
@@ -167,30 +213,40 @@ pull the latest changes.
...
+docker attach openproject-backend-1
 ```
-bin/compose run
-```
 
-If the frontend container is not running yet, it will be started. If the backend container is already running, it will
-be stopped. Instead it will be started in the foreground. This way you can debug using pry just as if you had started
-the server locally using `rails s`. You can stop it simply with Ctrl + C too.
+Inside the `backend` container you have a standard rails console. If you attached to the container **after** you run
+into your breakpoint, you won't the common lines pry will print before your prompt.
```
machisuji: Missing a "see" here, I reckon.

docs/development/development-environment-docker/README.md
```diff
@@ -167,30 +213,40 @@ pull the latest changes.
...
+The easiest way to do that is getting the container name from the docker compose list and attaching to it with the
+standard docker command.
 
+```bash
+# Check all running services and their containers.
+# As a default the `backend` container has the name `openproject-backend-1`
+docker compose ps
+
+# Attach to the container
+docker attach openproject-backend-1
```
machisuji, line 229: 👍

docs/development/development-environment-docker/README.md
Kharonus: added `see`

docs/development/development-environment-docker/README.md
Kharonus: changed to `for your`

frontend/package.json
```diff
@@ -180,7 +180,7 @@
     "build:watch": "node --max_old_space_size=4096 ./node_modules/@angular/cli/bin/ng build --watch --named-chunks",
     "tokens:generate": "theo src/app/spot/styles/tokens/tokens.yml --transform web --format sass,json --dest src/app/spot/styles/tokens/dist",
     "icon-font:generate": "node ./src/app/spot/icon-font/generate.js ./src/app/spot/icon-font",
-    "serve": "node --max_old_space_size=8096 ./node_modules/@angular/cli/bin/ng serve --host 0.0.0.0 --public-host http://localhost:4200",
```
Kharonus, line 183: I added this piece of code to be able to set any exposed port to the frontend service. And of course have it configured only at only single place (`.env`). Do you see any problems with that approach? 

`.env.example` was amended of course

machisuji: I'd probably rather use `export` than `eval` but other than that, that's fine!

machisuji (changes requested):
LGTM 👍 Just one small change and I'm happy.

docs/development/development-environment-docker/README.md
```diff
@@ -12,114 +12,159 @@ And nothing else!
...
+### Storybook
 
-You can stop the processes via Ctrl + C. You can also run everything in the background by adding the `-d` option as
-in `bin/compose up -d`. In that case you'll still be able to see the logs using `docker logs` with the respective
-container name. You can see the started containers using `docker ps`.
+There is a service to launch the storybook of the SPOT design system in the local development environment. To run it,
+simply use:
+
+```bash
+# Start the worker and let them run continuously
```
wielinde, line 163: @Kharonus I stumbled upon this comment. It is talking about starting a worker (singular) and letting them (plural) run continuously. The inconsistency of singular and plural made me think: What worker is this comment actually talking about? So if the comment is helpful and we decide that it is good to have it, could we be more specific?

Kharonus: true, I'd change it to `start the worker service and let it run continuously`.

we are talking here about the docker compose service named `worker`. How many worker processes it starts, you cannot see from this line.

machisuji (approved):
👍

PR #12781, opened 2023-06-05
Fix review errors from global index pages

klaustopher:
Fixes issues with https://community.openproject.org/work_packages/47850

We have identified some missing permission stuff. Those did not show up in development as I only tested with an admin user. Have now checked with an unpriviledged user and it works as expected.

### DIFF
--- a/modules/boards/app/controllers/boards/boards_controller.rb
+++ b/modules/boards/app/controllers/boards/boards_controller.rb
@@ -18,7 +18,7 @@ module ::Boards
     end
 
     def overview
-      projects = Project.allowed_to(User.current, :view_boards)
+      projects = Project.allowed_to(User.current, :show_board_views)
       @board_grids = Boards::Grid.includes(:project).where(project: projects)
     end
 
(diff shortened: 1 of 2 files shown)


cbliard (changes requested):
For team planner global page:
- an instance without enterprise token displays a 500 error page
- a user without access to a public project with team planners can see the team planners of this project, but then clicking the team planner link leads to a 403.
  - if this user goes to the public project page, the team planner menu entry is not displayed, so I suppose the team planners should not be visible publicly.
  - note: team planners from private projects are not shown as expected.
- when the team planner module is disabled for a project, and there were some team planners created, they are still visible in the global team planners page. They should be hidden. (clicking the links leads to a 403)

For boards global page:
- fixed: when a user is not a member of a private project, he does not see the boards anymore :+1: 
  - as a side effect when boards module is disabled in a project, the already created ones are not visible when the module is disabled :+1: 

For meetings
- I had a project with 2 meetings, and they were not visible in the global meetings page (using the administrator account). Not sure why, but I guess it's out of scope.
- I did not test this one further.

For anonymous users, they can access global projects, global work packages, and global news pages, but they can't access global boards page. Is it expected?

klaustopher:
> I had a project with 2 meetings, and they were not visible in the global meetings page (using the administrator account). Not sure why, but I guess it's out of scope.

It should only display future meetings on the global index page. Ivana tested the meetings page and found no errors, so I think everything in there should be fine

PR #12783, opened 2023-06-05
[#48114] Create Meetings outside of a Project

aaron-contreras:
See: https://community.openproject.org/work_packages/48114


modules/meeting/config/routes.rb
```diff
@@ -27,11 +27,11 @@
 #++
 
 OpenProject::Application.routes.draw do
-  scope 'projects/:project_id' do
-    resources :meetings, only: %i[new create index]
+  resources :projects, only: %i[] do
```
dombesz, line 30: This might be a good candidate to use [shallow route nesting](https://guides.rubyonrails.org/routing.html#shallow-nesting).

modules/meeting/app/controllers/projects/meetings_controller.rb
```diff
@@ -0,0 +1,109 @@
...
+  before_action :find_project, only: %i[index new create]
+  before_action :build_meeting, only: %i[new create]
+  before_action :convert_params, only: %i[create update]
+  before_action :authorize
+
+  def index
+    @meetings = @project.meetings
+  end
+
+  def create
```
dombesz: The authorization check for projects is missing from create. By altering the meeting form, I am able to create a meeting in a project where I don't have access to.
Jun-15-2023 13-34-53

dombesz (changes requested):
1. I'm not entirely sure if the two controllers are required, to simplify the code we could create a `meetings_scope` like:

```ruby
def meetings
  @projects = Project.allowed_to(User.current, :view_meetings)
  @projects = @projects.where(project_id:) if project_id   
  @meetings = Meeting.where(project: projects).from_today
end
```

This could be used in both of the index pages, without further ado.
The scope can be used to build the Meeting in the create action and it will automatically assign the `project_id`:

```ruby
def create
  @meeting = @meetings.build(meeting_params)
end
```

image

2. Submitting the meeting form without selecting a project results in an error page.

Jun-15-2023 13-06-44

app/views/layouts/_common_head.html.erb
```diff
@@ -6,7 +6,7 @@
 <% relative_url_root = OpenProject::Configuration['rails_relative_url_root'] || '' %>
 <meta name="app_base_path" content="<%= relative_url_root %>"/>
 <base href="<%= relative_url_root %>/" />
-<% if @project %>
+<% if !@global_create_context && @project %>
```
aaron-contreras: Thoughts on hiding the menu this way?

app/views/layouts/base.html.erb
```diff
@@ -62,7 +62,7 @@ See COPYRIGHT and LICENSE files for more details.
 <op-toasts-container></op-toasts-container>
 <op-modal-overlay></op-modal-overlay>
 <spot-drop-modal-portal></spot-drop-modal-portal>
-<% main_menu = render_main_menu(local_assigns.fetch(:menu_name, nil), @project) %>
+<% main_menu = render_main_menu(local_assigns.fetch(:menu_name, nil), !@global_create_context && @project) %>
```
aaron-contreras: Thoughts on hiding the menu this way?

modules/meeting/app/controllers/meetings_controller.rb
```diff
@@ -28,15 +28,17 @@
 
 class MeetingsController < ApplicationController
   around_action :set_time_zone
+  before_action :set_global_create_context, only: %i[new]
```
aaron-contreras: And it's set this way

aaron-contreras:
Converting back to a draft as I'll be using Turbo to fetch the participant list when selecting a project from the dropdown

modules/meeting/app/views/meetings/_form.html.erb
```diff
@@ -86,48 +103,12 @@ See COPYRIGHT and LICENSE files for more details.
...
-          </tr>
-          <% end -%>
-        </tbody>
-      </table>
-    </div>
-  </div>
+  <%= turbo_frame_tag "participants_section",
+                      src: new_form_refresh_url,
+                      data: { refresh_on_form_changes_target: 'turboFrame' },
+                      loading: :lazy do %>
```
ulferts: I don't think the `lazy` loading adds any value here as the element should be visible right from the beginning.

For the global creation case (selecting a project first) we wouldn't even need the `src` property to be set from the beginning as that will result in a useless request being issued where the response is always empty (except for the `turbo-frame`-tag.

In the project creation case no project is selected so the participant section will always stay empty. In there, the url should have the project included right from the start as that will be static anyway.

modules/meeting/app/views/meetings/_form.html.erb
```diff
@@ -35,6 +35,23 @@ See COPYRIGHT and LICENSE files for more details.
...
+                   {prompt: t(:project_selection_placeholder),
+                    container_class: '-wide'},
+                   class: project_select_initial_class_list,
+                   required: true,
+                   data: {
+                     'application-target': 'dynamic',
+                     controller: 'select-field-with-prompt',
+                     action: 'change->select-field-with-prompt#togglePromptStyling ' \
+                              'change->refresh-on-form-changes#triggerTurboReload'
+                   } %>
```
ulferts: There is an angular powered autocompleter that sports all the niceties of a typeahead and also supports hierarchies:

image

It is e.g. used on the [page for bulk moving work packages](https://github.com/opf/openproject/blob/dev/app/views/work_packages/moves/new.html.erb#L65-L77). It should also be used here.

Doing this will require sending the correct filters for the all projects, in which the user is allowed to create meetings. Please see the [page for bulk moving work packages (the `apiFilters` property)](https://github.com/opf/openproject/blob/dev/app/views/work_packages/moves/new.html.erb#L67) as an example again. In order for that to work, the [`create_meetings` permission](https://github.com/opf/openproject/blob/dev/modules/meeting/lib/open_project/meeting/engine.rb#L43) needs to have `contract_actions` added, see [project create](https://github.com/opf/openproject/blob/dev/config/initializers/permissions.rb#L36) for an example.

modules/meeting/app/controllers/meetings_controller.rb
```diff
@@ -123,14 +128,24 @@ def set_time_zone(&)
...
   def build_meeting
     @meeting = Meeting.new
+    if meeting_params.present?
+      convert_params
+      @meeting.participants.clear # Start with a clean set of participants
+      @meeting.participants_attributes = @converted_params.delete(:participants_attributes)
+      @meeting.attributes = @converted_params
+    end
     @meeting.project = @project
     @meeting.author = User.current
```
ulferts, line 151: Not necessary for this PR but it would make sense to move this functionality into a `Meetings::SetAttributesService`.

modules/meeting/app/controllers/meetings_controller.rb
```diff
@@ -123,14 +128,24 @@ def set_time_zone(&)
...
+  end
+
   def find_optional_project
-    return true unless params[:project_id]
+    return true if project_id.blank?
 
-    @project = Project.find(params[:project_id])
+    @project = Project.find(project_id)
     authorize
   rescue ActiveRecord::RecordNotFound
```
ulferts: Please check whether this method overwrite over what is already implemented in [the `ApplicationController`](https://github.com/opf/openproject/blob/dev/app/controllers/application_controller.rb#L284-L288) is actually necessary.

modules/meeting/spec/features/meetings_new_spec.rb
```diff
@@ -50,72 +50,150 @@
...
-      index_page.expect_no_create_new_button
+          show_page = new_page.click_create
+
+          show_page.expect_toast(message: 'Successful creation')
+
+          show_page.expect_invited(user, other_user)
+
+          show_page.expect_date_time "03/28/2013 01:30 PM - 03:00 PM"
+        end
+      end
```
ulferts, line 161: I don't understand why those passed on the CI. They fail for me locally and I think they do so rightfully as the `participants_section` does not show up on the project scoped new form.

ulferts (changes requested):
Hi @aaron-contreras please find my comments attached to the code. 

The part I am absolutely confused about is that the feature specs for the project scoped creation of meetings do not report the missing participant section.

modules/meeting/spec/features/meetings_new_spec.rb
aaron-contreras: Now we know :)

modules/meeting/app/controllers/meetings_controller.rb
aaron-contreras: It is no longer necessary by using the `op-project-autocompleter` as I can have `project_id` be a non-nested param within the `meeting` key.

aaron-contreras:
As discussed, I've moved the PR back to a draft state in order to modify the form implementation to use turbo-streams instead of turbo-frames.

modules/meeting/app/helpers/meetings_helper.rb
```diff
@@ -72,4 +72,14 @@ def render_journal_details(journal, header_label = :label_updated_time_by, _mode
 
     content_tag('div', "#{header}#{details}".html_safe, id: "change-#{journal.id}", class: 'journal')
   end
+
+  def global_create_context?
+    request.path == new_meeting_path
+  end
+
+  def options_for_project_selection
```
ulferts: This can now be removed since the functionality is now provided by the filters passed into the autocompleter.

modules/meeting/config/locales/en.yml
```diff
@@ -99,6 +101,8 @@ en:
 
   project_module_meetings: "Meetings"
 
+  project_selection_placeholder: "Select project"
```
ulferts: Can be removed since the autocompleter is implemented in Angular now.

modules/meeting/app/controllers/meetings_controller.rb
```diff
@@ -127,15 +128,6 @@ def build_meeting
...
 
-  def find_optional_project
-    return true unless params[:project_id]
-
-    @project = Project.find(params[:project_id])
-    authorize
-  rescue ActiveRecord::RecordNotFound
-    render_404
-  end
-
```
ulferts, line 161: Always good to get rid of those controller specific overrides if they are not needed. 🙇

modules/meeting/lib/open_project/meeting/engine.rb
```diff
@@ -40,7 +40,10 @@ class Engine < ::Rails::Engine
       project_module :meetings do
         permission :view_meetings, meetings: %i[index show], meeting_agendas: %i[history show diff],
                                    meeting_minutes: %i[history show diff]
-        permission :create_meetings, { meetings: %i[new create copy] }, require: :member
+        permission :create_meetings,
+                   { meetings: %i[new participants_section create copy] },
```
ulferts: With the change to turbo-stream, there no longer is a `participants_section` method added to the controller so this addition can be reverted.

ulferts (changes requested):
The things I am complaining about are leftovers of the pervious attempt. Feel free to merge this PR yourself, @aaron-contreras after those have been removed since the overall change is good.

PR #94, opened 2013-04-29
Feature/rails3 fix adding related issue

cratz:
[OPF #476] fixes adding related issue

PR #92, opened 2013-04-29
[OPF #478] fixes wiki page renaming

cratz:
- changed wiki#rename routes to rails default 'put'
- adapted tests

PR #89, opened 2013-04-26
fix user create with empty pwd

cratz:
- fixed wrong condition in users pwd validation methode

PR #12790, opened 2023-06-06
Track the cause for Journals to associate journal with the right cause

klaustopher:
Implements https://community.openproject.org/work_packages/46481

- [x] Add database migration to add the `cause` field
- [x] Add the `store_accessor`s to access JSON fields
- [x] `CreateService`: Add the cause to the SQL query
- [x] `CreateSerivce`: Change logic for aggregation of journals when a cause is involved
- [x] Implement cause for changing Working/NonWorking days 
- [x] Add the cause to the journals when changing data in Scheduling
- [x] Write tests for everything
- [x] Add i18n for displaying the cause types properly
- [x] Add cause to the API representation of journals

### DIFF
--- a/app/models/journal.rb
+++ b/app/models/journal.rb
@@ -30,6 +30,8 @@ class Journal < ApplicationRecord
   self.table_name = 'journals'
   self.ignored_columns += ['activity_type']
 
+  WorkingDayUpdate = Struct.new(:working_days, :non_working_days, keyword_init: true)
+
   include ::JournalChanges
   include ::JournalFormatter
   include ::Acts::Journalized::FormatHooks
@@ -52,11 +54,12 @@ class Journal < ApplicationRecord
   # Attributes related to the cause are stored in a JSONB column so we can easily add new relations and related
   # attributes without a heavy database migration. Fields will be prefixed with `cause_` but are stored in the JSONB
   # hash without that prefix
-  store_accessor :cause, %i[type work_package_id], prefix: true
+  store_accessor :cause, %i[type work_package_id changed_days], prefix: true
   VALID_CAUSE_TYPES = %w[
     work_package_predecessor_changed_times
     work_package_successor_changed_times
     work_package_parent_changed_times
+    working_days_changed
   ].freeze
 
   # Make sure each journaled model instance only has unique version ids
--- a/app/models/journal.rb
+++ b/app/models/journal.rb
@@ -53,7 +53,11 @@ class Journal < ApplicationRecord
   # attributes without a heavy database migration. Fields will be prefixed with `cause_` but are stored in the JSONB
   # hash without that prefix
   store_accessor :cause, %i[type work_package_id], prefix: true
-  VALID_CAUSE_TYPES = %w[work_package_predecessor_changed_times work_package_successor_changed_times].freeze
+  VALID_CAUSE_TYPES = %w[
+    work_package_predecessor_changed_times
+    work_package_successor_changed_times
+    work_package_parent_changed_times
+  ].freeze
 
   # Make sure each journaled model instance only has unique version ids
   validates :version, uniqueness: { scope: %i[journable_id journable_type] }
--- a/app/models/journal.rb
+++ b/app/models/journal.rb
@@ -60,6 +60,7 @@ class Journal < ApplicationRecord
     work_package_predecessor_changed_times
     work_package_parent_changed_times
     work_package_children_changed_times
+    work_package_related_changed_times
     working_days_changed
   ].freeze
 
--- a/app/models/journal.rb
+++ b/app/models/journal.rb
@@ -50,6 +50,7 @@ class Journal < ApplicationRecord
   register_journal_formatter :time_entry_hours, OpenProject::JournalFormatter::TimeEntryHours
   register_journal_formatter :wiki_diff, OpenProject::JournalFormatter::WikiDiff
   register_journal_formatter :time_entry_named_association, OpenProject::JournalFormatter::TimeEntryNamedAssociation
+  register_journal_formatter :cause, OpenProject::JournalFormatter::Cause
 
   # Attributes related to the cause are stored in a JSONB column so we can easily add new relations and related
   # attributes without a heavy database migration. Fields will be prefixed with `cause_` but are stored in the JSONB
--- a/app/models/journal.rb
+++ b/app/models/journal.rb
@@ -58,8 +58,8 @@ class Journal < ApplicationRecord
   store_accessor :cause, %i[type work_package_id changed_days], prefix: true
   VALID_CAUSE_TYPES = %w[
     work_package_predecessor_changed_times
-    work_package_successor_changed_times
     work_package_parent_changed_times
+    work_package_children_changed_times
     working_days_changed
   ].freeze
 
--- a/app/models/journal.rb
+++ b/app/models/journal.rb
@@ -49,8 +49,15 @@ class Journal < ApplicationRecord
   register_journal_formatter :wiki_diff, OpenProject::JournalFormatter::WikiDiff
   register_journal_formatter :time_entry_named_association, OpenProject::JournalFormatter::TimeEntryNamedAssociation
 
+  # Attributes related to the cause are stored in a JSONB column so we can easily add new relations and related
+  # attributes without a heavy database migration. Fields will be prefixed with `cause_` but are stored in the JSONB
+  # hash without that prefix
+  store_accessor :cause, %i[type work_package_id], prefix: true
+  VALID_CAUSE_TYPES = %w[work_package_predecessor_changed_times work_package_successor_changed_times].freeze
+
   # Make sure each journaled model instance only has unique version ids
   validates :version, uniqueness: { scope: %i[journable_id journable_type] }
+  validates :cause_type, inclusion: { in: VALID_CAUSE_TYPES, allow_blank: true }
 
   belongs_to :user
   belongs_to :journable, polymorphic: true
@@ -125,6 +132,10 @@ class Journal < ApplicationRecord
 
   def noop?
     (!notes || notes&.empty?) && get_changes.empty?
+  end
+
+  def has_cause?
+    cause_type.present?
   end
 
   private
--- a/app/services/journals/create_service.rb
+++ b/app/services/journals/create_service.rb
@@ -49,8 +49,12 @@ module Journals
     end
 
     def call(notes: '', cause: {})
+      # JSON columns read from the database always have string keys. As we do not know what is passed in here,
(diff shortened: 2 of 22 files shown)


app/services/journals/create_service.rb
```diff
@@ -161,9 +163,9 @@ def create_journal_sql(predecessor, notes)
         ), changes AS (
           #{select_changed_sql}
         ), insert_data AS (
-          #{insert_data_sql(notes, predecessor)}
+          #{insert_data_sql(predecessor, notes, cause)}
```
klaustopher, line 168: All other methods here used the order `predecessor, notes` so I changed the signature of this method to follow

app/models/journal.rb
```diff
@@ -49,8 +51,20 @@ class Journal < ApplicationRecord
...
 
+  # Attributes related to the cause are stored in a JSONB column so we can easily add new relations and related
+  # attributes without a heavy database migration. Fields will be prefixed with `cause_` but are stored in the JSONB
+  # hash without that prefix
+  store_accessor :cause, %i[type work_package_id changed_days], prefix: true
+  VALID_CAUSE_TYPES = %w[
+    work_package_predecessor_changed_times
+    work_package_successor_changed_times
+    work_package_parent_changed_times
+    working_days_changed
```
ulferts, line 64: I cannot think of a  case where changes to the successor (`work_package_successor_changed_times`) would lead to change of dates of the predecessor. It is specified in the OP WP but I don't think that the current implementation ever reschedules a work package because of it.

Is the `work_package_parent_changed_times` meant for when a parent's dates are changed because of changes to one of their children? If that is the case, my preferred wording would be `work_package_child_changed_times` as that is the the cause for the change. If it really is meant for when changes to the `parent` causes changes to its ancestors then I can only think of two scenarios where this might happen:
1) When the child is generated initially and the parent already has dates set. Since it isn't a rescheduling (only an initial scheduling and the user will see it right away) I wouldn't set the cause, then.
2) When the work package receives a new parent. If the work package altered does not have a start date of its own, it receives one. This might be a valid case but I am not completely convinced yet. I'd have to play around in the application to find out if that ever happens without the user having changed the exact work package that receives a new parent. If that holds true, the cause would reference the work package for which the journal is created.

The non discernible cause for the `changed_times` is missing. E.g. when the predecessor of the parent is altered.

app/services/journals/create_service.rb
```diff
@@ -67,20 +67,20 @@ def call(notes: '')
     # and only the predecessor or the journal to be created has notes, the changes are aggregated.
     # Instead of removing the predecessor, return it here so that it can be stripped in the journal creating
     # SQL to than be refilled. That way, references to the journal, including ones users have, are kept intact.
```
ulferts, line 73: Since the code in here isn't trivial, the documentation should be adapted.

app/services/journals/create_service.rb
```diff
@@ -145,7 +145,7 @@ def create_journal(notes)
     #
     # If a journal is created, all entries in the custom_values table associated to the journable are recreated as entries
     # in the customizable_journals table. Again, newlines are normalized.
```
ulferts, line 151: Since the code in here isn't trivial, the documentation should be adapted.

app/services/journals/create_service.rb
ulferts: 👍

app/services/work_packages/set_schedule_service.rb
```diff
@@ -122,6 +124,7 @@ def reschedule(scheduled, dependency)
   # descendants
   def reschedule_by_descendants(scheduled, dependency)
     set_dates(scheduled, dependency.start_date, dependency.due_date)
+    assign_cause_for_journaling(scheduled, :predecessor)
```
ulferts: The cause in this case should be `:child`.

app/services/work_packages/set_schedule_service.rb
```diff
@@ -147,6 +150,7 @@ def reschedule_by_predecessors(scheduled, dependency)
     new_start_date = [scheduled.start_date, dependency.soonest_start_date].compact.max
     new_due_date = determine_due_date(scheduled, new_start_date)
     set_dates(scheduled, new_start_date, new_due_date)
+    assign_cause_for_journaling(scheduled, :successor)
```
ulferts: In my opinion, the cause should be called `:predecessor` here. However, we cannot really know at this point whether it is caused by a predecessor of this work package or a predecessor of its ancestors.

app/services/work_packages/set_schedule_service.rb
```diff
@@ -174,4 +178,34 @@ def set_dates(work_package, start_date, due_date)
...
+    }
+  end
+
+  def assign_cause_initiated_by_changed_working_days(work_package)
+    work_package.journal_cause = {
+      "type" => 'working_days_changed',
+      "changed_days" => initiated_by.to_h
+    }
+  end
 end
```
ulferts, line 216: Given the sometimes hidden relationship (e.g. caused by a change to the predecessor or to a predecessor of an ancestor, it might be simpler to stick to finding out the kind of relationship not through the code paths but rather by the reference itself.

Since there will always only be one work package causing it, one could find out the direct predecessors, children and the parent for this work package (or at least their ids) and when setting the cause, check if the altered work package is in any of the sets and if that is the case, use the key for the set it is found in.. If it isn't, the general fallback will be used.

ulferts (commented):
I did not do  a thorough review so please take my remarks with a grain of salt. 

Apart form what I comment at the code, I am struggling with the way the the cause is passed through the layers. You had remarked the same before and I don't have a good solution. Would it be an option to at least have a stricter object for passing the values instead of a hash? Not really a solution for the passing itself but at least it might simplify using the interface.

app/models/journal.rb
klaustopher: 👍 I was not yet sure about the cases and was about to play around with this in the UI today. Makes sense, will take this into consideration

app/services/work_packages/set_schedule_service.rb
klaustopher: Changed

app/services/work_packages/set_schedule_service.rb
klaustopher: Changed 👍

app/services/work_packages/set_schedule_service.rb
klaustopher: You mean that **date changed by changes to predecessor 123** is only used in the direct followers and not in the entire chain?

app/services/journals/create_service.rb
klaustopher: done

lib/open_project/journal_formatter/cause.rb
```diff
@@ -0,0 +1,103 @@
...
+  def working_date_change_message(date, working)
+    I18n.t("journals.cause_descriptions.working_days_changed.dates.#{working ? :working : :non_working}",
+           date: I18n.l(Date.parse(date)))
+  end
+
+  # we need to tell the url_helper that there is not controller to get url_options
+  # so that we can later call link_to
+  def controller
+    nil
+  end
```
ulferts, line 102: nitpicking: you could also write `def controller; end`

klaustopher:
I created a follow-up PR that shows how we could re-implement the passing data as models: https://github.com/opf/openproject/pull/12882

I'm not too sure if I am really happy with that approach either.

app/services/work_packages/set_schedule_service.rb
ulferts: Yes, something along those lines. 

The specification is not really clear on this but my understanding of the talk with @psatyal was that the cause should reference the work package directly changed by the user. The current implementation reflects that and is thus in line with my understanding. When referencing the relationship of the directly manipulated work package it then to me only makes sense to use the term 'child' if it is the immediate child. For 'predecessor' my understanding of the term includes the transitive predecessors as well but it will quickly become very complicated to name the exact relationship.

Please see the following examples
 
### The predecessor to a parent is moved 
image
When the predecessor is moved so that the following work packages (parent and child) have to be rescheduled, the following causes are given:
* For the parent:
image

* For the child:
image

### The predecessor of a predecessor is moved
image

With "Predecessor" being moved, the subsequent work packages will have to be rescheduled. In this case, both ("Middlepackage" and "Successor") receive the message:
image

### A descendant (not the direct child) is moved
image

With "Grandchild" being moved, the other two both receive the message:
image

### A made up structure of followers and hierarchies
image

With "Epicenter" being moved:

* For "epicenter parent follower parent":
image
 
* For "epicenter parent follower":
image

* For "epicenter parent follower child":
image

* For "epicenter parent follower child follower"
image

## My conclusion

To me, the terms `predecessor`, `parent` and `child` should only be used if that relationship is direct. As soon as there is but one hop in between, the general term `related` should be used.

ulferts (changes requested):
The structure of the change and the code itself is good. There are however a couple of behavioural shortcomings. 

I've split those up between changes I would like to see in the way the message is printed and those where I have doubts about the overall behaviour. The message related shortcomings are attached to an in code comment. The more general concerns follow here:

* Having a work package with only a duration and the parent to be not having any dates as well. Setting the new parent will lead to
`Dates changed by changes to parent Task #89: Child` without changes to the due/start date actually done
* Having a work package with dates and a parent to be. Setting the new parent will lead to the message `Dates changed by changes to child Task #92: Child` although no changes to the child's dates have been performed. Only the parent has been set.
* Having a work package without dates and a parent to be which has dates. Setting the new parent will lead to the message `Dates changed by changes to child Task #92: Child` and all the dates of the parent being removed. Only the parent has been set.
* Having a work package with a due date (no start date) and a parent which is scheduled after the due date and has a predecessor to tie it down. Setting the parent will not lead to any "Dates changed" message but it should have been there (wrong indentation of the `assign_cause_for_journaling`. The due date is also lost.   
* Having a work package with a due date (no start date) and a parent to be. The parent to be is scheduled to be after the due date of the work package. Setting the parent will lead to the message `Dates changed by changes to parent Task #94: Child` although no changes have been made to the child.   

I am not sure myself on how all of the above should be resolved. Therefore I propose to go through them in a chat.

lib/open_project/journal_formatter/cause.rb
klaustopher: Copied this verbatim from another Formatter, will change

klaustopher:
Talked about the following changes:

- Instead of the parent/child/predecessor message we opt to always track it just as "related work package #xxx changed". We will see how it feels and if we really need the additional information
- We will only set a cause if any of the relevant fields have been changed

ulferts (approved):
Thanks for digging through this, @klaustopher

PR #12792, opened 2023-06-06
[#48420] Sums table in work plan pdf export

as-op:
https://community.openproject.org/work_packages/48420


klaustopher (approved):

as-op:
Thank you @klaustopher!

PR #12796, opened 2023-06-07
[#48454] [AppSignal] ERROR:  column settings.updated_at does not exist

dombesz:
https://community.openproject.org/work_packages/48454


machisuji:
@dombesz Thanks for the PR. I think it might be preferrable to just remove the caching altogether here, when in migrations, to make sure other migrations aren't hit by the same issue. I think there is a way to check for this. I still want to try this locally too. Will try to get to that tomorrow.

PR #12797, opened 2023-06-07
Implementation/48432 migrate existing roles to include subscribe to ical permission

jjabari-op:
https://community.openproject.org/work_packages/48432


modules/calendar/db/migrate/20230607101213_populate_share_calendar_role_permission.rb
```diff
@@ -0,0 +1,20 @@
...
+    # From https://community.openproject.org/projects/openproject/work_packages/15339/activity
+    # -> For the migration path, all the roles that has the setting "View calendars" active will
+    # also by default have "Subscribe to iCalendars" active
+    role_ids_with_view_calendar_permission = RolePermission.where(permission: "view_calendar").distinct(:role_id).pluck(:role_id)
+
+    new_permissions = role_ids_with_view_calendar_permission.map do |role_id|
+      { role_id:, permission: "share_calendars" }
+    end
+
+    RolePermission.insert_all(new_permissions)
```
ulferts: We have a [nifty little helper](https://github.com/opf/openproject/blob/dev/db/migrate/migration_utils/permission_adder.rb) for this used e.g. in [this migration](https://github.com/opf/openproject/blob/dev/db/migrate/20190527095959_set_manage_board_permission.rb)

modules/calendar/db/migrate/20230607101213_populate_share_calendar_role_permission.rb
```diff
@@ -0,0 +1,20 @@
...
+  end
+
+  def down
+    # Question: is following down migration desired? I tend to not include that as it (in theory) might
+    # remove manually set share_calendar permissions
+    # RolePermission.where(permission: "share_calendars").delete_all
```
ulferts: Nothing needed on the down migration since we cannot discern between those added by a user and those added automatically.

ulferts (changes requested):
Hi @jjabari-op can you please perform the changes requested. Hopefully that fixes the stubborn migrations magically.

ulferts (approved):

PR #12799, opened 2023-06-07
[#48441] Wrong bread crumbs in file picker

Kharonus:
- https://community.openproject.org/work_packages/48441
- replace nextcloud naming escape with simple replace to match nextcloud functionality

### DIFF
--- a/modules/storages/app/common/storages/peripherals/storage_interaction/nextcloud/files_query.rb
+++ b/modules/storages/app/common/storages/peripherals/storage_interaction/nextcloud/files_query.rb
@@ -36,7 +36,7 @@ module Storages::Peripherals::StorageInteraction::Nextcloud
     # rubocop:disable Metrics/AbcSize
     def call(user:, folder:)
       result = Util.token(user:, oauth_client: @oauth_client) do |token|
-        @base_path = Util.join_uri_path(@uri.path, "remote.php/dav/files", CGI.escapeURIComponent(token.origin_user_id))
+        @base_path = Util.join_uri_path(@uri.path, "remote.php/dav/files", token.origin_user_id.gsub(' ', '%20'))
 
         response = Util.http(@uri).propfind(
           Util.join_uri_path(@base_path, requested_folder(folder)),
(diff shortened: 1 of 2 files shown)


modules/storages/spec/common/peripherals/storage_requests_spec.rb
```diff
@@ -191,26 +192,18 @@ def origin_name
...
-              id: '12',
-              name: 'README.md',
-              mime_type: 'text/markdown',
-              permissions: include(:readable, :writeable)
-            )
-            expect(result.result.files[3]).to have_attributes(
-              mime_type: 'application/pdf',
-              permissions: %i[readable]
-            )
+            expect(result.result.files[0]).to have_attributes(id: '11',
```
Kharonus, line 195: everything below here is just reformatting triggered by rubymine autoformatter

modules/storages/app/common/storages/peripherals/storage_interaction/nextcloud/files_query.rb
```diff
@@ -36,7 +36,7 @@ def initialize(storage)
     # rubocop:disable Metrics/AbcSize
     def call(user:, folder:)
       result = Util.token(user:, oauth_client: @oauth_client) do |token|
-        @base_path = Util.join_uri_path(@uri.path, "remote.php/dav/files", CGI.escapeURIComponent(token.origin_user_id))
+        @base_path = Util.join_uri_path(@uri.path, "remote.php/dav/files", token.origin_user_id.gsub(' ', '%20'))
```
wielinde: @Kharonus Side note: The name `Util` is an abbreviation, right? AFAIK in Ruby the convention is trying to not abbreviate words so that the codes gets optimized for reading. A commonly used term could be `Helpers`

ba1ash (approved):
:+1:

modules/storages/app/common/storages/peripherals/storage_interaction/nextcloud/files_query.rb
Kharonus: I think we should avoid abbreviations, except very, very common abbreviations. I mean, we use http, uri, id. Alone in this line one ruby method is called `gsub`. 
From this point of view, I'd argue, that `util` is a quite common abbreviation.

PR #12803, opened 2023-06-07
Fix flaky `notification_center_date_alerts_spec`

aaron-contreras:
There are situations where a false assignable time_zone match would occur if
comparing `tz.utc_offset` to `local_offset` since the `utc_offset` stored in
the TZInfo object indeed matched the local_offset but when later utilizing the
returned `time_zone` object by calling `time_zone.now`, `ActiveSupport::TimeZone`
would perform a DST adjusment and potentially shift `time_zone.now` into a different
time/date than `Time.zone.now`.

Example from CI: Host is set to UTC time zone with a +0 offset. The `find_compatible_time_zone`
method would pick `London` since on paper, the offset was also +0; however, when using
`time_zone.now` the offset became +1 given the automatic Daylight Savings Time adjustment
since London does following DST. If UTC time was 23:XX, then the spec would fail.

This failed CI run demonstrates the issue/example above: https://github.com/opf/openproject/actions/runs/5182740870/jobs/9339858330


cbliard (approved):
Nice one :+1: 
Thanks for the detailed commit message

PR #12810, opened 2023-06-08
install correct gosu for each architecture

machisuji:
WP [#48395](https://community.openproject.org/projects/docker/work_packages/48395/activity)

This has been tested as described in the work package already, that is for x64, arm64 and ppc64. So all that's needed is a code review.


cbliard (approved):
:+1:

PR #12838, opened 2023-06-12
Restore some authentication settings in admin

cbliard:
Also makes ckEditorAugmentedTextarea read-only if the wrapped text area is disabled.

As for the invitation_expiration_days setting, commit 269f9416ddc and commit 286d862c38 got intertwined with merge commit eb63698a3dd, and resulted in hiding the setting when
`OpenProject::Configuration.registration_footer` is set.

As for the registration_footer setting, PR #6321 intent was to prevent modifying it if it was already set via env var or configuration file. When `OpenProject::Configuration` and `Settings` were merged, this code was not updated. With the default value being non-blank, it lead to hiding the setting from the administration pages.

This commit restores both settings.

### DIFF
--- a/spec/controllers/admin/settings/authentication_controller_spec.rb
+++ b/spec/controllers/admin/settings/authentication_controller_spec.rb
@@ -40,7 +40,8 @@ RSpec.describe Admin::Settings::AuthenticationSettingsController do
   describe 'PATCH #update' do
     render_views
 
-    current_user { create(:admin) }
+    shared_let(:admin) { create(:admin) }
+    current_user { admin }
 
     describe 'registration_footer' do
       let(:old_settings) do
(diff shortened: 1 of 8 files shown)


spec/controllers/admin/settings/authentication_controller_spec.rb
```diff
@@ -36,4 +36,66 @@ def fetch
 
     it_behaves_like 'a controller action with require_admin'
   end
+
+  describe 'PATCH #update' do
+    render_views
```
ulferts: Why is this needed here? I did not find any expectation related to the view.

ulferts (changes requested):
Thanks for addressing this. I found one small issue but overall, it should be mergeable soon.

spec/controllers/admin/settings/authentication_controller_spec.rb
cbliard: You're right, I copied it from another place and did not challenge its need.
I removed it.
Thanks for noticing.

PR #12839, opened 2023-06-12
[#41275] Perform default language reset on saving language settings

aaron-contreras:
See: https://community.openproject.org/work_packages/41275

Saving language settings should automatically reset the language of users with a non-available language
to the default language, and the obscure link in the bottom right corner to perform the language reset
manually will no longer be needed.

### DIFF
--- a/app/services/settings/language_update_service.rb
+++ b/app/services/settings/language_update_service.rb
@@ -1,0 +1,45 @@
+# frozen_string_literal: true
+
+# -- copyright
+# OpenProject is an open source project management software.
+# Copyright (C) 2023 the OpenProject GmbH
+#
+# This program is free software; you can redistribute it and/or
+# modify it under the terms of the GNU General Public License version 3.
+#
+# OpenProject is a fork of ChiliProject, which is a fork of Redmine. The copyright follows:
+# Copyright (C) 2006-2013 Jean-Philippe Lang
+# Copyright (C) 2010-2013 the ChiliProject Team
+#
+# This program is free software; you can redistribute it and/or
+# modify it under the terms of the GNU General Public License
+# as published by the Free Software Foundation; either version 2
+# of the License, or (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program; if not, write to the Free Software
+# Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+#
+# See COPYRIGHT and LICENSE files for more details.
+# ++
+#
+
+class Settings::LanguageUpdateService < Settings::UpdateService
+  def after_perform(call)
+    force_user_language
+
+    call
+  end
+
+  private
+
+  def force_user_language
+    User.where.not(language: Setting.available_languages)
+        .update_all(language: Setting.default_language)
+  end
+end
--- a/app/services/settings/language_update_service.rb
+++ b/app/services/settings/language_update_service.rb
@@ -31,14 +31,14 @@
 
 class Settings::LanguageUpdateService < Settings::UpdateService
   def after_perform(call)
-    force_user_language
+    force_users_to_use_only_available_languages
 
     call
   end
 
   private
 
-  def force_user_language
+  def force_users_to_use_only_available_languages
     User.where.not(language: Setting.available_languages)
         .update_all(language: Setting.default_language)
   end
--- a/spec/controllers/admin/settings/projects_settings_controller_spec.rb
+++ b/spec/controllers/admin/settings/projects_settings_controller_spec.rb
@@ -29,26 +29,17 @@
 require 'spec_helper'
 
 RSpec.describe Admin::Settings::ProjectsSettingsController do
-  before do
-    allow(@controller).to receive(:set_localization)
-    @params = {}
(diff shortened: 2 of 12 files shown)


spec/controllers/admin/settings/projects_settings_controller_spec.rb
```diff
@@ -29,26 +29,17 @@
...
 
-    context 'default project modules' do
-      before do
-        @previous_projects_modules = Setting.default_projects_modules
-      end
-
-      after do
-        Setting.default_projects_modules = @previous_projects_modules
-      end
-
```
cbliard, line 51: Nicely spotted :+1:.
I checked and the setting is reset after the test.

spec/services/settings/language_update_service_spec.rb
```diff
@@ -0,0 +1,63 @@
...
+        .to have_received(:force_user_language)
+    end
+
+    context 'when the contract is not successfully validated' do
+      let(:contract_success) { false }
+
+      include_examples 'unsuccessful call'
+    end
+  end
+end
```
cbliard: I'm not sure how this could be rewritten, but the spec feels really magical. It's hard to understand what's going on with all the stubs and the included context. I understand that it was already like this in `spec/services/settings/update_service_spec.rb`, but I really have a hard time understanding what is really tested, or what exactly would go wrong if the test fails.

Maybe it's also missing a `expect(instance).not_to have_received(:force_user_language)` in the case "when the contract is not successfully validated".

app/services/settings/language_update_service.rb
```diff
@@ -0,0 +1,45 @@
...
+class Settings::LanguageUpdateService < Settings::UpdateService
+  def after_perform(call)
+    force_user_language
+
+    call
+  end
+
+  private
+
+  def force_user_language
```
cbliard: I have a problem with the name `force_user_language`: it does not force only one user, so it should be plural, and it does not force all users either, so it should be more specific.

What about renaming it to `force_users_to_use_only_available_languages`?

cbliard (changes requested):
Great :+1: It works as expected.

I have some minor code remarks.
Feel free to merge when you feel you addressed them all.

spec/services/settings/language_update_service_spec.rb
cbliard: I won't mind if you let the test like it is, just please add the `expect(instance).not_to have_received(:force_user_language)` as it is really what is different between the two cases. Testing for successful/unsuccesful call is less important (and it's almost automatic as it is baked into the service code).

aaron-contreras: I'm honestly open and looking for suggestions on the clarity of the spec and what it's testing. I felt a bit that since this is inheriting the entire behavior from `Settings::UpdateService`, that re-testing any of what is already being tested over on the specs for the base service is redundant and just makes the suite more fragile.

I will be adding the test for not resetting the default language when the contract fails but other than that, do you feel this would be clearer if I didn't abstract the identical setup code into a `shared_context`?

cbliard: > I felt a bit that since this is inheriting the entire behavior from Settings::UpdateService, that re-testing any of what is already being tested over on the specs for the base service is redundant and just makes the suite more fragile.

Agreed.

> I will be adding the test for not resetting the default language when the contract fails

Great :+1: 

> other than that, do you feel this would be clearer if I didn't abstract the identical setup code into a shared_context?

Yes probably. I just gave it a try, and I came down with this:

```ruby
require 'spec_helper'

RSpec.describe Settings::LanguageUpdateService do
  subject(:service) do
    described_class.new(user: build_stubbed(:admin))
  end

  before do
    allow(service).to receive(:force_user_language)
  end

  it 'sets language of users having a non-available language to the default language' do
    service.call('available_languages' => ['de', 'fr'])

    expect(service)
      .to have_received(:force_user_language)
  end

  context 'when the contract is not successfully validated' do
    before do
      allow(service).to receive(:validate_contract).and_return(ServiceResult.failure(message: 'fake error'))
    end

    it 'does not change language of any users' do
      service.call('available_languages' => ['de', 'fr'])

      expect(service)
        .not_to have_received(:force_user_language)
    end
  end
end
```

It feels more descriptive and simpler to understand in my opinion. And it uses less mocking which makes it a little slower.

cbliard: On the other hand, it's only testing that a method has been called. The test to check that the language of users is actually changed is in another file (`spec/controllers/admin/settings/languages_settings_controller_spec.rb`). I wonder if this test is useful then.

aaron-contreras: I guess it might have some use in testing the alternate code path of not resetting users' languages unless successful and the controller test does only concern itself with the "happy path"?

cbliard: I think it's ok like this.
But generally, I like the unit test to test the thing properly, and then a higher-level test that tests the same code path could use some mocks, as it knows that the thing it mocks is already working as expected.

In this case, I think the test of `Settings::LanguageUpdateService` should test with real users that their language is reset, and the one testing `Admin::Settings::LanguagesSettingsController` simply ensuring that a `Settings::LanguageUpdateService` is instantiated and called is enough.

But in the end, both work.

cbliard (approved):
:ok_hand: Thanks!

Merging as the failing spec `./spec/features/wysiwyg/non_breaking_spaces_spec.rb:49` is passing locally. Probably a flickering one.

spec/services/settings/language_update_service_spec.rb
aaron-contreras: I'll keep this train of thought in mind; it makes sense in my opinion. Thanks for the having this discussion with me :)

PR #12850, opened 2023-06-13
Add share_calendars permission for roles having view_calendar

cbliard:
https://community.openproject.org/wp/48432


cbliard:
Hi @jjabari-op, 

The `share_calendars` was properly added to roles having the `view_calendar` permission as a migration, but in a newly installed instance, as the migrations run before the roles are created through seeding, the `share_calendars` permission would not be added to any roles (except project admin as it gets all assignable permissions).

To fix it, the permission must be added during the seeding, which is exactly what this PR does.

jonasjabari:
@cbliard thank you and sorry for not having seen this myself!

PR #12865, opened 2023-06-14
Feature/47512 access tokens settings improvement

jjabari-op:
https://community.openproject.org/projects/openproject/work_packages/47512


app/views/my/access_token_partials/_api_tokens_section.html.erb
```diff
@@ -0,0 +1,90 @@
...
+  </div>
+  <div>
+    <p>
+      <%= t("my_account.access_tokens.api.text_hint") %>
+    </p>
+    <% if Setting.rest_api_enabled? %>
+      <% if api_token %>
+        <div class="generic-table--container">
+          <div class="generic-table--results-container">
+            <table id="access-token-table" class="generic-table">
```
ulferts: This id is assigned to every table on the page resulting multiple elements on the page having the same id. I think it can simply be removed.

app/views/my/access_token_partials/_api_tokens_section.html.erb
```diff
@@ -0,0 +1,90 @@
...
+                            column_headers: [
+                              t('attributes.name'), 
+                              User.human_attribute_name(:created_at), 
+                              t('my_account.access_tokens.headers.expiration')
+                            ] 
+                          } 
+              %>
+              <tbody>
+                <tr>
+                  <td style="width: 60%;"><%= t("my_account.access_tokens.api.static_token_name") %></td>
```
ulferts: The use of inline styles for static behaviour is discouraged in OpenProject. I do feel a bit bad about saying this as it might change with the inclusion of Primer soon (although I haven't found a simple table component in the DS but maybe I was just searching for the wrong string).

Not so much at this place but at the other tables (i.e. the iCalendar one) I would also prefer a more flexible approach as the Name, Calendar and Project table might vary in length depending on the contents.

config/locales/crowdin/af.yml
```diff
@@ -269,7 +269,7 @@ af:
       notice_reset_token: "A new %{type} token has been generated. Your access token is:"
       token_value_warning: "Note: This is the only time you will see this token, make sure to copy it now."
       no_results_title_text: There are currently no access tokens available.
-      notice_ical_tokens_reverted: "iCalendar token \"%{token_name}\" for calendar \"%{calendar_name}\" of project \"%{project_name}\" has been revoked. The iCalendar URL with this token is now invalid."
+      notice_ical_token_revoked: "iCalendar token \"%{token_name}\" for calendar \"%{calendar_name}\" of project \"%{project_name}\" has been revoked. The iCalendar URL with this token is now invalid."
```
ulferts, line 270: Doesn't really matter but altering the files in `config/locales/crowdin/` is superfluous. They all are based on the `en.yml` and `js-en.yml` files, transmitted to Crowdin and then pulled from there again after being translated. So the line would be removed by the next pull from Crowdin anyway, as the key is removed from `en.yml`.

config/locales/en.yml
```diff
@@ -1521,6 +1523,26 @@ en:
...
+        disabled_text: "API tokens are not enabled by the administrator. Please contact your administrator to use this feature."
+      ical:
+        title: "iCalendar"
+        text_hint: "iCalendar tokens allow users to <a href=\"%{path}\" target=\"_blank\">subscribe to OpenProject calendars</a> and view up-to-date work package information from external clients."
+        disabled_text: "iCalendar subscriptions are not enabled by the administrator. Please contact your administrator to use this feature."
+        empty_text_hint: "To add an iCalendar token, subscribe to a new or existing calendar from within the Calendar module of a project. You must have the necessary permissions."
+      oauth:
+        title: "OAuth"
+        text_hint: "OAuth tokens allow third-party applications to connect with this OpenProject instance."
+        empty_text_hint: "There is no third-party application access configured and active for you. Please contact your administrator."
```
ulferts: I would rephrase the part 'Please contact your administrator' as it sounds as if the user should actually do it now because of some fault being present. I believe the phrasing used above:  'Please contact your administrator to use this feature.' or 'Please contact your administrator to activate this feature.' would be clearer.

spec/features/users/my_spec.rb
```diff
@@ -124,115 +124,303 @@ def expect_changed!
...
+                expect(page).to have_xpath("#{row_xpath}/td[2]/a", text: query.name)
+                expect(page).to have_xpath("#{row_xpath}/td[3]", text: query.project.name)
+              end
+            end
+          end
+
+          it 'single iCalendar tokens can be deleted' do
+            visit my_access_token_path
+            token_name = ical_token_for_query.ical_token_query_assignment.name
+            row_xpath = "//tr[td[1][contains(text(), '#{token_name}')]]"
```
ulferts: To avoid complicated statement like this, we introduced data attributes to be able to more easily target a DOM node in the specs: `data-qa-selector`

ulferts (changes requested):
Hi @jjabari-op, only small things to report (please see the code).

Among the feature specs failing, one is actually a correct positive: `rspec ./spec/features/oauth/authorization_code_flow_spec.rb:62` 

I guess it needs to be adapted.

It might also help to merge in the latest dev again to get more stable feature spec results.

app/views/my/access_token_partials/_api_tokens_section.html.erb
jjabari-op: I decided to set the table column widths statically as we show multiple of these tables on the access token page and I wanted them to have the same optical structure. In my opinion it's easier to digest the information when the table columns appear on the same horizontal position:
Screenshot from 2023-07-20 09-58-26
If the table columns have dynamic widths, the horizontal position of the "created on" and "expires" column will be different for each table.

config/locales/crowdin/af.yml
jjabari-op: I needed to adapt some translation files as GitHub raised merge conflicts while merging the dev branch back into this feature branch. I kind of thought as well that the crowdin workflow will make these changes obsolete at the end.

app/views/my/access_token_partials/_api_tokens_section.html.erb
ulferts: I am ok with having static table widths for the use case you describe (multiple tables on the same pabe). But please do it using css classes as opposed to inline styles as e.d. done for the [subject column in the wp table where the `-max` modifier is applied](https://github.com/opf/openproject/blob/dev/frontend/src/global_styles/content/_table.sass#L137-L139). Some similar, maybe page specific css, can govern the access tokens page, which would also ensure that all of those keep the same value which is currently spread out between partials.

ulferts:
@jjabari-op can you please resolve the conflicts and push again?

jjabari-op:
@ulferts I've implemented your feedback and fixed the affected spec.

Furthermore, I've re-enabled the hook without renaming it and wrapped it with an empty table, which is not visible when the hook is not used. I guess this is the least breaking change for anybody using this hook.

ulferts (approved):
Works and is looking shiny ✨

PR #12866, opened 2023-06-14
include schema into attributesByTimestamp

ulferts:
Include the reference to the schema valid for each attributeByTimestamp element. This is necessary as the work package schema is changed whenever the project or the type changes which might happen between those elements. 

https://community.openproject.org/wp/48567

### DIFF
--- a/spec/lib/api/v3/work_packages/work_package_collection_representer_spec.rb
+++ b/spec/lib/api/v3/work_packages/work_package_collection_representer_spec.rb
@@ -30,9 +30,11 @@ require 'spec_helper'
 
 RSpec.describe API::V3::WorkPackages::WorkPackageCollectionRepresenter do
   include API::V3::Utilities::PathHelper
+  create_shared_association_defaults_for_work_package_factory
 
   let(:self_base_link) { '/api/v3/example' }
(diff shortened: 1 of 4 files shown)


spec/lib/api/v3/work_packages/work_package_collection_representer_spec.rb
```diff
@@ -534,13 +532,58 @@
...
         .to be_json_eql(expected_path.to_json)
         .at_path('_embedded/schemas/_embedded/elements/0/_links/self/href')
+
+      # All work packages share the same project and type
+      expect(collection)
+        .to have_json_size(1)
+              .at_path('_embedded/schemas/_embedded/elements')
+    end
+
+    context 'when having timestamps leads to a different schema at the baseline time (because of a different type)',
```
aaron-contreras, line 550: To avoid unnecessary context switching in this spec (and to just be able to grok through it a little better), how about defining

```ruby
let(:first_wp) { WorkPackage.first } # or work_packages[0]
```

and refer to `first_wp` throughout the entire context?

aaron-contreras (approved):
Just left a comment to help grok through the spec a little simpler (Non-essential)

PR #12876, opened 2023-06-15
bump storybook angular

as-op:
storybook is not building atm. 

bumping `@storybook/angular` to `7.1.0-alpha34`


oliverguenther (approved):
Thanks :bow:

PR #12882, opened 2023-06-15
Implement models instead of hashes as journal changes

klaustopher:
Instead of passing hashes around, we have well defined models for causes, that can be instantiated and passed around. This also solves that we had to introduce the "helper" struct we needed to introduce for the working day update

### DIFF
--- a/app/services/journals/create_service.rb
+++ b/app/services/journals/create_service.rb
@@ -870,7 +870,7 @@ module Journals
 
     def cause_sql(cause)
       # Using the same encoder mechanism that ActiveRecord uses for json/jsonb columns
-      ActiveSupport::JSON.encode(cause&.to_h || {})
+      ActiveSupport::JSON.encode(cause || {})
     end
 
     # Because we added the journal via bare metal sql, rails does not yet
@@ -906,7 +906,7 @@ module Journals
     end
 
     def same_cause?(predecessor, cause)
-      (predecessor.cause.blank? && cause.blank?) || predecessor.cause == cause.to_h
+      (predecessor.cause.blank? && cause.blank?) || predecessor.cause == cause
     end
 
     def log_journal_creation(predecessor)
--- a/app/services/journals/create_service.rb
+++ b/app/services/journals/create_service.rb
@@ -48,13 +48,9 @@ module Journals
       self.journable = journable
     end
 
-    def call(notes: "", cause: {})
-      # JSON columns read from the database always have string keys. As we do not know what is passed in here,
-      # and we want to compare it to values read from the DB, we need to stringify the keys here as well
-      normalized_cause = cause.deep_stringify_keys
-
+    def call(notes: "", cause: CauseOfChange::NoCause.new)
       Journal.transaction do
-        journal = create_journal(notes, normalized_cause)
+        journal = create_journal(notes, cause)
 
         if journal
           reload_journals
@@ -270,28 +266,28 @@ module Journals
 
     def cleanup_predecessor_attachable(predecessor)
       cleanup_predecessor(predecessor,
-                          'attachable_journals',
+                          "attachable_journals",
                           :journal_id,
                           :id)
     end
 
     def cleanup_predecessor_customizable(predecessor)
       cleanup_predecessor(predecessor,
-                          'customizable_journals',
+                          "customizable_journals",
                           :journal_id,
                           :id)
     end
 
     def cleanup_predecessor_storable(predecessor)
       cleanup_predecessor(predecessor,
-                          'storages_file_links_journals',
+                          "storages_file_links_journals",
                           :journal_id,
                           :id)
     end
 
     def cleanup_predecessor_agenda_itemable(predecessor)
       cleanup_predecessor(predecessor,
-                          'meeting_agenda_item_journals',
+                          "meeting_agenda_item_journals",
                           :journal_id,
                           :id)
     end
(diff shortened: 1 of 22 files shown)


lib_static/plugins/acts_as_journalized/lib/cause_of_change/base.rb
```diff
@@ -0,0 +1,56 @@
...
+
+    def ==(other)
+      if other.is_a?(Hash)
+        to_h == other
+      else
+        other.class == self.class &&
+         type == other.type &&
+         additional_attributes == other.additional_attributes
+      end
+    end
```
dombesz: Implementing the `==` method with the `to_hash` call would also yield the benefit of explicit conversion. Hence we  can omit explicitly calling `to_h` in case of comparison with another hash.
The calls such as `predecessor.cause == cause.to_h` could be just simply written as `predecessor.cause == cause`

```suggestion
     def ==(other)
      other.to_hash == to_hash
    end
```

This is just a convenience suggestion.

app/services/journals/create_service.rb
```diff
@@ -910,7 +906,7 @@ def same_user?(predecessor)
     end
 
     def same_cause?(predecessor, cause)
-      (predecessor.cause.blank? && cause.blank?) || predecessor.cause == cause
+      (predecessor.cause.blank? && cause.blank?) || predecessor.cause == cause.to_h
```
dombesz: Defining the implicit `to_hash` method on the [cause class](https://github.com/opf/openproject/pull/12882/files#diff-1850d4434c148aa7386b223a7a7589175eaf720b49f94d9a128ea328619ffe4dR38), would automatically do the `to_hash` conversion when it is compared against another hash. 
```suggestion
      (predecessor.cause.blank? && cause.blank?) || predecessor.cause == cause
```

lib_static/plugins/acts_as_journalized/lib/cause_of_change/base.rb
```diff
@@ -0,0 +1,56 @@
...
+    attr_reader :type, :additional_attributes
+
+    def initialize(type, additional_attributes = {})
+      @type = type
+      @additional_attributes = additional_attributes
+    end
+
+    def to_h
+      { "type" => type }.merge(additional_attributes).deep_stringify_keys
+    end
```
dombesz: Defining the implicit conversion method `to_hash`  instead of the explicit `to_h` yields the benefit of automatic type conversion based on the callee method.
```suggestion
    def to_hash
      { "type" => type }.merge(additional_attributes).deep_stringify_keys
    end
    
    alias_method :to_h, :to_hash # alias it just for the convenience if we ever want to use the explicit conversion
```

For example there is non need to explicitly call `to_h` in the `ActiveSupport::JSON.encode(cause&.to_h || {})` call. It can be written as `ActiveSupport::JSON.encode(cause || {})`, and the receiver method will implicitly take care of the conversion if the argument object responds to `to_hash`.

This implicit conversion also happens, when comparing the cause object against a hash. The comparator will implicitly calls `to_hash` on the non-hash object and will use returned value if the comparison succeeds. In case there is no `to_hash` method on the object, it will raise an error that it cannot compare the 2 objects.

app/services/journals/create_service.rb
```diff
@@ -874,7 +870,7 @@ def normalize_newlines_sql(column)
 
     def cause_sql(cause)
       # Using the same encoder mechanism that ActiveRecord uses for json/jsonb columns
-      ActiveSupport::JSON.encode(cause || {})
+      ActiveSupport::JSON.encode(cause&.to_h || {})
```
dombesz: Using the `to_hash` method on the cause class, this can be written as:

```suggestion
      ActiveSupport::JSON.encode(cause || {})
```

lib_static/plugins/acts_as_journalized/lib/cause_of_change/no_cause.rb
```diff
@@ -0,0 +1,43 @@
...
+  class NoCause < Base
+    def initialize
+      super(nil)
+    end
+
+    def blank?
+      true
+    end
+
+    def to_h
```
dombesz:
```suggestion
    def to_hash
```

dombesz (approved):
Congrats on your 100th PR Klaus! 🥳 I just left a minor suggestion to improve the code ergonomics.

klaustopher:
Awesome, will take a look. Christophe also recently added a new cause, so I'll need to create a model there as well.

klaustopher:
@dombesz Not really sure if it would be right to implement a `to_hash` method here. It will cause some implicit conversion to be triggered by this. I will play around with your suggestion but reading through https://zverok.space/blog/2016-01-18-implicit-vs-expicit.html I'm not sure if this is the best way forward.

klaustopher:
Hey @dombesz I have implemented your suggested changes. Looks a lot cleaner now. I have also added classes for the new causes that @cbliard has introduced. Would be cool if you could take another quick glance over the changes.

dombesz (approved):
Nice job @klaustopher, I really like the introduction of Cause classes! 👍
Shall we change the base branch to 14.0? Feel free to decide on that and merge it.

klaustopher:
I think there's not much need to rush this

cbliard (approved):
Nice one 👍 I like this refactoring a lot.

Did you consider naming `Journal::CausedBySystemUpdate` like `Journal::Cause::SystemUpdate` so that they like like "cause" objects rather than "causedBySomething" objects?

PR #12884, opened 2023-06-15
Baseline: join with project at the time of the journal

ulferts:
When joining projects, the condition to join against has to be the project_id of `work_package_journals` and not the project_id of the `work_packages` table. The `work_packages` table contains the current state whereas the `work_package_journals` contains the state at the journal time. 

By that, changes between projects are correctly accounted for which is important especially in the context of a timestamped work package query where the project (and its descendants) serves as a context the query is executed in. In such a case, a condition like `AND (projects.id IN (4, 5, 8, 12)` is added by the query and needs to be checked agains the projects the work package used to be in and not the projects the work package is in.

https://community.openproject.org/wp/48592

### DIFF
--- a/app/models/journable/historic_active_record_relation.rb
+++ b/app/models/journable/historic_active_record_relation.rb
@@ -109,7 +109,7 @@ class Journable::HistoricActiveRecordRelation < ActiveRecord::Relation
     relation = substitute_database_table_in_where_clause(relation)
     relation = add_timestamp_condition(relation)
     relation = add_join_on_journables_table_with_created_at_column(relation)
-    relation = add_join_projects_on_journables(relation)
+    relation = add_join_projects_on_work_package_journals(relation)
     relation = select_columns_from_the_appropriate_tables(relation)
 
     # Based on the previous modifications, build the algebra object.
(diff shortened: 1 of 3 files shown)


app/models/journable/historic_active_record_relation.rb
```diff
@@ -253,22 +253,25 @@ def journals_join_statement
...
-  # Join the projects table on journables if :project is in the includes.
+  # Join the projects table on work_package_journals if :project is in the includes.
   # It is needed when projects are filtered by id, and has to be done manually
   # as eager_loading is disabled.
+  # It needs to be `work_package_journals` and not `journables` (the subselect of the work_packages table)
+  # because the journables table will contain the current project and not the project the work package was
+  # in at the time of the journal.
   # Does not work yet for other includes.
   #
   def add_join_projects_on_journables(relation)
```
cbliard: The method name is not right anymore. 
(I probably would not have noticed if the comment had not been updated.)

cbliard (approved):
It works as expected.
I only have a tiny remark about a method name. Apart from that it can be merged.

app/models/journable/historic_active_record_relation.rb
ulferts: I renamed it to reflect the change. Thanks for spotting.

PR #12891, opened 2023-06-16
Move lefthook to the development group only

dombesz:
Remove `lefthook` gem from the test environment, it takes 60MB with extensions and it is not being used in the test env.


crohr (approved):
thanks!

PR #12905, opened 2023-06-19
[#48538] Add iCalendar sub-level

mrmir:
https://community.openproject.org/work_packages/48538

Some of the description text is commented out as the docs linked within still don't exist. I have created https://community.openproject.org/projects/openproject/work_packages/48724 to handle those changes separately once that documentation is available. The documentation link in the 'Date format' section will probably also have to be changed eventually once the docs are updated to reflect the new settings groups.


ulferts (approved):
The change is looking fine @mrmir.

I was a bit surprised to not see the heading of the first menu hierarchy reflected in the breadcrumb above as in the specification:

Specification:
image

Implementation:
image

However, that behaviour is consistent with how the rest of the administration works, so I think it is acceptable.

Please fix the merge conflicts and merge at will.

PR #12916, opened 2023-06-20
Introduce RBS + Steep

klaustopher:
Playing around with adding type signatures. Let's see where this goes

## Tasks
- [x] Introduce RBS, Stepp and rbs_rails
- [x] Create needed config files
- [x] Generate stuff for rails models (fb496a3)
- [x] Generate stuff for a bunch of other files `using rbs prototype rb ...`
- [ ] Fix all errors or find a way to mark them as "okay" for now
- [ ] Add a GitHub action that runs steep
- [ ] Add a GitHub action that automatically updates RBS from gems for dependabot PRs (along the lines of https://github.com/klaustopher/pddln/blob/a09473a94e8553ff079ae5b061c7749920a9dc3f/.github/workflows/update-rbi.yml)
- [ ] Add a `lefthook` to run `steep check` as a before commit hook

## Notes
- The repository at https://github.com/ruby/gem_rbs_collection seems to have quite a few outdated definitions for gems. Half of rails is still stuck at 6.0. There seems to be a project out there thatt tries to make a better generator than what is included in `rbs collection`: https://github.com/ksss/orthoses
- The meta-tags gem we use ships with definitions, that cause a problem. There is an issue (https://github.com/kpumuk/meta-tags/issues/253), but this is pretty stale. It seems there is currently no way to ignore known problems in the type checker. Investigating further

---

Implements https://community.openproject.org/work_packages/47208


klaustopher:
Links:

- https://evilmartians.com/chronicles/climbing-steep-hills-or-adopting-ruby-types
- https://github.com/ruby/typeprof/blob/master/doc/doc.md

PR #12935, opened 2023-06-22
chore(View Components): Rename base class to `ApplicationComponent`

akabiru:
The name `RailsComponent` was a result of the now deprecated "RailsCells" which have been replaced by view components. ApplicationComponent is more conventional now.


dombesz (approved):
Thanks Kabiru! 👍

PR #12941, opened 2023-06-23
Replace Selenium with Cuprite

aaron-contreras:
This PR replaces Selenium with Cuprite in a [Strangler Fig Pattern approach](https://martinfowler.com/bliki/StranglerFigApplication.html).

Selenium and Cuprite live side by side and each example group/example defines whether
to use cuprite or not with the `:with_cuprite` metadata.

The short/medium term goal is to transition all our suite to running on Cuprite. However, we are at a point
right now where we can start benefiting from the performance improvement.

**Results:**
Our test suite is now consistently running in the 7 minute range with setup steps cached, and in the 9 minute range when setup steps are uncached.

Read more about [Cuprite](https://github.com/rubycdp/cuprite)

### DIFF
--- a/spec/features/custom_fields/activate_in_project_spec.rb
+++ b/spec/features/custom_fields/activate_in_project_spec.rb
@@ -29,7 +29,7 @@
 require 'spec_helper'
 require 'support/pages/custom_fields'
 
-RSpec.describe 'custom fields', js: true do
+RSpec.describe 'custom fields', js: true, with_cuprite: true do
   let(:user) { create(:admin) }
   let(:cf_page) { Pages::CustomFields.new }
   let(:for_all_cf) { create(:list_wp_custom_field, is_for_all: true) }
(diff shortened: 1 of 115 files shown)


spec/features/custom_fields/activate_in_project_spec.rb
```diff
@@ -29,23 +29,22 @@
...
 require 'support/pages/custom_fields'
 
-RSpec.describe 'custom fields', js: true do
+RSpec.describe 'custom fields', js: true, with_cuprite: true do
   let(:user) { create(:admin) }
   let(:cf_page) { Pages::CustomFields.new }
   let(:for_all_cf) { create(:list_wp_custom_field, is_for_all: true) }
   let(:project_specific_cf) { create(:int_wp_custom_field) }
   let(:work_package) do
     wp = build(:work_package).tap do |wp|
```
cbliard, line 38:
```suggestion
    build(:work_package).tap do |wp|
```

spec/features/notifications/settings/pause_reminder_settings_spec.rb
```diff
@@ -18,8 +20,6 @@
                                          first:,
                                          last:)
 
-      sleep 2
-
```
cbliard, line 22: I love seeing this

cbliard (approved):
Nice job :+1:

spec/features/custom_fields/activate_in_project_spec.rb
cbliard: but that's a detail

PR #12948, opened 2023-06-26
Added support for multi-value Version custom fields

virgo47:
- in admin, "Allow multi-select" is available for version CF
- added support for display/edit fields
- added support for multi-value in Activity formatting


virgo47:
Hello, we discussed this contribution via your support channel, and this is the result.

In custom fields administration:
image

In WP - display and edit:
image
image

And support for all values in Activities output:
image

We believe this will be very useful for many users/customers.
If something is missing, please, let me know. Thank you.

oliverguenther:
Hi @virgo47, looking good. To get this merged, we'll need a multi-select version spec similar to the user one: 

`spec/features/custom_fields/multi_user_custom_field_spec.rb`

oliverguenther:
https://community.openproject.org/work_packages/48745

virgo47:
Hi @oliverguenther, I added the test.

However, I'd welcome advice as I struggle a bit with failing project_attributes_activity_spec.rb. Normally, I see the custom version field in project activities just fine, but in the test, it seems, permission was missing. `Version.visible` (added by me in `custom_field.rb`) depends on `:view_work_packages`, I tried to add it to the initial `member_with_permissions` array in the test, but it still fails. I can skip the `visible` call, but it seems to be appropriate there (similar to User resolution in the same file).

Do you have some tip why the user in the test still doesn't see the version? I was trying to debug it with various printouts, but I still don't get it. Or can we assume that if the user knows Version id, it can know its name? It seems it worked that way previously anyway, but was it right?

I also have some less serious observations about the `multi_user_custom_field_spec.rb`, but that is not related to my PR per se and doesn't block it, so that can wait at this moment.

oliverguenther:
Hi @opf/backend, can someone take over the review here and help with the spec? Thanks 🙏

cbliard:
Hi @virgo47, 

Your analysis is right, the user is missing `:view_work_packages` permission. 

The reason why adding it to `member_with_permissions` array does not work is because the `work_package_tracking` module is not enabled in the project. Only `activity` is enabled. As [the `:view_work_packages` permission belongs to the `work_package_tracking`](https://github.com/opf/openproject/blob/d123e663ad7b37f1654e591ba01e26f0e7fc6592/config/initializers/permissions.rb#L138-L147), it needs to be enabled to make the associated permissions work.

So
```
  let(:user) do
    create(:user,
           member_in_project: project,
           member_with_permissions: %w[view_wiki_pages
                                       edit_wiki_pages
                                       view_wiki_edits
                                       view_work_packages])
  end
```
and 
```
  let(:project) { create(:project, parent: parent_project, active: false, enabled_module_names: %w[activity work_package_tracking]) }
```

will fix the test.

Please can you also modify the test data so that there are multiple versions set?

Do you need help with rubocop too?

cbliard:
And thinking about it, the test is not about permissions and enabled modules, so we should remove them: they make the test harder to understand and maintain.

You can remove `member_with_permissions` parameter, so that [it uses `%w[view_work_packages edit_work_packages]` by default](https://github.com/opf/openproject/blob/d123e663ad7b37f1654e591ba01e26f0e7fc6592/spec/factories/principal_factory.rb#L66-L68), and you can remove `enabled_module_names` parameter so that [it enables all modules by default](https://github.com/opf/openproject/blob/d123e663ad7b37f1654e591ba01e26f0e7fc6592/spec/factories/project_factory.rb#L41C32-L41C32):
```
  let(:user) { create(:user, member_in_project: project) }
  let(:project) { create(:project, parent: parent_project, active: false) }
```

cbliard:
@virgo47 I pushed a commit with the modifications I talked about. There are still the rubocop annotations which need to be fixed.

virgo47:
Thanks a lot, I'll look at that the next week, holiday time atm.

virgo47:
@cbliard I believe the rubocop should be fine now, at least locally it was.

virgo47:
Hi guys (@cbliard, @oliverguenther), we've got the conflict caused by the recent dev commit a5d5b5c6, should I resolve it or do you want to do it? If I should do it, what is your preferred way? Normally, I'd merge dev into the feature branch to see that the feature branch still works fine and the divergence is minimal.

oliverguenther (approved):

oliverguenther:
Thanks @virgo47 for your work (and patience on getting our feedback). 👏

virgo47:
Big thanks to you guys as well, this is extremely important for us and I hope it will help others as well.

PR #12965, opened 2023-06-27
Changes to documentation

MayaBerd:
Calendar FAQ - update non-working days feature availability 
Repositories - removed duplicate paragraph
Time and cost reports export - re-formulated the export process 
Time and costs FAQ - updated the possibility of logging time for other users 
Time Tracker integration - added wording to the last step


psatyal:
@MayaBerd Fixed the docs re "Set and change dates and duration of work packages" so that it is now coherent with how the date picker actually behaves.

PR #12967, opened 2023-06-27
Add some missing specs

mrmir:


modules/calendar/spec/features/calendar_sharing_spec.rb
```diff
@@ -59,58 +61,109 @@
...
+  #         page.click_button("Subscribe to iCalendar")
+
+  #         # modal should not be shown
+  #         expect(page).not_to have_selector('.spot-modal--header', text: "Subscribe to iCalendar")
+  #       end
+  #     end
+  #   end
+  # end
+
+  context 'with sufficient permissions and the ical_enabled setting enabled', with_settings: { ical_enabled: true } do
```
ulferts: Hey @mrmir, this line still has the `with_settings: { ical_enabled: true }` meta info for rspec which will lead to the setting being mocked. The scope of this context includes the specs where you have been trying to alter the setting. Removing it will lead to the specs passing.

oliverguenther (approved):

PR #12974, opened 2023-06-28
[Passion Feature] OP#48583 Add emojis mentions in OpEditor triggered by `:`

akabiru:
https://community.openproject.org/work_packages/48583

https://github.com/opf/commonmark-ckeditor-build/pull/82

Screenshot 2023-06-28 at 3 10 28 PM


oliverguenther (approved):
LGTM. I've seen you tested this upstream. My general approach to CKEditor feature customizations were that we would test them in a feature spec downstream, to make sure integration in OpenProject doesn't break due to external factors.

In this case, this is only a CKEditor internal feature without any references to OpenProject, so I'm fine with merging as is.

akabiru:
> LGTM. I've seen you tested this upstream. My general approach to CKEditor feature customizations were that we would test them in a feature spec downstream, to make sure integration in OpenProject doesn't break due to external factors.
> 
> In this case, this is only a CKEditor internal feature without any references to OpenProject, so I'm fine with merging as is.

Good looking out! ~Working~ will add a feature spec for this one, would be great to cover both ends. 👍🏾

PR #12986, opened 2023-06-29
Add `pry: true` aliases for quick and simple debugging

aaron-contreras:
Sets up aliases to examples/example groups in order to automatically set a `binding.pry` statement in the state of the spec when an expectation fails and debug why/what failed.

These failures usually occur deep inside the codebase and to debug it, you need to concern yourself with adding the `binding.pry` statements **in the right places** and ensuring you clean up afterwards.

With the `pry: true` metadata, these headaches/tediousness is gone

This is akin to using `fit` to set `focus: true` on an example.

### DIFF
--- a/spec/support/pry.rb
+++ b/spec/support/pry.rb
@@ -1,0 +1,61 @@
+# frozen_string_literal: true
+
+# -- copyright
+# OpenProject is an open source project management software.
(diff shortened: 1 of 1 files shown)


cbliard (approved):
Nice addition :+1:

PR #12996, opened 2023-06-30
Fix ng-select dropdown repositionning

cbliard:
I noticed the following javascript error while debugging something else:

```
Uncaught TypeError: component.dropdownPanel._updatePosition is not a function
    at op-autocompleter.component.ts:287:44
```

It looks like `_updatePosition` has been split into `_updateXPosition` and `_updateYPosition` in this [ng-select commit](https://github.com/ng-select/ng-select/commit/348c1179763f982e9a4c69b1013705110aca651a) on Oct 6, 2020.

I replaced the `_updatePosition` calls to use `_updateXPosition` and `_updateYPosition` instead.
I'm not sure how to extract a method to factorize this code. Any help appreciated.
I'm not sure about the eslint errors either.


oliverguenther (approved):
Thanks for looking into this, I noticed the error but somehow lost track of it.

I pushed a commit to extract it into a helper method, and get rid of the eslint errors except the dangling, which we can't avoid.

PR #13042, opened 2023-07-05
[48966] Refactored api group model

Kharonus:
- https://community.openproject.org/work_packages/48966


docs/api/apiv3/components/schemas/group_write_model.yml
```diff
@@ -0,0 +1,27 @@
...
+            - description: |-
+                A new member for the group.
+                
+                **Resource**: User
+
+examples:
+  - name: Emperor's guard
+    _links:
+      members:
+        - href: 'api/v3/users/42'
```
machisuji: These are missing a leading slash. Should be `/api/v3/users/42` etc.

machisuji (commented):
Do you know if there is any way to get a more useful example in the default body when creating a group? Right now it looks like this:

image

What would be more useful would be this:

image

You do define sensible examples in the write model yaml but I wonder how or if we can get them to be used by the openapi explorer.

machisuji:
Other than the not very helpful examples in both the requests and responses for update/create group it's all good! Thanks for the refactoring @Kharonus! 🦸

Kharonus:
I already noticed the not correctly used examples. Will give it a bit of investigation.

docs/api/apiv3/components/schemas/group_write_model.yml
Kharonus: fixed

Kharonus:
Interestingly, the examples must be written in singular form for openapi-explorer to render them correctly. Sadly this is kind of conflicting with the update rules for OAS3.1 (https://www.openapis.org/blog/2021/02/16/migrating-from-openapi-3-0-to-3-1-0).

There it says, we should replace basically EVERY example inside a schema by examples. But I noticed, that our redocly cannot validate examples e.g. when inside error models, and apparently openapi-explorer cannot render them. Will need to investigate a bit, and revert my examples changes for now.

ulferts:
> Interestingly, the examples must be written in singular form for openapi-explorer to render them correctly. 

How does that relate to the way is is handled for the [`View`-Resource](https://github.com/opf/openproject/blob/dev/docs/api/apiv3/paths/view.yml#L16-L20) It is the only case for having multiple examples that I found but there it seems to be working fine:

image

Is that a different case?

ulferts (approved):
Change is looking fine. Thanks for picking it back up again @Kharonus .

Kharonus:
Aha, good finding with the multiple examples. This is something, I can use for further changes.

PR #13045, opened 2023-07-05
Improve Docker TLS setup & docs

b12f:
Instead of using `extra_hosts` with `host-gateway`, step-ca now finds traefik via the internal network by setting extra aliases on the traefik container. This allows usage without modifying your system firewall.

Also improves the docs slightly.


docker/dev/tls/docker-compose.override.example.yml
```diff
@@ -3,8 +3,7 @@ services:
...
       - /etc/ssl/certs/ca-certificates.crt:/etc/ssl/certs/ca-certificates.crt:ro
-
-  step:
-    # add any host here run in a docker service, that need a TLS certificate
-    extra_hosts:
-      - "openproject.local:host-gateway"
+    networks:
+      external:
+        aliases:
+          - openproject.local
```
Kharonus, line 13: Can you shortly elaborate, why this alias has to be `openproject.local`? I mean, I know it, but for documentation sake we could add a comment here, telling, that for TLS validation step-ca needs to go through traefik to reach the correct target.

Nice find! 👍🏽

docker/dev/tls/docker-compose.yml
```diff
@@ -28,10 +28,6 @@ services:
     depends_on:
       step:
         condition: service_healthy
-    logging:
-      options:
-        max-size: "8m"
-        max-file: "10"
```
Kharonus, line 34: Can you specify the errors thrown by the logging options? @apfohl has a valid point, when he states, he introduced them to reduce produced logs and reduce occupied space on disk.

docs/development/development-environment-docker/README.md
```diff
@@ -350,6 +355,8 @@ security.pki.certificateFiles = [ path_to_root_ca.crt ];
 ```
 
 Then rebuild your system. After that the generated root CA should be inside `/etc/ssl/certs/ca-certificates.crt`.
+However, you cannot mount this directly into the containers, since it's just a link. Use `readlink -f` to find out where
+the link points in your nix store, and mount that file directly.
```
Kharonus: @ba1ash Didn't you mount the bundle directly? Why does the volume mount is different for you and @b12f ?

docs/development/development-environment-docker/README.md
```diff
@@ -385,6 +392,16 @@ After amending the override file and the `.env`, ensure that you restart the sta
 docker compose up -d frontend
 ```
 
+### Adding a new service
+
+Some development tasks require you to run separate services that interact with OpenProject. For example, you might want
+to have Nextcloud running to test the Nextcloud-OpenProject integration. To do this, you'll need to follow some steps:
+
+1. Add the Nextcloud service to your `docker-compose.override.yml`, with the appropriate traefik labels, network, and
+   ca-bundle mounted.
```
Kharonus, line 399: I'd challenge the idea of putting additional services into your core override. I'd at least show the second option, too: adding new services as seperate stacks (different docker compose, same external network)

Kharonus (changes requested):
Let's discuss the open questions, but I really like the way it evolves!

docker/dev/tls/docker-compose.yml
b12f:
```
Error response from daemon: unknown log opt 'max-file' for journald log driver
```

Same goes for `max-size`. I guess this is some version incompatibility somewhere. I can see why you'd do this, but it might be better to have it in the overrides then, especially because we don't enable this for every service.

docs/development/development-environment-docker/README.md
b12f: Feel free to do this, but I think that using the docs to explain the simplest possible setup is best, and then people can still fumble it all into different stacks with separate networks yada yada if they like

docker/dev/tls/docker-compose.override.example.yml
b12f: Done, good idea :+1:

docs/development/development-environment-docker/README.md
ba1ash: I tried to mount the link directly and it works for me. So, I switched to this approach.

b12f: Really? unless you mounted the whole nix store the link should be pointing at `/nix/store/...` which doesn't exist inside the container. I ran into this exact issue when trying it myself. Does `docker compose exec backend bash -c "cat /etc/ssl/certs/ca-bundle.crt | grep Open"` produce output?

ba1ash: Yes, it works for me..
image

Kharonus: this might be important to check it first

b12f: Hmm I tried now and you're right it works. Maybe this is new docker behavior?

ba1ash: > Hmm I tried now and you're right it works. Maybe this is new docker behavior?

Maybe. I did not manage to find out whether it is new or old behaviour.

Kharonus (approved):
LGTM

PR #13056, opened 2023-07-06
Add a lefthook config to run bin/setup_dev after a `git pull`

klaustopher:
I added a `post-merge` hook into the lefthook config, that will run `bin/setup_dev` when one of the following conditions apply:

- `Gemfile.lock` was modified in the merge
- `frontend/package-lock.json` was modified in the merge
- a migration was added to `db/migrate/` or any modules migration folder

`post-merge` might be confusing in name, but when running `git pull` this is basically doing a `git fetch && git merge` of the remote branch, so this will apply when using `git pull` as well. See [git docs](https://git-scm.com/docs/githooks#_post_merge)

### DIFF
--- a/lefthook.yml
+++ b/lefthook.yml
@@ -17,3 +17,10 @@ pre-commit:
       files: git diff --name-only --staged
       glob: "*.rb"
       run: bin/dirty-rubocop --uncommitted --force-exclusion {files}
+
+post-merge:
+  commands:
+    setup-dev:
+      files: git diff --name-only HEAD@\{1\} HEAD
+      glob: '{Gemfile.lock,frontend/package-lock.json,db/migrate/*.rb,modules/*/db/migrate/*.rb}'
+      run: bin/setup_dev
--- a/lefthook.yml
+++ b/lefthook.yml
@@ -20,7 +20,30 @@ pre-commit:
 
 post-merge:
   commands:
-    setup-dev:
+    bundle-install:
       files: git diff --name-only HEAD@\{1\} HEAD
-      glob: '{Gemfile.lock,frontend/package-lock.json,db/migrate/*.rb,modules/*/db/migrate/*.rb}'
-      run: bin/setup_dev
+      glob: Gemfile.lock
+      run: bin/bundle install
+    tasks:
+      run: lefthook run parallel-post-merge-jobs
+
+parallel-post-merge-jobs:
+  parallel: true
+  commands:
+    npm-install:
+      files: git diff --name-only HEAD@\{1\} HEAD
+      glob: frontend/package-lock.json
+      root: "frontend/"
+      run: npm ci
+    migrate:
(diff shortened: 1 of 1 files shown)


lefthook.yml
```diff
@@ -17,3 +17,10 @@ pre-commit:
       files: git diff --name-only --staged
       glob: "*.rb"
       run: bin/dirty-rubocop --uncommitted --force-exclusion {files}
+
+post-merge:
+  commands:
+    setup-dev:
+      files: git diff --name-only HEAD@\{1\} HEAD
+      glob: '{Gemfile.lock,frontend/package-lock.json,db/migrate/*.rb,modules/*/db/migrate/*.rb}'
+      run: bin/setup_dev
```
Kharonus: That won't work of course for devs with docker setup. So, please, announce it, once it is merged, so everyone can adapt their `lefthook-local.yml`

cbliard:
I'm not using `bin/setup_dev`, and that's ok as I can disable it with this in `lefthook-local.yml`:
```yaml
post-merge:
  commands:
    setup-dev:
      run: true
```

I wonder if there could be a way to leverage lefthook's ability to [run multiple processes in parallel](https://github.com/evilmartians/lefthook#parallel-execution).

lefthook.yml
Kharonus: you know by chance out of your head, you you could switch off a lefthook hook inside the local override? If I do not want to run the setup?

UPDATE: @cbliard provided the answer, thanks :D

klaustopher:
Of course, it would be possible to "extract" all the things that are done in `setup_dev` to run them isolated:

```yaml
post-merge:
  commands:
    bundle-install:
      files: git diff --name-only HEAD@\{1\} HEAD
      glob: Gemfile.lock
      run bundle install
    npm-install:
      files: git diff --name-only HEAD@\{1\} HEAD
      glob: frontend/package-lock.json
      root: "frontend/"
      run: npm ci
    migrate:
      files: git diff --name-only HEAD@\{1\} HEAD
      glob: '{db/migrate/*.rb,modules/*/db/migrate/*.rb}'
      run: bin/rails db:migrate
    extract-locales:
      files: git diff --name-only HEAD@\{1\} HEAD
      glob: '{config/locale/**/*,modules/*/config/locale/**/*}'
      run: bin/rails i18n:js:export
    register-frontend-js:
      files: git diff --name-only HEAD@\{1\} HEAD
      glob: 'modules/*/frontend/**/*'
      run: bin/rails openproject:plugins:register_frontend
```

It extracts the actions and would need to be manually maintained along the setup script. But I think for the few actions this would be fine.

`parallel` cannot be used as the `bin/rails` commands require the bundle to have been executed

cbliard (approved):
I like the way you managed to make the commands more atomic, and run them in parallel.

I would just change `bin/bundle install` to `bin/bundle check || bin/bundle install` which could be faster and put less garbage on screen if everything is already ok.

PR #13071, opened 2023-07-07
Fix/align headers on user activity page

cbliard:
Maya noticed that headers are not aligned on the users activity page.

Indeed there is a space where the user email and the custom fields would be displayed, but it is still eating place if there is nothing.
This PR removes the extra space when it's not allowed to expose user's email and there are no visible custom values.

Before:

image

After:

image


cbliard:
Failing tests are flaky ones, so the build is ok. I'll merge it.

PR #13072, opened 2023-07-07
Fix flaky test by adding change detection

cbliard:
Failing test was ./spec/features/work_packages/datepicker/datepicker_logic_spec.rb:168

The " days" suffix was not appended to the duration field. Duration field text is controlled with `[ngModel]="durationFocused ? duration : displayedDuration"`, but when `durationFocused` is changed, it sometimes is not detected.

Forcing the detection fixes it.


cbliard:
The failing test is another flaky one. I'll merge this one as it improves global build stability.

PR #13078, opened 2023-07-10
Inline some rspec-wait into our code

cbliard:
Doing so to remove the usage of Timeout.timeout as its behavior is unpredictable.


mereghost (approved):
Nice solution! 🎉

cbliard:
The failing test is a flaky one. Merging.

PR #12924, opened 2023-06-21
[Passion Chore] Adopt new Attribute Group (view) Components

akabiru:
Followup of #12917 replace markup with component calls

- [x] `oauth/applications/show`
- [ ] `app/views/admin/info`
- [ ] `app/views/users/change_status_info`
- [ ] `app/views/ldap_groups`
- [ ] `app/views/two_factor_authentication`
- [ ] Check up on Primer!

PR #13081, opened 2023-07-10
fix: align attribute groups across the application

b12f:
Attribute groups on the storages and oauth application admin pages looked different than on the work package pages. This was introduced by https://community.openproject.org/projects/openproject/work_packages/47680/activity

To align them, this commit reduces the scope of the `wp-attribute-group` css, instead applying those styles to the more general `attributes-group` block. It also fixes the HTML structure of the backend-rendered attribute groups to align with those in the frontend.

Ref https://community.openproject.org/work_packages/48640/activity


bsatarnejad (commented):
Hi @b12f ,
Thanks for the changes,
Are these changes applied to the OAuth application too? I just tested it and saw that attribute key values are still rendered horizontally.

Screenshot from 2023-07-12 08-27-09
Another issue is that the key and value are not aligned completely:

Screenshot from 2023-07-12 08-35-20

akabiru (commented):
Nice one @b12f 🙌🏾 

RE: @bsatarnejad's note in https://github.com/opf/openproject/pull/13081#pullrequestreview-1525616650 not all backend rendered views utilize the `AttributeGroupComponent` yet  - here's a list from the pending passion chore https://github.com/opf/openproject/pull/12924 I'd be happy to assist if you're tight on time.

- [x] `oauth/applications/show`
- [x] `app/views/admin/info`
- [x] `app/views/users/change_status_info`
- [x] `app/views/ldap_groups`
- [x] `app/views/two_factor_authentication`

oliverguenther:
I converted the remaining pages to use the same component. But the position of key - value seems off to me:

image

ulferts:
I found even more occurrences of the DOM structure and applied the component to them.

I also added a quick fix to almost get the key and value on the same height.

PR #13083, opened 2023-07-10
[#48966] refactored help text api models

Kharonus:
- https://community.openproject.org/work_packages/48966


ulferts (approved):
Thanks, @Kharonus

PR #13086, opened 2023-07-10
[#48647] PDF Export: Manage long query titles in footer of pdf export

as-op:
* refactor to draw all footer parts at once
* calculate available widths for footer parts
* implement footer line wrap
* implement multiline draw

https://community.openproject.org/work_packages/48647


ulferts (approved):
The change does what it advertises. 

I am not 100% myself whether the technical solution is actually the best one for the user. 

image

In the extreme, this will take away quite some space on every page. Maybe shortening the query name would be preferable.

as-op:
Thanks! I clarify with product about the truncation.

PR #13087, opened 2023-07-11
Fix flaky `team_planner_onboarding_tour_spec` test

aaron-contreras:
* Migrated the spec to using cuprite.
* Increased wait time on specific long-running expectation. It just
  takes a while to load as indicated in the already existing comment...

### DIFF
--- a/modules/team_planner/spec/features/onboarding/team_planner_onboarding_tour_spec.rb
+++ b/modules/team_planner/spec/features/onboarding/team_planner_onboarding_tour_spec.rb
@@ -30,8 +30,8 @@ require 'spec_helper'
 require_relative './../../support/onboarding/onboarding_steps'
 
 RSpec.describe 'team planner onboarding tour',
-               js: true,
-               with_cuprite: false,
+               :js,
+               :with_cuprite,
                with_ee: %i[team_planner_view] do
   let(:next_button) { find('.enjoyhint_next_btn') }
 
@@ -85,7 +85,7 @@ RSpec.describe 'team planner onboarding tour',
 
   after do
     # Clear session to avoid that the onboarding tour starts
-    page.execute_script("window.sessionStorage.clear();")
+    page.driver.cookies.clear
   end
 
   context 'as a new user' do
@@ -102,7 +102,7 @@ RSpec.describe 'team planner onboarding tour',
 
     it "I do not see the team planner onboarding tour in the scrum project" do
       # Set sessionStorage value so that the tour knows that it is in the scum tour
-      page.execute_script("window.sessionStorage.setItem('openProject-onboardingTour', 'startMainTourFromBacklogs');")
+      page.driver.set_cookie('openProject-onboardingTour', 'startMainTourFromBacklogs')
 
       # Set the tour parameter so that we can start on the wp page
       visit "/projects/#{scrum_project.identifier}/work_packages?start_onboarding_tour=true"
--- a/modules/team_planner/spec/support/onboarding/onboarding_steps.rb
+++ b/modules/team_planner/spec/support/onboarding/onboarding_steps.rb
@@ -29,22 +29,27 @@
 module OnboardingSteps
   def step_through_onboarding_team_planner_tour
     next_button.click
+    wait_for_reload
     expect(page).to have_text sanitize_string(I18n.t('js.onboarding.steps.team_planner.overview')), normalize_ws: true
 
     # The team planner is long to load
     next_button.click
+    wait_for_reload
     expect(page)
-      .to have_text sanitize_string(I18n.t('js.onboarding.steps.team_planner.calendar')), normalize_ws: true, wait: 10
+      .to have_text sanitize_string(I18n.t('js.onboarding.steps.team_planner.calendar')), normalize_ws: true, wait: 20
 
     next_button.click
+    wait_for_reload
     expect(page)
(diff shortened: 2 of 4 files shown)


aaron-contreras:
@akabiru This solves the flakiness on one of the specs you mentioned.

modules/team_planner/spec/features/onboarding/team_planner_onboarding_tour_spec.rb
```diff
@@ -30,8 +30,8 @@
 require_relative './../../support/onboarding/onboarding_steps'
 
 RSpec.describe 'team planner onboarding tour',
-               js: true,
-               with_cuprite: false,
+               :js,
+               :with_cuprite,
```
akabiru, line 34: cuprite ftw 🤩

akabiru (dismissed):
Nice one Aaron, looks great to me! 👍🏾  _ignored the failing spec as it seems to be another flaky one_

modules/team_planner/spec/features/onboarding/team_planner_onboarding_tour_spec.rb
```diff
@@ -85,7 +85,7 @@
 
   after do
     # Clear session to avoid that the onboarding tour starts
-    page.execute_script("window.sessionStorage.clear();")
+    page.driver.cookies.clear
```
klaustopher, line 88: `window.sessionStorage` is not the cookie jar, it's a session based local storage. Is that included in the  `cookies.clear` here?

modules/team_planner/spec/features/onboarding/team_planner_onboarding_tour_spec.rb
```diff
@@ -102,7 +102,7 @@
 
     it "I do not see the team planner onboarding tour in the scrum project" do
       # Set sessionStorage value so that the tour knows that it is in the scum tour
-      page.execute_script("window.sessionStorage.setItem('openProject-onboardingTour', 'startMainTourFromBacklogs');")
+      page.driver.set_cookie('openProject-onboardingTour', 'startMainTourFromBacklogs')
```
klaustopher, line 105: Same here. A cookie and the sessionStorage are not the same

modules/team_planner/spec/support/onboarding/onboarding_steps.rb
```diff
@@ -29,22 +29,27 @@
 module OnboardingSteps
   def step_through_onboarding_team_planner_tour
     next_button.click
+    wait_for_reload
```
cbliard, line 32: A method could be extracted here. Something like `next_tour_step` doing the click and the wait would be more intention revealing.

modules/team_planner/spec/support/onboarding/onboarding_steps.rb
```diff
@@ -29,22 +29,27 @@
...
     expect(page).to have_text sanitize_string(I18n.t('js.onboarding.steps.team_planner.overview')), normalize_ws: true
 
     # The team planner is long to load
     next_button.click
+    wait_for_reload
     expect(page)
-      .to have_text sanitize_string(I18n.t('js.onboarding.steps.team_planner.calendar')), normalize_ws: true, wait: 10
+      .to have_text sanitize_string(I18n.t('js.onboarding.steps.team_planner.calendar')), normalize_ws: true, wait: 20
```
cbliard, line 39: In b5b5b5bb (funny commit sha btw), I already raised timeout from 5 to 10, and now we're raising it from 10 to 20.

10 seconds seems to be already a lot of time. It's fine to raise it to 20, but then if it fails again sometimes, we'll have to search for a fix in another direction.

klaustopher: Waiting for 20 seconds for anything feels already reeeeeally excessive. Are we really expecting this JS to not finish before 20 seconds?

modules/team_planner/spec/features/onboarding/team_planner_onboarding_tour_spec.rb
aaron-contreras: I see. So, looks like cuprite (or Ferrum to be precise) doesn't provide access to the `sessionStorage`

```
     Failure/Error: page.driver.execute_script('window.sessionStorage.setItem("key", "value");')

     Ferrum::JavaScriptError:
       DOMException: Failed to read the 'sessionStorage' property from 'Window': Access is denied for this document.
```

However, now that you pointed this out, I noticed that these sessionStorage settings are not changing the spec's behavior at all (in the dev branch). I can remove both the lines that set the sessionStorage item and clear it and it doesn't change the spec's behavior. The way it was currently written doesn't really cause the onboarding tour to start from the backlogs module...

modules/team_planner/spec/support/onboarding/onboarding_steps.rb
aaron-contreras: I know it's a lot. I don't really know why this takes so long to load and it's definitely worth addressing but until then I'd prefer to wait 10 more seconds than to have to re-run the whole suite because this spec failed just because it wanted a bit more time. Do you have any ideas why the JS takes so long?

cbliard: Yep, same, it seems pragmatic.

Apart from that, I have no idea why it's taking so long. Even when running locally (so without all the 32 processes running tests like on CI), going to this screen takes 3 to 4 seconds.

aaron-contreras:
No longer valid as it was addressed in https://github.com/opf/openproject/pull/13214/files

PR #13088, opened 2023-07-11
Fix flaky Query Filter spec on Work Packages table

aaron-contreras:
https://community.openproject.org/wp/49072

* Migrates spec to cuprite
* Addresses flakiness cause being that assertions occur while loading is still going on. This being prevented now by `wait_for_reload`

### DIFF
--- a/frontend/src/app/features/work-packages/components/filters/quick-filter-by-text-input/quick-filter-by-text-input.component.ts
+++ b/frontend/src/app/features/work-packages/components/filters/quick-filter-by-text-input/quick-filter-by-text-input.component.ts
@@ -74,9 +74,7 @@ export class WorkPackageFilterByTextInputComponent extends UntilDestroyedMixin {
         }),
       )
       .subscribe((upstreamTerm:string) => {
-        console.log(`upstream ${upstreamTerm} ${(this.searchTerm as any).timestampOfLastValue}`);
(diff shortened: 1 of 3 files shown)


klaustopher (approved):

cbliard:
This won't fix the flakiness. It's more subtle than that. I managed to reproduce it locally by running `s-tui` locally to put my machine under load stress (like the CI machine). I tried to find out what's happening yesterday evening but failed so far. I'll continue this morning. 

What seems to happen (not totally clear to me atm) is that:
- the filter "created at" is selected
- the operator "between" is selected
- the first date field is filled.
  - the change is debounced by 250ms
- the second date field is filled
  - the change on the second field is debounced
  - the first change is triggering a request. **request is sent with only one date** (`filters            = [{"status":{"operator":"o","values":[]}},{"updatedAt":{"operator":"<>d","values":["2023-07-05T22:00:00Z"]}}]`)
- for some reason, the value change of the second field never triggers a second request to update the work package list. Two work packages are displayed: the one from today and the one from 3 days ago. Only the one from 3 days ago should be visible => test fails

The timing is really critical here. I can reproduce it only when stressing my machine, and if I had too many logs, the bug does not occur. There seem to be a short time frame where the second request is not sent or is cancelled. I'll continue investigating.

cbliard:
Yep, I managed to reproduce it while being on this branch.

image

PR #13094, opened 2023-07-11
[#48650] PDF Export: Style picture caption

as-op:
https://community.openproject.org/work_packages/48650

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -156,7 +156,7 @@ gem 'structured_warnings', '~> 0.4.0'
 gem 'airbrake', '~> 13.0.0', require: false
 
 gem 'prawn', '~> 2.2'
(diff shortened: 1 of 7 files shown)


ulferts (approved):
The images now have captions.

What I don't know is whether the left alignment of the caption is intended. This is different from how it is rendered in the UI currently (where it is middle aligned).

as-op:
> The images now have captions.
> 
> What I don't know is whether the left alignment of the caption is intended. This is different from how it is rendered in the UI currently (where it is middle aligned).

Thanks @ulferts, this is indeed not really intended, as it's not specified, but can be configured with the styling yaml. The default alignment is left. I'll take a note and bring it up with the product team.

PR #13109, opened 2023-07-12
[48967] Extract create_user permission out of manage_user

cbliard:
https://community.openproject.org/wp/48967


app/controllers/members_controller.rb
```diff
@@ -203,7 +203,7 @@ def create_members
     overall_result
   end
 
-  def with_new_member_params
+  def each_new_member_param
```
klaustopher, line 206: Renamed because I wanted to make it clearer that this will yield each param and is an iterator

spec/models/capabilities/scopes/default_spec.rb
```diff
@@ -367,6 +369,7 @@
               .map { |_, v| v[:actions].map { |vk, vv| vv.map { |vvv| item.call(vk, vvv, v[:global], v[:module_name]) } } }
               .flatten(2)
               .compact
+              .uniq { |v| v.join(",") }
```
klaustopher, line 372: This is needed, because `show` is allowed by the create and the manage user roles. So they will be added twice. This cleans it up. The `join` is needed because `uniq` uses the `eql?` methods and two arrays are not equal based on their contents

oliverguenther (approved):

PR #13110, opened 2023-07-12
[#48647] PDF Export: Manage long query titles in footer

as-op:
* limit footer lines amount to 3
* if there are more than 3 lines, last one is displayed with an ellipsis
Truncation

https://community.openproject.org/work_packages/48647


ulferts (approved):
👍

PR #13073, opened 2023-07-08
[#48545] Add Global Sidebar Menu

aaron-contreras:
This PR adds the global sidebar menu which allows navigation to and from global index pages. It contains some
drilled-down/descendant sub-menus still within a global context as specified in the Work Package.

See: https://community.openproject.org/work_packages/48545

**NOTE**: Took advantage and cleaned up some specs and moved them over to cuprite for some performance/clarity improvements😄

PR #13117, opened 2023-07-12
[#48604] Add "I am member" filter for projects

aaron-contreras:
**NOTE FOR REVIEWER**: This PR builds on the work done in https://github.com/opf/openproject/pull/13073. Therefore, only the [last commit](bb6a148fb22c051d004aa6e6ecad8a3273682d57) is part of this PR (at least until https://github.com/opf/openproject/pull/13073 is merged into dev)

See Work Package: https://community.openproject.org/work_packages/48604

### DIFF
--- a/app/models/project.rb
+++ b/app/models/project.rb
@@ -157,8 +157,8 @@ class Project < ApplicationRecord
   scope :visible, ->(user = User.current) { where(id: Project.visible_by(user)) }
   scope :newest, -> { order(created_at: :desc) }
   scope :active, -> { where(active: true) }
-  scope :with_member, ->(user = User.current) { where(id: user.memberships.pluck(:project_id)) }
-  scope :without_member, ->(user = User.current) { where.not(id: user.memberships.pluck(:project_id)) }
+  scope :with_member, ->(user = User.current) { where(id: user.memberships.select(:project_id)) }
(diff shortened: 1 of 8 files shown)


aaron-contreras:
First time creating a filter. Looking for comments/suggestions dropped here on both the filter's implementation/spec 🙏

app/models/project.rb
```diff
@@ -157,6 +157,8 @@ class Project < ApplicationRecord
   scope :visible, ->(user = User.current) { where(id: Project.visible_by(user)) }
   scope :newest, -> { order(created_at: :desc) }
   scope :active, -> { where(active: true) }
+  scope :with_member, ->(user = User.current) { where(id: user.memberships.pluck(:project_id)) }
+  scope :without_member, ->(user = User.current) { where.not(id: user.memberships.pluck(:project_id)) }
```
ulferts: Using `pluck` in this context leads to an unnecessary extra request as the `pluck` has to be resolved first with the results of that then being passed into the next query. Using `select` avoids that.

ulferts (changes requested):
Only a tiny thing @aaron-contreras. Merge at will after addressing it.

ulferts (approved):
👍

PR #13122, opened 2023-07-13
[#49095] Move removal of inactive users to be the last step of synchronization.

ba1ash:
https://community.openproject.org/work_packages/49095


akabiru (commented):
Looks good to me from my basic understanding. Q: Should we introduce a spec to ensure the ordering of actions is always honored?

ba1ash:
> Looks good to me from my basic understanding. Q: Should we introduce a spec to ensure the ordering of actions is always honored?

Sure, it makes sense. I will do it.

akabiru (approved):
🏆

PR #13129, opened 2023-07-13
[#49096] Check explicitly if user is a project member before show the link to project folder in the menu.

ba1ash:
https://community.openproject.org/work_packages/49096

- In case an admin is not a member show the link to a storage root folder
- Add a test case
- Enable cuprite for `modules/storages/spec/features/storages_menu_links_spec.rb`

### DIFF
--- a/modules/storages/spec/features/storages_menu_links_spec.rb
+++ b/modules/storages/spec/features/storages_menu_links_spec.rb
@@ -28,7 +28,7 @@
 
 require_relative '../spec_helper'
 
-RSpec.describe 'Project menu', js: true do
+RSpec.describe 'Project menu', js: true, with_cuprite: true do
   let(:storage) { create(:storage, name: "Storage 1") }
   let(:another_storage) { create(:storage, name: "Storage 2") }
   let(:unlinked_storage) { create(:storage, name: "Storage 3") }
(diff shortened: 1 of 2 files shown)


modules/storages/spec/features/storages_menu_links_spec.rb
```diff
@@ -49,14 +49,31 @@
...
+      let(:user) { create(:admin) }
+
+      it 'has a link to enabled storage, but not to hidden for him project' do
+        visit(project_path(id: project.id))
+
+        expect(page).to have_link(storage.name, href: storage.host)
+        project_folder_id = project_storage_with_manual_folder.project_folder_id
+        folder_href = "#{another_storage.host}/index.php/f/#{project_folder_id}?openfile=1"
+        expect(page).not_to have_link(another_storage.name, href: folder_href)
+        expect(page).to have_link(another_storage.name, href: another_storage.host)
```
Kharonus, line 73: wouldn't this check suffice? if the link has href to project folder, this expectation should fail, as it expects an href to only host url.

Kharonus (commented):
A comment you CAN check or leave it out. :D Not really important, LGTM otherwise.

modules/storages/spec/features/storages_menu_links_spec.rb
ba1ash: It should be enough. But I would prefer to be explicit here about the absence of the project folder link.

PR #13136, opened 2023-07-14
[#49140] Change default project folder setting

Kharonus:
- https://community.openproject.org/work_packages/49140
- added switch to controller setting correct default value


modules/storages/app/controllers/storages/admin/projects_storages_controller.rb
```diff
@@ -73,7 +73,9 @@ def new
 
     storage_id = params.dig(:storages_project_storage, :storage_id)
     if storage_id.present?
-      @project_storage.storage = available_storages.find_by(id: storage_id)
+      storage = available_storages.find_by(id: storage_id)
+      @project_storage.storage = storage
+      @project_storage.project_folder_mode = "automatic" if storage.automatically_managed?
     end
```
ba1ash: I assume that this logic should be located inside, because it seems to be responsible for setting new instance attributes. Wdyt?

ba1ash (changes requested):
Let's extend `modules/storages/spec/features/manage_project_storage_spec.rb` to have this behavior tested.

modules/storages/app/controllers/storages/admin/projects_storages_controller.rb
Kharonus: I don't get it. What do you mean by "inside"? I assumed this to be the correct place, as we prepare a project storage object, which is loaded in the view (setting the default value for the folder mode radio buttons).

Kharonus: I just tried a bit to move that code into the `SetAttributesService`, but I failed. Here is the why:

When I create the `@project_storage` from the `SetAttributeService`, I need to pass in the storage, that was selected (with the storage id from the URL), because I need the `automatically_managed?` property from the storage.

But I only get the storage, AFTER I created the `@project_storage`, from the contract.

I cannot find a way to make this code look better then it is now. Do you have any idea, @ba1ash ?

ba1ash: I had a look and added a commit with a possible implementation. @Kharonus have a look, please. I believe it makes sense.

ba1ash (approved):
LGTM

PR #13138, opened 2023-07-14
Prevent rendering of delete button on global calendars page

aaron-contreras:
As per requirements, the global index page for calendars, should not render the delete buttons for each query on the table (in tandem with other global index pages).

See: https://community.openproject.org/work_packages/47850/activity#activity-56

Also, extracted out some expectations into page objects in the calendars#index page. This reads nicer to me and helps keep better track of what is and isn't expected at a high level.


dombesz (approved):
LGMT!👍

PR #13139, opened 2023-07-14
[#49071] Trigger synchronization on oauth token creation and role storage permissions change.

ba1ash:
https://community.openproject.org/work_packages/49071

- `Roles::NotifyMixin` has been removed/reworked. Old events have not been subscribed to(at least I have not found the place).
- New set of events has been introduced inside `lib/open_project/events.rb`. 
- New events has been triggered in the appropriate places.
- Subscriptions to new events happen inside `modules/storages/lib/open_project/storages/engine.rb`
  - In subscriptions we `debounce` Nexcloud group folders synchronization to happen if needed.


modules/storages/lib/open_project/storages/engine.rb
```diff
@@ -32,6 +32,13 @@
 # gets loaded.
 module OpenProject::Storages
   class Engine < ::Rails::Engine
+    PERMISSIONS = %i[
+      read_files
+      write_files
+      create_files
+      delete_files
+      share_files
+    ].freeze
```
ulferts: The same keys exist as part of the [`Storages::GroupFolderPropertiesSyncService::PERMISSIONS_MAP`](https://github.com/opf/openproject/blob/dev/modules/storages/app/services/storages/group_folder_properties_sync_service.rb#L32-L38). Can they be merged or reference one another?

app/controllers/roles_controller.rb
```diff
@@ -81,10 +80,17 @@ def update
...
     @role.destroy
+
     flash[:notice] = I18n.t(:notice_successful_delete)
     redirect_to action: 'index'
-    notify_changed_roles(:removed, @role)
+
+    OpenProject::Notifications.send(
+      OpenProject::Events::ROLE_DESTROYED,
+      permissions:
+    )
```
ulferts: Not mandatory but it would be an improvement if all of this would be handled by a DeleteService.

app/services/roles/notify_mixin.rb
```diff
@@ -1,35 +0,0 @@
...
-# Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
-#
-# See COPYRIGHT and LICENSE files for more details.
-#++
-
-module Roles::NotifyMixin
-  private
-
-  def notify_changed_roles(action, changed_role)
-    OpenProject::Notifications.send(:roles_changed, action:, role: changed_role)
```
ulferts, line 33: I couldn't find any reference to `roles_changed` in the code. It was introduced about 8 years ago stating that external plugins would need it. Given that the newly introduced events have a better resolution, though, I have no problems with the changes in the PR.

modules/storages/lib/open_project/storages/engine.rb
```diff
@@ -62,6 +69,47 @@ class Engine < ::Rails::Engine
...
+              ::Storages::ManageNextcloudIntegrationEventsJob.debounce
+            end
+          end
+          OpenProject::Notifications.subscribe(
+            OpenProject::Events::ROLE_DESTROYED
+          ) do |payload|
+            if payload[:permissions].intersect?(PERMISSIONS)
+              ::Storages::ManageNextcloudIntegrationEventsJob.debounce
+            end
+          end
```
ulferts, line 83: I am surprised about the choice of events in here. My understanding of the intend is to update the permissions on Nextcloud in lockstep with the changes in OpenProject.

If my understanding is correct then ROLE_CREATE does not need to be listed here. It means a role has been created. But the creation of a role does not mean that a permission granted to the user is changed. Only if the newly created role is granted as part of the membership will there be a change in permission. As such, I don't think that listening for this change is necessary.

ulferts (changes requested):
The overall change is fine and in good shape. The old role based notification can be removed in my opinion and the addition of the new one more than replaces it. 

I had a few remarks next within the code but nothing too serious. 

The thing I miss the most though are tests, which I didn't find attached. With the changes I see in this PR I would e.g. write a suite of integration tests that assure that:
* When updating a role with a certain set of permissions, a job is scheduled
* When updating a project, a job is scheduled
* When destroying a role, a job is scheduled
* ...

Not really part of the PR but maybe worthwhile to know, that [there is a PR open to simplify the permission querying](https://github.com/opf/openproject/pull/12963). It does not change the existence of users, projects, members, roles, ... but rather focuses on simplifying the result of that structure by denormalizing it. This result is a database table of the structure:

```
| user_id | project_id | permission   |
|    5    |      23    | read_files   | 
|    8    |      34    | create_files |
```

If I manage to finish it, then it might be simpler to be notified about changes in the resulting permissions rather than listening in on the changes to the models. But that will likely not happen for 13.0.

app/models/role.rb
```diff
@@ -39,7 +39,13 @@ class Role < ApplicationRecord
     where("#{compare} builtin = #{NON_BUILTIN}")
   }
 
-  before_destroy :check_deletable
+  before_destroy(prepend: true) do
+    unless deletable?
+      errors.add(:base, "can't be destroyed")
+      throw :abort
```
mereghost: Are we using a `catch/throw` somewhere? Would raising an exception like `RecordNotDestroyed` be better?

ba1ash: I am not aware of the other places. My intention was just to stop the destroy method with something that is caught by ActiveRecord. So, we do not need to care about catching.
Sure, I will use the exception instead. The behaviour should be the same.

modules/storages/lib/open_project/storages/engine.rb
ba1ash: Done.

app/controllers/roles_controller.rb
ba1ash: Done.

modules/storages/lib/open_project/storages/engine.rb
ba1ash: Yes, you are absolutely right!
I unsubscribed from PROJECT_CREATED as well for similar reasons.

ba1ash:
@ulferts have an extra look, please.
I tested events issuing and the handling of issued events separately.

modules/storages/lib/open_project/storages/engine.rb
ulferts: Is is true that non members (i.e. a user in a public project) cannot have permissions in Nextcloud (I didn't check)? Only if that is true should the `PROJECT_CREATED` be removed as otherwise the project created might be public which would then grant permissions.

app/contracts/groups/delete_contract.rb
```diff
@@ -28,6 +28,6 @@
 
 module Groups
   class DeleteContract < ::DeleteContract
-    delete_permission -> { user.active? && user.admin? }
+    delete_permission(:admin)
```
ulferts, line 31: What is the reason for this change? I can't say that it is wrong but it seems to deliberate to be checking for whether the user is active that I would surmise that part being the result of a bug fix.

ulferts (approved):
Hi @ba1ash, thanks for adding the tests. I only have questions, but I expect the change to be mergeable.

app/contracts/groups/delete_contract.rb
ba1ash: My understanding is that there is no change in behavior.
If we check `DeleteContract` class. Then using `delete_permission(:admin)` here will force `authorized?` to fall into [this line](https://github.com/opf/openproject/blob/dev/app/contracts/delete_contract.rb#L59) which has the same logic.

PR #13141, opened 2023-07-16
feat[OP#48569] Validate application password against Nextcloud

akabiru:
https://community.openproject.org/work_packages/48569

Screenshot 2023-07-16 at 10 15 32 AM
Screenshot 2023-07-16 at 10 15 48 AM


modules/storages/app/validator/nextcloud_application_credentials_validator.rb
```diff
@@ -0,0 +1,71 @@
...
+    return unless contract.model.password_changed?
+
+    request = build_http_head_request
+    response = make_http_head_request_for(request)
+
+    case response
+    when Net::HTTPSuccess
+      true
+    when Net::HTTPUnauthorized
+      add_contract_error_for_password_with(:invalid_password)
```
ba1ash: Are you sure we need this method? For me it seems like a bit superfluous, because if I place implementation itself here it is descriptive enough and then it is shorter and no need for extra method. Wdyt?
```suggestion
      contract.errors.add(:password, :invalid_password)
```

modules/storages/app/validator/nextcloud_application_credentials_validator.rb
```diff
@@ -0,0 +1,71 @@
...
+    else
+      add_contract_error_for_password_with(:unknown_error)
```
ba1ash:
```suggestion
      contract.errors.add(:password, :unknown_error)
```

modules/storages/app/validator/nextcloud_application_credentials_validator.rb
```diff
@@ -0,0 +1,71 @@
...
+    @uri = URI(contract.host).normalize
+  end
+
+  def call
+    return unless contract.model.password_changed?
+
+    request = build_http_head_request
+    response = make_http_head_request_for(request)
+
+    case response
```
ba1ash: I would suggest to write it like this. Method's names explains returned values quite well.
```suggestion
    case make_http_head_request_for(build_http_head_request)
```

modules/storages/app/validator/nextcloud_application_credentials_validator.rb
```diff
@@ -0,0 +1,71 @@
...
+
+  def make_http_head_request_for(request)
+    Storages::Peripherals::StorageInteraction::Nextcloud::Util
+      .http(uri)
+      .request(request)
+  end
+
+  def add_contract_error_for_password_with(error_kind)
+    contract.errors.add(:password, error_kind)
+  end
```
ba1ash:
```suggestion
```

ba1ash (approved):
Well done! I tried it out and it looks good to me. I have a couple of code suggestions. Just ignore them if they don't make sense.

modules/storages/app/validator/nextcloud_application_credentials_validator.rb
akabiru: I agree it doesn't add too much value and is descriptive enough on it's own 👍🏾

modules/storages/app/validator/nextcloud_application_credentials_validator.rb
```diff
@@ -0,0 +1,64 @@
...
+# Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+#
+# See COPYRIGHT and LICENSE files for more details.
+#++
+class NextcloudApplicationCredentialsValidator
+  attr_reader :contract, :uri
+
+  def initialize(contract)
+    @contract = contract
+    @uri = URI(contract.host).normalize
```
Kharonus, line 33: we probably didn't test it yet with a storage with a sub path host (like `https://nextcloud.local/html`).

Should I run a test for that? Or you wanna setup your test nextcloud to run under a sub path? :D

Kharonus: Actually: this should be able to be tested by your unit tests. just using a host like `https://example.com/html` und check, if the webmock stub is correctly reached.

Kharonus (commented):
A small request regarding sub pathes for storages.

modules/storages/app/validator/nextcloud_application_credentials_validator.rb
akabiru: [URI.normalize](https://apidock.com/ruby/v2_5_5/URI/Generic/normalize) should handle subpaths similarly- we're using this format in other places as well. I'll add a unit test for posterity 👍🏾 

```ruby
[1] pry(main)> URI('https://nextcloud.local/html').normalize
=> #<URI::HTTPS https://nextcloud.local/html>
```

Done https://github.com/opf/openproject/pull/13141/commits/253446672c46c3b935a88d28c1ce6557cfb812c2

Kharonus: nice, thanks!

PR #13147, opened 2023-07-17
[#48464] Remove export options from backlog module

as-op:
* backlogs module: remove export option from UI 
* backlogs module: remove export function from controller
* backlogs module: remove export_card_configuration from admin UI
* openproject-pdf_export module: remove since it is now unused
* remove unused table "export_card_configurations" with migration
* remove relevant specs

https://community.openproject.org/work_packages/48464


db/migrate/20221026132134_bigint_primary_and_foreign_keys.rb
```diff
@@ -68,7 +68,6 @@ class BigintPrimaryAndForeignKeys < ActiveRecord::Migration[7.0]
     EnabledModule => %i[id project_id],
     EnterpriseToken => [:id],
     Enumeration => %i[id project_id parent_id color_id],
-    ExportCardConfiguration => [:id],
```
ulferts, line 71: please turn this into the table name `:export_card_configurations`. Because back then, the table still exists and should be adapted.

ulferts (changes requested):
Just one tiny thing that I would like to have changed so please merge yourself once that is addressed. 

Kudos for the spec_helper change.

db/migrate/20221026132134_bigint_primary_and_foreign_keys.rb
as-op: Thank you, @ulferts I tried that in a [previous commit](https://github.com/opf/openproject/pull/13147/commits/159ab50c53e9cd8b60802c3a133d23e96ed54513) but the result was 
https://github.com/opf/openproject/actions/runs/5574388967/jobs/10182939175#step:5:70
```
StandardError: An error has occurred, this and all later migrations canceled:

PG::UndefinedTable: ERROR:  relation "export_card_configurations" does not exist
LINE 9:  WHERE a.attrelid = '"export_card_configurations"'::regclass
                            ^
/app/db/migrate/migration_utils/column.rb:38:in `initialize'
/app/db/migrate/20221026132134_bigint_primary_and_foreign_keys.rb:215:in `new'
/app/db/migrate/20221026132134_bigint_primary_and_foreign_keys.rb:215:in `change_column_type'
/app/db/migrate/20221026132134_bigint_primary_and_foreign_keys.rb:204:in `block in change_columns_type'
/app/db/migrate/20221026132134_bigint_primary_and_foreign_keys.rb:222:in `block in for_each_klass_and_columns'
/app/db/migrate/20221026132134_bigint_primary_and_foreign_keys.rb:221:in `each'
/app/db/migrate/20221026132134_bigint_primary_and_foreign_keys.rb:221:in `for_each_klass_and_columns'
/app/db/migrate/20221026132134_bigint_primary_and_foreign_keys.rb:203:in `change_columns_type'
/app/db/migrate/20221026132134_bigint_primary_and_foreign_keys.rb:182:in `block in up'
/app/db/migrate/20221026132134_bigint_primary_and_foreign_keys.rb:1[81](https://github.com/opf/openproject/actions/runs/5574388967/jobs/10182939175#step:5:82):in `each'
/app/db/migrate/20221026132134_bigint_primary_and_foreign_keys.rb:181:in `up'
```

My assumption then was that since I removed the migrations that added the table, it does no longer exists in a fresh instance.

ulferts: True, that is because the migration that creates the table is removed in this PR. But for an instance that has had run the creating migration, the table would still be there. 

We could now do what you did in the first place and remove the reference as the change to the migration here will only exist in combination with the change removing the table. I guess it is the pragmatic thing to do. 

Admittedly I don't really like it since one then cannot traverse the migrations but that is more of a "That is the way it should be"-argument rather then one that actually matters. 

Going for pragmatic here so no change needed.

ulferts (approved):
👍

PR #13157, opened 2023-07-18
Journal the addition and removal of File Links on Work Packages

mereghost:
### Activities ☑️ 

With the context from the issue, first question is: what constitutes an **Activity**? Any changes of the state to the object of interest. A file added, an assignee added, the creation of the object itself from the aether.

#### How does it work? ⚙️ 

Internally, every activity is an entry in the `Journal` model. Every new object there, has a version attached to it and we compare different versions against each other to create the activities. Each of these changes are a `detail` in `Journal` parlance.

The actual mechanism of the comparison is kinda cool, but I'll leave you to explore that.

So, for `WorkPackage`s (or for any `journalized` model) every time changes happen and are saved to the database, a new `Journal` entry is created that points to the specialized `Journal` model for that object. The specialized `Journal` (`Journal::WorkPackageJournal` in this case) it snapshots the model at the time of saving.

When a `journalized` record is removed, so are all its journal entries.

### Attachments 📋 

Attachments aren't properly an attribute of `WorkPackage` but an associated model. Also, if when you destroy an object, the journal entries go with it, so how does it show the added and removed files on screen (or via the API)?

For those there's a couple of tricks. The `Journals::CreateService` instead of relying on the `Journal::AttachmentJournal` it creates a special type of journal called `Journal::AttachableJournal` that points to an `Attachment` and version of the `Journal` that itself points to the `Journal::WorkPackageJournal`.

When you remove the `Attachment` another `WorkPackageJournal`is added without the `AttachableJournal`, thus making the diff between the records show that the previous file was missing.

All this behaviour is dealt with within the `Journal::CreateService` massive SQL.

## The Solution 🧪 

I've tried to mimic the behaviour already existing by introducing the `Journal::StorableJournal` (yeah, I'm terrible at naming) that will behave exactly as the `Journal::AttachableJournal`. This meant editing the SQL inside the `Journal::CreateService` and adding the necessary queries.

Besides this, I had to add calls to the `Journal::CreateService` on both of our `...FileLink::CreateService` and `...FileLink::DeleteService`. The code is identical, but it felt like overkill to just create another abstraction for it (cue in Sandy Metz "the wrong abstraction is worse than repetition").

I also needed to add a formatter for our `file_links` on the `Journal`. This will allow us to change the rendered templates easily if needed. Besides this there was only a couple of touch-ups on other files to eager-load the new records.

### DIFF
--- a/db/migrate/20230713144232_create_storages_file_link_journals.rb
+++ b/db/migrate/20230713144232_create_storages_file_link_journals.rb
@@ -1,0 +1,42 @@
+# frozen_string_literal: true
+
+#-- copyright
+# OpenProject is an open source project management software.
+# Copyright (C) 2012-2023 the OpenProject GmbH
+#
+# This program is free software; you can redistribute it and/or
+# modify it under the terms of the GNU General Public License version 3.
+#
+# OpenProject is a fork of ChiliProject, which is a fork of Redmine. The copyright follows:
+# Copyright (C) 2006-2013 Jean-Philippe Lang
+# Copyright (C) 2010-2013 the ChiliProject Team
+#
+# This program is free software; you can redistribute it and/or
+# modify it under the terms of the GNU General Public License
+# as published by the Free Software Foundation; either version 2
+# of the License, or (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program; if not, write to the Free Software
+# Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+#
+# See COPYRIGHT and LICENSE files for more details.
+#++
+
+class CreateStoragesFileLinkJournals < ActiveRecord::Migration[7.0]
+  def change
+    # rubocop:disable Rails/CreateTableWithTimestamps
+    create_table :storages_file_links_journals do |t|
+      t.belongs_to :journal
+      t.belongs_to :file_link
+
+      t.string :link_name, default: ''
+    end
+    # rubocop:enable Rails/CreateTableWithTimestamps
+  end
+end
--- a/db/migrate/20230713144232_create_storages_file_link_journals.rb
+++ b/db/migrate/20230713144232_create_storages_file_link_journals.rb
@@ -32,10 +32,10 @@ class CreateStoragesFileLinkJournals < ActiveRecord::Migration[7.0]
   def change
     # rubocop:disable Rails/CreateTableWithTimestamps
     create_table :storages_file_links_journals do |t|
-      t.belongs_to :journal
-      t.belongs_to :file_link
+      t.belongs_to :journal, null: false, foreign_key: true
+      t.belongs_to :file_link, null: false
 
-      t.string :link_name, default: ''
+      t.string :link_name, null: false
     end
     # rubocop:enable Rails/CreateTableWithTimestamps
   end
--- a/lib/open_project/journal/attachment_helper.rb
+++ b/lib/open_project/journal/attachment_helper.rb
@@ -35,6 +35,13 @@ module OpenProject
           save
         end
       end
+
+      def file_links_changed(_obj)
+        unless new_record?
+          add_journal
+          save
+        end
+      end
     end
   end
 end
--- a/lib/open_project/journal/attachment_helper.rb
+++ b/lib/open_project/journal/attachment_helper.rb
@@ -35,13 +35,6 @@ module OpenProject
           save
         end
       end
-
-      def file_links_changed(_obj)
-        unless new_record?
-          add_journal
-          save
-        end
-      end
     end
   end
 end
--- a/lib/open_project/journal_formatter/file_link.rb
+++ b/lib/open_project/journal_formatter/file_link.rb
@@ -1,0 +1,58 @@
+# frozen_string_literal: true
+
+#-- copyright
+# OpenProject is an open source project management software.
+# Copyright (C) 2012-2023 the OpenProject GmbH
+#
+# This program is free software; you can redistribute it and/or
+# modify it under the terms of the GNU General Public License version 3.
+#
+# OpenProject is a fork of ChiliProject, which is a fork of Redmine. The copyright follows:
+# Copyright (C) 2006-2013 Jean-Philippe Lang
+# Copyright (C) 2010-2013 the ChiliProject Team
+#
+# This program is free software; you can redistribute it and/or
+# modify it under the terms of the GNU General Public License
+# as published by the Free Software Foundation; either version 2
+# of the License, or (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program; if not, write to the Free Software
+# Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+#
+# See COPYRIGHT and LICENSE files for more details.
+#++
+
+class OpenProject::JournalFormatter::FileLink < JournalFormatter::Base
+  include OpenProject::ObjectLinking
+
+  def render(key, values, options = { html: true })
+    id = key.to_s.sub('file_links_', '')
+    label, old_value, value = format_details(id, values)
+
+    if options[:html]
+      label, old_value, value = *format_html_details(label, old_value, value)
+      value = format_html_file_link_detail(id, value)
+    end
+
+    render_binary_detail_text(label, value, old_value)
+  end
+
+  private
+
+  # Based this off the Attachment formatter. Not sure if it is the best approach
+  def label(_key) = Storages::FileLink.model_name.human
+
+  def format_html_file_link_detail(key, value)
+    if value.present? && a = ::Storages::FileLink.find_by(id: key.to_i)
+      link_to_file_link(a, only_path: false)
+    elsif value.present?
+      value
+    end
+  end
+end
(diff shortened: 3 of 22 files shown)


db/migrate/20230713144232_create_storages_file_link_journals.rb
```diff
@@ -0,0 +1,42 @@
...
+# Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+#
+# See COPYRIGHT and LICENSE files for more details.
+#++
+
+class CreateStoragesFileLinkJournals < ActiveRecord::Migration[7.0]
+  def change
+    # rubocop:disable Rails/CreateTableWithTimestamps
+    create_table :storages_file_links_journals do |t|
+      t.belongs_to :journal
```
ulferts: Had to look it up myself since I keep on forgetting it but [belongs_to does not add a NOT NULL](https://edgeapi.rubyonrails.org/classes/ActiveRecord/ConnectionAdapters/SchemaStatements.html#method-i-add_reference) constraint by default which I think would be appropriate to apply here.
Same is true for a foreign key constraint.

db/migrate/20230713144232_create_storages_file_link_journals.rb
```diff
@@ -0,0 +1,42 @@
...
+      t.belongs_to :file_link
+
+      t.string :link_name, default: ''
```
ulferts: I'd remove the default and actually turn this into a non nullable column. I don't see why this should ever be empty as the purpose is to fall back to this value in case the actual file_link is removed.

lib/open_project/journal/attachment_helper.rb
```diff
@@ -35,6 +35,13 @@ def attachments_changed(_obj)
           save
         end
       end
+
+      def file_links_changed(_obj)
+        unless new_record?
+          add_journal
+          save
+        end
+      end
```
ulferts: This part is probably not needed. It is never referenced anywhere AFAIK

lib/open_project/object_linking.rb
```diff
@@ -147,5 +155,15 @@ def url_to_attachment(attachment, only_path: true)
         v3_paths.url_for(:attachment_content, attachment.id)
       end
     end
+
+    def url_to_file_link(file_link, only_path: true)
+      v3_paths = API::V3::Utilities::PathHelper::ApiV3Path
```
ulferts: if the comment above about the module breaking when including the module is still true, extracting the v3_paths into a method that is then called would be the next best thing I think.

lib_static/plugins/acts_as_journalized/lib/journal_changes.rb
```diff
@@ -41,6 +41,7 @@ def get_changes
 
     @changes.merge!(get_association_changes(predecessor, 'attachable', 'attachments', :attachment_id, :filename))
     @changes.merge!(get_association_changes(predecessor, 'customizable', 'custom_fields', :custom_field_id, :value))
+    @changes.merge!(get_association_changes(predecessor, 'storable', 'file_links', :file_link_id, :link_name))
```
ulferts: This will now be called for all journalized resources which includes work_packages but is not limited to it (e.g. Meeting and WikiPage). The same is of course potentially true for the calls that existed before fetching the changes  to attachments and custom values. E.g. WikiPage does not have a custom values. 

With more associations being checked now, I think it makes sense to introduce simple safeguards to each call to check if the change needs to be collected for this journable or if that would be futile anyway. 

In it's simplest form it could be checking for an included module or method to be there ala `self.class.included_modules.include?(Redmine::Acts::Customizable::InstanceMethods)` but it could also be build into the modules themselves so that we could check for something like `self.class.customized?` (analogous to how it is possible for `journaled?`

db/migrate/20230713144232_create_storages_file_link_journals.rb
```diff
@@ -0,0 +1,42 @@
...
+# You should have received a copy of the GNU General Public License
+# along with this program; if not, write to the Free Software
+# Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+#
+# See COPYRIGHT and LICENSE files for more details.
+#++
+
+class CreateStoragesFileLinkJournals < ActiveRecord::Migration[7.0]
+  def change
+    # rubocop:disable Rails/CreateTableWithTimestamps
```
akabiru, line 33: Q: Curious why timestamps are not needed? I would presume they'd provide useful timing metadata? E.g. If we were debugging for instance. _*disclaimer: the journals implementation is new to me_

lib/open_project/journal_formatter/file_link.rb
```diff
@@ -0,0 +1,58 @@
...
+      value = format_html_file_link_detail(id, value)
+    end
+
+    render_binary_detail_text(label, value, old_value)
+  end
+
+  private
+
+  # Based this off the Attachment formatter. Not sure if it is the best approach
+  def label(_key) = Storages::FileLink.model_name.human
```
akabiru, line 49: huh, didn't know this was valid syntax. 🎓 😄

lib/open_project/journal_formatter/file_link.rb
```diff
@@ -0,0 +1,58 @@
...
+
+  def format_html_file_link_detail(key, value)
+    if value.present? && a = ::Storages::FileLink.find_by(id: key.to_i)
+      link_to_file_link(a, only_path: false)
```
akabiru: Minor naming suggestion

```suggestion
    if value.present? && (file_link = ::Storages::FileLink.find_by(id: key.to_i))
      link_to_file_link(file_link, only_path: false)
```

modules/storages/app/services/storages/file_links/create_service.rb
```diff
@@ -39,6 +39,19 @@ def persist(service_result)
...
 
+  def after_perform(service_result)
+    # This only gets called if service_result is successful
+    container = service_result.result.container
+
+    # If the container isn't journaled, no need to proceed
+    return service_result unless container&.class&.journaled?
+
+    # If journal creation fails, we don't care for now
+    Journals::CreateService.new(container, service_result.result.creator).call
```
ulferts: Please use the `save_journals` method on the container. That will ensure the [`JOURNAL_CREATED` event being sent out](https://github.com/opf/openproject/blob/dev/lib_static/plugins/acts_as_journalized/lib/acts/journalized/save_hooks.rb#L67-L69) which e.g. the sending of notifications (leading to in app notifications and mails) relies on.

Admittedly, it would also make sense to move that notification sending into the `Journals::CreateService` since it should be service's responsibility.

Regardless, it is also shorter to just call `container.save_journal` :)

A nice side effect of this change should be, that the container will receive a touch which will make for a consistent experience for the users, seeing the upload in the journals and seeing that the work package was updated at the same time.

Since I haven't tested the code out myself, please also check whether the sending of notifications and mails works as expected. I would expect this to be the case as the changes in here follow the established patterns but double checking is advisable.

akabiru (commented):
Top work! 🏆  The detailed [PR description](https://github.com/opf/openproject/pull/13157) is a really nice touch 👍🏾 

_*I don't have enough context here so will leave the end approval to Jens :)_

ulferts (changes requested):
The change is good and follows along the established patterns. Nothing in here surprised me in a bad way so the change itself is on a good track. I added some remarks but it should be easy to address. 

What I have been wondering for some time is how to turn the `Journals::CreateService` to become more flexible so that it is easier to journalize associations and to only do the journalization of e.g. custom_values where it is necessary. But this part does not need to be addressed within this PR. 

I haven't tested the code myself as don't have the env set up for it. I'll see how much of a hassle it turns out to be. If I don't manage to do it, somebody else would have to cover that part.

lib/open_project/journal/attachment_helper.rb
mereghost: Yup, it was a leftover from my previous quackery.

db/migrate/20230713144232_create_storages_file_link_journals.rb
mereghost: My current understanding is that these journals are generated the same time as the main `Journal` entry. Timestamps are left off in this case as they don't really matter.

lib/open_project/journal_formatter/file_link.rb
mereghost: This is new to 3.0, it is called `endless method` which makes me think of a very different thing. 🤣

db/migrate/20230713144232_create_storages_file_link_journals.rb
mereghost: Oh, yeah. IIRC it is because of the `dependent: :nullify` option and the FK part is because DHH hates FKs for some reason. 🙄 I'll amend this to add both.

modules/storages/app/services/storages/file_links/create_service.rb
mereghost: After the change notifications are working as expected.

lib/open_project/object_linking.rb
mereghost: Including the `PathHelper` seemed safe,but we would still need to prefix the calls with `ApiV3Path` which is shorter but I opted for the method as, if the breaking is subtle, it would be hard to me identifying it.

lib_static/plugins/acts_as_journalized/lib/journal_changes.rb
mereghost: Agree that it isn't great from a performance perspective, even if the query returns nothing as it is still a FK lookup (which is fast but not so fast as not running).

That makes me think of a couple of changes:

1. `Journal::CreateService` could probably tailor its SQL based on `journable` capabilities (i.e. not run the extra sql for attachables etc).
2. `JournalChanges`, through journal, could ask the `journable` and only run what's necessary.
3. These changes feel big. 

Would you like this addressed in this or in a following `code_maintenance` PR?

lib/open_project/journal_formatter/file_link.rb
aaron-contreras: 👨‍🎓

lib_static/plugins/acts_as_journalized/lib/journal_changes.rb
ulferts: I see that you now started on 2. to improve the code in `JournalChanges` and that is by far enough for this PR. Doing 1. makes a ton of sense since we should be conscious of performance with this service but it needn't be in this PR. For motivation, though: When the dates of a  work package being part of an elaborate ancestry/follows relationship is updated, the other work packages being part of the relationship might also have to be updated. I have seen cases where because of a date change to a work package 50+ other work packages were updated. In such a case, every ms of improvement counts.

ulferts (approved):
Amends look good. So from my side, this is good to merge at will.

PR #13158, opened 2023-07-18
Implementation/47064 support custom field  in the attributesbytimestamp

dombesz:
https://community.openproject.org/wp/47064


lib_static/plugins/acts_as_journalized/lib/acts/journalized/journable_differ.rb
```diff
@@ -30,25 +30,29 @@ module Acts::Journalized
...
         data.each_with_object({}) do |e, h|
-          h[e[0]] = (e[1].is_a?(String) ? e[1].gsub(/\r\n/, "\n") : e[1])
+          h[e[0]] = (e[1].is_a?(String) ? e[1].gsub("\r\n", "\n") : e[1])
         end
       end
 
       def no_nil_to_empty_strings?(normalized_old_data, attribute, new_value)
         old_value = normalized_old_data[attribute]
-        new_value != old_value && (new_value.present? || old_value.present?)
+        new_value != old_value
```
ulferts: I thinks this method no longer serves it's purpose of checking whether value might have changed from e.g. `nil` to `''` as the method name suggests. 

```
new_value = nil
old_value = ''

new_value != old_value && (new_value.present? || old_value.present?)
=> false

new_value != old_value
=> true
```

To have the implementation be more explicit, one could change it into
```
 ([new_value, old_value] - ['', nil]).any?
```

*Update note* In an earlier version of this `old_value = nil` was specified which was incorrect.

ulferts: It is unsettling that no spec pics up on this. So I don't know how relevant my complaint is.

app/models/journable/with_historic_attributes/loader.rb
```diff
@@ -86,6 +116,28 @@ def journalized_class
       journables.first.class
     end
 
+    def load_journals_by_id(journal_ids)
+      Journal
+        .where(id: journal_ids)
+        .includes({ customizable_journals: :custom_field })
+        .index_by(&:id)
```
ulferts: Tiny thing: The `Journal` needn't be involved in the loading. The same result can be achieved by

```
Journal::CustomizableJournal
  .where(journal_id: journal_ids)
  .includes(:custom_field)
  .index_by(&:journal_id)
```

ulferts (changes requested):
Hey Attila, I like the change. The number of places touched is really kept in check nicely.

I have two tiny remarks in the code and one issue that can also be addressed in a subsequent PR. That is actually what I would suggest, to fix the two small issues and have the bigger issue addressed separately.

The bigger issue is an N+1 query when the diff for the representer is created:

image

We also have https://community.openproject.org/wp/48572 where we might place the fix in. It is not strictly the same but related.

lib_static/plugins/acts_as_journalized/lib/acts/journalized/journable_differ.rb
dombesz: I made the changes here 953bdf0f2ddf3e6d9dcb06b40ebc919fbd3cebb1. The expectations now fail in case any of the `nil -> ''` attribute transitions are included.
Some minor adjustment was required on the conditions, because `[false].any?` returns `false` and it breaks the expectation when a transition happens from `nil -> false`. I used `([new_value, old_value] - ['', nil]).present?` instead.

app/models/journable/with_historic_attributes/loader.rb
dombesz: Done 57c3bf093da05728519db5c0e28f267524258281

dombesz:
Thanks for the thorough review @ulferts. I addressed your remarks except the N+1 issue, that will be addressed separately. From my side this PR is good to go, once the CI is green.

ulferts (approved):
:+1: merge at will

PR #13160, opened 2023-07-18
[#45011] No info about changes in activity when use checkbox in custom fields

dombesz:
https://community.openproject.org/work_packages/45011


HDinger (approved):
👍 The failing spec seems to be unrelated

PR #13036, opened 2023-07-05
Introduce queries for meetings

klaustopher:
- [x] Add Query for Meetings
- [x] Add Filters for query
  - [x] Filter for `future` and `past` meetings
  - [ ] potentially: Filter for specific time range? (see discussion in ticket)
  - [x] Filter for invited users (including me)
  - [x] Filter for participating users (including me)
  - [x] Filter for author (including me)
- [ ] ~Integrate UI into the frontend~
- [x] Test it

---

https://community.openproject.org/work_packages/48547/activity

PR #13165, opened 2023-07-19
[#48546] Add Submenus for Meetings

aaron-contreras:
Adds a submenu for both the:

- Project Meetings Index Page
- Global Meetings Index Page

with the set of queries built in https://github.com/opf/openproject/pull/13036 that allows users to filter out the subset of meetings they're interested in.

See: https://community.openproject.org/work_packages/48546


ulferts (changes requested):
The code changes are all fine and the functionality is as well except for an omission. The specification says 

> The default sorting order of the meetings in the views will always be using the current day as reference so in the top of table there will be the meetings of today and then going down to the future or to the past (upcoming or past).

This is not yet implemented.

The specification doesn't state the desired sort order of the 'Involvement' filters so that might be something where Product should be asked. We could of course just do it by time desc but I believe that it might be different between the 'Invited'  and the 'Attendee' for example.

klaustopher:
@ulferts Should this be done in the filter? That `past` means `order by start_time DESC` and `future` means `order by start_time ASC`. Might make sense to add the defaults in there, instead of in the links?

PR #13182, opened 2023-07-19
Fix a bug that did not allow removing permissions from a global role

klaustopher:


spec/features/global_roles/global_role_update_spec.rb
```diff
@@ -0,0 +1,52 @@
...
+# You should have received a copy of the GNU General Public License
+# along with this program; if not, write to the Free Software
+# Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+#
+# See COPYRIGHT and LICENSE files for more details.
+#++
+
+require 'spec_helper'
+
+RSpec.describe 'Role updating', js: true, with_cuprite: true do
```
aaron-contreras, line 31: ❤️

aaron-contreras (approved):
Took me a while to get that it was the `memberPermissions` being rendered as well and thus causing an override. Nice catch!

PR #13189, opened 2023-07-20
Implementation/48587 copy project folder on project copy

apfohl:
https://community.openproject.org/projects/openproject/work_packages/48587


modules/storages/app/services/projects/copy/storage_project_folders_dependent_service.rb
```diff
@@ -0,0 +1,96 @@
...
+            project_storage.update!(project_folder_id: file_id)
+          end,
+          on_failure: ->(r) { add_error!(destination_folder_name, r.to_active_model_errors) }
+        )
+    end
+
+    def project_folder_path(project_storage)
+      project = project_storage.project
+      "#{project_storage.storage.group_folder}/#{project.name.gsub('/', '|')} (#{project.id})/"
+    end
```
ba1ash: I think we should put this logic inside `Storages::ProjectStorage` class, because we have the same path constructor in `Storages::GroupFolderPropertiesSyncService` already. So, the following implementation should fit good here and there.
```ruby
Storages::ProjectStorage
  # ...other code...
  def project_folder_path
    "#{storage.group_folder}/#{project.name.gsub('/', '|')} (#{project.id})/"
  end
end  
```
Wdyt?

ba1ash (commented):
I tested it locally. It works fine for me. I have only one suggestion. Let me know what you think about it.

ba1ash (approved):

PR #12953, opened 2023-06-26
Meetings with agenda tops prototype

jjabari-op:
Evaluating Primer and Hotwire/ViewComponents in combination while prototyping a new feature (agenda items for meetings)

-> Navigate to project meetings and create a new meeting. You will find the option to create agenda_items there.

Findings towards Primer so far:

- Our list styles break the border_box component (which renders rows as `li` elements) -> our margin-left conflicts with the Primer style
- Our form styles break the Primer form styles -> Primer inputs should have a border-radius and different label margins (new or edit state of agenda_items show the broken Primer forms)
- I've created a primer adjustments css which would fix these issues but then disabled it in order not to hide these effects (frontend/src/global_styles/openproject/_primer-adjustments.sass)

Finding towards Hotwire&ViewComponents:

- In order to avoid some "low-level" implementations when defining turbo-streams, I've come up with a prototype approach on how to use base ViewComponent classes (modules/meeting/app/components/base/turbo_component.rb) to create an abstraction layer on top of turbo-streams. You see the usage in the meeting_agenda_item_controller (modules/meeting/app/controllers/meeting_agenda_items_controller.rb) In comparison to that, you can review the implementation without ViewComponents just with some helper methods here: https://github.com/opf/openproject/blob/0b10f60ea0a323f8c19f2f4af0f3cac27b6c1151/modules/meeting/app/controllers/meeting_agenda_items_controller.rb

The long term goal of this spike is to develop a concept for this EPIC: https://community.openproject.org/projects/openproject/work_packages/37297/activity

PR #13194, opened 2023-07-20
Meeting with agenda tops v1 stable

jjabari-op:
Old state of #12953 in order to get it deployed via pullpreview


ulferts:
@jonasjabari I believe this PR can be closed as #3729 exists. Is that correct?

PR #13204, opened 2023-07-21
[chore] Remove project folder mode default

Kharonus:
Removed default as aftermath to discussions in #13136. Kudos to @ba1ash .


ba1ash (approved):
:+1:

PR #13214, opened 2023-07-24
fix flaky team planner onboarding spec - hacky way

ulferts:
Uses a hacky workaround to the problem of the team planner (among other `PartitionedQuerySpacePageComponent` incarnations) not loading without an external event (e.g. the mouse moving or the notification bell update).

https://community.openproject.org/wp/49073


ulferts:
I restarted the specs on the CI 3 times now and every run passed.

PR #13216, opened 2023-07-24
[#49096] Hide storage menu link if user is not a project member.

ba1ash:
https://community.openproject.org/work_packages/49096


Kharonus (approved):
LGTM

PR #13222, opened 2023-07-24
[OP#48413] Danger zone when deactivating a storage on a project

akabiru:
https://community.openproject.org/work_packages/48413

Screenshot 2023-07-26 at 1 47 01 PM

### DIFF
--- a/modules/storages/app/controllers/storages/admin/projects_storages_controller.rb
+++ b/modules/storages/app/controllers/storages/admin/projects_storages_controller.rb
@@ -39,7 +39,7 @@ class Storages::Admin::ProjectsStoragesController < Projects::SettingsController
   # This defines @object as the model instance.
   model_object Storages::ProjectStorage
 
-  before_action :find_model_object, only: %i[edit update destroy] # Fill @object with ProjectStorage
+  before_action :find_model_object, only: %i[edit update destroy destroy_info] # Fill @object with ProjectStorage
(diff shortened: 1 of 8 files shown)


modules/storages/app/controllers/storages/admin/projects_storages_controller.rb
```diff
@@ -146,6 +146,13 @@ def destroy
     redirect_to project_settings_projects_storages_path
   end
 
+  def destroy_info
+    @project_storage_to_destroy = @object
+
+    hide_project_in_layout
```
ba1ash: Are we sure to hide project's layout? Maybe, It makes sense not to, because we remove storage in the context of a project. Wdyt?

ba1ash (approved):
It works fine for me: Well done :+1: 
I have only one question which is not critical.

modules/storages/app/controllers/storages/admin/projects_storages_controller.rb
akabiru: Good point, I mostly followed the project delete example which actually shouldn't apply completely here. Great catch! I'll update 👍🏾

akabiru: Done in https://github.com/opf/openproject/pull/13222/commits/2fc81fd9f9b97b227263151689d06740de130beb

Screenshot 2023-07-26 at 1 47 01 PM

PR #13223, opened 2023-07-24
Chore/file link sync service refactoring

apfohl:


modules/storages/app/common/storages/peripherals/storage_file_info_converter.rb
```diff
@@ -0,0 +1,46 @@
...
+# along with this program; if not, write to the Free Software
+# Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+#
+# See COPYRIGHT and LICENSE files for more details.
+#++
+
+module Storages::Peripherals
+  module StorageFileInfoConverter
+    def to_storage_file(storage_file_info)
+      Storages::StorageFile.new(
```
mereghost, line 32: 🟢 My first time seeing this, but WDYT about making the `Storages::StorageFile` into a [`Data`](https://docs.ruby-lang.org/en/3.2/Data.html) or a [`Struct`](https://docs.ruby-lang.org/en/3.2/Struct.html)?

It would provide the same funcionality, plus keyword arguments and a couple extras that might come in handy.

modules/storages/app/common/storages/peripherals/storage_interaction/nextcloud/file_id_query.rb
```diff
@@ -0,0 +1,45 @@
...
+#++
+
+module Storages::Peripherals::StorageInteraction::Nextcloud
+  class FileIdQuery
+    def initialize(storage)
+      @query = ::Storages::Peripherals::StorageInteraction::Nextcloud::Internal::PropfindQuery.new(storage)
+    end
+
+    def call(path:)
+      query_params = {
```
mereghost: 🟡 Since `query_params` is treated like a local variable and the big difference is just the depth argument, it is very hard to notice the difference between this an the `FileIdsQuery`.

🤔 Both call the same endpoints, differ by 1 parameter. Should they be separate entities at all?

modules/storages/app/common/storages/peripherals/storage_interaction/nextcloud/files_info_query.rb
```diff
@@ -0,0 +1,157 @@
...
+          Util.error(:error, 'Outbound request failed!', response_object)
+        end
+      end
+    end
+
+    # rubocop:disable Metrics/AbcSize
+    def create_storage_file_infos
+      ->(response_object) do
+        ServiceResult.success(
+          result: response_object.ocs.data.to_h.map do |key, value|
```
mereghost: 🟡 `to_h.map` seems superfluous, specially since you are iterating over key and value, a simpler call would be `map do |(key, value)|`.

Are we covering for possible `null` data? Would a call to the more strict `Hash()` method be better then? This method raises an `ArgumentError` in case the underlying value is impossible to transform into a hash (think of a number or string).

modules/storages/app/common/storages/peripherals/storage_interaction/nextcloud/files_info_query.rb
```diff
@@ -0,0 +1,157 @@
...
+            if value.statuscode == 200
+              ::Storages::StorageFileInfo.new(
```
mereghost, line 104: 🟢 Long list of positional arguments give me the creeps. >.<

modules/storages/app/common/storages/peripherals/storage_interaction/nextcloud/files_info_query.rb
```diff
@@ -0,0 +1,157 @@
...
+                nil,
+                nil,
+                nil,
+                nil,
+                nil,
+                nil,
+                nil,
+                nil,
+                nil,
+                nil
```
mereghost: 🟢 Long list `nil` obligatory args makes me very `nil`hilist.

modules/storages/app/models/storages/storage_file_info.rb
```diff
@@ -0,0 +1,51 @@
...
+    @mime_type = mime_type
+    @size = size
+    @owner_name = owner_name
+    @owner_id = owner_id
+    @trashed = trashed
+    @last_modified_by_name = last_modified_by_name
+    @last_modified_by_id = last_modified_by_id
+    @permissions = permissions
+    @location = location
+  end
```
mereghost: 🟢 Written as a Struct this would look like:

```suggestion
Storages::StorageFileInfo = Struct.new('StorageFileInfo'
  :status, 
  :status_code, 
  :id, 
  :name, 
  :last_modified_at, 
  :created_at,
   :mime_type,
   :size,
   :owner_name,
   :owner_id, 
   :trashed, 
   :last_modified_by_name, 
   :last_modified_by_id, 
   :permissions, 
   :location
)
```

This gives us the ability to: use positional args, use keyword args, just pass the necessary values with keyword args, normal attribute getters, hash like access to them by use of [] among others.

We could use Data, and then override the `initialize` method to make some of the properties optional (by having defaults)

modules/storages/app/services/storages/file_link_sync_service.rb
```diff
@@ -26,205 +26,89 @@
...
-      sync_nextcloud(storage_id, storage_file_links)
+    file_links
+      .group_by(&:storage_id)
+      .map { |storage_id, storage_file_links| sync_nextcloud(storage_id, storage_file_links) }
+      .reduce(ServiceResult.success(result: [])) do |state, sync_result|
+      state >> ->(state_result) do
+        sync_result.match(
+          on_success: ->(sr) { ServiceResult.success(result: state_result + sr) },
+          on_failure: ->(_) { ServiceResult.success(result: state_result) }
+        )
```
mereghost: 🟡 I'm all for functional programming, but won't this generate a new `ServiceResult` per interaction? Wouldn't be easier to instantiate it once with the aggregated results in order to not put the `GarbageCollector` under pressure?

modules/storages/app/services/storages/file_link_sync_service.rb
```diff
@@ -26,205 +26,89 @@
...
-  # Get the OAuthClientToken that will authenticate us against Nextcloud
-  def get_connection_manager(storage_id)
-    oauth_client = OAuthClient.find_by(integration_id: storage_id)
-    ::OAuthClients::ConnectionManager.new(user: @user, oauth_client:)
+  def sync_nextcloud(storage_id, file_links)
+    ::Storages::Peripherals::StorageRequests
+      .new(storage: ::Storages::Storage.find(storage_id))
+      .files_info_query
+      .call(user: @user, file_ids: file_links.map(&:origin_id))
+      .map(&to_hash)
```
mereghost: 🟡 I'd expand this block as this notation might lead people to wonder if it is correct or missing the `:`

modules/storages/app/services/storages/file_link_sync_service.rb
```diff
@@ -26,205 +26,89 @@
...
-      response = request_files_info(token, file_links.map(&:origin_id))
-      Rails.logger.debug { "Got nextcloud filesinfo response: #{response.inspect}" }
-      # Parse HTTP response an return a ServiceResult with:
-      #   success: result= Hash with Nextcloud filesinfo (name of endpoint) data
-      #   failure: result= :error or :not_authorized
-      parse_files_info_response(response)
-    end
-    # Pass errors from Nextcloud into service result
-    @service_result.merge!(nextcloud_request_result)
+  def set_file_link_permissions(file_links)
```
mereghost, line 73: 🟡 In general, having a lambda defined in a method, means that every time the method is called a new lambda is created. Again, this can put pressure on the GarbageCollector and generate a large number of trash objects.

One way around this is to define them as constants and refer to those.

mereghost (commented):
That's a ton. >.< I've used an traffic light system to make notes on what I think is:

🟢 Just me being weird or minor suggestions
🟡 Points that warrant clarification discussion
🔴 (which I use sparingly) Major issues that need to be addressed before a merge

modules/storages/app/common/storages/peripherals/storage_file_info_converter.rb
apfohl: Never thought about this tbh. Then it would make sense for all domain objects to be that way. At least if they don't have persistence. I will try it out to model the `StorageFile` as a `Data.define`. Good idea.

modules/storages/app/common/storages/peripherals/storage_interaction/nextcloud/file_id_query.rb
apfohl: Please define "entity". For me an entity is a data object in persistence.

If you mean, having multiple slightly different queries. It's because of in-code visibility. I want to make them distinguishable by name, not by a parameter that might be left out.

We can try to find a more elegant solution while keeping the explicit names intact though.

modules/storages/app/common/storages/peripherals/storage_interaction/nextcloud/files_info_query.rb
apfohl: > `to_h.map` seems superfluous, specially since you are iterating over key and value, a simpler call would be `map do |(key, value)|`.

Didn't know that, thx. (was not suggested by the IDE and I don't read all the APIs ;)

> Are we covering for possible null data? Would a call to the more strict Hash() method be better then? This method raises an ArgumentError in case the underlying value is impossible to transform into a hash (think of a number or string).

Good point. Would it be a domain error or a technical one? A domain error I would transpose into a `StorageError`.

modules/storages/app/common/storages/peripherals/storage_interaction/nextcloud/files_info_query.rb
apfohl: Yes, but as it's a class and needs to have that many attributes because it's basically a DTO. It needs to be like that.

Addition: I will try to resolve this issue with `Data.define`.

modules/storages/app/common/storages/peripherals/storage_interaction/nextcloud/files_info_query.rb
apfohl: Yes, but as it's a class and needs to have that many attributes because it's basically a DTO. It needs to be like that.

Addition: I will try to resolve this issue with Data.define or default parameter values.

modules/storages/app/models/storages/storage_file_info.rb
apfohl: This.

modules/storages/app/services/storages/file_link_sync_service.rb
apfohl: I can try, but it will look uglier.

modules/storages/app/services/storages/file_link_sync_service.rb
apfohl: I could also make `to_hash` an actual function that does not return a function and use `.map(:method(to_hash))`?

wdyt?

modules/storages/app/services/storages/file_link_sync_service.rb
apfohl: Could you give an example please?

modules/storages/app/common/storages/peripherals/storage_interaction/nextcloud/files_info_query.rb
mereghost: In this case I'd make it a technical one, as the remote response sent us garbage or changed its behaviour and we didn't react in time.

modules/storages/app/services/storages/file_link_sync_service.rb
mereghost: You could make it into a function that takes a parameter and write the block in a straight forward manner `map {|info| info_to_hash(info) }` would be clearer, IMO.

modules/storages/app/services/storages/file_link_sync_service.rb
mereghost: For the file info one, this gets tricky as it depends on instance methods, so the natural solution would be not making it a lambda. One workaround would be to make the method just return the lambda, assign its return to a local variable and reuse that variable.

```ruby
def set_link_permissions
  ->(file_links) { magic_with(file_links) }
end

def call
  permissions_setter = set_link_permissions
  Request.match(on_success: permissions_setter.call(file_links), on_failure: panic! }
end

File: modules/storages/app/common/storages/peripherals/storage_interaction/nextcloud/file_id_query.rb (hunk 2 above)
Comment (mereghost) [reply modules/storages/app/common/storages/peripherals/storage_interaction/nextcloud/file_id_query.rb]: I was using the term loosely here. Most meaning Object, from a behaviour standpoint both do the same thing, return a list of `file_info`. It just happens that one is a list consisting of one Object and the other possibly a longer list (but can also be just one Object).

Is it worth to have 2 different interfaces? Or have an unified one differing by what we are telling it to do?

Comment (apfohl) [reply modules/storages/app/common/storages/peripherals/storage_interaction/nextcloud/file_id_query.rb]: No, you are right, having the one with arrays is enough. I did the same for the `FilesInfoQuery`.

File: modules/storages/app/services/storages/file_link_sync_service.rb (hunk 7 above)
Comment (apfohl) [reply modules/storages/app/services/storages/file_link_sync_service.rb]: Actually better ;)

File: modules/storages/app/common/storages/peripherals/storage_interaction/nextcloud/files_info_query.rb (hunk 3 above)
Comment (apfohl) [reply modules/storages/app/common/storages/peripherals/storage_interaction/nextcloud/files_info_query.rb]: I tried doing it both ways, `Hash(response_object.ocs.data)` which leads to:

```
TypeError:
       can't convert OpenStruct into Hash
```

With `response_object.ocs.data.map do |(key, value)|` the result is somehow `nil`???

mereghost: Since that's a `ostruct`, you can call `.each_pair.map` which will avoid the intermediate transformation into a hash.

modules/storages/app/services/storages/file_link_sync_service.rb
apfohl: Okay, understood. In this case the function is only called as many times as there are storages configured in the specific project. I would think, that it is therefor okay for now. wdyt? It's hard to extract right now, because it depends on the list of file links for the selected storage which are kind of partially applied. The only place would be inside the `sync_nextcloud` function, but it would be only called once.

mereghost: I'm all for going forward as it is and later on addressing it. =)

mereghost (approved):
LGTM! 🎉

PR #13235, opened 2023-07-25
[#49331] Added project folder deletion on project storage deletion

Kharonus:
- https://community.openproject.org/work_packages/49331
- added before_perform hook to project storage delete service
  - skip if project folder was not found
  - cancel deletion, if deletion is not possible

### DIFF
--- a/modules/storages/app/services/storages/project_storages/delete_service.rb
+++ b/modules/storages/app/services/storages/project_storages/delete_service.rb
@@ -30,6 +30,8 @@ module Storages::ProjectStorages
   # Performs the deletion in the superclass. Associated FileLinks are deleted
   # by the model before_destroy hook.
   class DeleteService < ::BaseServices::Delete
+    using Storages::Peripherals::ServiceResultRefinements
+
     # rubocop:disable Metrics/AbcSize
     def before_perform(params, service_result)
       before_result = super(params, service_result)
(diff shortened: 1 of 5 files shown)


modules/storages/app/services/storages/project_storages/delete_service.rb
```diff
@@ -30,6 +30,30 @@ module Storages::ProjectStorages
...
+      before_result = super(params, service_result)
+      return before_result if before_result.failure? || !model.storage.is_a?(Storages::NextcloudStorage)
+
+      Storages::Peripherals::StorageRequests
+        .new(storage: model.storage)
+        .delete_folder_command
+        .call(location: model.project_folder_path)
+        .on_success { |result| result.result = model }
+        .on_failure do |result|
+        if result.errors.code == :not_found
```
mereghost: Should we change the semantics of the command then?

mereghost (approved):
Looks good, just a minor question about the semantics of the delete folder service.

modules/storages/app/services/storages/project_storages/delete_service.rb
Kharonus: you think, it should not return an error if `not_found`? I thought about it, but IMHO it is the right place to check it here. Usually I would expect an error, when trying to delete a non-existing file. But the deletion service needs to harden itself against the error itself and continue operating, because this error does not block execution.

You think differently? At least for my part this code makes more sense like this.

- we have an error while deleting
- we do want to continue if it is the very specific error of not finding stuff

I'm just not very satisfied with the readability. :/

mereghost: I think that the `delete_folder_command` should probably return a success if it returns a 404, since, well... if it doesn't exist - either because I can't see it or it was never there - it is gone now! :P

apfohl: I think there is not right or wrong. There are two ways to look at it:

1. You observe the result of that call. In that case @mereghost is right, only the outcome counts. But it indeed hides something. If we don't need to care about it that's a valid option.
2. You observe the actual functioning of the call. That would mean to make it not lie and therefore don't hide the fact that it can fail. You would need to take care of that by handling the error after the call. That would make it explicit for other programmers what is happening and don't hide something.

If you would ask me, I would lean a bit towards the 2nd approach. But not because either of them is wrong, but due to the fact of ruby being the way it is (no clear function signatures). Also the error handling is super simple to do with `.match()`.

PR #13238, opened 2023-07-25
[#49081] PDF-Export: Add tests for pdf content

as-op:
This PR adds:
* Best case tests for WorkPackage::PDFExport::WorkPackageListToPdf
* Best case tests for WorkPackage::PDFExport::WorkPackageToPdf

https://community.openproject.org/work_packages/49081


spec/models/work_packages/pdf_export/work_package_to_pdf_spec.rb
```diff
@@ -0,0 +1,120 @@
...
+    WorkPackage.human_attribute_name(column_name)
+  end
+
+  subject(:pdf) do
+    content = export_pdf.content
+    # File.binwrite('WorkPackageToPdf-test-preview.pdf', content)
+    { strings: PDF::Inspector::Text.analyze(content).strings,
+      images: PDF::Inspector::XObject.analyze(content).page_xobjects.flat_map do |o|
+        o.values.select { |v| v.hash[:Subtype] == :Image }
+      end }
```
klaustopher, line 89: ❓ In the other test we are writing the PDF to disk, here we are checking the content directly. Can't we use the same procedure in the other file as well?

spec/models/work_packages/pdf_export/work_package_to_pdf_spec.rb
```diff
@@ -0,0 +1,120 @@
...
+    end
+  end
+
+  def column_title(column_name)
+    label_title(column_name).upcase
+  end
+
+  def label_title(column_name)
+    WorkPackage.human_attribute_name(column_name)
+  end
```
klaustopher: It's only 2 methods, but they are the same here and in the other test. We could put them in a module in `spec/support/` and include that module in both tests

klaustopher (approved):
Left 2 remarks inside, otherwise this is fine

spec/models/work_packages/pdf_export/work_package_to_pdf_spec.rb
as-op: The (other) query based export does have a feature to batch exporting thousands of work packages (with images) to different pdfs and concats them in the end with an external tool. Granted, the batching is currently not triggered in the spec, but this change wouldn't have much benefits.

klaustopher: 👍

spec/models/work_packages/pdf_export/work_package_to_pdf_spec.rb
as-op: Thanks, done!

PR #13241, opened 2023-07-25
fix(doc): incoming email: config:set instead of config:get

d--j:


ulferts (approved):
Hi @d--j , you are right of course. Thanks for fixing.

PR #13250, opened 2023-07-26
[#48649] PDF Export: Filter attributes table

as-op:
This PR
* refactors work package attribute table generation exports to avoid dup code for single/multi-wp-export
* removes all attributes with empty values from the table

https://community.openproject.org/work_packages/48649


app/models/work_package/pdf_export/work_package_detail.rb
```diff
@@ -95,54 +91,62 @@ def attributes_table_column_widths
...
 
-  def build_attributes_table_rows(work_package)
-    # get work package attribute table rows data [[label, value, label, value]]
-    attrs = %i[
+  def attribute_data_list(work_package)
+    list = if respond_to?(:column_objects)
+             attributes_list_by_columns(work_package)
+           else
+             attributes_list_by_wp(work_package)
+           end
```
aaron-contreras: It seems these don't need `work_package` as an argument at all. Perhaps

```suggestion
    list = if respond_to?(:column_objects)
             attributes_list_by_columns
           else
             attributes_list_by_wp
           end
```

app/models/work_package/pdf_export/work_package_detail.rb
```diff
@@ -95,54 +91,62 @@ def attributes_table_column_widths
...
+    list
+      .map { |entry| entry.merge({ value: get_column_value_cell(work_package, entry[:name]) }) }
+      .select { |attribute_data| can_show_attribute?(attribute_data) }
+  end
+
+  def can_show_attribute?(attribute_data)
+    attribute_data[:value].present?
+  end
+
+  def attributes_list_by_columns(_work_package)
```
aaron-contreras:
```suggestion
  def attributes_list_by_columns
```

app/models/work_package/pdf_export/work_package_detail.rb
```diff
@@ -95,54 +91,62 @@ def attributes_table_column_widths
...
+    column_objects
+      .reject { |column| column.name == :subject }
+      .map do |column|
+      { label: column.caption || '', name: column.name }
+    end
+  end
+
+  def attributes_list_by_wp(_work_package)
```
aaron-contreras:
```suggestion
  def attributes_list_by_wp
```

aaron-contreras (changes requested):
Very nice refactor. Only itch would be the unused parameter for both `#attributes_list_by_columns` and `#attributes_list_by_wp`.

as-op:
🙏 Thank you @aaron-contreras!

PR #13257, opened 2023-07-26
load the environment to have the crowdin path added to I18n.load_path

ulferts:
Loading the environment will trigger the [assignment of the crowdin files to I18n.load_path](https://github.com/opf/openproject/blob/dev/config/application.rb#L139)


aaron-contreras (approved):
👍

PR #13272, opened 2023-07-27
[#49425] PDF export of descriptions - HTML in table with header column

as-op:
This PR changes the way unknown/unexpected HTML tags are handled in
telling md-to-pdf gem to iterate through the HTML-AST to 
* print out text nodes
* handle known HTML tags in with unknown tags

https://community.openproject.org/work_packages/49425


ulferts (approved):
👍

PR #13284, opened 2023-07-28
[#48623] Missing spacing between fields on project information form

mrmir:
https://community.openproject.org/work_packages/48623

### DIFF
--- a/frontend/src/app/shared/components/forms/form.sass
+++ b/frontend/src/app/shared/components/forms/form.sass
@@ -24,3 +24,7 @@
     > .button
       &:not(:last-child)
         margin-bottom: 1rem
+
+  &--field
+    .spot-form-field
+      margin-bottom: 1rem


HDinger (commented):
The issue appears because the fieldset is expecting the fields to be direct children (see [here](https://github.com/opf/openproject/blob/5101ba3fd36f2bd64a2b0627ab5e58e3d557aecc/frontend/src/app/shared/components/forms/form.sass#L17C1-L26C1)). The html structure of these pages have changed. There are two additional tags (`formly-field` and `formly-group`) in between.
Please check wether these additional elements are necessary (at first glance it seems like they don't but I did not check the code yet).

HDinger (approved):
It seems like the changed structure is a side effect of an update, so we won't touch that. Therefore, the changes are fine. Thanks 👍

PR #13286, opened 2023-07-28
[#49248] Copy FileLinks respecting the location of the files

ba1ash:
https://community.openproject.org/work_packages/49248


modules/storages/app/services/projects/copy/file_links_dependent_service.rb
```diff
@@ -39,31 +39,88 @@ def source_count
...
-      state.work_package_id_lookup.each do |old_wp_id, new_wp_id|
-        create_work_package_file_links(old_wp_id, new_wp_id)
-      end
-    end
+      source_wp_ids = state.work_package_id_lookup.keys
+      Storages::FileLink
+        .where(container_id: source_wp_ids, container_type: "WorkPackage")
+        .group_by(&:storage_id)
+        .each do |(storage_id, source_file_links)|
+        if source_file_links.any?
```
apfohl: Maybe use a filtering above, to exclude these cases entirely from `.each` block. That would reduce the nesting and noise.

apfohl (approved):
Generally I would suggest breaking the `file_links_dependent_service` down into smaller methods. But due to the fact getting it merged I leave this discussion for later.

There is one small thing, I would like you to think about tho. I marked it.

Besides that I'm fine with the result. Good job.

PR #13295, opened 2023-07-31
Fix misaligned sidemenu icons

b12f:
Ref https://community.openproject.org/work_packages/49316/activity


HDinger (commented):
Icons are correctly aligned 👍 

I am not so happy with putting the svg path the test expectation. Is there a way to avoid that and simply test for a `svg` independtly of the content?
Other than that, it is fine to merge once the tests are green 👍

PR #13296, opened 2023-07-31
preview and restore backups

machisuji:
WP [#47096](https://community.openproject.org/projects/openproject/work_packages/47096/activity)


app/components/backups/row_component.rb
```diff
@@ -0,0 +1,79 @@
...
+    def button_links
+      links = []
+
+      if backup.ready?
+        links << helpers.link_to("Download", "/attachments/#{backup.attachments.first.id}", class: "icon icon-download")
+        links << helpers.link_to("Preview", preview_admin_backup_path(backup.id), class: "icon icon-watched")
+        links << helpers.link_to("Restore", restore_admin_backup_path(backup.id), class: "icon icon-import")
+      end
+
+      links << helpers.link_to("Delete", admin_backup_path(backup.id), method: :delete, class: "icon icon-delete")
```
ulferts: These links need to be i18ned

app/components/backups/row_component.rb
```diff
@@ -0,0 +1,79 @@
...
+      else
+        "?"
+      end
+    end
+
+    def button_links
+      links = []
+
+      if backup.ready?
+        links << helpers.link_to("Download", "/attachments/#{backup.attachments.first.id}", class: "icon icon-download")
```
ulferts: Use the routes helper so that relative url paths don't mess with it.

app/controllers/admin/backups_controller.rb
```diff
@@ -53,7 +74,142 @@ def show
...
+    schema = RestoreBackupJob.preview_schema_name backup_id: backup_id
+
+    cookies.signed[:backup_preview] = backup.attributes
+      .merge(schema: schema, previous_user_id: current_user.id)
+      .symbolize_keys
+      .to_yaml
+
+    cookies.signed[:login_user_id] = get_schema_user_id schema
+
+    redirect_to "/admin/backups/"
```
ulferts: Using `redirect_to action: :index` should be safe against having a relative url path. I am not sure if that would be the case with this string as well.

app/controllers/admin/backups_controller.rb
```diff
@@ -53,7 +74,142 @@ def show
...
+
+      return redirect_back_or_default home_path
+    end
+
+    backup_id = backup_preview[:id]
+
+    if RestoreBackupJob.preview_active?(backup_id: backup_id).blank?
+      flash[:error] = "Not currently previewing anything"
+
+      return redirect_to "/admin/backups/"
```
ulferts: Same as above

app/controllers/admin/backups_controller.rb
```diff
@@ -53,7 +74,142 @@ def show
...
+    backup.save!
+
+    backup.size_in_mb = (backup.attachments.first.filesize / 1024.0 / 1024.0).round(2)
+    backup.save!
+
+    JobStatus::Status.create(reference: backup, message: "imported", status: :success)
+
+    flash[:info] = I18n.t("backup.notice_uploaded", comment: backup.comment)
+
+    redirect_to "/admin/backups"
```
ulferts: Same as above

app/controllers/admin/backups_controller.rb
```diff
@@ -53,7 +74,142 @@ def show
...
+  end
+
+  def destroy
+    backup = Backup.find params[:id]
+
+    backup.destroy!
+
+    flash[:info] = I18n.t("backup.notice_deleted")
+
+    redirect_to "/admin/backups/"
```
ulferts: Same as above

app/helpers/omniauth_consent_helper.rb
```diff
@@ -0,0 +1,33 @@
...
+      end
+    end
+  end
+
+  def request_omniauth_consent(user)
+    raise "User does not authenticate using OmniAuth" unless user.uses_external_authentication?
+
+    login_flash[:omniauth_consent_requested] = true
+
+    redirect_to "/auth/#{current_user.authentication_provider.downcase}?prompt=consent"
```
ulferts, line 25: define a proper path, please

config/routes.rb
```diff
@@ -422,8 +422,19 @@
...
+        get :preview
+        get :restore
+      end
+
       collection do
+        get :back
+        get :restored
+
+        get :upload
+        post :upload, action: :perform_upload
```
ulferts, line 436: The `perform_upload` method is private so that the ActionDispatch cannot find it.

app/controllers/admin/backups_controller.rb
```diff
@@ -65,31 +139,138 @@ def perform_token_reset
...
+
+    JobStatus::Status.create(reference: backup, message: "imported", status: :success)
+
+    flash[:info] = I18n.t("backup.notice_uploaded", comment: backup.comment)
+
+    redirect_to "/admin/backups"
+  end
+
+  def create_uploaded_backup # rubocop:disable Metrics/AbcSize
+    Backup.new creator: current_user, comment: params[:comment]
```
ulferts: There is a variable assignment to `backup` missing as that is referred to down in the rest of the method

app/models/backup.rb
```diff
@@ -1,9 +1,13 @@
...
+    def create_permission
       :create_backup
     end
 
+    def restore_permission
+      :restore_backup
+    end
+
     def include_attachments?
       val = OpenProject::Configuration.backup_include_attachments
```
ulferts, line 12: Not really strictly related to this PR but what about files in Nextcloud? Those are not part of the backup which might be ok since they are hosted in a different, user accessible, system. But if we now restore a backup there might now be `Storage::FileLink` records referencing files not existing in Nextcloud any more. And I don't know if there is also information stored on Nextcloud side which could become outdated by restoring a backup.

app/controllers/admin/backups_controller.rb
```diff
@@ -65,31 +139,138 @@ def perform_token_reset
...
+    backup.attachments.build file: params[:backup_file], author: current_user
+    backup.save!
```
ulferts: In case there is an error, an exception is thrown which will result in a 500 being presented to the user instead of error messages the user can then react to.

app/controllers/admin/backups_controller.rb
```diff
@@ -65,31 +139,138 @@ def perform_token_reset
...
+
+    backup.size_in_mb = (backup.attachments.first.filesize / 1024.0 / 1024.0).round(2)
+    backup.save!
+
+    backup
```
ulferts: A backup file can quickly become quite large. This might then conflict with the `Settings.attachment_max_size` value. My advise at this point is to extract this into a service + contract where you can then work around the setting for this scenario. This might also make it easier to deal with error messages of the attachments.

ulferts (changes requested):
As agreed upon, the intermediary review results, @machisuji .

app/models/backup.rb
machisuji: I think it is ok not to include the NextCloud files but yes, restoring a backup might mess things up there. I think that's unavoidable, though. It might lose the connection to NextCloud altogether because it hadn't been configured yet at the time of the backup or perhaps the credentials have changed in the meantime.

@Kharonus Can you think of something on the NextCloud side that could break when one restores a backup? I guess it could be that work packages referenced in files no longer exist. Can the OpenProject extension in NextCloud handle that?

I reckon it can. I think this is just something one has to accept when doing this.

PR #13297, opened 2023-07-31
Implementation/48572 adapt work package eager loading after custom fields are included in baseline

ulferts:
Moves the eager loading of work package custom fields in the call stack involved in the rendering of work package collections in the API v3.  

This mostly involves the loading of custom fields. Before, a delegation object was wrapping every work package rendered and that wrapper got the `#available_custom_fields` prefilled. With the addition of `attributesAtTimestamp`, where work packages at a different time are rendered, this model proved too unflexible. 

With the change in this PR, the available custom fields are cached in the request store. That enables to have the fields of shared between work packages of the same project and type. There is also a method added to preload the available custom fields for a set of work packages (`WorkPackage.preload_available_custom_fields`). That method is used in the eager loading to replace the wrapper mentioned before. The implementation of this is also way simpler.

There is the limitation of not being able to include the work packages in the `attributesAtTimestamp` in the preloading. So if a work package changed project or type to a tuple not shared by a work packages in the original set, the database is hit again. While not being an edge case, the assumption behind accepting the limitation is still, that it should not be very common. Removing that limitation has proven to be quite hard given the order the eager loading takes place (`::API::V3::WorkPackages::EagerLoading::CustomValue` before `::API::V3::WorkPackages::EagerLoading::HistoricAttributes`). Changing the order leads to new N+1 queries as the historic values require to have the available custom fields in order to create the diffs.

Not included in the optimization are also the loading of models referenced in the `attributesAtTimestamp` (e.g. a status). 

The optimization yields considerable improvements in the dataset used. With a page size of 500 work packages, the response time dropped from roughly 5s to 2.7s.

https://community.openproject.org/wp/48572


lib/api/v3/work_packages/eager_loading/custom_value.rb
```diff
@@ -126,66 +125,14 @@ def custom_values_of(field_format)
...
-          end
-
-          def usage_hash
-            Hash.new do |by_project_hash, project_id|
-              by_project_hash[project_id] = Hash.new do |by_type_hash, type_id|
-                # Use a set to ensure CFs are only available once
-                by_type_hash[type_id] = Set.new
-              end
-            end
+          def timestamped?
```
dombesz, line 134: I cannot find any reference for the use of this method, is it used?

dombesz (approved):
Nice job @ulferts, the refactoring made the code simpler and the number of queries are also significantly reduced by this change.👏

lib/api/v3/work_packages/eager_loading/custom_value.rb
ulferts: You are right, @dombesz . It is a leftover of a previous attempt. I'll remove it.

PR #13300, opened 2023-07-31
[49395] Adress style issues coming from Primers different base styles

bsatarnejad:
- [x] Change font-size from 16px to 14px
- [x] Remove back-ground color of filters
- [x] Override styles for button in toolbar to remove underline style while hovering over it
- [x] Fix Enterprise banners misalignment both home page block and EE banner component
- [x] Remove extra spacing from user-activity header
- [x] Fix bold labels in notification settings
- [x] fix misalignment of checkboxes
- [x] Fix top bar text of project selection modal which is cut off from bottom
- [x] User created contend should have the base body font size
- [x] Override styles for button in spot-action-bar to remove underline style while hovering over it
- [x] Inputs shouldn't be bold
- [x] Labels with option-label class should not be bold
- [x] Remove underline for buttons inside the upsale page

https://community.openproject.org/projects/openproject/work_packages/49395/activity


frontend/src/app/features/user-preferences/notifications-settings/page/notifications-settings-page.component.html
```diff
@@ -8,6 +8,7 @@ <h5>{{ text.notifyImmediately.title }}</h5>
   <p>{{ text.notifyImmediately.description }}</p>
 
   <spot-selector-field
+    labelWeight="bold"
```
HDinger: Since bold should be the default, I would not add it to every spot-selector field, but rather change the default in the [component itself](https://github.com/opf/openproject/blob/b99d0381a53aa1ee0e57f24fd95239b28bba3a66/frontend/src/app/spot/components/selector-field/selector-field.component.ts#L40)

Otherwise it would be regular again if someone forgets to add the `labelWeight` attribute.

frontend/src/app/features/work-packages/components/wp-baseline/baseline/baseline.component.sass
```diff
@@ -51,7 +51,7 @@
       border: 1px solid $spot-color-basic-gray-4
       border-left: none !important
       padding: $spot-spacing-0_5 $spot-spacing-0_25
-      line-height: unset
+      line-height: 1rem
```
HDinger, line 54: This change results in the time info being not correclty aligned

Bildschirmfoto 2023-08-03 um 09 26 58

HDinger (changes requested):
Hi @bsatarnejad , great job! 👍  I only have very minor remarks. 

Please also have a look at the toggle button in ProjectInclude and take care that the labels are not bold:

Bildschirmfoto 2023-08-03 um 09 12 49

HDinger (approved):
👍 Good to merge once the tests are green

PR #13304, opened 2023-07-31
Docker slim image

crohr:
Adds a `slim` build target for our official docker images, which can be used for more advanced deployments where you don't need all the packages that come with the default `all-in-one` build.

before vs after

Changes:
* builds and compiles gems, node modules, and assets in temporary stages, and only copies what's needed in `slim`. Resulting image is ~800MB (vs 2.7GB for all-in-one).
* introduces a `base` image for both `slim` and `all-in-one` targets.
* base ruby image is the `slim` variant, which has a lot less build dependencies included. It may cause some issues with the downstream flavours, if some specific system dependencies are missing.
* removes the need for the `BASE_PREFIX` arg, which was used to force a specific platform. buildx natively supports multi-platform.
* updates the `docker.yml` workflow so that it pushes the `slim` variant anytime the default image is generated.

For now I didn't make too many changes to the way the `all-in-one` image is generated. Keep in mind that this target theoretically needs to be able to recompile additional plugins / assets.

@machisuji looks like the push did work: https://hub.docker.com/repository/docker/openproject/community-test/tags?page=1&ordering=last_updated


machisuji (approved):
👍

PR #13308, opened 2023-07-31
[#49410] Ensure Baseline functionality is reset when switching to a Builtin View

aaron-contreras:
The Baseline component was not watching for changes to the "pristine" state of
the current timestamp query being requested.

By watching for the pristine state being requested, now we can disable and reset
the Baseline details when selecting a builtin view (like "All open") which
doesn't include a timestamp (which would default the timestamp to 'PT0S').

However, when a saved query that includes a timestamp on it is selected due to
it have been saved that way, this will also cover the case of ensuring that
Baseline is not disabled and reset as the timestamp would not be the default.

See: https://community.openproject.org/work_packages/49410


dombesz (approved):
LGTM, thanks Aaron!

PR #13323, opened 2023-08-01
[#48160] Fix 403 when creating a Meeting globally without a Project

aaron-contreras:
Addresses note 1 in [this comment left by Ivana](https://community.openproject.org/work_packages/48160/activity#activity-21)

Ensure `authorize_global` is the hook used for the create action instead of `authorize`. By doing this, we ensure that based on `@project` being set or not, we authorize appropriately and allow the failure to save branch a chance of executing.

**BONUSES:**

1. I went ahead and optimized `meetings_new_spec` in https://github.com/opf/openproject/commit/c401daa022b5b95ac57e8bfd2f4c7f87187be8bf with an explanation to it.
2. The same issue found in note 3 in [the same comment as above left by Ivana](https://community.openproject.org/work_packages/48160/activity#activity-21) was occurring in the meetings page as well. Even though it wasn't reported, I went ahead and addressed it in https://github.com/opf/openproject/pull/13323/commits/d66ed3d3ac788188e225fcda0aab1bfb5a23bf25

### DIFF
--- a/modules/meeting/app/controllers/meetings_controller.rb
+++ b/modules/meeting/app/controllers/meetings_controller.rb
@@ -32,8 +32,8 @@ class MeetingsController < ApplicationController
   before_action :build_meeting, only: %i[new create]
   before_action :find_meeting, except: %i[index new create]
   before_action :convert_params, only: %i[create update]
-  before_action :authorize, except: %i[index new]
-  before_action :authorize_global, only: %i[index new]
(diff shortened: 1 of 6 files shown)


dombesz (approved):
Works as expected, but the experience seems to be a bit inconsistent. The title has the `required="true"` attribute, but the project and the start date do not have the required attribute. I'll create a separate PR for those.

image
image

image

PR #13332, opened 2023-08-02
Adds a section with Storage OAuth Tokens to te Profile Page

mereghost:
Implements OP#49434

Adds an extra section over the Access Tokens page, so you can "logout" from a certain storage.

image

### DIFF
--- a/app/controllers/my_controller.rb
+++ b/app/controllers/my_controller.rb
@@ -78,7 +78,27 @@ class MyController < ApplicationController
   end
 
   # Administer access tokens
-  def access_token; end
+  def access_token
+    @storage_tokens = OAuthClientToken
+                        .preload(:oauth_client)
+                        .joins(:oauth_client)
+                        .where(user: @user, oauth_client: { integration_type: 'Storages::Storage' })
+  end
+
+  def delete_storage_token
+    token = OAuthClientToken
+      .preload(:oauth_client)
+      .joins(:oauth_client)
+      .where(user: @user, oauth_client: { integration_type: 'Storages::Storage' }).find_by(id: params[:id])
+
+    if token&.destroy
+      # TODO i18n this
+      flash[:notice] = I18n.t('my.access_tokens.storages.removed')
+    else
+      flash[:error] = I18n.t('my.access_tokens.storages.failed')
+    end
+    redirect_to action: :access_token
+  end
 
   # Configure user's in app notifications
   def notifications
--- a/app/controllers/my_controller.rb
+++ b/app/controllers/my_controller.rb
@@ -93,7 +93,7 @@ class MyController < ApplicationController
 
     if token&.destroy
       # TODO i18n this
-      flash[:notice] = I18n.t('my.access_tokens.storages.removed')
+      flash[:info] = I18n.t('my.access_tokens.storages.removed')
     else
       flash[:error] = I18n.t('my.access_tokens.storages.failed')
     end
--- a/app/controllers/my_controller.rb
+++ b/app/controllers/my_controller.rb
@@ -92,7 +92,6 @@ class MyController < ApplicationController
       .where(user: @user, oauth_client: { integration_type: 'Storages::Storage' }).find_by(id: params[:id])
 
     if token&.destroy
-      # TODO i18n this
       flash[:info] = I18n.t('my.access_tokens.storages.removed')
     else
       flash[:error] = I18n.t('my.access_tokens.storages.failed')
--- a/app/views/my/access_token_partials/_storage_tokens_section.html.erb
+++ b/app/views/my/access_token_partials/_storage_tokens_section.html.erb
@@ -26,62 +26,58 @@ Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
 See COPYRIGHT and LICENSE files for more details.
 
 ++#%>
-<div class="attributes-group" id="oauth-token-section">
+<div class="attributes-group" id="storage-token-section">
   <div class="attributes-group--header">
     <div class="attributes-group--header-container">
-      <h3 class="attributes-group--header-text"><%= t("my_account.access_tokens.oauth.title") %></h3>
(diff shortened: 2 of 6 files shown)


mereghost:
@Kharonus ideally removing a token would trigger a revoke of it on the remote server, but I haven't found a way already implemented to do it. =/

Kharonus:
I'll have a look, in our API we added an endpoint to do so, will check NC API.

mereghost:
> I'll have a look, in our API we added an endpoint to do so, will check NC API.

Cool. I was afraid I was missing something that already existed.

app/controllers/my_controller.rb
```diff
@@ -78,7 +78,27 @@ def change_password
...
+  end
+
+  def delete_storage_token
+    token = OAuthClientToken
+      .preload(:oauth_client)
+      .joins(:oauth_client)
+      .where(user: @user, oauth_client: { integration_type: 'Storages::Storage' }).find_by(id: params[:id])
+
+    if token&.destroy
+      # TODO i18n this
```
akabiru: Seems like it was solved?
```suggestion
```

Kharonus: 🔧 remove todo comment

app/controllers/my_controller.rb
```diff
@@ -78,7 +78,27 @@ def change_password
   end
 
   # Administer access tokens
-  def access_token; end
+  def access_token
+    @storage_tokens = OAuthClientToken
+                        .preload(:oauth_client)
+                        .joins(:oauth_client)
+                        .where(user: @user, oauth_client: { integration_type: 'Storages::Storage' })
```
Kharonus, line 85: ❓ does this match for `integration_type: Storages::NextcloudStorage`? I'm not sure, what type we enter here. But if someone changes that in future, will it be a problem? Just asking out of curiosity, doesn't think this line needs a change.

app/views/my/access_token_partials/_storage_tokens_section.html.erb
```diff
@@ -0,0 +1,83 @@
...
+
+See COPYRIGHT and LICENSE files for more details.
+
+++#%>
+<div class="attributes-group" id="storage-token-section">
+  <div class="attributes-group--header">
+    <div class="attributes-group--header-container">
+      <h3 class="attributes-group--header-text"><%= t("my_account.access_tokens.storages.title") %></h3>
+    </div>
+  </div>
```
akabiru, line 34: 🟠  Should we use the [`AttributeGroups::AttributeGroupHeaderComponent`](https://github.com/opf/openproject/blob/dev/app/components/attribute_groups/attribute_group_header_component.rb)? AFAIK, we're leaning more towards view components with the primer adoption. Context #12917


```suggestion
  <%= render(AttributeGroups::AttributeGroupHeaderComponent.new(title: t("my_account.access_tokens.storages.title"))) %>
```

app/views/my/access_token_partials/_storage_tokens_section.html.erb
Kharonus, line 1: 📝 Those tables are not yet structured as `table-` and `row-components`? shame, but we don't have to make a lunge right now, as they will get replaced by primer components anyway.

Kharonus (approved):
Just a couple of comments, and please remove the `TODO` comment, otherwise LGTM.

spec/controllers/my_controller_spec.rb
```diff
@@ -345,5 +345,29 @@
...
+      it 'can remove the token' do
+        expect do
+          delete :delete_storage_token, params: { id: token.id }
+        end.to change(OAuthClientToken, :count).by(-1)
+
+        expect(flash[:info]).to be_present
+        expect(flash[:error]).not_to be_present
+        expect(response).to redirect_to(action: :access_token)
+      end
+    end
```
akabiru, line 371: Nice one 👍🏾 
🟢  Noticed we don't have a feature spec, do you think it's worth repurposing into a feature spec instead? I reckon it simulates a more accurate user journey

akabiru (commented):
Looking good to me! I'll leave to @Kharonus for the final approval as you already have a conversation going 🙂 

And yes, I ~stole~ borrowed your [traffic light review system](https://github.com/opf/openproject/pull/13223#pullrequestreview-1547996208) 😄

app/controllers/my_controller.rb
mereghost: 🤦🏼

spec/controllers/my_controller_spec.rb
Kharonus: Everything we can test in unit test I'd rather test in unit tests. Feature specs do have a much bigger ramp up time.

So, I'd say, unless we encounter a nasty regression we couldn't test in a unit test, we should bother with a feature specs in this view.

app/controllers/my_controller.rb
mereghost: Good question!

⚠️ I'm citing this out of the top of my mind, so details might be a little off ⚠️ 

`Storages::NextcloudStorage` is implemented as a STI on `Storages::Storage`, meaning that to for the purposes of associations it is a `Storage::Storage`.

If and when we load the association, though, the STI mechanism will instantiate the actual `Storages::SpecializedStorage`.

You can see it in place by running `OAuthClientToken.first.oauth_client.integration`

spec/controllers/my_controller_spec.rb
mereghost: @akabiru I thought about it, but seemed so minor that the overhead of the feature spec would be way to high for such a small piece of code.

@Kharonus agreed.

mereghost:
> And yes, I ~stole~ borrowed your [traffic light review system](https://github.com/opf/openproject/pull/13223#pullrequestreview-1547996208) smile

Oh, I stole it from someone else, feel free to plunder it!

app/views/my/access_token_partials/_storage_tokens_section.html.erb
mereghost: This page isn't using any components, so I just kept with what was in place. I'm not sure we should update it. =/

PR #13336, opened 2023-08-02
[#49489] Fix required inputs on the meetings form

dombesz:
https://community.openproject.org/work_packages/49489
This PR fixes the missing `required` attribute on the date selector of the new meetings form.
The Project input cannot simply receive the `required` attribute as it is an `ng-select` input, which does not supports the required tag.


aaron-contreras (approved):
Thank you! 💥

PR #13338, opened 2023-08-02
fix: make all attribute groups except for wps single column

b12f:
Ref: https://community.openproject.org/work_packages/49400/activity

### DIFF
--- a/frontend/src/app/features/work-packages/components/wp-form-group/wp-attribute-group.component.sass
+++ b/frontend/src/app/features/work-packages/components/wp-form-group/wp-attribute-group.component.sass
@@ -1,8 +1,47 @@
 @import "src/assets/sass/helpers"
 
 .wp-attribute-group
+  display: flex
+  flex-direction: row
+  flex-wrap: wrap
+
+  // We want a margin between items, but don't know if any one item is on the right,
+  // the left, or spanning both columns. The easy way to get a gutter is to give
+  // *all* elements margins on both sides, and then negating this margin on the parent.
+  // This also explains all of the + and - xrem calc() values below.
+  margin: 0 -1 * $spot-spacing-1_5
+  width: calc(100% + #{$spot-spacing-3})
+  
   &--attribute
+    flex-basis: 400px
+    flex-grow: 1
+    flex-shrink: 1
+    min-width: calc(50% - 3rem)
+    max-width: calc(100% - 3rem)
+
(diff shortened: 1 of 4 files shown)


frontend/src/global_styles/content/_attributes_key_value.sass
```diff
@@ -31,11 +31,6 @@
   // flex-wrap: wrap
   font-size: 0.875rem
   margin-bottom: 0.1875rem
-  flex-basis: 400px
-  flex-grow: 1
-  flex-shrink: 1
-  min-width: calc(50% - 3rem)
-  max-width: calc(100% - 3rem)
   margin: 0 1.5rem
```
HDinger: As the attributes are now all below each other we can get rid of this extra margin and the magical calculation done here: https://github.com/opf/openproject/blob/a705bc530f4a2bd87d0f8e03ae4244b9b8b82bca/frontend/src/global_styles/content/_attributes_group.sass#L85C1-L90C1

frontend/src/app/features/work-packages/components/wp-form-group/wp-attribute-group.component.sass
```diff
@@ -1,8 +1,47 @@
 @import "src/assets/sass/helpers"
 
 .wp-attribute-group
+  display: flex
```
HDinger, line 23: Wouldn't it be better to use a `grid` here?  Thus we could avoud all these magic calculations

HDinger (changes requested):
It works fine. 👍 I have some remarks to the code.

frontend/src/app/features/work-packages/components/wp-form-group/wp-attribute-group.component.sass
b12f: Unfotunately no, because of the requirement that the the fields go single column at some point, but we cannot use breakpoints since the width of the layout is independent of the width of the screen. You can show/hide and drag the sidemenu and split screen.

HDinger (approved):
@b12f  I changed the wp attribute to a grid. I _think_ that I catched all cases, but please double check that everything works as expected before merging.

b12f:
Hi @HDinger unfortunately this breaks mobile screens of less than 432px width because the 400px minimum are a hard limit:

20230808_10h53m08s_grim

PR #13339, opened 2023-08-02
[#49430] Ensure long text values don't spill over in the WP view

aaron-contreras:
By adding `overflow: hidden` to the `.attributes-key-value--value-container` we ensure that no fields (name or not) will spill over into their adjacent fields.

See: https://community.openproject.org/work_packages/49430

**Previews:**
image
image


HDinger (approved):
LGTM 👍

PR #13357, opened 2023-08-03
[#49518] Uploading a folder to Nextcloud fails with wrong error message

Kharonus:
- https://community.openproject.org/work_packages/49518
- add additional check if drop content was a folder


ulferts (approved):
Thanks for fixing.

PR #13364, opened 2023-08-04
improve performance of validity_period migration

ulferts:
Duplicating the `successors.created_at <= predecessors.created_at` condition to also be in the inner subselect (the one calculating the successors) reduces the set of successors necessary for the later calculation at an earlier state already.


cbliard (approved):
I tried the migration and got similar results with the updated version.

PR #13366, opened 2023-08-04
Optimized Docker TLS dev stack

apfohl:
- removed volume mount to host ceatificates for traefik
- mount step ca root certificate directly into traefik
- moved TLS settings to global traefik CLI parameters


Kharonus:
@machisuji Not sure, if you want to have a quick review of it. ;)

@apfohl and me are trying to make it a bit easier to setup.

PR #13367, opened 2023-08-04
[#45533] Work package relationships are lost when copying

mrmir:
https://community.openproject.org/wp/45533


mrmir:
Closed in favour of #13713

PR #13382, opened 2023-08-07
[#49501] Log time modal is not scrollable

Kharonus:
- https://community.openproject.org/work_packages/49501
- removed `no_scroll` modifier for `spot-model--body`

⚠️ IMPORTANT:
This change basically reverts what was introduced (among other changes, of course) in [this commit](https://github.com/opf/openproject/commit/a066f83350483da738170528a124e639e8dec71e#diff-d589a9a84f284fb0615c75cd63c3754f16a148e23bb4d7fc51ea148f90448748R2) by @b12f .

I'm not sure, if the requirements are simply clashing with what was true in January. So, please check extra carefully. ;)

Just, FYI: if the datepicker is open, if would of course clash with a very narrow window height:
image

Take notice of the extra scrollbars (h and v), but those are on edge, too.

### DIFF
--- a/frontend/src/app/shared/components/time_entries/shared/modal/base.modal.html
+++ b/frontend/src/app/shared/components/time_entries/shared/modal/base.modal.html
@@ -9,7 +9,7 @@
 
   <div class="spot-divider"></div>
 
-  <div class="spot-modal--body spot-modal--body_no-scroll spot-container">
+  <div class="spot-modal--body spot-container">
     <te-form
       #editForm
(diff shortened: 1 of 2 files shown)


HDinger (approved):
I cannot think of a reason to keep the "no-scroll" behavior, so for me it is fine to merge. Maybe @b12f remembers why this was added?

Kharonus:
keeping it a bit open for @b12f to answer ;)

PR #13387, opened 2023-08-07
[49581] Translate names from common files when seeding

cbliard:
https://community.openproject.org/wp/49581

So that the role names, the color names, and the document categories are correctly translated when another language than English is selected.

### DIFF
--- a/app/seeders/source/seed_data_loader.rb
+++ b/app/seeders/source/seed_data_loader.rb
@@ -52,18 +52,24 @@ class Source::SeedDataLoader
   end
 
   def seed_data
-    @seed_data ||= Source::SeedData.new(translate(raw_seed_files_content))
+    @seed_data ||= Source::SeedData.new(translated_seed_files_content)
   end
 
-  def raw_seed_files_content
-    seed_files.reduce({}) do |raw_content, seed_file|
-      raw_content.deep_merge(seed_file.raw_content)
-    end
+  def translated_seed_files_content
+    seed_files
+      .map { |seed_file| translate_seed_file(seed_file) }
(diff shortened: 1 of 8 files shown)


aaron-contreras (commented):
Ah! Very nice 🖌️

Just a question to confirm my understanding of the fix:

So only `standard.yml` and `bim.yml` were being picked up for translation since those were the arguments to `seed_name` `Source::SeedDataLoader` but not `common.yml`, however now when using `seed_file.name` and translating each file individually we're able to search for the `common` key under the translation files as well?

cbliard:
> So only `standard.yml` and `bim.yml` were being picked up for translation since those were the arguments to `seed_name` `Source::SeedDataLoader` but not `common.yml`, however now when using `seed_file.name` and translating each file individually we're able to search for the `common` key under the translation files as well?

Not exactly. Before the fix, both `standard.yml` and `common.yml` files [were picked](https://github.com/opf/openproject/blob/0a7a93b2b27e56ee04e22389151f138b2cfa4900/app/seeders/source/seed_data_loader.rb#L67), and they were then merged together in a single hash, and then translated using `seeds.#{seed_name}` as i18n prefix. This works for strings coming from `standard.yml`, but not for the ones coming for `common.yml` as the prefix should be `seeds.common`, but the information about the originating file is lost as the hashes have been merged together

The fix is to do things in the right order: get `standard.yml` and `common.yml`, translate them using the appropriate i18n prefix, and then merge them.

I'll have a look on why the tests are failing.

cbliard:
Tests are failing because of https://community.openproject.org/wp/49583. The last commit fixes it.

aaron-contreras (approved):

PR #13401, opened 2023-08-08
[OP#43836] Fix the remaining hours calculation 

mereghost:
##### Relevant Work Package: [OP#43836](https://community.openproject.org/projects/openproject/work_packages/43836/github?query_id=491)

The issue is derived from the fact that we bubble up the remaining hours and during the summing process we add all the child `remaining_hours`.

The appropriate solution for this is to give it the same treatment that we have for `estimated_hours`, which also unlocks the possibility to add remaining hours to parent work package.

Previous:
Screenshot from 2023-08-10 10-15-19

Current:
Screenshot from 2023-08-10 10-27-44

### DIFF
--- a/db/migrate/20230808140921_add_derived_remaining_hours_to_work_packages.rb
+++ b/db/migrate/20230808140921_add_derived_remaining_hours_to_work_packages.rb
@@ -1,0 +1,69 @@
+# frozen_string_literal: true
+
+#--copyright
+# OpenProject is an open source project management software.
+# Copyright (C) 2012-2023 the OpenProject GmbH
+#
+# This program is free software; you can redistribute it and/or
+# modify it under the terms of the GNU General Public License version 3.
+#
+# OpenProject is a fork of ChiliProject, which is a fork of Redmine. The copyright follows:
+# Copyright (C) 2006-2013 Jean-Philippe Lang
+# Copyright (C) 2010-2013 the ChiliProject Team
+#
+# This program is free software; you can redistribute it and/or
+# modify it under the terms of the GNU General Public License
+# as published by the Free Software Foundation; either version 2
+# of the License, or (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program; if not, write to the Free Software
+# Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+#
+# See COPYRIGHT and LICENSE files for more details.
+#++
+
+class AddDerivedRemainingHoursToWorkPackages < ActiveRecord::Migration[7.0]
+  def change
+    add_column :work_packages, :derived_remaining_hours, :float
+    add_column :work_package_journals, :derived_remaining_hours, :float
+
+    reversible do |change|
+      change.up do
+        WorkPackage.transaction do
+          migrate_to_derived_remaining_hours!
+        end
+      end
+
+      change.down do
+        WorkPackage.transaction do
+          rollback_from_derived_remaining_hours!
+        end
+      end
+    end
+  end
+
+  def work_packages_to_update
+    leaf_node_ids = WorkPackage.leaves.pluck(:id)
+    WorkPackage.where.not(id: leaf_node_ids)
+  end
+
+  def migrate_to_derived_remaining_hours!
+    scope = work_packages_to_update
+              .where.not(remaining_hours: nil)
+    scope.update_all("derived_remaining_hours = remaining_hours, remaining_hours = NULL")
+  end
+
+  def journal_user = SystemUser.first
+
+  def rollback_from_derived_remaining_hours!
+    scope = work_packages_to_update
+              .where.not(derived_remaining_hours: nil)
+    scope.update_all("remaining_hours = derived_remaining_hours")
+  end
+end
--- a/db/migrate/20230808140921_add_derived_remaining_hours_to_work_packages.rb
+++ b/db/migrate/20230808140921_add_derived_remaining_hours_to_work_packages.rb
@@ -35,22 +35,17 @@ class AddDerivedRemainingHoursToWorkPackages < ActiveRecord::Migration[7.0]
 
     reversible do |change|
       change.up do
-        WorkPackage.transaction do
-          migrate_to_derived_remaining_hours!
-        end
+        migrate_to_derived_remaining_hours!
       end
 
       change.down do
-        WorkPackage.transaction do
-          rollback_from_derived_remaining_hours!
-        end
+        rollback_from_derived_remaining_hours!
       end
     end
   end
 
   def work_packages_to_update
-    leaf_node_ids = WorkPackage.leaves.pluck(:id)
-    WorkPackage.where.not(id: leaf_node_ids)
+    WorkPackage.where.not(id: WorkPackage.leaves)
   end
 
   def migrate_to_derived_remaining_hours!
--- a/modules/backlogs/lib/open_project/backlogs/patches/update_ancestors_service_patch.rb
+++ b/modules/backlogs/lib/open_project/backlogs/patches/update_ancestors_service_patch.rb
@@ -39,12 +39,13 @@ module OpenProject::Backlogs::Patches::UpdateAncestorsServicePatch
     def inherit_attributes(ancestor, loader, attributes)
       super
 
-      inherit_remaining_hours(ancestor, loader) if inherit?(attributes, :remaining_hours)
+      derive_remaining_hours(ancestor, loader) if inherit?(attributes, :remaining_hours)
     end
 
-    def inherit_remaining_hours(ancestor, loader)
-      ancestor.remaining_hours = all_remaining_hours(loader.leaves_of(ancestor)).sum.to_f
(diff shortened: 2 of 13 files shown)


db/migrate/20230808140921_add_derived_remaining_hours_to_work_packages.rb
```diff
@@ -0,0 +1,69 @@
...
+        WorkPackage.transaction do
+          rollback_from_derived_remaining_hours!
+        end
+      end
+    end
+  end
+
+  def work_packages_to_update
+    leaf_node_ids = WorkPackage.leaves.pluck(:id)
+    WorkPackage.where.not(id: leaf_node_ids)
```
cbliard: :yellow_circle: doing `WorkPackage.where.not(id: WorkPackage.leaves)` would be faster and consume less memory.

db/migrate/20230808140921_add_derived_remaining_hours_to_work_packages.rb
```diff
@@ -0,0 +1,69 @@
...
+  end
+
+  def migrate_to_derived_remaining_hours!
+    scope = work_packages_to_update
+              .where.not(remaining_hours: nil)
+    scope.update_all("derived_remaining_hours = remaining_hours, remaining_hours = NULL")
+  end
+
+  def journal_user = SystemUser.first
```
cbliard, line 57: :green_circle: the new "endless" method definition syntax. I'm not yet used to seeing it. Thanks for using it.

db/migrate/20230808140921_add_derived_remaining_hours_to_work_packages.rb
```diff
@@ -0,0 +1,69 @@
...
+
+  def rollback_from_derived_remaining_hours!
+    scope = work_packages_to_update
+              .where.not(derived_remaining_hours: nil)
+    scope.update_all("remaining_hours = derived_remaining_hours")
```
cbliard, line 62: :green_circle: I would have expected the rollback to fail actually, as the `derived_remaining_hours` column being read from has just been removed. But I tested it and it works. Nice!

db/migrate/20230808140921_add_derived_remaining_hours_to_work_packages.rb
```diff
@@ -0,0 +1,69 @@
...
+
+    reversible do |change|
+      change.up do
+        WorkPackage.transaction do
+          migrate_to_derived_remaining_hours!
+        end
+      end
+
+      change.down do
+        WorkPackage.transaction do
```
cbliard: :green_circle: The `transaction` block should [not be necessary](https://edgeguides.rubyonrails.org/active_record_migrations.html#migration-overview).

> On databases that support transactions with statements that change the schema, each migration is wrapped in a transaction.

Any reason for this extra transaction block?

modules/backlogs/lib/open_project/backlogs/patches/update_ancestors_service_patch.rb
```diff
@@ -39,12 +39,13 @@ module InstanceMethods
...
+      derive_remaining_hours(ancestor, loader) if inherit?(attributes, :remaining_hours)
     end
 
-    def inherit_remaining_hours(ancestor, loader)
-      ancestor.remaining_hours = all_remaining_hours(loader.leaves_of(ancestor)).sum.to_f
-      ancestor.remaining_hours = nil if ancestor.remaining_hours == 0.0
+    def derive_remaining_hours(work_package, loader)
+      descendants = loader.descendants_of(work_package)
+
+      work_package.derived_remaining_hours = not_zero(all_remaining_hours(descendants).sum.to_f)
```
cbliard, line 48: :green_circle: Sorry, not strictly related to your changes, but the method `all_remaining_hours(work_packages)` seems a bit tedious. It could be rewritten as 
```
def all_remaining_hours(work_packages)
  work_packages.sum(:remaining_hours)
end
```
It would avoid loading work packages in memory while being simpler to read.

In fact it could even be inlined:
```
work_package.derived_remaining_hours = not_zero(descendants.sum(:remaining_hours))
```
Same could be done to `derived_estimated_hours` computation.

Yeah, but in fact it would not work because `descendants` is not an ActiveRecord relation but an array with hashes and records. Just realized after reviewing more. Sorry for the burden.

modules/backlogs/lib/open_project/backlogs/patches/work_packages/update_ancestors/loader_patch.rb
```diff
@@ -32,8 +32,12 @@ def self.included(base)
...
     private
 
+    def selected_descendants_attributes
+      super + %i(remaining_hours)
+    end
+
     def selected_leaves_attributes
-      super + [:remaining_hours]
+      super + %i[remaining_hours derived_remaining_hours]
     end
```
cbliard, line 41: :yellow_circle: I understand why `:remaining_hours` must be added to the `selected_descendants_attributes` array, as it is used in the derivation computation. And I suppose you fixed a bug because previously the `:remaining_hours` attribute would not be fetched in the hash, and so the sum would be incorrect. It would be correct only with the leaf work package, as the leaf is a real work package instance and not a hash. And it was not caught by the tests because in `modules/backlogs/spec/services/work_packages/update_ancestors_service_spec.rb` the parent work package did not have any remaining hours. Thanksfully you fixed it when you added `let(:parent_remaining_hours) { 2.0 }` in the spec.

But for the `selected_leaves_attributes` array, I do not see why `:remaining_hours` and `:derived_remaining_hours` are needed: in `WorkPackages::UpdateAncestors::Loader` class, `selected_leaves_attributes` is called only when `related_of(wp, :leaves)` is called, which is called only when `leaves_of` is called. And `leaves_of` is called only in `WorkPackages::UpdateAncestors#aggregate_done_ratio`. As remaining hours are not needed for done ratio computation, I would try to remove the `selected_leaves_attributes` method from the patch, commit, and see if something breaks.

It looks like the methods `selected_descendants_attributes` and `selected_leaves_attributes` were swapped by mistake in the original implementation.

What do you think?

cbliard (approved):
Well done!

I have adopted your traffic light system. I have some remarks about things that could be improved or are maybe not necessary. Yet I see no blockers so you can merge anytime you feel you addressed my comments.

db/migrate/20230808140921_add_derived_remaining_hours_to_work_packages.rb
mereghost: Just did it without even thinking out of habit. I'll remove it. =)

db/migrate/20230808140921_add_derived_remaining_hours_to_work_packages.rb
mereghost: the blocks are actually run before any actual changes happen in the database, which is kinda neat for cases like this one.

modules/backlogs/lib/open_project/backlogs/patches/update_ancestors_service_patch.rb
mereghost: I tried to keep as close to the original as possible but you are totally right.

modules/backlogs/lib/open_project/backlogs/patches/work_packages/update_ancestors/loader_patch.rb
mereghost: Quick update: removing the `derived_remaining_hours` from the list breaks all the tests on `UpdateAncestorsService` due to differing values on the children work packages (`nil` vs `0.0`).

mereghost: @cbliard Probably. This area could seriously get some love, in order to make the code clearer and maybe less "smart".

Would you keep this blocked off or move forward and add a code maintenance task to go over the possible misnomer among other small improvements?

db/migrate/20230808140921_add_derived_remaining_hours_to_work_packages.rb
cbliard: Nice.
And it probably runs in reverse only when rolling back, or the migration would not work (it could run the block which populates the columns before the column is added).

modules/backlogs/lib/open_project/backlogs/patches/work_packages/update_ancestors/loader_patch.rb
cbliard: As I said in my review comment, I see no blockers in you PR: it works. I just have some tiny remarks and this one is a tiny remark too. I think the code is good enough like this. Not perfect but good enough. I would merge and move forward.

cbliard: I tried removing the `selected_leaves_attributes` method and I did not experience any failures when running `bundle exec rspec -fd spec/services/work_packages/update_ancestors_service_spec.rb modules/backlogs/spec/services/work_packages/update_ancestors_service_spec.rb`. Which test was failing exactly in your case?

mereghost: IIRC All of them. >.< It seems it wasn't able to match the objects as it was comparing `nil`s to `0.0`.

But now that you mention it, I didn't remove the method from here, I just kept it as it was originally.

By looking at the code I get the feeling that maybe, it was intended to eventually (that's too many layers of uncertainty :P ) it was considered to make `remaining_time` part of the `done_ratio` calculation, but we never fully committed to the idea. Probably @ulferts, @wielinde or @oliverguenther would know.

PR #13417, opened 2023-08-09
Refactor/change error rendering

cbliard:
While analyzing what needs to be changed in https://community.openproject.org/wp/40427, I saw some opportunities for improvement.


cbliard:
Closed in favor of #13442

PR #13419, opened 2023-08-09
[#48642] Add missing link to archived projects on `Type` deletion error   

aaron-contreras:
**NOTE:** The Work Package's comments suggest using a `contains` operator (`~`) for filtering the archived project. I preferred the use of an `is` operator (`=`) since there might be cases where two archived projects contain the word "Project" ("Project A", "Project B"), but only "Project A" is being flagged because of the type attempting to be deleted. Using a `contains` operator would result in both "Project A" and "Project B" being listed in the projects list the user is directed to, however "Project B" is not in that list for the right reasons; it just happens to be similarly named.

See: https://community.openproject.org/work_packages/48642

### DIFF
--- a/app/controllers/types_controller.rb
+++ b/app/controllers/types_controller.rb
@@ -178,11 +178,11 @@ class TypesController < ApplicationController
         )
       ]
 
-      archived_projects = @type.projects.filter(&:archived?)
-      if !archived_projects.empty?
-        error_message.push(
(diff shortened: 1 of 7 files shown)


oliverguenther (approved):
Fully agree on the filter choice, contains is weird for this use-case. LGTM 👍

PR #13432, opened 2023-08-10
Allow Memberships for sub-project resources

klaustopher:
This is the follow-up to #13386. After some more discussions, we have identified, that all the resources we are giving memberships to, are mostly project-scoped. For all of the checks, it will be significantly easier to leave the `project_id` in the members table and optionally add the entity.

So, when a membership is to an entire project, we leave `entity = nil`.

---

Implements https://community.openproject.org/projects/openproject/work_packages/49492/activity

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -43,8 +43,6 @@ class User < Principal
   include ::Users::Avatars
   extend DeprecatedAlias
 
-  has_many :categories, foreign_key: 'assigned_to_id',
-                        dependent: :nullify
   has_many :watches, class_name: 'Watcher',
                      dependent: :delete_all
   has_many :changesets, dependent: :nullify
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -466,7 +466,7 @@ class User < Principal
 
   # Return user's roles for project
   def roles_for_project(project)
-    project_role_cache.fetch(project:)
+    project_role_cache.fetch(project)
   end
   alias :roles :roles_for_project
 
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -466,7 +466,7 @@ class User < Principal
 
   # Return user's roles for project
   def roles_for_project(project)
-    project_role_cache.fetch(project)
+    project_role_cache.fetch(project:)
   end
   alias :roles :roles_for_project
 
--- a/app/models/users/project_role_cache.rb
+++ b/app/models/users/project_role_cache.rb
@@ -47,7 +47,7 @@ class Users::ProjectRoleCache
     # Return all roles if user is admin
     return all_givable_roles if user.admin?
 
-    ::Authorization.roles(user, project).eager_load(:role_permissions)
+    ::Authorization.roles(user, project:).eager_load(:role_permissions)
   end
 
   def cache
--- a/app/models/users/project_role_cache.rb
+++ b/app/models/users/project_role_cache.rb
@@ -33,21 +33,21 @@ class Users::ProjectRoleCache
     self.user = user
   end
 
-  def fetch(project: nil, entity: nil)
-    cache[entity || project] ||= roles(project:, entity:)
+  def fetch(cacheable)
+    cache[cacheable] ||= roles(cacheable)
   end
 
   private
 
-  def roles(project: nil, entity: nil)
+  def roles(context)
     # Return all roles if user is admin
     return all_givable_roles if user.admin?
 
     # Project is nil if checking global role
     # No roles on archived projects, unless the active state is being changed
-    return [] if project && archived?(project)
+    return [] if context.is_a?(Project) && archived?(context)
 
-    ::Authorization.roles(user, project:, entity:).eager_load(:role_permissions)
+    ::Authorization.roles(user, context).eager_load(:role_permissions)
   end
 
   def cache
--- a/app/models/users/project_role_cache.rb
+++ b/app/models/users/project_role_cache.rb
@@ -34,40 +34,24 @@ class Users::ProjectRoleCache
   end
 
   def fetch(project: nil, entity: nil)
-    if project
-      project_cache[project] ||= roles(project:)
-    elsif entity
-      entity_cache[entity] ||= roles(entity:)
-    else
-      []
-    end
+    cache[entity || project] ||= roles(project:, entity:)
   end
 
   private
 
   def roles(project: nil, entity: nil)
-    if project
-      # Project is nil if checking global role
-      # No roles on archived projects, unless the active state is being changed
-      return [] if archived?(project)
-      # Return all roles if user is admin
-      return all_givable_roles if user.admin?
+    # Return all roles if user is admin
+    return all_givable_roles if user.admin?
 
-      ::Authorization.roles(user, project:).eager_load(:role_permissions)
-    elsif entity
-      # Return all roles if user is admin
-      return all_givable_roles if user.admin?
+    # Project is nil if checking global role
+    # No roles on archived projects, unless the active state is being changed
+    return [] if project && archived?(project)
 
-      ::Authorization.roles(user, entity:).eager_load(:role_permissions)
-    end
+    ::Authorization.roles(user, project:, entity:).eager_load(:role_permissions)
   end
 
-  def project_cache
-    @project_cache ||= {}
-  end
-
-  def entity_cache
-    @entity_cache ||= {}
+  def cache
+    @cache ||= {}
   end
 
   def all_givable_roles
--- a/app/models/users/project_role_cache.rb
+++ b/app/models/users/project_role_cache.rb
@@ -33,25 +33,39 @@ class Users::ProjectRoleCache
     self.user = user
   end
 
-  def fetch(project)
-    cache[project] ||= roles(project)
+  def fetch(project: nil, entity: nil)
+    if project
+      project_cache[project] ||= roles(project:)
+    elsif entity
+      entity_cache[entity] ||= roles(entity:)
+    else
+      []
+    end
   end
 
   private
 
-  def roles(project)
-    # Project is nil if checking global role
-    # No roles on archived projects, unless the active state is being changed
-    return [] if project && archived?(project)
-
+  def roles(project: nil, entity: nil)
     # Return all roles if user is admin
     return all_givable_roles if user.admin?
 
-    ::Authorization.roles(user, project:).eager_load(:role_permissions)
+    if project
+      # Project is nil if checking global role
+      # No roles on archived projects, unless the active state is being changed
+      return [] if archived?(project)
+
+      ::Authorization.roles(user, project:).eager_load(:role_permissions)
+    elsif entity
+      ::Authorization.roles(user, entity:).eager_load(:role_permissions)
+    end
   end
 
-  def cache
-    @cache ||= {}
+  def project_cache
+    @project_cache ||= {}
+  end
+
+  def entity_cache
+    @entity_cache ||= {}
   end
 
   def all_givable_roles
--- a/app/models/users/project_role_cache.rb
+++ b/app/models/users/project_role_cache.rb
@@ -46,16 +46,18 @@ class Users::ProjectRoleCache
   private
 
   def roles(project: nil, entity: nil)
-    # Return all roles if user is admin
-    return all_givable_roles if user.admin?
-
     if project
       # Project is nil if checking global role
       # No roles on archived projects, unless the active state is being changed
       return [] if archived?(project)
+      # Return all roles if user is admin
+      return all_givable_roles if user.admin?
 
       ::Authorization.roles(user, project:).eager_load(:role_permissions)
     elsif entity
+      # Return all roles if user is admin
+      return all_givable_roles if user.admin?
+
       ::Authorization.roles(user, entity:).eager_load(:role_permissions)
     end
   end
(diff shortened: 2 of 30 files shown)


docs/development/concepts/permissions/README.md
```diff
@@ -7,27 +7,23 @@ keywords: permissions, roles, RBAC
...
 
+With RBAC, the application defines a set of roles that users and groups can be individually assigned to within the scope of a project, resource or globally. In OpenProject, the roles and permissions contained within are freely configurable. There can be an arbitrary number of roles defined.
 
 ## Key takeaways
 
 *Permissions in OpenProject...*
 
 - use the Role-based access control (RBAC) approach to allow fine-grained access to authorized resources
-- are assigned to users and groups through roles on a per-project or global level
+- are assigned to users and groups through roles on a per-project, per-resource or global level
```
aaron-contreras: Most specific to most general :)
```suggestion
- are assigned to users and groups through roles on a per-resource, per-project, or global level
```

spec/models/member_spec.rb
```diff
@@ -30,66 +30,66 @@
...
-      it 'is valid' do
-        expect(new_member)
-          .to be_valid
-      end
-    end
-
-    context 'without a project (global)' do
-      let(:project) { nil }
+  describe 'validations' do
+    it { expect(member).to validate_uniqueness_of(:user_id).scoped_to(%i[project_id entity_type entity_id]) }
```
aaron-contreras, line 53: I continue to be surprised by the matchers provided by `type: :model`

spec/models/member_spec.rb
```diff
@@ -30,66 +30,66 @@
...
+    it { expect(member).to validate_inclusion_of(:entity_type).in_array(["WorkPackage"]).allow_blank }
```
aaron-contreras, line 54: I continue to be surprised by the matchers provided by `type: :model`

klaustopher: It is https://github.com/thoughtbot/shoulda-matchers

aaron-contreras: Ah, neat!

app/services/authorization.rb
```diff
@@ -39,9 +39,11 @@ def projects(action, user)
     Authorization::ProjectQuery.query(user, action)
   end
 
-  # Returns all roles a user has in a certain project or globally
-  def roles(user, project = nil)
-    if project
+  # Returns all roles a user has in a certain project, for a specific enttiy or globally
```
aaron-contreras:
```suggestion
  # Returns all roles a user has in a certain project, for a specific entity or globally
```

app/models/user.rb
```diff
@@ -43,8 +43,6 @@ class User < Principal
   include ::Users::Avatars
   extend DeprecatedAlias
 
-  has_many :categories, foreign_key: 'assigned_to_id',
-                        dependent: :nullify
```
ulferts, line 47: 🤷  Why is this removed?

app/models/work_package.rb
```diff
@@ -67,6 +67,9 @@ class WorkPackage < ApplicationRecord
 
   has_and_belongs_to_many :github_pull_requests # rubocop:disable Rails/HasAndBelongsToMany
 
+  has_many :members, as: :entity, dependent: :destroy
+  has_many :principals, through: :members
```
ulferts: With the other associations (assignee, responsible and watcher) also referencing users/principals, I'd prefer having a more specific name for the principals associated via the members.

app/services/authorization.rb
```diff
@@ -39,9 +39,11 @@ def projects(action, user)
     Authorization::ProjectQuery.query(user, action)
   end
 
-  # Returns all roles a user has in a certain project or globally
-  def roles(user, project = nil)
-    if project
+  # Returns all roles a user has in a certain project, for a specific entity or globally
+  def roles(user, project: nil, entity: nil)
```
ulferts: An interface like `def roles(user, context)` where context can be both Project or Entity (i.e. WorkPackage) seems less error prone to me. With the current interface a caller might provide `roles(some_user, project: some_project, entity: some_work_package)` even though that does not make sense.

The method would then have to switch between the queries called base on the context provided.

app/services/authorization/user_entity_roles_query.rb
```diff
@@ -0,0 +1,41 @@
...
+  transformations.register :all, :user_restriction do |statement, user, _|
+    statement.where(users_table[:id].eq(user.id))
+  end
+
+  transformations.register users_members_join, :entity_restriction do |statement, _, entity|
+    statement
+      .and(members_table[:entity_type].eq(entity.class.to_s))
+      .and(members_table[:entity_id].eq(entity.id))
+
+    statement.and(members_table[:project_id].eq(entity.project.id)) if entity.respond_to?(:project)
```
ulferts: Is that just an extra safety net or is it actually required? It might be slightly better to call `project_id` to avoid loading the entitie's project.

app/models/users/project_role_cache.rb
```diff
@@ -33,21 +33,21 @@ def initialize(user)
     self.user = user
   end
 
-  def fetch(project)
-    cache[project] ||= roles(project)
+  def fetch(project: nil, entity: nil)
```
ulferts: Since this is effectively an XOR in that only one value is used, I'd structure the interface accordingly and have it accept only a single parameter.

spec/services/authorization/user_entity_roles_query_spec.rb
```diff
@@ -0,0 +1,66 @@
...
+require 'spec_helper'
+
+RSpec.describe Authorization::UserEntityRolesQuery do
+  let(:user) { build(:user) }
+  let(:project) { build(:project, public: false) }
+  let(:work_package) { build(:work_package, project:) }
+  let(:work_package2) { build(:work_package, project:) }
+  let(:wp_role) { build(:role) }
+  let(:non_member) { build(:non_member) }
+  let(:member) { build(:member, project:, roles: [wp_role], principal: user, entity: work_package) }
```
ulferts, line 40: Adding a second member (not based on the entity) with a different role to check that it isn't returned would be advisable.

app/services/authorization/user_entity_roles_query.rb
```diff
@@ -0,0 +1,41 @@
...
+
+class Authorization::UserEntityRolesQuery < Authorization::UserRolesQuery
+  transformations.register :all, :user_restriction do |statement, user, _|
+    statement.where(users_table[:id].eq(user.id))
+  end
+
+  transformations.register users_members_join, :entity_restriction do |statement, _, entity|
+    statement
+      .and(members_table[:entity_type].eq(entity.class.to_s))
+      .and(members_table[:entity_id].eq(entity.id))
```
ulferts, line 37: The `.and`s don't stick since they don't change the statement but return a new arel scope instead. So the new scope needs to be assigned again probably leading to something ugly like:

```
statement = statement.and(...)  
```

And that altered statement then needs to be returned from the method to actually find its way into the SQL. Currently the sql is e.g. for `Authorization::UserEntityRolesQuery.query(User.first, WorkPackage.first).to_sql`

```
SELECT "roles".* FROM "roles" WHERE "roles"."id" IN (SELECT "roles"."id" FROM "users" LEFT OUTER JOIN "members" ON "users"."id" = "members"."user_id" AND "members"."project_id" = 1 LEFT OUTER JOIN "member_roles" ON "members"."id" = "member_roles"."member_id" LEFT OUTER JOIN "roles" ON "roles"."id" = "member_roles"."role_id" WHERE "users"."id" = 4)
```

app/services/work_packages/update_service.rb
```diff
@@ -108,6 +109,12 @@ def move_time_entries(work_packages, project_id)
       .update_all(project_id:)
   end
 
+  def move_work_package_memberships(work_packages, project_id)
+    Member
+      .where(entity: work_packages)
+      .update_all(project_id:)
+  end
+
```
ulferts, line 117: :+1:

The contracts for members should also check that if there is an entity having a project_id, then that value is the value for the member's project_id as well. Can also be part of a subsequent PR, though.

spec/models/users/allowed_to_spec.rb
```diff
@@ -33,642 +33,538 @@
...
+          it do
+            # TODO: Figure this one out. In the documentation it says:
+            #
+            # Surprisingly, when looking up permissions, the non member role is always factored in, even it the user
+            # does have other roles within the project as well. That means that if the user has a role in a project
+            # not granting "Create new work package", but the non member role is granting the permission, the user
+            # will in effect have that permission.
+            #
+            # So, in my eyes, with the permission being granted to the non-member role, it should be allowed to do the
+            # requested action.
```
ulferts: I believe the documentation is wrong/outdated. I was also of the impression that the non member role would be factored in up until recently but when trying it and digging into the code I came to the conclusion that it in fact is not taken into account.

ulferts (changes requested):
Definitely moving into the correct direction @klaustopher. 

Please find my remarks mostly attached to the code. 

When looking for other side effects I also checked the memberships API (e.g. /api/v3/memberships). Not surprisingly, an entity focused membership also appears there. But I think it should be hidden unless we adapt the Membership resource's properties to also include the entity. As I don't see us doing that, I'd simply hide such memberships. 

Another place where the entity memberships turn up unwanted is in the capabilities API, e.g. `http://localhost:3000/api/v3/capabilities?pageSize=1000&filters=[{"principal":{"operator":"=","values":["%ID OF A USER WITH AN ENTITY MEMBERSHIP%"]}}]`. Here, we could extend the API without changing the structure of the resource as it is already designed for polymorphism but it is probably not required to be available. It might turn out that in fact is necessary if the frontend depends on some information but we can find that out when tackling the frontend.

app/models/user.rb
klaustopher: It is already defined in the `Principal` model and was redundant

app/services/authorization/user_entity_roles_query.rb
klaustopher: Yeah, makes sense to save that one step

app/services/authorization/user_entity_roles_query.rb
klaustopher: You are right 👍

app/models/work_package.rb
klaustopher: done

app/services/authorization.rb
klaustopher: Personally, I am a fan of explicitly putting arguments in, instead of havin a catch-all argument that you can pass basically everything into. I understand that having the context is more aligned with the rest of the code. I'll refactor and try out how it feels.

app/services/authorization/user_entity_roles_query.rb
klaustopher: done

app/services/authorization/user_entity_roles_query.rb
klaustopher: done

spec/services/authorization/user_entity_roles_query_spec.rb
klaustopher: done. Added tests for another work package and a project

spec/models/users/allowed_to_spec.rb
klaustopher: Removed the spec and removed the sentence from the docs

app/services/authorization.rb
klaustopher: I have refactored it in d543cf3738. Let me know what you think

lib/api/v3/memberships/memberships_api.rb
```diff
@@ -35,7 +35,9 @@ class MembershipsAPI < ::API::OpenProjectAPI
         resources :memberships do
           get &::API::V3::Utilities::Endpoints::Index
                  .new(model: Member,
-                      scope: -> { Member.includes(MembershipRepresenter.to_eager_load) },
+                      # TODO: For now we exclude entity specific memberships in the API until we have updated the
+                      # frontend and representers to show those properly
+                      scope: -> { Member.where(entity: nil).includes(MembershipRepresenter.to_eager_load) },
```
ulferts, line 40: This fixes the issue for the index action. The show endpoint does still return an entity membership if the right id is provided.

ulferts (changes requested):
Thanks for the changes @klaustopher. 

I still have some small remarks. There is one attached to the code but the other one is actually the same I had before where the capabilities API (e.g. `http://localhost:3000/api/v3/capabilities?pageSize=1000&filters=[{"principal":{"operator":"=","values":["15"]}}, {"context": {"operator":"=", "values":["p4"]}}]` where 15 is the id of a user and 4 the id of the project assigned to entity membership) returns the capabilities originating from entity memberships. The only possibility I currently see for preventing the leaking is doing it right in the [SQL reading the capabilities from the DB](https://github.com/opf/openproject/blob/dev/app/models/capabilities/scopes/default.rb)

klaustopher:
Both raised issues fixed for now, by excluding them. Also added issue https://community.openproject.org/work_packages/49919/activity so that we can properly re-introduce it.

ulferts (approved):
Thanks for the amends, @klaustopher. It is good to be merged.

There is still one thing that does not reflect the changes in the [capabilities default scope](https://github.com/opf/openproject/blob/dev/app/models/capabilities/scopes/default.rb#L111-L115) but that part has been wrong in the first place. It signals to the API consumers that users have at least the non member role's permission which, as established by this PR, is incorrect. I'll issue a subsequent PR to have that part removed.

PR #13433, opened 2023-08-10
fix wiki_contents with missing user as author

machisuji:
otherwise the migration will fail trying to insert wiki pages due to a foreign key constraint error like the following:

```
-- execute("UPDATE wiki_pages SET text = wiki_contents.text, author_id = wiki_contents.author_id, lock_version = wiki_contents.lock_version FROM wiki_contents WHERE wiki_contents.page_id = wiki_pages.id")
   (4.3ms)  UPDATE wiki_pages SET text = wiki_contents.text, author_id = wiki_contents.author_id, lock_version = wiki_contents.lock_version FROM wiki_contents WHERE wiki_contents.page_id = wiki_pages.id
  TRANSACTION (0.8ms)  ROLLBACK
   (1.0ms)  SELECT pg_advisory_unlock(2193071229178148760)
StandardError: An error has occurred, this and all later migrations canceled:

PG::ForeignKeyViolation: ERROR:  insert or update on table "wiki_pages" violates foreign key constraint "fk_rails_4189064f3f"
DETAIL:  Key (author_id)=(1) is not present in table "users".

from /home/ec2-user/.rbenv/versions/3.2.1/lib/ruby/gems/3.2.0/gems/activerecord-7.0.6/lib/active_record/connection_adapters/postgresql/database_statements.rb:48:in `exec'
Caused by ActiveRecord::InvalidForeignKey: PG::ForeignKeyViolation: ERROR:  insert or update on table "wiki_pages" violates foreign key constraint "fk_rails_4189064f3f"
DETAIL:  Key (author_id)=(1) is not present in table "users".

from /home/ec2-user/.rbenv/versions/3.2.1/lib/ruby/gems/3.2.0/gems/activerecord-7.0.6/lib/active_record/connection_adapters/postgresql/database_statements.rb:48:in `exec'
Caused by PG::ForeignKeyViolation: ERROR:  insert or update on table "wiki_pages" violates foreign key constraint "fk_rails_4189064f3f"
DETAIL:  Key (author_id)=(1) is not present in table "users".
```

### DIFF
--- a/db/migrate/20230322135932_merge_wiki_content_into_page.rb
+++ b/db/migrate/20230322135932_merge_wiki_content_into_page.rb
@@ -26,6 +26,8 @@
 # See COPYRIGHT and LICENSE files for more details.
 #++
 
+require_relative './migration_utils/utils'
+
 class MergeWikiContentIntoPage < ActiveRecord::Migration[7.0]
   include ::Migration::Utils
 
--- a/db/migrate/20230322135932_merge_wiki_content_into_page.rb
+++ b/db/migrate/20230322135932_merge_wiki_content_into_page.rb
@@ -27,11 +27,14 @@
 #++
 
 class MergeWikiContentIntoPage < ActiveRecord::Migration[7.0]
+  include ::Migration::Utils
+
   def change
     add_contents_columns_to_pages
 
     reversible do |dir|
       dir.up do
+        change_null_values_on_contents
         update_wiki_pages_from_contents
         update_journals_to_pages
       end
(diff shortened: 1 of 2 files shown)


db/migrate/20230322135932_merge_wiki_content_into_page.rb
```diff
@@ -27,11 +27,14 @@
 #++
```
dombesz, line 28:
```suggestion

require_relative './migration_utils/utils'

```

dombesz (changes requested):
The `::Migration::Utils` only work if we require it in the migration. I added a suggestion.
Otherwise, LGTM.👍 Feel free to merge once the require is solved.

PR #13438, opened 2023-08-10
fix: Make activity page columns the same width

b12f:
When having long texts in the changes column, it could take up a ton of space, making the projects column very narrow. This commit hardcodes the widths to be 50% each.

This commit also fixes mobile by wrapping the two columns if there isn't enough horizontal space.

Ref: https://community.openproject.org/projects/openproject/work_packages/49595/activity


aaron-contreras (approved):
Nice one! 🇫🇷

PR #13442, opened 2023-08-11
Better error shown when conflict on wiki page edition

cbliard:
https://community.openproject.org/wp/40427

Original error:

image

Improved error:

image

Additionally, base errors are now separated from field errors. This should enhance the way errors are displayed in rails-rendered views. Example:

image


app/views/common/_validation_error.html.erb
```diff
@@ -27,12 +27,7 @@ See COPYRIGHT and LICENSE files for more details.
 
 ++#%>
 
-<% error_message = t("activerecord.errors.template.header.#{(error_messages.count == 1) ? 'one' : 'other'}",
-                     count: error_messages.count,
-                     model: object_name)
-%>
-
-<% single_content_for :error_details do %>
+<% content_for :error_details, flush: true do %>
```
aaron-contreras, line 30: Didn't know about the `:flush` option. Nice 👍

aaron-contreras (approved):
Great! 🖌️ Just one question:

Do you happen to know why when executing a `dry-run` on `spec/helpers/error_message_helper_spec.rb` I get this output?

```
ErrorMessageHelper
  #error_messages_for_contract
    with two field errors
      example at ./spec/helpers/error_message_helper_spec.rb:60
      example at ./spec/helpers/error_message_helper_spec.rb:61
      example at ./spec/helpers/error_message_helper_spec.rb:59
    with one base error
      example at ./spec/helpers/error_message_helper_spec.rb:69
      example at ./spec/helpers/error_message_helper_spec.rb:71
      example at ./spec/helpers/error_message_helper_spec.rb:70
      example at ./spec/helpers/error_message_helper_spec.rb:72
    with two base errors and two field errors
      example at ./spec/helpers/error_message_helper_spec.rb:96
      example at ./spec/helpers/error_message_helper_spec.rb:94
      example at ./spec/helpers/error_message_helper_spec.rb:95
    with one field error
      example at ./spec/helpers/error_message_helper_spec.rb:48
      example at ./spec/helpers/error_message_helper_spec.rb:50
      example at ./spec/helpers/error_message_helper_spec.rb:49
    when no errors
      example at ./spec/helpers/error_message_helper_spec.rb:40
    with one base error and one field error
      example at ./spec/helpers/error_message_helper_spec.rb:81
      example at ./spec/helpers/error_message_helper_spec.rb:83
      example at ./spec/helpers/error_message_helper_spec.rb:82
  #error_messages_for
    accesses the model from instance variables if a name is given
    renders nothing if there is no instance variable with the given name
    with two field errors
      example at ./spec/helpers/error_message_helper_spec.rb:61
      example at ./spec/helpers/error_message_helper_spec.rb:59
      example at ./spec/helpers/error_message_helper_spec.rb:60
    with one base error
      example at ./spec/helpers/error_message_helper_spec.rb:71
      example at ./spec/helpers/error_message_helper_spec.rb:70
      example at ./spec/helpers/error_message_helper_spec.rb:69
      example at ./spec/helpers/error_message_helper_spec.rb:72
    when no errors
      example at ./spec/helpers/error_message_helper_spec.rb:40
    with one base error and one field error
      example at ./spec/helpers/error_message_helper_spec.rb:82
      example at ./spec/helpers/error_message_helper_spec.rb:81
      example at ./spec/helpers/error_message_helper_spec.rb:83
    with two base errors and two field errors
      example at ./spec/helpers/error_message_helper_spec.rb:96
      example at ./spec/helpers/error_message_helper_spec.rb:95
      example at ./spec/helpers/error_message_helper_spec.rb:94
    with one field error
      example at ./spec/helpers/error_message_helper_spec.rb:48
      example at ./spec/helpers/error_message_helper_spec.rb:49
      example at ./spec/helpers/error_message_helper_spec.rb:50
```

When I execute the spec run I do get the output you intended ("is expected to..."). Is it because the syntax in the examples is not `it { is_expected.to be_truthy }`? I would expect this syntax to be needed since there's a named subject defined.

aaron-contreras:
Ah! Seems unrelated. Even with the `it { is_expected.to be_truthy }` syntax it still renders as above. I guess this is due to dry-run simply basing itself off raw output without evaluating the example's contents (which is how I believe then the correct example description is generated with `subject`s).

cbliard:
> Ah! Seems unrelated. Even with the `it { is_expected.to be_truthy }` syntax it still renders as above. I guess this is due to dry-run simply basing itself off raw output without evaluating the example's contents (which is how I believe then the correct example description is generated with `subject`s).

Yes that would be my guess too. In dry-run mode, the code does not get executed so it has no way to know what the actual matcher is. This is needed to generate the test description.

PR #13447, opened 2023-08-11
[Code Maintenance] Introduce a Storage Interaction Registry

mereghost:
⚠️ THIS IS A VERY INFLUX WORK AS I EXPERIMENT WITH DIFFERENT APPROACHES ⚠️

##### Related Work Package: [[OP#49699](https://community.openproject.org/projects/openproject/work_packages/49699/activity)](https://community.openproject.org/projects/openproject/work_packages/49699/activity)

This PR is aimed at solving the extensibility of `StorageRequests` as it currently is pretty tied to `Nextcloud` and with `Sharepoint` coming soon:tm:, we will need shuffle things around.

With this in mind I decided to introduce an `Inversion of Control` container.

### How did I do it?

I'm leveraging `dry-container` as a library here to allow us to register pieces of behaviour (classes, lambdas etc) to a certain key.

I also made sure that the classes returned respond to a class level `call` method, basically making them behave like `lambda`s, which makes it easy to stub them with just a normal lambda that does what we need on tests.

### How does it looks like?

I've added, in the `Nextcloud` module the registration of all their queries and commands, and later imported them in the `Registry`.

Currently the keys follow the pattern below:

```ruby
Storages::Peripherals::Registry.register('operation_type.storage_short_name.operation_name', SomeCallableClass)
```

That being said we should standardize the names across all the storages as patterns emerge.

### What are the advantages?

From my POV there's a couple:

1. Ease of code switching during runtime: we can base all the calls on the `storage_short_name` and let the container pull in the correct object.
2. Ease of extensibility: adding a new storage is just a matter of creating the necessary commands and queries and registering them. The rest of the codebase never need to know about it.
3. Ease for 3rd parties: given we document the barebones commands and queries necessary for integrating a new file storage, we can allow 3rd parties to register their own against the registry and won't need to touch the rest of the codebase.

:book: TO BE CONTINUED


apfohl:
Good job, this looks absolutely right to me. One thing I might have tried differently is, that I would have tried to not use the registry statically. That means I would have provided a parameter to where it will be needed. But maybe I need to accept Rails and it's "I have all the things on class methods" way of doing it. That breaks IoC IMO. Because you can just call it from everywhere and that's hard to control.

apfohl (approved):

PR #13340, opened 2023-08-02
[OP#49397] Add an admin view for project members storage authorization status

akabiru:
#### https://community.openproject.org/work_packages/49397

- [x] Define base query: fetch members belonging to current project including principal and oauth client tokens
- [x] Update Breadcrumbs
- [x] Add ⚠️ Icon to error messages
- [x] Add Feature specs
- ~Investigate scoping permissions check to Project Role~
- [x] Resolve Fixmes
   - [x] FIXME: Status based on Nextcloud OAuth Client Token Presence `modules/storages/app/components/storages/projects_storages/members/row_component.rb`
   - [x] ~update no members template~ handled by `TableComponent#empty_row_message`
   - [x] Fix Link in instructions
- [x] Add pagination to member check view

Screenshot 2023-08-08 at 1 33 49 PM

Screenshot 2023-08-08 at 1 19 24 PM

PR #13461, opened 2023-08-14
[#49673] Limit members check icon to automatically managed project folders

akabiru:
#### https://community.openproject.org/work_packages/49673

Only the icon is hidden. If the user manually accesses the UI, then all members will read as "Not Connected".

This follows #13340 

Screenshot 2023-08-14 at 5 41 19 PM


ba1ash (approved):
:+1:

PR #13477, opened 2023-08-15
[#49630] Nextcloud/Administration link is stripped when host has a subpath

akabiru:
##### https://community.openproject.org/work_packages/49630

Use singular interface [`Storages::Peripherals::StorageInteraction::Nextcloud::Util.join_uri_path`](https://github.com/opf/openproject/blob/dev/modules/storages/app/common/storages/peripherals/storage_interaction/nextcloud/util.rb#L50-L56) that uses `File.join` given [`URI.join`](https://apidock.com/ruby/URI/join/class) is buggy with subpaths.

> [URI.join](https://apidock.com/ruby/URI/join/class) uses a delimiter – forward slash (/) – to decide if joined strings are a path or endpoint. In order to include strings as part of the path, they must end with a forward slash (/). Otherwise, they are assumed to be an endpoint and are overritten by your new “endpoint”.

> ```ruby
>  > [URI](https://apidock.com/ruby/URI).[join](https://apidock.com/ruby/URI/join/class)("http://localhost/", "test", "main.json")
>  => #<URI::HTTP:0x007fa68cec0ba0 URL:http://localhost/main.json> 
> 
>  > [URI](https://apidock.com/ruby/URI).[join](https://apidock.com/ruby/URI/join/class)("http://localhost/", "test/", "main.json")
>  => #<URI::HTTP:0x007fa68ce14c60 URL:http://localhost/test/main.json>
> ```

image

### DIFF
--- a/modules/storages/app/views/storages/admin/storages/automatically_managed_project_folders/edit.html.erb
+++ b/modules/storages/app/views/storages/admin/storages/automatically_managed_project_folders/edit.html.erb
@@ -80,7 +80,8 @@ See COPYRIGHT and LICENSE files for more details.
     <p class="form--field-instructions -xwide">
       <%= t(:"storages.instructions.managed_project_folders_application_password") %>
       <a
-        href="<%= URI.join(@storage.oauth_application.integration.host, 'settings/admin/openproject') %>"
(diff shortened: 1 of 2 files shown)


ba1ash (approved):
:+1:

PR #13501, opened 2023-08-17
[#49690] Bugfix PDF Export: Sums table only shows total sum

as-op:
https://community.openproject.org/work_packages/49690


aaron-contreras (approved):
👍

aaron-contreras:
Failing spec is unrelated.

PR #13505, opened 2023-08-17
[#49507] File Drag and Drop

dombesz:
https://community.openproject.org/work_packages/49507

The issue is, when the active file drag arrives over the description input (ckeditor), then it loses the dragging state and it does not allow the upload to happen. Removing the counter and relying on the `dragenter` `dragleave` events only fixes the issue.
<details>
  <summary>Demo</summary>

Aug-17-2023 14-15-09

</details>

### DIFF
--- a/frontend/src/app/shared/components/attachments/attachments.component.ts
+++ b/frontend/src/app/shared/components/attachments/attachments.component.ts
@@ -113,7 +113,13 @@ export class OpAttachmentsComponent extends UntilDestroyedMixin implements OnIni
   };
 
   private onGlobalDragEnter:(_event:DragEvent) => void = (_event) => {
-    this.dragging += 1;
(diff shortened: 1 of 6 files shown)


Kharonus:
🔴 This change basically reverts what I have implemented to the drag and drop. The reason, we could not use a bool and instead hat to use a stack counter, was described here: 
https://community.openproject.org/projects/openproject/work_packages/47557/activity#activity-4

I might find more sources, but the problem is not as simply solvable.

dombesz:
Thanks @Kharonus for letting me know about the [existing ticket and corner case](https://community.openproject.org/wp/47557/activity#activity-6), updated the fix to take that into account.

Kharonus (approved):
LGTM

PR #13532, opened 2023-08-21
[#49493] Add a `WorkPackageRole` type

aaron-contreras:
## Milestones
1. A new Role type will become available (`WorkPackageRole`) in order to capture/assign roles to users and groups appropriately for Work-Package-specific permissions in order to grant access to OpenProject on a Work-Package-specific basis.

2. Ability to specify/determine which role types permissions are assignable to. This will help simplify the form implementation for Role Creation.

3. Seed the three default Work Package roles and corresponding permissions that will come pre-packaged with OpenProject
  - Editor
  - Commenter
  - Viewer 

## Notes
Work Package: https://community.openproject.org/work_packages/49493

### DIFF
--- a/app/seeders/basic_data/base_role_seeder.rb
+++ b/app/seeders/basic_data/base_role_seeder.rb
@@ -26,14 +26,12 @@
 # See COPYRIGHT and LICENSE files for more details.
 #++
 module BasicData
-  class RoleSeeder < ModelSeeder
-    self.model_class = Role
-    self.seed_data_model_key = 'roles'
+  class BaseRoleSeeder < ModelSeeder
     self.needs = []
 
     def model_attributes(role_data)
       {
-        type: type(role_data['global']),
+        type:,
         name: role_data['name'],
         position: role_data['position'],
         permissions: role_data['permissions'].uniq,
@@ -43,8 +41,8 @@ module BasicData
 
     private
 
-    def type(global_value)
-      true?(global_value) ? 'GlobalRole' : 'Role'
+    def type
+      model_class.to_s
     end
 
     def builtin(value)
@@ -74,7 +72,9 @@ module BasicData
       when Array
         value
       when :all_assignable_permissions
-        Roles::CreateContract.new(Role.new, nil).assignable_permissions.map(&:name)
+        Roles::CreateContract.new(model_class.new, nil)
+                             .assignable_permissions
+                             .map(&:name)
       end
     end
 
--- a/app/seeders/common.yml
+++ b/app/seeders/common.yml
@@ -240,7 +240,32 @@ open_color_palette:
   - '#d9480f'
 
 work_package_roles:
-  # TODO: Fill in with work package default roles and permissions
+  - reference: :default_role_work_package_editor
+    t_name: Work Package Editor
+    position: 7
+    permissions:
+      - :view_work_packages
+      - :edit_work_packages
+      - :work_package_assigned
+      - :add_work_package_notes
+      - :edit_own_work_package_notes
+      - :manage_work_package_relations
+      - :export_work_packages
+  - reference: :default_role_work_package_commenter
+    t_name: Work Package Commenter
+    position: 8
+    permissions:
+      - :view_work_packages
+      - :work_package_assigned
+      - :add_work_package_notes
+      - :edit_own_work_package_notes
+      - :export_work_packages
+  - reference: :default_role_work_package_viewer
+    t_name: Work Package Viewer
+    position: 9
+    permissions:
+      - :view_work_packages
+      - :export_work_packages
 project_roles:
   - reference: :default_role_non_member
     t_name: Non member
--- a/app/seeders/common.yml
+++ b/app/seeders/common.yml
@@ -239,112 +239,114 @@ open_color_palette:
   - '#e8590c'
   - '#d9480f'
 
-roles:
-- reference: :default_role_non_member
-  t_name: Non member
-  position: 0
-  builtin: :non_member
-  permissions:
-  - :view_work_packages
-  - :view_calendar
-  - :comment_news
-  - :browse_repository
-  - :view_changesets
-  - :view_wiki_pages
-- reference: :default_role_anonymous
-  t_name: Anonymous
-  position: 1
-  builtin: :anonymous
-  permissions:
-  - :view_work_packages
-  - :browse_repository
-  - :view_changesets
-  - :view_wiki_pages
-- reference: :default_role_member
-  t_name: Member
-  position: 3
-  permissions:
-  - :view_work_packages
-  - :export_work_packages
-  - :add_work_packages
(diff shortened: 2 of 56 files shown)


app/seeders/basic_data/base_role_seeder.rb
```diff
@@ -74,7 +72,9 @@ def permissions(value)
       when Array
         value
       when :all_assignable_permissions
-        Roles::CreateContract.new(Role.new, nil).assignable_permissions.map(&:name)
+        Roles::CreateContract.new(self.class.model_class.new, nil)
+                             .assignable_permissions
+                             .map(&:name)
```
aaron-contreras: I don't know if this is a bug or intended but this would always assign all `assignable_permissions` for Project Roles even if a `GlobalRole` is being seeded. @cbliard

spec/seeders/basic_data/project_role_seeder_spec.rb
```diff
@@ -103,42 +105,15 @@
...
-    end
-  end
-
   context 'with permissions: :all_assignable_permissions' do
     let(:data_hash) do
       YAML.load <<~SEEDING_DATA_YAML
         roles:
-        - reference: :role_project_admin
-          name: Project admin
-          global: true
```
aaron-contreras, line 139: @cbliard Following up on the `assignable_permissions` comment. This was intended to be a global role, but based on the spec it was expected to be assigned all `assignable_permissions` for the `Role` type.

Now that I've split these specs into `project_role_seeder_spec` and `global_role_seeder_spec` I've added an expectation on the `global_role_seeder_spec` that the assigned roles to the created role, only gets the `assignable_permissions` for `GlobalRole` types.

app/seeders/common.yml
```diff
@@ -240,111 +240,115 @@ open_color_palette:
...
+        - :view_changesets
+        - :view_team_planner
+    - reference: :default_role_project_admin
+      t_name: Project admin
+      position: 5
+      permissions: :all_assignable_permissions
+  global:
+    - reference: :default_role_project_creator_and_staff_manager
+      t_name: Staff and projects manager
+      global: true
```
cbliard: `global: true` is not useful anymore.

spec/seeders/basic_data/project_role_seeder_spec.rb
cbliard: I think I did a mistake originally here. As it is the "Project admin" role, it should not have been `global: true`.
Thanks for having fixed it!

spec/support/seeders.rb
```diff
@@ -48,7 +54,7 @@ def add_needed_seeders_for(seeder_class, needed_acc)
   shared_let(:needed_seeders_keys) do
     # warning: there is one limitation: RoleSeeder will only pick 'roles' key,
     # and not 'modules_permissions' key.
-    needed_seeders.map { _1.try(:seed_data_model_key) }.compact
```
cbliard: I do not like introducing exceptions. It can then evolve to accidental complexity. The idea was to keep a 1-1 relation between the key name in the seed files and the class name of the record. This way I do not have to think too much to lookup the right key given a model name.

Instead of having `roles.project`, `roles.global`, and `roles.work_package` , why not use `project_roles`, `global_roles`, and `work_package_roles` directly? This way the role seeders would work as any other seeder.

What do you think about it?

spec/seeders/basic_data/global_role_seeder_spec.rb
```diff
@@ -0,0 +1,87 @@
...
+              global: true
+              position: 1
+              permissions: :all_assignable_permissions
+      SEEDING_DATA_YAML
+    end
+
+    it 'gives all assignable permissions to the role' do
+      expect(GlobalRole.find_by(name: 'Project admin').permissions)
+        .to match_array(Roles::CreateContract.new(GlobalRole.new, nil).assignable_permissions.map { _1.name.to_sym })
+    end
```
cbliard, line 83: I think that's ok. When I run it in a console, it returns `[:add_project, :create_backup, :create_user, :manage_user, :manage_placeholder_user]`

Maybe the test can be enhanced to check that the permissions include `:add_project` and `:create_user`, and do not include `view_project` and `view_work_packages`. Just to be sure that they are global permissions and not project permissions.

(As a side note, I'm always suspicious when we assert equality using business code in the comparison: in this case, what if a bug is introduced in `Roles::CreateContract.new(GlobalRole.new, nil).assignable_permissions`? I suppose we have to trust that a unit test in `Roles::CreateContract` exists for when a `GlobalRole` is given. In that case, hardening the test would be useless. And after checking, it looks like `spec/contracts/roles/shared_contract_examples.rb` is covering it, so no need to add the "test enhancements" I suggested earlier: that would only duplicate code)

app/seeders/basic_data/base_role_seeder.rb
```diff
@@ -43,8 +41,8 @@ def model_attributes(role_data)
 
     private
 
-    def type(global_value)
-      true?(global_value) ? 'GlobalRole' : 'Role'
+    def type
+      self.class.model_class.to_s
```
cbliard: `self.class` is not needed. `model_class.to_s` is enough.

app/seeders/basic_data/base_role_seeder.rb
cbliard: That's not what I am observing. I tried it and it returned the expected global permissions.

We can have a look together. Ping me when you're available.

app/seeders/basic_data/base_role_seeder.rb
```diff
@@ -74,7 +72,9 @@ def permissions(value)
       when Array
         value
       when :all_assignable_permissions
-        Roles::CreateContract.new(Role.new, nil).assignable_permissions.map(&:name)
+        Roles::CreateContract.new(self.class.model_class.new, nil)
```
cbliard: `self.class` not needed here.

spec/seeders/basic_data/global_role_seeder_spec.rb
aaron-contreras: Discussed in a call

klaustopher (commented):
Looked through everything. Very thorough!

lib/open_project/access_control/permission.rb
```diff
@@ -42,6 +42,7 @@ def initialize(name,
                      hash,
                      public: false,
                      require: nil,
+                     permissible_on: [],
```
klaustopher: I'd opt to remove the default value here and make it mandatory to add this to a  permission definiton

aaron-contreras: Ah, yes! I was thinking about adding a warning somewhere in case someone forgets about defining it but the error message raised by removing the default is much better. :)

klaustopher (approved):
Awesome work!

PR #13534, opened 2023-08-21
Address "implicit block expectation syntax" deprecation warning

aaron-contreras:
Fixes

```
The implicit block expectation syntax is deprecated, you should pass
a block rather than an argument to `expect` to use the provided block
expectation matcher or the matcher must implement
`supports_value_expectations?`.

e.g  `expect { value }.to change `User#local_avatar_attachment``
not `expect(value).to change `User#local_avatar_attachment``
```

and

```
The implicit block expectation syntax is deprecated, you should
pass a block rather than an argument to `expect` to use the provided
block expectation matcher or the matcher must implement
`supports_value_expectations?`.

e.g  `expect { value }.to raise Exception`
not `expect(value).to raise Exception`
```

### DIFF
--- a/modules/avatars/spec/models/user_spec.rb
+++ b/modules/avatars/spec/models/user_spec.rb
@@ -28,8 +28,8 @@ RSpec.describe User do
     context "when the uploaded file is a good image" do
       subject { lambda { user.local_avatar_attachment = avatar_file } }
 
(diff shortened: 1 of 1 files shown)


akabiru (approved):
Good looking out 🏆

PR #13535, opened 2023-08-21
Address deprecation warning for Firefox Driver's `:capabilities` parameter

aaron-contreras:
Fixes:

```
WARN Selenium [:capabilities] [DEPRECATION] The :capabilities parameter
for Selenium::WebDriver::Firefox::Driver is deprecated. Use :options
argument with an instance of Selenium::WebDriver::Firefox::Driver
instead.
```


oliverguenther (approved):
🙇‍♂️

PR #13543, opened 2023-08-22
Creates the OneDriveStorage model

mereghost:
##### Implements [OP#49759](https://community.openproject.org/projects/openproject/work_packages/49759/relations)

The most bare bones model creation.


Kharonus (approved):
If this really suffices to add new storages of type onedrive, I find this nice! ;)

PR #13545, opened 2023-08-22
Improve wording of journal entry of migration step adding existing file links

wielinde:
https://community.openproject.org/work_packages/49770

With the release of 13.0 we had to create a journal entry for all those file links that were already there. Unfortunately that journal entry is not understandable. The information value is very low. The complexity is high. It would be best to not show it. However, the cost for hiding it appear very high in development, maintenance and run time. And it is not likely that we will run into a similar issue. Also, I don't believe that so many installations already use the Nextcloud integration. That is why I opt for simply improving the texts within the activity entry and live with it.

Before:
image

After:

image

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -1528,8 +1528,8 @@ en:
           non_working: "%{date} is now non-working"
       system_update:
         file_links_journal: >
-          From now on you will see links to files that are stored in external file storages mentioned here in the
-          Activity tab. The following additions of links actually already existed before this system migration:
+          From now on, activity related to file links (files stored in external storages) will appear here in the
+          Activity tab. The following represent activity concerning links that already existed:
 
 
   links:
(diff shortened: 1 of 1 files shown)


psatyal:
Hi @wielinde,

I understand. If we've got to have a message, then yes, the "after" one is much clearer. I'd recommend a few changes in the interest of clarity:

> **OpenProject system upgrade:** From now on, activity related to file links (files stored in external storages) will appear here in the Activity tab. The following represent activity concerning links that already existed:
> 
> - **File link** to {filename} ({storage-name}) added
> - **File link** to {filename} ({storage-name}) added

Also, if we can use a slightly different format for the actual journal entries in this particular case, we could even do:

> **OpenProject system upgrade:** From now on, activity related to file links (files stored in external storages) will appear here in the Activity tab. The following represent activity concerning links that already existed:
> 
> - {filename} ({storage-name})
> - {filename} ({storage-name})

PR #13555, opened 2023-08-23
[49771] Set statement timeout to 15 minutes for the validity period migration

cbliard:
https://community.openproject.org/wp/49771

Introduce a `set_minimum_statement_timeout` helper to be used in a migration. It sets the statement timeout to a minimum value for the migration. It will do nothing if the statement timeout is already set to a higher value.

### DIFF
--- a/lib_static/open_project/migration_statement_timeout/manager.rb
+++ b/lib_static/open_project/migration_statement_timeout/manager.rb
@@ -1,0 +1,83 @@
+# frozen_string_literal: true
+
+#-- copyright
+# OpenProject is an open source project management software.
+# Copyright (C) 2012-2023 the OpenProject GmbH
+#
+# This program is free software; you can redistribute it and/or
+# modify it under the terms of the GNU General Public License version 3.
+#
+# OpenProject is a fork of ChiliProject, which is a fork of Redmine. The copyright follows:
+# Copyright (C) 2006-2013 Jean-Philippe Lang
+# Copyright (C) 2010-2013 the ChiliProject Team
+#
+# This program is free software; you can redistribute it and/or
+# modify it under the terms of the GNU General Public License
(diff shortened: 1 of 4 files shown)


cbliard:
I'll rebase it on the correct branch `release/13.0`.

lib_static/open_project/migration_statement_timeout/manager.rb
```diff
@@ -0,0 +1,83 @@
...
+#
+# See COPYRIGHT and LICENSE files for more details.
+#++
+
+module MigrationStatementTimeout
+  module Manager
+    def exec_migration(conn, direction)
+      min_timeout = self.class.minimum_statement_timeout
+      return super unless min_timeout
+      return super unless direction == :up
```
aaron-contreras, line 36: Is this because rollbacks are fine using the default and don't usually exceed it?

lib_static/open_project/migration_statement_timeout/migration_extensions.rb
```diff
@@ -0,0 +1,53 @@
...
+    # interpreted as milliseconds.
+    #
+    # When the given value is a string with units, it is interpreted
+    # accordingly. Valid units for this parameter are "ms", "s", "min", and "h".
+    # Examples: "15min", "90s", "2h".
+    #
+    # @param [Integer|String] timeout duration
+    def set_minimum_statement_timeout(timeout)
+      self.minimum_statement_timeout = timeout
+    end
```
aaron-contreras, line 51: Very nice 🖌️

lib_static/open_project/migration_statement_timeout/manager.rb
cbliard: No, it's just that it's about the migration, and it always goes up. I took inspiration from some other gems doing similar things. They also bother only in one direction.

PR #13557, opened 2023-08-23
Fix/safari versions support test

as-op:


as-op:
Closed in favor of https://github.com/opf/openproject/pull/13561

PR #13560, opened 2023-08-23
 [#49749] Identity providers sign in icons overlap with the text or not enough spacing between them

mrmir:
https://community.openproject.org/wp/49749

Old:

image

New:

image


aaron-contreras (approved):
👍

PR #13565, opened 2023-08-24
[#49782] Fix out-of-bounds widgets error when creating Version Boards

aaron-contreras:
## Bug description
An ["out-of-bounds" validation](https://github.com/opf/openproject/blob/21a696ef9b170e14ad2daf53364a4c2113822c2f/modules/grids/app/contracts/grids/base_contract.rb#L139-L144) was rightfully raising an error since when a Project had more than 4 versions which applied for being a column on a version board, the amount of columns being set on the board was 4 (as done in the original implementation of Board creation).

Version boards however, are a special case, as the number of columns to be created initially with it is varying based on the amount of open versions for the project at the moment of creating the board.

## Fix
The `VersionBoardCreateService` now sets the `column_count_for_board` dynamically based on the versions that a widget will be created for; this ensures they're mapped 1-to-1 (or the default if no versions apply) and we don't run into out-of-bounds widgets for them anymore.

Added a modification to the setup code in the `VersionBoardCreateService` spec file demonstrates the failing behavior and success behavior after the override.
## Notes
See Bug Work Package: https://community.openproject.org/work_packages/49782

### DIFF
--- a/modules/boards/app/services/boards/version_board_create_service.rb
+++ b/modules/boards/app/services/boards/version_board_create_service.rb
@@ -15,7 +15,7 @@ module Boards
     private
 
     def column_count_for_board
-      versions(params).count
+      [super, versions(params).count].max
     end
 
     def create_queries(params)
--- a/modules/boards/app/services/boards/version_board_create_service.rb
+++ b/modules/boards/app/services/boards/version_board_create_service.rb
@@ -14,6 +14,10 @@ module Boards
 
     private
 
+    def column_count_for_board
+      versions(params).count
+    end
+
     def create_queries(params)
       versions(params).map do |version|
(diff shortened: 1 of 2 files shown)


cbliard (approved):

cbliard (changes requested):
There is a failing test, and it highlights an issue when the project has no versions:

image

failing test is `modules/boards/spec/features/boards_sorting_spec.rb:49`

aaron-contreras:
Thanks for catching that bug @cbliard !

cbliard:
> Thanks for catching that bug @cbliard !

Let's be honest, I did not catch anything: it's an existing test which detected the issue. The person to thank is @bsatarnejad. Thank you!

cbliard (approved):

PR #13568, opened 2023-08-25
Correctly create default values for multi-value list custom field

cbliard:
## Bug description

Two bugs

https://community.openproject.org/wp/49765

Creating a new list custom field and then adding a comment to an existing work package would be impossible: an error "<custom field> was attempted to be written but is not writable" was displayed and the comment was not saved.

https://community.openproject.org/wp/49784

In a project with a multi-value list custom field with multiple default values, the work package new form would not preselect the default values. Instead it would preselect a "-" value.

## Analysis

In `ModelContract`, it checks that readonly attributes are not changed. If there are some `unauthenticated_changes`, then some readonly attributes have been changed. To see what has been changed by the user, for custom fields, it checks that the given value is different from the custom field default value, in which case it has not been changed by the user. This check is done in `CustomValue#default?`.

Custom values are populated by `Redmine::Acts::Customizable#custom_field_values` using the `#available_custom_fields`. This ensure there is at least one `CustomValue` for each `CustomField` of the work package.

This check is flawed for list custom field:
* for non-multi-value list, `CustomField#default_value` returns an integer, and `CustomValue#value` returns a string. Checking something like `367 == "367"` always returns false. So the field is considered changed by the user.
* for multi-value list, `CustomField#default_value` returns an array of integers, and `CustomValue#value` returns a stringified array of integers. Checking something like `[367, 368] == "[367, 368]"` always returns false as well. So the field is considered changed by the user.

Also the behavior is wrong for multi-value list: when creating value, it should create as many `CustomValue` instances as there are default options for the multi-value list custom field. For instance if the default options are `367` and `368`, then two `CustomValue`s must be created: one with value 367, second with value 368. This is why the values were not preselected in the new work package form (bug #49784). The value returned by the work package form api was actually `"[367, 368] not found"`, which was then displayed by the frontend as the value `"-"`.

## Fix

* Return strings for default value of list custom fields in `CustomField#default_value` and check inclusiveness instead of equality for multi-value custom field in `CustomValue#default?` (bug #49765).
* Create multiple CustomValue for multi-value list custom fields in `Redmine::Acts::Customizable#custom_field_values` (bug #49784).

### DIFF
--- a/app/models/custom_field.rb
+++ b/app/models/custom_field.rb
@@ -80,7 +80,7 @@ class CustomField < ApplicationRecord
 
   def default_value
     if list?
-      ids = custom_options.select(&:default_value).map(&:id)
+      ids = custom_options.where(default_value: true).pluck(:id).map(&:to_s)
 
       if multi_value?
         ids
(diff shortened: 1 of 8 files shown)


klaustopher (approved):
LGTM. Good job!

PR #13570, opened 2023-08-25
[#49760] Creating and editing of OneDrive storages

apfohl:
- https://community.openproject.org/wp/49760

### What is part of this PR?

- added a new provider type option to storage creation dropdown
- added additional forms for the flow of creating onedrive storages
  - refactored the existing forms and controllers to be less specific for nextcloud and more agnostic for different storage types
- amended tests to work with different storage provider types

### General assumptions and information

- only one provider type is needed to serve onedrive and sharepoint, as both share the same API
- creating an oauth client for onedrive access need some guidance, as the app has to be created in azure portal
- adding a project storage will not (yet) show onedrive storages in the "available storages" dropdown

### What is NOT included?

- unit test for new contracts, will be delivered in a next PR


app/services/oauth_clients/set_attributes_service.rb
```diff
@@ -28,4 +28,19 @@
 
 # See also: create_service.rb for comments
 class OAuthClients::SetAttributesService < BaseServices::SetAttributes
+  private
+
+  def set_attributes(params)
+    super(replace_client_secret_with_nil(params))
+  end
+
+  def replace_client_secret_with_nil(params)
```
ba1ash, line 37: Could you please clarify for me why we have to replace client secret with nil here? And can we remove it from params without clonning?

modules/storages/app/models/storages/common/configuration_checks.rb
```diff
@@ -42,9 +42,12 @@ def configured?
     end
 
     def configuration_checks
-      { host_name_configured: (host.present? && name.present?),
-        openproject_oauth_application_configured: oauth_application.present?,
-        storage_oauth_client_configured: oauth_client.present? }
+      { storage_oauth_client_configured: oauth_client.present? }.tap do |configuration_checks|
+        if provider_type_nextcloud?
```
ba1ash, line 46: Imho, it would be more simple to not have this module/concern and put the logic inside `storage.rb` base class and define/redefine `configuration_checks` in the descendants. Wdyt?

modules/storages/app/models/storages/storage.rb
```diff
@@ -61,11 +61,12 @@ class Storages::Storage < ApplicationRecord
   has_one :oauth_application, class_name: '::Doorkeeper::Application', as: :integration, dependent: :destroy
 
   PROVIDER_TYPES = [
-    PROVIDER_TYPE_NEXTCLOUD = 'Storages::NextcloudStorage'.freeze
+    PROVIDER_TYPE_NEXTCLOUD = 'Storages::NextcloudStorage'.freeze,
+    PROVIDER_TYPE_ONE_DRIVE = 'Storages::OneDriveStorage'.freeze
   ].freeze
 
   # Uniqueness - no two storages should  have the same host.
```
ba1ash: Actually, do we need this comment? I would say the meaning of the following line is obvious enough.

modules/storages/app/services/storages/storages/set_attributes_service.rb
```diff
@@ -39,14 +39,24 @@ def set_default_attributes(_params)
     private
 
     def set_attributes(params)
-      super(params)
+      super(replace_empty_host_with_nil(params))
```
ba1ash, line 42: Why do we have to replace it? And do we really have to clone? Can we modify params, by removing the key, for instance?

modules/storages/app/views/storages/admin/storages/edit.html.erb
```diff
@@ -1,76 +1,91 @@
...
-                data: { confirm: t(:'storages.confirm_replace_oauth_client') },
-                class: 'button -with-icon icon-reload') %>
-  <% else %>
-    <%= link_to(t("js.label_create"), new_admin_settings_storage_oauth_client_path(@object), class: 'button -with-icon icon-add') %>
-  <% end %>
+<% if @storage.provider_type_nextcloud? %>
+  <%= render partial: 'storages/admin/storages/nextcloud/oauth_client_section' %>
+<% elsif @storage.provider_type_one_drive? %>
+  <%= render partial: 'storages/admin/storages/one_drive/oauth_client_section' %>
 <% end %>
```
ba1ash: What about deriving the partial name from the provider type?
```suggestion
 <%= render partial: "storages/admin/storages/#{@storage.short_provider_type}/oauth_client_section" %>
```

modules/storages/app/views/storages/admin/storages/new_oauth_client.html.erb
```diff
@@ -7,40 +35,10 @@
...
-
-  <% if @storage.oauth_client %>
-    <%= styled_button_tag t(:button_replace), class: "-highlight -with-icon icon-checkmark" %>
+  <% if @storage.provider_type_nextcloud? %>
+    <%= render partial: 'storages/admin/storages/nextcloud/oauth_client', locals: { f: f } %>
   <% else %>
-    <% label_action = @storage.automatic_management_unspecified? ? 'continue' : 'complete' %>
-    <%= styled_button_tag t("storages.buttons.save_and_#{label_action}_setup"), class: "-highlight -with-icon icon-checkmark" %>
+    <%= render partial: 'storages/admin/storages/one_drive/oauth_client', locals: { f: f } %>
   <% end %>
```
ba1ash: Let's derive it here as well, maybe?

modules/storages/lib/api/v3/storages/storages_api.rb
```diff
@@ -26,39 +26,17 @@
 # See COPYRIGHT and LICENSE files for more details.
 #++
 
-# This class provides definitions for API routes and endpoints for the storages namespace. It inherits the
-# functionality from the Grape REST API framework. It is mounted in lib/api/v3/root.rb.
-# `modules/storages/lib/` is a defined root directory for grape, providing a root level look up for the namespaces.
-# Hence, the modules of the class have to be represented in the directory structure.
-
-# OpenProjectAPI is a simple subclass of Grape::API that handles patches.
```
ba1ash, line 34: :+1: Thanks for removing all these comments!

modules/storages/app/models/storages/common/configuration_checks.rb
Kharonus: This is a fine question for @akabiru ;)

app/services/oauth_clients/set_attributes_service.rb
Kharonus: Two answers:
- we replace the value of `""` with `nil`, because otherwise we get `""` in the database as a value, instead of `null`.
- we clone the params, because we want to maintain the immutability of method params.

app/services/oauth_clients/set_attributes_service.rb
```diff
@@ -28,4 +28,19 @@
...
+    cloned_param = params.clone
+
+    if cloned_param[:client_secret] == ''
```
mereghost, line 40: 🟢 This could be changed to a `cloned_params[:client_secret].blank?`

Also what's the behaviour if a key is missing? If it is no big deal you could call `.compact_blank` over it so that all blank pairs are removed.

modules/storages/app/services/storages/storages/set_attributes_service.rb
Kharonus: Same answer as above ;)

modules/storages/app/views/storages/admin/storages/edit.html.erb
Kharonus: good hint!

modules/storages/app/common/storages/errors.rb
```diff
@@ -31,5 +31,9 @@
 module Storages
   module Errors
```
mereghost, line 32: 🟢 One thing that came to me while looking at this part is that we could have one `Error` (or `BaseError` as RoR loves `BaseSomething` names) and make our other errors inherit from it.

Then we can just rescue our own errors by going with `rescue Storages::Errors::BaseError` and let other more esoteric errors bubble up. WDYT?

modules/storages/app/common/storages/peripherals/registry.rb
```diff
@@ -37,8 +37,14 @@ class Resolver < Dry::Container::Resolver
...
-          _, storage, operation = key.split('.')
-          raise ::Storages::Errors::OperationNotSupported, "Operation #{operation} not support by provider: #{storage}"
+          case key.split('.')
+          in ['contracts', storage]
+            raise ::Storages::Errors::MissingContract, "No contract defined for provider: #{storage}"
+          in [_, storage, operation]
+            raise ::Storages::Errors::OperationNotSupported, "Operation #{operation} not support by provider: #{storage}"
+          else
+            raise ::Storages::Errors::ResolverStandardError, "Cannot resolve key #{key}."
+          end
```
mereghost, line 47: <3

modules/storages/app/models/storages/storage.rb
```diff
@@ -89,7 +90,7 @@ class Storages::Storage < ApplicationRecord
   def self.shorten_provider_type(provider_type)
     case /Storages::(?'provider_name'.*)Storage/.match(provider_type)
     in provider_name:
-      provider_name.downcase
+      provider_name.underscore
     else
       raise ArgumentError,
             "Unknown provider_type! Given: #{provider_type}. " \
```
mereghost, line 95: 🟡 I have this gut feeling that this method could be removed or at least made into a good old `if` as we only have 2 cases.

modules/storages/app/models/storages/storage.rb
```diff
@@ -104,4 +105,8 @@ def short_provider_type
   def provider_type_nextcloud?
     provider_type == ::Storages::Storage::PROVIDER_TYPE_NEXTCLOUD
   end
+
+  def provider_type_one_drive?
+    provider_type == ::Storages::Storage::PROVIDER_TYPE_ONE_DRIVE
+  end
 end
```
mereghost, line 111: 🟡 WDYT of making this into a query method that takes in a parameter? Also this is basically a type check. Can't we substitute its uses by duck typing?

```ruby
def provider_type?(storage_klass)
  self.class == storage_klass
end
```

modules/storages/app/views/storages/admin/storages/edit.html.erb
mereghost: 🟡 Good point, @ba1ash.

mereghost (approved):
Overall looks good. 

I left some comments but I don't see any of it as a blocker for a merge.

modules/storages/app/views/storages/admin/storages/new_oauth_client.html.erb
Kharonus: done

modules/storages/app/views/storages/admin/storages/edit.html.erb
Kharonus: done

modules/storages/app/models/storages/storage.rb
Kharonus: done

app/services/oauth_clients/set_attributes_service.rb
Kharonus: well, the replacement taking place here is very specific. we replace the empty string with `nil`. The empty string is the standard value, when the input field is left empty. we do not want to replace `[]`, `{}`, or such things, those should still result in errors.

modules/storages/app/common/storages/peripherals/registry.rb
Kharonus: icecream? 🍦

modules/storages/app/models/storages/storage.rb
Kharonus: we use it in a view, where we not yet have storage instances, but instead call in on the class. I'd keep it.

modules/storages/app/models/storages/common/configuration_checks.rb
ba1ash: Not exactly, before it was a set of methods that were extracted into a module, but the behavior is the same for the whole hierarchy. So, it is fine. But with this PR we are introducing the check on the type of the class inside the module which is included inside a base class. So, it smells like `configuration_checks` method is now a responsibility of descendants.

modules/storages/app/models/storages/storage.rb
Kharonus: What is the advantage of this method over simply using `==`? Both have to use this long fully qualified name. The advantage of having the explicit check is less characters in the calling places:

```erb
<% if @storage.provider_type_nextcloud? %>
<% end %>

<% if @storage == ::Storages::Storage::PROVIDER_TYPE_NEXTCLOUD %>
<% end %>
```

so, this method is just for convenience for callers.

mereghost: Yes, but this can blow up once we have multiple provider types. There must be a way to not have to add this on a per provider basis. Right now we need to add it to a list of constants and add a method. =/

modules/storages/app/common/storages/errors.rb
Kharonus: I'd like to have a clear error structure, alas we still have a bit of a mess there: 
- `modules/storages/app/models/storages/storage_error.rb`: this are storage errors used to be returned into service results
- `modules/storages/app/common/storages/errors.rb`: those errors are used within the peripherals

so, I'd like to rework all errors, into a clear structure and knowledge of how to use them. Just introducing a `BaseError` right now is something, I'd not do right now. WDYT?

modules/storages/app/models/storages/common/configuration_checks.rb
Kharonus: I don't see it as a violation of responsibility.

the concern is scoping everything it needs to give you an answer about the validity of your configuration. independent of the storage provider type. putting those check into the storage class descendants would again bloat them, while having them within the concern makes it easier to grasp all differences in configuration checking within all provider types.

modules/storages/app/models/storages/storage.rb
Kharonus: right, but what is the alternative? we can remove them and use `==` everywhere. No big reason to not do it. The provider types are in the end those longer variables. These bool methods are in the end not needed, they are just sugar.

Can you think of a convenient form, that is not using the original variable names? what about dynamic method generation: we iterate over the provider types array and generate a method `provider_type_{short_name}?` for every type.

modules/storages/app/common/storages/errors.rb
mereghost: Sounds like a plan to me. =)

modules/storages/app/models/storages/storage.rb
mereghost: That's not a hill I'm willing to die on. :rofl: I'll see if I can come up with an idea later on. :P

modules/storages/app/models/storages/common/configuration_checks.rb
ba1ash: But it is already dependent, because the behavior differs for each subclass inside `configurations_check`. This piece belongs there in the first place, because of the type context. And the module presence in this case just decrease the code readability, imho.

Kharonus: As discussed with the devs, we will destruct the concern and integrate the check to the specific descendants.

The storage.rb spec will be dismantled into a shared_example, to be executed in `NextcloudStorage` as well as in `OneDriveStorage`. the test cases from the concern will be written explicitly for both provider types.

There will be a next PR including those changes, link will be added by edith.

ba1ash (approved):
LGTM :+1:

PR #13571, opened 2023-08-25
[#49750] SAML icon too small when compared to other icons

mrmir:
https://community.openproject.org/work_packages/49750

Before:

image

After:

image


as-op (approved):
👍

PR #13575, opened 2023-08-28
Make chinese zh-CN locale up to date

cbliard:
https://community.openproject.org/wp/49795

Previously, downloading translations files from crowdin would write both zh-CN and zh-TW to the same file, overwriting the previous one. This commit adds a mapping to the crowdin.yml file to avoid this.

Last update of the Chinese Simplified translations was on [Jan 3 2022](https://github.com/opf/openproject/commit/2a1de3ebff) (12.1 release).

As a side effect, the script `fix_crowdin_translation_filenames.rb` is now superfluous.

Admin interface before:

image

Admin interface after:

image


klaustopher:
LGTM. Cannot tell if removing the file breaks anything else, so do not feel comfortable to approve here.

klaustopher (approved):
Approved after talked about changes with @cbliard

PR #13582, opened 2023-08-29
[48355] Save Team Planner view as 2 weeks view not only 1 week view.

bsatarnejad:
https://community.openproject.org/projects/openproject/work_packages/48355/activity


HDinger (changes requested):
`displayRepresentation` is the attribute used to define whether a Work package query is in list view or card view. So you should not use it for the week display of calendars. We already have a queryAttribute for this (`cview`).
With your current changes, the `displayRepresentation` is always set to the value of `cview` param. However, this does not exist in a WP page, making it impossible to save a WP card view.

bsatarnejad:
Close this PR since this can be a feature.

PR #13589, opened 2023-08-30
[49767] Number of wp no longer shown in bars on the graph

bsatarnejad:
In new version of chart.js we need to add DataLabelsPlugin plugin to show the labels.

https://community.openproject.org/projects/openproject/work_packages/49767/activity


ulferts (approved):
👍

PR #13591, opened 2023-08-30
use case insensitive login matching for oauth remapping

machisuji:
WP [#49834](https://community.openproject.org/projects/openproject/work_packages/49834/activity)


aaron-contreras (approved):
👍

Nice to know we override `find_by_login` to handle this.

Went ahead and fixed a Rubocop warning that was having the CI fail. As of around 13:00 my time, the build step seems to fail on all open pull requests. It'll probably self-correct by tomorrow so please re-run it and merge. I'll leave it approved as it's fine locally :)

PR #13595, opened 2023-08-30
[49831] Description in a box having too little height when the browser window's width is decreased 

bsatarnejad:
The layout of WP pages is changed from one-column to two-columns when the screen width is greater than 1279px (xl), so flex-basis of 530px for tabs should be applied only when the screen is larger than xl size. 

https://community.openproject.org/work_packages/49831/activity


ulferts (approved):
Thanks for the fix

PR #13596, opened 2023-08-30
Use `VCR` for stubbing outbound HTTP requests

akabiru:
### https://community.openproject.org/projects/openproject/work_packages/49797

The current static web mocks can become stale without any alarms if there are breaking changes with external integration APIs or if there's a misrepresentation of the request mocks. Instead, it is more accurate and deterministic to record real HTTP interactions that are refreshed periodically and use those in our tests.

🚧  This is a pilot to solicit feedback and then cascade the update in other storage peripherals.

#### Proposed Solution:

Hook in [VCR](https://github.com/vcr/vcr) into WebMock, record real HTTP interactions and replay them during tests runs for fast, determistic, accurate tests.

#### Out of Scope

Automatic re recording of cassettes. See [Docs](https://benoittgt.github.io/vcr/#/cassettes/automatic_re_recording)

#### Links

* https://benoittgt.github.io/vcr/#/
* https://github.com/vcr/vcr


docker/dev/tls/docker-compose.core-override.example.yml
```diff
@@ -13,6 +13,19 @@ services:
       # It must be amended accordingly to OS.
       - /etc/ssl/certs/ca-certificates.crt:/etc/ssl/certs/ca-certificates.crt:ro
 
+  backend-test:
+    # Connect the backend-test container to the same network as the backend for nextcloud HTTP interactions
+    networks:
+      - external
+    extra_hosts:
+      - "nextcloud.local:host-gateway"
```
Kharonus: usually this should be nedded, as long as the `backend-test` service is in the external network. With the proposed setup, the `traefik` service should have the alias `nextcloud.local`.

modules/storages/spec/factories/storage_factory.rb
Kharonus, line 1: those changes might conflict with what I just merged to `dev` :/

docker/dev/tls/docker-compose.core-override.example.yml
akabiru: Thanks! Will update 👍🏾

modules/storages/spec/factories/storage_factory.rb
akabiru: No worries, let me run a second rebase 👍🏾

Kharonus (approved):
From my pov this looks good.

Are there any technical implication, we need to consider? 

One remark I still have, up for discussion:
Should there one vcr cassette config per test? Or should we actually configure some "server states", that are used with multiple different requests and document their behaviour inside the yml file?

mereghost (approved):
Nice work! :)

You may want to rebase, as I did some changes to the factories that got merged yesterday.

akabiru:
> From my pov this looks good.
> 
> Are there any technical implication, we need to consider?
> 

Thanks @Kharonus, yes- a few come to mind:

1. We _still_ need to agree on a cassette refresh period- right now it's not enforced in the specs
2. In order to generate new cassettes, one needs to set up the `backend-test` -> `nextcloud` connection on docker; the credentials are manually copied over right now and set in the `.env.test.local`. _We should be able to automate this with a simple script_

> One remark I still have, up for discussion: Should there one vcr cassette config per test? Or should we actually configure some "server states", that are used with multiple different requests and document their behaviour inside the yml file?

Good question, depends on our team decision & "system under test" but it's not uncommon to reuse cassettes across multiple tests E.g. a Login cassette

PR #13606, opened 2023-08-31
docs[#49218] Remove storage configuration checks feature that didn't make it into 13.0.1

akabiru:
https://community.openproject.org/work_packages/49218 did not make it into 13.0.1 as the workpackage version had not been updated.


as-op (approved):
👍

PR #13607, opened 2023-08-31
[49848] Fix some untranslated strings

cbliard:
https://community.openproject.org/wp/49848


ulferts (approved):
👍 

Thanks for the cleanups as well

PR #13608, opened 2023-08-31
 [#47871]  [AppSignal] Performance MessagesController#show

mrmir:
https://community.openproject.org/work_packages/47871


mrmir:
This eliminates some of the CACHE Project and Attachment calls, but the EnabledModule calls are still there for the most part and I can't figure out how to get rid of them.

However, the duration for rendering messages/show is noticeably lower now, and so is the allocations count.

ulferts (approved):
👍 The response time drops to ~ two thirds of the former value on the page I tested with so quite good.

PR #13611, opened 2023-09-01
[#49872] Extend CSP frame-ancestors with active nextcloud storage hosts

ba1ash:
https://community.openproject.org/work_packages/49872

**Why?**
We need to allow, at least, work_package creation form to be opened inside an iframe within Nextcloud.

**How?**

The PR extends the hook which modifies CSP `frame-ancestors` field with enabled Nextcloud storages hosts for every request.

https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/frame-ancestors


modules/storages/spec/lib/append_storages_hosts_to_csp_hook_spec.rb
```diff
@@ -52,12 +52,12 @@ def trigger_application_controller_before_action_hook
     hook_listener.application_controller_before_action(controller:)
   end
 
-  shared_examples 'appends storage host to CSP' do
-    it 'adds CSP directive to allow connection to storage host' do
+  shared_examples 'modifies CSP' do
```
akabiru: 🟢 I tend to prefer the previous description as it was more specific _"appends storage host to CSP"_

Or we could name it like the method

```suggestion
  shared_examples 'content security policy directives' do
```

```ruby
include_examples 'content security policy directives'
```

akabiru (approved):
Top work 🏆  nice refresher on CSP for me 😄

PR #13617, opened 2023-09-04
Fix/reenable specifying eager loading in endpoints

ulferts:
Eager loading for the base scopes specified in the API endpoints was disabled by a [change done to the default index action](https://github.com/opf/openproject/commit/f83d4f0b47a971dffa10fa79253ceca013f6823a). Because of that, the performance of some endpoints, e.g. /api/v3/principals degraded.

https://community.openproject.org/wp/49915


aaron-contreras (approved):
👍

PR #13619, opened 2023-09-04
[chore] removed configuration check concern

Kharonus:
- remove inclusion of concern from storage base class
- moved provider type related configuration checks into descendant models
- refactored unit tests
- changed scope :configured to iterate over subclasses and concatenate results

### What is the problem?

- the concern for configuration checks is not really satisfying the separation of concerns principle
- instead, every model should contain the logic of their own configurational completeness

### What happened so far?

- removed concern
- added logic to comply to the inheritance schema of the storage types for
  - `configured` scope
  - instance method `configured?`
- refactored unit tests for models to no longer instantiate a base storage
  - this triggers A LOT of small changes around the specs
  - review can be shortened, as it is very repetitive


modules/storages/app/models/storages/storage.rb
```diff
@@ -82,6 +80,10 @@ class Storages::Storage < ApplicationRecord
     end
   end
 
+  scope :configured, -> do
+    subclasses.map { |clazz| clazz.configured.to_a }.flatten
```
Kharonus: this is very much to be discussed. Any better idea I'm fine to try. :D

IMHO, we need a scope on the base class, that can be executed e.g. in the API or the project_storages controller, like `Storages::Storage.configured`. To make that work, I currently execute the scopes of all subclasses and concatenate the results. This is working, but it might appear a bit smelly.

WDYT?

ba1ash: I'd suggest to not implement the scope if we do not need it explicitly somewhere. Maybe, let's use `raise NotImplementedError` here as well.

klaustopher: Just FYI, Ruby has [`flat_map`](https://ruby-doc.org/3.2.2/Enumerable.html#method-i-first) that does map & flatten in one step

Kharonus: I'm confused? Like I said, we DO need the scope.

In the locations, where we use it, you do not want to fetch first all provider types, to iterate over them and call the scope on each. you do want to say: "Give me all configured storages, please!"

Kharonus: discussion cancellation: I removed the scope and replaced the usage with the instance method `configured?`

Kudos to @mereghost for brainstorming.

modules/storages/app/controllers/storages/admin/project_storages_controller.rb
```diff
@@ -169,6 +171,6 @@ def available_storages
     Storages::Storage
       .visible
       .not_enabled_for_project(@project)
-      .configured
+      .select(&:configured?)
```
akabiru, line 174: 🟡 Switching to a ruby instance method is less performant and could become an issue if the number of storages are many (this is unlikely), as they are loaded into memory. Most importantly we also loose out on (lazy) chain able scopes- which is handy for forming SQL queries.

I see I missed a discussion in https://github.com/opf/openproject/pull/13619#discussion_r1315762693 with @mereghost 😬

mereghost: I do agree with you @akabiru, but as you noted we the performance hit shouldn't be that terrible here.

If push comes to shove we could store the configured status in the DB to allow for easy queries, rather than trying to make crazy unions or cross joins.

Kharonus: That is the point, we can reintroduce it, but it probably would require a db migration. This solution right now is the "simple" one. As the PR was not about the scope in the very beginning, but it stumbled across it to make one drive storages work at all, any improvements to the scope should come to a follow up PR.

If you already have a great idea, @akabiru, I'd be very intrigued to see it :D

akabiru: Makes sense, let's cross that bridge when we get there then 😄

akabiru (approved):
Nice one, looks good to me! 👍🏾 :shipit:

mereghost (approved):
Let's :shipit:

akabiru (approved):
Haha @Kharonus I had already reviewed & approved 😄 https://github.com/opf/openproject/pull/13619#pullrequestreview-1610963757

ba1ash (approved):
Thanks @Kharonus. Well done! :muscle:

PR #13626, opened 2023-09-05
Follow up for Share Work Package permission

klaustopher:
Follow Up to https://community.openproject.org/work_packages/49494

I have made a mistake in the definition of the ticket and used the wrong identifier for the permission. This renames it and adds a translation for it

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -2490,6 +2490,7 @@ en:
   permission_save_queries: "Save views"
   permission_select_custom_fields: "Select custom fields"
   permission_select_project_modules: "Select project modules"
+  permission_share_work_packages: "Share work packages"
   permission_manage_types: "Select types"
   permission_view_changesets: "View repository revisions in OpenProject"
   permission_view_commit_author_statistics: "View commit author statistics"
--- a/db/migrate/20230905090205_add_share_workpackages_permission.rb
+++ b/db/migrate/20230905090205_add_share_workpackages_permission.rb
@@ -1,0 +1,46 @@
+# frozen_string_literal: true
+
+#-- copyright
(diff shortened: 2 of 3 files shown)


db/migrate/20230905090205_add_share_workpackages_permission.rb
```diff
@@ -0,0 +1,47 @@
...
+# See COPYRIGHT and LICENSE files for more details.
+#++
+
+require_relative 'migration_utils/permission_adder'
+
+class AddShareWorkpackagesPermission < ActiveRecord::Migration[7.0]
+  # Decouple ActiveRecord model from Migration
+  class MigrationRolePermission < ApplicationRecord
+    self.table_name = 'role_permissions'
+  end
```
aaron-contreras, line 37: ❤️

config/locales/en.yml
```diff
@@ -2490,6 +2490,7 @@ en:
   permission_save_queries: "Save views"
   permission_select_custom_fields: "Select custom fields"
   permission_select_project_modules: "Select project modules"
+  permission_share_work_packages: "Share work packages"
```
aaron-contreras, line 2493: Thank you, I missed this in the previous PR even though I ran into a bug with it before. It would be nice to have a spec file in core and each module that runs through all defined permissions and ensures that there is a translation present for them. I'll bring up a WorkPackage for it and assign it to myself

aaron-contreras (approved):
🦭 of approval

config/locales/en.yml
klaustopher: Should be easy enough. We can iterate through all permissions with `OpenProject::AccessControl.permissions.map(&:name)` and then check that we have a translation for each one

PR #13629, opened 2023-09-05
Add rubocop-performance

klaustopher:
As I recently saw a case of `map().flatten` and suggested the use of `flat_map`, I saw that we do not yet use the `rubocop-performance`. It suggests some nice not-too-well-known functions, that can combine multiple steps and also some general things that just can be done quicker.

Let me know what you think :)

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -377,7 +377,7 @@ class ApplicationController < ActionController::Base
                                 .order('id ASC')
     fail ActiveRecord::RecordNotFound if @work_packages.empty?
 
-    @projects = @work_packages.map(&:project).compact.uniq
+    @projects = @work_packages.filter_map(&:project).uniq
     @project = @projects.first if @projects.size == 1
   rescue ActiveRecord::RecordNotFound
     render_404
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -441,7 +441,7 @@ class ApplicationController < ActionController::Base
 
   # Returns a string that can be used as filename value in Content-Disposition header
   def filename_for_content_disposition(name)
-    request.env['HTTP_USER_AGENT'] =~ %r{(MSIE|Trident)} ? ERB::Util.url_encode(name) : name
+    %r{(MSIE|Trident)}.match?(request.env['HTTP_USER_AGENT']) ? ERB::Util.url_encode(name) : name
   end
 
   def api_request?
--- a/app/models/principals/scopes/possible_assignee.rb
+++ b/app/models/principals/scopes/possible_assignee.rb
@@ -49,7 +49,7 @@ module Principals::Scopes
               .assignable
               .of(project)
               .group('user_id')
-              .having(["COUNT(DISTINCT(project_id, user_id)) = ?", Array(project).count])
+              .having(["COUNT(DISTINCT(project_id, user_id)) = ?", Array(project).size])
               .select('user_id')
         )
       end
--- a/app/models/queries/filters/strategies/boolean_list.rb
+++ b/app/models/queries/filters/strategies/boolean_list.rb
@@ -37,7 +37,7 @@ module Queries::Filters::Strategies
     end
 
     def valid_values!
-      filter.values &= allowed_values.map(&:last).map(&:to_s)
+      filter.values &= allowed_values.map { |v| v.last.to_s }
     end
 
     private
@@ -51,7 +51,7 @@ module Queries::Filters::Strategies
     end
 
     def too_many_values
-      values.count { |element| !element.blank? } > 1
+      values.count(&:present?) > 1
     end
   end
 end
--- a/app/models/queries/filters/strategies/boolean_list.rb
+++ b/app/models/queries/filters/strategies/boolean_list.rb
@@ -51,7 +51,7 @@ module Queries::Filters::Strategies
     end
 
     def too_many_values
-      values.reject(&:blank?).length > 1
+      values.count { |element| !element.blank? } > 1
     end
   end
 end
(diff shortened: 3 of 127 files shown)


klaustopher:
Not 100% done with all fixes

oliverguenther (approved):
Great idea, thanks for addressing the issues in one go 👍

app/controllers/application_controller.rb
```diff
@@ -441,7 +441,7 @@ def accept_key_auth_actions
 
   # Returns a string that can be used as filename value in Content-Disposition header
   def filename_for_content_disposition(name)
-    request.env['HTTP_USER_AGENT'] =~ %r{(MSIE|Trident)} ? ERB::Util.url_encode(name) : name
```
klaustopher, line 444: Just for context, `match` and `=~` fill [some pseudo global variables](https://docs.ruby-lang.org/en/3.2/Regexp.html#class-Regexp-label-Regexp+Global+Variables) (one of the things that shows Matz's love for Perl). Even though they use the `$` prefix they are thread- and method local. Also, the match object is created.

Using `match?` only checks that the string matches the regex without generating a match object and not filling the pseudo globals

aaron-contreras:
Love it! ❤️

app/services/custom_fields/create_service.rb
```diff
@@ -29,7 +29,7 @@
 module CustomFields
   class CreateService < ::BaseServices::Create
     def self.careful_new_custom_field(type)
-      if type.to_s =~ /.+CustomField\z/
+      if /.+CustomField\z/.match?(type.to_s)
```
cbliard, line 32: Wouldn't `type.to_s.end_with?('CustomField')` be better here?

Replying to myself: probably not, because `CustomField` does not match in the original case, but matches in my suggested change. So the behavior would be different.

You can ignore my comment :)

lib/open_project/access_control.rb
```diff
@@ -134,7 +134,7 @@ def modules_permissions(modules)
       end
 
       def module_enterprise_feature?(name)
-        current_module = modules.select { |m| m[:name] == name }[0]
+        current_module = modules.find { |m| m[:name] == name }
```
cbliard, line 137: :pray: Fixing all these ones is a blessing!

app/models/queries/filters/strategies/boolean_list.rb
```diff
@@ -51,7 +51,7 @@ def operator_map
     end
 
     def too_many_values
-      values.reject(&:blank?).length > 1
+      values.count { |element| !element.blank? } > 1
```
cbliard: Could be `count(&:present?) > 1` instead.

app/models/queries/filters/strategies/relation.rb
```diff
@@ -54,7 +54,7 @@ def valid_values!
     private
 
     def too_many_values
-      values.reject(&:blank?).length > 1
+      values.count { |element| !element.blank? } > 1
```
cbliard: Could be `count(&:present?) > 1` instead.

modules/backlogs/lib/open_project/backlogs/hooks/layout_hook.rb
```diff
@@ -86,7 +86,7 @@ def controller_work_package_new_after_save(context = {})
         end
 
         if params[:copy_tasks]
-          params[:copy_tasks] += ':' if !/:/.match?(params[:copy_tasks])
+          params[:copy_tasks] += ':' if !params[:copy_tasks].include?(':')
```
cbliard: And then the Rails/NegateInclude cop will suggest that `!foo.include?('bar')` could be replaced by `foo.exclude?('bar')` :)

modules/reporting/spec/models/cost_query/operator_spec.rb
```diff
@@ -100,15 +100,15 @@ def create_project(options = {})
...
+      expect(cost_query('projects', 'name', '~', 'o').size).to eq(Project.all.count { |p| p.name.include?('o') })
+      expect(cost_query('projects', 'name', '~', 'test').size).to eq(Project.all.count { |p| p.name.include?('test') })
+      expect(cost_query('projects', 'name', '~', 'child').size).to eq(Project.all.count { |p| p.name.include?('child') })
     end
 
     it "does !~ (not contains)" do
-      expect(cost_query('projects', 'name', '!~', 'o').size).to eq(Project.all.count { |p| !(p.name =~ /o/) })
-      expect(cost_query('projects', 'name', '!~', 'test').size).to eq(Project.all.count { |p| !(p.name =~ /test/) })
-      expect(cost_query('projects', 'name', '!~', 'child').size).to eq(Project.all.count { |p| !(p.name =~ /child/) })
+      expect(cost_query('projects', 'name', '!~', 'o').size).to eq(Project.all.count { |p| !(p.name.include?('o')) })
```
cbliard: Same with Rails/NegateInclude: Use `.exclude?` and remove the negation part.

app/models/principals/scopes/possible_assignee.rb
```diff
@@ -49,7 +49,7 @@ def possible_assignee(project)
               .assignable
               .of(project)
               .group('user_id')
-              .having(["COUNT(DISTINCT(project_id, user_id)) = ?", Array(project).count])
+              .having(["COUNT(DISTINCT(project_id, user_id)) = ?", Array(project).size])
```
cbliard, line 52: I didn't know that `count` is slower than `size` or `length`.
https://github.com/fastruby/fast-ruby#arraylength-vs-arraysize-vs-arraycount-code

cbliard (approved):
Love it! Thanks for doing the extra work of fixing current code offenses.

I have some very minor suggestions.

klaustopher:
I just ran the auto-fixer did not look to detailed into what it produced 🙈. Will apply your suggestions

PR #13632, opened 2023-09-05
[#49935] Fix flaky create meeting spec

aaron-contreras:
## Flakiness description
`./modules/meeting/spec/features/meetings_new_spec.rb:167` has been failing quite frequently on CI runs lately. 

After some inspection to it and attempting to reproduce it locally, using the `stress` tool, I found that it was caused by the participant section (which is fetched via a turbo stream) not loaded in time before the form submission occurred while under heavy load.

By using `wait_for_network_idle` we can ensure that the participant section is loaded and therefore, its params also bundled in the form submission when clicking create.

## Notes
See: https://community.openproject.org/work_packages/49935


mrmir (approved):
Nice! :+1:

PR #13641, opened 2023-09-06
Remove debug statement in the middle of the case statement of RDM Mailhandler

klaustopher:
There is a debug statement in the middle of the case statement, that makes the code invalid (at least in Ruby 3.2). It was introduced in 2ae53bd which was fixing typos. Looks like it landed there by accident.

File seems not to be in use, but it is causing problems when running rubocop on the entire project


cbliard (approved):
Not sure if the file is still relevant or is dead code, but it was definitely invalid ruby code. 
Thanks for fixing it. If it's dead code it can still be deleted in another PR.

PR #13646, opened 2023-09-06
fix font-size of ng-value in wp type status row

ulferts:
Before:

image

After:

image

This fix could have been applied to both the generic `_autocomplete.sass` or the specific `_type_status_row.sass` and I am not sure about where to place it best. It seems prudent to have the font-size of inputs governed generally throughout the system as it is done in `_forms.sass` for all the normal input fields. On the other hand, d26b76f63a5 removed the font-size statements explicitly so I am unsure.

https://community.openproject.org/wp/49858

### DIFF
--- a/frontend/src/global_styles/content/_autocomplete.sass
+++ b/frontend/src/global_styles/content/_autocomplete.sass
@@ -61,6 +61,7 @@ div.autocomplete
 // -------------------------- ng-select --------------------------
 .ng-select
   width: 100%
+  font-size: var(--body-font-size)
 
   .ng-value-container
     min-height: 2rem
@@ -92,7 +93,6 @@ div.autocomplete
   .ng-value
     @include text-shortener
     line-height: 22px
-    font-size: var(--body-font-size)
 
 .ng-select.ng-select-multiple .ng-select-container
   height: initial !important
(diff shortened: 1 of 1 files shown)


ulferts:
@bsatarnejad / @HDinger I'd welcome your input on this one.

bsatarnejad:
> @bsatarnejad / @HDinger I'd welcome your input on this one.

Hi @ulferts , 
The font-size of ng-select is inherited from its parent, which is 14px (body font size). In this specific case (type-status-select-component which I hadn’t checked), the parent has a font-size of 24px. So this font-size raised this bug by overriding the font-size of ng-select. In order to avoid issues like this, I think it would be better to revert my previous changes in frontend/src/global_styles/content/_autocomplete.sass and add the font-size of var(--body-font-size) to ng-select styles like the way you changed it for now.

wielinde:
@ulferts @bsatarnejad Am just making this up or did the status and type selection also change the color of those fields? Right now, that seems lost.

bsatarnejad:
> @ulferts @bsatarnejad Am just making this up or did the status and type selection also change the color of those fields? Right now, that seems lost.

Hi @wielinde No, @ulferts and I just checked the release12.5, and it doesn't change the color, but it can be a feature.

bsatarnejad (approved):

PR #13647, opened 2023-09-06
[49460] Enable Primer's high-contrast mode

bsatarnejad:
https://community.openproject.org/projects/openproject/work_packages/49460/activity


HDinger (commented):
Hi @bsatarnejad 

really great work 👍 I only have some minor remarks:

- [x] Please remove the option for all the "dark" modes from the selector in the personal settings
- [x] In the menus, there is a jump when I hover due to the border being shown. Could you please check whether we can avoid that easily? If not, I can live with it as well.
- [x] The green create button in the sidebar, as well as in the top bar have too little contrast due to the dark backgrounds. I think we have to add a border for them.
- [x] Although it does not make that much difference, I would set the colors for the grey default buttons in the `primer_color_mapping` file as well (similar to how it is done for the alternative buttons).
- [x] I think we should add a border around the status button of a WorkPackage as well.
- [x] I have the feeling that inside user generated content (e.g. the description) the borders are too bright (e.g. tables, checkboxes, dividers etc)
- [x] We still have a lot of custom border-colors in the code which are not changed when changing to the high-contrast mode (I was searching for "border-color: #")
- [x] The accessibility test showed that at many places the font is too bright and the contrast too low. This is the case when we use `color: var(--gray-dark)`. Could you please check whether we can replace that with a Primer color to increase the contrast?

One major thing that we have to discuss is how to deal when there is a custom theme applied. The current solution can only deal partially with that. I am in favour of completely overriding that but we have to talk with product for that.

- [x] When having the high-contrast mode selected, all custom theme variables are ignored. That means that the header and the sidebar switch to the light colors which we also have in the OP light theme.

app/views/custom_styles/_primer_color_mapping.erb
```diff
@@ -5,15 +5,38 @@
...
-    <% end %>
-    <% if design_color.variable == "content-link-color" %>
-        --color-accent-fg: <%= design_color.hexcode %> !important;
-    <% end %>
-  <% end %>
+  }
+  [data-light-theme=light_high_contrast]{
+    --button--alt-highlight-background-hover-color: var(--color-btn-primary-hover-bg);
+    --button--alt-highlight-background-color: var(--color-btn-primary-bg);
+    --button--alt-highlight-border-color: var(--main-menu-font-color);
```
HDinger: Why did you choose the `main-menu-font-color` for the border of the button? That seems wrong, esecially since you overwrite that variable a few lines later.

app/views/custom_styles/_primer_color_mapping.erb
```diff
@@ -5,15 +5,38 @@
...
+    --main-menu-bg-color: var(--AppHeader-bg);
+    --main-menu-hover-font-color: var(--color-fg-default);
+    --main-menu-bg-selected-background: var(--color-action-list-item-default-hover-bg);
+    --main-menu-bg-hover-background: var(--color-action-list-item-default-hover-bg);
+    --main-menu-font-color: var(--color-fg-default);
+    --main-menu-selected-font-color: var(--color-fg-default);
+    --main-menu-border-color: var(--color-border-muted);
+    --main-menu-resizer-color: var(--color-border-muted);
+    --main-menu-bg-selected-border: var(--color-border-muted);
+    --header-border-bottom-width: 1px;
```
HDinger: Please take the time and sort these so the belonging variables are shown below each other (so that all button styles are in a block, all main menu styles, all header styles etc.). Currently it is really hard to gasp all variables when trying to understand the effects on one part of the application.

HDinger (changes requested):
Hi @bsatarnejad 

I think we are nearly done, I just have some little remarks:

- [x] When opening a menu, the items change their color on hover
Bildschirmfoto 2023-10-06 um 13 27 39

- [x] I think we need to override all variables that are customizable via the admin/design page. Otherwise I can set the `primary_color` to any value and the `-highlight` buttons are shown in that color.
Bildschirmfoto 2023-10-06 um 13 35 58

- [x] I thought about the logo again, and I think we have to touch it. At least in the case, when the logo has not been changed (so still the OpenProject logo is shown). For these cases we could switch to the dark version of it (as we have it available). We can discuss next week how that could be done.

- [x] Please rename "theme" to "mode" on the settings page, to reduce conflicts with the custom (instance) theme

oliverguenther:
@lindenthal @aaron-contreras you need to activate a feature flag to see the high contrast theme. You can do that through the UI using the hidden flag page https://pr-13647-49460-enable-prime-ip-18-192-58-179.my.pullpreview.com/admin/settings/experimental

I've gone ahead and enabled it for the admin user

aaron-contreras:
@oliverguenther Ah, thanks! I ended up trying it out with a local clone

frontend/src/app/spot/styles/sass/components/list.sass
```diff
@@ -29,7 +29,6 @@
         @include spot-focus
 
       &:hover
-        color: var(--content-link-color)
```
HDinger: We do want this effect in the "normal" mode. Only in the high contrast mode, this color should not change depending on the custom theme.

app/views/custom_styles/_primer_color_mapping.erb
```diff
@@ -38,6 +38,8 @@
     --button--active-border-color: var(--color-border-default);
     --button--background-hover-color: var(--color-btn-hover-bg);
     --button--background-color: var(--color-btn-bg);
-   
+    --button--highlight-background-color: #1A67A3;
```
HDinger: Where do these magic numbers come from? Is there a reason not to use a primer variable? I thought we agreed, to use the existing button variables for these buttons as well?

frontend/src/global_styles/content/_buttons.sass
```diff
@@ -90,7 +90,7 @@ a.button
   &.-alt-highlight
     @include button-style(var(--button--alt-highlight-background-color), var(--button--alt-highlight-background-hover-color), var(--button--alt-highlight-font-color))
     @include button-style(var(--button--alt-highlight-background-color), var(--button--alt-highlight-background-hover-color), var(--button--alt-highlight-font-color))
-    border-color: var(--button--alt-highlight-border-color)
+    border-color: var(--button--alt-highlight-background-color)
```
HDinger: Are you sure that this is accepted by the product team? Please double check with them in doubt.

frontend/src/global_styles/layout/_main_menu.sass
```diff
@@ -51,6 +51,7 @@ $arrow-left-width: 40px
     +allow-vertical-scrolling
     height: calc(100vh - var(--header-height))
     position: relative
+    border-right: 1px solid var(--main-menu-border-color)
```
HDinger: What is the reason for that change? The commit message says it is for the case that there is no EE token present. However, this is not reflected in the code.

HDinger (changes requested):
Hi @bsatarnejad 

I did not have the chance to test the new changes, so I only looked at the code. I have some remarks, most of them questions about certain changes.

frontend/src/app/spot/styles/sass/components/list.sass
bsatarnejad: Hi @HDinger, thanks for remarks.
I checked Github, while hovering over list items, the color is not changed both in normal and high-contrast modes. 
I thought maybe we want it to look like Github.

app/views/custom_styles/_primer_color_mapping.erb
bsatarnejad: these variables are used for styling primary buttons (blue ones), maybe I made a mistake here, we don't want to have them with our primary colors in high-contrast mode? Github only has gray and green buttons.

frontend/src/global_styles/layout/_main_menu.sass
bsatarnejad: It was not correct, the issue was related to the variable 'main-menu-border-width' which was 0px, I set it to 1px for high contrast mode.

frontend/src/app/spot/styles/sass/components/list.sass
HDinger: Mhh I see.. Could you please ask the product team about that?

app/views/custom_styles/_primer_color_mapping.erb
HDinger: yes, I agree, we don't want our primary colors, but the ones from Primer. But why didn't you use the variable from primer but the hexcode instead?

frontend/src/app/spot/styles/sass/components/list.sass
bsatarnejad: Hi @HDinger ,
I asked, in normal mode, it will be kept as it is now (color change on hover effect). In high-contrast mode, the color won't change while hovering over items.

app/views/custom_styles/_primer_color_mapping.erb
bsatarnejad: I made a mistake here and thought we want to use our primary color for buttons in high-contrast mode. So I used the hexcode of colors we used in light mode for primary buttons. I rechecked it with Marc and now I map the color used for our primary buttons to primer primary color.

frontend/src/app/features/work-packages/components/wp-buttons/wp-status-button/wp-status-button.component.sass
```diff
@@ -6,7 +6,7 @@
     padding: 5px
     margin-bottom: 0
     background-color: var(--status-selector-bg-color)
-    border: none
+    border: 1px solid var(--button--border-color)
```
HDinger: This will effect the normal mode as well. Is this really intended?

frontend/src/app/spot/styles/sass/components/list.sass
```diff
@@ -106,9 +109,10 @@
     padding: $spot-spacing-0_25 $spot-spacing-1
 
   &_active &--item-action_active
-    color: var(--content-link-hover-active-color)
+    font-weight: var(--base-text-weight-semibold, 600)
```
HDinger: We have an internal variable `var(--base-text-weight-bold)` for bold text (which then uses the Primer variable). Please use that for consistency.

frontend/src/global_styles/common/header/app-menu.sass
```diff
@@ -22,7 +22,10 @@
 
   &--item-action
     background: transparent
-    border: 0
+    border-right: 1px solid transparent
+    border-left: 1px solid transparent
+    border-top: 0
+    border-bottom: var(--header-border-bottom-width) solid transparent
```
HDinger, line 28: Is there a reason that the border is always transparent, but on hover (see below) it is defined by a variable?

frontend/src/global_styles/content/_autocomplete.sass
```diff
@@ -127,6 +128,7 @@ div.autocomplete
 
   &.ng-option-marked
     background-color: $spot-color-basic-gray-6 !important
+    border: 1px solid var(--list-item-hover--border-color) !important
```
HDinger: In the normal mode, this change leads to the items jumping when hovering over them

frontend/src/global_styles/content/_avatar.sass
```diff
@@ -21,10 +19,10 @@ $user-avatar-default-bg-color: white
   vertical-align: middle
   cursor: inherit
   user-select: none
-  background-color: $user-avatar-default-bg-color
   // Default size
   line-height: 36px
   font-size: 16px
+  border: 1px solid var(--avatar-border-color)
```
HDinger, line 25: Please double check the alignment of the initals in the avatar. Due to the border, there are now too low.

With border (current state):
Bildschirmfoto 2023-10-20 um 08 48 57

Without border (old state):
Bildschirmfoto 2023-10-20 um 08 49 08

frontend/src/global_styles/layout/_drop_down.sass
```diff
@@ -102,7 +103,7 @@
 
   LI > .selected
     background: var(--drop-down-selected-bg-color)
-    color: var(--drop-down-selected-font-color)
+    font-weight: var(--base-text-weight-semibold, 600)
```
HDinger: Same as above: Please use our internal variable for consistency (`var(--base-text-weight-bold)`)

HDinger (changes requested):
Hi @bsatarnejad 

I am sorry for the many iterations 🙈 I found some issues in the normal mode which we have to adress especially the jumping in the ngSelect options. I think once you adressed my last remarks we should merge it ☺️

frontend/src/app/features/work-packages/components/wp-buttons/wp-status-button/wp-status-button.component.sass
bsatarnejad: Since it's a button and each button on Github has a border, I added a border for WP status button. If it's not intended, I can change it to have border only in high-contrast mode.

frontend/src/global_styles/common/header/app-menu.sass
bsatarnejad: Yes, if we don't set a transparent border for an item, while hovering over it, it will be jumpy.

frontend/src/app/features/work-packages/components/wp-buttons/wp-status-button/wp-status-button.component.sass
HDinger: I see.. Please double check with product about this.

bsatarnejad: Hi @HDinger 
I talked with Marc about this issue, and we decided to have a border only in high-contrast mode.

HDinger (approved):
Great job 👍 Thanks for all the effort!

bsatarnejad:
Happy to have high-contrast mode in OP from now on! ;)

PR #13648, opened 2023-09-06
Implementation/49112 enable login to microsoft graph api

apfohl:
https://community.openproject.org/projects/openproject/work_packages/49112/activity

This PR implements to parts that are needed to login to Azure for gaining access to the OneDrive/SharePoint API.

What was done?

- Storage representer no longer returns an oauth application for a OneDrive storage
- Adding a OneDrive storage to a project now skips the `project folder` configuration, as it's not available for OneDrive
- `folder_mode` is set to `inactive` on default for OneDrive storages
- CSP configuration filters out OneDrive as they have not `host`

### DIFF
--- a/modules/storages/lib/api/v3/storages/storage_representer.rb
+++ b/modules/storages/lib/api/v3/storages/storage_representer.rb
@@ -178,12 +178,14 @@ module API::V3::Storages
     end
 
     associated_resource :oauth_application,
-                        skip_render: ->(*) { !current_user.admin? },
+                        skip_render: ->(*) { !current_user.admin? || !represented.provider_type_nextcloud? },
                         getter: ->(*) {
+                          next unless current_user.admin? && represented.provider_type_nextcloud?
+
                           ::API::V3::OAuth::OAuthApplicationsRepresenter.create(represented.oauth_application, current_user:)
                         },
                         link: ->(*) {
-                          next unless current_user.admin?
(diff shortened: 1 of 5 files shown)


modules/storages/lib/api/v3/storages/storage_representer.rb
```diff
@@ -178,12 +178,14 @@ def initialize(model, current_user:, embed_links: nil)
...
-                        skip_render: ->(*) { !current_user.admin? },
+                        skip_render: ->(*) { !current_user.admin? || !represented.provider_type_nextcloud? },
                         getter: ->(*) {
+                          next unless current_user.admin? && represented.provider_type_nextcloud?
+
                           ::API::V3::OAuth::OAuthApplicationsRepresenter.create(represented.oauth_application, current_user:)
                         },
                         link: ->(*) {
-                          next unless current_user.admin?
+                          next unless current_user.admin? && represented.provider_type_nextcloud?
```
Kharonus: 🟡 extract this bool condition to a meaningful method name and use it at all three places (first one after applying De Morgan's rule).

e.g.

```ruby
def represent_oauth_application?
  current_user.admin? && represented.provider_type_nextcloud?
end
```

Kharonus (changes requested):
This should be just the first step, right? I would expect some FE changes, if login is already possible.

You mentioned CSP configs, what is to be done about them?

Kharonus (approved):

PR #13652, opened 2023-09-06
[#49956] Accidentaly granting access to Nextcloud project folders that are no members of the project

wielinde:
https://community.openproject.org/work_packages/49956

This PR attempts to fix a problem with the access control lists of project folders in Nextcloud. Apparently a memoized object (for performance optimization) containing the ACL settings for Nextcloud users that have an admin pendant in OpenProject got unexpectedly extended with other users. The other users were members of other projects that had their persmissions already set in a previous loop for a different project. So the bug made memoized object grow with each setting of ACLs per project folder. Non admin users falsely were treated as admins. And admins always have access to all project folders.

### DIFF
--- a/modules/storages/app/services/storages/group_folder_properties_sync_service.rb
+++ b/modules/storages/app/services/storages/group_folder_properties_sync_service.rb
@@ -201,24 +201,15 @@ class Storages::GroupFolderPropertiesSyncService
   end
 
   def project_folder_permissions(project:)
-    tokens_query = OAuthClientToken
-                 .where(oauth_client: @storage.oauth_client)
-                 .where.not(id: @admin_tokens_query)
-                 .includes(:user)
-    # The user scope is required in all cases except one:
-    #   when the project is public and non member has at least one storage permission
-    #   then all non memebers should have access to the project folder
-    if !(project.public? && Role.non_member.permissions.intersect?(PERMISSIONS_KEYS))
-      tokens_query = tokens_query.where(users: project.users)
-    end
-    tokens_query.each_with_object({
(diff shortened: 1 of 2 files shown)


modules/storages/app/services/storages/group_folder_properties_sync_service.rb
```diff
@@ -212,7 +212,7 @@ def project_folder_permissions(project:)
       tokens_query = tokens_query.where(users: project.users)
     end
     tokens_query.each_with_object({
-                                    users: admins_project_folder_permissions,
+                                    users: admins_project_folder_permissions.clone,
```
wielinde: @ba1ash `admins_project_folder_permissions` is a memoized object. We need to clone it first, if we want to modify the object as we do in line 219. Othwerwise we also modify the memoized object and that would be a bad side effect.

aaron-contreras (approved):
Discussed the bug description with Wieland. The thesis adds up 👍 Only thing I'd like to see is a spec that demonstrates the current unexpected behavior and green with the fix.

aaron-contreras:
Hi, @wielinde I've stayed a bit after hours and refactored the service to hopefully make the intention a lot clearer and remove any hackiness or hidden assumptions. Also, I made sure to update the current specs so that they would do two things:

1. Demonstrate the buggy behavior with the former implementation of `#project_folder_permissions` by raising errors of un-stubbed requests (because the request body would include unexpected users).
2. Be green only if we don't cascade permissions and bundle permissions down to subsequent project folders.

wielinde:
@aaron-contreras Thank you for improving the code and adding specs. Thank you @ba1ash for reviewing and simplifying it.

PR #13654, opened 2023-09-07
[Maintenace] Refactor of the Connection Manager

mereghost:
## What's this?

~~A Pull Request.~~
This is an attempt of fixing some of the more glaring issues with the `OAuthClients::ConnectionManager` (or `CM`) like the fact that it is tailored for `Nextcloud` usage.

## Why do we need it?

~~To merge code~~
With the introduction of the `OneDrive/Sharepoint` integration, we have new authorization and  refresh URIs, plus potential extra headers for some requests.

So we now have a reason and an opportunity to uncouple this.

## How does it work?

~~It is a mystery.~~
The trick was to remake this N+1 times. 🤣 This allowed me to identify some similar patterns around the storages and commands/queries.

I introduced the `Storages::Peripherals::OAuthConfigurations` namespace as a place to gather all the different details needed by the `CM`.

These objects then were made available directly from the `Storage` by calling the `#oauth_configuration` method, meaning that at any point in the code the type of storage is irrelevant.

I also introduced the `SubclassResponsibility` error, to signal that a method should be implemented by a subclass of the current class.


mereghost:
AHAHAHAHA! I love that you liked my original text on the PR @akabiru so I preserved it, just ~~with a strikethrough~~.

app/services/oauth_clients/connection_manager.rb
```diff
@@ -57,7 +55,9 @@ def get_access_token(scope: [], state: nil)
 
       # Return the Nextcloud OAuth authorization URI that a user needs to open to grant access and eventually obtain
       # a token.
-      @redirect_url = get_authorization_uri(scope:, state:)
+      full_scope = @config.compute_scopes(scope)
```
apfohl: 🟡🔴 I'm pretty sure this in not necessary as the scope from the `@config` is spliced in line 94. Therefore it would do it twice.

modules/storages/app/common/storages/peripherals/oauth_configurations/nextcloud_configuration.rb
```diff
@@ -0,0 +1,83 @@
...
+# along with this program; if not, write to the Free Software
+# Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+#
+# See COPYRIGHT and LICENSE files for more details.
+#++
+
+module Storages
+  module Peripherals
+    module OAuthConfigurations
+      class NextcloudConfiguration
```
apfohl: 🟢 This point in Ruby is super annoying. You can't have an interface defined for different implementations. It honestly makes me super angry.

apfohl: Maybe we can solve this as well with the `SubClassResponsibility`?

apfohl (changes requested):
Overall it looks very good. We now have storage type independent ConnectionManager.

Maybe it was a bit ambiguous of me wanting to have the name `ConnectionManager` wiped out after this PR. One thing I would like to take a few minutes to think about is indeed the `get_access_token` method. It still returns two completely different things. Even just making a plan and writing a new WP for it would be good enough.

mereghost:
@apfohl wrote:

> Maybe it was a bit ambiguous of me wanting to have the name `ConnectionManager` wiped out after this PR. 

I thought of renaming it but was at a loss of what to call it.

> One thing I would like to take a few minutes to think about is indeed the `get_access_token` method. It still returns two completely different things.

Yeah. I don't like it too, albeit one can make a point that it returns always the same thing a `ServiceResult`. I thought of making it return a failure on not having an access token and hooking a `on_failure` block asking for the `uri`, but that would basically emulate the behaviour in there (i.e. Either(Token, URI)).

```ruby
result = get_access_token
result.on_failure { |result| follow result.result }
      .on_success { |result| do_oauth_things_with result.result }
```

Do you have a suggestion on how you'd prefer the interaction go?

modules/storages/app/common/storages/peripherals/oauth_configurations/nextcloud_configuration.rb
mereghost: I tried adding a superclass to it, but the benefits where... questionable, so I left it like this.

But yeah, we could introduce something that would reinforce, on the code level, that interface.

app/services/oauth_clients/connection_manager.rb
mereghost: Good catch! :tada: I didn't notice as I replaced this piecemeal and the operation wouldn't return duplicates!

mereghost: There is even a conditional that don't need to exist anymore on line 94. Thanks again.

modules/storages/app/common/storages/peripherals/oauth_configurations/nextcloud_configuration.rb
apfohl: At least I don't want to give up on having some sort of interfaces ;).

apfohl:
> Do you have a suggestion on how you'd prefer the interaction go?

I need to think about this for a while and look through the code. But I'm pretty sure get_access_token must return a result or an error. Nothing else. At the place where it is called, there must be a way to act on the case of a failure and obtain the URL from a different place.

But this is not holding back the merging of this PR.

apfohl (approved):

apfohl (approved):
I like the approach, I've done it in the past as well.

app/services/oauth_clients/connection_manager.rb
```diff
@@ -91,7 +90,8 @@ def refresh_token
     # @param scope (OAuth2 RFC) specifies the resources to access. Nextcloud has only one global scope.
     def get_authorization_uri(scope: [], state: nil)
       client = rack_oauth_client # Configure and start the rack-oauth2 client
-      client.authorization_uri(scope:, state:)
+      joined_scopes = @config.compute_scopes(scope)
```
Kharonus: 🟡 I tend to question the need of merging scopes. Do we have the use case, that we use the same `oauth_configuration` on two different places with two different scopes? It would restrict the interface of that method and make it simpler to use.

modules/storages/app/common/storages/peripherals/oauth_configurations/one_drive_configuration.rb
Kharonus, line 1: 🟢 Like! this config classes really combine the differences in the oauth setup in one place.

modules/storages/spec/factories/storage_factory.rb
```diff
@@ -29,16 +29,22 @@
...
+    # rubocop:disable FactoryBot/FactoryAssociationWithStrategy
+    # For some reason the order of saving breaks STI
     trait :with_oauth_client do
-      oauth_client
+      oauth_client { build(:oauth_client) }
     end
+    # rubocop:enable FactoryBot/FactoryAssociationWithStrategy
 
-    factory :nextcloud_storage, class: '::Storages::NextcloudStorage' do
-      provider_type { Storages::Storage::PROVIDER_TYPE_NEXTCLOUD }
```
Kharonus, line 41: 🔴 ❓  Why is this no longer needed? It had the effect, that e.g. running `create(:storage)` inside a spec would return an error, as you cannot instantiate a storage without the mandatory attribute of `provider_type`. 
As it is removed from the sub factories, I believe, this check will no longer work? Or does it?

Kharonus (changes requested):
one question and one idea.

But overall quite excited about the improvement! 🥂

modules/storages/spec/factories/storage_factory.rb
mereghost: `provider_type` will get automatically filled in when the record is saved as RoR STI mechanism kicks in. So this wasn't needed unless we wanted to use the `:storage` factory -  which we shouldn't as, we discussed previously - it is basically an `abstract class`.

app/services/oauth_clients/connection_manager.rb
mereghost: You have a point. I preserved the ability to pass in more scopes as this could come from the outside of the `ConnectionManager`, but that is not a present use case.

I'll remove this and make the `CM` ask the config about the scopes.

modules/storages/spec/factories/storage_factory.rb
Kharonus: Ah, it it still works. cool! Then no change requested. 🟢

Kharonus (approved):
LGTM

PR #13658, opened 2023-09-07
New Permission Query Interface

klaustopher:
Discussed this in detail with @ulferts and also with @aaron-contreras. Asking for more feedback.

---

Currently most of our permission checks are done via the `allowed_to?(permission, context, global:)` method on the User object. This currently does one of 2 obvious things:

- `allowed_to?(:global_permission, nil, global: true)` - Checks if the user has been given a **global permission** via a global role
- `allowed_to?(:work_package_permission, @work_package)` - Checks if the user has been given a **work package permission** either on the specific work package or on the entire project, the work package contains

It also does the following:

- `allowed_to?(:project_permission, nil, global: true)` - Checks if the user has been given the **project permission** on any project

The last case is described in the permission docs, but it's not 100% obvious how this differs from a global permission.

We do have the `allowed_to_in_project?` and `allowed_to_globally?` methods on the user, but they are sparsely used.

---

We have already introduced entity specific memeberships. During that process @aaron-contreras has added a new key `permissible_on` to each permission. We want to use this newly collected information to also explicitly tell the developer if they are doing something wrong (i.e. check a global permission on a project). For this, we will **remove** the `allowed_to?` method and split it into specific methods, that will then be used throughout the codebase.

I will write the implementation in a bit of pseudo code, this will not be the implementation, but it should give you the idea what will happen.

## Global Permission (i.e. `create_project`, `create_user`):

```ruby
def allowed_globally?(permission)
  # Check that permission can be checked globally
  perm = OpenProject::AccessControl.permission(permission)
  raise IllegalPermissionCheck.new(permission) unless perm.global?

  RolePermission
    .joins(:members)
    .where(permission: perm, members: { user: self, project: nil, entity: nil })
    .exists?
end
```

### Example

```erb
<%- if user.allowed_globally?(:create_user) %>
  <%= link_to new_user_path %>
<%- end %>
```

## Project Permission for a specific project (i.e. `add_work_packages`, `manage_members`)

```ruby
def allowed_in_project?(permission, project)
  # Check that permission can be checked for a project
  perm = OpenProject::AccessControl.permission(permission)
  raise IllegalPermissionCheck.new(permission) unless perm.project?

  RolePermission
    .joins(:members)
    .where(permission: perm, members: { user: self, project: project, entity: nil })
    .exists?
end
```

### Example

```erb
<%- if user.allowed_in_project?(:add_work_packages, @project) %>
  <%= link_to new_project_work_package_path(@project) %>
<%- end %>
```

## Project Permission for <ins>any</ins> project (i.e. `add_work_packages` on a global index page)

```ruby
def allowed_in_any_project?(permission)
  # Check that permission can be checked for a project
  perm = OpenProject::AccessControl.permission(permission)
  raise IllegalPermissionCheck.new(permission) unless perm.project?

  RolePermission
    .joins(:members)
    .where(permission: perm, members: { user: self, project: !nil, entity: nil })
    #                                                        ↳ yes, I know this does not work 👀
    .exists?
end
```

### Example

```erb
<%- if user.allowed_in_any_project?(:add_work_packages) %>
  <%= link_to new_global_work_package_path %>
<%- end %>
```

## Entity specific (i.e. `WorkPackage`) permission for a specific entity (i.e. `view_work_package`)

```ruby
def allowed_in_entity?(permission, entity)
  # Check that permission can be checked for that specific entity
  perm = OpenProject::AccessControl.permission(permission)
  raise IllegalPermissionCheck.new(permission) unless perm.permissible_on.include?(entity.model_name.element.to_sym)

  RolePermission
    .joins(:members)
    .where(permission: perm, members: { user: self, entity: entity })
      .or(permission: perm, members: { user: self, entity: nil, project: entity.project })
    .exists?
end
```

### Example

```erb
<%- if user.allowed_in_entity?(:copy_work_package, @work_package) %>
  <%= link_to copy_project_work_package_path(@project, @work_package) %>
<%- end %>
```

## Entity specific (i.e. `WorkPackage`) permission for <ins>any</ins> entity (i.e. `view_work_package` to determine if we allow access to the global work packages index page)

```ruby
def allowed_for_any_entity?(permission, entity_class)
  # Check that permission can be checked for that specific entity
  perm = OpenProject::AccessControl.permission(permission)
  raise IllegalPermissionCheck.new(permission) unless perm.permissible_on.include?(entity_class.model_name.element.to_sym)

  RolePermission
    .joins(:members)
    .where(permission: perm, members: { user: self, project: nil, entity_type: entity.model_name.name })
      .or(permission: perm, members: { user: self, entity: nil, project: !nil })
    #                                                                    ↳ yes, I still know this is illegal 😇
    .exists?
end
```

### Example

```ruby
class GlobalWorkPackagesController < ApplicationController
  def index
    if !user.allowed_for_any_entity?(:view_work_package, WorkPackage)
      render nothing: true, status: :forbidden
    else
      render
    end
  end
end
```

This last one will need one special case, where we check if the user is allowed the permission to any entity *within a specific project*

## Entity specific (i.e. `WorkPackage`) permission for <ins>any</ins> entity within a given project (i.e. `view_work_package` to determine if we allow access to the <ins>project's</ins> work packages index page)

```ruby
def allowed_for_any_entity?(permission, entity_class, in_project: nil)
  # Check that permission can be checked for that specific entity
  perm = OpenProject::AccessControl.permission(permission)
  raise IllegalPermissionCheck.new(permission) unless perm.permissible_on.include?(entity_class.model_name.element.to_sym)

  if in_project
    RolePermission
      .joins(:members)
      .where(permission: perm, members: { user: self, project: project, entity_type: entity.model_name.name })
        .or(permission: perm, members: { user: self, entity: nil, project: project })
      .exists?
  else
    RolePermission
      .joins(:members)
      .where(permission: perm, members: { user: self, project: nil, entity_type: entity.model_name.name })
        .or(permission: perm, members: { user: self, entity: nil, project: !nil })
      #                                                                    ↳ yes, I still know this is illegal 😇
      .exists?
  end
end
```

### Example

```ruby
class ProjectWorkPackagesController < ApplicationController
  def index
    if !user.allowed_for_any_entity?(:view_work_package, WorkPackage, in_project: @project)
      render nothing: true, status: :forbidden
    else
      render
    end
  end
end
```

---

Those methods will then be used *everywhere* in the codebase, instead of the `allowed_to?` method, so we always explicitly
know what type of permission we are checking for and what context we want (i.e. global vs any project).


cbliard:
First, I like the `↳ yes, I know this does not work 👀` :smile: 

And I like the intent of making things clearer. I think your proposition goes in the right direction.

Why are `allowed_in_project?` and `allowed_in_entity?` distinct? I know that's because of the underlying implementation but from a caller's point of view it makes less sense. It could either be the same: `allowed_in?(permission, project_or_entity)`; or all be distinct: `allowed_in_project?(permission, project)`, `allowed_in_work_package?(permission, wp)`, `allowed_in_meeting?(permission, meeting)`, and so on.

I think also that it is missing the case where the permission is checked for multiple projects. It is not used a lot, I do not remember where it's used, but it's used.

I was thinking that maybe a more fluent interface could be interesting, so here is another suggestion:

```ruby
# global permission
user.allowed_globally.to?(:permission)

# project permission
user.allowed_in_project(project).to?(:permission)
# permission in at least one project
user.allowed_in_project(:any).to?(:permission) # use of a special argument value to change the check from specific to any
user.allowed_in_at_least_one_project.to?(:permission) # easier to understand for me
user.allowed_in_any_project.to?(:permission) # shorter to write/read


# work package permission
user.allowed_for_work_package(wp).to?(:permission)
# work package permission within a project
user.allowed_for_work_package(wp).in_project(project).to?(:permission)
# any work package
user.allowed_for_work_package(:any).to?(:permission)
user.allowed_for_any_work_package.to?(:permission)

# same for other entities
```

Maybe it could be made possible with the builder pattern. Each composition adds an argument to the builder, and the last `to?` builds the query and executes it. The builder could be returned by `#allowed`. Interface would then become:

```ruby
# global permission
user.allowed.to?(:permission)
# or this one, and the one above is an error
user.allowed.globally.to?(:permission)

# project permission
user.allowed.in_project(project).to?(:permission)
# permission in at least one project
user.allowed.in_project(:any).to?(:permission)
user.allowed.in_at_least_one_project.to?(:permission)
user.allowed.in_any_project.to?(:permission)


# work package permission
user.allowed.for_work_package(wp).to?(:permission)
# work package permission within a project
user.allowed.for_work_package(wp).in_project(project).to?(:permission)
user.allowed.in_project(project).for_work_packaget(wp).to?(:permission) # same
# any work package
user.allowed.for_work_package(:any).to?(:permission)
user.allowed.for_any_work_package.to?(:permission)
```

klaustopher:
I am not a fan of introducing a builder pattern with a complex method chain that than needs to be mocked. Permission checking should be as simple as possible, in my opinion. And remembering a complex builder interface is not really simple. 

---

The case of multiple projects is not in my pseudo code, but it would be covered by simply passing in an array of projects/work packages/etc

---

I like the idea of having explicit methods that specify the type of entity instead of just using the generic term entity. That would also eliminate the need to specify the work package class.

```ruby
def allowed_for_any_work_package?(permission)
  allowed_for_any_entity?(permission, WorkPackage)
end
```

to make it easier for the user 👍 Will add this

cbliard:
Indeed, keeping the `allowed_*` methods would be simpler for the caller, and the IDE would also suggest completions.

klaustopher:
This PR will be ripped apart in smaller PRs

PR #13660, opened 2023-09-07
[#49932] Convert file storages index page to primer design

akabiru:
#### https://community.openproject.org/projects/openproject/work_packages/49932

- [x] Adopt [OpenProject::PageHeader](https://github.com/opf/primer_view_components/pull/22) for page heading
   - [x] Confirm whether breadcrumbs remain unchanged
- [x] File storages "table" as [BorderBox](https://primer.style/design/components/border-box/rails/beta)
  - [x] Empty state
  - Possibility to edit/delete a storage ([Separate WP](https://community.openproject.org/work_packages/50212/relations))
- [x] Check on mobile 

Screenshot 2023-10-03 at 3 15 10 PM

Screenshot 2023-10-03 at 2 51 26 PM

<details>
<summary>Mobile Screenshots 📱 </summary>

Screenshot 2023-10-03 at 3 15 27 PM

Screenshot 2023-10-03 at 2 51 42 PM

</details>


modules/storages/app/views/storages/admin/storages/index.html.erb
```diff
@@ -1,31 +1,48 @@
...
-  <li class="toolbar-item">
-    <%= link_to new_admin_settings_storage_path,
-                { class: 'button -alt-highlight',
-                  aria: { label: t('storages.label_new_storage') },
-                  title: t('storages.label_new_storage') } do %>
-      <%= op_icon('button--icon icon-add') %>
-      <span class="button--text"><%= ::Storages::Storage.model_name.human %></span>
+<% if OpenProject::FeatureDecisions.storage_primer_design_active? %>
+  <%= render(Primer::OpenProject::PageHeader.new) do |component| %>
+    <% component.with_title(classes: "f1-light") do %>
```
HDinger: Hi @akabiru 
after longer discussion with the product team about the desired font size (and font weight), we agreed on 24px as default for the PageHeader Component. I just changed that in our viewComponents fork. Thus, the extra class here is not needed any more as soon you rebase.

akabiru: Thank you @HDinger , I'll rebase and update! 👍🏾

modules/storages/app/components/storages/admin/storage_list_component.rb
```diff
@@ -0,0 +1,118 @@
...
+    end
+  end
+
+  class BorderBoxComponent < Primer::Beta::BorderBox
+    def before_render
+      super
+
+      # Remove the _base.sass margin-left from <ul> tag
+      @list_arguments[:classes] = "ml-0"
+    end
```
akabiru: Overriding the default list arguments- that are not overrideable via [the constructor](https://github.com/primer/view_components/blob/main/app/components/primer/beta/border_box.rb#L85) afaics. cc @HDinger

modules/storages/app/components/storages/admin/new_storage_button_component.rb
```diff
@@ -0,0 +1,58 @@
...
+    end
+
+    private
+
+    def button_options
+      { scheme:,
+        size:, tag:,
+        href: new_admin_settings_storage_path,
+        aria: { label: },
+        title: label }
```
cbliard: This will not be helpful from a screen reader perspective. The text being read with Orca is "Storage (pause) visited link".
[The title element should always be avoided](https://www.a11yproject.com/posts/title-attributes/), except for iframe tag. It is not accessible.
The aria-label is useless if it is the same as the text of the element.

To improve it from an accessibility stand point:
* if it acts as a button, use `role="button"`
* the `+` icon is invisible to the user. A sighted user might understand that it is a metaphor for "Add", but blind people will not. They will only "see" a "Storage" button without knowing the action
  * You can add `aria-label="Add new storage"` (for instance)
  * Changing the label from "Storage" to "Add new storage" would maybe make it clearer for some sighted users. That's something to discuss with the UX people.

With these modifications, Orca now reads it as "Add new storage (pause) push button".

modules/storages/app/components/storages/admin/storage_list_component.rb
cbliard: This is necessary because `frontend/src/global_styles/openproject/_base.sass` defines this at the end of the file:

```
ul,
ol
  margin-left: 2rem
```

It works fine with a plain `Primer::Beta::BorderBox` if these styles are removed. What would be the side-effects if these lines were removed?

HDinger: The lines in `frontend/src/global_styles/openproject/_base.sass` are necessary for now to intend our list. Without them, all of our list would show their bullet points too far on the left (maybe even out of view). So we need that. I am aware, that inside the Primer::Beta::BorderBox it looks weird if we do not override it.
However, I am reyling on [this PR](https://github.com/opf/openproject/pull/13527) as it will add a general rule for [lists inside BorderBoxes](https://github.com/opf/openproject/pull/13527/files#diff-ebc8270818b7f26cb6ecc2c0c6b4752069453d5b02a91f6f08c5738ac8387c3cR34). We have to wait for that PR anyway to benefit from the flexbox helpers.

modules/storages/app/components/storages/admin/storage_list_component.rb
```diff
@@ -0,0 +1,72 @@
...
+      if storage.configured?
+        span_tag(storage.name)
+      else
+        render(Primer::Beta::Octicon.new(:'alert-fill', size: :small, color: :severe)) +
+          span_tag(storage.name, classes: 'pl-2')
+      end
+    end
+
+    def formatted_datetime(storage)
+      span_tag(" #{I18n.t('activity.item.created_on', datetime: helpers.format_time(storage.created_at.to_fs(:long)))}")
```
cbliard: It should be `helpers.format_time(storage.created_at))` to properly format with time zones.

cbliard:
I tried to extract the html rendering into erb files too, but with more html than ruby. I find having methods rendering html being called here and there harder to read. Having everything in one erb template seems simpler.

The modifications are in the branch implementation/49932-convert-the-index-page-to-primer-design--with-erb-template. Differences can be seen with this link https://github.com/opf/openproject/compare/implementation/49932-convert-the-index-page-to-primer-design...implementation/49932-convert-the-index-page-to-primer-design--with-erb-template although I did not find how to compare a tip of a branch to the tip of another branch.

Thoughts?

modules/storages/app/components/storages/admin/storage_list_component.rb
akabiru: Good shout 👍🏾

modules/storages/app/components/storages/admin/new_storage_button_component.rb
akabiru: This is super helpful Christophe, thank you 👍🏾

akabiru:
> I tried to extract the html rendering into erb files too, but with more html than ruby. I find having methods rendering html being called here and there harder to read. Having everything in one erb template seems simpler.
> 
> The modifications are in the branch implementation/49932-convert-the-index-page-to-primer-design--with-erb-template. Differences can be seen with this link [implementation/49932-convert-the-index-page-to-primer-design...implementation/49932-convert-the-index-page-to-primer-design--with-erb-template](https://github.com/opf/openproject/compare/implementation/49932-convert-the-index-page-to-primer-design...implementation/49932-convert-the-index-page-to-primer-design--with-erb-template) although I did not find how to compare a tip of a branch to the tip of another branch.
> 
> Thoughts?



> I tried to extract the html rendering into erb files too, but with more html than ruby. I find having methods rendering html being called here and there harder to read. Having everything in one erb template seems simpler.
> 
> The modifications are in the branch implementation/49932-convert-the-index-page-to-primer-design--with-erb-template. Differences can be seen with this link [implementation/49932-convert-the-index-page-to-primer-design...implementation/49932-convert-the-index-page-to-primer-design--with-erb-template](https://github.com/opf/openproject/compare/implementation/49932-convert-the-index-page-to-primer-design...implementation/49932-convert-the-index-page-to-primer-design--with-erb-template) although I did not find how to compare a tip of a branch to the tip of another branch.
> 
> Thoughts?

Cheers @cbliard , responded on Element- one thing that's missing is the `ul` margin left workaround https://github.com/opf/openproject/pull/13660#discussion_r1324388283 unless there's a different approach?

cbliard:
> Cheers @cbliard , responded on Element- one thing that's missing is the `ul` margin left workaround [#13660 (comment)](https://github.com/opf/openproject/pull/13660#discussion_r1324388283) unless there's a different approach?

As Henriette dropped it in [her last commit](https://github.com/opf/openproject/pull/13660/commits/d0a9410a497bea1302e3e63537b8187e432ea25d#diff-d890fe219e1f4dcb4aa4299a43c1c01b51553958dadc108191b688404549765fL109-L117), I chose to ignore it as well.

She said: 
> However, I am reyling on [this PR](https://github.com/opf/openproject/pull/13527) as it will add a general rule for [lists inside BorderBoxes](https://github.com/opf/openproject/pull/13527/files#diff-ebc8270818b7f26cb6ecc2c0c6b4752069453d5b02a91f6f08c5738ac8387c3cR34). We have to wait for that PR anyway to benefit from the flexbox helpers.

spec/support/pages/messages/create.rb
```diff
@@ -44,7 +44,7 @@ def add_text(text)
       find('.ck-content').base.send_keys text
     end
 
-    def click_save
+    def click_save(expect_success: true)
```
akabiru: @HDinger @oliverguenther I'm not sure I understand this change https://github.com/opf/openproject/pull/13660/commits/84cb848e418a7e599aa531388b17822581375a42 as it doesn't seem to change any behaviour 🤔

config/locales/en.yml
```diff
@@ -1318,6 +1318,8 @@ en:
       x_seconds:
         one: "1 second"
         other: "%{count} seconds"
+    formats:
+      long: "%Y-%m-%d %H:%M:%S %Z"
```
HDinger: We already have a time format defined. So please double check with the product team whether this format is really required, or whether our existing one can be used.

frontend/src/stimulus/controllers/dynamic/buttons/op-primer-responsive-button.controller.ts
```diff
@@ -0,0 +1,60 @@
...
+ * along with this program; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ *
+ * See COPYRIGHT and LICENSE files for more details.
+ * ++
+ */
+
+import { Controller } from '@hotwired/stimulus';
+
+export default class OpPrimerResponsiveButtonController extends Controller {
```
HDinger: I'd really really like to avoid that class. First, it is super hard to maintain. You are applying primer classes and breakpoints in here which will not be updated if Primer changes them. Second, I'd question the need of it completely. Is it really necessary to switch to a smaller button on mobile?

modules/storages/app/components/storages/admin/storage_list_component.html.erb
```diff
@@ -0,0 +1,78 @@
...
+      end
+    end
+  else
+    render Primer::Beta::Blankslate.new(border: true) do |component|
+      component.with_visual_icon(icon: :cloud)
+      component.with_heading(tag: :h2).with_content(I18n.t('storages.storage_list_blank_slate.heading'))
+      component.with_description { I18n.t('storages.storage_list_blank_slate.description') }
+
+      component.with_primary_action(
+        href: new_admin_settings_storage_path,
```
HDinger: Are these always the same `button_options` as in the `NewStorageButtonComponent`? If yes, we should reuse them.

modules/storages/app/components/storages/admin/storage_list_component.html.erb
```diff
@@ -0,0 +1,78 @@
...
+
+      storages.map do |storage|
+        component.with_row(scheme: :default, id: storage_row_css_id(storage)) do
+          grid_layout('op-storage-list', tag: :div, align_items: :center) do |grid|
+            grid.with_area(:name, tag: :div, mr: 1) do
+              render(Primer::Beta::Link.new(href: edit_admin_settings_storage_path(storage), scheme: :primary)) { storage.name }
+            end
+
+            unless storage.configured?
+              grid.with_area(:label, tag: :div) do
```
HDinger: `label` is a bad name for a grid-area, as it is not not really explanatory. Something like `error-label` would be better

modules/storages/app/components/storages/admin/storage_list_component.sass
```diff
@@ -0,0 +1,27 @@
+.op-storage-list
+  display: grid
+  grid-template-columns: fit-content(180px) 1fr 180px 100px 240px
+  grid-template-areas: "name label user provider time" "host host user provider time"
+
+  .op-principal--name
+    @extend .text-light
```
HDinger: Colors should be set via the system_arguments of a component because it is less error-prone and we will notice when something changes on the Primer side

modules/storages/app/components/storages/admin/storage_list_component.sass
```diff
@@ -0,0 +1,27 @@
...
+
+  @media screen and (max-width: $breakpoint-md)
+    grid-template-columns: fit-content(180px) 1fr
+    grid-template-areas: "name label" "host host" "user user" "time time"
+    grid-row-gap: 5px
+
+    &--provider
+      display: none
```
HDinger: There is a system argument, which you can set to hide elements on certain breakpoints `hide: :sm`. Same for the user and time element

modules/storages/app/controllers/storages/admin/storages_controller.rb
```diff
@@ -166,7 +166,7 @@ def default_breadcrumb
   # See: default_breadcrum above
   # Defines whether to show breadcrumbs on the page or not.
   def show_local_breadcrumb
-    true
+    !OpenProject::FeatureDecisions.storage_primer_design_active?
```
HDinger, line 169: I am wondering whether we really need the feature flag after all. 🤔 Maybe we can discuss that later.

modules/storages/spec/features/admin_storages_spec.rb
```diff
@@ -31,12 +31,65 @@
...
+
+      before do
+        complete_storage
+        incomplete_storage
+      end
+
+      it 'renders a list of storages' do
+        visit admin_settings_storages_path
+
+        expect(page).to have_css('.op-storage-list--name', text: complete_storage.name)
```
HDinger: We have the patter to not test on classes any more as you'd have to adapt all tests if you change a class name. Instead we have a data attribute called `data-test-selector` which you set on those elements that you need, and then test for it. This selector is independent of the class and idealy not touched any more in the future.

modules/storages/app/components/storages/admin/storage_list_component.sass
```diff
@@ -0,0 +1,27 @@
+.op-storage-list
+  display: grid
+  grid-template-columns: fit-content(180px) 1fr 180px 100px 240px
+  grid-template-areas: "name label user provider time" "host host user provider time"
```
HDinger: We have to do some fine-tuning at the grid I think. On smaller screens the timestamp overflows quite easily:
Bildschirmfoto 2023-10-04 um 09 17 50

Further, the label should probably not be the element that has `1fr` because it is not even visible in most cases. This leads to the name being cut off weirdly.
Bildschirmfoto 2023-10-04 um 09 19 08

HDinger (changes requested):
Hi @akabiru 

the page looks really great! 🥳  I have some remarks, but most of them are rather small. I can also help you with the needed CSS changes, if you want to. Further, I think we can discuss the need of the feature flag. If we remove that, we could get rid of some old code as well which would be nice :)

config/locales/en.yml
akabiru: Good shout, I'll check with Marc 👍🏾

frontend/src/stimulus/controllers/dynamic/buttons/op-primer-responsive-button.controller.ts
akabiru: Yeah, working this out with CSS was rather difficult as we have to btn switch classes- however I do agree that we can get rid of this if it's not a must have- will check with product 👍🏾

modules/storages/app/components/storages/admin/storage_list_component.html.erb
akabiru: Ah, yeah- good point

modules/storages/app/components/storages/admin/storage_list_component.html.erb
akabiru: They are, although It seemed rather hacky to try and reuse them as primer primary_action slot is tightly coupled- Ideally it would have been better to just pass the `NewStorageButtonComponent`.

We could however do sth like

```ruby
component.with_primar_action(**NewStorageButtonComponent.new.button_options) do
  button.with_leading_visual_icon(icon: :plus)
  I18n.t("storages.label_storage")
end

```

modules/storages/app/components/storages/admin/storage_list_component.sass
akabiru: AFAICT, the `User::AvatarComponent` is not a primer component- we're still reusing the legacy avatar helper 🤔 

https://github.com/opf/openproject/blob/dev/app/components/users/avatar_component.rb

modules/storages/app/components/storages/admin/storage_list_component.sass
akabiru: Oh, I didn't know that- nice one 👍🏾

modules/storages/app/controllers/storages/admin/storages_controller.rb
akabiru: Since we've moved the edit & delete buttons into the storage edit/view page- removing the feature flag at this stage would be disruptive to the user as they wouldn't have a delete action 😬 My plan was to retain the flag and remove it once we have the edit page built out-

spec/support/pages/messages/create.rb
akabiru: ping on this one @HDinger 😄

frontend/src/stimulus/controllers/dynamic/buttons/op-primer-responsive-button.controller.ts
akabiru: Marc- we can remove this

config/locales/en.yml
akabiru: Marc- revert back to activity timestamp

HDinger (approved):
Great job @akabiru 🥳 Thanks for all the iterations! The result was definitely worth it. Once the test selector is fixed, this can be merged 👍

spec/support/pages/messages/create.rb
oliverguenther: you can remove it, it was part of a temporary fix and I seem to have overlooked that part

PR #13669, opened 2023-09-08
[#49120] Get single file of onedrive storage

Kharonus:
[OP#49120](https://community.openproject.org/wp/49120)

### What is in here?

- added the `files_into_query` for one drive to the code base
- added unit tests for this query
- added test data from https://developer.microsoft.com/en-us/graph/graph-explorer/

### Why do we need it?

- step by step we need to implement all interaction queries and commands for the one drive storage provider type

### Information for reviewers:

- one review of @apfohl or @mereghost should suffice, as this only adds a single query
- functionality of all one drive queries are not yet tested against real sharepoint tenants

### DIFF
--- a/modules/storages/spec/common/peripherals/storage_interaction/one_drive/files_info_query_spec.rb
+++ b/modules/storages/spec/common/peripherals/storage_interaction/one_drive/files_info_query_spec.rb
@@ -1,0 +1,200 @@
+# frozen_string_literal: true
+
+#-- copyright
+# OpenProject is an open source project management software.
+# Copyright (C) 2012-2023 the OpenProject GmbH
+#
+# This program is free software; you can redistribute it and/or
+# modify it under the terms of the GNU General Public License version 3.
+#
+# OpenProject is a fork of ChiliProject, which is a fork of Redmine. The copyright follows:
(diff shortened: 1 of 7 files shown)


modules/storages/spec/common/peripherals/storage_interaction/one_drive/files_info_query_spec.rb
```diff
@@ -0,0 +1,200 @@
...
+
+  private
+
+  def read_json(name)
+    File.readlines(payload_path.join("#{name}.json")).join
+  end
+
+  def payload_path
+    Pathname.new(Rails.root).join('modules/storages/spec/support/payloads/')
+  end
```
mereghost: WDYT of moving this to a `one_drive_test_helper` as I get the feeling we will be using this a lot? :P

mereghost (approved):
LGTM! Just a minor suggestion that can surely be done in another PR later.

modules/storages/spec/common/peripherals/storage_interaction/one_drive/files_info_query_spec.rb
Kharonus: :P makes sense, will add a commit

PR #13672, opened 2023-09-08
Add an extra test case for file_links_api_spec.rb

ba1ash:
- The test case checks the response when the links are tried to be
  created for unknown storage.
- Speed up. On my machine is a huge improvement. Thanks @cbliard for your recommendations!
  - Use shared_let and friends to save time spend in factories.
  - Merge some test cases within the same context.

**Before**:
```
Finished in 2 minutes 17.3 seconds (files took 21.25 seconds to load)
132 examples, 0 failures

[TEST PROF INFO] Time spent in factories: 01:28.528 (61.21% of total time)
```
**After**:
```
Finished in 30.12 seconds (files took 16.82 seconds to load)
131 examples, 0 failures

[TEST PROF INFO] Time spent in factories: 00:01.786 (5.05% of total time)
```

p.s.: @wielinde the specs we thought are absent actually present. We looked into `POST /api/v3/file_links` specs, but we should have checked specs for `POST /api/v3/work_packages/:work_package_id/file_links`.

### DIFF
--- a/modules/storages/spec/requests/api/v3/file_links/file_links_api_spec.rb
+++ b/modules/storages/spec/requests/api/v3/file_links/file_links_api_spec.rb
@@ -34,24 +34,49 @@ require_module_spec_helper
 RSpec.describe 'API v3 file links resource' do
   include API::V3::Utilities::PathHelper
 
-  let(:permissions) { %i(view_work_packages view_file_links) }
-  let(:project) { create(:project) }
-
-  let(:current_user) do
-    create(:user, member_in_project: project, member_with_permissions: permissions)
+  def add_permissions(user, *permissions)
+    role = Role.joins(members: :principal).where('users.id': user).first
+    role.add_permission!(*permissions)
+    user.reload # clear user's project_role_cache
   end
 
-  let(:work_package) { create(:work_package, author: current_user, project:) }
-  let(:another_work_package) { create(:work_package, author: current_user, project:) }
+  def remove_permissions(user, *permissions)
+    role = Role.joins(members: :principal).where('users.id': user).first
+    role.remove_permission!(*permissions)
+    user.reload # clear user's project_role_cache
+  end
 
-  let(:oauth_application) { create(:oauth_application) }
-  let(:storage) { create(:nextcloud_storage, creator: current_user, oauth_application:) }
-  let(:another_storage) { create(:nextcloud_storage, creator: current_user) }
+  def enable_module(project, modul)
+    project.enabled_module_names = project.enabled_module_names + [modul]
+    project.save
+  end
 
-  let(:oauth_client) { create(:oauth_client, integration: storage) }
-  let(:oauth_client_token) { create(:oauth_client_token, oauth_client:, user: current_user) }
+  def disable_module(project, modul)
+    project.enabled_module_names = project.enabled_module_names - [modul]
+    project.save
+  end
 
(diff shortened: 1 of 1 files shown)


cbliard:
If the goal is to save time, when running the tests, it looks like a lot of time is spent in the factories:

```
rspec modules/storages/spec/requests/api/v3/file_links/file_links_api_spec.rb
[...]
Finished in 29 seconds (files took 7.75 seconds to load)
132 examples, 0 failures
[...]
[TEST PROF INFO] Time spent in factories: 00:25.524 (80.7% of total time)
```

To reduce this time, you can rely on `shared_let` to create the objects in database only once. They are then reloaded from database before each test case. Much cheaper than creating on each test.

`current_user` and `project` are good candidates.

When doing so:
```
Finished in 15.28 seconds (files took 7.8 seconds to load)
132 examples, 35 failures

[TEST PROF INFO] Time spent in factories: 00:11.517 (64.13% of total time)
```

There are now failures, because some tests need to be rewritten. For instance:

```ruby
    context 'if user has not sufficient permissions' do
      let(:permissions) { %i(view_work_packages) }

      it_behaves_like 'API V3 collection response', 0, 0, 'FileLink', 'Collection' do
        let(:elements) { [] }
      end
    end
```

This wont work anymore as the permissions are applied to the project/user membership which has already been created. It can then be rewritten as

```
    def remove_permissions(user, *permissions_to_remove)
      user_permissions = RolePermission.joins(role: { members: :principal }).where('users.id': user)
      user_permissions.where(permission: permissions_to_remove).delete_all
      user.reload # clear user's project_role_cache
    end

    context 'if user has not sufficient permissions' do
      before do
        remove_permissions(current_user, :view_file_links)
        get path
      end

      it_behaves_like 'API V3 collection response', 0, 0, 'FileLink', 'Collection' do
        let(:elements) { [] }
      end
    end
```

Also the `get path` has to be repeated because it needs to occur after the permissions have been changed.

These changes can be a pain to implement but the time savings are real.
Here it saved half of the total time (Time spent in factories went from 00:25.524 to 00:11.517).

Another easy saving: when creating a work package through the factory, it also creates a user (the author), a project, a status and a priority, unless they are specified on creation. In this test, `author` and `project` are specified for each work package creation, but not `status` nor `priority`.

To have a default priority and status created automatically, you can use `shared_association_default` like so:
```ruby
  shared_association_default(:priority) { create(:priority) }
  shared_association_default(:status) { create(:status) }
```

This saves 2 seconds: Time spent in factories went from 00:11.517 to 00:09.928

And I did not realize that the 2 work packages instances never change. They can be changed to a `shared_let` as well. This saves 2 more seconds again: Time spent in factories went from 00:09.928 to 00:07.792

And `oauth_application`, `storage`, and `another_storage` can also use `shared_let`. It saves another 3 seconds: Time spent in factories went from 00:07.792 to 00:04.512, for a total time of 8.52 seconds (it was 29 seconds at the beginning).

And `oauth_client`, `oauth_client_token`, and `project_storage` can be converted as well. It saves another 3 seconds: Time spent in factories went from 00:04.512 to 00:00.822 (10% of total time) for a total time of 5.18 seconds.

It needs rewriting part of the tests, but going from 30 seconds to 5 seconds could be worth it.

mereghost (approved):
Nice improvements, @ba1ash! :D

PR #13680, opened 2023-09-11
[#49948] PDF export does not contain custom fields and some core values

as-op:
https://community.openproject.org/work_packages/49948

### DIFF
--- a/app/models/exports/formatters/custom_field.rb
+++ b/app/models/exports/formatters/custom_field.rb
@@ -18,10 +18,14 @@ module Exports
       # Print the value meant for export.
       #
       # - For boolean values, don't use the Yes/No formatting for the UI
+      #   treat nil as false
       # - For long text values, output the plain value
       def format_for_export(object, custom_field)
         case custom_field.field_format
-        when 'bool', 'text'
+        when 'bool'
+          value = object.typed_custom_value_for(custom_field)
+          value == nil ? false : value
+        when 'text'
           object.typed_custom_value_for(custom_field)
         else
           object.formatted_custom_value_for(custom_field)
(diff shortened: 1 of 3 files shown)


HDinger (commented):
Hi @as-op 

I am not super familiar with this feature, so please ignore my comments if they are out of scope. I found some fields that are missing in the PDF export:

* A boolean custom field which has value `false` is not shown. Propably because it is considered empty and thus filtered out.
* The field "Spent units" ("Gebuchte Einheiten") is missing although there is a value in it.
* The order of the fields is confusing. I would have expected them to be in the same order as in the WP view, but it isn't

as-op:
Thanks a lot, @HDinger! 🙏 Very good catches, I check these two issues:

> * A boolean custom field which has value `false` is not shown. Propably because it is considered empty and thus filtered out.
> * The field "Spent units" ("Gebuchte Einheiten") is missing although there is a value in it.

There is no intended order right now, it's the same as in `::Query.available_columns(work_package.project)`, I could sort alphabetically, but that's not good either. 
In the [next step](https://community.openproject.org/projects/openproject/work_packages/49977/activity) the visibility and order will taken by the WP view.

> * The order of the fields is confusing. I would have expected them to be in the same order as in the WP view, but it isn't

as-op:
Hello @HDinger, may I kindly ask you to re-review? 

* I've added support for a **spent units** entry in the attributes table. - It is quite an exception as it's neither an available column nor a work package attribute. 😓  I adapted the handling I found in the API representer.
* **Custom boolean fields** may have `nil `as value if they have never been toggled, the UI shows nil as "no". I adapted this behaviour for the export.

Thanks again!

HDinger (commented):
The fields are now shown 👍 One minor thing for the "spent units" field: The unit seems to be wrong:
WP view:
Bildschirmfoto 2023-09-12 um 14 16 07

PDF: 
Bildschirmfoto 2023-09-12 um 14 16 17

as-op:
Hi @HDinger, 
I saw that but it is more of an UI issue. WP view uses the name of the cost type. While the export uses the singluar/plural versions, I then preferred that usage.
image_config
WP view:
image_ui
PDF:
image_export

Thanks for reviewing!

PR #13689, opened 2023-09-12
[chore] reintroduced file info query

Kharonus:
- added file info query for one drive
- let one drive files info use file info query
- reorganize spec folder to satisfy rubocop
- added file info for nextcloud
- amended api logic to use file info query

### What is in this PR?

I reintroduced the `file_info_query` (singular) for both nextcloud and one drive.

### But... why?

We have the use cases to use them. It was merged into `nextcloud/files_info_query`, at the price of having some logic in the API to handle the errors differently. Now, with one drive provider type, it makes less sense.

### How can I test it?

Either manipulate and run the unit tests for it, or enter some auth data for a one drive storage in the database and run thise endpoint against your live data:

- `/api/v3/storages/:storage-id/files/:file-id`


modules/storages/app/common/storages/peripherals/storage_error_helper.rb
```diff
@@ -42,5 +42,20 @@ def raise_error(error)
         raise API::Errors::InternalError.new
       end
     end
+
+    def error_to_code(error)
+      case error.code
+      when :bad_request
```
m-graf: You can get a similar mapping out of [HTTP Response](https://www.rubydoc.info/gems/http/HTTP/Response/Status#SYMBOLS-constant) :)

Kharonus: True, nice hint, thanks. Sadly we do not use that gem (yet? :D). But I won't introduce a new dependency just because of some http code mappings. ^^

Kharonus:
Added @akabiru to the reviewers, as the query specs were altered. Maybe we want to raise the question here, if more specs should use `vcr`.

modules/storages/spec/features/manage_project_storage_spec.rb
```diff
@@ -67,19 +63,14 @@
...
-          '11': {
-            status: 'OK',
-            statuscode: 200,
-            id: 11,
-            name: 'Folder1',
-            path: 'files/Folder1',
-            mtime: 1682509719,
-            ctime: 0
-          }
+          status: 'OK',
```
akabiru, line 67: Q: did the file info payload from nextcloud response change? 🤔  _Another great case for `VCR`, less ambiguity_ 😄

akabiru: Oh, I see it's now a `get` for a single folder

modules/storages/app/common/storages/peripherals/storage_error_helper.rb
```diff
@@ -42,5 +42,20 @@ def raise_error(error)
...
+      when :not_authorized
+        401
+      when :forbidden
+        403
+      when :not_found
+        404
+      else
+        500
+      end
+    end
```
akabiru: 🟢 One option is to use [`Rack::Utils::SYMBOL_TO_STATUS_CODE`](https://github.com/rack/rack/blob/main/lib/rack/utils.rb#L542-L544). Only issue is we'd have to rename `:not_authorized` to `:unauthorized`

```ruby
[17] pry(main)> %i[bad_request not_authorized forbidden not_found].map { |s| Rack::Utils::SYMBOL_TO_STATUS_CODE[s] }
=> [400, nil, 403, 404]
[18] pry(main)> Rack::Utils::HTTP_STATUS_CODES[401]
=> "Unauthorized"
[20] pry(main)> %i[bad_request unauthorized forbidden not_found].map { |s| Rack::Utils::SYMBOL_TO_STATUS_CODE[s] }
=> [400, 401, 403, 404]
```

modules/storages/spec/common/storages/peripherals/storage_interaction/one_drive/file_info_query_spec.rb
```diff
@@ -0,0 +1,110 @@
...
+  let(:user) { create(:user) }
+  let(:token) { create(:oauth_client_token, user:, oauth_client: storage.oauth_client) }
+  let(:file_id) { '01BYE5RZ5MYLM2SMX75ZBIPQZIHT6OAYPB' }
+  let(:not_found_json) { not_found_response }
+  let(:forbidden_json) { forbidden_response }
+
+  before do
+    stub_request(:get, "https://graph.microsoft.com/v1.0/drives/#{storage.drive_id}/items/#{file_id}")
+      .with(headers: { 'Authorization' => "Bearer #{token.access_token}" })
+      .to_return(status: 200, body: read_json('folder_drive_item'), headers: {})
```
akabiru, line 51: 🟠 [As you pointed out](https://github.com/opf/openproject/pull/13689#issuecomment-1717432225), good time to discuss adopting `VCR` here. Set up is simpler as we don't have to connect to a local docker container and can hit the graph API directly 👍🏾

akabiru (approved):
Top work! 🏆 Two non-blocking remarks for consideration 👍🏾

modules/storages/app/common/storages/peripherals/storage_error_helper.rb
Kharonus: that's a good hint. Will opt for it

modules/storages/spec/common/storages/peripherals/storage_interaction/one_drive/file_info_query_spec.rb
Kharonus: yes... so, should we have a look next week and introduce a vcr example for one drive?

akabiru: Sounds good 👍🏾

mereghost (approved):
LGTM.

We might to fuse the Utils from both nextcloud and onedrive at one point. Or maybe make the necessary objects smarter so we don´t need it.

Kharonus:
Definitely, I tend to do those things in small chunks. right now I moved some logic from one drive queries to a util. We still need to continue on this constantly to keep up a good structure in code.

PR #13698, opened 2023-09-13
[#49495] Add "Shared with users" common query for work packages

aaron-contreras:
## Description
This common query would allow a quick filter for all shared work packages by utilizing the `*` operator

### Creates a new `SharedWithUser` work package filter to achieve this query
* Allows filtering for work packages shared only with users whom are members of a project that the current user is also a member of.
* A new filter strategy and set of operators was created to filter with the following operators:
  * is (OR): filters for work packages shared with _**any**_ of the users in the provided list
  * is (AND): filters for work packages shared with _**every**_ user in the provided list
  * any: filters for any shared work package

## Notes
See: https://community.openproject.org/work_packages/49495


app/models/queries/work_packages/filter/shared_user_filter.rb
```diff
@@ -0,0 +1,80 @@
...
+# See COPYRIGHT and LICENSE files for more details.
+# ++
+
+class Queries::WorkPackages::Filter::SharedUserFilter <
+  Queries::WorkPackages::Filter::PrincipalBaseFilter
+  def available?
+    super && User.current
+                 .allowed_to?(:view_shared_work_packages,
+                              nil,
+                              global: true)
```
klaustopher: Do we know if we exectue that filter within a project context? If so, it would make more sense to check the permission for the current project

app/models/queries/work_packages/filter/shared_user_filter.rb
```diff
@@ -0,0 +1,80 @@
...
+  def shared_work_package_ids
+    base_query = Member.where(entity_type: 'WorkPackage')
+
+    unless %w[* !*].include?(operator)
+      base_query = base_query.where(user_id: values_replaced)
+    end
+
+    base_query.select('entity_id')
+              .distinct
+              .pluck(:entity_id)
```
klaustopher: Question is if we would need a bit more filtering logic herE? 

With the `available?` check above, you would allow access to the filter if the user has the `view_shared_work_packages` permission in **any** project. But here we are not filtering anymore, but just plainly returning all entity ids that are shared in the entire system. Would we need to join the projects permissions here as well and only return work packages belonging to products that the user has the `view_shared_work_packages` permission on?

app/models/queries/work_packages/filter/shared_user_filter.rb
aaron-contreras: You're right, thanks. Diving into this `allowed_to?` stuff has me mixing up the use-case for whether context is present and global is true (and all other variations).

klaustopher (approved):

PR #13706, opened 2023-09-14
Standardize resolution in Cuprite Browser

aaron-contreras:
* Sets 1920x1080 as the default resolution for both headless and headful mode.

* Adds an ENV variable to customize the browser's window resolution when running in headful mode (perhaps because you're working on your laptop and 1920x1080 is too big to fit your screen).


oliverguenther (approved):
Excellent, thank you 🙇

PR #13713, opened 2023-09-15
[#45533] Work-package relationships are lost when copying

mrmir:
https://community.openproject.org/work_packages/45533


app/services/work_packages/bulk/copy_service.rb
```diff
@@ -32,11 +32,48 @@ class CopyService < BulkedService
...
+          wp_copy = alter_work_package(work_package, params)
+
+          result.add_dependent!(wp_copy)
+
+          @wp_map.store(work_package.id, wp_copy.result.id)
+        end
+
+        result.result = false if result.failure?
+
+        copy_relations
```
dombesz: I believe the relations should be copied only when the `result` service call is successful. Using the `on_success` method can be a good approach, similar to what we have in the [`WorkPackages::CopyService`](https://github.com/opf/openproject/blob/c427d1ba3514ffe5d2731d9233f5fe87e797c92c/app/services/work_packages/copy_service.rb#L56-L60).

app/services/work_packages/bulk/copy_service.rb
```diff
@@ -32,11 +32,48 @@ class CopyService < BulkedService
       def initialize(user:, work_packages:)
         super
 
+        @wp_map = {}
```
dombesz, line 37: Just for convenience we can declare an `attr_reader :wp_map` for this variable. Then later on it can be referenced without the `@`. ie: `wp_map.store(work_package.id, wp_copy.result.id)`

dombesz (changes requested):
Overall looks good, I left 2 observations.

app/services/work_packages/bulk/copy_service.rb
```diff
@@ -56,10 +56,9 @@ def bulk(params)
           @wp_map.store(work_package.id, wp_copy.result.id)
         end
 
+        copy_relations unless result.failure?
```
mrmir: I feel this works better than using `on_success`. What do you think @dombesz?

dombesz (approved):
Well done @mrmir ! 👏 Feel free to merge, once the spec failing is fixed.

PR #13724, opened 2023-09-19
Fix the interface for OneDrive FilesQuery

mereghost:
I had the `OneDrive::FilesQuery` returning the wrong object. This PR aims to fix it.

### DIFF
--- a/modules/storages/app/common/storages/peripherals/storage_interaction/one_drive/files_query.rb
+++ b/modules/storages/app/common/storages/peripherals/storage_interaction/one_drive/files_query.rb
@@ -33,7 +33,7 @@ module Storages
     module StorageInteraction
       module OneDrive
         class FilesQuery
-          FIELDS = "?$select=id,name,size,webUrl,lastModifiedBy,createdBy,fileSystemInfo,file,folder,parentReference"
+          FIELDS = "?$select=id,name,size,webUrl,lastModifiedBy,createdBy,fileSystemInfo,file,folder"
 
           using ServiceResultRefinements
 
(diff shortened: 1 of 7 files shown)


modules/storages/app/models/storages/storage_file.rb
```diff
@@ -30,14 +32,14 @@ module Storages
...
     :mime_type,
-    :created_at,
-    :last_modified_at,
+    :created_at, # Time? DateTime (deprecated)?
+    :last_modified_at, # Time? DateTime (deprecated)?
     :created_by_name,
     :last_modified_by_name,
-    :location,
-    :permissions
+    :location, # Should always start with a '/'
```
apfohl, line 41: 🟢 There could be a check in the `initialize` method, to check this and `raise` if it doesn't met the criteria.

apfohl (approved):

modules/storages/app/models/storages/storage_file.rb
mereghost: I'm thinking of adding some type of validation somewhere down the line. I played around a bit with adding it, but for now that's not a pressing matter.

modules/storages/spec/common/storages/peripherals/storage_interaction/one_drive/files_query_spec.rb
```diff
@@ -68,38 +71,67 @@
...
+
+    it 'has one ancestors and one parent' do
+      stub_request(:get, "https://graph.microsoft.com/v1.0/drives/#{storage.drive_id}/items/01BYE5RZ2LCUJQWLOZYNDLDC4DRRGJ7OEN/children#{described_class::FIELDS}")
+        .with(headers: { 'Authorization' => "Bearer #{token.access_token}" })
+        .to_return(status: 200, body: payload, headers: {})
+
+      storage_files = described_class.call(storage:, user:, folder: "01BYE5RZ2LCUJQWLOZYNDLDC4DRRGJ7OEN").result
+
+      expect(storage_files.ancestors.size).to eq(2)
+      expect(storage_files.ancestors.map(&:location)).to contain_exactly("/", "/Contoso Clothing")
```
Kharonus, line 127: ❓ ah, so the decision for now is to cut everything before the colon and just return the location on the assumed (and configured drive)?

🟢 I think it'll work. Fine with this approach.

modules/storages/app/models/storages/storage_file.rb
```diff
@@ -30,14 +32,14 @@ module Storages
   StorageFile = Data.define(
     :id,
     :name,
-    :size,
+    :size, # Integer >= 0
     :mime_type,
-    :created_at,
-    :last_modified_at,
+    :created_at, # Time? DateTime (deprecated)?
+    :last_modified_at, # Time? DateTime (deprecated)?
```
Kharonus, line 38: ❓ What exactly is deprecated?

modules/storages/app/common/storages/peripherals/storage_interaction/one_drive/files_query.rb
```diff
@@ -81,21 +81,65 @@ def uri_path_for(folder)
...
-                name: json[:name],
-                size: json[:size],
-                mime_type: Util.mime_type(json),
-                created_at: DateTime.parse(json.dig(:fileSystemInfo, :createdDateTime)),
-                last_modified_at: DateTime.parse(json.dig(:fileSystemInfo, :lastModifiedDateTime)),
-                created_by_name: json.dig(:createdBy, :user, :displayName),
-                last_modified_by_name: json.dig(:lastModifiedBy, :user, :displayName),
-                location: json.dig(:parentReference, :path),
-                permissions: nil
+                id: Digest::SHA256.hexdigest(component),
```
Kharonus, line 133: 🟡 not sure, this will work, as we probably want the breadcrumbs to contain the ancestors real ids (as for one drive we do not access any "parent" by location. 

Doesn't matter for now, I guess, as this can be fixed, when the breadcrumbs are touched and tested.

Kharonus (approved):
In the end, I only have a 🟡 and two ❓ , so it is approved from my side. Both are no blocker, I'd just like them to be discussed.

modules/storages/app/models/storages/storage_file.rb
mereghost: DateTime is deprecated in favor of Time.

modules/storages/spec/common/storages/peripherals/storage_interaction/one_drive/files_query_spec.rb
mereghost: It keeps the values similar to what we had for nextcloud. =)

modules/storages/app/common/storages/peripherals/storage_interaction/one_drive/files_query.rb
mereghost: The approach is similar to what we do on nextcloud, for now it will suffice. It is a pain to reconstruct all ancestors for OneDrive unless we ditch the `id` approach and move towards a `path` based one (which is "flakier").

Kharonus: yes, might be a pain, but this decision must be made within the next days. When we touch the frontend I suppose. It might also has a slight impact on the API.

modules/storages/app/models/storages/storage_file.rb
Kharonus: Why not starting using time directly?

PR #13726, opened 2023-09-19
Check which tests is covering some code

cbliard:
do not merge please.


cbliard:
`admin` params is only covered by 

```
'./spec/models/permitted_params_spec.rb:526'
'./spec/models/permitted_params_spec.rb:535'
```

PR #13731, opened 2023-09-20
Direct download of file from OneDrive storage

mereghost:
## What?

This PR introduces the `DownloadLinkQuery` to the `OneDrive` storage. It follows the same semantics that the `Nextcloud` version, therefore keeping the contract with the `API`

## How?

When you request a download from the MS Graph API, it returns a redirect with an `authenticated` download URL in its `Location` header.

Also I split the tests from the `Nextcloud` version from the `Registry` tests in its own file.


apfohl (approved):

modules/storages/app/common/storages/peripherals/storage_interaction/one_drive/download_link_query.rb
```diff
@@ -0,0 +1,85 @@
...
+            new(storage).call(user:, file_link:)
+          end
+
+          def call(user:, file_link:)
+            Util.using_user_token(@storage, user) do |token|
+              response = Net::HTTP.start(@uri.host, @uri.port, use_ssl: true) do |http|
+                http.get(uri_path_for(file_link.origin_id), { 'Authorization' => "Bearer #{token.access_token}" })
+              end
+
+              if Net::HTTPRedirection === response
```
Kharonus, line 51: 🔴 I don't understand, why this ever should be true. The request is a simple GET on `"/v1.0/drives/#{@storage.drive_id}/items/#{file_id}"`. This responds with the normal json, we know, and what we need is to extract the `@microsoft.graph.downloadUrl` from it.

There is a hint, for what I did in my PR: I extracted all requests to the standard drive item to an [internal query](https://github.com/opf/openproject/pull/13732/files#diff-3564ab603c086431dba829fc8a9a60ba49195e068db1779898633149222f25ad). This way other queries, as `file_info`, `open_link` and also your `download_link` can then use this internal one and simply map the result to anything else.

modules/storages/spec/common/storages/peripherals/storage_interaction/one_drive/download_link_query_spec.rb
```diff
@@ -0,0 +1,66 @@
...
+  let(:file_link) { create(:file_link) }
+  let(:token) { create(:oauth_client_token, oauth_client: storage.oauth_client, user:) }
+
+  subject(:download_link_query) { described_class.new(storage) }
+
+  before do
+    allow(Storages::Peripherals::StorageInteraction::OneDrive).to receive(:token).and_yield(token)
+    stub_request(:get, "https://graph.microsoft.com/v1.0/drives/#{storage.drive_id}/items/#{file_link.origin_id}")
+      .with(headers: { 'Authorization' => "Bearer #{token.access_token}" })
+      .and_return(status: 302, body: nil, headers: { 'Location' => 'https://somecool.link/from/microsoft' })
```
Kharonus, line 46: 🔴 This is not, how the real graph API would respond to `https://graph.microsoft.com/v1.0/drives/#{storage.drive_id}/items/#{file_link.origin_id}`. See other comment.

Kharonus (changes requested):

modules/storages/app/common/storages/peripherals/storage_interaction/one_drive/download_link_query.rb
mereghost: The [download API](https://learn.microsoft.com/en-us/graph/api/driveitem-get-content?view=graph-rest-1.0&tabs=http) responds on the `/drives/{drive-id}/items/{item-id}/content` URI.

While this returns the same URL as the `@microsoft.graph.downloadUrl`, it is simpler than actually getting the entire response, parsing it and forwarding it, give that both are "short lived urls".

mereghost: Which makes me notice that I forgot the `/content` 🤦🏼

Kharonus: Now it makes sense :P

modules/storages/spec/common/storages/peripherals/storage_interaction/one_drive/download_link_query_spec.rb
Kharonus: solved

Kharonus (approved):
LGTM

PR #13733, opened 2023-09-20
allow adding plugins via Gemfile.{modules,plugins}

machisuji:
before this change you would get a deployment mode error since you would also need to update Gemfile.lock before the build.

Alternative to https://github.com/opf/openproject-flavours/pull/4. We should pick one.

### DIFF
--- a/docker/prod/Dockerfile
+++ b/docker/prod/Dockerfile
@@ -169,9 +169,14 @@ COPY vendor ./vendor
 # some gemspec files of plugins require files in there, notably OpenProject::Version
 COPY lib ./lib
 
-RUN bundle install --quiet --deployment --path vendor/bundle --no-cache \
-  --with="$RAILS_GROUPS" --without="test development" --jobs=8 --retry=3 && \
-  rm -rf vendor/bundle/ruby/*/cache && rm -rf vendor/bundle/ruby/*/gems/*/spec && rm -rf vendor/bundle/ruby/*/gems/*/test
+RUN \
+  bundle config set --local path 'vendor/bundle' && \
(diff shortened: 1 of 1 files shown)


docker/prod/Dockerfile
```diff
@@ -169,9 +169,15 @@ COPY vendor ./vendor
 # some gemspec files of plugins require files in there, notably OpenProject::Version
 COPY lib ./lib
 
-RUN bundle install --quiet --deployment --path vendor/bundle --no-cache \
-  --with="$RAILS_GROUPS" --without="test development" --jobs=8 --retry=3 && \
-  rm -rf vendor/bundle/ruby/*/cache && rm -rf vendor/bundle/ruby/*/gems/*/spec && rm -rf vendor/bundle/ruby/*/gems/*/test
+RUN \
+  bundle config set --local path 'vendor/bundle' && \
+  bundle config set --local with '$RAILS_GROUPS' && \
```
cbliard: Shouldn't it be double quotes?
This would set the following in `.bundle/config`:
```
BUNDLE_WITH: "$RAILS_GROUPS"
```
But I could not figure out where this `RAILS_GROUPS` is set.

cbliard: `RAILS_GROUPS` was set to `production` before 997dd6eaeb where this variable was removed.

machisuji: Thanks for having a look @cbliard. The variable substitution here is actually done by docker, not by bash. So the quotes don't matter. But you are right, that this variable seems to be empty looking at my output of the dockerization. So I propose to just remove that line altogether.

PR #13735, opened 2023-09-20
[#49977] PDF export (single work package): Include all attributes and fields according to the work package type form configuration

as-op:
For a single work package PDF Export this PR:

* displays attributes and custom fields according to the work package type form configuration in the attributes table
* refactors method for wp form type `spent_units` into format exporter
* adds format exporter  for `derived_remaining_hours`
* splits wp form type `date` into `start_date` `due_date` `duration`

https://community.openproject.org/work_packages/49977


HDinger (commented):
Hi @as-op 

I did not have a look at the code as that is not my field of expertise but just tested it. I noticed two things:

* All custom fields are shown with their ID instead of their name. (see green marks in the picture)
* For Custom Fields of type "long text", the value is shown as raw text in the table, and then again compiled below the table. Is that intended? (see blue marks in the picture)

Bildschirmfoto 2023-09-28 um 08 09 50

as-op:
Thanks a lot, @HDinger ! Both issues are not intended and are indeed bugs. I'll fix those.

PR #13751, opened 2023-09-21
Permission mocking

klaustopher:
Allows nicely mocking permissions for users

### DIFF
--- a/app/services/authorization/user_permissible_service.rb
+++ b/app/services/authorization/user_permissible_service.rb
@@ -1,16 +1,4 @@
 module Authorization
-  class UnknownPermissionError < StandardError
-    def initialize(permission_name)
-      super("Tried to check permission #{permission_name} that is not defined as a valid permission. It will never return true")
-    end
-  end
-
-  class IllegalPermissionCheck < StandardError
-    def initialize(permission, permissions, context)
-      super("Tried to check permission #{permission} which maps to #{permissions.map(&:name).join(', ')} in #{context} context. Permissible contexts for this permission are: #{permissions.flat_map(&:permissible_on).uniq.join(', ')}.")
-    end
-  end
-
   class UserPermissibleService
     attr_accessor :user
 
@@ -65,19 +53,37 @@ module Authorization
       AllowedInAnyEntityQuery.new(user:, permissions: perms, entity_class:, in_project:).exists?
     end
 
-    def self.permissions_for(action)
-      return [action] if action.is_a?(OpenProject::AccessControl::Permission)
-      return action if action.is_a?(Array) && action.all? { |a| a.is_a?(OpenProject::AccessControl::Permission) }
+    class << self
+      def permissions_for(action) # rubocop:disable Metrics/PerceivedComplexity
+        return [action] if action.is_a?(OpenProject::AccessControl::Permission)
+        return action if action.is_a?(Array) && action.all?(OpenProject::AccessControl::Permission)
 
-      if action.is_a?(Hash)
-        if action[:controller]&.to_s&.starts_with?('/')
-          action = action.dup
-          action[:controller] = action[:controller][1..]
+        if action.is_a?(Hash)
+          if action[:controller]&.to_s&.starts_with?('/')
+            action = action.dup
+            action[:controller] = action[:controller][1..]
+          end
+
+          OpenProject::AccessControl.allow_actions(action)
+        else
+          [OpenProject::AccessControl.permission(action)].compact
+        end
+      end
+
+      def normalized_permissions(permission, context, raise_on_unknown: false)
+        perms = permissions_for(permission)
+
+        if perms.blank?
+          Rails.logger.warn "Used permission \"#{permission}\" that is not defined. It will never return true."
+          raise UnknownPermissionError.new(permission) if raise_on_unknown
+
+          return []
         end
 
-        OpenProject::AccessControl.allow_actions(action)
-      else
-        [OpenProject::AccessControl.permission(action)].compact
+        context_perms = perms.select { |p| p.permissible_on?(context) }
+        raise IllegalPermissionContextError.new(permission, perms, context) if context_perms.blank?
+
+        context_perms
       end
     end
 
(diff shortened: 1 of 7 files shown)


app/services/authorization/user_permissible_service.rb
```diff
@@ -65,19 +53,37 @@ def allowed_in_any_entity?(permission, entity_class, in_project: nil)
...
-      return [action] if action.is_a?(OpenProject::AccessControl::Permission)
-      return action if action.is_a?(Array) && action.all? { |a| a.is_a?(OpenProject::AccessControl::Permission) }
+    class << self
+      def permissions_for(action) # rubocop:disable Metrics/PerceivedComplexity
+        return [action] if action.is_a?(OpenProject::AccessControl::Permission)
+        return action if action.is_a?(Array) && action.all?(OpenProject::AccessControl::Permission)
+
+        if action.is_a?(Hash)
+          if action[:controller]&.to_s&.starts_with?('/')
+            action = action.dup
```
aaron-contreras, line 63: 💭 Curious: Why do we `#dup` the action? I don't see it being used anywhere downstream again within the `.permissions_for` method. Is it preventing something?

spec/support/mocked_permission_helper.rb
```diff
@@ -0,0 +1,137 @@
...
+# You should have received a copy of the GNU General Public License
+# along with this program; if not, write to the Free Software
+# Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+#
+# See COPYRIGHT and LICENSE files for more details.
+#++
+
+# Provides a nice DSL to mock permissions on the user. It also validates that the permissions are
+# mocked in the right context. So when you try to mock a project permission globally, it will complain.
+# For examples on usage, see spec/support_spec/mocked_permission_helper_spec.rb
```
aaron-contreras, line 31: 🖌️

app/services/authorization/user_permissible_service.rb
klaustopher: It's copied over from the other implementations of that method. As we don't know where the hash is coming from, it's best practice to either `dup` it or extract keys from it with i.e. `slice` or `except` **if we want to modifiy it internally**. Otherwise the changes we make would also change the hash outside of the method. And that's not an intended side effect here.

Think of something like:

```ruby
link_params = { controller: 'foo', action: 'index' }
if user.allowed_in_project(link_params, @project)
  link_to link_params
end
```

If our method did not `dup` the hash, our changes would also change what gets passed into `link_to`

spec/support/mocked_permission_helper.rb
```diff
@@ -0,0 +1,137 @@
...
+
+    # Permission is allowed on any entity, if
+    #   - filtering for one specific project, when
+    #     - the permission has been given to that project
+    #     - the permission has been given to any work package belonging to that project
+    #   - NOT filtering for one specific project, when
+    #     - the permission has been given to any project
+    #     - the permission has been given to any entity
+    allow(permissible_service).to receive(:allowed_in_any_entity?) do |permission, entity_class, in_project:|
+      all_permitted_entties = permission_mock.permitted_entities
```
aaron-contreras:
```suggestion
      all_permitted_entities = permission_mock.permitted_entities
```

spec/support/mocked_permission_helper.rb
```diff
@@ -0,0 +1,137 @@
...
+
+      next true if in_project && all_permitted_entties[in_project].include?(permission)
```
aaron-contreras:
```suggestion
      next true if in_project && all_permitted_entities[in_project].include?(permission)
```

spec/support/mocked_permission_helper.rb
```diff
@@ -0,0 +1,137 @@
...
+
+      filtered_entities = if in_project
+                            all_permitted_entties.select do |k, _|
```
aaron-contreras:
```suggestion
                            all_permitted_entities.select do |k, _|
```

spec/support/mocked_permission_helper.rb
```diff
@@ -0,0 +1,137 @@
...
+                              k.is_a?(entity_class) && k.respond_to?(:project) && k.project == in_project
+                            end
+                          else
+                            all_permitted_entties.select { |k, _| k.is_a?(entity_class) || k.is_a?(Project) }
```
aaron-contreras:
```suggestion
                            all_permitted_entities.select { |k, _| k.is_a?(entity_class) || k.is_a?(Project) }
```

spec/support_spec/mocked_permission_helper_spec.rb
```diff
@@ -0,0 +1,170 @@
...
+      expect(user).not_to be_allowed_in_work_package(:view_work_packages, other_work_package)
+    end
+  end
+
+  context 'when mocking a permission in the work package' do
+    before do
+      mock_permissions_for(user) do |mock|
+        mock.in_work_package :view_work_packages, work_package: work_package_in_project
+        mock.in_work_package :view_work_packages, :edit_work_packages, work_package: other_work_package_in_project
+      end
```
aaron-contreras, line 131: Love the DSL ❤️

spec/support_spec/mocked_permission_helper_spec.rb
```diff
@@ -0,0 +1,170 @@
...
+# You should have received a copy of the GNU General Public License
+# along with this program; if not, write to the Free Software
+# Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+#
+# See COPYRIGHT and LICENSE files for more details.
+#++
+
+require 'spec_helper'
+
+RSpec.describe MockedPermissionHelper do
```
aaron-contreras, line 31: I'd add some association defaults since we're creating work packages for every example and there will be some factory cascading otherwise
 
```suggestion
RSpec.describe MockedPermissionHelper do
  shared_association_default(:author, factory_name: :user) { create(:user) }
  shared_association_default(:priority) { create(:priority) }
  shared_association_default(:status) { create(:status) }
```

(These should work)

app/services/authorization/user_permissible_service.rb
aaron-contreras: Thank you 👍
Makes sense

spec/support_spec/mocked_permission_helper_spec.rb
klaustopher: I switched to `build` strategy, as the mocking is explicitly for the cases where we only build :)

aaron-contreras (approved):

PR #13754, opened 2023-09-21
[#49113] ensure to display file links for one drive storages

Kharonus:
- https://community.openproject.org/wp/49113
- refactored errors returned by connection manager to be storage errors
- added source indicator to storage error data
- removed nextcloud specific method names from sync service
- check all three states of the file link list
  - not logged in
  - with auth error
  - happy case

### DIFF
--- a/modules/storages/app/common/storages/peripherals/storage_interaction/one_drive/files_info_query.rb
+++ b/modules/storages/app/common/storages/peripherals/storage_interaction/one_drive/files_info_query.rb
@@ -55,7 +55,14 @@ module Storages
             end
 
             result = file_ids.map do |file_id|
-              wrap_storage_file_error(file_id, FileInfoQuery.call(storage: @storage, user:, file_id:))
+              file_info_result = FileInfoQuery.call(storage: @storage, user:, file_id:)
+              if file_info_result.failure? &&
+                file_info_result.errors.data.source.is_a?(::OAuthClients::ConnectionManager)
+                # errors in the connection manager must short circuit the query and return the error
+                return file_info_result
+              end
+
+              wrap_storage_file_error(file_id, file_info_result)
             end
 
             ServiceResult.success(result:)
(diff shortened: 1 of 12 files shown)


modules/storages/app/common/storages/peripherals/storage_interaction/one_drive/files_info_query.rb
```diff
@@ -55,7 +55,14 @@ def call(user:, file_ids:)
             end
 
             result = file_ids.map do |file_id|
-              wrap_storage_file_error(file_id, FileInfoQuery.call(storage: @storage, user:, file_id:))
+              file_info_result = FileInfoQuery.call(storage: @storage, user:, file_id:)
+              if file_info_result.failure? &&
+                file_info_result.errors.data.source.is_a?(::OAuthClients::ConnectionManager)
```
mereghost: 🟡 What do you think about adding a method `error_source` to the `ServiceResult` that does that digging around or returns something to represent an `UnknownSource` (just not `nil` :P)

Right now, we assume a lot about the internals of another object here.

modules/storages/app/common/storages/peripherals/storage_interaction/one_drive/files_info_query.rb
```diff
@@ -70,7 +77,7 @@ def wrap_storage_file_error(file_id, query_result)
               storage_error = query_result.errors
               StorageFileInfo.new(
                 id: file_id,
-                status: storage_error.data.dig(:error, :code),
+                status: storage_error.data.payload.dig(:error, :code),
```
mereghost: 🟡 Same as above. Maybe we could add this to our refinements.

mereghost (approved):
Overall looks great, just pointed out some "nice to have" improvements that don't need to be addressed for a merge.

modules/storages/app/common/storages/peripherals/storage_interaction/one_drive/files_info_query.rb
Kharonus: added suggestion

PR #13756, opened 2023-09-21
Fix incorrect execution order of `OpenProject::AccessControl` spec setup

aaron-contreras:
Given that `:all_former_permissions` and
`:former_repository_permissions` were memoized with a `let!` construct their execution was happening after the

```
described_class.remove_modules_permissions(:repository)
```

call in the around hook. Therefore `:all_former_permissions` would lack the repository permissions when resetting it back after `example.run` in `reset_former_permissions_and_clear_caches`.

This caused other specs that required repository permissions down the line in the test suite to fail due to there being no permissions for the `:repository` project module. (and false positives for this spec file).

This order dependency was detected and reproducible by the following `rspec --bisect`s output

```
rspec './spec/lib/open_project/access_control_spec.rb[1:1:3]' \
'./spec/requests/api/v3/repositories/revisions_by_work_package_resource_spec.rb[1:1:3:2:1:2]' \
 --seed 44936
```

### DIFF
--- a/spec/lib/open_project/access_control_spec.rb
+++ b/spec/lib/open_project/access_control_spec.rb
@@ -84,10 +84,10 @@ RSpec.describe OpenProject::AccessControl do
   end
 
   describe '.remove_modules_permissions' do
-    let!(:all_former_permissions) { described_class.permissions }
-    let!(:former_repository_permissions) do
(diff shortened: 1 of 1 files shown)


cbliard (approved):
:ok_hand:

PR #13770, opened 2023-09-22
Do not set statement timeout if already set to 0

cbliard:
A statement_timeout of 0 disables the statement timeout. So when setting the minimum to `15min`, it should do nothing as `15min` is less than infinite.
The reverse is true: when setting the minimum statement timeout to `0`, it should always apply it (unless it is already 0).

### DIFF
--- a/lib_static/open_project/migration_statement_timeout/manager.rb
+++ b/lib_static/open_project/migration_statement_timeout/manager.rb
@@ -30,6 +30,74 @@
 
 module MigrationStatementTimeout
   module Manager
+    class SetMinimumStatementTimeout
+      attr_reader :conn
+
+      def initialize(conn, migration)
+        @conn = conn
+        @migration = migration
+      end
+
+      delegate :say, to: :@migration
+
+      def call(min_timeout)
+        if same_timeout?(min_timeout)
+          say "ignore set statement_timeout to #{min_timeout}: " \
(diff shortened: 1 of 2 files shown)


lib_static/open_project/migration_statement_timeout/manager.rb
```diff
@@ -30,6 +30,74 @@
...
+        when /\A\d+(ms)?\z/
+          timeout.to_i
+        when /\A\d+s\z/
+          timeout.to_i * 1000
+        when /\A\d+min\z/
+          timeout.to_i * 1000 * 60
+        when /\A\d+h\z/
+          timeout.to_i * 1000 * 60 * 60
+        else
+          raise "Unrecognized statement timeout duration #{timeout.inspect}"
```
m-graf: Maybe an `ArgumentError` instead of a generic `raise`?

lib_static/open_project/migration_statement_timeout/manager.rb
```diff
@@ -30,6 +30,74 @@
 
 module MigrationStatementTimeout
   module Manager
+    class SetMinimumStatementTimeout
```
m-graf, line 33: Very cool, I like this

m-graf (commented):
Looks good!

oliverguenther (approved):

PR #13788, opened 2023-09-26
fix polish translation for deleted attachment journal

machisuji:
WP [#50197](https://community.openproject.org/projects/openproject/work_packages/50197/activity)


machisuji:
Lesson learned: This has to be corrected in crowdin otherwise the change will be overriden. Closing the PR.

PR #13800, opened 2023-09-27
Implement the UploadLinkQuery for OneDrive

mereghost:
##### Related WorkPackage: [OP#49126](https://community.openproject.org/projects/openproject/work_packages/49126)

Pretty much what the title says. 
Nothing new at this point.


modules/storages/app/common/storages/peripherals/storage_interaction/one_drive/upload_link_query.rb
```diff
@@ -0,0 +1,86 @@
...
+            @storage = storage
+            @uri = storage.uri
+          end
+
+          def self.call(storage:, user:, data:)
+            new(storage).call(user:, data:)
+          end
+
+          def call(user:, data:)
+            folder, filename = data.slice(:parent, :fileName).values
```
ba1ash: Should we switch to snake_case in key names or camelCase is more appropriate here?

ba1ash: And what do you think about using `data.fetch` instead of slicing? `#fetch` raises an error in case of key absence which is useful, because both keys are mandatory as I see.

mereghost: That's a good question. I just went with what the API endpoint gets and never thought about it. 🤔

mereghost: it shouldn't be an issue as this is passed from `API::V3::StorageFiles::StorageFilesAPI`, so the data is already sort of validated.

What I did miss is that the keys are strings. 🤦🏼

Kharonus: I think, we have the convention: in ruby variables and keys are written in snake_case. camelCase is only relevant when the data is represented in the API response. Looking at the code, it should be possible to change this, as the data goes through some validation logic [here](https://github.com/opf/openproject/blob/1a9d031a77bbb2e067ec2e6a7429e5f846cfe169/modules/storages/lib/api/v3/storage_files/storage_files_api.rb#L66-L74).

Kharonus (approved):
LGTM. I'd like to solve that casing question from pascal before merge, but from code perspective I have no objections.

modules/storages/app/common/storages/peripherals/storage_interaction/one_drive/upload_link_query.rb
ba1ash: For me it is fine.

ba1ash (approved):

PR #13803, opened 2023-09-27
[#50219] Custom export cover background overlay color won't return to…

as-op:
… the default one after deleting

This PR:
* allows blank parameter to remove the stored color
* adds a test for this

https://community.openproject.org/work_packages/50219


dombesz (approved):
LGTM👍

PR #13683, opened 2023-09-11
Fix/improve project allowed to scope performance

ulferts:
Reworks the SQL statement underlying

`Project.allowed_to?(some_user, :some_permission)`

that showed poor performance in some datasets. 

### Background

The problem manifests in statements where the aforementioned statement takes part as a subquery in combination with a `LIMIT`, as e.g. in 

```sql

SELECT "work_packages"."id"                      AS t0_r0,
       "work_packages"."type_id"                 AS t0_r1,
       "work_packages"."project_id"              AS t0_r2,
       "work_packages"."subject"                 AS t0_r3,
       "work_packages"."description"             AS t0_r4,
       "work_packages"."due_date"                AS t0_r5,
       "work_packages"."category_id"             AS t0_r6,
       "work_packages"."status_id"               AS t0_r7,
       "work_packages"."assigned_to_id"          AS t0_r8,
       "work_packages"."priority_id"             AS t0_r9,
       "work_packages"."version_id"              AS t0_r10,
       "work_packages"."author_id"               AS t0_r11,
       "work_packages"."lock_version"            AS t0_r12,
       "work_packages"."done_ratio"              AS t0_r13,
       "work_packages"."estimated_hours"         AS t0_r14,
       "work_packages"."created_at"              AS t0_r15,
       "work_packages"."updated_at"              AS t0_r16,
       "work_packages"."start_date"              AS t0_r17,
       "work_packages"."responsible_id"          AS t0_r18,
       "work_packages"."position"                AS t0_r19,
       "work_packages"."story_points"            AS t0_r20,
       "work_packages"."remaining_hours"         AS t0_r21,
       "work_packages"."budget_id"               AS t0_r22,
       "work_packages"."derived_estimated_hours" AS t0_r23,
       "work_packages"."schedule_manually"       AS t0_r24,
       "work_packages"."parent_id"               AS t0_r25,
       "work_packages"."duration"                AS t0_r26,
       "work_packages"."ignore_non_working_days" AS t0_r27,
       "work_packages"."derived_remaining_hours" AS t0_r28,
       "projects"."id"                           AS t1_r0,
       "projects"."name"                         AS t1_r1,
       "projects"."description"                  AS t1_r2,
       "projects"."public"                       AS t1_r3,
       "projects"."parent_id"                    AS t1_r4,
       "projects"."created_at"                   AS t1_r5,
       "projects"."updated_at"                   AS t1_r6,
       "projects"."identifier"                   AS t1_r7,
[...]

PR #13807, opened 2023-09-27
Permission Rework [1/x]: `.allowed_to` scope for WorkPackages

klaustopher:
Based on the work that @ulferts did in #13683, this PR adds a similiar `.allowed_to` scope to the `WorkPackage` model. It already optimizes by also including WorkPackages if

- the requested permission is allowed on the WorkPackage
- the requested permission is allowed on the Project the Work Package belongs to

Also, I have extracted the `allowed_to_permissions` method out of the scope into the `Authorization` module. We will be using the method a lot more in the future, so there's a good place to park it.

Also, we added the `Authorization.contextual_permissions` method that allows filtering permissions for a given context. This helps us when checking permissions or mocking permissions, that we don't accidentally try to check i.e. a global permission (like `:create_user` in a project context).

--- 

### DIFF
--- a/app/models/work_packages/scopes/allowed_to.rb
+++ b/app/models/work_packages/scopes/allowed_to.rb
@@ -26,64 +26,75 @@
 # See COPYRIGHT and LICENSE files for more details.
 # ++
 
-module Projects::Scopes
+module WorkPackages::Scopes
   module AllowedTo
     extend ActiveSupport::Concern
 
     class_methods do
-      # Returns an ActiveRecord::Relation to find all projects for which
-      # +user+ has the given +permission+
-      def allowed_to(user, permission)
-        permissions = Authorization.contextual_permissions(permission, :project, raise_on_unknown: true)
+      # Returns an ActiveRecord::Relation to find all work packages for which
+      # +user+ has the given +permission+ either directly on the work package
+      # or by the linked project
+      def allowed_to(user, permission) # rubocop:disable Metrics/AbcSize
+        permissions = Authorization.contextual_permissions(permission, :work_package, raise_on_unknown: true)
 
         if user.admin? && permissions.all?(&:grant_to_admin?)
           where(id: allowed_to_admin_relation(permissions))
         elsif user.anonymous?
           where(id: allowed_to_non_member_relation(user, permissions))
         else
-          where(arel_table[:id].in(allowed_to_member_relation(user, permissions)
-                                     .arel
-                                     .union(:all, allowed_to_non_member_relation(user, permissions).arel)))
+          union = Arel::Nodes::UnionAll.new(
+            Arel::Nodes::UnionAll.new(
+              allowed_to_member_relation(user, permissions).select(arel_table[:id]).arel,
+              allowed_to_non_member_relation(user, permissions).select(arel_table[:id]).arel
+            ),
+            where(project_id: Project.allowed_to(user, permission).select(:id)).select(arel_table[:id]).arel
+          )
+
+          where(arel_table[:id].in(union))
         end
       end
 
       private
 
       def allowed_to_admin_relation(permissions)
-        joins(allowed_to_enabled_module_join(permissions))
-          .where(arel_table[:active].eq(true))
+        joins(:project)
+        .joins(allowed_to_enabled_module_join(permissions))
+          .where(Project.arel_table[:active].eq(true))
       end
 
       def allowed_to_non_member_relation(user, permissions)
-        joins(allowed_to_enabled_module_join(permissions))
+        joins(:project)
+          .joins(allowed_to_enabled_module_join(permissions))
           .joins(allowed_to_builtin_roles_in_active_project_join(user))
           .joins(allowed_to_role_permission_join(permissions))
-          .select(:id)
+          .select(arel_table[:id])
       end
 
       def allowed_to_member_relation(user, permissions)
         Member
           .joins(allowed_to_member_in_active_project_join(user))
+          .joins(allowed_to_member_in_work_package_join)
           .joins(allowed_to_enabled_module_join(permissions))
           .joins(:roles, :member_roles)
           .joins(allowed_to_role_permission_join(permissions))
-          .select('projects.id')
+          .select(arel_table[:id])
       end
 
-      def allowed_to_enabled_module_join(permissions)
+      def allowed_to_enabled_module_join(permissions) # rubocop:disable Metrics/AbcSize
         project_module = permissions.filter_map(&:project_module).uniq
         enabled_module_table = EnabledModule.arel_table
+        projects_table = Project.arel_table
 
         if project_module.any?
           arel_table.join(enabled_module_table, Arel::Nodes::InnerJoin)
-                    .on(arel_table[:id].eq(enabled_module_table[:project_id])
+                    .on(projects_table[:id].eq(enabled_module_table[:project_id])
                           .and(enabled_module_table[:name].in(project_module))
-                          .and(arel_table[:active].eq(true)))
+                          .and(projects_table[:active].eq(true)))
                     .join_sources
         end
       end
 
-      def allowed_to_role_permission_join(permissions)
+      def allowed_to_role_permission_join(permissions) # rubocop:disable Metrics/AbcSize
         return if permissions.all?(&:public?)
 
         role_permissions_table = RolePermission.arel_table
@@ -110,10 +121,9 @@ module Projects::Scopes
       def allowed_to_members_condition(user)
         members_table = Member.arel_table
 
-        members_table[:project_id].eq(arel_table[:id])
+        members_table[:project_id].eq(arel_table[:project_id])
                                   .and(members_table[:user_id].eq(user.id))
-                                  .and(members_table[:entity_type].eq(nil))
-                                  .and(members_table[:entity_id].eq(nil))
+                                  .and(members_table[:entity_type].eq(model_name.name))
       end
 
       def allowed_to_builtin_roles_in_active_project_join(user)
@@ -131,13 +141,23 @@ module Projects::Scopes
       end
 
       def allowed_to_member_in_active_project_join(user)
-        arel_table.join(arel_table)
-                  .on(arel_table[:active].eq(true)
-                                         .and(allowed_to_members_condition(user)))
+        Project.arel_table
+          .join(Project.arel_table)
+                  .on(Project.arel_table[:active].eq(true)
+                   .and(allowed_to_members_condition(user)))
                   .join_sources
       end
 
+      def allowed_to_member_in_work_package_join
+        members_table = Member.arel_table
+        arel_table.join(arel_table)
+        .on(members_table[:entity_id].eq(arel_table[:id]))
+        .join_sources
+      end
+
       def allowed_to_built_roles_in_active_project_condition(user)
+        projects_table = Project.arel_table
+
         builtin = if user.logged?
                     Role::BUILTIN_NON_MEMBER
                   else
@@ -147,8 +167,8 @@ module Projects::Scopes
         roles_table = Role.arel_table
 
         roles_table[:builtin].eq(builtin)
-                             .and(arel_table[:active])
-                             .and(arel_table[:public])
+                             .and(projects_table[:active])
+                             .and(projects_table[:public])
       end
 
       def allowed_to_no_member_exists_condition(user)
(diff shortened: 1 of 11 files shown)


cbliard (approved):
Nice work!

lib/open_project/access_control/permission.rb
```diff
@@ -72,15 +73,19 @@ def public?
...
+        permissible_on? :project
       end
 
       def global?
-        @permissible_on.include? :global
+        permissible_on? :global
+      end
+
+      def permissible_on?(context_type)
+        @permissible_on.include?(context_type)
```
ulferts, line 88: Nice little dry-ing up.

app/models/work_packages/scopes/allowed_to.rb
```diff
@@ -0,0 +1,184 @@
...
+          .join(Project.arel_table)
+                  .on(Project.arel_table[:active].eq(true)
+                   .and(allowed_to_members_condition(user)))
+                  .join_sources
+      end
+
+      def allowed_to_member_in_work_package_join
+        members_table = Member.arel_table
+        arel_table.join(arel_table)
+        .on(members_table[:entity_id].eq(arel_table[:id]))
```
ulferts: ~~As long as there is only one `entity` we are dealing with but as soon as there can be any other value for `entity_type` and by that duplicate `ids`, this will break. So adding another condition on the `entity_type` needs to be added.~~

~~Edit 1: In the context it is used, there is another join which secures this condition by already having a check for the `entity_type`. I personally would still add it to the condition just to for the risk it poses but it is not strictly necessary.~~

Edit2: Now with the statement this is used in needing to be reworked, I'd definitely put the check for the `entity_type` in.

app/models/work_packages/scopes/allowed_to.rb
```diff
@@ -0,0 +1,184 @@
...
+
+        if user.admin? && permissions.all?(&:grant_to_admin?)
+          where(id: allowed_to_admin_relation(permissions))
+        elsif user.anonymous?
+          where(id: allowed_to_non_member_relation(user, permissions))
+        else
+          union = Arel::Nodes::UnionAll.new(
+            Arel::Nodes::UnionAll.new(
+              allowed_to_member_relation(user, permissions).select(arel_table[:id]).arel,
+              allowed_to_non_member_relation(user, permissions).select(arel_table[:id]).arel
```
ulferts: This part of the query isn't necessary. 

It covers the case where a work package is in a project that is public and the non member role has the necessary permission to grant. In other words, it will return the id of all work packages in a public project where the user has no membership in but it will do so only if the non member role has the permission.

That part is already covered by the `Project.allowed_to` which itself includes that part in its statement. It will return the id of all projects where the user is member with the permissions OR those public projects where the user has no membership if the non member role has the permission.

app/models/work_packages/scopes/allowed_to.rb
```diff
@@ -0,0 +1,184 @@
...
+        permissions = Authorization.contextual_permissions(permission, :work_package, raise_on_unknown: true)
+
+        if user.admin? && permissions.all?(&:grant_to_admin?)
+          where(id: allowed_to_admin_relation(permissions))
+        elsif user.anonymous?
+          where(id: allowed_to_non_member_relation(user, permissions))
+        else
+          union = Arel::Nodes::UnionAll.new(
+            Arel::Nodes::UnionAll.new(
+              allowed_to_member_relation(user, permissions).select(arel_table[:id]).arel,
```
ulferts: There is a very subtle bug in this part of the statement. Abbreviated the statement looks like this:

```sql
SELECT "work_packages".*
FROM "work_packages"
WHERE "work_packages"."id" IN ((
	(SELECT "work_packages"."id"
		FROM "members"
		INNER JOIN "projects" ON "projects"."active" = TRUE
		AND "members"."project_id" = "work_packages"."project_id"
		AND "members"."user_id" = 5
		AND "members"."entity_type" = 'WorkPackage'
		INNER JOIN "work_packages" ON "members"."entity_id" = "work_packages"."id"
		INNER JOIN "member_roles" ON "member_roles"."member_id" = "members"."id"
		INNER JOIN "roles" ON "roles"."id" = "member_roles"."role_id"
		INNER JOIN "member_roles" "member_roles_members" ON "member_roles_members"."member_id" = "members"."id"
		INNER JOIN "role_permissions" ON "roles"."id" = "role_permissions"."role_id"
		AND (FALSE
							OR "role_permissions"."permission" = 'view_work_packages'
							AND "enabled_modules"."name" = 'work_package_tracking'))
UNION ALL
  // work packages in public project
UNION ALL
  // work packages in public project or in projects in which the user has the necessary permission via a role
))
```

Or further reduced to the interesting part

```sql
SELECT "work_packages".*
FROM "work_packages"
WHERE "work_packages"."id" IN ((
	(SELECT "work_packages"."id"
		FROM "members"
		INNER JOIN "projects" ON "projects"."active" = TRUE
		AND "members"."project_id" = "work_packages"."project_id"
		// some more conditions
		INNER JOIN "work_packages" ON "members"."entity_id" = "work_packages"."id"
		// some more joins
UNION ALL
  // work packages in public project
UNION ALL
  // work packages in public project or in projects in which the user has the necessary permission via a role
))
```

Please notice that the `AND "members"."project_id" = "work_packages"."project_id"` condition is before the `INNER JOIN "work_packages"` in that first part of the UNIONed id collection. 

Because the `work_packages` table hasn't been joined in the subquery yet, PostgreSQL will reference the work packages table in the outer statement  
```sql
SELECT "work_packages".*
FROM "work_packages"
```
and do ... something. It is just horribly slow on a larger dataset so I never managed to find out if the results are correct in the end.

Swapping the order fixes the problem (and performs way quicker). 

Swapping would lead to the following statement

```sql
(SELECT "work_packages"."id"
		FROM "members"
	    INNER JOIN "work_packages" ON "members"."entity_id" = "work_packages"."id" AND "members"."entity_type" = 'WorkPackage'
		INNER JOIN "projects" ON "projects"."active" = TRUE
		AND "members"."project_id" = "work_packages"."project_id"
		AND "members"."user_id" = 1445
		AND "members"."entity_type" = 'WorkPackage'		
		INNER JOIN "member_roles" ON "member_roles"."member_id" = "members"."id"
		INNER JOIN "roles" ON "roles"."id" = "member_roles"."role_id"
		INNER JOIN "member_roles" "member_roles_members" ON "member_roles_members"."member_id" = "members"."id"
		INNER JOIN "role_permissions" ON "roles"."id" = "role_permissions"."role_id"
		AND (FALSE
							OR "role_permissions"."permission" = 'view_work_packages'
							AND "enabled_modules"."name" = 'work_package_tracking'))
```

Given the `project_id` column on the `members` table this can be simplified. There is also an unnecessary second join with member_roles in here `INNER JOIN "member_roles" "member_roles_members" ON "member_roles_members"."member_id" = "members"."id"` That can be removed. The result could look like this:

```sql
(SELECT "members".entity_id
    FROM members	    
		INNER JOIN "projects" ON "projects"."active" = TRUE
		AND "members"."project_id" = "projects"."id"
		AND "members"."user_id" = 1445
		AND "members"."entity_type" = 'WorkPackage'		
		INNER JOIN "member_roles" ON "member_roles"."member_id" = "members"."id"
		INNER JOIN "roles" ON "roles"."id" = "member_roles"."role_id"
		INNER JOIN "role_permissions" ON "roles"."id" = "role_permissions"."role_id"
		AND (FALSE
							OR "role_permissions"."permission" = 'view_work_packages'
							AND "enabled_modules"."name" = 'work_package_tracking'))
```							

That unnecessary second join with member_roles `INNER JOIN "member_roles" "member_roles_members" ON "member_roles_members"."member_id" = "members"."id"` seems to have crept in from the `Project.allowed_to` scope implementation and can be removed there as well.

I don't know though whether removing the join with the `work_packages` table out of the statement is worth that much. Performance-wise, it did not make a difference in my tests. So if the statement for fetching the allowed WorkPackages is the only (or the most important) use case for populating the `project_id` column for work package members, then we should rethink whether this is worth the effort of maintaining the column (e.g. when moving a work package between projects).

app/models/work_packages/scopes/allowed_to.rb
```diff
@@ -0,0 +1,184 @@
...
+
+      def allowed_to_no_member_exists_condition(user)
+        Member
+          .select(1)
+          .where(allowed_to_members_condition(user))
+          .arel
+          .exists
+          .not
+      end
+    end
```
ulferts, line 69: This copies a lot of the code from the `Projects::AllowedTo` scope. We'll have to see how much of that is still there after the rework on the work package members check but I suspect that there is still a substantial overlap. Given the complexity and similarity of the use case of the two scopes, let's extract what they share into a module.

app/services/authorization.rb
```diff
@@ -36,7 +36,12 @@ def users(action, project)
 
   # Returns all projects a user has a certain permission in
   def projects(action, user)
-    Project.allowed_to(action, user)
+    Project.allowed_to(user, action)
```
ulferts, line 39: Very good catch 👍 Thank you.

spec/models/work_packages/scopes/allowed_to_spec.rb
```diff
@@ -0,0 +1,138 @@
...
+    context 'and requesting a permission that is granted on the project and the work package' do
+      subject { described_class.allowed_to(user, :view_work_packages) }
+
+      it 'returns the authorized work packages' do
+        expect(subject).to contain_exactly(work_package_in_private_project, other_work_package_in_private_project)
+      end
+    end
+  end
+
+  # TODO: Add more tests that check anonymous and non-member permissions
```
ulferts: What you said 😄 Maybe the brunt of the tests for when there is a permission on a project membership can be copied over from the specs on `Project.allowed_to`.

ulferts (changes requested):
Definitely heading into the right direction @klaustopher and some nice improvements in there already. There are a couple of things though that need to be addressed.

app/models/work_packages/scopes/allowed_to.rb
```diff
@@ -0,0 +1,184 @@
...
+
+    class_methods do
+      # Returns an ActiveRecord::Relation to find all work packages for which
+      # +user+ has the given +permission+ either directly on the work package
+      # or by the linked project
+      def allowed_to(user, permission) # rubocop:disable Metrics/AbcSize
+        permissions = Authorization.contextual_permissions(permission, :work_package, raise_on_unknown: true)
+
+        if user.admin? && permissions.all?(&:grant_to_admin?)
+          where(id: allowed_to_admin_relation(permissions))
```
ulferts: This is a rather simple statement so it is ok to leave around. But there would also be the option to remove it completely for the sake of just using `where(project_id: Project.allowed_to)`

app/models/work_packages/scopes/allowed_to.rb
```diff
@@ -0,0 +1,184 @@
...
+        elsif user.anonymous?
+          where(id: allowed_to_non_member_relation(user, permissions))
```
ulferts: With the ` allowed_to_non_member_relation` implementation not being needed in the case below, it might be simpler to just use `where(project_id: Project.allowed_to)` here as well. 

In total, this might lead to the whole implementation becoming way simpler as there is then only the case of a non admin logged in user where the work package member specific case needs to be added on the one hand, and then there are the rest of the cases where only `Project.allowed_to` is used.

app/models/work_packages/scopes/allowed_to.rb
klaustopher: I was unsure here. Good to know that this can be removed 👍

app/models/work_packages/scopes/allowed_to.rb
ulferts: I have been thinking about this some more. Maybe the better approach is to introduce a scope on Member which allows filtering for all members that grant a permission in some context (project, global, entity). This would exclude the case for non members/anonymous of course but those are rather project specific currently so it might be fine. That scope could then be reused for other entities as well. So ideally, an entity specific scope would just have to be combined out of scopes from the project and members.

app/models/work_packages/scopes/allowed_to.rb
klaustopher: done

app/models/work_packages/scopes/allowed_to.rb
klaustopher: done

PR #13809, opened 2023-09-27
Permission Rework [2/x]: Extract all permission methods from `User` model and implement a new Permission Query Service

klaustopher:
In this PR, we are extracting out all the `allowed_to...?` methods from the `User` model into a concern. This will make it easier in the future to remove the deprecated methods. 

The deprecation warnings are currently commented out, as they will spam test runs, because we are still using the `allowed_to` methods everywhere. This will be adressed in a later PR, and when our code is not using any of the old method anymore, we will switch on the deprecation warning to warn 3rd party devs.

We also added a new query interface, that is being more straight forward with what permissions we are really wanting to check for:

| Permission check method | Description | example where this should be used |
|-------------------------|-------------|-----------------------------------|
| `user.allowed_globally?(:manage_user)` | **Only** checks for a global permission | Pages in administration to see if a user is allowed to manage a certain part of the application |
| `user.allowed_in_project?(:manage_members, project)` | Checks whether the user has the permission on the entire project | For project specific areas like managing members |
| `user.allowed_in_project?(:manage_members, [project1, project2])` | Checks whether the user has the permission on **all** given projects | An action that involves 2 projects (i.e. moving a WP) |
| `user.allowed_in_any_project?(:view_meetings)` | Checks whether the user has the permission in **any** project | To check if a global index page should be shown, where we have not introduced roles for the entity (i.e. meetings) |
| `user.allowed_in_work_package?(:view_work_packages, work_package)` | Checks whether the user has the permission in the specified work package **OR** in the project the WP belongs to | When displaying the work package page, to figure out if a certain section should be shown |
| `user.allowed_in_work_package?(:view_work_packages, [wp1, wp2])` | Checks whether the user has the permission in **all of** the specified work packages **OR** in the projects the WPs belong to | An action that involves multiple work packages (i.e. an export) |
| `user.allowed_in_any_work_package?(:view_work_packages)` | Checks whether the user has the permission in **any** work package | To check if a global index page should be shown, where we have introduced roles for the entity (i.e. work packages) |
| `user.allowed_in_any_work_package?(:view_work_packages, in_project: project)` | Checks whether the user has the permission in any work package **within the given project** | To check if we should show menu items in the Project's menu, where we expect to have users that will not have access to an entire project, but only single work packages |

--- 


app/services/authorization/user_permissible_service.rb
```diff
@@ -0,0 +1,138 @@
...
+      context = entity_class.model_name.element.to_sym
+      perms = Authorization.contextual_permissions(permission, context, raise_on_unknown: true)
+      return true if admin_and_all_granted_to_admin?(perms)
+
+      if in_project
+        WorkPackage.allowed_to(user, perms).exists?(project: in_project)
+      else
+        WorkPackage.allowed_to(user, perms).exists?
+      end
+    end
```
cbliard: Shouldn't it be using `entity_class` instead of assuming `WorkPackage`?

cbliard (approved):
I really like it :100:

app/services/authorization/user_permissible_service.rb
klaustopher: Absolutely!

app/models/user.rb
```diff
@@ -170,8 +171,6 @@ def update_password
 
   def reload(*args)
     @name = nil
```
ulferts: Not strictly part of this PR but with the removal of this line in #13829, the whole method can be removed.

app/models/users/permission_checks.rb
```diff
@@ -0,0 +1,118 @@
...
+
+  def reload(*args)
+    @user_permissible_service = nil
+    @user_allowed_service = nil
+    @project_role_cache = nil
+
+    super
+  end
+
+  # All the new methos to check for permissions. This will completely replace the old interface:
```
ulferts: typo: methos -> methods

app/services/authorization/user_permissible_service.rb
```diff
@@ -0,0 +1,138 @@
...
+        .where(members: { user_id: user.id, entity: nil, project: })
+        .pluck(:permission)
+        .map(&:to_sym)
+        .select { |permission| project.allows_to?(permission) }
+    end
+
+    def entity_permissions(entity)
+      RolePermission
+        .joins(role: { member_roles: :member })
+        .where(members: { user_id: user.id, entity:, project: entity.porject })
```
ulferts: typo: porject -> project

app/services/authorization/user_permissible_service.rb
```diff
@@ -0,0 +1,138 @@
...
+
+      return false if permissions_filtered_for_project.empty?
+      return true if admin_and_all_granted_to_admin?(permissions)
+
+      cached_permissions(project).intersect?(permissions_filtered_for_project.map(&:name))
+    end
+
+    def allowed_in_single_entity?(permissions, entity)
+      return false if entity.nil?
+
```
ulferts, line 93: The state (`active?`)  of the project needs to be checked here as well. Without the check, the case where an entity is in an archived project might return truthy.

app/services/authorization/user_permissible_service.rb
```diff
@@ -0,0 +1,138 @@
...
+        .map(&:to_sym)
+    end
+
+    def project_permissions(project)
+      RolePermission
+        .joins(role: { member_roles: :member })
+        .where(members: { user_id: user.id, entity: nil, project: })
+        .pluck(:permission)
+        .map(&:to_sym)
+        .select { |permission| project.allows_to?(permission) }
```
ulferts: Permissions granted to non members and to the anonymous user are not covered by this. The `Authorization::UserProjectRolesQuery` covers this and could be reused.

app/services/authorization/user_permissible_service.rb
```diff
@@ -0,0 +1,138 @@
...
+      context = entity_class.model_name.element.to_sym
+      perms = Authorization.contextual_permissions(permission, context, raise_on_unknown: true)
+      return true if admin_and_all_granted_to_admin?(perms)
+
+      if in_project
+        entity_class.allowed_to(user, perms).exists?(project: in_project)
+      else
+        entity_class.allowed_to(user, perms).exists?
+      end
+    end
```
ulferts, line 62: The former implementation treated `Authorization.roles(some_user)` to mean: "All roles in any project or globally". There are quite a lot of calls ala `user.allowed_to_globally?(:some_permission)` in the code base where :some_permission is e.g. :add_work_packages or :view_members which are project permissions. I get that we want to separate here between the two cases of having a global permission and having a project permission in any project. I just want to make sure that you are aware of this case and have the necessary changes handled in the subsequent PRs.

app/models/users/permission_checks.rb
```diff
@@ -0,0 +1,118 @@
...
+    @user_allowed_service ||= ::Authorization::UserAllowedService.new(self, role_cache: project_role_cache)
+  end
+
+  def user_permissible_service
+    @user_permissible_service ||= ::Authorization::UserPermissibleService.new(self)
+  end
+
+  def project_role_cache
+    @project_role_cache ||= ::Users::ProjectRoleCache.new(self)
+  end
```
ulferts, line 117: The `Authorization::UserPermissibleService` serves similar needs as the `Users::ProjectRoleCache`. Is the ProjectRoleCache to be replaced by the UserPermissibleService?

spec/services/authorization/user_permissible_service_spec.rb
```diff
@@ -0,0 +1,84 @@
...
+
+  describe '#allowed_in_any_entity?' do
+    context 'when asking for a permission that is not defined' do
+      let(:permission) { :not_defined }
+
+      it 'raises an error' do
+        expect { subject.allowed_in_any_entity?(permission, WorkPackage) }.to raise_error(Authorization::UnknownPermissionError)
+      end
+    end
+  end
```
ulferts, line 354: A whole lot of additional tests are required for those cases.

app/services/authorization/user_permissible_service.rb
```diff
@@ -0,0 +1,138 @@
+module Authorization
+  class UserPermissibleService
+    attr_accessor :user
+
+    def initialize(user)
+      @user = user
+      @cached_permissions = Hash.new do |hash, key|
+        hash[key] = Set.new
+      end
```
ulferts: If the checks for the current permission turn out to be more complicated (e.g. because of Non member and Anonymous), it would be an option to use a utility class inheriting from `Hash` that will do the fetching internally. Part of the implementation in this class could be moved into the utility class: 
* https://github.com/opf/openproject/pull/13809/files#diff-2c66bd35b7924325e768d995449b664d8a7828da69cde64deab84919bf3a1777R112-R136
* https://github.com/opf/openproject/pull/13809/files#diff-2c66bd35b7924325e768d995449b664d8a7828da69cde64deab84919bf3a1777R68-R74

ulferts (changes requested):
I like the direction this change is heading into. There are however a couple of amends that need to be done before the PR can be merged.

app/models/user.rb
klaustopher: done

app/services/authorization/user_permissible_service.rb
klaustopher: done

app/models/users/permission_checks.rb
klaustopher: The role cache will be removed with the old UserAllowedService

ulferts (approved):

PR #13810, opened 2023-09-27
Permission Rework [3/x]: Easier membership setup for factories

klaustopher:
We already have an option to generate memberships along in a principal factory. Unfortunately, it only allows to specify one role or one set of permissions that will be added to all memberships. 

Since we needed to extend the interface to build memberships on Work Packages, we had to touch the interface, I took inspiration from the `Project factory` that allows a hash interface for the memberships.

We can now do the same thing in the `principal` factory and all derivatives thereof:

```ruby
# when we have a role
role = create(:role, permissions: [:a, :b, :c]
create(:user, member_with_roles: { project => role })

# when we don't have a role, we can let the helper generate one for us:
create(:user, member_with_permissions: { project => [:a, :b, :c], work_package => [:c, :d, :e] })
```


cbliard (approved):
Apart from the need of this refactoring, this now looks a lot more cohesive to have project and roles/permissions together in a hash. Thanks for bringing this improvement.

spec/factories/principal_factory.rb
```diff
@@ -29,61 +29,62 @@
...
+      #   }
+      member_with_roles { {} }
 
-      unusable_attribute = %i[member_through_role member_with_permissions].find { |attr| evaluator.send(attr).present? }
-      next unless unusable_attribute
-
-      raise ArgumentError, "Cannot use :#{unusable_attribute} without :member_in_project or :member_in_projects " \
-                           "for #{principal.class} factory"
+      global_roles { [] }
+      global_permissions { [] }
```
ulferts, line 47: Ideally, the global roles/permissions would fall into the same interface, but then one would have to write something like: 

```
create(:user,
       member_with_permissions: {
          project => %w(view_work_packages),
          global: %w(add_project)
       })
```

As there is no actual object to create the member against. Any ideas?

ulferts (approved):
Nice improvement. 

I'd merge it but I don't know if that will lead to the previous PR getting more changes.

ulferts:
Come to think of it, the PR could also be changed to target `dev` and thus be merged quicker.

spec/factories/principal_factory.rb
klaustopher: Personally, I would not want to have it this way. You are not a member of "global". Yes, we handle it that way internally, but I'm not a huge fan of including the global roles and permissions inside of this hash.

PR #13812, opened 2023-09-27
Permission Rework [4/x]: Nice interface to mock permissions for stubbed users

klaustopher:
Currently, mocking the `allowed_to?` method on the user is an easy way to bypass the permission checking system. In the future we will have a lot more methods how we check permissions. I don't want the person writing tests to keep in mind if they need to mock `allowed_in_project` or `allowed_in_any_entity(..., in_project:)`.

The new interface can be called in a before filter and you specify what permissions you want to give the user in projects, work packages, or globally.

```ruby
mock_permissions_for(user) do |mock|
  mock.in_project :view_work_packages, project: some_project
  mock.in_work_package :edit_work_packages, work_package: some_wp
  mock.globally :create_users
end
```

Sometimes, we just want to allow everything and completely bypass the permission system. This is possible:

```ruby
mock_permissions_for(user) do |mock|
  mock.all_permissions_allowed!
end
```

If we want to deny every permission, we can do the following:

```ruby
mock_permissions_for(user)
# since we do not mock any permissions here, every permission check will return false!
```

As this might be a bit un-intuitive, there is this nicer way to write it:

```ruby
mock_permissions_for(user) do |mock|
  mock.forbid_everything
end
```

Internally this then correctly mocks all calls to `allow_globally?`, `allow_in_project?`, `allow_in_any_project?`, `allow_in_work_package?`, `allow_in_any_work_package?` and also in the legacy interface `allowed_to?`.

One more hint: The calls to `mock_permissions_for` **do not stack**. So only the mocks defined in the latest run will be used!

To see some useful usages, check `spec/support_spec/mocked_permission_helper_spec.rb`

--- 


aaron-contreras (commented):
I reviewed most of this in the original PR, awesome job; I really like the DSL. I don't recall the `#all_permissions_allowed!` and `#forbid_everything!` from the previous PR though. How about having those two methods more similarly named given they are just the inverse of each other? It would be easier to remember and keeps the naming at the same abstraction level that `#forbid_everything!` does

e.g.

Instead of
```ruby
mock.allow_all_permissions! # to allow everything
mock.forbid_everything! # to forbid everything
```

we have

```ruby
mock.allow_everything! # to allow everything
mock.forbid_everything! # to forbid everything
```

**NOTE**: Do let me know if you like this idea. I can take care of replacing throughout the PR if so.

spec/features/menu_items/top_menu_item_spec.rb
```diff
@@ -120,11 +126,12 @@ def click_link_in_open_menu(title)
 
     context 'as a regular user' do
       it 'only displays projects, activity and news' do
+        skip "Figure out why the news item is not shown here any more"
```
aaron-contreras: This will let you know once it starts showing up again
```suggestion
        pending "Figure out why the news item is not shown here any more"
```

spec/features/menu_items/top_menu_item_spec.rb
```diff
@@ -134,6 +141,7 @@ def click_link_in_open_menu(title)
       let(:user) { create(:anonymous) }
 
       it 'displays only projects, activity and news' do
+        skip "Figure out why the news item is not shown here any more"
```
aaron-contreras: This will let you know once it starts showing up again
```suggestion
        pending "Figure out why the news item is not shown here any more"
```

klaustopher:
> I don't recall the #all_permissions_allowed! and #forbid_everything! from the previous PR though.

Yeah, I added this, when I saw that we had a bunch of occurences where it just mocked `allow(user).to receive(:allowed_to?).and_return(true)`. 

I like your suggestion to name it better, will change

cbliard:
I like this new helper and I would have some suggestions for the dsl:

1. Prefix methods with `allow_`

   First time I saw it used in the `mocked_permission_helper_spec.rb`, it took me some time to understand what it does.
   Instead of
   ```ruby
   mock_permissions_for(user) do |mock|
     mock.in_project :view_work_packages, project: some_project
     mock.in_work_package :edit_work_packages, work_package: some_wp
     mock.globally :create_users
   end
   ```
   I find that this reads better:
   ```ruby
   mock_permissions_for(user) do |mock|
     mock.allow_in_project :view_work_packages, project: some_project
     mock.allow_in_work_package :edit_work_packages, work_package: some_wp
     mock.allow_globally :create_users
   end
   ```

2. Remove the bang at the end of methods
   This is subjective. I feel like the bang should be reserved for "dangerous" or suprising usages: if it can have unexpected side-effects, if it mutates the object in-place, if it raises exceptions (contrary to a non-bang version), and so on.
   I feel like `allow_everything!` and `forbid_everything!` are just doing what they are being told and should be named `allow_everything` and `forbid_everything`

3. Raise if no block given
   > If we want to deny every permission, we can do the following:
   >
   > ```ruby
   > mock_permissions_for(user)
   > # since we do not mock any permissions here, every permission check will return false!
   > ```
   >
   > As this might be a bit un-intuitive, there is this nicer way to write it:
   >
   > ```ruby
   > mock_permissions_for(user) do |mock|
   >  mock.forbid_everything
   > end
   > ```
   I totally agree that the first one is a bit counter-intuitive. We could forbid it by raising if no block is given.

And Aaron already suggested to rename `all_permissions_allowed!` into `allow_everything!`, and you already commited it :+1:.

spec/contracts/projects/shared_contract_examples.rb
```diff
@@ -29,15 +29,15 @@
...
-  let!(:allowed_to) do
-    allow(current_user)
-      .to receive(:allowed_to?) do |permission, permission_project|
-      permissions.include?(permission) && project == permission_project
+  let(:current_user) { build_stubbed(:user) }
+
+  before do
+    mock_permissions_for(current_user) do |mock|
+      mock.in_project(*permissions, project:)
+      mock.in_project :add_subprojects, project:
```
ulferts: Why is the :add_subprojects permission added here? It wasn't part of the original setup.

spec/lib/api/v3/wiki_pages/wiki_page_representer_rendering_spec.rb
```diff
@@ -48,13 +48,6 @@
 
   subject { representer.to_json }
 
-  before do
-    allow(user)
-      .to receive(:allowed_to?) do |permission, _project|
-      permissions.include?(permission)
-    end
```
ulferts, line 55: The `permissions`/`all_permissions` let statements can also be replaced

spec/lib/api/v3/work_packages/work_package_collection_representer_spec.rb
```diff
@@ -77,6 +77,7 @@
   current_user { user }
 
   before do
+    # TODO: Figure this one out ...
```
ulferts: This is a TODO

ulferts (changes requested):
I only have three tiny things but the overall change is good. I like it.

klaustopher:
@cbliard 

- Raising an exception when no block is given
- Removed the bang methods

Not sure yet about the renaming of the other methods. Let me think about this

ulferts (approved):
Given your vacation, I amended the tiny things I found in the PR when reviewing it the first time.

I also fixed the specs.

Guess what is missing now is your decision on whether you want to rename the mocked methods to have a `allow_` prefix like @cbliard suggested.

klaustopher:
Thanks @ulferts for all the small fixes and also helping understand the weird situation in https://github.com/opf/openproject/pull/13812/commits/f0acb95c90a9ba67be3110cec402532d8695b874 I was scratching my head on this one.

I have decided to follow @cbliard's suggestion and add the `allow_` prefix for the methods in the mocking helper. Makes the intention a lot clearer. Thanks for the suggestion 👍

PR #13819, opened 2023-09-28
Add concurency setting for test-core workflow.

ba1ash:
Expected behavior:
  - all pending and in-progress workflow runs are canceled on the PR update.
  - it should not change the behavior of workflow runs on push to dev/release branches, because `github.head_ref` is only availabe in PR context and if it is absent then `github.run_id` is used as building block for concurency group value. And `github.run_id` is unique for each workflow run.

### DIFF
--- a/.github/workflows/test-core.yml
+++ b/.github/workflows/test-core.yml
@@ -70,3 +70,7 @@ jobs:
       with:
         node-version: '14'
     - run: ./script/api/validate_spec
+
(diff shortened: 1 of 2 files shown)


mereghost (approved):

crohr (approved):
Looks good thanks! I would add a comment explaining the rationale for the concurrency group right within the workflow file though.

PR #13820, opened 2023-09-28
Fix `:remove` action for component turbo stream responses

aaron-contreras:
## Bug description
While working on attempting to remove a component from the DOM via a turbo stream response, I noticed that the `:remove` action results in an error every time.

Given I attempt to remove a component via a turbo stream, this resulted
in a

```
"You need to wrap your component in a `component_wrapper` block in order to use the turbo-stream methods"
```

error. This is due to the component template not being rendered and
the `@component_wrapped_used` instance variable not being set (because
the `component_wrapper` invocation in the component template is never
called.

## Proposed fix
The proposed fix for this is to render JUST the `component_wrapper` (without
its block content) in order to get the dom id that needs removal
and turbo streams is able to do its magic.

This is achieved by providing a `@wrapper_only` flag so when the
`component_wrapper` method is called within the component's template,
only the wrapper tag is rendered, making it enough to perform a
successful turbo stream removal of the component by handling this
case with the short circuit. We don't really care in this case about
rendering the entire component given all we care about is that
outer tag.


ulferts (approved):
LGTM

PR #13832, opened 2023-10-02
remove methods no longer in use - especially pagination

ulferts:


app/controllers/roles_controller.rb
```diff
@@ -115,20 +115,6 @@ def bulk_update
     end
   end
 
-  def autocomplete_for_role
```
dombesz, line 118: Line 34 of this file also needs to be changed.
https://github.com/opf/openproject/blob/158f3e4e3ac86706212430d736e525edfb52159a/app/controllers/roles_controller.rb#L34

dombesz (approved):
Just 1 comment, please feel free to merge after it's addressed.

app/controllers/roles_controller.rb
ulferts: 👍 well spotted

PR #13837, opened 2023-10-03
[#50295] copying projects with nextcloud project folders does not recreate file links to folders

wielinde:
https://community.openproject.org/wp/50295

The filesinfo endpoint returns folder location paths without trailing slashes, but the WebDAV PROPFIND does. Thus, the paths are not directly mappable. This leads to a situation where one cannot be mapped on the other.

### DIFF
--- a/modules/storages/app/common/storages/peripherals/storage_interaction/nextcloud/files_info_query.rb
+++ b/modules/storages/app/common/storages/peripherals/storage_interaction/nextcloud/files_info_query.rb
@@ -138,6 +138,8 @@ module Storages::Peripherals::StorageInteraction::Nextcloud
       return '/' if idx == nil
 
       idx += prefix.length - 1
+      # Remove the following when /filesinfo starts responding with a trailing slash for directory paths
+      # in all supported versions of OpenProjectIntegation Nextcloud App.
       file_path << '/' if mimetype == 'application/x-op-directory' && file_path[-1] != '/'
       Util.escape_path(file_path[idx..])
     end
--- a/modules/storages/app/common/storages/peripherals/storage_interaction/nextcloud/files_info_query.rb
+++ b/modules/storages/app/common/storages/peripherals/storage_interaction/nextcloud/files_info_query.rb
@@ -116,7 +116,7 @@ module Storages::Peripherals::StorageInteraction::Nextcloud
                 last_modified_by_name: value.modifier_name,
                 last_modified_by_id: value.modifier_id,
                 permissions: value.dav_permissions,
-                location: location(value.path)
+                location: location(value.path, value.mimetype)
               )
             else
               ::Storages::StorageFileInfo.new(
@@ -132,13 +132,13 @@ module Storages::Peripherals::StorageInteraction::Nextcloud
 
     # rubocop:enable Metrics/AbcSize
 
-    def location(file_path)
+    def location(file_path, mimetype)
       prefix = 'files/'
       idx = file_path.rindex(prefix)
       return '/' if idx == nil
 
       idx += prefix.length - 1
-
+      file_path << '/' if mimetype == 'application/x-op-directory' && file_path[-1] != '/'
       Util.escape_path(file_path[idx..])
     end
   end
--- a/modules/storages/app/services/projects/copy/file_links_dependent_service.rb
+++ b/modules/storages/app/services/projects/copy/file_links_dependent_service.rb
@@ -118,12 +118,26 @@ module Projects::Copy
         .result
     end
 
+    def remove_trailing_slashes_from_hash_keys(original_hash)
+      # Remove trailing slashes of directory paths.
+      # Nextcloud WebDAV PROPFIND requests for return for forlders location paths with trailing slashes.
+      # The `filesinfo` endpoint of the `OpenProject integration` Nextcloud app does not have those
+      # trailing slashes for folder paths. For matching both types of paths, we need to remove the trailing slashes.
+      clean_hash = {}
+      original_hash.each do |key, value|
+        clean_hash[key.chomp('/')] = value
+      end
+
+      clean_hash
+    end
+
     def folder_files_file_ids_deep_query(storage_requests:, path:)
-      storage_requests
-        .folder_files_file_ids_deep_query
-        .call(path:)
-        .on_failure { |r| add_error!("folder_files_file_ids_deep_query", r.to_active_model_errors) }
-        .result
+      result = storage_requests
+                 .folder_files_file_ids_deep_query
+                 .call(path:)
+                 .on_failure { |r| add_error!("folder_files_file_ids_deep_query", r.to_active_model_errors) }
+                 .result
+      remove_trailing_slashes_from_hash_keys(result)
     end
   end
 end
--- a/modules/storages/app/services/projects/copy/file_links_dependent_service.rb
+++ b/modules/storages/app/services/projects/copy/file_links_dependent_service.rb
@@ -118,26 +118,12 @@ module Projects::Copy
         .result
     end
 
-    def remove_trailing_slashes_from_hash_keys(original_hash)
-      # Remove trailing slashes of directory paths.
-      # Nextcloud WebDAV PROPFIND requests for return for forlders location paths with trailing slashes.
-      # The `filesinfo` endpoint of the `OpenProject integration` Nextcloud app does not have those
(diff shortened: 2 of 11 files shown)


modules/storages/app/services/projects/copy/file_links_dependent_service.rb
```diff
@@ -118,12 +118,26 @@ def files_info_query(storage_requests:, file_ids:)
...
-        .folder_files_file_ids_deep_query
-        .call(path:)
-        .on_failure { |r| add_error!("folder_files_file_ids_deep_query", r.to_active_model_errors) }
-        .result
+      result = storage_requests
+                 .folder_files_file_ids_deep_query
+                 .call(path:)
+                 .on_failure { |r| add_error!("folder_files_file_ids_deep_query", r.to_active_model_errors) }
+                 .result
+      remove_trailing_slashes_from_hash_keys(result)
```
Kharonus: 🔴 This is not the right location to fix it. One hint is, that you have added a comment describing nextcloud behaviour.

The sanitation of response data to fit the expected data must be done by the query. And we should add a unit test to the query to check, that the data is in the expected format.

Kharonus (changes requested):

modules/storages/app/services/projects/copy/file_links_dependent_service.rb
wielinde: @Kharonus (@ba1ash) I get what you say and I agree. Two thoughts:
- Having the fix here solves a problem that we imminently encounter only here: It fixes the copy of the projects. It is a working patch for 13.0. We could release it and help our users and customers and support team.
- If we move it further upwards to the query, I do not know the side effects. The QA gets complex. Released as a patch it might bring new bugs (as other places might rely on the trailing slashes?...for me that is hard to judge).

So we could leave the code where it is, create a CODE MAINTENANCE ticket for 13.1 and have it go through a propper QA. Patch level releases should not bring bugs. Mayor and Minor releases always bring some bugs (however hard we try).

Different topic:
The code in this file has a very long method `copy_dependency` that does a lot of different stuff, one after the other. It was hard to understand what every section is doing. Having separate methods for copying automatically managed folders and manually managed folders would be start. Some method names are not really helpful (What is `folder_files_file_ids_deep_query`? Why is it named like that? If we have such unclear names then at least I would need a comment.)

modules/storages/spec/features/delete_project_storage_and_file_links_spec.rb
```diff
@@ -46,7 +46,7 @@
   let(:file_link) { create(:file_link, storage:, container: work_package) }
   let(:second_file_link) { create(:file_link, container: work_package, storage:) }
   let(:delete_folder_url) do
-    "#{storage.host}/remote.php/dav/files/#{storage.username}/#{project_storage.project_folder_path.chop}"
+    "#{storage.host}/remote.php/dav/files/#{storage.username}/#{project_storage.project_folder_path.chop}/"
```
wielinde, line 49: @ba1ash isn't the `chop` removing the last slash? And now you add one again... Can't we simply remove both?

modules/storages/spec/requests/api/v3/storages/storages_spec.rb
```diff
@@ -334,7 +334,7 @@
   describe 'DELETE /api/v3/storages/:storage_id' do
     let(:path) { api_v3_paths.storage(storage.id) }
     let(:delete_folder_url) do
-      "#{storage.host}/remote.php/dav/files/#{storage.username}/#{project_storage.project_folder_path.chop}"
+      "#{storage.host}/remote.php/dav/files/#{storage.username}/#{project_storage.project_folder_path.chop}/"
```
wielinde, line 337: @ba1ash Same here

modules/storages/spec/services/storages/project_storages/delete_service_spec.rb
```diff
@@ -44,7 +44,7 @@
     let(:file_link) { create(:file_link, container: work_package, storage:) }
     let(:other_file_link) { create(:file_link, container: other_work_package, storage:) }
     let(:delete_folder_url) do
-      "#{storage.host}/remote.php/dav/files/#{storage.username}/#{project_storage.project_folder_path.chop}"
+      "#{storage.host}/remote.php/dav/files/#{storage.username}/#{project_storage.project_folder_path.chop}/"
```
wielinde, line 47: @ba1ash and here

modules/storages/spec/services/storages/project_storages/delete_service_spec.rb
```diff
@@ -67,7 +67,7 @@
     context 'with Nextcloud storage' do
       let(:storage) { create(:nextcloud_storage) }
       let(:delete_folder_url) do
-        "#{storage.host}/remote.php/dav/files/#{storage.username}/#{project_storage.project_folder_path.chop}"
+        "#{storage.host}/remote.php/dav/files/#{storage.username}/#{project_storage.project_folder_path.chop}/"
```
wielinde, line 70: @ba1ash and here

spec/services/projects/delete_service_spec.rb
```diff
@@ -55,7 +55,7 @@
           work_package = create(:work_package, project:)
           create(:file_link, container: work_package, storage:)
           delete_folder_url =
-            "#{storage.host}/remote.php/dav/files/#{storage.username}/#{project_storage.project_folder_path.chop}"
+            "#{storage.host}/remote.php/dav/files/#{storage.username}/#{project_storage.project_folder_path.chop}/"
```
wielinde, line 58: @ba1ash and here.

modules/storages/app/services/projects/copy/file_links_dependent_service.rb
ba1ash: I moved this logic to `files_info_query`. So, the logic is the following: we save the slash last if it is directory. Doing so will be in sync with the behavior of `propfind_query` where we are not able to distinguish directories easily if we remove all last slashes.
As well I modified  `Storages::Peripherals::StorageInteraction::Nextcloud::Util.escape_path` to not eat the last slash if `path` has it at the end. So, it has more clear behavior now.

modules/storages/app/common/storages/peripherals/storage_interaction/nextcloud/files_info_query.rb
```diff
@@ -132,13 +132,15 @@ def create_storage_file_infos
...
+    def location(file_path, mimetype)
       prefix = 'files/'
       idx = file_path.rindex(prefix)
       return '/' if idx == nil
 
       idx += prefix.length - 1
-
+      # Remove the following when /filesinfo starts responding with a trailing slash for directory paths
+      # in all supported versions of OpenProjectIntegation Nextcloud App.
+      file_path << '/' if mimetype == 'application/x-op-directory' && file_path[-1] != '/'
```
Kharonus, line 143: 🟡 why does this fix doesn't require amending the nextcloud files info query spec??

-> Answer: because it currently doesn't assert the real data. But it should. Please compare to the one drive files info query spec.

Kharonus (changes requested):
Can we agree to inflate the query spec a bit to test the new path data?

modules/storages/app/common/storages/peripherals/storage_interaction/nextcloud/files_info_query.rb
ba1ash: The request are stubbed, but they are stubbed with actual responses there. I added appropriate location expectation to the spec.

modules/storages/spec/common/peripherals/files_info_query_spec.rb
```diff
@@ -134,6 +134,40 @@
...
+                    created_at: Time.parse("Thu, 01 Jan 1970 00:00:00.000000000 UTC +00:00"),
+                    mime_type: "application/x-op-directory",
+                    size: 989752,
+                    owner_name: "admin",
+                    owner_id: "admin",
+                    trashed: false,
+                    last_modified_by_name: nil,
+                    last_modified_by_id: nil,
+                    permissions: "RMGDNVCK",
+                    location: "/OpenProject/Demo%20project%20%281%29/" },
```
Kharonus, line 153: perfect, this is what I wanted to see it tested! 👍🏽

Kharonus (approved):

PR #13838, opened 2023-10-03
[#50278] Journalize the MeetingAgendaItem class

dombesz:
https://community.openproject.org/wp/50278

### DIFF
--- a/app/services/journals/create_service.rb
+++ b/app/services/journals/create_service.rb
@@ -783,9 +783,7 @@ module Journals
 
       data_changes = (journable.journaled_columns_names - text_column_names).map do |column_name|
         <<~SQL
-          (#{journable_table}.#{column_name} != #{data_table}.#{column_name})
-          OR (#{journable_table}.#{column_name} IS NULL AND #{data_table}.#{column_name} IS NOT NULL)
-          OR (#{journable_table}.#{column_name} IS NOT NULL AND #{data_table}.#{column_name} IS NULL)
+          (#{journable_table}.#{column_name} IS DISTINCT FROM #{data_table}.#{column_name})
         SQL
       end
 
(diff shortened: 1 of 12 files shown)


app/services/journals/create_service.rb
```diff
@@ -696,9 +783,7 @@ def data_changes_condition_sql
 
       data_changes = (journable.journaled_columns_names - text_column_names).map do |column_name|
         <<~SQL
-          (#{journable_table}.#{column_name} != #{data_table}.#{column_name})
-          OR (#{journable_table}.#{column_name} IS NULL AND #{data_table}.#{column_name} IS NOT NULL)
-          OR (#{journable_table}.#{column_name} IS NOT NULL AND #{data_table}.#{column_name} IS NULL)
+          (#{journable_table}.#{column_name} IS DISTINCT FROM #{data_table}.#{column_name})
```
ulferts, line 786: 🙇 thanks

modules/meeting/app/services/meeting_agenda_items/create_service.rb
```diff
@@ -28,5 +28,15 @@
...
+    include TouchMeeting
+
+    def after_perform(call)
+      meeting_agenda_item = call.result
+      meeting = meeting_agenda_item.meeting
+
+      touch(meeting) unless meeting.nil?
+
+      call
+    end
```
ulferts: Is there a reason for doing this over adding

```
touch: true
```

to the `belongs_to :meeting` association of the `MeetingAgendaItem` model?

I don't see the use case of having an invalid meeting and still wanting to add agenda items.

ulferts (approved):
Nicely done, @dombesz. The functionality is solid.

I added a question to the PR which is why I am not merging it myself.

There is also one change the journalization does not pick up on but I think the behaviour enabling that is incorrect in the first place:

When having an agenda item referencing a work package a journal is created for the item just as it should be. But if the referenced work package is deleted, which will [delete the agenda item in turn](https://github.com/opf/openproject/blob/dev/app/models/work_package.rb#L73) no new journal is generated so the for the journal, the agenda item still exists.  

IMO the agenda item shouldn't be just deleted, at least not without warning as information about the meeting simply vanishes. 

But if that functionality is in fact desired, changing to a 
```
has_many :meeting_agenda_items, dependent: :destroy, touch: true
```
might just do the trick.

ulferts (approved):
Nicely addressed a complicated requirement there, @dombesz

PR #13843, opened 2023-10-04
Fix: `@component_wrapper_used` should be set on component and not target

aaron-contreras:
The purpose of `@component_wrapper_used` in this scenario is to check if the component that is to be inserted into the target has used a `component_wrapper` in its template rendering.

This PR both:
* Fixes the issue where this will always raise an error as the target component is not being rendered (and doesn't need to be).

* Ensures that there is a unique ID set at the top-most level of the turbo stream response and turbo can de-duplicate any existing components that match the current ID of the component before appending/prepending.

### DIFF
--- a/app/components/concerns/op_turbo/streamable.rb
+++ b/app/components/concerns/op_turbo/streamable.rb
@@ -28,6 +28,10 @@
 
 module OpTurbo
   module Streamable
+    # rubocop:disable OpenProject/AddPreviewForViewComponent
+    class MissingComponentWrapper < StandardError; end
+    # rubocop:enable OpenProject/AddPreviewForViewComponent
+
     extend ActiveSupport::Concern
 
     class_methods do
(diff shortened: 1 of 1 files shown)


app/components/concerns/op_turbo/streamable.rb
```diff
@@ -66,7 +66,7 @@ def render_as_turbo_stream(view_context:, action: :update)
       def insert_as_turbo_stream(component:, view_context:, action: :append)
         template = component.render_in(view_context)
 
-        unless @component_wrapper_used
+        unless component.instance_variable_get(:@component_wrapper_used)
```
ulferts: Instead of accessing the private instance variable, the variable needs to be exposed e.g. via  a `wrapped?` method.

ulferts (changes requested):
The change itself is fine. It is still somewhat confusing if compared to the rest of the file so it might make sense to add some explanatory comment.

ulferts (approved):
Reads way better now, @aaron-contreras. Thanks for that bit of extra care.

PR #13844, opened 2023-10-04
remove entity members from global roles query

ulferts:
Avoids returning members on entities (e.g. WorkPackage) on the `Authorization::UserGlobalRolesQuery`, so when `Authorization.roles(some_user)` is called.

To reproduce, first create such a member:
```Member.create(entity: WorkPackage.first, project: WorkPackage.first.project, roles: [WorkPackageRole.first], principal: User.active.first)```

Then fetch the global roles for the user:
`Authorization.roles(User.active.first)`

https://community.openproject.org/wp/50287

### DIFF
--- a/spec/services/authorization/user_global_roles_query_spec.rb
+++ b/spec/services/authorization/user_global_roles_query_spec.rb
@@ -125,7 +125,7 @@ RSpec.describe Authorization::UserGlobalRolesQuery do
         global_member.save!
       end
 
-      it 'is the global role and non member role' do
+      it 'is the global role, member role and non member role' do
         expect(described_class.query(user)).to contain_exactly(global_role, role, non_member)
       end
     end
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: spec/services/authorization/user_global_roles_query_spec.rb [hunk 1]
```diff
@@ -70,57 +76,69 @@
...
     end
 
-    context 'w/ the user having a global role' do
+    context 'with the user having a global role and a member role' do
       before do
+        member.save!
         global_member.save!
       end
 
       it 'is the global role and non member role' do
```
aaron-contreras:
```suggestion
      it 'is the global role, member role and non member role' do
```

spec/services/authorization/user_global_roles_query_spec.rb
```diff
@@ -70,57 +76,69 @@
...
 
-    context 'w/ the user having a global role and a member role' do
+    context 'with the user having a global role, a member role and a work package role' do
       before do
         member.save!
         global_member.save!
+        work_package_member.save!
       end
 
       it 'is the global role and non member role' do
```
aaron-contreras:
```suggestion
      it 'is the global role, member role and non member role' do
```

aaron-contreras (commented):
Two spec description updates, that's it 👍

ulferts:
Thanks for catching those, @aaron-contreras . I fixed the descriptions

aaron-contreras (approved):

PR #13845, opened 2023-10-04
[#50235] Granularly update Share Modal's components on CRUD actions

aaron-contreras:
## Description
Instead of re-rendering the whole modal on every CRUD action, we can perform more granular turbo stream responses:
### Create
1. New shares are prepended to the list, while preserving the alphabetical order of all initially rendered shares.
2. Updating a share via the invite form, removes the original share on the list and prepends the updated share to the the list.
Kapture 2023-10-05 at 10 37 34

### Update
1. Ability to respond with a `head :ok`, meaning no re-rendering is needed.
2. Tab indexes are preserved (big on accessibility)
Kapture 2023-10-05 at 10 44 47

### Destroy
1. Performs a `:remove` turbo stream action on the removed share
2. Preserves tab index (big on accessibility again)
Kapture 2023-10-05 at 10 47 14

### Noteworthy commits
https://github.com/opf/openproject/pull/13845/commits/ecc7d4d360475f9cfb63c7ecfe85572ad94bab8e: When performing a turbo stream response for a View Component slot, we need a way to specify the `with_*` call or whatever slot invocation mechanism is used as the `component_wrapper` and not only be able to provide a tag name as it stood before this PR (e.g. plain `div` with the wrapper key as its ID and some attributes).

This commit extends `component_wrapper` which can optionally take a method name as a symbol, which should render the custom component_wrapper you would like (e.g. a view component slot call to `box.with_row`) and adds the `wrapper_key` as its `id`.
### Work Package
https://community.openproject.org/work_packages/50235

### TODO
When https://github.com/opf/openproject/pull/13843 is merged this should be ready to review, as this commit is based off that PR and not based on dev

### DIFF
--- a/app/components/work_packages/share/modal_body_component.rb
+++ b/app/components/work_packages/share/modal_body_component.rb
@@ -38,7 +38,6 @@ module WorkPackages
         super
 
         @work_package = work_package
-        @shared_users = find_shared_users
       end
 
       def self.wrapper_key
@@ -46,8 +45,6 @@ module WorkPackages
       end
 
       private
-
-      attr_reader :shared_users
 
       def insert_target_modified?
         true
@@ -76,8 +73,8 @@ module WorkPackages
         list_container.instance_variable_set(:@list_arguments, new_list_arguments)
       end
 
-      def find_shared_users
-        @shared_users = User
+      def shared_users
+        @shared_users ||= User
                           .having_entity_membership(@work_package)
                           .includes(work_package_shares: :roles)
                           .where(work_package_shares: { entity: @work_package })
--- a/app/components/work_packages/share/modal_body_component.rb
+++ b/app/components/work_packages/share/modal_body_component.rb
@@ -53,6 +53,10 @@ module WorkPackages
         true
       end
 
+      def insert_target_modifier_id
+        'op-share-wp-active-shares'
+      end
+
       # There is currently no available system argument for setting an id on the
       # rendered <ul> tag that houses the row slots on Primer::Beta::BorderBox components.
       def invited_user_list(&)
--- a/app/components/work_packages/share/modal_body_component.rb
+++ b/app/components/work_packages/share/modal_body_component.rb
@@ -33,11 +33,11 @@ module WorkPackages
       include OpTurbo::Streamable
       include OpPrimer::ComponentHelpers
 
-      def initialize(work_package:, shared_users: nil)
+      def initialize(work_package:)
         super
 
         @work_package = work_package
-        @shared_users = shared_users || find_shared_users
+        @shared_users = find_shared_users
       end
 
       def self.wrapper_key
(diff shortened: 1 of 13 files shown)


app/components/work_packages/share/modal_body_component.rb
```diff
@@ -45,20 +47,39 @@ def self.wrapper_key
...
+      def insert_target_modified?
+        true
+      end
+
+      def insert_target_modifier_id
+        'op-share-wp-active-shares'
+      end
+
+      # There is currently no available system argument for setting an id on the
+      # rendered <ul> tag that houses the row slots on Primer::Beta::BorderBox components.
```
ulferts, line 57: It would help in the future to document the motivation for setting the id.

app/controllers/work_packages/shares_controller.rb
```diff
@@ -39,41 +39,77 @@ def index
...
+    respond_with_turbo_streams
+  end
+
+  def respond_with_remove_share
+    remove_via_turbo_stream(
+      component: WorkPackages::Share::ShareRowComponent.new(share: @share)
+    )
+
+    update_via_turbo_stream(
+      component: WorkPackages::Share::ShareCounterComponent.new(count: @shared_users.size)
```
ulferts: This only works because the `@shared_users` variable is set as a side effect in a method called by the same caller before this here is called. 

Using `find_shared_users` would be less surprising. Same applies to [further up](https://github.com/opf/openproject/pull/13845/files#diff-1044ba07b8889f8ff65fb7b1338806f8909221c74d285aadcb659334a3ca169dR95)

app/controllers/work_packages/shares_controller.rb
```diff
@@ -92,4 +128,12 @@ def find_role_ids(builtin_value)
     # is why we unscope.
     WorkPackageRole.unscoped.where(builtin: builtin_value).pluck(:id)
   end
+
+  def find_shared_users
+    @shared_users = User
+                      .having_entity_membership(@work_package)
+                      .includes(work_package_shares: :roles)
+                      .where(work_package_shares: { entity: @work_package })
+                      .ordered_by_name
```
ulferts: For the use cases in here, where only the existence of any share is checked for, something like

```
Member.of_work_package(@work_package)
```
would suffice.

Given the uses cases, it might be even better to just memoize the count in this method.

spec/features/work_packages/share_spec.rb
```diff
@@ -134,15 +129,15 @@
...
-      share_modal.expect_shared_with(comment_user, 'Comment')
-      share_modal.expect_shared_with(shared_project_user, 'Edit')
-      share_modal.expect_shared_with(current_user, editable: false)
+      share_modal.expect_shared_with(comment_user, 'Comment', position: 1)
       # This user's role was updated
-      share_modal.expect_shared_with(not_shared_yet_with_user, 'Comment')
+      share_modal.expect_shared_with(not_shared_yet_with_user, 'Comment', position: 2)
+      share_modal.expect_shared_with(shared_project_user, 'Edit', position: 3)
+      share_modal.expect_shared_with(current_user, position: 4, editable: false)
+      share_modal.expect_shared_with(view_user, 'View', position: 5)
```
ulferts, line 137: 👍 for those changes to the specs. Quite nice to have that behaviour safeguarded.

app/components/work_packages/share/modal_body_component.rb
```diff
@@ -45,20 +47,39 @@ def self.wrapper_key
...
       end
 
-      def share_editable?(share)
-        User.current != share.principal && sharing_manageable?
+      def find_shared_users
+        @shared_users = User
+                          .having_entity_membership(@work_package)
+                          .includes(work_package_shares: :roles)
+                          .where(work_package_shares: { entity: @work_package })
+                          .ordered_by_name
```
ulferts, line 80: This is a bit weird to me. The method is [called once like this](https://github.com/opf/openproject/pull/13845/files#diff-dcc2c760c3da0ae92ef0691162adad19cc81662a330b2a062405da3c68fa139fR41)

```
@shared_users = find_shared_users
```

I'd thus remove the variable assignment here.

ulferts (changes requested):
I like that change a lot. The behaviour is much nicer. 

I still have to get used to the way the controllers are written, somewhat stateful, but that is on me. 

The only tiny thing that needs handling is resetting the autocompleter after adding an item:

image

My statement somewhat contradicts what I wanted earlier, saying that I don't want to rerender the autocompleter. But to be precise, I only want the autocompleter to be cleared after the invite button is pressed, not rerendered/refocused. Although, resetting the focus to the autocompleter after pressing the invite would still be nice. And since the clear behaviour will need some js anyway, you might be able to throw that in as well :). But to be clear, the refocusing should not happen when removing/editing existing shares.

app/controllers/work_packages/shares_controller.rb
aaron-contreras: I like the idea of memoizing the count. It'd read better

aaron-contreras:
@ulferts Thanks for the feedback 👍 I've addressed all your remarks.

As discussed in our chat today, we compromised on replacing the `InviteUserFormComponent` on create/updates via the form. At least until our direction with the auto-completers is clear.

ulferts (approved):
I like it. Feels good to be using it.

Well done @aaron-contreras

PR #13852, opened 2023-10-05
[#50234] add storage filter to project and project storage collection

Kharonus:
[OP#50234](https://community.openproject.org/work_packages/50234)

### What is here?

- add storage_id and storage_url filter to project collection
- add storage_url to project storage collection
- added storage collection endpoint
- added requests tests
- refactored projects api specification as I touched it

### But... why??

- the storage filters are very handy when fetching a project list from a storage context, e.g. a specific nextcloud.

### DIFF
--- a/modules/storages/app/models/queries/storages/storages/storage_query.rb
+++ b/modules/storages/app/models/queries/storages/storages/storage_query.rb
@@ -28,27 +28,12 @@
 # See COPYRIGHT and LICENSE files for more details.
 #++
 
-class API::V3::Storages::StoragesAPI < API::OpenProjectAPI
-  helpers Storages::Peripherals::Scopes
-
-  resources :storages do
-    post &API::V3::Utilities::Endpoints::Create.new(model: Storages::Storage).mount
-
-    get &API::V3::Utilities::Endpoints::Index.new(model: Storages::Storage, scope: -> { visible_storages }).mount
-
-    route_param :storage_id, type: Integer, desc: 'Storage id' do
-      after_validation do
-        @storage = visible_storages.find(params[:storage_id])
+module Queries::Storages::Storages
+  class StorageQuery < Queries::BaseQuery
+    class << self
+      def model
+        @model ||= ::Storages::Storage
       end
-
-      get &API::V3::Utilities::Endpoints::Show.new(model: Storages::Storage).mount
-
-      patch &API::V3::Utilities::Endpoints::Update.new(model: Storages::Storage).mount
-
-      delete &API::V3::Utilities::Endpoints::Delete.new(model: Storages::Storage).mount
-
-      mount API::V3::StorageFiles::StorageFilesAPI
-      mount API::V3::OAuthClient::OAuthClientCredentialsAPI
     end
   end
 end
--- a/modules/storages/lib/api/v3/storages/storage_representer.rb
+++ b/modules/storages/lib/api/v3/storages/storage_representer.rb
@@ -182,12 +182,14 @@ module API::V3::Storages
     associated_resource :oauth_application,
                         skip_render: ->(*) { !represent_oauth_application? },
                         getter: ->(*) {
-                          next unless represent_oauth_application?
+                          next unless represent_oauth_application? && represented.oauth_application.present?
 
                           ::API::V3::OAuth::OAuthApplicationsRepresenter.create(represented.oauth_application, current_user:)
                         },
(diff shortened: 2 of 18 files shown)


modules/storages/app/models/queries/storages/storages/storage_query.rb
```diff
@@ -0,0 +1,39 @@
...
+# Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+#
+# See COPYRIGHT and LICENSE files for more details.
+#++
+
+module Queries::Storages::Storages
+  class StorageQuery < Queries::BaseQuery
+    class << self
+      def model
+        @model ||= ::Storages::Storage
```
ba1ash, line 35: Why do we need t cache the model class object into an instance variable here?

modules/storages/lib/api/v3/storages/storage_representer.rb
```diff
@@ -179,13 +179,15 @@ def initialize(model, current_user:, embed_links: nil)
     associated_resource :oauth_application,
                         skip_render: ->(*) { !represent_oauth_application? },
                         getter: ->(*) {
-                          next unless represent_oauth_application?
+                          next unless represent_oauth_application? && represented.oauth_application.present?
```
ba1ash, line 185: Let's put the extra condition in `represent_oauth_application?` if it is possible. It seems like the place for it. Now we have the answer to question `should oauth application be represented?` distributed some how through the code. Maybe `represent_oauth_application?` should be renamed.

modules/storages/spec/requests/api/v3/projects/projects_storages_filter_spec.rb
```diff
@@ -0,0 +1,118 @@
...
+  let!(:project4) { create(:project, members: { user => role }) }
+  let!(:storage1) { create(:nextcloud_storage) }
+  let!(:storage2) { create(:one_drive_storage) }
+  let!(:storage3) { create(:one_drive_storage) }
+  let!(:project_storage11) { create(:project_storage, project: project1, storage: storage1) }
+  let!(:project_storage12) { create(:project_storage, project: project1, storage: storage2) }
+  let!(:project_storage13) { create(:project_storage, project: project1, storage: storage3) }
+  let!(:project_storage23) { create(:project_storage, project: project2, storage: storage3) }
+  let!(:project_storage21) { create(:project_storage, project: project2, storage: storage1) }
+  let!(:project_storage31) { create(:project_storage, project: project3, storage: storage1) }
```
ba1ash: Can be these defined using `shared_let`? Then we could save some seconds.

ba1ash (approved):
It looks good to me. I have some questions, but they are not critical.

modules/storages/app/models/queries/storages/storages/storage_query.rb
Kharonus: I just copy-pasted another standard query. I really have no idea :D You need me to check this? @ba1ash

modules/storages/lib/api/v3/storages/storage_representer.rb
Kharonus: Actually, no, it can't be moved to `represent_oauth_application?`, as the result is different. 4 examples:

1. OneDrive storage -> there is no oauth application -> JSON response does not contain link property
2. Nextcloud storage, but user is no admin -> oauth application should not be exposed -> JSON response does not contain link property
3. Nextcloud storage, user is admin, but no oauth application is created -> missing oauth application must be visible -> JSON response contains the link property, which has a `null` value
4. Nextcloud storage, user is admin, oauth application is created -> render oauth application in JSON response

to achieve that, we need to distinguish between `represent_oauth_application?` and the presence of the oauth application itself

modules/storages/spec/requests/api/v3/projects/projects_storages_filter_spec.rb
Kharonus: How does it work? Can you maybe share some knowledge here?

modules/storages/app/models/queries/storages/storages/storage_query.rb
ba1ash: In my understanding there is no reason for doing this. Let it be like everywhere that's fine.

mereghost (approved):

ba1ash (approved):
:+1:

PR #13855, opened 2023-10-05
hide public permissions from role administration

ulferts:
Public permissions should never be manageable in the role administration. They are granted as soon as a project membership is created and therefore need to be hidden in the administration.

https://community.openproject.org/wp/50392


aaron-contreras (approved):
👍

PR #13854, opened 2023-10-05
Linking all the pieces of the Sharepoint Integration into the FilePicker/Workpackage view

mereghost:
##### Related Work Package: [OP#49128](https://community.openproject.org/projects/openproject/work_packages/49128), [OP#49127](https://community.openproject.org/projects/openproject/work_packages/49127), [OP#49126](https://community.openproject.org/projects/openproject/work_packages/49126), [OP#49125](https://community.openproject.org/projects/openproject/work_packages/49125)

On this PR, we link all the Sharepoint/OneDrive pieces for actual usage.

This meant changing the behaviour of the `FilesQuery` - in order to accommodate the inner workings of the File Picker - so that it can use a path based approach, plus some changes to our contract validators.

This also prepares for upload (it needs more changes on the FilePicker and is being handled by @Kharonus) by allowing connection to OneDrive and Sharepoint for uploads.

image
image
image

PR #13865, opened 2023-10-06
[chore] added one drive upload strategy for frontend

Kharonus:
### What?

- refactored nextcloud upload service to storage upload service using different strategies for each provider type
- fixed type urn in storage representer
- amended storage component to seed correct strategy

### Why?

- there was an idea of adding one upload service for each provider type, but I couldn't get it to work with angular DI

### ⚠️ IMPORTANT 

This PR currently points to a feature branch from @mereghost , as it is directly tight to the changes there. You can find the PR here: #13854

### DIFF
--- a/frontend/src/app/core/upload/convert-http-event.ts
+++ b/frontend/src/app/core/upload/convert-http-event.ts
@@ -26,14 +26,13 @@
 // See COPYRIGHT and LICENSE files for more details.
 //++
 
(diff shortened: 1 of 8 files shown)


mereghost (approved):
LGTM, but I´d like @apfohl view on this before a merge. =)

apfohl (approved):

PR #13875, opened 2023-10-09
Remove cron jobs from db with old handler names.

ba1ash:
After renaming some cron jobs records with old handler class name can be left in the database which produce unnecessary noise.

It resolves errors like:
https://appsignal.com/openproject-gmbh/sites/62bf0cc2d2a5e4832c923ff0/exceptions/incidents/119
https://appsignal.com/openproject-gmbh/sites/62d6c55083eb673480474bf9/exceptions/incidents/6543


wielinde (approved):
LGTM

PR #13877, opened 2023-10-09
[chore] added vcr data to one drive files query spec

Kharonus:
### What?

- refactored test for `one_drive/files_query` to use vcr data
- fixed found bugs
- added error data to files query

### Why?

- because live data is more accurate then stubs


mereghost (approved):
Awesome work! :D

PR #13883, opened 2023-10-10
[#50236] Allow sharing with groups

aaron-contreras:
#### Given the following group setup
Group setup

#### 1. Including Groups in share modal's auto-completer search results
Autocompleter demo

#### 2. Only displaying the Group's share unless a specific Role is granted to a Group Member

Displaying correct shares demo

#### 3. Updating a group's share also propagates to its inherited member roles
Hard to demonstrate on a screen recording but there's a feature spec :)

Updating a group share demo

#### 4. Removing a Group's share preserves the non-inherited shares a member might have
Removing group shares demo

#### 5. Mailer specific copy when sharing with Groups
Mailer screenshot

## Notes
See: https://community.openproject.org/work_packages/50236


app/components/work_packages/share/modal_body_component.html.erb
```diff
@@ -23,11 +23,13 @@
...
-              render(WorkPackages::Share::ShareCounterComponent.new(count: shared_users.count))
+              render(WorkPackages::Share::ShareCounterComponent.new(count: shared_principals.count))
             end
 
-            shared_users.each do |user|
-              share = user.work_package_shares.where(entity: @work_package).first
+            shared_principals.each do |principal|
+              share = principal.work_package_shares
+                               .where(entity: @work_package)
+                               .first
```
ulferts, line 32: Not caused by this PR but we need to find a better way of dealing with this. The way we are currently doing it will lead to n+1 queries. 

Not part of the PR so does not need to be fixed in it.

app/components/work_packages/share/permission_button_component.rb
```diff
@@ -70,8 +70,14 @@ def options
...
-        # TODO: handle having more than one role
-        share.roles.first
+        if share.persisted?
+          share.roles
+               .joins(:member_roles)
+               .where(member_roles: { inherited_from: nil })
+               .first
+        else
+          share.roles.first
+        end
```
ulferts, line 79: Same as above, we need to avoid n+1 loading this information. But as above, it can also happen outside of this PR.

app/components/work_packages/share/permission_button_component.rb
```diff
@@ -70,8 +70,14 @@ def options
       end
 
       def active_role
-        # TODO: handle having more than one role
-        share.roles.first
+        if share.persisted?
+          share.roles
+               .joins(:member_roles)
+               .where(member_roles: { inherited_from: nil })
+               .first
```
ulferts, line 76: `share.roles` will already have `member_roles` joined as it is the join table between `members` and `roles`.

app/controllers/work_packages/shares_controller.rb
```diff
@@ -131,7 +131,11 @@ def find_role_ids(builtin_value)
...
   end
 
-  def current_member_count
-    @current_member_count ||= Member.of_work_package(@work_package).size
+  def current_visible_member_count
+    @current_visible_member_count ||= Member
+                                        .joins(:member_roles)
+                                        .of_work_package(@work_package)
+                                        .where(member_roles: { inherited_from: nil })
+                                        .size
```
ulferts, line 139: In this PR you are redoing the pattern `Member.joins(:member_roles).where(member_roles: { inherited_from: nil))`. Additionally I see  `Role.join(:member_roles).where(member_roles: { inherited_from: nil))`.  

You could have a `MemberRole.not_inherited` scope. And by that have:
`Member.joins(:member_roles).where(MemberRole.not_inherited))`
`Role.join(:member_roles).where(MemberRole.not_inherited)`

You'd still end up repeating the joins. That can also be placed into a scope if you deem it appropriate. All of this is not super important admittedly.

app/forms/work_packages/share/invitee.rb
```diff
@@ -42,7 +42,7 @@ class Invitee < ApplicationForm
           # but restrict the type of the principal to users only. Subject to change
           # as we want to support groups soon.
           resource: 'principals',
-          filters: [{ name: 'type', operator: '=', values: ['User'] },
+          filters: [{ name: 'type', operator: '=', values: %w[User Group] },
                     { name: 'id', operator: '!', values: [::Queries::Filters::MeValue::KEY] }],
```
ulferts, line 46: There is an edge case hidden in here:

If the user is part of a group, that user can now give permissions to the group they are member of and by that in effect manage their own sharing. 

I don't think it is super critical for now since share_work_packages requires edit_work_packages. But this might be subject to change. And even now, the user potentially gains the ability to:
* be assignee 
* log own time and see them
* manage relations
over what the edit_work_package permission allows.

The only thing we could do to avoid this would be to limit the groups to those the user is not a member in and that might also be weird as it is hard to communicate that to the user. 

Or we exclude the user from being added when adding their group.

But ideas sound bad and I don't have better ones. One could also argue that we have the same problem with members where having the manage members permission allows to set additional permission for oneself.

app/models/principal.rb
```diff
@@ -57,12 +57,19 @@ class Principal < ApplicationRecord
            dependent: :nullify,
            class_name: 'Member',
            foreign_key: 'user_id'
+  has_many :work_package_shares,
+           -> { where(entity_type: WorkPackage.name) },
+           inverse_of: :principal,
+           dependent: :delete_all,
+           class_name: 'Member',
+           foreign_key: 'user_id'
```
ulferts, line 65: We now have:
```
has_many :members, dependent: :destroy
has_many :memberships, dependent: :nullify
has_many :work_package_shares, dependent: :delete_all
```

Which all reference the same entries in the database. I don't know which one wins.

An inherited problem but I would like to know what actually happens.

app/workers/mails/work_package_shared_job.rb
```diff
@@ -31,14 +31,44 @@ class Mails::WorkPackageSharedJob < ApplicationJob
...
+  # user has no other current shares on the work package (via a group or independently)
+  # and should be sent a mail.
+  def newly_invited_group_user_member_ids(work_package_member:)
+    Member.of_work_package(work_package_member.entity)
+          .joins(:member_roles)
+          .references(:member_roles)
+          .where(principal: work_package_member.principal.users)
+          .group('members.id')
+          .having("COUNT(member_roles.id) = 1")
+          .select('members.id')
```
ulferts, line 72: Minor thing and not awfully relevant here but typically you want to use `COUNT(*)` over `COUNT(column)`. The only reason for using `COUNT(column)` is if you want to exclude NULL values which that statement filters out. But it has to check that value for doing it which is why it is slower.

spec/features/work_packages/share_spec.rb
```diff
@@ -82,66 +96,160 @@
...
+        # Sent out email only on first share and not again when updating so the
+        # count should still be 1.
+        perform_enqueued_jobs
+        expect(ActionMailer::Base.deliveries.size).to eq(1)
+      end
 
-      # Reopening the modal will show the same state as before.
-      work_package_page.visit!
+      share_modal.close
+      click_button 'Share'
```
ulferts: Why close the modal at this step?

spec/features/work_packages/share_spec.rb
```diff
@@ -82,66 +96,160 @@
...
+        # This group's share was revoked
+        share_modal.expect_not_shared_with(not_shared_yet_with_group)
+        # This user's share was revoked via its group
+        share_modal.expect_not_shared_with(richard)
+        # This user's share was revoked
+        share_modal.expect_not_shared_with(edit_user)
+        # This user has never been added
+        share_modal.expect_not_shared_with(non_shared_project_user)
+
+        share_modal.expect_shared_count_of(7)
```
ulferts, line 249: Kudos 🙇 for the love you showed on this feature spec.

spec/support/components/work_packages/share_modal.rb
```diff
@@ -78,7 +80,12 @@ def change_role(user, role_name)
         end
       end
 
+      # Typing +:escape+ when the auto-completer is in focus doesn't exit the modal
+      # as the auto-completer doesn't seem to allow losing focus from itself with +:escape+
+      # Obtaining focus on the modal itself and then and then sending +:escape+ does the
+      # trick but this feels buggy.
       def close
+        modal_element.click
         modal_element.send_keys(:escape)
```
ulferts: By now there is a close button on the share modal so we can just press that one. So good if the usability is right. 😄

ulferts (changes requested):
I like the PR. Reads straightforward and the added tests give a feeling of being solid.

When trying out the functionality a couple of scenarios fail. Since the code for fanning out and destroying group memberships wasn't touched, it might well be that the problem is caused by code that existed before and has had the same problem all the time.

### Scenario 1

Given: 
* User is part of the group

Steps:
* Add user with permission "View"
* Add group with permission "Edit"
* Edit user's share to have permission "Edit"
* Remove group's share

Expected
* Group's share is removed
* Users' share stays

Actual:
* Both shares are removed

### Scenario 2

Given: 
* User is part of the group

Steps:
* Add user with permission "View"
* Add group with permission "Edit"
* Edit user's share to have permission "Edit"
* Re-Add group to have permission "View"

Expected
* Group's share is updated
* Users' share stays the same

Actual:
* Group's share is updated
* User's share is removed

### Scenario 3

Given: 
* User is part of the group

Steps:
* Add group with permission "Edit"
* Add user with permission "Edit"

Expected
* Group's share is unchanged
* Users' share is added with "Edit" permission

Actual:
* Group's share is unchanged
* Users' share is is not added

### Scenario 4

Given: 
* User is part of the group

Steps:
* Add group with permission "Edit"
* Add user with permission "View"
* Update user's share to have permission "Edit"
* Close and reopen modal

Expected
* Group's share is showing "Edit"
* Users' share is showing "Edit"

Actual:
* Group's share is showing "Edit"
* Users' share vanished

aaron-contreras:
Hi @ulferts , thanks for catching those edge cases❗️ As discussed in our quick-chat this morning, the pattern indeed re-occurs when both the Group and User are granted the same Role and I have tracked down what's going on. I've detailed my findings with a walkthrough of what's happening and I've verified the current behavior of how Project memberships with Users and Groups are currently behaving and actually there is no bug coming from before, what was discovered in this PR is expected behavior (at least for Projects):

## Why is this happening?

The underlying issue lies in the `WorkPackageMembers::CreateService`, and `WorkPackageMembers::UpdateService`; specifically, in the `Members::Concerns::RoleAssignment` concern. Let's take the first mentioned scenario:

### Walkthrough

Given:
- User is part of the group

Steps:
 1. Add user with permission "View"
 2. Add group with permission "Edit"
 3. Edit user's share to have permission "Edit"
 4. Remove group's share
 
Expected:
- Group's share is removed
- User's share stays

Actual:
- Both shares are removed

**Step 1**

Sharing the work package with the user with the "View" role creates the following DB structure:

image

**Step 2**

Subsequently, sharing the work package with the group with the "Edit role" creates the following DB structure:

image


All is fine after this step. However, when executing...

**Step 3**

The PATCH request performed by the select dropdown is providing `Member1` (The `Member` record for Dinesh). What happens next in the `SetAttributesService` is the root cause of the problem. I've annotated the code snippets with a step by step walkthrough:
```ruby
module WorkPackageMembers
  class SetAttributesService < ::BaseServices::SetAttributes
    prepend Members::Concerns::RoleAssignment

    private

    def set_attributes(params)
      # The call to super invokes the `set_attributes` method in the prepended concern
      super

      model.change_by_system do
        model.project = model.entity&.project
      end
    end
  end
end
```

```ruby
module Members::Concerns::RoleAssignment
  private

  # Which invokes this method
  def set_attributes(params)
    assign_roles(params)

    super
  end

  def assign_roles(params)
    raise ArgumentError, 'Cannot handle changing `roles`. Use role_ids instead.' if model.persisted? && params.key?(:roles)

    return unless params[:role_ids]

    role_ids = params
                 .delete(:role_ids)
                 .select(&:present?)
                 .map(&:to_i)
    # role_ids is the id of "EditRole"
    
    existing_ids = model.member_roles.map(&:role_id)
    # Dinesh's roles via his member_roles are "EditRole" and "ViewRole"

    # ["EditRole", "ViewRole"] - ["EditRole"] => ["ViewRole"]
    # Which means Dinesh's MemberRole1 (the non-inherited one) gets
    # marked for destruction.
    mark_roles_for_destruction(existing_ids - role_ids)

    # ["EditRole"] - ["EditRole", "ViewRole"] => []
    # Which means that Dinesh is not getting a new "MemberRole4"
    # with the "EditRole" that is not inherited.
    # Here lies the bug!
    build_roles(role_ids - existing_ids)
  end

  def mark_roles_for_destruction(role_ids)
    role_ids.each do |role_id|
      model
        .member_roles
        .detect { |mr| mr.inherited_from.nil? && mr.role_id == role_id }
        &.mark_for_destruction
    end
  end

  def build_roles(role_ids)
    role_ids.each do |role_id|
      model.member_roles.build(role_id:)
    end
  end
end
```

And here it is. After performing **Step 3**, Dinesh ends up solely with the `MemberRole3` which he inherited from his Group

image


So then when performing **Step 4**, the cleanup job is doing its job properly and deleting `Member2` along with `MemberRole2` (Cool Group's `Member` record along with its `MemberRole` record), **AND** `Member1` along with its only left `MemberRole3` (which was inherited). That is why all shares for the work package are gone and this is correct behavior on the cleanup job's end.

The 3rd described scenario in the failing cases demonstrates exactly what's happening in the `SetAttributesService`:

**Scenario 3**

Given:
* User is part of the group

Steps:
* Add group with permission "Edit"
* Add user with permission "Edit"

Expected:
* Group's share is unchanged
* Users' share is added with "Edit" permission

Actual:
* Group's share is unchanged
* Users' share is is not added


The reason why this is behaving in this manner can be digested in the following two bullet points providing a digest of how Project memberships for Users and Groups have worked thus far:

Given "Role A" exists
Given you're a member of "Cool Group", "Cooler Group", and _n_ other number of groups.

* "Role A" can only ever be granted to you via _n_ groups you're a member of (each creating a separate `member_role` record in your Project membership with its inherited_from column`inherited_from: <group_member_role_id>`) **OR** as a direct membership in a project. Meaning if you were granted "Role A" independently and then any group you belong to is granted "Role A", you would only keep the inherited `member_role` with "Role A" just granted to you via the group. This sort of makes sense because why would you have the same role assigned from two different sources on the same resource. At least, I believe that was the train of thought back in time.
* Following up with the previous bullet point, you can be granted "Role A" multiple times **as long as it is granted via a different group**. Meaning if you are part of "Cool Group" and "Cooler Group" and both groups were granted "Role A", you would have two distinct `member_roles` each with a different `inherited_from` id (respectively, "Cool Group"'s `member_role` and "Cooler Group"'s `member_role`). That's how the `CleanupInheritedRolesService` is able to work its magic when you're removed from a group or when a group's membership is revoked, given that there are two separate `member_role`s granted to you by both groups.

Following all of the above, this raises the following questions:
1. Do we want to be able to grant the same Role to a User independently **AND** via any _n_ number of groups?
2. If so, which benefits does this yield?

I would find reasonable if this works exactly as described in the Project membership mechanics, but I would also find it somewhat reasonable to be able to grant the same Role both independently and via _n_ number of groups: on the one hand, it'd be nice that if I had specifically granted Dinesh "Role A", and I granted a group he belonged to "Role A", that upon modifying the Role that the group is shared the work package with that Dinesh's role is intact given I specifically granted him "Role A". However, it seems to me like the former is exactly what a sharer might **NOT** desire.

I see two options:
1. Allowing the described behavior in entity memberships.
2. Modifying the share modal behavior:
   * Hide/disable the roles in the dropdowns in both the row and invite form that have been granted to a User via a group.
   * Given "Role A", "Role B", "Role C", and "Dinesh" is a member in "Cool Group"
   * If "Cool Group" was invited with "Role A", when attempting to invite "Dinesh" individually, only "Role B" and "Role C" should be available roles. 
   * If "Dinesh" was invited with "Role A", and subsequently "Cool Group" is invited with "Role A", then "Dinesh"'s share row should be removed from the dropdown modal (this could be surprising, some indication of what happened should be made visible)

aaron-contreras:
Discussed this today with @klaustopher and we DO want to be able to grant the same role to a user independently and via a group. Converting back to a draft to account for this

spec/features/work_packages/share_spec.rb
aaron-contreras: No reason. It's a left-over

spec/features/work_packages/share_spec.rb
aaron-contreras: Thank you :)

app/components/work_packages/share/modal_body_component.html.erb
aaron-contreras: https://community.openproject.org/work_packages/50544/activity

app/components/work_packages/share/permission_button_component.rb
aaron-contreras: https://community.openproject.org/work_packages/50544/activity

klaustopher (commented):
PR looks good from my side. I'll leave the final review for @ulferts as he has brought up some valid concern and I cannot 100% assess if they are fulfilled.

app/forms/work_packages/share/invitee.rb
ulferts: I copied that question to the Share-Epic on OpenProject community.

ulferts (approved):
Thanks for the detective work, @aaron-contreras. Feels good now and I did not run into any more weird cases.

PR #13894, opened 2023-10-11
Fix/fix attachments unit test

cbliard:
While reviewing #13861, I spoted a unit test which did not behave correctly: it was expecting a 403 unauthorized because of missing permissions, but when using the correct permissions, it was still failing with 403 because it was using the anonymous user. Then it was a 422 for missing file size, and then a 422 again because the fog provider was not properly configured.

This was all fixed in the first commit.

Then the second commit adds a positive test case to ensure there will not be any regression on this anymore, and also test the `:add_work_package_attachments` permission introduced in #13861.


ulferts (approved):
👍 thanks for following up on this.

PR #13908, opened 2023-10-12
Update README.md

williamhatch:
development environment setup in mac, need to run `arch -arm64 brew install rbenv ruby-build` in the M1/M2 CPU (not intel)

### DIFF
--- a/docs/development/development-environment-osx/README.md
+++ b/docs/development/development-environment-osx/README.md
@@ -69,7 +69,7 @@ Next, install a PostgreSQL database. If you wish to use a MySQL database instead
 
 ```shell
 # Install postgres database
-$ brew install postgres
+$ brew install postgresql
 
 # Create the database instance
 $ postgres -D /usr/local/var/postgres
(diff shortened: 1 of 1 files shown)


klaustopher:
Hey @williamhatch, thanks for the contribution. 

I am a bit confused about the need for the explicit switch of the architecture. I am running on a M2 mac as well and for me, `rbenv` automatically picked up the correct architecture. Are you maybe running your terminal under Rosetta to run in x86 mode? Normally, a terminal should be running in arm64 mode and this should be picked up by rbenv and ruby build as well.

What's the output in your console, when you run `arch` command. It should be

```shell
$ arch
arm64
```

williamhatch:
Thanks @klaustopher , you are right. If I run in the default terminal, it is arm64 mode.  But I found another issue: https://thedevpost.com/blog/install-ruby-with-rbenv-on-m1-mac/.

klaustopher:
Ruby 3.2 should install without any flags on arm64. Those flags should only be needed to install ruby 2.x which is not officially supported. As I said, using an m2 and I could just install ruby without any of this.

as-op:
Thank you @williamhatch for the docs contribution! 
Unfortunately I cannot reproduce it, too. Ruby 3.1/3.2 installs just fine on a arm64 Mac.
I'm closing this PR.

PR #13911, opened 2023-10-12
Use entity_class instead of hard coded class name in UserPermissibleService

klaustopher:
Just one small addition, as this will be re-usable in the future, we can already future prove it by using the `entity_class` here instead of the hard coded `WorkPackage`


ulferts (approved):
👍

PR #13912, opened 2023-10-12
Allow multiple-directory dynamic Stimulus controller namespacing

aaron-contreras:
As per the docs:
https://devdocs.io/javascript/global_objects/string/replace

> "If pattern is a string, only the first occurrence will be replaced."

---

This should allow a stimulus controller to be namespaced upon multiple subdirectories. For example:

`dynamic/work-packages/shares/x.controller.ts`


oliverguenther (approved):
Good catch 🙏

PR #13915, opened 2023-10-13
29203 Fix missing permission on attachment prepare

cbliard:
https://community.openproject.org/wp/29203

This fixes the error when having only the `:add_work_package_attachments` permission and trying to upload an attachment to a work package while doing a direct upload (like with s3).

This was actually a proposed implementation originally in https://github.com/opf/openproject/pull/13861/commits/19e5d9890710d9bfa7289b6d1ee2bf094493b17d#diff-b9569a0f85cab0526410ea11ced6758c01f535d5f152b2789456157cc477ce95L59


oliverguenther (approved):
Thanks for the correction! 🙇

PR #13916, opened 2023-10-13
[#50907] Implement /oauth_clients/:oauth_client_id/ensure_access 

ba1ash:
See https://community.openproject.org/work_packages/50907 for the route description.
There are two additional changes to note:
- `require_login` check has been enabled for both `/ensure_access` and `/callback`
- setting up `oauth_state_*` cookie on frontend was modified a bit to have expiration set to ~1 hour.


javiervalerocelada:
Hello, I have configured the Open ID plugin for 4 months now. Today, October 13, 2023, it has stopped working. I always get the same error: user=12 Invalid or expired access token: Invalid or expired access token

I have updated to openproject 13.0.5 but the error follow. Can you help me?




Captura desde 2023-10-13 19-43-45

ba1ash:
> Hello, I have configured the Open ID plugin for 4 months now. Today, October 13, 2023, it has stopped working. I always get the same error: user=12 Invalid or expired access token: Invalid or expired access token
> 
> I have updated to openproject 13.0.5 but the error follow. Can you help me?
> 
> Captura desde 2023-10-13 19-43-45

Hello @javiervalerocelada,
Thank you for reporting the issue. Unfortunately, it is not the right place for it. Could you please create a bug through https://community.openproject.org/ in OpenProject project. Here you will find a guide how to do it: https://www.openproject.org/docs/development/report-a-bug/.

Speaking about the issue you described here: there is a lack of information to help with it right away(the log is too short and it is not clear doing what specifically you experience the error. I assume you try to log in, right?). Please, create a bug report following the guide above and we will do our best to help you.

Best,
Pavel

app/controllers/oauth_clients_controller.rb
```diff
@@ -68,6 +69,41 @@ def callback
...
+    if oauth_client.nil?
+      flash[:error] = [I18n.t('oauth_client.errors.oauth_client_not_found'),
+                       I18n.t('oauth_client.errors.oauth_client_not_found_explanation')]
+
+      if User.current.admin?
+        redirect_to admin_settings_storages_path
+      else
+        redirect_to root_url
+      end
+    else
```
mereghost: WDYT of extracting this into a method to reduce branching on this action? Maybe something along the lines of

```ruby
some_pretty_descriptive_name unless oauth_client
```

app/controllers/oauth_clients_controller.rb
```diff
@@ -68,6 +69,41 @@ def callback
...
+      connection_manager = OAuthClients::ConnectionManager.new(
+        user: User.current,
+        configuration: oauth_client.integration.oauth_configuration
+      )
+
+      # check if the origin is the same
+      destination_url = if params[:destination_url].present? && params[:destination_url].start_with?(root_url)
```
mereghost: Could this test become a `params.fetch(:destination_url, '').start_with?(root_url)`?

modules/storages/app/common/storages/peripherals/oauth_configurations/nextcloud_configuration.rb
```diff
@@ -41,7 +41,7 @@ def initialize(storage)
         end
 
         def authorization_state_check(token)
-          util = StorageInteraction::Nextcloud::Util
+          util = ::Storages::Peripherals::StorageInteraction::Nextcloud::Util
```
mereghost, line 44: You shouldn't need the full qualifier here as the `Util` object is in down the namespace chain. Did you ran into some issue due to that?

modules/storages/config/locales/en.yml
```diff
@@ -213,7 +213,8 @@ en:
     member_connection_status:
       connected: "Connected"
       connected_no_permissions: "User role has no storages permissions"
-      not_connected: "Not connected. The user should login to the storage via the %{files_label} tab of a work package to connect."
+      not_connected: "Not connected. The user should login to the storage via the following %{link}."
+      link: "link"
```
Kharonus: 🟡 don't we have some generic `link_label`?

modules/storages/app/common/storages/peripherals/oauth_configurations/nextcloud_configuration.rb
ba1ash: Yes, I do not remember the exact case. Probably, it happened when I run one of the specs alone. I should have created a note about that.

modules/storages/config/locales/en.yml
ba1ash: I do not know. I'll have a look.

app/controllers/oauth_clients_controller.rb
ba1ash: Personally, I'd stick to current implementation, but let's try.

app/controllers/oauth_clients_controller.rb
ba1ash: Yes, it looks better to me. Thanks!

modules/storages/config/locales/en.yml
ba1ash: I have not found one. It seems like we do not have it.

Kharonus: Maybe we should insert it at the same level as `label_list` in `config/locales/en.yml`? WDYT? It is a way to generic term for being so deeply nested.

ba1ash: Sure, it makes sense. I'll add it there.

PR #13922, opened 2023-10-16
Enable bullet gem again

dombesz:
Adding the Bullet gem to the test environment. It is configured to raise test error when an N+1 query is detected.
https://community.openproject.org/wp/50620


mrmir:
@dombesz this would still be good to have, yes?

PR #13925, opened 2023-10-16
Permission Rework [5/x]: More places in the codebase that still mocked allowed_to calls

klaustopher:
Found a few more places, where we still mocked `allowed_to?` calls on the user. Replaced those as well.

Also I found that in a bunch of places we silently mocked permissions on `nil`. This is now prevented in the mocking helper and I have removed all places where we silently did this.


aaron-contreras:
I'm loving https://github.com/opf/openproject/pull/13925/commits/bd660582c84f4848c710123a0a0290c4b06743c7 These mocks are getting an upgrade to their "verifying double" ability on every commit ❗️

modules/costs/spec/contracts/time_entries/update_contract_spec.rb
```diff
@@ -55,28 +55,27 @@
 
     context 'if project changed' do
       let(:new_project) do
-        build_stubbed(:project).tap do |p|
+        build_stubbed(:project).tap do |new_project|
```
aaron-contreras:
```suggestion
        build_stubbed(:project) do |new_project|
```

modules/documents/spec/api/v3/documents/document_representer_rendering_spec.rb
```diff
@@ -49,13 +49,6 @@
 
   subject { representer.to_json }
 
-  before do
-    allow(user)
-      .to receive(:allowed_to?) do |permission, _|
-      permissions.include?(permission)
-    end
-  end
-
```
aaron-contreras, line 58: Couldn't find where this spec is mocking its permissions. I see nothing going on there in the shared examples. Is no mocking needed?

aaron-contreras (approved):
👍 Looks good

modules/documents/spec/api/v3/documents/document_representer_rendering_spec.rb
klaustopher: The `it_behaves_like 'has a titled link'` shared example has the mocks

ulferts (approved):
I like it. It is good that you continue to clean those specs up.

ulferts:
It might be an option to introduce a rubocop rule to nudge people to using the mocks.

PR #13929, opened 2023-10-17
Extract method to request all permissions on a given context for a user

klaustopher:
Discussion in Element spawned this. The method helps to grab all permissions a user has on a given context. This can be used in multiple places, one of them is the UserPermissibleService, but other's can use this as well (👀 to the storages team)


mereghost (approved):
Nice! Thanks a lot for this, @klaustopher

PR #13931, opened 2023-10-17
[#50543] Failing GRAPH API call leads to rails exception

Kharonus:
[OP#50543](https://community.openproject.org/work_packages/50543)

### What???

- added new endpoints for opening storages
  - `/api/v3/storages/:id/open`
  - `/api/v3/project_storages/:id/open`
- amended queries
  - `open_file_link_query` used for generating open links of items with an id
  - `open_storage_query` used for generating open links of a storage file root
- removed `open_link` method from storage records
- added vcr based unit tests for new queries
- added api requests tests for new endpoints

### But... Why??

- providing and using static links for opening external resources has some benefits
  - links can be used in UI without making external request whil bootstrapping the app

### DIFF
--- a/modules/storages/app/common/storages/peripherals/storage_interaction/nextcloud/open_storage_query.rb
+++ b/modules/storages/app/common/storages/peripherals/storage_interaction/nextcloud/open_storage_query.rb
@@ -43,7 +43,7 @@ module Storages
 
           # rubocop:disable Lint/UnusedMethodArgument
           def call(user:)
-            ServiceResult.success(result: File.join(@uri.to_s, 'index.php/apps/files'))
+            ServiceResult.success(result: Util.join_uri_path(@uri, 'index.php/apps/files'))
           end
 
           # rubocop:enable Lint/UnusedMethodArgument
--- a/modules/storages/app/common/storages/peripherals/storage_interaction/nextcloud/open_storage_query.rb
+++ b/modules/storages/app/common/storages/peripherals/storage_interaction/nextcloud/open_storage_query.rb
@@ -28,28 +28,27 @@
 # See COPYRIGHT and LICENSE files for more details.
 #++
 
-class API::V3::Storages::StoragesAPI < API::OpenProjectAPI
-  helpers Storages::Peripherals::Scopes
+module Storages
+  module Peripherals
+    module StorageInteraction
+      module Nextcloud
+        class OpenStorageQuery
+          def initialize(storage)
+            @uri = storage.uri
+          end
 
-  resources :storages do
-    post &API::V3::Utilities::Endpoints::Create.new(model: Storages::Storage).mount
+          def self.call(storage:, user:)
+            new(storage).call(user:)
+          end
 
-    get &API::V3::Utilities::Endpoints::Index.new(model: Storages::Storage, scope: -> { visible_storages }).mount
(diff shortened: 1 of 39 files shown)


modules/storages/app/common/storages/peripherals/storage_interaction/nextcloud/open_storage_query.rb
```diff
@@ -0,0 +1,54 @@
...
+            @uri = storage.uri
+          end
+
+          def self.call(storage:, user:)
+            new(storage).call(user:)
+          end
+
+          # rubocop:disable Lint/UnusedMethodArgument
+          def call(user:)
+            ServiceResult.success(result: File.join(@uri.to_s, 'index.php/apps/files'))
```
ba1ash: Could it be just like that without path? If no, I'd suggest to use `Storages::Peripherals::StorageInteraction::Nextcloud::Util.join_uri_path`. I believe we agreed to use this one instead of `File.join`.
```suggestion
            ServiceResult.success(result: @uri.to_s)
```

ba1ash (commented):
I find it strange to have a redirect action in the API somehow. I believe that it should be placed in an ordinary controller. In the API I would vote for just extending storage and project_storage endpoint with a link like you did. So, instead of having `/api/v3/project_storages/{id}/open` and `/api/v3/storages/{id}/open` doing redirects I suggest having something like `storages/{id}/open` and `projects/{project_id_or_name}/project_storages/{id}/open`. It feels more natural to me. What do you think?

Kharonus:
TBH I don't understand the question :D what is the difference of your suggestion? apart from the fact, that project_storages is not nested under project context in the API.

Edit: Ah, wait, the difference is, that it is no API endpoint. But I don't get the reason? HTTP redirects are a very common way to resolve external links. What is the benefit of a rails controller?

modules/storages/app/common/storages/peripherals/storage_interaction/nextcloud/open_storage_query.rb
Kharonus: without the path it opens the dashboard. some time ago, we agreed on the requirement, that the root open link should point to files app.

ba1ash: Okay, I see, thanks!
What about using `Storages::Peripherals::StorageInteraction::Nextcloud::Util.join_uri_path`?

ba1ash (approved):
LGTM
:shipit:

ba1ash:
> TBH I don't understand the question :D what is the difference of your suggestion? apart from the fact, that project_storages is not nested under project context in the API.
> 
> Edit: Ah, wait, the difference is, that it is no API endpoint. But I don't get the reason? HTTP redirects are a very common way to resolve external links. What is the benefit of a rails controller?

Just for history and clarification, because the question was discussed in the team call. We agreed to proceed with the current approach, because we use it already(for attachments, for instance) and it works fine. Therefore it does not make sense to behave differently in this scenario.

modules/storages/app/common/storages/peripherals/storage_interaction/nextcloud/open_storage_query.rb
Kharonus: Yes, changed that, thanks for noticing!

PR #13932, opened 2023-10-17
[Op#50212]: Add Turbo Stream and Frame View Components

akabiru:
https://community.openproject.org/wp/50212

Define `OpTurbo::{FrameComponent,StreamComponent}`

- [x] Remove "wrapper" from name- see @cbliard's https://github.com/opf/openproject/pull/13866#discussion_r1357813436
- [x] Add `OpTurbo::FrameComponent`
- [x] Define Lookbook previews

Screenshot 2023-10-17 at 7 10 10 PM
Screenshot 2023-10-18 at 7 32 43 PM


lookbook/previews/op_turbo/stream_component_preview.rb
```diff
@@ -0,0 +1,24 @@
+# frozen_string_literal: true
+
+module OpTurbo
+  # @logical_path OpenProject/OpTurbo
+  class StreamComponentPreview < Lookbook::Preview
+    # Renders a turbo-stream tag with given action and target
+    # @param action select { choices: [append, prepend, replace, update, remove, before, after] }
```
akabiru: 🚧  _Rails overrides the `action:` local var in Lookbook preview, one way would be to rename the variable but doesn't seem like a good enough reason._ Thoughts / Ideas? 🤔

aaron-contreras: How about:

```ruby
    # Renders a turbo-stream tag with given action and target
    # @param _action select { choices: [append, prepend, replace, update, remove, before, after] }
    # @param target text
    def default(_action: 'append', target: 'model_id')
      template = template_example_from_action(_action, target)
      render_with_template(locals: { template:, action: _action, target: })
    end
```

Preview demo

akabiru: Worked beautifully, cheers Aaron ⭐

aaron-contreras (approved):
Awesome!

PR #13937, opened 2023-10-18
fix typo in heading

lindenthal:


lindenthal:
Hi @birthe,

is it correct to only capitalize the first word in a heading?

Best
Niels

birthe (approved):
Thank you, @lindenthal, Yes, it's correct to only capitalize at the beginning of a sentence (also for titles) and proper names.

PR #13940, opened 2023-10-18
Bug/49947 project storage can be set to automatically managed while storage is not configured for it

akabiru:
#### https://community.openproject.org/work_packages/49947

When storage does not support automatic folder management, do not allow the automatic option when adding the storage into a project

Screenshot 2023-10-17 at 3 47 27 PM
Screenshot 2023-10-17 at 3 45 30 PM

### DIFF
--- a/modules/storages/app/models/storages/project_storage.rb
+++ b/modules/storages/app/models/storages/project_storage.rb
@@ -52,7 +52,7 @@ class Storages::ProjectStorage < ApplicationRecord
   scope :automatic, -> { where(project_folder_mode: 'automatic') }
 
   def automatic_management_possible?
-    storage.provider_type_nextcloud? && storage.automatically_managed?
+    storage.present? && storage.provider_type_nextcloud? && storage.automatically_managed?
   end
 
   def project_folder_path
--- a/modules/storages/app/models/storages/project_storage.rb
+++ b/modules/storages/app/models/storages/project_storage.rb
@@ -51,6 +51,10 @@ class Storages::ProjectStorage < ApplicationRecord
 
   scope :automatic, -> { where(project_folder_mode: 'automatic') }
 
+  def automatic_management_possible?
+    storage.provider_type_nextcloud? && storage.automatically_managed?
+  end
+
   def project_folder_path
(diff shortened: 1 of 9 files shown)


modules/storages/app/models/storages/project_storage.rb
```diff
@@ -51,6 +51,10 @@ class Storages::ProjectStorage < ApplicationRecord
 
   scope :automatic, -> { where(project_folder_mode: 'automatic') }
 
+  def automatic_management_possible?
+    storage.present? && storage.provider_type_nextcloud? && storage.automatically_managed?
```
Kharonus, line 55: 🟡 not very happy about the naming of `storage.automatically_managed?` sounds very off in this context, when checking if auto management is possible for a project storage.

but, this is part of database already, just for a bad gut feeling, I wouldn't trigger a new migration yet.

Kharonus (approved):
LGTM so far.

But tests are failing, so there is probably something still missing?

modules/storages/app/models/storages/project_storage.rb
akabiru: Hmm, I'm not sure I see the disconnect in the name- seems quite fitting from my POV as a project folder can only be automatically managed, if the linked storage is automatically managed. Am I missing sth? 🤔 

FWIW, if indeed the name is not fitting, it's fairly reachable to introduce an alias for `storage.automatically_managed?` that would be more fitting for this context- we can do so without a database migration. Any alternate names that come to mind?

akabiru:
> LGTM so far.
> 
> But tests are failing, so there is probably something still missing?

Cheers, yeah- had some factory mismatch issues - patched https://github.com/opf/openproject/pull/13940/commits/f1b7439f59e80df38a2eb8dcd83b6117f3fb0a50

modules/storages/app/models/storages/project_storage.rb
Kharonus: resolved in live discussion. @akabiru will raise another short PR introducing an alias here to get a more understandable interface to the storage.

PR #13942, opened 2023-10-18
[50436] Attribute help text icon overlaps with the field

bsatarnejad:
https://community.openproject.org/projects/openproject/work_packages/50436/activity


HDinger (commented):
Hi @bsatarnejad 

although the solution works, I am wondering whether it is correct to always truncate the labels. Many of our customers have custom fields that only differentiate at the very end. So I think we should rather only truncate long single words. When the name consists of multiple works, we should rather wrap.

HDinger (approved):

PR #13943, opened 2023-10-18
Optimize Journals::CreateService to not include unnecessary association

dombesz:
The generated sql query of the `Journals::CreateService` always included the conditions of all the possible journaled relations of a journable object. This can slow the query down, executing unnecessary selects and inserts that do not make a difference. 

For example, when the journable is a `Meeting` object, the only journaled relation is the `agenda_items`. The previous create service would still do calculations for the other journaled relations such as the `attachable`, `customizable` etc, which are not available for the `Meeting` object. Eliminating those extra queries should make journaling faster.


ulferts (approved):
Very nicely done, @dombesz

PR #13945, opened 2023-10-19
 Permission Rework [6/x]: Move `allowed_to_globally` checks to new checks

klaustopher:
This takes all occurences of `allowed_to_globally?` and moves them to:

- `allowed_globally?` when it's a global permission
- `allowed_in_any_project?` when it's a project permission
- `allowed_in_any_work_package?` when it's a WP permission

This also uncovered two issues in the `UserPermissibleService`:
- The `allowed_globally?` methods did not return `false` for a locked user
- The `allowed_in_any_entity?` method did not return true, when the user had the permission in the project, but there was no entity (i.e. WorkPackage) created yet, the permission check returned `false`


app/services/authorization/user_permissible_service.rb
```diff
@@ -29,7 +30,12 @@ def allowed_in_any_project?(permission)
       perms = contextual_permissions(permission, :project)
       return true if admin_and_all_granted_to_admin?(perms)
 
-      Project.allowed_to(user, perms).exists?
+      # If no projects matching the scope exists, it could still be that the user has access to the
+      # permission either via the anonymous or non-member role, especially when there are no projects
+      #  in the database. So if this returns false, we check non-member and anonymous permissions as well.
+      Project.allowed_to(user, perms).exists? ||
+        non_member_permissions.intersect?(perms.map { |perm| perm.name.to_sym }) ||
+        anonymous_permissions.intersect?(perms.map { |perm| perm.name.to_sym })
```
klaustopher: Very interesting case. We rely on the `Project.allowed_to` scope here. If there are no projects in the DB, this will return `false`, even if the user has the permission either via the non-member or anonymous role (that would correctly be checked, when there were projects in the DB). So, to have parity with the previous implementation, I have added those checks here as well

app/contracts/users/base_contract.rb
```diff
@@ -96,5 +95,9 @@ def existing_auth_source
       end
     end
     # rubocop:enable Rails/DynamicFindBy
+
+    def can_create_or_manage_users?
+      user.allowed_globally?(:manage_user) || user.allowed_globally?(:create_user)
```
ulferts, line 100: Good cleanup

app/models/users/permission_checks.rb
```diff
@@ -108,7 +108,7 @@ def allowed_to_in_project?(action, project)
   end
 
   def allowed_to_globally?(action)
-    # OpenProject::Deprecation.replaced(:allowed_to_globally?, :allowed_globally?, caller)
+    OpenProject::Deprecation.replaced(:allowed_to_globally?, :allowed_globally?, caller)
```
ulferts, line 111: 👍

app/services/authorization/user_permissible_service.rb
ulferts: I wonder about the use case in which this distinction actually comes into play. Can you please help me out here, @klaustopher?

As I am still wondering about the use case my opinion is to not have that fallback for a number of reasons:
* All permissions Anonymous and NonMembers have are project permissions. Within a project, the permissions are only granted if the module that permission is part of is active. A project not existing and the module not being active to me is the same thing. So I'd argue that if the project does not exist, the permission is not granted.
* This creates a different behaviour between the builtin roles and the user configured roles. There, the permissions are definitely only granted if the project exists. 
* Normally the builtin roles only come into play in case the project is public. Now they are always applied and serve as a fallback.


Not so much an argument but a flaw in the implementation, the current code would always grant the permissions as fallbacks even if the user configured roles do not grant them. So the minimal permissions the user would currently have are the union between the NonMember and the Anonymous roles permissions. If the special case is to be catered for, I'd add a check so that they are only applied in case of `Project.none?`.

ulferts (changes requested):

app/services/authorization/user_permissible_service.rb
klaustopher: The problem occured in some tests, especially in some of the tests related to showing menu items. For example, we show the `Meetings` entry if `user.allowed_in_any_project?(:view_meetings)`. Currently the implentation is `Project.allowed_to?(user, :view_meetings).exists?`. This only returns true, when there is **at least one project** that fulfills this. If there are no projects in the DB, this returns `false`.

The current implenentation of `allowed_to_globally?(:view_meetings)` returns `true` when the non-member has the permission, regardless if a project exists or not. You can check that by giving the permission to the non-member and then checking all permissions with `User.all_permissions_for(nil)`. This returns all permission that the `non-member` role has.

I am fine with changing the behavior back to the previous implementation (which is more correct, as you state), but then we would also start hiding menu items, that were previously shown, until at least one product is there.

klaustopher: I have added the additional check that non-member and anonymous roles are only checked if there are no projects. Eventhough this does not 100% match with the current implementation of what is returned in the `allowed_to_globally?` method, but I think this way it is a bit more correct.

klaustopher: As a bit more context:

```ruby

[7] pry(main)> Project.where(public: true).count
  Project Count (1.4ms)  SELECT COUNT(*) FROM "projects" WHERE "projects"."public" = $1  [["public", true]]
=> 0

[8] pry(main)> Authorization.roles(User.find(5), nil)
  User Load (0.5ms)  SELECT "users".* FROM "users" WHERE "users"."type" IN ($1, $2) AND "users"."id" = $3 LIMIT $4  [["type", "User"], ["type", "AnonymousUser"], ["id", 5], ["LIMIT", 1]]
  Role Load (1.4ms)  SELECT "roles".* FROM "roles" WHERE "roles"."id" IN (SELECT "roles"."id" FROM "users" LEFT OUTER JOIN "members" ON "users"."id" = "members"."user_id" AND "members"."entity_type" IS NULL AND "members"."entity_id" IS NULL LEFT OUTER JOIN "member_roles" ON "members"."id" = "member_roles"."member_id" LEFT OUTER JOIN "roles" ON ("roles"."id" = "member_roles"."role_id" OR "roles"."builtin" = 1) WHERE "users"."id" = 5 GROUP BY "roles"."id")
  RolePermission Load (0.4ms)  SELECT "role_permissions".* FROM "role_permissions" WHERE "role_permissions"."role_id" IN ($1, $2, $3, $4)  [["role_id", 9], ["role_id", 1], ["role_id", 4], ["role_id", 5]]
=> [#<GlobalRole:0x000000011ac6a460
  id: 9,
  name: "Staff and projects manager",
  position: 6,
  builtin: 0,
  type: "GlobalRole",
  created_at: Wed, 13 Sep 2023 09:04:10.703167000 UTC +00:00,
  updated_at: Wed, 13 Sep 2023 09:04:10.703167000 UTC +00:00>,
 #<ProjectRole:0x000000010bac9e30
  id: 1,
  name: "Non member",
  position: 1,
  builtin: 1,
  type: "ProjectRole",
  created_at: Wed, 13 Sep 2023 09:04:10.385863000 UTC +00:00,
  updated_at: Wed, 13 Sep 2023 09:04:10.385863000 UTC +00:00>,
 #<ProjectRole:0x000000010bac9d90
  id: 4,
  name: "Reader",
  position: 4,
  builtin: 0,
  type: "ProjectRole",
  created_at: Wed, 13 Sep 2023 09:04:10.486629000 UTC +00:00,
  updated_at: Wed, 13 Sep 2023 09:04:10.486629000 UTC +00:00>,
 #<ProjectRole:0x000000010bac9cf0
  id: 5,
  name: "Project admin",
  position: 5,
  builtin: 0,
  type: "ProjectRole",
  created_at: Wed, 13 Sep 2023 09:04:10.527569000 UTC +00:00,
  updated_at: Wed, 13 Sep 2023 09:04:10.527569000 UTC +00:00>]
```

There's no public projects in the DB, but when i ask for the user's roles for `nil` as the project/context, I get all roles that potentially maps to the the user including those for the non-member role.

But I think the way how it is now is the better way to do it.

klaustopher: Discussed with @ulferts. We know that behavior is changing. But we think it is more correct that `allowed_in_any_project?(permission)` means that there must be a project, where the user has the permission. And not "would the user theoretically have the permission via the non-member role, regardless if the project exists" like it is defined today.

ulferts (approved):
LGTM 🎈

PR #13947, opened 2023-10-19
Added possibilty to add tenant and drive ID for a OneDrive storage

apfohl:
https://community.openproject.org/projects/openproject/work_packages/50141

This PR is implementing the function of providing the Tenant ID and the Drive ID during the setup of a OneDrive/Sharepoint storage.

- A new page between the storage type selection and OAuth settings
- Contract for validating that TenantId and DriveId are UUIDs (solved with a regex)
- `before_action` for parsing the parameter for the update action to get the drive id

### DIFF
--- a/modules/storages/app/contracts/storages/storages/one_drive_contract.rb
+++ b/modules/storages/app/contracts/storages/storages/one_drive_contract.rb
@@ -32,5 +32,9 @@ module Storages::Storages
   class OneDriveContract < ::ModelContract
     attribute :host
     validates :host, absence: true
+    attribute :tenant_id
+    validates :tenant_id, format: { with: /\A(?:[a-f0-9]{8}-[a-f0-9]{4}-[a-f0-9]{4}-[a-f0-9]{4}-[a-f0-9]{12}|consumers)\z/i }
+    attribute :drive_id
+    validates :drive_id, format: { with: /\A[a-f0-9]{8}-[a-f0-9]{4}-[a-f0-9]{4}-[a-f0-9]{4}-[a-f0-9]{12}\z/i }, allow_nil: true
   end
 end
--- a/modules/storages/app/controllers/storages/admin/storages_controller.rb
+++ b/modules/storages/app/controllers/storages/admin/storages_controller.rb
@@ -42,6 +42,7 @@ class Storages::Admin::StoragesController < ApplicationController
   # and set the @<controller_name> variable to the object referenced in the URL.
   before_action :require_admin
   before_action :find_model_object, only: %i[show destroy edit update replace_oauth_application]
+  before_action :parse_drive_id, only: %i[update]
 
   # menu_item is defined in the Redmine::MenuManager::MenuController
   # module, included from ApplicationController.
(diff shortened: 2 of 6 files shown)


modules/storages/app/contracts/storages/storages/one_drive_contract.rb
```diff
@@ -32,5 +32,9 @@ module Storages::Storages
   class OneDriveContract < ::ModelContract
     attribute :host
     validates :host, absence: true
+    attribute :tenant_id
+    validates :tenant_id, format: { with: /\A(?:[a-f0-9]{8}-[a-f0-9]{4}-[a-f0-9]{4}-[a-f0-9]{4}-[a-f0-9]{12}|consumers)\z/i }
+    attribute :drive_id
+    validates :drive_id, format: { with: /\A[a-f0-9]{8}-[a-f0-9]{4}-[a-f0-9]{4}-[a-f0-9]{4}-[a-f0-9]{12}\z/i }, allow_nil: true
```
Kharonus, line 38: 🔴 why is the drive id allowed to be `nil`? at least some of our queries will fail on a `nil` value here.

modules/storages/app/controllers/storages/admin/storages_controller.rb
```diff
@@ -171,6 +172,16 @@ def show_local_breadcrumb
 
   private
 
+  def parse_drive_id
```
Kharonus: 🟡 just naming: ^^

`parse_drive_id` is invalid, if we are viewing this controller from a nc storage perspective. So, the question, can we extract the condition to the `before_action` (or is `@storage` not available there)? 

At least, we should rename the method here to something like: `prepare_update_params`, to be less OD specific.

modules/storages/config/locales/en.yml
```diff
@@ -126,6 +126,10 @@ en:
           Copy the client secret from the Azure portal. For a newly created application the secret first needs to be
           created manually. For authorization of web applications a secret is mandatory.
         missing_client_id_for_redirect_uri: "Client ID missing to provide redirect URI."
+        tenant_id: >
+          Please insert the tenant ID you got from your SharePoint administrator.
+        drive_id: >
+          The drive ID can be obtained by you SharePoint Administrator.
```
Kharonus: 🟡 `Administrator` -> `administrator`

Kharonus (changes requested):

modules/storages/app/contracts/storages/storages/one_drive_contract.rb
apfohl: This is due to the fact, how the solution is build. The contract is evaluated on both `create` and `update`. Therefor and because the drive ID is inserted via the `update` action on the second `page`, the `create` action is called with an empty drive ID. That is of cause not ideal, but it will also be reworked with the refactoring of the UI to Primer.

modules/storages/app/controllers/storages/admin/storages_controller.rb
apfohl: `parse_drive_id` is called as a `before_action` and `@storage` is available. But I don't understand the question here.

I agree on the naming. I'll change it.

modules/storages/config/locales/en.yml
apfohl: good.

modules/storages/app/contracts/storages/storages/one_drive_contract.rb
Kharonus: Okay, that is a bit cracked, but I'd accept it due to the intermediate state this implementation has (it is to be replaced by primer form very soon).

Kharonus (approved):

PR #13950, opened 2023-10-19
Feature 50526, support for non-open versions in version custom fields

virgo47:
Proposed changes for: https://community.openproject.org/projects/openproject/work_packages/50526/activity

Eventually, I named the CF column/attribute `allow_non_open_versions`. It's shorter than `...locked_and_closed...` and more inclusive. I also left the word "allow" there, as without it it didn't imply the meaning as well (open for discussion, of course).

This option is available for CFs of type Version, but obviously not for others. In this spirit, I added one rather talkative test to `custom_fields_spec.rb` because I couldn't find a similar one.

When enabled, edit offers also locked and closed versions as values - this works for both single and multi-value CFs. This is covered by the new `non_open_version_custom_field_spec.rb`.


app/models/custom_field.rb
```diff
@@ -281,6 +285,14 @@ def multi_value_possible?
       [ProjectCustomField, WorkPackageCustomField].include?(self.class)
   end
 
+  def allow_non_open_versions?
+    allow_non_open_versions
+  end
+
+  def allow_non_open_versions_possible?
+    version? && [ProjectCustomField, WorkPackageCustomField].include?(self.class)
```
ulferts: Why support the feature also for projects? Is this a drive by gain? I am mostly asking because you hadn't included projects in the specifying work package so I am not sure about the reason for adding it. And why isn't it allowed for `TimeEntryCustomField` which can also custom fields of type version?

That being said, I also don't see a reason not to have it on `Project` I would just like to know the reason for the selection.

app/models/project.rb
```diff
@@ -274,6 +274,12 @@ def assignable_versions
     @assignable_versions ||= shared_versions.references(:project).with_status_open.order_by_semver_name.to_a
   end
 
+  # Returns all versions, including closed and locked, as an array.
+  # This is useful for custom fields allowing non-open versions as input values.
+  def all_versions
+    @all_versions ||= shared_versions.references(:project).order_by_semver_name.to_a
+  end
```
ulferts, line 282: I think the name of the method can be improved. On the level of the project model, having a method `all_versions` can very easily be confused with `rolled_up_versions` and `shared_versions`. To me, the fact that the caller wants to have non open versions included would be expressed best via a param of assignable_versions so we'd end up with `def assignable_versions(only_open: true)`

app/models/work_package.rb
```diff
@@ -262,6 +262,10 @@ def assignable_versions
     end
   end
 
+  def all_versions
+    @all_versions ||= project&.all_versions
+  end
+
```
ulferts: The remark I have for the project model applies here as well.

ulferts (changes requested):
First of all, thank you @virgo47 for this work. I think the functionality will come in handy.

I have smaller remarks on the code which you can find below.

It would also help if you could add some unit tests to:

* https://github.com/opf/openproject/blob/dev/spec/models/custom_value_spec.rb#L309-L325
* https://github.com/opf/openproject/blob/dev/spec/models/custom_value_spec.rb#L399-L708
* https://github.com/opf/openproject/blob/dev/spec/contracts/projects/shared_contract_examples.rb#L250-L264

app/models/custom_field.rb
virgo47: There was no reason not to support it on the other types, but I missed the `TimeEntryCustomField` because I realized it supports version CF after this change. I'd add that. The support for `all_version` is actually there, at least in `app/contracts/time_entries/base_contract.rb` because I encountered the problem there while developing it. At this moment it is likely dormant, but I'd allow it there as well. There is on reason not to, if anyone finds a usage for it.
So I wanted to have it whenever Version CF can be - for the parity's sake - I just didn't know about time entries using it when I added it into that array (one of the first changes). I'll add it, if you don't have objection to it.

app/models/project.rb
virgo47: I was considering exactly this! But I wasn't brave enough to mess with an existing method. :-) But I like this more and i will redo it. Thanks for the prod.

app/models/custom_field.rb
virgo47: One more note - the same problem is related to the code above for multi-value CFs (which is an omission from my previous PR #12948). I'll fix it both.

app/contracts/concerns/assignable_custom_field_values.rb
```diff
@@ -35,7 +35,7 @@ def assignable_custom_field_values(custom_field)
       when 'list'
         custom_field.possible_values
       when 'version'
-        assignable_versions
+        assignable_versions(only_open: !custom_field.allow_non_open_versions?)
```
ulferts, line 38: 👍

app/models/project.rb
ulferts: 👍 reads way nicer

ulferts (approved):
Once again, thanks @virgo47. 

The functionality looks very good now. Can be merged once the tests are fixed.

virgo47:
Thanks for the review @ulferts, I also added some low-level tests for the `allow_non_open_versions_possible?` method (omitted the `_possible` suffix by accident in the commit message).
However, I struggle to come up with any idea for the other tests you suggested, because they are mocked quite a lot. E.g. in:
````
context 'for a version custom field' do
  let(:custom_field) { build_stubbed(:version_project_custom_field) }
  let(:versions) { double('versions') }
````
I can replace the `let(:version)` by this:
````
let(:versions) { [build_stubbed(:version), build_stubbed(:version, status: 'closed')] }
````
And it doesn't change a bit, because it simply tests that whatever "goes in" (gets returned by the method on the Project) "gets out" on the other end of the contract. I feel like the integration tests cover the usage of `only_open` argument better - although I understand the distinction between unit and integration tests.
Do you have something concrete on your mind for the unit tests?

File: spec/models/custom_field_spec.rb [hunk 5]
```diff
@@ -448,7 +448,7 @@
     context 'with a project int cf' do
       let(:field) { build_stubbed(:integer_project_custom_field) }
 
-      it 'is true' do
+      it 'is false' do
```
ulferts, line 451: Thanks for fixing the descriptions here

ulferts (approved):
👍

virgo47:
Thanks for the support and merge. :-)

lindenthal:
Hi @virgo47, cool feature! 👍👎🎉👏

PR #13951, opened 2023-10-19
fix reversing the relation if necessary before validation

ulferts:
Reactivates inverting the relation if necessary before the validation takes place. This got broken by renaming the `BaseContract`'s super method. 

But this then leads to the problem that a relation can no longer be changing its direction, e.g. by turning a 'blocks' relation into a 'blocked by' relation, since the `WorkPackage.relatable` scope, when determining the work packages with which an origin work package is related, would include the relation to be reversed. To circumvent this, the modified relation needs to be excluded from the calculation as if it did not exist.

### DIFF
--- a/app/contracts/relations/base_contract.rb
+++ b/app/contracts/relations/base_contract.rb
@@ -46,7 +46,7 @@ module Relations
       Relation
     end
 
-    def validate!(*args)
+    def valid?(*args)
       # same as before_validation callback
       model.send(:reverse_if_needed)
 
@@ -65,7 +65,7 @@ module Relations
 
     def validate_nodes_relatable
       if (model.from_id_changed? || model.to_id_changed?) &&
-         WorkPackage.relatable(model.from, model.relation_type).where(id: model.to).empty?
+         WorkPackage.relatable(model.from, model.relation_type, ignore_relation: model).where(id: model.to).empty?
         errors.add :base, I18n.t(:'activerecord.errors.messages.circular_dependency')
       end
     end
--- a/app/contracts/relations/base_contract.rb
+++ b/app/contracts/relations/base_contract.rb
@@ -63,7 +63,7 @@ module Relations
 
     def validate_nodes_relatable
       if (model.from_id_changed? || model.to_id_changed?) &&
-         WorkPackage.relatable(model.from, model.relation_type, ignored_relation: model).where(id: model.to).empty?
+         WorkPackage.relatable(model.from, model.relation_type, ignored_relation: model).where(id: model.to_id).empty?
         errors.add :base, I18n.t(:'activerecord.errors.messages.circular_dependency')
       end
     end
--- a/app/contracts/relations/base_contract.rb
+++ b/app/contracts/relations/base_contract.rb
@@ -26,8 +26,6 @@
 # See COPYRIGHT and LICENSE files for more details.
 #++
 
-require 'model_contract'
-
 module Relations
   class BaseContract < ::ModelContract
     attribute :relation_type
@@ -65,7 +63,7 @@ module Relations
 
     def validate_nodes_relatable
       if (model.from_id_changed? || model.to_id_changed?) &&
-         WorkPackage.relatable(model.from, model.relation_type, ignore_relation: model).where(id: model.to).empty?
+         WorkPackage.relatable(model.from, model.relation_type, ignored_relation: model).where(id: model.to).empty?
         errors.add :base, I18n.t(:'activerecord.errors.messages.circular_dependency')
       end
     end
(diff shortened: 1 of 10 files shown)


app/models/work_packages/scopes/relatable.rb
```diff
@@ -356,6 +363,7 @@ def existing_relation_of_type_lateral(relation_type, limit_direction: false)
           WHERE (relations.#{direction2} = related.id AND relations.relation_type = :relation_type)
             AND NOT related.from_#{direction2}
             #{direction_limit ? "AND NOT #{direction_limit}" : ''}
+            #{ignored_relation&.id ? " AND NOT relations.id = #{ignored_relation.id}" : ''}
```
dombesz, line 366: 📝TIL: `NOT a = b` and `a != b` are equal in sql.

spec/contracts/relations/create_contract_spec.rb
```diff
@@ -27,51 +27,17 @@
...
-          .to receive_message_chain(:visible, :exists?)
-          .with(from.id)
-          .and_return(false)
-      end
-
-      it 'is invalid' do
-        expect(subject).not_to be_valid
-      end
-    end
+    subject(:contract) { described_class.new relation, current_user }
```
dombesz: It seems that both the `let(:relation)` and the `subject` could be declared inside the `spec/contracts/relations/shared_contract_examples.rb` and still work for the create and update contracts. 
Is better readability the intention for these declarations?

spec/contracts/relations/shared_contract_examples.rb
```diff
@@ -0,0 +1,144 @@
...
+      end
+    end
+
+    Relation::TYPES.keys.select { Relation::TYPES[_1][:reverse].present? }.each do |available_type|
+      context "when having the type '#{available_type}'" do
+        let(:relation_type) { available_type }
+
+        it_behaves_like 'contract is valid'
+      end
+    end
```
dombesz: The two contexts have the same expectation and the `available_types` are complimenting each other. If this is the correct expectation, then they can be combined into a single context by going through the `Relation::TYPES.keys` without the `select`. Is this also for improving readability?

spec/models/work_packages/scopes/relatable_spec.rb
```diff
@@ -293,7 +304,30 @@
...
+      end
+    end
+
+    context "for a 'blocks' relation and the existing relations being 'blocks' when ignoring origin`s relation" do
+      let(:relation_type) { Relation::TYPE_BLOCKS }
+      let(:directly_related_work_package_type) { Relation::TYPE_BLOCKS }
+      let(:transitively_related_work_package_type) { Relation::TYPE_BLOCKS }
+      let(:ignored_relation) { origin.relations.first }
+
+      it 'contains the the related work packages' do
```
dombesz:
```suggestion
      it 'contains the related work packages' do
```

spec/models/work_packages/scopes/relatable_spec.rb
```diff
@@ -150,6 +151,16 @@
             .to be_empty
         end
       end
+
+      context "with the existing relation and the queried being '#{current_type}' typed but explicitly ignoring the existing" do
+        let(:relation_type) { current_type }
+        let(:ignored_relation) { directly_related_work_package.relations.first }
+
+        it 'is empty' do
```
dombesz: There is a typo here, the spec should be called something like :

```suggestion
        it 'contains the directly related work package' do
```

spec/models/work_packages/scopes/relatable_spec.rb
```diff
@@ -1515,8 +1549,65 @@
...
+      let(:ignored_relation) { additional_direct_relation }
+
+      it 'is empty' do
+        expect(relatable)
+          .to be_empty
       end
     end
   end
+
+  context 'when ignoring anything else then a single relation' do
```
dombesz:
```suggestion
  context 'when ignoring anything else than a single relation' do
```

app/contracts/relations/base_contract.rb
```diff
@@ -65,7 +63,7 @@ def validate_to_exists
 
     def validate_nodes_relatable
       if (model.from_id_changed? || model.to_id_changed?) &&
-         WorkPackage.relatable(model.from, model.relation_type).where(id: model.to).empty?
+         WorkPackage.relatable(model.from, model.relation_type, ignored_relation: model).where(id: model.to).empty?
```
dombesz: Calling `model.to` will force loading the `to` relation resulting in an extra query. Using `model.to_id` will not result in an extra query.

dombesz (commented):
Just a few questions and observations.

ulferts:
Thanks for the review @dombesz. Your remarks are very welcomed.

dombesz (approved):
Nice work @ulferts on taking care of this complicated edge case of validation!👏

PR #13952, opened 2023-10-19
Apply errors from contract validations to model

ulferts:
Changes the contracts so that their error object is the error object of the model they validate. That way, all the error are in a single place, the model's error object.

This has the advantage of then being propagated up, through the service and controller layer to the view where errors are then displayed to the user. In particular, this helps with view form objects that rely on the model's errors to display errors next to the input.

The validation of the model layer now happens before the validation of the contracts because the default implementation of `ActiveModel:Validations#valid?` will clear the errors before validation. This no longer happens on the contract level. Because of the switch in order, before_validation callbacks defined on the model level no longer need to be triggered from the contract (as e.g. in  [821d48b](https://github.com/opf/openproject/pull/13952/commits/821d48b9db3f3283c78ec4c3f44dfbeeac94ea09))

As an improvement on the side, the no longer necessary error handling for `typed_dag` is removed since the gem is no longer part of the code base.


akabiru (approved):
Thanks Jens! 🥇

PR #13957, opened 2023-10-20
Reshape the error handling of the Nextcloud's GroupSyncService

mereghost:
https://community.openproject.org/work_packages/50894
https://community.openproject.org/work_packages/50616

## [Wat?](https://www.destroyallsoftware.com/talks/wat)

This PR aims to reduce the noise on AppSignal by reducing the number of raised exceptions when running.

## Why?

We keep reporting exceptions to AppSignal that are either non actionables or irrelevant, this adds noise and makes unearthing real glaring issues harder for everyone.

Besides, who does love getting AppSignal emails every 5 mins. :P

## How?

Most of the errors aren't something we can act immediately - like misconfiguration - nor are they exceptional so I started to log this info on the RoR log so, in case of failures we have just one point to look at.

All logs were made into JSON so, if structured logging is around, this becomes simpler to deal with (aaaaand it might help later on some other ideas on next steps).

Also in this PR, I extended the `ServiceResultRefinements` with the methods `result_or` and `result_and` (and their reversed forms `error_and` and `error_or`). Those 2 methods take in blocks and unwrap the `ServiceResult` `result` or `error` as needed.

```ruby
create_folder(folder) >> obtain_id >> update_storage
```

The line above calls 3 methods/lambdas taking the `ServiceResult.result` of each of then and executes then in order. How do you stop the code flow? By returning a failure.

```ruby
create_folder(folder)
  .result_and { |result| obtain_id(result) }
  .result_and {|result| update_storage(result) }
```

The line above is the same but written with the `result_and` method. If someone isn't versed in `monadic vernacular` I'd say it is slightly more legible.

But what I'd prefer is:

```ruby
create_folder(folder).result_or { |error| return error }
folder_id = obtain_folder_id(folder).result_or { |error| return error }
update_storage(folder_id).result_or { |error| return error }
```

As it conveys that I'm interested in the results, the context will return when met with errors and there's a clear order of argument passing. It is more verbose for sure, but allows us too keep cause and effect pretty close.

That being said, I've added it as an example and it can easily be converted to any other of the options we have if you folks feel it isn't great.

### DIFF
--- a/modules/storages/app/common/storages/peripherals/storage_interaction/nextcloud/set_permissions_command.rb
+++ b/modules/storages/app/common/storages/peripherals/storage_interaction/nextcloud/set_permissions_command.rb
@@ -1,3 +1,5 @@
+# frozen_string_literal: true
+
 #-- copyright
 # OpenProject is an open source project management software.
 # Copyright (C) 2012-2023 the OpenProject GmbH
@@ -90,16 +92,16 @@ module Storages::Peripherals::StorageInteraction::Nextcloud
       when Net::HTTPSuccess
         doc = Nokogiri::XML(response.body)
         if doc.xpath("/d:multistatus/d:response/d:propstat[d:status[text() = 'HTTP/1.1 200 OK']]/d:prop/nc:acl-list").present?
-          ServiceResult.success
+          ServiceResult.success(result: :success)
         else
           Util.error(:error, "nc:acl properly has not been set for #{path}")
         end
       when Net::HTTPNotFound
-        Util.error(:not_found)
+        Util.error(:not_found, 'Outbound request destination not found!', response)
       when Net::HTTPUnauthorized
-        Util.error(:unauthorized)
+        Util.error(:unauthorized, 'Outbound request not authorized!', response)
       else
-        Util.error(:error)
+        Util.error(:error, 'Outbound request failed!', response)
       end
     end
     # rubocop:enable Metrics/AbcSize
--- a/modules/storages/app/common/storages/peripherals/storage_interaction/nextcloud/set_permissions_command.rb
+++ b/modules/storages/app/common/storages/peripherals/storage_interaction/nextcloud/set_permissions_command.rb
@@ -97,11 +97,11 @@ module Storages::Peripherals::StorageInteraction::Nextcloud
           Util.error(:error, "nc:acl properly has not been set for #{path}")
         end
       when Net::HTTPNotFound
-        Util.error(:not_found, 'Outbound request destination not found!', response)
+        Util.error(:not_found, 'Outbound request destination not found', response)
       when Net::HTTPUnauthorized
-        Util.error(:unauthorized, 'Outbound request not authorized!', response)
+        Util.error(:unauthorized, 'Outbound request not authorized', response)
       else
-        Util.error(:error, 'Outbound request failed!', response)
+        Util.error(:error, 'Outbound request failed', response)
       end
     end
     # rubocop:enable Metrics/AbcSize
--- a/modules/storages/app/services/storages/group_folder_properties_sync_service.rb
+++ b/modules/storages/app/services/storages/group_folder_properties_sync_service.rb
@@ -101,7 +101,8 @@ module Storages
           OpenProject.logger.warn({ command: 'nextcloud.remove_user_from_group',
                                     group: @storage.group,
                                     user:,
-                                    message: error.log_message }.to_json)
+                                    message: error.log_message,
+                                    data: { status: error.data.code, body: error.data.body } }.to_json)
         end
       end
 
@@ -110,7 +111,8 @@ module Storages
           OpenProject.logger.warn({ command: 'nextcloud.add_users_to_group',
                                     group: @storage.group,
                                     user:,
-                                    message: error.log_message }.to_json)
+                                    message: error.log_message,
+                                    data: { status: error.data.code, body: error.data.body } }.to_json)
         end
       end
     end
@@ -149,7 +151,12 @@ module Storages
       Peripherals::Registry
         .resolve("commands.nextcloud.set_permissions")
         .call(storage: @storage, **command_params)
-        .result_or { |error| OpenProject.logger.warn "PROPPATCH #{path} #{error.log_message}. Response: #{error.data.inspect}" }
+        .result_or do |error|
+        OpenProject.logger.warn({ command: 'nextcloud.set_permissions',
+                                  folder: project_storage.project_folder_path,
+                                  message: error.log_message,
+                                  data: { status: error.data.code, body: error.data.body } }.to_json)
+      end
     end
 
     def project_tokens(project_storage)
@@ -179,12 +186,10 @@ module Storages
           .resolve("commands.nextcloud.set_permissions")
           .call(storage: @storage, **command_params)
           .result_or do |error|
-          error_msg = { command: 'nextcloud.set_permissions',
-                        folder: path,
-                        message: error.log_message,
-                        data: { status: error.data.code, body: error.data.body } }.to_json
-
-          OpenProject.logger.warn error_msg
+          OpenProject.logger.warn({ command: 'nextcloud.set_permissions',
+                                    folder: path,
+                                    message: error.log_message,
+                                    data: { status: error.data.code, body: error.data.body } }.to_json)
         end
       end
     end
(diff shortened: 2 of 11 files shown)


modules/storages/app/common/storages/peripherals/storage_interaction/nextcloud/set_permissions_command.rb
```diff
@@ -90,16 +92,16 @@ def call(path:, permissions:)
...
         end
       when Net::HTTPNotFound
-        Util.error(:not_found)
+        Util.error(:not_found, 'Outbound request destination not found!', response)
       when Net::HTTPUnauthorized
-        Util.error(:unauthorized)
+        Util.error(:unauthorized, 'Outbound request not authorized!', response)
       else
-        Util.error(:error)
+        Util.error(:error, 'Outbound request failed!', response)
```
Kharonus: Shouldn't we use `StorageError` here, along with error source and stuff like in the one drive queries?

modules/storages/spec/services/storages/group_folder_properties_sync_service_spec.rb
```diff
@@ -853,11 +837,11 @@
...
-        expect(OpenProject.logger).to have_received(:warn) do |msg, _|
-          expect(msg).to eq "Nextcloud user Darth Maul has not been removed from Nextcloud group " \
-                            "OpenProject: 'Failed to remove user Darth Maul from group OpenProject: Not viable to remove " \
-                            "user from the last group you are SubAdmin of'"
-        end
+        # expect(OpenProject.logger).to have_received(:warn) do |msg, _|
+        #   expect(msg).to eq "Nextcloud user Darth Maul has not been removed from Nextcloud group " \
+        #                     "OpenProject: 'Failed to remove user Darth Maul from group OpenProject: Not viable to remove " \
+        #                     "user from the last group you are SubAdmin of'"
+        # end
```
Kharonus: 🟡 missing a TODO or remove this?

modules/storages/app/common/storages/peripherals/storage_interaction/nextcloud/set_permissions_command.rb
mereghost: I think so, I didn't want to change this at this time but we need to standardize the classes, errors, messages etc.

modules/storages/spec/services/storages/group_folder_properties_sync_service_spec.rb
mereghost: I'm rewriting the error cases. =)

modules/storages/app/common/storages/peripherals/storage_interaction/nextcloud/set_permissions_command.rb
Kharonus: Yeah, I leave it to your judgement, if it would blow the scope now and if you feel well armed to tackle that foe right now. :P

Kharonus (approved):
LGTM

I find the reworked sync service readable and understandable. Any glimpses regarding functionality is better to be judged by @ba1ash, as he wrote the first iteration.

From my pov this one could be merged as is.

modules/storages/app/services/storages/group_folder_properties_sync_service.rb
```diff
@@ -28,266 +28,268 @@
...
+      add_remove_users_to_group
+    end
 
-  def rename_folder(source:, target:)
-    Storages::Peripherals::Registry.resolve("commands.#{@storage.short_provider_type}.rename_file")
-      .call(storage: @storage, source:, target:)
-      .on_failure(&failure_handler('rename_file_command', { source:, target: }))
-  end
+    def add_remove_users_to_group
+      remote_users = remote_group_users.result_or { puts :error }
```
ba1ash: Should we log it or have you planned it for next iteration?

modules/storages/app/services/storages/group_folder_properties_sync_service.rb
```diff
@@ -28,266 +28,268 @@
...
+                      folder: folder_path,
+                      data: { status: error.data.code, body: error.data.body } }.to_json
+
+        return OpenProject.logger.warn(error_msg)
       end
+
+      folder_ids = Peripherals::Registry
+                     .resolve('queries.nextcloud.file_ids')
+                     .call(storage: @storage, path: folder_path)
+                     .result_or { |_error| puts :error } # Log this
```
ba1ash: Probably should be logged as well.

modules/storages/app/services/storages/group_folder_properties_sync_service.rb
```diff
@@ -28,266 +28,268 @@
...
+        .call(storage: @storage, folder_path:)
+        .result_or do |error|
+        error_msg = { command: 'nextcloud.create_folder',
+                      folder: folder_path,
+                      data: { status: error.data.code, body: error.data.body } }.to_json
+
+        return OpenProject.logger.warn(error_msg)
       end
+
+      folder_ids = Peripherals::Registry
```
ba1ash: `folder_id` would be more appropriate name.

modules/storages/app/services/storages/group_folder_properties_sync_service.rb
```diff
@@ -28,266 +28,268 @@
...
+            # we need to stop as this would mess with the other processes
+            return OpenProject.logger.warn error_msg
+          end
+        end
+      end
     end
 
-    tokens_query
-  end
+    def move_folder(project_storage, current_name)
```
ba1ash: I suggest `rename_folder` as a method name.

modules/storages/app/services/storages/group_folder_properties_sync_service.rb
```diff
@@ -28,266 +28,268 @@
...
+
+        return OpenProject.logger.warn(error_msg)
       end
+
+      folder_ids = Peripherals::Registry
+                     .resolve('queries.nextcloud.file_ids')
+                     .call(storage: @storage, path: folder_path)
+                     .result_or { |_error| puts :error } # Log this
+
+      project_storage.update(project_folder_id: folder_ids.dig(folder_path, 'fileid'))
```
ba1ash: `folder_ids.dig(folder_path, 'fileid')` can return `nil` potentially and then we are left with nil project_folder_id. Do you think it is a problem? I am not sure, but maybe we should handle the nil case, because later it could be really not easy to debug if it happens.

modules/storages/app/services/storages/group_folder_properties_sync_service.rb
```diff
@@ -28,266 +28,268 @@
...
-      tokens_query = tokens_query.where(users: project.users)
+    def ensure_folders_exist(remote_folders)
+      id_folder_map = remote_folders.to_h { |folder, properties| [properties['fileid'], folder] }
+
+      active_project_storages_scope.map do |project_storage|
+        next create_folder(project_storage) unless id_folder_map.key?(project_storage.project_folder_id)
+
+        current_path = id_folder_map[project_storage.project_folder_id]
+        if current_path == project_storage.project_folder_path
+          project_storage.project_folder_id
```
ba1ash, line 211: It seems we do not need `project_folder_id` value up on the stack and it has been used as a boolean. Let's return a boolean explicitly, maybe?

modules/storages/app/services/storages/group_folder_properties_sync_service.rb
mereghost: HAHAHAHAHAHAHAHAH! I forgot about these cases. Yes, I need to add a log.

modules/storages/app/services/storages/group_folder_properties_sync_service.rb
mereghost: Good one.

modules/storages/app/services/storages/group_folder_properties_sync_service.rb
mereghost: If we update it with nil it won't be used later on so no harm done.

modules/storages/app/services/storages/group_folder_properties_sync_service.rb
mereghost: I'm sorry, but I'm kinda slow today. What do you mean in this case? Could you express this in code?

ba1ash: Nevermind. After having a closer look I see that what I suggested does not make sense. Current implementation is fine :+1:

ba1ash (approved):
LGTM :+1:

PR #13962, opened 2023-10-20
feat[#50499]: Add navigational elements (breadcrumb and back buttons)

dominic-braeunlein:
Work package: https://community.openproject.org/wp/50499

This PR adds '`OpPrimer::BackButtonComponent`' in `app/components/op_primer` 
This button inherits from Primer::Beta::IconButton and needs only the `href` as parameter.

This PR also adds this BackButton and breadcrumbs in storage edit view.
The breadcrumbs are outside of the header because `Primer::OpenProject::PageHeader` doesn't have a `slot` for breadcumbs.

There is a `left-margin` problem that I need to clarify with @HDinger (see picture below). Or maybe @akabiru you know too where the `margin-left: 2rem;` for `<ol>` tags should be overridden.  

Screenshot 2023-10-20 at 14 34 56

I will clarify this before merging

### DIFF
--- a/app/helpers/breadcrumb_helper.rb
+++ b/app/helpers/breadcrumb_helper.rb
@@ -30,13 +30,13 @@ module BreadcrumbHelper
   def full_breadcrumbs
     items = breadcrumb_paths.compact
     render(Primer::Beta::Breadcrumbs.new(data: { 'test-selector': "op-breadcrumb" })) do |breadcrumbs|
-      items.each do |item|
+      items.each_with_index do |item, index|
         item = anchor_string_to_object(item) if item.is_a?(String) && item.start_with?("\u003c")
 
         if item.is_a?(Hash)
-          breadcrumbs.with_item(href: item[:href]) { item[:text] }
+          breadcrumbs.with_item(href: item[:href], classes: index == 0 ? "first-breadcrumb-element" : nil) { item[:text] }
         else
-          breadcrumbs.with_item(href: "#") { item }
+          breadcrumbs.with_item(href: "#", classes: index == 0 ? "first-breadcrumb-element" : nil) { item }
         end
       end
     end
--- a/app/helpers/breadcrumb_helper.rb
+++ b/app/helpers/breadcrumb_helper.rb
@@ -43,9 +43,7 @@ module BreadcrumbHelper
   end
 
   def breadcrumb_paths(*args)
-    if args.nil?
-      nil
-    elsif args.empty?
+    if args.empty?
       @breadcrumb_paths ||= [default_breadcrumb]
     else
       @breadcrumb_paths ||= []
--- a/app/helpers/breadcrumb_helper.rb
+++ b/app/helpers/breadcrumb_helper.rb
@@ -29,7 +29,7 @@
 module BreadcrumbHelper
   def full_breadcrumbs
     items = breadcrumb_paths.compact
-    render(Primer::Beta::Breadcrumbs.new(data: { 'test-selector': "op-breadcrumb" })) do |breadcrumbs|
+    render(Primer::Beta::Breadcrumbs.new(test_selector: "op-breadcrumb" )) do |breadcrumbs|
       items.each_with_index do |item, index|
         item = anchor_string_to_object(item) if item.is_a?(String) && item.start_with?("\u003c")
 
--- a/app/helpers/breadcrumb_helper.rb
+++ b/app/helpers/breadcrumb_helper.rb
@@ -27,30 +27,19 @@
 #++
 
 module BreadcrumbHelper
-  def full_breadcrumb
-    breadcrumb_list(*breadcrumb_paths)
-  end
+  def full_breadcrumbs
+    items = breadcrumb_paths.compact
+    render(Primer::Beta::Breadcrumbs.new) do |breadcrumbs|
+      items.each do |item|
+        item = anchor_string_to_object(item) if item.is_a?(String) && item.start_with?("\u003c")
 
-  def breadcrumb(*args)
-    elements = args.flatten
-    elements.any? ? content_tag('p', (args.join(' &#187; ') + ' &#187; ').html_safe, class: 'op-breadcrumb') : nil
-  end
-
-  def breadcrumb_list(*args)
-    elements = args.flatten
-    breadcrumb_elements = [content_tag(:li,
-                                       elements.shift.to_s,
-                                       class: 'first-breadcrumb-element')]
-
-    breadcrumb_elements += elements.map do |element|
-      if element
-        content_tag(:li,
-                    h(element.to_s),
-                    class: "icon4 icon-small icon-arrow-right5")
+        if item.is_a?(Hash)
+          breadcrumbs.with_item(href: item[:href]) { item[:text] }
+        else
+          breadcrumbs.with_item(href: "#") { item }
+        end
       end
     end
-
-    content_tag(:ul, breadcrumb_elements.join.html_safe, class: 'op-breadcrumb', 'data-test-selector': 'op-breadcrumb')
   end
 
   def breadcrumb_paths(*args)
@@ -70,5 +59,17 @@ module BreadcrumbHelper
     else
       false
     end
+  end
+
+  private
+
+  # transform anchor tag strings to {href, text} objects
+  # e.g "\u003ca href=\"/admin\"\u003eAdministration\u003c/a\u003e"
+  def anchor_string_to_object(html_string)
+    # Parse the HTML
+    doc = Nokogiri::HTML.fragment(html_string)
+    # Extract href and text
+    anchor = doc.at("a")
+    { href: anchor["href"], text: anchor.text }
   end
 end
--- a/app/helpers/breadcrumb_helper.rb
+++ b/app/helpers/breadcrumb_helper.rb
@@ -29,7 +29,7 @@
 module BreadcrumbHelper
   def full_breadcrumbs
     items = breadcrumb_paths.compact
-    render(Primer::Beta::Breadcrumbs.new) do |breadcrumbs|
+    render(Primer::Beta::Breadcrumbs.new(data: { 'test-selector': "op-breadcrumb" })) do |breadcrumbs|
       items.each do |item|
         item = anchor_string_to_object(item) if item.is_a?(String) && item.start_with?("\u003c")
 
--- a/modules/storages/app/views/storages/admin/storages/edit.html.erb
+++ b/modules/storages/app/views/storages/admin/storages/edit.html.erb
@@ -28,6 +28,7 @@ See COPYRIGHT and LICENSE files for more details.
 ++#%>
 
 <% html_title t(:label_administration), t("project_module_storages"), t('label_edit_x', x: @storage.name) %>
+<% local_assigns[:additional_breadcrumb] = @storage.name %>
 
 <% if OpenProject::FeatureDecisions.storage_primer_design_active? %>
   <%= render(Primer::OpenProject::PageHeader.new) do |header| %>
@@ -37,6 +38,8 @@ See COPYRIGHT and LICENSE files for more details.
       <% end %>
     <% end %>
 
+    <% header.with_back_button(href: admin_settings_storages_path, 'aria-label':  I18n.t("button_back")) %>
+   
     <% header.with_actions do %>
       <%=
         primer_form_with(
@@ -64,7 +67,6 @@ See COPYRIGHT and LICENSE files for more details.
 
   <%= render(::Storages::Admin::StorageViewComponent.new(@storage)) %>
 <% else %>
-  <% local_assigns[:additional_breadcrumb] = @storage.name %>
   <%= toolbar title: t('label_edit_x', x: @storage.name) %>
 
   <%= render(::Storages::Admin::ConfigurationChecksComponent.new(storage: @storage)) %>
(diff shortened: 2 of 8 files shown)


HDinger:
Hi @dominic-braeunlein 

we have a file called [primer-adjustements.sass](https://github.com/opf/openproject/blob/dev/frontend/src/global_styles/openproject/_primer-adjustments.sass) where you can add the override for the `ol` tag inside the Breadcrumb. We already do something similar with lists inside the `Box` component.

With regards to the Breadcrumb and the BackButton being outside of the PageHeader: I am curious whether you discussed the option to add these elements directly to the `PageHeader` component itself, instead of writing them next to it? At first glance it would be worth the effort, as this is not the only page that will need this, and the custom backButton would then be obsolete as well. @akabiru how do you think about it?

app/components/op_primer/back_button_component.rb
```diff
@@ -0,0 +1,44 @@
...
+        href: href,
+        icon: "arrow-left",
+        scheme: :invisible,
+        size: :medium,
+        tag: :a,
+        aria: { label: I18n.t("button_back") }
+      )
+    end
+  end
+end
```
akabiru: 🟠 One convention we have (https://github.com/opf/openproject/pull/13769#issuecomment-1736801115) is we opted out of extending Primer components directly because this opens us up to breaking changes, if primer base classes are changed. Instead, we compose primer components, extending our own [`ApplicationComponent`](https://github.com/opf/openproject/blob/dev/app/components/application_component.rb) which is a ViewComponent

```ruby
# app/components/op_primer/back_button_component.rb
module OpPrimer
  class BackButtonComponent < ApplicationComponent
    options icon: "arrow-left",
            size: :medium

    def button_options
      { href: options[:href],
        icon: "arrow-left",
        scheme: :invisible,
        size:,
        tag: :a,
        aria: { label: I18n.t("button_back") } }
    end
  end
end
```

```ruby
# app/components/op_primer/back_button_component.html.erb
<%=
  render(Primer::Beta::IconButton.new(**button_options))
%>

```

modules/storages/app/views/storages/admin/storages/edit.html.erb
```diff
@@ -29,9 +29,21 @@ See COPYRIGHT and LICENSE files for more details.
 
 <% html_title t(:label_administration), t("project_module_storages"), t('label_edit_x', x: @storage.name) %>
 
+
+
+
```
akabiru:
```suggestion
```

modules/storages/app/views/storages/admin/storages/edit.html.erb
```diff
@@ -29,9 +29,21 @@ See COPYRIGHT and LICENSE files for more details.
...
+    
 <% if OpenProject::FeatureDecisions.storage_primer_design_active? %>
+
+  <%= render(Primer::Beta::Breadcrumbs.new) do |component|
+    component.with_item(href: "admin/settings") { "Administration" }
+    component.with_item(href: "admin/settings/storages") { "File Storages" }
+    component.with_item(href: "admin/settings/storages/#{@storage.id}/edit") { @storage.name }
+  end %>
```
akabiru: 🟠 Agree with @HDinger https://github.com/opf/openproject/pull/13962#issuecomment-1772761276 that we should make this component more generic. Two topics to keep in mind:

- Generally text is translated, so `Administration`,  `File Storages` should be read from `I18n`
- We should consider harmonizing the new `BreadcrumbComponent` with the [`BreadcrumbHelper#breadcrumb_paths`](https://github.com/opf/openproject/blob/dev/app/helpers/breadcrumb_helper.rb#L56) similar to the current convention
- BreadcrumbHelper allows a `#show_local_breadcrumb` control method in controllers that is toggled on demand. E.g. [see matches](https://github.com/search?q=repo%3Aopf%2Fopenproject%20show_local_breadcrumb&type=code) I imagine we want to maintain this contract for backwards compatibility

modules/storages/app/views/storages/admin/storages/edit.html.erb
```diff
@@ -29,9 +29,21 @@ See COPYRIGHT and LICENSE files for more details.
...
+
   <%= render(Primer::OpenProject::PageHeader.new) do |header| %>
+
     <% header.with_title(test_selector: 'storage-name-title') do %>
+      <%= render(OpPrimer::BackButtonComponent.new(href: url_for(controller: '/storages/admin/storages', action: :index))) %>
```
akabiru:
```suggestion
      <%= render(OpPrimer::BackButtonComponent.new(href: admin_settings_storages_path) %>
```

🟢 We generally use named `Rails.application.url_helpers` path names- syntactic sugar on top of `url_for` https://guides.rubyonrails.org/routing.html#path-and-url-helpers

💡  To list only storages urls, the `bin/rails routes` cli command supports grepping functionality

i.e. you can pick any of the prefixes and append `path`, to generate a URL
```
$ bin/rails routes -g storages

                                                           Prefix Verb   URI Pattern                                                                               Controller#Action
                          new_admin_settings_storage_oauth_client GET    /admin/settings/storages/:storage_id/oauth_client/new(.:format)                           storages/admin/oauth_clients#new
                              admin_settings_storage_oauth_client POST   /admin/settings/storages/:storage_id/oauth_client(.:format)                               storages/admin/oauth_clients#create
 new_admin_settings_storage_automatically_managed_project_folders GET    /admin/settings/storages/:storage_id/automatically_managed_project_folders/new(.:format)  storages/admin/automatically_managed_project_folders#new
edit_admin_settings_storage_automatically_managed_project_folders GET    /admin/settings/storages/:storage_id/automatically_managed_project_folders/edit(.:format) storages/admin/automatically_managed_project_folders#edit
     admin_settings_storage_automatically_managed_project_folders PATCH  /admin/settings/storages/:storage_id/automatically_managed_project_folders(.:format)      storages/admin/automatically_managed_project_folders#update
                                                                  PUT    /admin/settings/storages/:storage_id/automatically_managed_project_folders(.:format)      storages/admin/automatically_managed_project_folders#update
                                 edit_host_admin_settings_storage GET    /admin/settings/storages/:id/edit_host(.:format)                                          storages/admin/storages#edit_host
                 replace_oauth_application_admin_settings_storage DELETE /admin/settings/storages/:id/replace_oauth_application(.:format)                          storages/admin/storages#replace_oauth_application
                                          admin_settings_storages GET    /admin/settings/storages(.:format)                                                        storages/admin/storages#index
                                                                  POST   /admin/settings/storages(.:format)                                                        storages/admin/storages#create
                                       new_admin_settings_storage GET    /admin/settings/storages/new(.:format)                                                    storages/admin/storages#new
                                      edit_admin_settings_storage GET    /admin/settings/storages/:id/edit(.:format)                                               storages/admin/storages#edit
                                           admin_settings_storage PATCH  /admin/settings/storages/:id(.:format)                                                    storages/admin/storages#update
                                                                  PUT    /admin/settings/storages/:id(.:format)                                                    storages/admin/storages#update
                                                                  DELETE /admin/settings/storages/:id(.:format)                                                    storages/admin/storages#destroy
                 confirm_destroy_project_settings_project_storage GET    /projects/:project_id/settings/project_storages/:id/destroy_info(.:format)                storages/admin/project_storages#destroy_info
                         project_settings_project_storage_members GET    /projects/:project_id/settings/project_storages/:project_storage_id/members(.:format)     storages/project_settings/project_storage_members#index
                                project_settings_project_storages GET    /projects/:project_id/settings/project_storages(.:format)                                 storages/admin/project_storages#index
                                                                  POST   /projects/:project_id/settings/project_storages(.:format)                                 storages/admin/project_storages#create
                             new_project_settings_project_storage GET    /projects/:project_id/settings/project_storages/new(.:format)                             storages/admin/project_storages#new
                            edit_project_settings_project_storage GET    /projects/:project_id/settings/project_storages/:id/edit(.:format)                        storages/admin/project_storages#edit
                                 project_settings_project_storage PATCH  /projects/:project_id/settings/project_storages/:id(.:format)                             storages/admin/project_storages#update
                                                                  PUT    /projects/:project_id/settings/project_storages/:id(.:format)                             storages/admin/project_storages#update
                                                                  DELETE /projects/:project_id/settings/project_storages/:id(.:format)                             storages/admin/project_storages#destroy
```

akabiru (commented):
Great start @dominic-braeunlein 🙌🏾  A few remarks on general convention

🔴  For any UI interactions, we generally add feature tests. See https://github.com/opf/openproject/blob/dev/modules/storages/spec/features/admin_storages_spec.rb

---

fyi, I'm using @mereghost 's traffic light system for code review comments 😄 

> https://github.com/opf/openproject/pull/13223#pullrequestreview-1547996208
> 🟢 Just me being weird or minor suggestions
> 🟡 Points that warrant clarification discussion
> 🔴 (which I use sparingly) Major issues that need to be addressed before a merge

dominic-braeunlein:
I redid that PR from scratch.

The `BreadcrumbHelper.full_breadcrumb` method is replaced with one that renders `Primer::Beta::Breadcrumbs`.  
I renamed `full_breadcrumb` to `full_breadcrumbs`. The method is only used in `app/views/layouts/base.html.erb` as far as I can see.

@HDinger 
The margin-left from OpenProject is overridden in [_primer-adjustments.sass](https://github.com/opf/openproject/pull/13962/files#diff-ebc8270818b7f26cb6ecc2c0c6b4752069453d5b02a91f6f08c5738ac8387c3c) with following CSS. I am not sure if this selector is specific enough. What do you think?
```
nav
  > ol 
    margin-left: 0
```

@akabiru 
FYI: I have the same update problem with the storage name in the Breadcrumb as in this [WP-50738](https://community.openproject.org/projects/openproject/work_packages/50738/activity). I still have to work on that.
And I set `show_local_breadcrumb` in [storages_controller.rb](https://github.com/opf/openproject/pull/13962/files#diff-9fd520ca22b1e49857c2eb3f76cb5f13166a1fc140dacc62a6a5658c6c9e16c7) to `true` with this PR so they show again.


This PR should only be merged after https://github.com/opf/primer_view_components/pull/45 is merged.

HDinger:
> @HDinger The margin-left from OpenProject is overridden in [_primer-adjustments.sass](https://github.com/opf/openproject/pull/13962/files#diff-ebc8270818b7f26cb6ecc2c0c6b4752069453d5b02a91f6f08c5738ac8387c3c) with following CSS. I am not sure if this selector is specific enough. What do you think?
> 
> ```
> nav
>   > ol 
>     margin-left: 0
> ```


Hi @dominic-braeunlein 

I think you could limit this a bit by using:

```
 #breadcrumb
   ol 
      margin-left: 0
```

Thus we could be sure, that the effects are really limited to the breadcrumb and not other `nav` tags where we might not want that. 🤷‍♀️

app/helpers/breadcrumb_helper.rb
```diff
@@ -27,36 +27,23 @@
...
-  end
-
-  def breadcrumb_list(*args)
-    elements = args.flatten
-    breadcrumb_elements = [content_tag(:li,
-                                       elements.shift.to_s,
-                                       class: 'first-breadcrumb-element')]
+  def full_breadcrumbs
+    items = breadcrumb_paths.compact
+    render(Primer::Beta::Breadcrumbs.new(data: { 'test-selector': "op-breadcrumb" })) do |breadcrumbs|
```
akabiru: 💡  Primer providers a [`test_selector`](https://github.com/opf/primer_view_components/blob/main/app/components/primer/base_component.rb#L151) system argument that will only _"add the given value in **non-Production environments for testing purposes**"_

```suggestion
    render(Primer::Beta::Breadcrumbs.new(test_selector: "op-breadcrumb")) do |breadcrumbs|
```

Screenshot 2023-11-21 at 5 48 17 PM

app/helpers/breadcrumb_helper.rb
```diff
@@ -27,36 +27,23 @@
...
 
-    breadcrumb_elements += elements.map do |element|
-      if element
-        content_tag(:li,
-                    h(element.to_s),
-                    class: "icon4 icon-small icon-arrow-right5")
+        if item.is_a?(Hash)
+          breadcrumbs.with_item(href: item[:href], classes: index == 0 ? "first-breadcrumb-element" : nil) { item[:text] }
+        else
+          breadcrumbs.with_item(href: "#", classes: index == 0 ? "first-breadcrumb-element" : nil) { item }
```
akabiru, line 39: Q: Should we render these are links or normal text? Primer design doesn't seem to specify but I'm curious if this came up before 🤔 

The focus state seems misleading- and perhaps just raw text would be better
Screenshot 2023-11-21 at 6 01 08 PM

akabiru (approved):
Top work @dominic-braeunlein 👏🏾 🥇 

----

Just a few non-blocking remarks 👍🏾

app/helpers/breadcrumb_helper.rb
HDinger: I just doubled checked with Primer, and they are saying in their definition of the breadcrumbs that the last element should be a plain text. So I think we should stick to that as well.

Bildschirmfoto 2023-11-22 um 14 16 41

HDinger (commented):
Looks good to me 👍 🥳  Next to the things @akabiru menitoned you should probably rebase, as the primer bump has already been merged to dev.

app/helpers/breadcrumb_helper.rb
dominic-braeunlein: I agree, unfortunately the [Primer breadcrumbs implementation](https://github.com/opf/primer_view_components/blob/main/app/components/primer/beta/breadcrumbs.rb) doesn't allow for any text items.

I checked in the React implementation. It is also a link there but the CSS makes it look like a text by setting `pointer-events: none;` on the anchor:
https://primer.style/react/storybook/?path=/story/components-breadcrumbs--default&globals=colorScheme%3Alight

So I copied the approach and added that to `_primer-adjustments.sass`

HDinger: mhh.. again its really hard to understand Primer's intentions here. 🤷‍♀️ The override is fine for me for now. We should however keep an eye open to changes on the breadcrumb component in the future.

PR #13970, opened 2023-10-23
Permission Rework [7/x]: Move `allowed_to_in_project?` checks to new checks

klaustopher:
Fortunately we only had a few calls, so this is a rather short PR.


dombesz (approved):
LGTM👍 Not sure if this one should be merged, of wait for #13945 to be merged. Feel free to proceed.

PR #13971, opened 2023-10-23
fix: correct typo in example parameter

shaman-apprentice:
#  description for what it does and what it's intentions are

I was confused why you called the parameter `sortyBy` instead of `sortBy`. Until I noticed, that it is a typo in the example. So this PR corrects the typo with the intention to prevent following devs to copy paste the typo and wonder why the parameter has no effect.


dombesz:
Thanks @shaman-apprentice for your contribution! 👋

dombesz (approved):

PR #13973, opened 2023-10-23
[50652] Hide the sidebar in all tappable screens (tablet and mobile)

bsatarnejad:
https://community.openproject.org/projects/openproject/work_packages/50652/activity

### DIFF
--- a/frontend/src/app/core/browser/device.service.ts
+++ b/frontend/src/app/core/browser/device.service.ts
@@ -3,8 +3,13 @@ import { Injectable } from '@angular/core';
 @Injectable({ providedIn: 'root' })
 export class DeviceService {
   public mobileWidthTreshold = 544;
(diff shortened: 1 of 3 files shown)


HDinger (commented):
The higher threshold results in some weird effects, where the sidebar is already hidden but the content still has a lot of spacing because it is not changing to the mobile layout. 

for example: 544px (no mobile layout applied) 
Bildschirmfoto 2023-10-24 um 09 35 14

compared to: 543px (when the mobile breakpoint is reached)
Bildschirmfoto 2023-10-24 um 09 35 26

For me, the question is whether we want to switch completely to the Primer medium breakpoint (768px) as our general breakpoint now (cc @marcalcobe)

HDinger (approved):
I talked with @marcalcobe about this, and he agreed to accept the higher spacings.

PR #13986, opened 2023-10-24
[#50684] Migrate to httpx http ruby client

ba1ash:
https://community.openproject.org/work_packages/50684
It is a draft. I migrated nextcloud queries/command to use HTTPX.
- https://gitlab.com/os85/httpx
- An article with ruby HTTP clients comparison: https://honeyryderchuck.gitlab.io/2023/10/15/state-of-ruby-http-clients-use-httpx.html


mereghost:
I like it. Well... liking something more than `Net:HTTP` isn't that much of a high bar to set, but it looks nice.

mereghost (approved):
LGTM!

Nice work here, this lays down the groundwork for even bigger changes. \o/

PR #14010, opened 2023-10-26
Permission Rework [8/x]: Fix public permissions not being added

klaustopher:
In the current implementation we cache all roles the user has on a project and if the permission is allowed on any of them, we allow it. This check includes checking the public permissions as well:

https://github.com/opf/openproject/blob/8bf10659b80f1843bda5b39c4bcec61696727391/app/models/role.rb#L162-L164

With the new system we skip that step of caching the roles and cache the permissions directly. This lead to the public permissions being ignored. This PR fixes this.

---

After some back and forth with @ulferts we decided to look into adding the public permissions to the roles automatically (in seeds, set attribute service and via a migration)

### DIFF
--- a/app/models/users/permission_checks.rb
+++ b/app/models/users/permission_checks.rb
@@ -86,13 +86,18 @@ module Users::PermissionChecks
   end
 
   def all_permissions_for(context)
-    Authorization
+    permissions = Authorization
       .roles(self, context)
       .includes(:role_permissions)
       .pluck(:permission)
       .compact
-      .uniq
       .map(&:to_sym)
+
+    if context.is_a?(Project) || context.nil?
+      permissions + OpenProject::AccessControl.public_permissions.map(&:name)
+    else
+      permissions
+    end.uniq
   end
 
   # Old allowed_to? interface. Marked as deprecated, should be removed at some point ... Guessing 14.0?
--- a/app/models/users/permission_checks.rb
+++ b/app/models/users/permission_checks.rb
@@ -86,18 +86,13 @@ module Users::PermissionChecks
   end
 
   def all_permissions_for(context)
-    permissions = Authorization
+    Authorization
       .roles(self, context)
       .includes(:role_permissions)
       .pluck(:permission)
       .compact
       .map(&:to_sym)
-
-    if context.is_a?(Project) || context.nil?
-      permissions + OpenProject::AccessControl.public_permissions.map(&:name)
-    else
-      permissions
-    end.uniq
+      .uniq
   end
 
   # Old allowed_to? interface. Marked as deprecated, should be removed at some point ... Guessing 14.0?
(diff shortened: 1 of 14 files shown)


app/models/users/permission_checks.rb
```diff
@@ -86,13 +86,18 @@ def member_of?(project)
...
       .pluck(:permission)
       .compact
-      .uniq
       .map(&:to_sym)
+
+    if context.is_a?(Project) || context.nil?
+      permissions + OpenProject::AccessControl.public_permissions.map(&:name)
+    else
+      permissions
+    end.uniq
```
ulferts: There is one, potentially two, problems with this:
1) Even if the user has no permission in the project at all (e.g. not a member), will the public permissions be added.
2) There is no check on whether the modules the permission belongs to is active in the project. This might not be an actual problem if all callers check for this. I assume that they do as the problem is not specific to the now added public permissions. But please double check this.

ulferts (changes requested):
Hi @klaustopher , 

adding the public permissions definitely makes sense. I think it cannot be done in the way the PR currently does it though. Please see below.

For such a security sensitive change, it would also help to have additional positive and negative cases in the specs for the `UserPermissibleService.allowed_*` methods.

app/models/users/permission_checks.rb
klaustopher: > Even if the user has no permission in the project at all (e.g. not a member), will the public permissions be added.

Isn't that the case right now as well? When the user is not a member of the project, the non-member permission is returned and the public permissions are also added?

app/services/roles/set_attributes_service.rb
```diff
@@ -28,8 +28,16 @@
...
+      super(params)
+
+      if model.is_a?(ProjectRole)
+        model.permissions += OpenProject::AccessControl.public_permissions.map(&:name)
+      end
+    end
+
     def set_default_attributes(*)
-      model.permissions = ProjectRole.non_member.permissions if model.permissions.blank?
+      model.permissions = ProjectRole.non_member.permissions if model.permissions.blank? && model.is_a?(ProjectRole)
```
ulferts, line 40: 👍

spec/migrations/migrate_team_planner_permissions_spec.rb
```diff
@@ -26,8 +26,8 @@
 # See COPYRIGHT and LICENSE files for more details.
 #++
 
-require Rails.root.join("db/migrate/20220414085531_migrate_team_planner_permissions.rb")
 require 'spec_helper'
+require Rails.root.join("db/migrate/20220414085531_migrate_team_planner_permissions.rb")
```
klaustopher, line 30: When running the migration by itself, it failed because `Rails` was not defined yet. Switched the require order around

ulferts (approved):

PR #14020, opened 2023-10-26
[#50729] API schema is missing parameters to create memberships

Kharonus:
[OP#50729](https://community.openproject.org/work_packages/50729)

### What?

- reworked specs for endpoints behind /api/v3/memberships
- introduced model for schema property

### But... Why??

- better documentation of API behavior
- checked current API behavior, if still consistent


docs/api/apiv3/paths/memberships_form.yml
```diff
@@ -1,34 +1,51 @@
...
     '200':
       description: OK
-      headers: {}
+      content:
+        application/hal+json:
+          schema:
+            $ref: '../components/schemas/membership_form_model.yml'
+          examples:
+            form:
+              $ref: '../components/examples/membership-form-response.yml'
```
cbliard, line 32: As there is only one example, maybe it can be simplified to 
```
          example:
            $ref: '../components/examples/membership-form-response.yml'
```
or the example title could be more explicit:
```
          examples:
            'form response with a validation error on roles attribute':
              $ref: '../components/examples/membership-form-response.yml'
```

docs/api/apiv3/components/examples/membership-schema-response.yml
```diff
@@ -0,0 +1,67 @@
...
+  project:
+    type: 'Project'
+    name: 'Project'
+    required: false
+    hasDefault: false
+    writable: true
+    location: '_links'
+    _links:
+      allowedValues:
+        href: '/api/v3/memberships/available_projects?filters=%5B%7B%22principal%22%3A%7B%22operator%22%3A%22%21%22%2C%22values%22%3A%5B%2214%22%5D%7D%7D%5D'
```
cbliard, line 44: This does not correctly reflect the endpoint response: I tried it and the `project._links` is empty (no `allowedValue`). Same for `principal._links` and `roles._links`.

docs/api/apiv3/paths/memberships_available_projects.yml
```diff
@@ -1,97 +1,30 @@
 # /api/v3/memberships/available_projects
 ---
 get:
+  summary: Available projects for memberships
+  operationId: memberships_available_projects
```
cbliard: Did you consider prefixing it with `get_` like other operations?

docs/api/apiv3/paths/membership.yml
```diff
@@ -128,58 +93,62 @@ get:
...
-    required: true
-    schema:
-      type: integer
+    - name: id
+      description: Membership id
+      in: path
+      required: true
+      example: 1
+      schema:
+        type: integer
```
cbliard, line 116: Tthe `update_membership` operation should accept the same `requestBody` parameter as the form-update-membership endpoint:

```
  requestBody:
    content:
      application/json:
        schema:
          $ref: '../components/schemas/membership_write_model.yml'
        examples:
          'no notification':
            $ref: '../components/examples/membership-create-request-no-notification.yml'
          'custom message':
            $ref: '../components/examples/membership-create-request-custom-message.yml'
```

It would be nice to create specific update example instead of reusing the `create` examples, like 'update roles' to have 2 roles to a membership.

cbliard (changes requested):
Nice :+1: 

The update_membership endpoint is missing the `requestBody` parameter, preventing tools relying on the openapi spec to actually update something. 

This is the sole change request. The other comments are just remarks which can be safely ignored.

There is also the broader change of renaming operations like `view_xxx` or `list_xxx` to `get_xxx`. I do not have strong opinion about which name to choose, but I think the whole spec should be consistent. Maybe we should issue another PR to harmonize all operation ids so that the same pattern is used everywhere.

Aside from that, I really like that you took the time to reorder things to make description, name, summary, tags, and other relevant information appear first. It helps reading/understanding the yaml a lot.

docs/api/apiv3/paths/memberships_form.yml
Kharonus: that is a very good hint, and it took me a while to figure out, why I prefer `examples` over `example` in a [Media Type Object](https://spec.openapis.org/oas/latest.html#mediaTypeObject). The answer is a bit hidden inside this link.

`example` is of type `any`. Meaning, if I want to extract and reference it, the extracted object would no longer be an object of type [Example Object](https://spec.openapis.org/oas/latest.html#exampleObject).

So, my approach here is this:
- use `example` if I would inline any example object, as we are doing it for some error objects
- use `examples` if we extract an example object into a separate file

regarding the explicit naming: I admit, this is born out of laziness, but there is a truth hidden behind that. A explicit text could turn into a misleading, or wrong, description if one changes the example. This cannot happen in a generic text. This is, why I would tend to use explicit namings only to distinguish between multiple examples, because only then the reader of the spec will have a dropdown with those names for toggling the shown example.

What do you think about this?

docs/api/apiv3/paths/memberships_available_projects.yml
Kharonus: yes, good catch.

docs/api/apiv3/paths/membership.yml
Kharonus: really nice catch! thanks

docs/api/apiv3/components/examples/membership-schema-response.yml
Kharonus: I admit, that it can be confusing, but our API does only fill out `allowedValues` if there is already ANY value posted against the form endpoint. e.g. run a request `POST /api/v3/memberships/form` with the body

```json
{
  "_links": {
    "principal": {
      "href": "/api/v3/users/14"
    }
  }
}
```

and you will see those links.

docs/api/apiv3/paths/memberships_form.yml
cbliard: I don't mind about preferring `example` or `examples`. I saw somewhere else that you turned a `examples` into a `example`, so I thought that was a pattern you wanted to reproduce. Now I understand that it was isolated and simple enough and so inlining was better.

About naming examples, I think it is highly useful to understand what is being done. At first when reading the example, I thought that it was wrong because there were some validation errors and it was returning a 200 OK. I thought it was actually a "misleading, or wrong example" but then I read more carefully (and tried the endpoint locally with invalid data) and it is correct.

I think a good descriptive title helps the reader. That the goal of a doc. The "can-be-outdated" reason is also valid for the endpoint description, the endpoint summary, the error code explanations and so on. I don't think that's a good reason to avoid adding them. They provide value for the reader.

In fact, I started appreciating descriptive example names with the 3 different examples you introduced in the patch endpoint here: https://github.com/opf/openproject/pull/14020/files#diff-17a4cc3645bd5b987b7b6b3cb538a7d5dcde99a95ab603fafcfbf1055951c532R122-R128 (edit: well, it was on the create endpoint originally)

cbliard: But it's true that when there is only one example in an `examples`, the name is not shown in the rendered api documentation on the website or locally on dev instance. So the name does not really matter as it is not visible from the reader. I assumed it was.

If the name is not visible, then yes, no need to bother with it. Still that would be useful to have an example description for the reader instead of trying to infer what the example is trying to demonstrate.

docs/api/apiv3/components/examples/membership-schema-response.yml
cbliard: True for the form endpoints.

And this specific `membership-schema-response.yml` schema is used only to describe the `/api/v3/memberships/schema` endpoint, and on this endpoint, the returned `_links` are empty. Is it correct?

Also on the form-create and form-update endpoints, the schema are not quite the same as the project and principal are writable for create, but not for update.

docs/api/apiv3/paths/membership.yml
```diff
@@ -114,6 +114,18 @@ patch:
...
+      application/json:
+        schema:
+          $ref: '../components/schemas/membership_write_model.yml'
+        examples:
+          'global role':
+            $ref: '../components/examples/membership-create-request-global-role.yml'
+          'no notification':
+            $ref: '../components/examples/membership-create-request-no-notification.yml'
+          'custom message':
+            $ref: '../components/examples/membership-create-request-custom-message.yml'
```
cbliard: These examples will not work for the patch endpoint. Only roles is writable for the patch endpoint.

docs/api/apiv3/components/examples/membership-create-request-custom-message.yml
```diff
@@ -0,0 +1,13 @@
...
+    principal:
+      href: '/api/v3/users/14'
+    project:
+      href: '/api/v3/projects/21'
+    roles:
+      - href: '/api/v3/roles/4'
+  _meta:
+    notificationMessage:
+      format: markdown
+      raw: 'Execute Order *66*'
```
cbliard: Should be `You are granted permissions to execute order *66*` ;)

cbliard (changes requested):
Thanks for addressing my previous remarks.

My only request would be to update the examples of the update-membership endpoint as the examples from the create-membership can't be used: `principal` and `project` are not writable, only `roles` is.

Same applies for form-update-membership.

Could be interesting to have an example with multiple roles being assigned.

docs/api/apiv3/paths/membership.yml
Kharonus: very correct as usual, thanks for the hint!

docs/api/apiv3/paths/memberships_form.yml
Kharonus: This can also be achieved by added a `description` property to the example object itself. Do you think, this is needed by the examples introduced here?

docs/api/apiv3/components/examples/membership-schema-response.yml
Kharonus: Yes, I noticed that. But this is also the reason for those links not being `required`. But if the CAN be in there, they must be specified.

docs/api/apiv3/components/examples/membership-create-request-custom-message.yml
Kharonus: adapted

docs/api/apiv3/paths/membership.yml
Kharonus: - removed the required properties from the write model to make it more flexible
- added an update only example
- added descriptions to examples

docs/api/apiv3/paths/memberships_form.yml
cbliard: ah good to know.

docs/api/apiv3/components/examples/membership-schema-response.yml
cbliard: My point is only that the `/api/v3/memberships/schema` endpoint return empty `_links` for `roles`, `project`, and `principal`, and yet the example response has something there, so that is not consistent.

Either the example should be adapted to have `_links: {}` like the endpoint is returning, or the endpoint code should be updated to return something like `_links: { allowedValues: { href: 'foo' } }` but I do not know what is the right thing to do here.

Anyway, I think it is minor and the main issue of missing parameters to create / update memberships is now fixed.

docs/api/apiv3/components/examples/membership-create-request-custom-message.yml
```diff
@@ -10,4 +11,4 @@ value:
   _meta:
     notificationMessage:
       format: markdown
-      raw: 'Execute Order *66*'
+      raw: 'Granted permission to execute Order *66*.'
```
cbliard, line 14: :heart:

cbliard (approved):

cbliard:
Great! :+1:

PR #14031, opened 2023-10-27
Permission Rework [13/x]: Replace `allowed_to?` calls in controllers

klaustopher:
All controllers except the `ApplicationController` with the main authorization method

### DIFF
--- a/app/controllers/work_packages_controller.rb
+++ b/app/controllers/work_packages_controller.rb
@@ -128,11 +128,17 @@ class WorkPackagesController < ApplicationController
   end
 
   def protect_from_unauthorized_export
-    if (supported_list_formats + %w[atom]).include?(params[:format]) &&
-       !User.current.allowed_to?(:export_work_packages, @project, global: @project.nil?)
-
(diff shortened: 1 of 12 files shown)


app/controllers/work_packages_controller.rb
```diff
@@ -128,14 +128,20 @@ def authorize_on_work_package
   end
 
   def protect_from_unauthorized_export
-    if (supported_list_formats + %w[atom]).include?(params[:format]) &&
-       !User.current.allowed_to?(:export_work_packages, @project, global: @project.nil?)
-
+    if (supported_list_formats + %w[atom]).include?(params[:format]) && !user_allowed_to_export?
```
ulferts, line 131: Since you are introducing a method for this here, the name could also be inverted to something like `export_unauthorized?` as that would read better.

ulferts (approved):

app/controllers/work_packages_controller.rb
klaustopher: Personally, I see this 100% the other way. `if not_allowed_to_something?` reads completely out of place to me. `if !allowed_to_something` is clearer. But if you feel strongly about it, I can change it

PR #14035, opened 2023-10-27
Permission Rework [12/x]: Replace `allowed_to?` calls in Helpers

klaustopher:


modules/meeting/lib/open_project/meeting/engine.rb
```diff
@@ -92,7 +94,7 @@ class Engine < ::Rails::Engine
                    permissible_on: :project,
                    require: :member
         permission :send_meeting_minutes_notification,
-                   { meeting_minutes: [:notify] },
+                   { meeting_minutes: %i[notify icalendar] },
```
klaustopher: Currently the `icalendar` action in the `MeetingMinutesController` was not backed by any permission. I think the one here makes most sense, but please advice me to a different permission where it makes more sense

klaustopher: Checked with @oliverguenther and it would not make sense to add the permission, because the controller action is not backed by a route, so the link that would be added will lead nowhere. So we temporarily added catching the `UnknownPermissionError` in the helper method and returning false.

PR #14036, opened 2023-10-27
Fixed storage edit page heading change on rename

apfohl:
The name of the storage in the heading of the edit page needs to change when the storage is renamed.

Because we use Turbo for this now, the heading became a turbo frame and gets updated with a turbo stream with the response of the `update` action.

### DIFF
--- a/modules/storages/app/views/storages/admin/storages/update.turbo_stream.erb
+++ b/modules/storages/app/views/storages/admin/storages/update.turbo_stream.erb
@@ -1,0 +1,6 @@
+<%= turbo_stream.update ActionView::RecordIdentifier.dom_id(@storage, :storage_view_general_info) do %>
+  <%= render(::Storages::Admin::StorageViewComponent.new(@storage)) %>
+<% end %>
+<%= turbo_stream.update ActionView::RecordIdentifier.dom_id(@storage, :edit_storage_header) do %>
+  <%= @storage.name %>
+<% end %>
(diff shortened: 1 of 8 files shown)


modules/storages/app/views/storages/admin/storages/update.turbo_stream.erb
```diff
@@ -0,0 +1,6 @@
+<%= turbo_stream.update ActionView::RecordIdentifier.dom_id(@storage, :storage_view_general_info) do %>
+  <%= render(::Storages::Admin::StorageViewComponent.new(@storage)) %>
```
akabiru: It seems rendering complete `StorageViewComponent` renders- replaces it within general info frame 😬 

Screenshot 2023-10-30 at 1 12 45 PM

One option is to extract the general info into a component after all

akabiru (changes requested):
Coming along well 👍🏾  

- [x] ⚠️  re-rendering the `StorageViewComponent` renders the full component within the general info section
- [x] 🚧  Worth adding an assertion into the feature spec https://github.com/opf/openproject/blob/dev/modules/storages/spec/features/admin_storages_spec.rb#L110-L121

modules/storages/app/views/storages/admin/storages/update.turbo_stream.erb
```diff
@@ -0,0 +1,6 @@
...
+<% end %>
+<%= turbo_stream.update ActionView::RecordIdentifier.dom_id(@storage, :edit_storage_header) do %>
```
akabiru: 💡 we can do without the absolute namespacing as dom_id is available as a view helper method 😉 

```suggestion
<%= turbo_stream.update dom_id(@storage, :storage_view_general_info) do %>
  <%= render(::Storages::Admin::StorageViewComponent.new(@storage)) %>
<% end %>
<%= turbo_stream.update dom_id(@storage, :edit_storage_header) do %>
```

modules/storages/app/views/storages/admin/storages/update.turbo_stream.erb
apfohl: I'm already on to fixing that.

akabiru (approved):
🥇

PR #14053, opened 2023-10-30
[Op#50737] Convert new storage page to primer design

akabiru:
https://community.openproject.org/wp/50737

- [x] Select Provider form
     - [x] Add [Docs Link](https://www.openproject.org/docs/system-admin-guide/integrations/nextcloud/)
     - [x] Retrieve provider form on select change
     - [x] Switch select caption by provider type
     - [x] TODO: Create Task to add One Drive Storage Documentation (https://community.openproject.org/wp/50983) (https://github.com/opf/openproject/pull/14140)
- [x] Nextcloud Form
     - [x] 🐛  fix automatically managed validation errors
     - [x] 🔧  oauth client should return turbo stream partials, not the full layout html
     - [x] Do not render application password on edit
- [x] One Drive Form
- [x] Update Edit view with OneDrive context
     - [x] Fix configuration check for one drive general info
     - [x] Fix clipboard copy action
- ~Manage page refresh in new storage follow, maintain state~ 
- [x] Setup Task for Primer Flash (https://community.openproject.org/wp/50985)
- [x] Checkup Mobile View
   - [x] 🐛 Long text overflows  (https://community.openproject.org/wp/51008)

#### Select Provider 📹 

https://github.com/opf/openproject/assets/17295175/cc977b57-8867-4219-8c2d-c4782fd29c28

#### Create Nextcloud Storage 📹 

https://github.com/opf/openproject/assets/17295175/611dbe72-7be2-4ef3-9354-8874a71e51eb

#### One Drive Storage 📹 

https://github.com/opf/openproject/assets/17295175/a5788034-0b58-47b0-a221-8ccc0426f92a


akabiru (commented):
🚧 Bug logged for Mobile view overflow issues- to be resolved separately. https://community.openproject.org/wp/51008

frontend/src/stimulus/controllers/dynamic/storages/oauth-client-form.controller.ts
```diff
@@ -30,27 +30,53 @@
...
+
+  public toggleSubmitButtonDisabled():void {
+    const targetsConfigured = this.hasClientIdTarget && this.hasClientSecretTarget && this.hasSubmitButtonTarget;
+
+    if (!targetsConfigured) {
+      return;
+    }
 
-    this.redirectUriTarget.value = this.redirectUri();
+    if (this.clientIdTarget.value === '' || this.clientSecretTarget.value === '') {
```
HDinger, line 67: It is somehow inconsistent that you disable the button here but nowhere else. On the other places you let the user press the button and render the error message if there is a value missing. I think this would be fine here as well.

modules/storages/app/components/storages/admin/automatically_managed_project_folders_info_component.html.erb
```diff
@@ -0,0 +1,32 @@
...
+      concat(render(Primer::Beta::Text.new(font_weight: :semibold, mr: 1, test_selector: 'storage-managed-project-folders-label')) { I18n.t('storages.file_storage_view.automatically_managed_folders') })
+      concat(automatically_managed_project_folders_status_label)
+    end
+
+    grid.with_area(:description, tag: :div, color: :subtle, test_selector: 'storage-automatically-managed-project-folders-description') do
+      render(Primer::Beta::Truncate.new(font_weight: :light)) { I18n.t('storages.page_titles.managed_project_folders.subtitle_short') }
+    end
+
+    if editable_storage?
+      grid.with_area(:"icon-button", tag: :div, color: :subtle) do
```
HDinger: The color should be `:muted` accoriding to Figma. This is especially important for the high-contrast mode. (Same is true for the other icons on the right side)

modules/storages/app/components/storages/admin/oauth_client_info_component.html.erb
```diff
@@ -0,0 +1,36 @@
...
+      concat(
+        render(
+          Primer::Beta::Text.new(font_weight: :semibold, mr: 1, test_selector: 'storage-oauth-client-label')
+        ) { I18n.t("storages.file_storage_view.#{storage.short_provider_type}_oauth") }
+      )
+      concat(configuration_check_label_for(:storage_oauth_client_configured))
+    end
+
+    grid.with_area(:description, tag: :div, color: :subtle, test_selector: 'storage-oauth-client-id-description') do
+      render(Primer::Beta::Truncate.new(font_weight: :light)) { provider_oauth_client_description }
```
HDinger: This looks a bit different than specified in figma due to the `font_weight`. I noticed that "automatically managed folder" form as well.

modules/storages/app/components/storages/admin/oauth_client_info_component.html.erb
```diff
@@ -0,0 +1,36 @@
...
+    end
+
+    if editable_storage?
+      grid.with_area(:"icon-button", tag: :div, color: :subtle) do
```
HDinger: The button should have the color `:muted`

HDinger (commented):
Hi @akabiru 

great job, it works already quite nice! 🥳 @apfohl and I had a look together and only found some minor issues while clickling through:

- [x] From time to time the routing felt weird. I try to describe the cases where we expected something else:
  - [x] After I chose a type, and then press "Cancel", I expected to go back to the index page. Instead I was able to chose a type again. Further, I expected to be able to change the type afterwards in the select field. It felt weird, that this is not possible. From an accessibility point, it is also hard to communicate that the options of a select have changed.
  - [x] When closing any other block than the "general information", it is exactly reversed: I get redirected to the Index Page, although I would expect to stay on that page where I am. This is especially an issue when I try to go one step back and see the oauth credentials I already saved (see video below). This happens however only on the Nextcloud page. For Onedrive, it behaves like expected: closing a block, leads to the block being closed and there is no redirect.

https://github.com/opf/openproject/assets/7457313/7223b14a-85b5-4e56-ada6-dce724883588

  - [x] I sometines found it hard to get back the index page, as the breadcrumbs are missing and it is not intuitive that I have to press Save or Cancel in one of the boxes.. E.g. when I create a new OneDrive storage, enter the general information, but cancel the oauth credentials, I am on a page with no return.  This is however an issue on the edit pages as well. I think the breadcrumbs would solve that for all of them. (#13962)

Bildschirmfoto 2023-11-15 um 09 38 36

- [x] The box headers all have the wrong colour as they should be grey like on the index page
- [x] There are some places where the font-weight or the color does not match exactly the figma specification. I tried to point out those that I found, but maybe you could also keep an eye open for it.
- [x] From an accessibility point of view, I am wondering whether this "magical" appearance of new blocks after the type selection is catched by a screenreader 🤔 Do you know anything about whether the turbo responses are accessible out of the box? (https://community.openproject.org/wp/51021)


@apfohl and I also tried to have a look at the code, but due to the pure mass it was kind of hard for us to really get a good overview of it. IMaybe you could give us a short introdction and point out the most important parts?

frontend/src/stimulus/controllers/dynamic/storages/oauth-client-form.controller.ts
akabiru: Makes sense, I went by the designs here- but I agree that consistency is the better option here 👍🏾

modules/storages/app/components/storages/admin/automatically_managed_project_folders_info_component.html.erb
akabiru: Thanks for catching this! 👍🏾

modules/storages/app/components/storages/admin/forms/oauth_client_form_component.html.erb
```diff
@@ -2,23 +2,81 @@
...
+                          'input->storages--oauth-client-form#setRedirectUriValue'
+                }
+              },
+              client_secret_input_options: {
+                data: {
+                  'storages--oauth-client-form-target': 'clientSecret',
+                  action: 'input->storages--oauth-client-form#toggleSubmitButtonDisabled'
+                }
+              }
+            )
```
aaron-contreras, line 40: Is there a reason you're explicitly specifying the `input` event as the listener here? Stimulus provides an "Event Shorthand" notation which lets you rely on default events for different elements (unless you explicitly want to listen on a different event (e.g. `change`):

https://stimulus.hotwired.dev/reference/actions#event-shorthand

image

akabiru: Just out of preference as I find the explicit definitions more clear and easier to follow

aaron-contreras: Got it, thought it was some special case.

frontend/src/stimulus/controllers/dynamic/storages/oauth-client-form.controller.ts
akabiru: I spent (significant) time on this today and it's not a simple change as we'll have to rework the OAuth Clients service workflow to accommodate this behaviour. Hence, I broker it down in to a [CODE MAINTENANCE wp](https://community.openproject.org/wp/51033)

akabiru (commented):
Thank you for the detailed review @HDinger ! I've addressed majority of the issues and the larger topics assigned to work packages. As agreed, we'll go ahead and merge this first version. 🙇🏾

aaron-contreras:
Great work! @akabiru

akabiru:
> Great work! @akabiru

Thank you @aaron-contreras ! ❤️

PR #14063, opened 2023-10-31
Permission Rework [15/x]: Replace `allowed_to?` calls in models

klaustopher:


lib/open_project/access_control.rb
```diff
@@ -70,7 +70,7 @@ def permission(action)
 
           permissions.detect { |p| p.controller_actions.include?(permission_path) }
         else
-          permissions.detect { |p| p.name == action }
+          permissions.detect { |p| p.name == action&.to_sym }
```
dombesz: This change is a bit unexpected, is this tested anywhere?

dombesz (approved):
I left one question regarding the access_control.rb, othwerise LGTM!👍

lib/open_project/access_control.rb
klaustopher: I'll put it in the other place where this is needed, and not in the general class 👍

PR #14065, opened 2023-10-31
Permission Rework [17/x]: Replace `allowed_to?` calls in several straight forward remaining places

klaustopher:


app/models/users/permission_checks.rb
```diff
@@ -95,6 +95,26 @@ def all_permissions_for(context)
       .uniq
   end
 
+  # Helper method to be used in places where we just throw anything into the permission check and don't know what
+  # context it should be checked on. Things like menu item checks, controller action checks, generic services, etc.
+  def allowed_based_on_permission_context?(permission, project: nil, entity: nil) # rubocop:disable Metrics/PerceivedComplexity
```
klaustopher: This method is not used in this PR but will be used to fix the few remaining instances

app/models/work_package.rb
```diff
@@ -138,7 +138,7 @@ class WorkPackage < ApplicationRecord
          :relatable,
          :directly_related
 
-  acts_as_watchable
+  acts_as_watchable(permission: :view_work_packages)
```
klaustopher, line 141: We have classes that are derived from the `WorkPackage` class like `Task` or `Impediment`. When working with those classes, the permission check used to auto-generate the permission name that was checked. This generated `view_tasks` and `view_impediments`, which don't exist. Specifying the permission prevents the auto generation and always uses the `view_work_packages` permission correctly.

config/initializers/permissions.rb
```diff
@@ -171,6 +171,24 @@
...
+                     {},
+                     permissible_on: :global,
+                     require: :loggedin,
+                     grant_to_admin: true
+
+      map.permission :edit_grids,
+                     {},
+                     permissible_on: :global,
+                     require: :loggedin,
+                     grant_to_admin: true
```
klaustopher: Previously we had a system in place, that not-existing permissions wer always marked as granted to admins. And this way the permission was silently granted to each admin:

https://github.com/opf/openproject/blob/fc6f39ff3d50b2a287730021781f11f5e37ffff9/lib/open_project/access_control.rb#L176-L180

As we are explicitly checking that all permissions exist now, this lead to errors. So, here we have added all permissions that were previously granted to admins behind the scenes.

lib/open_project/access_control.rb
```diff
@@ -176,6 +176,7 @@ def contract_actions_map
       def grant_to_admin?(permission_name)
         # Parts of the application currently rely on granting not defined permissions,
         # e.g. :edit_attribute_help_texts to administrators.
+        # TODO: When we have added all permissions, we should remove this check.
```
klaustopher, line 179: The method is also no longer used in the `UserPermissibleService`. When we turn off the `UserALlowedService`, we can get rid of this as well

config/initializers/permissions.rb
ulferts: The permissions :view_grids and :edit_grids should, based on my digging so take it with a grain of salt, never be queried for in production, because the subclasses of `Grid` overwrite the required permissions:
* [`MyPage`](https://github.com/opf/openproject/blob/dev/modules/my_page/app/models/grids/my_page.rb#L38-L42)
* [`Dashboard`](https://github.com/opf/openproject/blob/dev/modules/dashboards/app/models/grids/dashboard.rb#L33-L35)
* [`Overview`](https://github.com/opf/openproject/blob/dev/modules/overviews/app/models/grids/overview.rb#L33-L35) 

Ideally, the superclass would complain if another subclass is added without specifying the permissions but for those that exists it is covered, so I don't think that we need to define those permissions. But I guess that you didn't simply add this by chance. When did you run into a `UnknownPermissionError` for that permission?

ulferts: For the `edit_attribute_help_text` I can reproduce the problem and understand where it comes from. However I don't think that adding the permission is the right solution. It creates new problems by adding the permission to the list of global permissions:

image

But even with the permission checkbox checked, a non admin user cannot make use of the permission. This will in my opinion only cause confusion. 

I see four options:

1.  Extend the functionality so that non admins can actually manipulate `AttributeHelpText` models. They would need to have access to the controller and the contracts would have to be adjusted as well.
2. Introduce a specific option to `acts_as_journalized only_admin: true` so that only admins are allowed to manipulate thus avoiding the generic permission check.
3. Allow the `acts_as_journalized` options to take a proc that is evaluated rather then the permission check.
4. Misuse an existing global permission e.g. `:manage_user`. 

The last one is dirty without question. It has the potential of causing problems later on and is just surprising to do. The first one is the option that will yield the most benefits in the long run but might require some work. Option 2 and 3 are rather similar. Given the `viewable_by_all_users` option introduced apparently only for the `AttributeHelpText` functionality, if I had to choose between the two, I'd go for option 3 so the option can be removed again. Option 2 looks like it will complicate the code. 

My recommendation is to check if option 1 really boils down to only altering the permission checks in the controller and contract. If it does, I'd go with it. If not, I'd go for option 3.

modules/grids/app/models/grids/grid.rb
```diff
@@ -45,6 +45,10 @@ def to_s
       name.presence || self.class.to_s.demodulize
     end
 
-    acts_as_attachable allow_uncontainered: false
+    acts_as_attachable allow_uncontainered: false,
+                       view_permission: :view_grids,
+                       add_on_new_permission: :edit_grids,
+                       add_on_persisted_permission: :edit_grids,
+                       delete_permission: :edit_grids
```
ulferts: See above as to why I don't think this change to be necessary.

ulferts (changes requested):
Most of what is in the PR is straight forward and can be merged from my perspective. But the juicy part of it, adding the permissions, is something where we need a different solution IMO. Please see above for more details. Sorry, that you were the one to run into the problem.

config/initializers/permissions.rb
klaustopher: Looking into option 1.

klaustopher: > Ideally, the superclass would complain if another subclass is added without specifying the permissions but for those that exists it is covered, so I don't think that we need to define those permissions. But I guess that you didn't simply add this by chance. When did you run into a `UnknownPermissionError` for that permission?

When removing the permissions in the grid.rb model, the following tests fail:

```
rspec './modules/grids/spec/models/grids/grid_spec.rb[1:1:2:3:1]' # Grids::Grid behaves like grid attributes behaves like acts_as_attachable included #attachments_visible allows access to a logged user when viewable_by_all_users is set
rspec './modules/grids/spec/models/grids/grid_spec.rb[1:1:2:1:4:1]' # Grids::Grid behaves like grid attributes behaves like acts_as_attachable included #validations on attachments_claimed for attachments that are attached, created by the current_user, added to attachments_claimed and the user having the permission is invalid
rspec './modules/grids/spec/models/grids/grid_spec.rb[1:1:2:1:3:1]' # Grids::Grid behaves like grid attributes behaves like acts_as_attachable included #validations on attachments_claimed for attachments that are unattached, created by the current_user, added to attachments_claimed and the user not having the permission is invalid
rspec './modules/grids/spec/models/grids/grid_spec.rb[1:1:2:1:2:1]' # Grids::Grid behaves like grid attributes behaves like acts_as_attachable included #validations on attachments_claimed for attachments that are unattached, created by the current_user, added to attachments_claimed and the user having the permission is valid
rspec './modules/grids/spec/models/grids/grid_spec.rb[1:1:2:1:5:1]' # Grids::Grid behaves like grid attributes behaves like acts_as_attachable included #validations on attachments_claimed for attachments that are unattached, not created by the current_user, added to attachments_claimed and the user having the permission is invalid
rspec './modules/grids/spec/models/grids/grid_spec.rb[1:1:2:2:1]' # Grids::Grid behaves like grid attributes behaves like acts_as_attachable included #attachments_claimed updates all attachments to be linked to the model before saving
```

klaustopher:
All requests are worked in:

- Added `skip_permission_checks` option to `acts_as_attachable` to circumvent permission checks when we don't need them
- Removed unneeded permissions for grids
- The `edit_attribute_help_texts` permission now allows users with that global role to edit help texts
- Simplified the check for the admin menu item in the admin menu

ulferts (approved):
Thanks for doing the extra work. Nice to see that it wasn't that complicated to turn the newly added permission into being useful.

PR #14074, opened 2023-11-01
allow writing with correct suppl. group

machisuji:


docker/prod/Dockerfile
```diff
@@ -187,10 +192,6 @@ RUN cp Gemfile.lock.bak Gemfile.lock && rm Gemfile.lock.bak
 
 RUN ./docker/prod/setup/postinstall.sh
 
-# We need this so puma is allowed to create the tmp/pids folder and files
```
machisuji, line 190: This was actually meant to happen for the slim, not the AIO container. The latter doesn't need it. So I simply moved this to the slim container build.

oliverguenther (approved):

PR #14092, opened 2023-11-06
Permission Rework [18/18]: Use new permission check methods in ApplicationController and action authorization

klaustopher:
Final step 🎉

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -293,14 +293,14 @@ class ApplicationController < ActionController::Base
   # Authorize the user for the requested controller action.
   # To be used in before_action hooks
   def authorize(ctrl = params[:controller], action = params[:action])
-    OpenProject::Deprecation.deprecate_method(ApplicationController, :authorize)
+    # OpenProject::Deprecation.deprecate_method(ApplicationController, :authorize)
     do_authorize({ controller: ctrl, action: }, global: false)
   end
 
   # Authorize the user for the requested controller action outside a project
   # To be used in before_action hooks
   def authorize_global
-    OpenProject::Deprecation.deprecate_method(ApplicationController, :authorize_global)
+    # OpenProject::Deprecation.deprecate_method(ApplicationController, :authorize_global)
 
     action = { controller: params[:controller], action: params[:action] }
     do_authorize(action, global: true)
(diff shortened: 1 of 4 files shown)


app/controllers/application_controller.rb
```diff
@@ -249,9 +311,21 @@ def authorize_global
...
+    is_authorized = begin
+      if global
+        User.current.allowed_based_on_permission_context?(action)
+      else
+        User.current.allowed_based_on_permission_context?(action, project: @project || @projects,
+                                                                  entity: @work_package || @work_packages)
+      end
+    rescue Authorization::UnknownPermissionError
+      false
+    end
```
klaustopher, line 328: This would be the quick & dirty version before we switch everything over to the the proper `before_action`. I think this might be a quick switch to get to testing quickly and then go through the controllers one by one to switch to the newly added methods above

app/services/authorization/user_permissible_service.rb
```diff
@@ -48,15 +49,17 @@ def allowed_in_entity?(permission, entities_to_check, entity_class)
...
+      return false unless authorizable_user?
       return true if admin_and_all_granted_to_admin?(perms)
 
       # entity_class.allowed_to will also check whether the user has the permission via a membership in the project.
+      # TODO: ^--- thats' actually not the case weirdly...
       allowed_scope = entity_class.allowed_to(user, perms)
 
       if in_project
-        allowed_scope.exists?(project: in_project)
+        allowed_in_single_project?(perms, in_project) || allowed_scope.exists?(project: in_project)
```
klaustopher, line 60: I have to figure out why, but the `WorkPackage.allowed_to` scope does not properly check project permissions. This is a band-aid fix to get the method working and I will fix the scope afterwards and remove this

config/initializers/permissions.rb
```diff
@@ -63,7 +63,7 @@
 
       map.permission :manage_user,
                      {
-                       users: %i[index show edit update],
+                       users: %i[index show edit update change_status change_status_info],
```
klaustopher, line 66: Another instance where this controller action was not covered before and was secretly granted to all admins.

ulferts (approved):
👍

lindenthal:
Very cool @klaustopher 👏👍

PR #14097, opened 2023-11-06
[#50119] Deal with cancelling and data loss

dombesz:
https://community.openproject.org/work_packages/50119

- [X] Render single Agenda entries after the crud actions
- [X] Update the dropdown menus of each item after crud and reordering actions.
- [x] Change the views in a way to re-render the show agenda items when a crud or move action happens


oliverguenther (approved):
Works very nicely, great stuff @dombesz 🙌

PR #14102, opened 2023-11-06
Bug/50677 users that are shared with a wp cannot be added as project members any more

ulferts:
Removes `Member`s on entities like WorkPackages from two scopes and by that:
* avoids them showing up in the administration page of a user where the current memberships in projects are listed:
 
image

* allows to add users as members with whom work packages inside the project have been shared before
* fixes showing the work package members in the members list: https://community.openproject.org/wp/50916

https://community.openproject.org/wp/50677


klaustopher (approved):

klaustopher:
Would it maybe make sense that we add scopes on memberships?

```ruby
scope :on_project, -> { where(entity: nil).where.not(project_id: nil) }
scope :global, -> { where(entity: nil).where(project_id: nil) }
scope :on_entity, -> { where.not(entity: nil, project_id: nil) }
```

Then we could use those instead of spreading `where(entity: nil)` in places outside of the model itself

ulferts:
@klaustopher all the scopes should now be in as discussed.

klaustopher (approved):
Awesome!

PR #14108, opened 2023-11-07
add work package editor role to workflow administration

ulferts:
Adds the 'Work Package Editor' role to the roles selectable when administrating the workflows:

image

Doing so, enables users having the role when having a work package shared with them to alter the status according to the workflows.

Administrators are notified of the need to configure a workflow via a warning:

image

### TODOs

* [x] Add editor role to workflow administration page
  * [x] limit this to only happen if an enterprise token grants the ability to have shared work packages
* [x] Display a warning to administrators in case no workflow is configured for that role  
  * [x] limit this to only happen if an enterprise token grants the ability to have shared work packages
  * [x] Cache the state of this to avoid querying the database on every request
* [x] introduce `work_package_sharing` EE action. Currently this is only limited to the workflow administration but will have to be extended to at least the share modal. 
* [x] Seed of a default workflow for new instances (same as the Project admin/member)

### Work package

https://community.openproject.org/wp/49482

### DIFF
--- a/app/helpers/warning_bar_helper.rb
+++ b/app/helpers/warning_bar_helper.rb
@@ -39,12 +39,28 @@ module WarningBarHelper
       (setting_protocol_mismatched? || setting_hostname_mismatched?)
   end
 
+  def render_workflow_missing_warning?
+    current_user.admin? &&
+      no_workflow_for_wp_edit_role?
+  end
+
   def setting_protocol_mismatched?
     request.ssl? != OpenProject::Configuration.https?
   end
 
   def setting_hostname_mismatched?
     Setting.host_name.gsub(/:\d+$/, '') != request.host
+  end
+
+  def no_workflow_for_wp_edit_role?
+    workflow_exists = OpenProject::Cache.read('no_wp_share_editor_workflow')
+
+    if workflow_exists.nil?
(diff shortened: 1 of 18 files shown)


app/helpers/warning_bar_helper.rb
```diff
@@ -47,6 +53,17 @@ def setting_hostname_mismatched?
...
+  def no_workflow_for_wp_edit_role?
+    workflow_exists = OpenProject::Cache.read('no_wp_share_editor_workflow')
+
+    if workflow_exists.nil?
+      workflow_exists = Workflow.exists?(role_id: Role.where(builtin: Role::BUILTIN_WORK_PACKAGE_EDITOR).select(:id))
+      OpenProject::Cache.write('no_wp_share_editor_workflow', workflow_exists) if workflow_exists
+    end
+
+    !workflow_exists
+  end
```
klaustopher, line 65: I don't see that the cache is ever cleaned? So wouldn't the warning always persist even after the workflow has been created?

ulferts: If I wrote it correctly, and there might be an easier way of doing this, the cache is only written if workflows exist:

`OpenProject::Cache.write(...) if workflow_exists`

That is why I did not use the `Cache.fetch` method which would be way simpler to use.

klaustopher: 👍

klaustopher (approved):

PR #14117, opened 2023-11-08
[#50896] Disabled custom field (on a project level) is still appearing in PDF file

as-op:
https://community.openproject.org/work_packages/50896


klaustopher (approved):

as-op:
Wow, you are fast @klaustopher :D Thanks!

PR #14120, opened 2023-11-08
Feature/50849 change work packages status

virgo47:
Implementation of feature [WP 50849](https://community.openproject.org/projects/openproject/work_packages/50849/activity) (Allow status change without Edit WP rights). The new permission is called `change_work_package_status`.

The permission is analogous to an existing `assign_versions` and just like it, the new permission actually takes away the capability to change the WP status from `edit_work_packages` alone (migration adding the new permission to the roles with `edit_work_packages` is provided).

The status change is a bit special and this is still WIP. Currently, most tests are fixed, but these used `edit_work_packages` and new permission was just added (as it would be with the migration). Some new tests will be added (for the new permission on its own).

But then there are tests using `add_work_packages` and just adding the new permission is the wrong fix. This is related to a new WP creation, e.g. in backlogs adding new WP complains that status can't be changed. This will be fixed in the code by further commits.


virgo47:
I've got most of the work done, I believe, but there are some unresolved questions:

1. Both `change_work_package_status` and previously existing `assign_versions` take away the editability from the general `edit_work_packages` permission. These permissions are for EDIT - and the typical scenario is to add them either to `edit_work_packages` (to preserve original full editability) or add them selectively to allow editing these fields. BUT this also affects the creation form where the Version field is completely omitted if `assign_versions` is missing. While for version it is not a problem, it is for the status. I'd argue that both fields should be editable for new WP unless we have more fine-grained ways how to specify whether the permission is for edit or for add. This affects some implementation in the `BaseContract#reduce_by_writable_permissions` where I added a "shortcut" for new models. This may be harsh and I'd welcome input on this one.
2. During the implementation I discovered a non-ideal behavior of the edit WP form. I reported this as [WP 50920](https://community.openproject.org/projects/openproject/work_packages/50920/activity). The form asks for the schema for the project and WP type - but the schema is computed with default permissions edit AND add WP, so if user is allowed to add new WP, all the fields are initially rendered as editable. Only after the click the form API delivers schema which is consulted against `edit_work_packages` permission only (which is right).

Ad 1), when adding new WP without the "shortcut" (that is with original behaviour when per-field permissions are consulted), when user has only the `change_work_package_status` permission the Status field is indicated as writable=false, but funny enough it actually shows dropdown for the status. If the status is indicated as writable, it shows only the text you can click on - and only then shows the dropdown. This visual difference would affect some tests with `add_work_packages` and without change status permission. The default status is eventually filled in - so no problem here. But this also affects some tests for Boards and Backlogs, it is not even possible to add a WP to the backlog without the new `change_work_package_status`. This is the reason why I believe that all fields should be editable when a new WP is created. That's why I added that "shortcut" - but the current implementation may be clumsy. However, it fixed most tests.

Currently failing tests - one would benefit from the proper writable indication in the schema call for edit WP page (WP 50920), the other I have to check yet.

Please, let me know how you see the whole situation.

virgo47:
@cbliard In a private communication you suggested a few tricks how to make the rights work differently for edit and new scenario - the most promising of these was to utilize subclasses of `WorkPackages::BaseContract` - that is `WorkPackages::CreateContract` and `UpdateContract`.

At first, I thought I can just move the `permission` specification for `status_id` to `UpdateContract` only. After all, that is where I needed to restrict it - so for create it would still be available as before (in contrast to how `version_id` behaves). However, this didn't play well with current UI, probably because of the [WP 50920](https://community.openproject.org/projects/openproject/work_packages/50920/activity) bug I mentioned - because in extreme cases there is no field indicated as writable before the `form` call for the WP is made - and that one is never made, because you have no writable field to click on.

So I decided to go the other way around - to leave the `status_id` permission specification in WP BaseContract *and* override it in `CreateContract` so that it has no permission. This may look like this:

````
    # Code in WP/BaseContract:
    attribute :status_id,
              permission: :change_work_package_status, # added permission spec
              writable: ->(*) { # original code from before
                model.status_id_change || !closed_version_and_status?
              }

    # And code in WP/CreateContract:
    attribute :status_id,
              # Overriding permission from WP base contract to ignore change_work_package_status for creation,
              # because we don't require that permission for writable status during WP creation.
              # Note that nil would not override and [] would ignore the default permission, so we use the default here:
              permission: :add_work_packages,
              writable: ->(*) { # probably not necessary, should be "inherited" from BaseContract...
````

With this, the "shortcut" code added for new models [here](https://github.com/opf/openproject/pull/14120/commits/234c0a750dc66c1ddbe5c40027643d9737a4bec2#diff-67216a47ea195c0f9f692831495250f98141a7b245f8ad79e1014bc7409326cfR245) is unnecessary and many tests start to work fine.

Alas, many other tests started to fail for very strange reasons and it took me quite some time to find the cause. I believe the implementation of `collect_ancestor_attributes_by` in the common base contract is not right. [This line](https://github.com/opf/openproject/blob/dev/app/contracts/base_contract.rb#L207) takes the info for the attribute from the more general class (superclass) and overrides with it the potentially more specific info. I guess the problem hardly occurs anywhere - but when I specify something else in WP's `CreateContract` (subclass), I expect it to be used when the method is run on that contract. Instead, the info from WP's `BaseContract` is used.

I'd suggest something like this instead:
````
      super_attributes = klass.send(attribute_to_collect).dup
      # attributes.send(...) would override more specific attributes in subclasses with more generic ones
      attributes = super_attributes.send(combination_method, attributes)
````

Now the `dup` needs to be called on attributes for each class, unless we rework the whole method to first create a list of classes and then walk it backwards.

Please, @cbliard, let me know what you think about the method - is there any reason why it overrides more specific attribute info with less specific ones? If you agree with my assumption, I'd love to fix it in a separate minimal PR, to see the effects of it first without all my other modifications on this PR. After that, I'd merge that PR (when accepted) to this one and focus on the problems caused only by the "change status" functionality.

**Comment by cbliard** [comment]
@virgo47 sorry for coming back late about this.

> I'd suggest something like this instead:
> 
> ```
>       super_attributes = klass.send(attribute_to_collect).dup
>       # attributes.send(...) would override more specific attributes in subclasses with more generic ones
>       attributes = super_attributes.send(combination_method, attributes)
> ```
> 
> Now the `dup` needs to be called on attributes for each class, unless we rework the whole method to first create a list of classes and then walk it backwards.
> 
> Please, @cbliard, let me know what you think about the method - is there any reason why it overrides more specific attribute info with less specific ones? If you agree with my assumption, I'd love to fix it in a separate minimal PR, to see the effects of it first without all my other modifications on this PR. After that, I'd merge that PR (when accepted) to this one and focus on the problems caused only by the "change status" functionality.

Your analysis is good, and I think your fix is working as expected, but I would not fix it this way. I would rather change it in `#collect_ancestor_attributes`:
```ruby
    combination_method, cleanup_method = if self.class.send(attribute_to_collect).is_a?(Hash)
                                           %i[merge with_indifferent_access]
                                         else
                                           %i[concat uniq]
                                         end
```
and change `%i[merge with_indifferent_access]` to `%i[reverse_merge with_indifferent_access]`. I think `#reverse_merge` makes it more obvious that the ancestors keys are merged only if they do not exist in the collected attributes yet.

By the way, as the hash is `dup`ed, it could even be `reverse_merge!` to make the merge in place.

Good idea to make a separate PR for that to evaluate the side effects.

db/migrate/20231106083056_set_change_work_package_status_permission.rb
```diff
@@ -0,0 +1,41 @@
...
+
+class SetChangeWorkPackageStatusPermission < ActiveRecord::Migration[7.0]
+  def up
+    ::Migration::MigrationUtils::PermissionAdder
+      .add(:edit_work_packages,
+           :change_work_package_status)
+  end
+
+  def down
+    # nothing to do
```
cbliard: The `#down` part of the migration could be done with something similar to `db/migrate/20230912185647_add_view_shared_work_packages_permission.rb`: 

```ruby
class SetChangeWorkPackageStatusPermission < ActiveRecord::Migration[7.0]
  # Decouple ActiveRecord model from Migration
  class MigrationRolePermission < ApplicationRecord
    self.table_name = 'role_permissions'
  end

  def up
    ::Migration::MigrationUtils::PermissionAdder
      .add(:edit_work_packages,
           :change_work_package_status)
  end

  def down
    MigrationRolePermission.where(permission: 'change_work_package_status').delete_all
  end
end
```

virgo47: Done as suggested, thanks.

virgo47:
Finally, with the CreateContract + collect_ancestor_attributes fixes (thank you, @cbliard, for all the suggestions) and the tests running, I consider this done. Of course, I'm open to further review suggestions.

app/contracts/work_packages/create_contract.rb
```diff
@@ -34,6 +34,16 @@ class CreateContract < BaseContract
...
+    attribute :status_id,
+              # Overriding permission from WP base contract to ignore change_work_package_status for creation,
+              # because we don't require that permission for writable status during WP creation.
+              # Note that nil would not override and [] would ignore the default permission, so we use the default here:
+              permission: :add_work_packages,
+              writable: ->(*) {
+                # If we did not change into the status,
+                # mark unwritable if status and version is closed
+                model.status_id_change || !closed_version_and_status?
+              }
```
cbliard: Is it necessary to override the `writable` property here? It's the same as in `BaseContract`.

app/policies/work_package_policy.rb
```diff
@@ -94,4 +95,12 @@ def comment_allowed?(work_package)
...
     user.allowed_in_project?(:assign_versions, work_package.project)
   end
+
+  def change_status_allowed?(work_package)
+    @change_status_cache ||= Hash.new do |hash, project|
+      hash[project] = user.allowed_in_project?(:change_work_package_status, work_package.project)
+    end
+
+    @change_status_cache[work_package.project]
+  end
```
cbliard: As user Alice, I tried sharing a work package with edit rights to user Bob having no access to a project.

As expected Bob now have limted access to the project, and can only see and edit the work package that Alice shared. Bob can edit everything but the status. Yet, the `WorkPackageRole` with name `Work Package Editor` has the permission `:change_work_package_status` being listed. (You can run `WorkPackageRole.find_by(builtin: Role::BUILTIN_WORK_PACKAGE_EDITOR).permissions` in a console to check).

I think it's lacking a work package level permission check, checking that either `user.allowed_in_project?(:change_work_package_status, work_package.project)` or `user.allowed_in_work_package?(:change_work_package_status, work_package)` returns true.

I tried changing the implementation to a naive `user.allowed_in_project?(:change_work_package_status, work_package.project) || user.allowed_in_work_package?(:change_work_package_status, work_package)` but it did not work. I suppose the schema is also interfering as it does its checks at the project level.

I am not comfortable yet with the new work package permissions system. I'll ask for some help here.

cbliard (changes requested):
Thanks a lot for this! It looks solid.

I found a case where it does not work as expected: when sharing a work package with edit permissions to someone else, this person can edit everything but the status. I'm not comfortable with `WorkPackageRole`s, so I'll ask for some help about it.

Also, I'm still not sure about the split of `edit_work_package` and `change_work_package_status`. As you already stated there are two possibilities (please correct me if I'm wrong): 
- It can behave like `add_work_package_attachments` so that `edit_work_package` allows to edit everything including the status while `change_work_package_status` allows to change only the status. This could be useful to allow a group of people to change work package status while forbidding them to modify other properties.
- Or it can behave like `assign_versions` so that `edit_work_package` is not enough to change the status and `change_work_package_status` is needed as well. This is the current implementation. This could be useful to forbid a group of people to change work package status while allowing them to modify all other properties.

I'll ask this tomorrow to our next Product meeting.

app/contracts/work_packages/create_contract.rb
virgo47: My bad, left there from when I wasn't sure about the granularity of the attribute's properties. I'll fix that.

app/policies/work_package_policy.rb
virgo47: I tried my best, but even while implementing this I was still learning more and more about OP's secrets, and I'm not aware of all the features obviously. So I'll welcome some more direct help with this, if possible.

virgo47:
Thanks for the review.

> Also, I'm still not sure about the split of `edit_work_package` and `change_work_package_status`...

Originally, I didn't know that `:assign_versions` is disjunct with WP edit, I'd not expect it - but I understand how that can be useful. That's why I wanted to model it this way at first. But I got some doubts lately, to be honest, especially because `status` is more essential to WP than version.

But instead of veering in the middle of the struggles with some tests still broken I rather finished it as originally proposed.

That said, I'd gladly do it as an additional permission only. I feel that cutting from the Edit WP capabilites and putting them to a separate permission is not very intuitive. It may make sense for `assign_versions` - as it also has a different wording. It's more about the relation between the version and the WP than about the `version` field alone, if I get it right. But for status, I'd gladly do it as an additional permission only, leaving Edit WP with that capability as well. That would also make the DB migration useless, but that's the least of my concerns. ;-)

I'll wait for the results of your discussions. However, without overly pushing you, I hope we can make it right and get it to the upstream during January, so we can use it from your authoritative source without the need to patch our installation. Of course, to make it "right" is important for me as well. I'll let you tell me what "right" is.

Thanks

app/policies/work_package_policy.rb
cbliard: I can relate to that, and the work package permissions is very new. No problem, we'll see how it can be fixed.
And if `:change_work_package_status` is no more distinct from `:edit_work_package`, that would solve the issue as well.

cbliard:
> > Also, I'm still not sure about the split of `edit_work_package` and `change_work_package_status`...
> 
> Originally, I didn't know that `:assign_versions` is disjunct with WP edit, I'd not expect it - but I understand how that can be useful. That's why I wanted to model it this way at first. But I got some doubts lately, to be honest, especially because `status` is more essential to WP than version.
> 
> But instead of veering in the middle of the struggles with some tests still broken I rather finished it as originally proposed.
> 
> That said, I'd gladly do it as an additional permission only. 

We discussed it, and we agreed on having `:edit_work_package` including `:change_work_package_status`. @psatyal will update the work package accordingly. 

In the "Role permissions" page, it will be reflected in the UI by automatically checking and disable the "change work package status" permission check box whenever the "edit work packages" permission check box is ticked. This is not done to include the permission (even if it feels correct) but rather to inform the user that the edit work packages permission includes the other one.

For consistency: the same behavior should be done for `:add_work_package_attachments`: when ticking "edit work packages", then "add work package attachments" gets disabled and ticked, and when unticking "edit work packages" then "add work package attachments" gets enabled and unticked.

On the backend side, the check is done by checking presence of either of `:edit_work_package` and `:change_work_package_status` permissions. It will also have the nice side-effect to fix the issue with sharing a work package for edition mentioned before.

Does it make sense?

app/policies/work_package_policy.rb
cbliard: It has been decided that `:edit_work_package` will encompass `:change_work_package_status`, so this problem will go away.

Just for information, if it was not, then something like `user.allowed_in_any_work_package?(:change_work_package_status)` would have been needed for the schema, because the schema and the writability of each field is scoped to the project. It would allow the user to click any status, and upon clicking the update_form endpoint would be queried, and that would return the effective permissions for the work package, forbidding or allowing the user to modify the field.

That's not really nice but that's how it should have been implemented.

psatyal:
Thanks @cbliard, I've updated the work package.

virgo47:
@cbliard I think I understand everything, thank you and the team for the clarification.

As for the backend, I'm not sure when I will start with the changes, in the worst case I may start on/after Jan 8, but I will definitely work on this.

The UI is by far the most problematic part for me... I'm not quite sure about visually checking something and then ignoring it on save to the model, not to mention you'll want some tests for it as well, I guess. For me, it is a lot of work (figuring out how to do it) with very little value. Wouldn't the permission explanation (rephrased, of course) be enough?

psatyal:
Hello @virgo47. Good point, automatic check/uncheck descoped for now. Thanks again for your contribution!

Cheers
Parimal

app/policies/work_package_policy.rb
virgo47: This should be resolved/not relevant with the change to the original `:edit_work_package` behaviour that allows status change without the new permission.

virgo47:
Hello @cbliard (and happy New Year!), I've made the changes, now `:edit_work_package` behaves as before and allows for status change. Strangely, one test failed, but it worked when I ran it on my station. Can it be a glitch?

Note that I needed to add support for the permission *symbols* array into `Authorization#permission_for`, it is in a separate commit above, you may want to review that - if it is the right place. Without it, the array of Edit WP + Change WP status permissions didn't work.

spec/lib/api/v3/work_packages/schema/work_package_schema_representer_spec.rb
```diff
@@ -864,7 +864,7 @@
     end
 
     describe 'status' do
-      context 'if having the assign_versions permission' do
+      context 'if having the change_work_package_status permission' do
```
cbliard, line 867: Nice catch

cbliard (approved):
Hello @virgo47 and happy new year to you too!
Thank you for keeping up with this :+1:

For the one failing test, yes some flickering tests occasionally slip into the code. A fix for this one [is being worked on](https://github.com/opf/openproject/pull/14498).

Good that you changed `Authorization#permission_for` to allow an array of symbols. I wanted to request changes about this commit to add a test, update the doc and fix the rubocop violations, and finally I found it simpler to push a commit than to write a review. I hope you don't mind.

It looks all good for me. I'll wait for the CI and then merge it once it's green, then there will be a last round of QA before we can consider it done.

virgo47:
@cbliard, I definitely don't mind, I actually thank you very much for the commit and also for the whole review process and merge. :-)

PR #14124, opened 2023-11-08
[#50910] enable 'EqualsAll' (&=) operator for the user_action filter

Kharonus:
[OP#50910](https://community.openproject.org/wp/50910)

- change user action filter type to `list_all`
- add operator `&=` to user action filter, which is true if ALL values are true.


wielinde:
@Kharonus The "all" operator is AFAIK an operator that does not need a value. It simply returns all entries for a requested resource. I guess that operator comes handy when the default query has a filter with a stricter operator, e.g. all "open" work packages.

So, I guess we need to have some other name such as `EqualsAll` with the operator `&=` (Here is that operator documented https://github.com/opf/openproject/blob/27bcaa9acfe4e10a3d193b6207d83dea9558e88b/docs/api/apiv3/tags/filters.yml#L89).

Kharonus:
You are right, what do you think to change the `list_all` filter type to actually contain `&=` instead of `*` operator?

Kharonus:
@wielinde I replaced the operator. What do you think about the current solution?

wielinde:
@Kharonus I am not an expert. I cannot see any usage of the `list_all` strategy anywhere. So, it might be safe to change it. However, I'd prefer to have @ulferts to have a look at this PR.

oliverguenther (approved):

PR #14130, opened 2023-11-09
Enable the Nextcloud Sync Service to mark a storage as unhealthy

mereghost:
## What?

The sync jobs run on the background, frequently failing without knowledge of the admins. This PR aims to lay the groundwork for the SyncService/CronJob pair to mark the storage as unhealthy.

## How?

I added a new `jsonb` field to the storage, containing the attributes that we would need to notify an admin that something is wrong with their integration.

The names right now are... dumb, I'm all ears for better suggestions.

Moreover I've made changes so that the entire SyncService works with service results, which will make it easier to return the necessary info.


modules/storages/app/common/storages/peripherals/storage_interaction/nextcloud/add_user_to_group_command.rb
```diff
@@ -53,33 +53,36 @@ def call(user:, group: @group)
           )
       )
 
+      error_data = Storages::StorageErrorData.new(source: self.class, payload: response)
```
Kharonus, line 56: 🟢 Nice, thanks!

modules/storages/app/models/storages/nextcloud_storage.rb
```diff
@@ -41,20 +41,7 @@ class NextcloudStorage < Storage
     store_attribute :provider_fields, :group, :string
     store_attribute :provider_fields, :group_folder, :string
 
-    def self.sync_all_group_folders
```
Kharonus, line 44: ❓  can you shortly elaborate, why this class method is no longer needed? I fail to see the connection to the scope of the PR, yet.

modules/storages/app/services/storages/group_folder_properties_sync_service.rb
```diff
@@ -46,40 +46,40 @@ class GroupFolderPropertiesSyncService
...
+    def self.call(storage)
+      new(storage).call
+    end
+
     def initialize(storage)
       @storage = storage
     end
 
     def call
-      prepare_remote_folders and apply_permissions_to_folders
```
Kharonus, line 54: 🟡 why can't we bind both calls together? so that a failure in `prepare_remote_folder` would already return failure service result? We did this with the `>>` operator, didn't we?

The new lines contain the same error handling twice.

modules/storages/app/services/storages/group_folder_properties_sync_service.rb
```diff
@@ -90,35 +90,27 @@ def apply_permissions_to_folders
...
     end
 
     def add_remove_users_to_group
       remote_users = remote_group_users.result_or do |error|
-        return OpenProject.logger.warn({ command: 'nextcloud.group_users',
-                                         group: @storage.group,
-                                         message: error.log_message,
-                                         data: { status: error.data.code, body: error.data.body } }.to_json)
+        return format_and_log_error(error, group: @storage.group)
+        # ServiceResult.failure(errors: error)
```
Kharonus: 🔴 commented line

modules/storages/app/models/storages/nextcloud_storage.rb
Kharonus: Ah, is it, that you basically moved the code into the cron job?

modules/storages/app/workers/storages/manage_nextcloud_integration_cron_job.rb
```diff
@@ -26,14 +28,25 @@
...
 
-  def perform
-    Storages::NextcloudStorage.sync_all_group_folders
+    def perform
+      OpenProject::Mutex.with_advisory_lock(NextcloudStorage, 'sync_all_group_folders', timeout_seconds: 0, transaction: false) do
+        NextcloudStorage.automatically_managed.includes(:oauth_client).find_each do |storage|
+          result = GroupFolderPropertiesSyncService.call(storage)
+          result.error_and do |error|
+            # Sample code
+            storage.mark_as_unhealthy(reason: error.code)
```
Kharonus: 🟡 a bit confused: how does a storage will get healthy again? I see no branch where it is marked as healthy.

Kharonus (changes requested):

modules/storages/app/models/storages/nextcloud_storage.rb
mereghost: Yep. This was moved to bring it closer to what triggers the behaviour.

modules/storages/app/services/storages/group_folder_properties_sync_service.rb
mereghost: `bind` is sure an option. I've changed it a bit to a syntax that may look clearer to the casual rubyist.

modules/storages/app/services/storages/group_folder_properties_sync_service.rb
```diff
@@ -46,40 +46,37 @@ class GroupFolderPropertiesSyncService
...
-        error_msg = { command: 'nextcloud.file_ids',
-                      folder: @storage.group_folder,
-                      message: error.log_message,
-                      data: { status: error.data.code, body: error.data.body } }.to_json
+        format_and_log_error(error, { folder: @storage.group_folder })
 
-        raise error_msg if error.code == :error
-
-        return OpenProject.logger.warn error_msg
+        # This hides the folder, but it will be in the logs
```
ba1ash: I cannot understand the comment here.
The return of failure directly will not hide the folder, right. I suggest  to be more explicit here:
1. clarify which folder: `the root group folder`
2. have a reference how returning a failure result leads to hiding.

mereghost: ~~The comment is wrong XD It should say, "don't hide the folder" 🤦🏼 ~~

Copied the comment when replacing lines 🤦🏼 ^2

modules/storages/app/models/storages/storage.rb
```diff
@@ -98,6 +88,24 @@ def self.shorten_provider_type(provider_type)
...
+    def healthy?
+      mark_as_healthy if health_status.nil?
+      health_status == 'healthy'
+    end
+
+    def unhealthy?
+      mark_as_healthy if health_status.nil?
+      health_status == 'unhealthy'
+    end
+
```
apfohl: I Suggest adding a default value to the health status in the database and simplify the logic here to side effect free methods.

apfohl: I suggest `unhealthy` as the default and setting it to `health` upon completion of the setup.

modules/storages/spec/models/storages/nextcloud_storage_spec.rb
```diff
@@ -42,6 +42,74 @@
...
+      expect(storage).to be_healthy
+    end
+
+    it 'does not mark as healthy if health_status has been defined earlier' do
+      storage.mark_as_healthy
+      expect do
+        storage.unhealthy?
+      end.not_to(change(storage, :health_status).from('healthy'))
+      expect(storage).not_to be_unhealthy
+    end
```
apfohl: With the changes above in the storage model these specs need to be reworked.

apfohl (changes requested):
Please create a default value for the health status as described in my comments.

Kharonus (approved):
I have nothing to add to this change anymore. So I agree for a merge. If we find anything not fitting, we can still adapt the code. from maintenance perspective code looks good! 👍🏽

PR #14149, opened 2023-11-14
[#50987] Fix storage connection error

dominic-braeunlein:
- [OP#50987](https://community.openproject.org/wp/50987)
- check for blank folder in empty_response to account for empty root folders

### DIFF
--- a/modules/storages/app/common/storages/peripherals/storage_interaction/nextcloud/files_query.rb
+++ b/modules/storages/app/common/storages/peripherals/storage_interaction/nextcloud/files_query.rb
@@ -72,9 +72,9 @@ module Storages::Peripherals::StorageInteraction::Nextcloud
     private
 
     def requested_folder(folder)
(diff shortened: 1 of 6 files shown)


Kharonus (changes requested):
🔴 I would not solve the problem by changing the interface of the `files_query`. In the manner that `folder` can no longer be `nil`.

The problem occurs due to the fact, that the responses for a storage differ heavily. Meaning, in nextcloud we always get the requested folder back, too. In onedrive, the response is empty.

So, this problem is a thing for the query implementation to solve. 

To be honest, I didn't knew, this bug ticket exist. So I actually issued a fix here: #14237

And I think, it is really as simple as that: tell the method in the one drive query, that builds "empty responses", that there is no parent, because we are root.

open for discussion. ;)

dominic-braeunlein:
Looks great!

PR #14161, opened 2023-11-16
Use capybara accessible selectors for share wp

cbliard:
Introduce [capybara_accessible_selectors](https://github.com/citizensadvice/capybara_accessible_selectors) for better accessibility.

- Use accessible selectors `within_modal` / `find_modal` in share_spec.rb and define the modal as an accessible one: role=dialog, `aria-modal=true, aria-labelledby=spotModalTitle
- Set `spotModalTitle` id on all modals.
- Introduce new custom accessible capybara selectors for `list` and `list_item` with custom description as well as rspec matchers (`have_list` / `have_list_item`).

### DIFF
--- a/spec/support/capybara/additional_accessible_selectors.rb
+++ b/spec/support/capybara/additional_accessible_selectors.rb
@@ -1,3 +1,5 @@
+# frozen_string_literal: true
+
 #-- copyright
 # OpenProject is an open source project management software.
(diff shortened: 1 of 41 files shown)


spec/support/capybara/additional_accessible_selectors.rb
aaron-contreras, line 1: 🎨🖌️

aaron-contreras (approved):
Awesome job!

PR #14165, opened 2023-11-16
Fixed storage oauth credential replace to show plaintext secret

apfohl:


ba1ash (approved):

akabiru (commented):
Nice one! Failing feature test at https://github.com/opf/openproject/blob/dev/modules/storages/spec/features/admin_storages_spec.rb#L156

PR #14171, opened 2023-11-17
[#50782] "Edit meeting dateils" tooltip visible when meeting details modal is canceled

dombesz:
https://community.openproject.org/work_packages/50782


HDinger (approved):
Looks kind of hacky but until Primer provides a fix for the original bug, it is defintely good enough. 👍

PR #14172, opened 2023-11-17
[51135] Fix selection of "group by" option in work package list view

cbliard:
https://community.openproject.org/wp/51135

Selecting "Group by" view option, selecting the first attribute from the drop down, and then clicking "Apply" has no effect.
It should apply the grouping.

This bug blocks me for renaming "Progress (%)" into "% Complete" because there is a feature test where the work packages are grouped by `done_ratio`, and after renaming it, "% Complete" is the first one in the list and the grouping is not applied.


cbliard:
Tests are passing, the error comes from eslint which I did not introduce and did not know how to fix.

HDinger (approved):

cbliard:
So just adding `void` is enough to get the eslint error fixed. Good to know.
Thanks for taking the time to fix it :pray:

PR #14185, opened 2023-11-20
[50749] Contrast issues

bsatarnejad:
https://community.openproject.org/projects/openproject/work_packages/50749/activity


frontend/src/app/shared/components/colors/colors.service.ts
```diff
@@ -18,4 +18,8 @@ export class ColorsService {
 
     return hash % 360;
   }
+
+  public colorMode():string {
+    return document.body.getAttribute('data-light-theme') ?? 'light';
```
HDinger: Currently you are checking (at multiple places) directly for the string Primer sets on the body. So whenever Primer changes that string, all of these checks will fail. Please make that a bit more future-proof and set `colorMode` to some kind of constant provided by us and then ckeck against that constant. Thus we will only have to adapt one place if Primer changes the string.

e.g.
```
   export const colorModes = {
     lightHighContrast: 'lightHighContrast
   }

   ...

    return document.body.getAttribute('data-light-theme') === 'light_high_contrast' ? colorModes.lightHighContrast : 'light';
    
    ... 

    if colorMode === colorModes.lightHighContrast ...
```

frontend/src/global_styles/content/_project_status.sass
```diff
@@ -1,11 +1,11 @@
...
-$project-status-deep-blue: #006E8F
-$project-status-light-blue: #84D6E5
-$project-status-deep-green: #007528
+$project-status-gray: var(--fgColor-subtle, var(--color-subtle-fg))
+$project-status-red: var(--fgColor-danger, var(--color-danger-fg))
+$project-status-orange: var(--fgColor-severe, var(--color-severe-fg))
+$project-status-green: var(--fgColor-success, var(--color-success-fg))
+$project-status-deep-blue: var(--fgColor-attention, var(--color-attention-fg))
+$project-status-light-blue: var(--fgColor-accent, var(--color-accent-fg))
+$project-status-deep-green: var(--fgColor-success, var(--color-success-fg))
```
HDinger: Please rename these variables. The variables don't necessarily translate to the colors in the names any more. Further, `status-green` and `status-deep-green` now have the same color. I know, that it was specified like this, so I asked the product team in the ticket to double check.

HDinger (changes requested):
Hi @bsatarnejad 

Looks good overall. I found some place that still don't have enough contrast:

- [x] In the files tab (`To add a link, see or upload files...`) the text is too bright. (contrast 2,8:1)
Bildschirmfoto 2023-11-24 um 09 26 48


- [ ] There are multiple places where we reached a ratio of 5:1 which is not sufficient for wcag AAA. I am not sure at the moment whether we aim for AA or AAA. Please double check that. For AAA we would need a ratio of 7:1. Examples are the sidebar headlines, locked users in the administration, filtered out parents in the WP table, the placeholder in date modal or greyed out options in the configuration dropdown of the WP table.
Bildschirmfoto 2023-11-24 um 09 25 46
Bildschirmfoto 2023-11-24 um 09 24 42
Bildschirmfoto 2023-11-24 um 09 35 54

bsatarnejad:
> Hi @bsatarnejad
> 
> Looks good overall. I found some place that still don't have enough contrast:
> 
>  
>- [ ] In the files tab (`To add a link, see or upload files...`) the text is too bright. (contrast 2,8:1)
> 
> Bildschirmfoto 2023-11-24 um 09 26 48
> [ ]  There are multiple places where we reached a ratio of 5:1 which is not sufficient for wcag AAA. I am not sure at the moment whether we aim for AA or AAA. Please double check that. For AAA we would need a ratio of 7:1. Examples are the sidebar headlines, locked users in the administration, filtered out parents in the WP table, the placeholder in date modal or greyed out options in the configuration dropdown of the WP table.
> 
> Bildschirmfoto 2023-11-24 um 09 25 46 Bildschirmfoto 2023-11-24 um 09 24 42 Bildschirmfoto 2023-11-24 um 09 35 54

As we discussed, We need to pass AA level, so for now a ratio of 5:1 is sufficient.

HDinger (approved):

PR #14199, opened 2023-11-21
Fix historic attributes

klaustopher:
The historic attributes query needs to modify the `WHERE` conditions as well now. Previously everything was fixed because we simply `JOIN`ed the permission related stuff. Now we have the `UNION` query that selects allowed_to for projects and work packages. So we need to build a new AST Walker, that also modifies all the other conditions where `work_packages` table is joined and move it to `journalized_work_packages`.


klaustopher:
Errors come from the underlying PR that is hiding those errors because the failing unit tests here prevent the feature tests. I'll merge this into the other branch to fix everything in there.

PR #14204, opened 2023-11-21
[#50908] Implement redirect to project folder with waiting under /projects/:project_id/project_storages/:id/open

ba1ash:
See feature for description: https://community.openproject.org/wp/50437
Link to the implementation chunk(no useful information there): https://community.openproject.org/work_packages/50908
Some technical implementation details worth mentioning:

1. A new flash message type has been introduced. A sample of usage:
   ```ruby
       flash[:modal] = {
         type: 'Storages::OpenProjectStorageModalComponent',
         parameters: {
           project_storage_open_url: request.path,
           redirect_url: storage_open_url,
           state: :waiting
           }
         } 
   ```
   `type` is a class name for desired modal view_component.
   `parameter` are attributes that are passed to the modal class constructor.
2. The waiting modal has a polling mechanism implemented. It is located in `storages/open-project-storage-modal.controller.ts`.
  In short: 
   1. It starts polling.
   2. Wait for response to arrive.
   3. Is it 200?
      1. Yes, then there is a turbo_stream in the body with updated modal content. Apply the turbo_stream. Stop polling.
      2. No, it returns to it starts from the beginning.
 3. The logic that responds to the polling is located in `storages/project_storages_controller.rb`. Check the test for it: `modules/storages/spec/requests/project_storages_open_spec.rb`. I believe I was quite descriptive there.
 4. Storage links in project menu has been updated to be `[ensure_connection_url]?destination_url=[project_storage_open_url]
 5. To see the waiting modal locally you need to set nextcloud integration up, enable a project storage to be automatically managed, run a background synchronization job once, then add a user to the project, click on the link in a project menu, then there should be a waiting modal, then fire Nextcloud sync job again, the modal should be updated. Or contact me and let's test it together :)


modules/storages/app/controllers/storages/project_storages_controller.rb
```diff
@@ -0,0 +1,104 @@
...
+    )
+    if @object.project_folder_automatic?
+      storage = @object.storage
+      # check if user "see" project_folder
+      result = ::Storages::Peripherals::Registry
+                 .resolve("queries.#{storage.short_provider_type}.file_info")
+                 .call(storage:,
+                       user: current_user,
+                       file_id: @object.project_folder_id)
+      result.match(
```
apfohl: 🟢 I would chain it and not have a `result` variable here.

modules/storages/app/controllers/storages/project_storages_controller.rb
```diff
@@ -0,0 +1,104 @@
...
+        on_success: ->(_) do
+          respond_to do |format|
+            format.turbo_stream do
+              render(
+                turbo_stream: OpTurbo::StreamComponent.new(
```
apfohl: 🟢 I'm thinking about having an option for our "base component" to pass to the ctor an option to set the TurboFrames name and it'll wrap the component you want to pass in here automagically into a frame. Just an idea. Not to be solved here.

modules/storages/app/controllers/storages/project_storages_controller.rb
```diff
@@ -0,0 +1,104 @@
...
+    if @object.project_folder_automatic?
+      storage = @object.storage
+      # check if user "see" project_folder
+      result = ::Storages::Peripherals::Registry
+                 .resolve("queries.#{storage.short_provider_type}.file_info")
+                 .call(storage:,
+                       user: current_user,
+                       file_id: @object.project_folder_id)
+      result.match(
+        on_success: ->(_) do
```
apfohl: 🟡 I would like to have the functions for `on_success` and `on_failure` extracted to untangle the spaghetti. :spaghetti:

modules/storages/lib/api/v3/project_storages/project_storage_open_api.rb
```diff
@@ -33,24 +33,7 @@ class API::V3::ProjectStorages::ProjectStorageOpenAPI < API::OpenProjectAPI
...
-                         .resolve("queries.#{@project_storage.storage.short_provider_type}.open_file_link")
-                         .call(
-                           storage: @project_storage.storage,
-                           user: current_user,
-                           file_id: @project_storage.project_folder_id
-                         )
-                     end
-
-      query_result.match(
+      @project_storage.open(current_user).match(
```
apfohl, line 36: 🟢 YEAH :metal:

modules/storages/lib/open_project/storages/engine.rb
```diff
@@ -159,12 +159,21 @@ def self.permissions
...
-            href = "/api/v3/project_storages/#{project_storage.id}/open"
-            icon = if storage.provider_type_nextcloud?
-                     'nextcloud-circle'
-                   else
-                     'hosting'
-                   end
+            next unless storage.configured?
+
+            url_helpers = Rails.application.routes.url_helpers
+            open_project_storage_url = url_helpers.open_project_storage_url(
```
apfohl, line 163: 🟡🟢 I don't get, why we have these URL data in some completely unrelated file then what they belong to. I know that this comes from :railway_track: but it would be better placed closes to the storage models.

apfohl (changes requested):
No major things to change. Mostly one 🟡 I would like you to change. The other 🟡 and 🟢 are just topics I would like to talk about with someone.

modules/storages/app/controllers/storages/project_storages_controller.rb
ba1ash: Sure, I'll extract it.

modules/storages/lib/open_project/storages/engine.rb
ba1ash: Yes, the piece of logic is kinda big here. We could think together how to improve it or leave it as it is for now.

apfohl (approved):

PR #14213, opened 2023-11-22
[OP#50985] use primer flash (banner) for file storage administration settings

akabiru:
#### https://community.openproject.org/wp/50985

To test, create a storage

- [x] Support dismiss action _(Switches from [`Flash`](https://primer.style/components/flash) to [`Banner`](https://primer.style/components/banner/rails/alpha) as flash does not yet have a working dismiss action)_

See: https://github.com/primer/view_components/issues/2394

Screenshot 2023-11-23 at 1 34 14 PM

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -115,19 +115,19 @@ module ApplicationHelper
     end
   end
 
-  def render_primer_flash_message?
-    flash[:primer_flash].present?
+  def render_primer_banner_message?
+    flash[:primer_banner].present?
   end
 
-  def render_primer_flash_message
-    return unless render_primer_flash_message?
+  def render_primer_banner_message
+    return unless render_primer_banner_message?
 
-    render(FlashMessageComponent.new(**flash[:primer_flash].to_hash))
+    render(BannerMessageComponent.new(**flash[:primer_banner].to_hash))
   end
 
   # Renders flash messages
-  def render_legacy_flash_messages
-    return if render_primer_flash_message?
+  def render_flash_messages
+    return if render_primer_banner_message?
 
     messages = flash
       .reject { |k, _| k.start_with? '_' }
@@ -151,24 +151,20 @@ module ApplicationHelper
     end
   end
 
-  def render_legacy_flash_message(type, message, html_options = {}) # rubocop:disable Metrics/AbcSize
+  def render_flash_message(type, message, html_options = {})
     if type.to_s == 'notice'
       type = 'success'
     end
-
     toast_css_classes = ["op-toast -#{type}", html_options.delete(:class)]
-
     # Add autohide class to notice flashes if configured
     if type.to_s == 'success' && User.current.pref.auto_hide_popups?
       toast_css_classes << 'autohide-toaster'
     end
-
     html_options = { class: toast_css_classes.join(' '), role: 'alert' }.merge(html_options)
     close_button = content_tag :a, '', class: 'op-toast--close icon-context icon-close',
                                        title: I18n.t('js.close_popup_title'),
                                        tabindex: '0'
     toast = content_tag(:div, join_flash_messages(message), class: 'op-toast--content')
-
     content_tag :div, '', class: 'op-toast--wrapper' do
       content_tag :div, '', class: 'op-toast--casing' do
         content_tag :div, html_options do
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -115,66 +115,6 @@ module ApplicationHelper
     end
   end
 
-  def render_primer_banner_message?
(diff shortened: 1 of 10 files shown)


HDinger (commented):
Hi @akabiru 

The banners look really nice 👍  I have some remarks:

- [x] When I delete a storage there is still the old success message shown
- [ ] When I edit a storage (e.g change the name), I don't see any banner. However when I click afterwards on any link in the sidebar, I see on the new page the old flash message "Successfully updated".
- [ ] On mobile, the banner looks off. I assume that this happens because the `content-wrapper` has additional padding on mobile and the banner is a child of that. So we might have to move the banner a bit up in the DOM

Bildschirmfoto 2023-11-24 um 08 43 31



**Some additional notes, that do not necessariliy require changes on this PR:**

- [ ] When I create a OneDrive storage, and press cancel in the "OAuth applications" section, the section simply closes and I see no success banner. (Same when canceling in the nextcloud setup). I don't know though if the banner here is the right behaviour, or whether it would be better to redirect to the index page and show the banner there. What do you think?
- [ ] I was wondering whether we want to show a banner in the error cases as well? However, I saw that there is nothing designed for that in Figma either, so I am fine with leaving it out for now.
- [ ] Just a note: It feels weird, that the dismiss icon is focused and the tooltip shown initially. But since this is probably coming from Primer (and is accessibility conform) I am fine with keeping it like this.
- [ ] Currently the tooltip of the closing icon "Dismiss" is not translated. I know that this is coming from Primer, but I would like to have a follow-up ticket to change that.

akabiru:
Thank you fro the review @HDinger , some responses
> 
>     * [x]  When I delete a storage there is still the old success message shown

Thanks, Done ✅  _initially the designs didn't cover this but it makes sense 👍🏾_

>     * [ ]  When I edit a storage (e.g change the name), I don't see any banner. However when I click afterwards on any link in the sidebar, I see on the new page the old flash message "Successfully updated".

With the unified form, this behaviour is expected. As soon as a storage is created- the label changes from "incomplete" to "completed"- this is the indicator that the **_first step_** is complete- note that the completion flash is only rendered after all steps in the form have been completed.

>     * [ ]  On mobile, the banner looks off. I assume that this happens because the `content-wrapper` has additional padding on mobile and the banner is a child of that. So we might have to move the banner a bit up in the DOM
> 
> 
> Bildschirmfoto 2023-11-24 um 08 43 31

Good catch, the `full_when_narrow: false` prop should cover this but it doesn't seem to work. I'm also not sure we can (should) move the flash any further as moving it out of the `content-wrapper` broke the page. Seems ok for now but I can track this in a WP?

> **Some additional notes, that do not necessariliy require changes on this PR:**
> 
>     * [ ]  When I create a OneDrive storage, and press cancel in the "OAuth applications" section, the section simply closes and I see no success banner. (Same when canceling in the nextcloud setup). I don't know though if the banner here is the right behaviour, or whether it would be better to redirect to the index page and show the banner there. What do you think?

Ditto: The success banner is only rendered after successful completion- note that it's not a "success created" message, rather a reminder to activate the storage in projects, which is not possible if the storage has not been fully configured.

>     * [ ]  I was wondering whether we want to show a banner in the error cases as well? However, I saw that there is nothing designed for that in Figma either, so I am fine with leaving it out for now.

Good eye, error cases are covered in form validation errors- but for cases like delete failure (edge-case), it's will render a danger banner.

>     * [ ]  Just a note: It feels weird, that the dismiss icon is focused and the tooltip shown initially. But since this is probably coming from Primer (and is accessibility conform) I am fine with keeping it like this.
>     * [ ]  Currently the tooltip of the closing icon "Dismiss" is not translated. I know that this is coming from Primer, but I would like to have a follow-up ticket to change that.

Good eye, this is [hardcoded atm](https://github.com/opf/primer_view_components/blob/main/app/components/primer/alpha/banner.html.erb#L25)- we would have to fix it. https://community.openproject.org/wp/51360

HDinger (approved):
Hi @akabiru 

> Good catch, the full_when_narrow: false prop should cover this but it doesn't seem to work. I'm also not sure we can (should) move the flash any further as moving it out of the content-wrapper broke the page. Seems ok for now but I can track this in a WP?

It is okay for me to extract the mobile issue into a separate ticket 👍  Good to merge then from my side 🥳

akabiru:
> Hi @akabiru
> 
> > Good catch, the full_when_narrow: false prop should cover this but it doesn't seem to work. I'm also not sure we can (should) move the flash any further as moving it out of the content-wrapper broke the page. Seems ok for now but I can track this in a WP?
> 
> It is okay for me to extract the mobile issue into a separate ticket 👍 Good to merge then from my side 🥳

Thank you! Tracked in https://community.openproject.org/wp/51370

PR #14232, opened 2023-11-23
[#51015] Mobile, the participant section should move to details section

dombesz:
https://community.openproject.org/work_packages/51015

<details><summary>Demo</summary>
<p>

https://github.com/opf/openproject/assets/83396/11720179-4a86-4601-85b2-9ee762cef9e4

</p>
</details>


oliverguenther (approved):
I've pushed a change to use the AsyncDialogComponent so we can lazy load the dialog content. Could you take another look?

PR #14238, opened 2023-11-24
Fix/make custom actions spec faster and more reliable

cbliard:
It may not fix the `Capybara::ElementNotFound: Unable to find visible css ".ng-option" with text "Progress (%)" within #<Capybara::Node::Element tag="div" path="/HTML/BODY[1]/NG-DROPDOWN-PANEL[1]/DIV[1]">` error yet, but it fixes other issues that occured on my machine preventing me from investigating the real issue.

### DIFF
--- a/spec/support/pages/admin/custom_actions/form.rb
+++ b/spec/support/pages/admin/custom_actions/form.rb
@@ -111,10 +111,10 @@ module Pages
 
           Array(value).each do |val|
             within field do
-              if has_selector?('.form--selected-value--container', wait: 1)
+              if has_selector?('.form--selected-value--container', wait: 0)
                 find('.form--selected-value--container').click
                 autocomplete = true
-              elsif has_selector?('.autocomplete-select-decoration--wrapper', wait: 1)
+              elsif has_selector?('.autocomplete-select-decoration--wrapper', wait: 0)
                 autocomplete = true
               end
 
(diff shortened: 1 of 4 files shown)


spec/support/pages/admin/custom_actions/form.rb
```diff
@@ -111,10 +113,10 @@ def set_action_value(name, value)
 
           Array(value).each do |val|
             within field do
-              if has_selector?('.form--selected-value--container', wait: 1)
+              if has_selector?('.form--selected-value--container', wait: 0)
```
aaron-contreras, line 116: 👏

spec/support/shared/ferrum_patches.rb
```diff
@@ -41,3 +41,24 @@ def ignore_ferrum_javascript_error
...
+    end
+
+    def pending_connections
+      main_frame_id = @traffic.first&.request&.frame_id
+      current_navigation = @traffic.reverse.find { |conn| conn.navigation_request?(main_frame_id) }
+      current_traffic = @traffic.filter { |exchange| exchange.request.loader_id == current_navigation.request.loader_id }
+      current_traffic.count(&:pending?)
+    end
+  end
+end
```
aaron-contreras, line 64: 🎨🖌️

aaron-contreras (approved):
Awesome! 🇫🇷

PR #14241, opened 2023-11-24
[#50265] Allow Shared with users to become Watchers

aaron-contreras:
## Notes
This PR has @klaustopher 's https://github.com/opf/openproject/pull/14118 as its base. It can either be merged into it or rebased on `dev` once the aforementioned PR is merged in.

See: https://community.openproject.org/work_packages/50265

### DIFF
--- a/app/models/users/permission_checks.rb
+++ b/app/models/users/permission_checks.rb
@@ -56,6 +56,15 @@ module Users::PermissionChecks
     def allowed_members(action, project)
       Authorization.users(action, project).where.not(members: { id: nil })
     end
+
+    def allowed_on_work_package(action, work_package)
+      project_members = allowed_members(action, work_package.project)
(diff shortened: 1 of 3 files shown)


oliverguenther (approved):

app/models/users/permission_checks.rb
```diff
@@ -56,6 +56,15 @@ def allowed(action, project)
     def allowed_members(action, project)
       Authorization.users(action, project).where.not(members: { id: nil })
     end
+
+    def allowed_on_work_package(action, work_package)
```
klaustopher: Can we rename this to `allowed_members_on_work_package`? It returns a list of members, and that should be representend in the name

aaron-contreras: I thought of it but couldn't decide on which of the two method names to pick. Thanks

PR #14242, opened 2023-11-24
[#51229] Remove feature flags for storage_primer_design and storage_one_drive_integration

ba1ash:
https://community.openproject.org/work_packages/51229


apfohl (approved):
Looks very clean now with all the components.

modules/storages/spec/features/admin_storages_spec.rb
```diff
@@ -537,7 +537,8 @@
     end
   end
 
-  it 'creates, edits and deletes storages', :webmock do
+  # skipped to be revised later. broken due to removal of storage_primer_design feature flag
+  xit 'creates, edits and deletes storages', :webmock do
```
akabiru, line 541: @ba1ash we can delete these specs, they're for the old flow

modules/storages/spec/features/admin_storages_spec.rb
```diff
@@ -753,7 +754,8 @@
     end
     let!(:unconfigured_storage) { create(:nextcloud_storage) }
 
-    it 'reports storages that are not configured correctly' do
+    # skipped to be revised later. broken due to removal of storage_primer_design feature flag
+    xit 'reports storages that are not configured correctly' do
```
akabiru, line 758: Ditto: would also need deleting the configuration checks component

modules/storages/spec/features/admin_storages_spec.rb
ba1ash: Sure, I will do it. I postponed it because I was not sure. Thanks for clarifying!

PR #14243, opened 2023-11-24
Added DangerZone for storage deletion in administration

apfohl:
https://community.openproject.org/projects/openproject/work_packages/50609/activity

This PR adds the **Danger Zone** to confirm a file storage deletion from the *Administration* of *File Storages*.

- [x] Triggered by clicking the `delete` button on the *File Storage Edit* page.
- [x] Upon entering the correct name of the storage that should be deleted, the "Delete" button gets active.
- [x] After successful deletion, the list of remaining storages is show.
- [x] Cancelling the **Danger Zone** leads back to the list of storages.

### DIFF
--- a/modules/storages/config/locales/en.yml
+++ b/modules/storages/config/locales/en.yml
@@ -95,6 +95,7 @@ en:
     page_titles:
       file_storages:
         subtitle: "Add an external file storage in order to upload, link and manage files in work packages."
+        delete: "Delete file storage"
       managed_project_folders:
         title: "Automatically managed project folders"
         subtitle_short: "Let OpenProject create folders per project automatically."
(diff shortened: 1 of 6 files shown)


modules/storages/config/locales/en.yml
```diff
@@ -166,11 +167,14 @@ en:
         one_drive: "Allow OpenProject to access Azure data using OAuth to connect OneDrive/Sharepoint."
     delete_warning:
       storage: >
-        Are you sure you want to delete this storage? This will also delete the storage from all projects where it is used.
-        Further, it will also delete all links from work packages to files that are stored in that storage.
+        Are you sure you want to delete %{file_storage}? To confirm this action please introduce the
+        storage name in the field below, this will:
+      storage_delete_result_1: "Remove all storage setups for all projects using this storage."
+      storage_delete_result_2: "Remove all links from work packages of all projects to files and folders of that storage."
+      storage_delete_result_3: "In case this storage has an automatically managed project folders, these and its files will be deleted forever."
```
ba1ash: Is this true?
I think that the files and folders will be hidden after the deletion.

wielinde: @ba1ash I checked the Delete service of storagas. It iterates over all ProjectStorages and calls its Delete service, which in turn deletes the project folder. So, I guess all automatically managed project folders get deleted. At least that is how I understand the code.

apfohl: @wielinde @ba1ash If I see this correctly in `modules/storages/app/services/storages/group_folder_properties_sync_service.rb:160`, we only remove permissions from the folders in Nextcloud. We do not remove them. So the text needs to be different. You agree?

ba1ash: @apfohl, I agree with text changes. We only hide folders(files) on NC side.
@wielinde yes we delete project_storages record, but my point was about `these and its files will be deleted forever.`

apfohl: I changed the text.

ba1ash (approved):
LGTM :+1:

PR #14247, opened 2023-11-27
Fixed meetings copy spec that used Time's next_week in a wrong way

apfohl:
`next_week` of `Time` was used in a wrong way, resulting in a spec failing on certain days of the week.


klaustopher (approved):
👍

PR #14252, opened 2023-11-27
Improve instructions for supporting self signed certificates

wielinde:


docs/system-admin-guide/integrations/nextcloud/README.md
```diff
@@ -303,7 +311,15 @@ On OpenProject inside the storage administration (*Administration → File stora
 
 ##### While using a self signed TLS/SSL certificate you receive "certificate verify failed"
 
-Some administrators setup Nextcloud using a self signed TLS/SSL certificate with their own CA (certificate authority). The CA needs to be known on the OpenProject server. On a Debian/Ubuntu based server, make sure you add the CA certificate for your Nextcloud certificate to `/usr/local/share/ca-certificates` and run `sudo update-ca-certificates` afterwards. Then OpenProject's Ruby code should be able to verify your Nextcloud TLS/SSL certificate when emitting HTTPS requests to your Nextcloud server.
+Some administrators setup Nextcloud using a self signed TLS/SSL certificate with their own CA (certificate authority). The CA needs to be known on the OpenProject server. On a Debian/Ubuntu based server, make sure you add the CA certificate for your Nextcloud certificate to `/usr/local/share/ca-certificates` and run `sudo update-ca-certificates` afterwards.
+
+Add the following line to the file `.env` in your application's home directory:
+
+```ENV
```
wielinde: @Kharonus

- I am not sure if `ENV` as a syntax actually exists.
- I am not sure if using an `.env` file is the right, the go-to approach to set ENV variables.

Kharonus: I'd try not to be too specific and instead say:

````markdown
Add the following environmental variable to the OpenProject server and restart it:

```
SSL_CERT_FILE=/etc/ssl/certs/ca-certificates.crt
```
````

point is, that we do not want to contain every possible way out there to handle env variables. `.env` in this case is a file only interpreted by the docker daemon when composing configurations.

Kharonus (changes requested):

Nudelsalad:
Wondering why this hasn't been merged yet, spent past two hours searching this information. Couldn't also find sth in the forum. 
Pls for gods sake, merge it :v:

wielinde:
@NobodysNightmare May I ask you to take over this PR? When using self signed certificates you need to tell both, PHP and Ruby where to find the CA. And @Kharonus is right, my solution is a bit too specific. Still, we should add this to our troubleshooting guides. Just a week ago I ran into the same problem myself again and could not find any clear help searching online. I needed the info here.

NobodysNightmare:
I rebased the PR onto `release/16.0` and also targeted the PR to that branch, so that the docs updates get visible with the upcoming release.

The proposal of @Kharonus was included as suggested, but keeping the specific instructions as an example.

@MayaBerd asking for your editorial review on this.

MayaBerd:
@NobodysNightmare looks good to me

MayaBerd (approved):

wielinde:
Thank you @NobodysNightmare!

PR #14255, opened 2023-11-28
Honor exisiting builtin roles when seeding

klaustopher:
#14254 introduced the WP roles beeing seeded via a migration. This causes the seeds to run after the migrations to fail. This PR sets the RoleSeeders up to look for existing `builtin` roles before seeding. This way the seeder works correctly even with the existing WP roles already beeing seeded by the migration

### DIFF
--- a/app/seeders/basic_data/base_role_seeder.rb
+++ b/app/seeders/basic_data/base_role_seeder.rb
@@ -28,6 +28,7 @@
 module BasicData
   class BaseRoleSeeder < ModelSeeder
     self.needs = []
+    self.attribute_names_for_lookups = %i[builtin name]
 
     def model_attributes(role_data)
       {
--- a/app/seeders/basic_data/base_role_seeder.rb
+++ b/app/seeders/basic_data/base_role_seeder.rb
@@ -28,7 +28,6 @@
 module BasicData
   class BaseRoleSeeder < ModelSeeder
     self.needs = []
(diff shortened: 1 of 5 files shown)


cbliard (changes requested):
It fails if run in another language than English, because the lookup acts on the name. 
Try this to reproduce:
```
OPENPROJECT_DEFAULT_LANGUAGE=de rails db:seed
```
I think it's fine to only lookup by attribute `[:builtin]` alone. This would prevent this issue.

And I think there is a more general issue that the work package role name is not translated anymore. We can merge this one once fixed, and then we'll see how to make the work package role translated again.

cbliard:
And I am surprised that the issue that you are fixing now has not been caught by unit tests, especially `spec/seeders/root_seeder_standard_edition_spec.rb`. I'll investigate that as well.

Edit: `db:test:purge` rake task is called so the roles added by the migration are obliviated. I added a unit test to manually run the migration before running the tests.

klaustopher:
> I think it's fine to only lookup by attribute [:builtin] alone. This would prevent this issue.

This will not work, as all non-builtin roles use builtin=0 and this way no other role would be seeded.

cbliard (approved):
Nice :+1: 
Thanks for detecting and fixing it!

klaustopher:
Thanks for adding the tests 👍 and pointing me to the edge cases like translations

PR #14256, opened 2023-11-28
[#51274] return correct parent id

Kharonus:
[OP#51274](https://community.openproject.org/wp/51274)

### WAT?

- one drive files query now returns real parent id when folder is empty, not the fake id like before

### But ... WHY?

- upload needs it for the prepare upload request, fake id led to error

### DIFF
--- a/modules/storages/app/common/storages/peripherals/storage_interaction/one_drive/files_query.rb
+++ b/modules/storages/app/common/storages/peripherals/storage_interaction/one_drive/files_query.rb
@@ -50,14 +50,14 @@ module Storages
             result = Util.using_user_token(@storage, user) do |token|
               # Make the Get Request to the necessary endpoints
               response = Net::HTTP.start(@uri.host, @uri.port, use_ssl: true) do |http|
-                http.get(uri_path_for(folder) + FIELDS, { 'Authorization' => "Bearer #{token.access_token}" })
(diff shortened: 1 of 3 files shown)


Kharonus:
I still need to update the VCR cassette.

modules/storages/app/common/storages/peripherals/storage_interaction/one_drive/files_query.rb
```diff
@@ -166,12 +174,18 @@ def root(id)
                             permissions: %i[readable writeable])
           end
 
-          def uri_path_for(folder)
+          def children_uri_path_for(folder)
             return "/v1.0/drives/#{@storage.drive_id}/root/children" if folder.root?
```
apfohl, line 178: 🟡 Can't you use `location_uri_path_for` here somehow? Looks a bit WET ;).

apfohl (changes requested):
LGTM

modules/storages/app/common/storages/peripherals/storage_interaction/one_drive/files_query.rb
Kharonus: I wouldn't do so, as it will still need an `if` (appending `:/children` or `/children`).

so I'd argue, that it is actually better readable having it WET.

apfohl: okay

apfohl (approved):

PR #14258, opened 2023-11-28
[50609] changed result text of storage deletion

apfohl:


modules/storages/config/locales/en.yml
```diff
@@ -171,7 +171,7 @@ en:
         storage name in the field below, this will:
       storage_delete_result_1: "Remove all storage setups for all projects using this storage."
       storage_delete_result_2: "Remove all links from work packages of all projects to files and folders of that storage."
-      storage_delete_result_3: "In case this storage has an automatically managed project folder, the permissions to access the folder and its files will be removed."
+      storage_delete_result_3: "In case this storage has an automatically managed project folders, these and its files will be deleted forever."
```
wielinde:
```suggestion
      storage_delete_result_3: "In case this storage has automatically managed project folders, those and and the contained files will be deleted forever."
```

modules/storages/config/locales/en.yml
```diff
@@ -171,7 +171,7 @@ en:
         storage name in the field below, this will:
       storage_delete_result_1: "Remove all storage setups for all projects using this storage."
       storage_delete_result_2: "Remove all links from work packages of all projects to files and folders of that storage."
-      storage_delete_result_3: "In case this storage has an automatically managed project folder, the permissions to access the folder and its files will be removed."
+      storage_delete_result_3: "In case this storage has automatically managed project folders, those and and the contained files will be deleted forever."
```
wielinde:
```suggestion
      storage_delete_result_3: "In case this storage has automatically managed project folders, those and their contained files will be deleted forever."
```

PR #14265, opened 2023-11-28
[chore] Refactor nextcloud files query spec to vcr

Kharonus:
### WAT?

- refactored files query spec for nextcloud query to use VCR cassettes
- some small adjustments of production code to satisfy common interface
  - root entity has now name `Root` in both queries, and no longer `/` in nextcloud
  - error data contains now source in payload for files query of nextcloud

### But ... WHY?

- because VCR is the much more awesome test setup


akabiru (commented):
Nice one! 🥇 Seems there're failing specs- good to merge from POV on green 👍🏾

Kharonus:
Yeah, specs were red, because I recorded the cassettes with my local nextcloud on `nextcloud-sub.local/html` which caused the cassette matching to fail.

I rerecorded them now with a `nextcloud.local` instance.

akabiru (approved):
:shipit:

PR #14215, opened 2023-11-22
#50921 Add storage health information in admin page

dominic-braeunlein:
##### Related PR: #14130 

This PR adds labels to unhealthy storages in the index page.
And adds additional health information on the edit page.

PR #14268, opened 2023-11-28
fix[#50921] Remove name sequence from storage child classes

dominic-braeunlein:
This removes previously added code in storage child factories which broke other tests relying on those factories.

The breaking code was merged with #14215.

### DIFF
--- a/modules/storages/spec/factories/storage_factory.rb
+++ b/modules/storages/spec/factories/storage_factory.rb
@@ -54,7 +54,6 @@ FactoryBot.define do
           class: '::Storages::NextcloudStorage' do
     provider_type { Storages::Storage::PROVIDER_TYPE_NEXTCLOUD }
     sequence(:host) { |n| "https://host#{n}.example.com" }
-    sequence(:name) { |n| "Storage #{n}" }
 
     trait :as_automatically_managed do
       automatically_managed { true }
@@ -135,7 +134,6 @@ FactoryBot.define do
           parent: :nextcloud_storage,
           traits: [:as_automatically_managed] do
     sequence(:host) { |n| "https://host-complete#{n}.example.com" }
-    sequence(:name) { |n| "Storage complete #{n}" }
 
     after(:create) do |storage|
       create(:oauth_client, integration: storage)


aaron-contreras:
Interesting. I'm curious: what kind of conflict did it cause by setting a sequence on the parent and nested factory, and what would the proper way of accomplishing this be to have a unique autogenerated attribute on each child factory while also having one on the parent? (or is this an incorrect usage of the sequence macro?)

dominic-braeunlein:
**Short answer**
The error is caused by incorrect usage. I added the same sequence to a child factory as it is on the parent.



**Long answer**

_Pre-info
Storages need to have a unique name and there are two different types of storages.
The parent factory had a sequence for the name already `sequence(:name) { |n| "Storage #{n}" }`_

When I created tests creating different states of NextCloud storages I got `Validation failed: Name has already been taken.` validation error. 
As a response I added sequences to some child factories to see if that solves the problem. Together with some more sequence changes I got my tests working. 
Unfortunately I left the exact same naming pattern `sequence(:name) { |n| "Storage #{n}" }` on one child as it is on the parent. This ended up creating the failing tests.
It didn't fail my tests as I only tested Nextcloud storages. But it fails the uniqueness of the names when the tests create Nextcloud and OneDrive storages.

**What caused the Validation error then?**
The Validation errors were wrong. It's weird but I got misleading validation errors locally.
There is also another Validation error I got where the message is pointing to the wrong factory and attribute.
It's still not clear what causes it.

**Why did it got merged?**
First, on my PR I saw the pipeline failing and was wondering if these (for me unknown) tests are failing because of my changes but didn't investigate too much.
Then I rebased my PR with dev and pushed additionally some cleanup commit. 
After that the PR pipeline was successful and I came to the conclusion that the tests are fixed because of the rebase.
I merged the PR to dev and see the previous tests failing :/

**Conclusion**
I ended up removing the sequences on the child factories and see no problems.
The sequences seem to work as one would expect and the sequences in the child factories seem to be unnecessary.

ba1ash (approved):

aaron-contreras:
Gotchu! This seems like a non-trivial thing to debug. Good catch

PR #14269, opened 2023-11-28
Stabilize the Custom Actions Feature Spec

aaron-contreras:
Adding the action doesn't seem to be flaky as it's treated like an ng-select autocomplete. When inspecting the failing screenshots on the CI, it seems that the value on the project condition is not even set:

CI failure screenshot

This tells me that there's probably a timing issue with ng-select that a simple `fill_in :x, with: x` doesn't handle too well. However, I haven't seen a single failure on the prior step which uses the autocomplete logic for filling up the form field.

I'm giving this a try to see if just treating it like another ng-select autocomplete stabilizes the spec.

### DIFF
--- a/spec/support/pages/admin/custom_actions/form.rb
+++ b/spec/support/pages/admin/custom_actions/form.rb
@@ -85,21 +85,11 @@ module Pages
         end
 
         def set_condition(name, value)
-          page.within('#custom-actions-form--conditions') do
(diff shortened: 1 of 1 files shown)


aaron-contreras:
I've re-ran it a couple of times already and not a single failure @cbliard

klaustopher (approved):

PR #14270, opened 2023-11-28
[#51153] Only schedule CRON background job for setting project folder permissions when needed.

ba1ash:
https://community.openproject.org/work_packages/51153

### DIFF
--- a/modules/storages/app/models/storages/project_storage.rb
+++ b/modules/storages/app/models/storages/project_storage.rb
@@ -26,10 +26,6 @@
 # See COPYRIGHT and LICENSE files for more details.
 #++
 
-# A ProjectStorage is a kind of relation between a Storage and
-# a Project in order to enable or disable a Storage for a specific
-# WorkPackages in the project.
-# See also: file_link.rb and storage.rb
 class Storages::ProjectStorage < ApplicationRecord
   using Storages::Peripherals::ServiceResultRefinements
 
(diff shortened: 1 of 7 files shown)


modules/storages/app/models/storages/project_storage.rb
```diff
@@ -51,6 +47,12 @@ class Storages::ProjectStorage < ApplicationRecord
   }.freeze, _prefix: 'project_folder'
 
   scope :automatic, -> { where(project_folder_mode: 'automatic') }
+  scope :active, -> { joins(:project).where(project: { active: true }) }
+  scope :active_nextcloud_automatically_managed, -> do
+    automatic
+      .active
+      .where(storages: Storages::NextcloudStorage.automatically_managed)
+  end
```
mereghost, line 55: WDYT about making a scope `nextcloud_automatically_managed` and use a method to collect the multiple scopes `active.automatic.nextcloud_automatically_managed`?

modules/storages/app/workers/storages/manage_nextcloud_integration_cron_job.rb
```diff
@@ -35,5 +35,13 @@ class ManageNextcloudIntegrationCronJob < Cron::CronJob
     queue_with_priority :low
 
     self.cron_expression = '*/5 * * * *'
+
+    def self.ensure_scheduled!
```
ba1ash, line 39: This method is used by a rake task for scheduling cron_jobs. So, the job is still registered on boot time. Then the rake task calls this method for all registered jobs.

modules/storages/app/services/storages/group_folder_properties_sync_service.rb
```diff
@@ -260,13 +260,18 @@ def remote_group_users
...
+                                      data: { status: error.data.payload.code, body: error.data.payload.body } })
+
+      OpenProject.logger.warn error_message.to_json
     end
 
+    ### Model Scopes
+
     def active_project_storages_scope
-      project_storage_scope.where(project: { active: true })
+      ::Storages::ProjectStorage.active_nextcloud_automatically_managed
```
mereghost: `project_storage_scope` here includes the `automatic` scope, is already scoped to a nextcloud storage (as only those are queried for the job) that has `automatically_managed` set to `true`, so unless you ran into some error (and i'd love to debug this with you) this change might not be needed.

mereghost (approved):

PR #14280, opened 2023-11-29
Retry until ng-select is ready to accept events in `custom_actions_spec.rb`

aaron-contreras:
It seems that  ng-select is not ready to receive events for some reason. I'm unaware of a way to expect it to be ready to receive events other attempting to set the value on the conditions multiple times.

Because the problem doesn't seem to be in finding the dropdown item to click on in the autocompleter but in the expectation for the the value to ACTUALLY be set, retrying on this sequence seems the way to go.

I did a bulk run while under 96% CPU usage with all passing.

This seems to be it! 🍾

```shell
-> % RSPEC_RETRY_COUNT=2 script/bulk_run_rspec ./spec/features/work_packages/custom_actions/custom_actions_spec.rb:137
1 tests to run
================================================================================
Running ./spec/features/work_packages/custom_actions/custom_actions_spec.rb:137 5 times
================================================================================
------------------------------------ Run 0 -------------------------------------
ok
------------------------------------ Run 1 -------------------------------------
ok
------------------------------------ Run 2 -------------------------------------
ok
------------------------------------ Run 3 -------------------------------------
ok
------------------------------------ Run 4 -------------------------------------
ok
PASSED
```

### DIFF
--- a/spec/features/work_packages/custom_actions/custom_actions_spec.rb
+++ b/spec/features/work_packages/custom_actions/custom_actions_spec.rb
@@ -316,15 +316,24 @@ RSpec.describe 'Custom actions', :js, :with_cuprite,
 
     wp_page.click_custom_action('Unassign')
     wp_page.expect_attributes assignee: '-'
+    within '.work-package-details-activities-list' do
+      expect(page)
+        .to have_css('.op-user-activity .op-user-activity--user-name',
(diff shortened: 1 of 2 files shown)


cbliard (approved):
Thanks! Let's see how it goes :crossed_fingers:

PR #14288, opened 2023-11-30
[#51318] add localization key to project storage row component

Kharonus:
[OP#51318](https://community.openproject.org/wp/51318)

- insert localization key


akabiru (approved):
👍🏾

PR #14291, opened 2023-11-30
Allow multiple dynamic Stimulus controllers on a single element

aaron-contreras:
Given the following setup:

```html
<div data-controller="a-controller b-controller"
     data-application-target="dynamic">
  <!-- More HTML -->
</div>
```

We'd run into the following console error:

```
op-application.controller.ts:18 Failed to load dyanmic controller chunk user-limit work-packages--share--user-selected: Error: Cannot find module './user-limit work-packages/share/user-selected.controller'
```

This is because the way the handling for dynamic controllers currently only supports a single controller at a time, causing "a-controller b-controller" to be interpreted as a single Stimulus controller.

My intent with this commit is to be able to have multiple dynamic Stimulus controllers driving an element and therefore "a-controller" AND "b-controller" being loaded as the individual controllers they are.

### DIFF
--- a/frontend/src/stimulus/controllers/op-application.controller.ts
+++ b/frontend/src/stimulus/controllers/op-application.controller.ts
@@ -7,17 +7,20 @@ export class OpApplicationController extends ApplicationController {
   private loaded = new Set<string>();
 
   dynamicTargetConnected(target:HTMLElement) {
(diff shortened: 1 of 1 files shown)


oliverguenther (approved):
Good catch 👍

PR #14315, opened 2023-12-04
[chore] attempt to fix exception in error handling

Kharonus:
Update: this fixes [#51756](https://community.openproject.org/wp/51756)

### Wat?
- oauth controller needs to specifically react to StorageError

### Discussion needed

- The fix here is merely a hack.
- The oauth controller is actually only used by storages code right now, yet it exists outside of the storages namespace.
- This is one of the topics, where we need to be sure, what interface we have. So, right now, we return here a `ServiceResult` with the occasional `StorageError` in the property `errors` of the result (! no array, but a singular error object).

Any suggestions, how we can fix that more gracefully?

### DIFF
--- a/app/controllers/oauth_clients_controller.rb
+++ b/app/controllers/oauth_clients_controller.rb
@@ -118,6 +118,10 @@ class OAuthClientsController < ApplicationController
   end
 
   def set_oauth_errors(service_result)
+    if service_result.errors.is_a?(::Storages::StorageError)
+      service_result.errors = service_result.errors.to_active_model_errors
+    end
+
     flash[:error] = ["#{t(:'oauth_client.errors.oauth_authorization_code_grant_had_errors')}:"]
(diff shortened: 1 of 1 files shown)


Kharonus:
@mereghost and @apfohl 

Can we once more discuss this? Because even in `release/13.2` I still get those bad exceptions, when trying to log in to a storage with bad credentials.

mereghost:
@Kharonus asked:
> Any suggestions, how we can fix that more gracefully?

image

but in all seriousness we need to review this entire flow and figure out a way to make this info to bubble up gracefully.

mereghost (approved):
Solves the issue? ✅ 
Is kinda messy? ✅ 
Good enough while we work on a more graceful solution? ✅ 

By the powers granted to me by the great machine god, I approve this amendment to the sacred code-routines.

image

PR #14316, opened 2023-12-04
Allow users to access the project site when they only allowed to access WorkPackages

klaustopher:
Fixes https://community.openproject.org/work_packages/51267

### DIFF
--- a/app/models/users/permission_checks.rb
+++ b/app/models/users/permission_checks.rb
@@ -118,9 +118,9 @@ module Users::PermissionChecks
   def allowed_based_on_permission_context?(permission, project: nil, entity: nil) # rubocop:disable Metrics/PerceivedComplexity, Metrics/AbcSize
     permissions = Authorization.permissions_for(permission, raise_on_unknown: true)
 
-    entity_blank_or_not_project_scoped = (entity.blank? || !entity.respond_to?(:project) || (entity.respond_to?(:project) && entity.project.blank?))
-    entity_is_work_package_or_list = ((entity.is_a?(WorkPackage) && entity.persisted?) || (entity.is_a?(Enumerable) && entity.all?(WorkPackage)))
-    entity_is_project_scoped_and_project_is_present = (entity.respond_to?(:project) && entity.project.present?)
+    entity_blank_or_not_project_scoped = entity.blank? || !entity.respond_to?(:project) || (entity.respond_to?(:project) && entity.project.blank?)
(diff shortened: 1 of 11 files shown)


ulferts (changes requested):
I believe the approach in here where the calling user would need to have all the permissions defined in the allowed lambda is too restrictive. Take the work_package_table widget for example. It requires having either the `save_queries` or `manage_public_queries` permission. There will be projects where few users have that, e.g. on community. But even without those permissions, the user are rightfully still allowed to see the work packages such widgets show as they have the `view_work_packages` permission. And if they don't then the widget will be empty so no information will leak. The same is true for the members widget. The `allowed` lambda was only meant to restrict the available widgets on administration. There, a user without the `save_queries` permission wouldn't be allowed to use the `work_packages_table` widget because the system would prevent the creation of a query which is necessary for the widget to work. But once the query is set up, the results of the query can be used by all users, each according to their permission sets. 

Removing all the widgets the user is not allowed to see will additionally pose a problem when two people with different permission sets administrate the same overview page. The person with fewer permissions will be tempted to fill the gaps not knowing that in fact there are no gaps which will lead to errors being reported the user cannot understand. 

But I might well miss information on the scenario where users with only shared work packages access the overview page. Why was it necessary to remove the widgets in this scenario?

ulferts:
~~There might be a simpler solution rather then working within the overview page. A user with only shared work packages doesn't necessarily have to see the overview page at all. We could modify the~~ `OverviewsController` ~~to check whether the user is allowed to see the overview page. If not, we redirect the user to the first module the user is allowed to see.~~

Does not work because the information will still get out via the API.

klaustopher:
I can remove it, no problem. The problem I saw was that the widgets Project Description, Project Status would leak information when being displayed to users who only have WP access. I will remove this. Maybe we can refactor the `allowed_to` to `allowed_to_add` or something, that makes the intention clearer

klaustopher:
Removed the widget filtering. Thanks for helping me better understand what it's about :)

ulferts:
> I can remove it, no problem. The problem I saw was that the widgets Project Description, Project Status would leak information when being displayed to users who only have WP access. I will remove this. Maybe we can refactor the `allowed_to` to `allowed_to_add` or something, that makes the intention clearer

Yes, in hindsight, using a different name would have been better. 

And you are right about the Project description and Project status. That information should not get out. The problem with you being right is that we need to filter that at a different layer, in the API response. And with that information being filtered out, the widgets will then no longer be able to show the information which is desired but the widgets are not set up to do it, yet.

I'll open a ticket to remove the information from the project resource.

klaustopher:
Will take a look at this as well. Thanks for pointing out. Is this ok to merge then?

ulferts (approved):

PR #14317, opened 2023-12-04
[#51185] Fix flickering `multi_invite_spec.rb`

aaron-contreras:
## Description
By the time the modal was requested, the Enterprise stub was yet to be registered. Under very rare circumstances did the ViewComponent start rendering after the stub, but most of the time, the request went through quickly enough that the before hook was registered after the user limit section rendered.

Ensuring the before hook is registered before visiting the modal, ensures that the user limit text is present on the first request.

## Notes

Work Package: https://community.openproject.org/work_packages/51185

### DIFF
--- a/spec/features/work_packages/share/multi_invite_spec.rb
+++ b/spec/features/work_packages/share/multi_invite_spec.rb
@@ -191,12 +191,10 @@ RSpec.describe 'Work package sharing',
     let(:global_manager_user) { create(:user, global_permissions: %i[manage_user create_user]) }
     let(:current_user) { global_manager_user }
 
-    before do
+    it 'allows creating multiple users at once' do
       work_package_page.visit!
       click_button 'Share'
-    end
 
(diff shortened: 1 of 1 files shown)


klaustopher (approved):

klaustopher:
Yeah, I think we should use this as a showcase, that using `before` to trigger the rendering of the page is not a good idea if we intend to change things that affect what is rendered in the page. `before` blocks are executed in the order of definition, and there is nothing in RSpec, where we can manually chang the order. Things like `prepend_before` [do not exist](https://stackoverflow.com/questions/17778368/change-the-order-of-before-hooks-in-rspec). We should avoid using `before` to execute anything that is related to running the actual thing under test. Instead we could for example define it in `let(:run_tests)` and then start every test with a `run_test`

aaron-contreras:
Agreed @klaustopher !

PR #14318, opened 2023-12-04
Improve stability of custom_actions_spec

cbliard:
It was often failing with:

```
  1) Custom actions viewing workflow buttons
     Failure/Error:
       expect(page)
         .to have_css('.op-user-activity a.user-mention',
                      text: other_member_user.name,
                      wait: 10)

       expected to find css ".op-user-activity a.user-mention" within #<Capybara::Node::Element tag="div" path="//HTML[1]/BODY[1]/DIV[1]/DIV[1]/MAIN[1]/DIV[2]/OPENPROJECT-BASE[1]/DIV[1]/UI-VIEW[1]/WORK-PACKAGES-BASE[1]/DIV[1]/UI-VIEW[1]/WP-FULL-VIEW-ENTRY[1]/DIV[1]/DIV[2]/DIV[2]/DIV[1]/DIV[1]/OP-WP-TAB[1]/WP-ACTIVITY-TAB[1]/WORK-PACKAGE-COMMENT[1]/DIV[1]/DIV[1]/DIV[1]/DIV[1]"> but there were no matches
```

The activities were correctly fetched from the server, but the last one was not displayed. It's because the activities tab displays acitivites with this for loop: `*ngFor="let inf of visibleActivities; trackBy:trackByIdentifier"`, and the identifier for `ActivityEntryInfo` is composed of `href`, `version`, and `updatedAt` properties from the activity.

The problem is that the `updatedAt` property only has second resolution, and the two custom actions could be triggered within the same second. In that case, the second one would not be displayed as the identifier did not change.

The fix is to send milliseconds from the API for iso8601 date times.

This is only a partial fix as there are other flickering issues with the custom_actions_spec at custom actions creation when setting the project filter. Sometimes the selection does not work and it fails. Still, this is a quality of dev improvement.

It will probably break a few specs. I'll fix them.

### DIFF
--- a/docs/api/apiv3/components/examples/date_alert_notification.yml
+++ b/docs/api/apiv3/components/examples/date_alert_notification.yml
@@ -5,8 +5,8 @@ value:
   id: 1
   readIAN: false
   reason: dateAlert
-  createdAt: '2022-04-05T14:38:28Z'
-  updatedAt: '2022-04-06T09:03:24Z'
+  createdAt: '2022-04-05T14:38:28.361Z'
+  updatedAt: '2022-04-06T09:03:24.347Z'
   _embedded:
     project:
       _hint: Project resource shortened for brevity
--- a/docs/api/apiv3/components/examples/grid-simple-collection-response.yml
+++ b/docs/api/apiv3/components/examples/grid-simple-collection-response.yml
@@ -31,8 +31,8 @@ value:
             endRow: 3
             startColumn: 3
             endColumn: 6
-        createdAt: '2018-12-03T16:58:30Z'
-        updatedAt: '2018-12-13T19:36:40Z'
+        createdAt: '2018-12-03T16:58:30.297Z'
+        updatedAt: '2018-12-13T19:36:40.352Z'
         _links:
           scope:
             href: "/my/page"
--- a/docs/api/apiv3/components/examples/grid-simple-response.yml
+++ b/docs/api/apiv3/components/examples/grid-simple-response.yml
@@ -24,8 +24,8 @@ value:
       endRow: 3
       startColumn: 3
       endColumn: 6
-  createdAt: '2018-12-03T16:58:30Z'
-  updatedAt: '2018-12-13T19:36:40Z'
+  createdAt: '2018-12-03T16:58:30.915Z'
+  updatedAt: '2018-12-13T19:36:40.588Z'
   _links:
     self:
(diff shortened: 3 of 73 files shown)


klaustopher:
Wouldn't it make sense to instead add specific `updated_at` timestamps when creating the models, if they are relevant. Because even if adding the miliseconds to the iso8601 version of the timestamp, we are not guaranteed for them to be different (i.e. when the models are generated within a transaction, and `NOW()` is used to set the timestamp.

cbliard:
> Wouldn't it make sense to instead add specific `updated_at` timestamps when creating the models, if they are relevant. Because even if adding the miliseconds to the iso8601 version of the timestamp, we are not guaranteed for them to be different (i.e. when the models are generated within a transaction, and `NOW()` is used to set the timestamp.

Your idea seems interesting, what do you mean exactly by adding "a specific `updated_at` timestamp when creating the model"? Would you have an example?

Anyway if there are multiple models, they'll have a different href (because different id) and probably different version. The issue is only when a model gets updated within the same second.

klaustopher:
You can specify the `updated_at` attribute in the `create` call, for an AR model and also for a factory:

```ruby
let(:foo1) { create(:user, updated_at: 1.minute.ago) }
let(:foo2) { create(:user, updated_at: 30.seconds.ago) }
```

When the `updated_at` field is relevant, we should give it an explicit value in our tests and not rely on implicit properties that could lead to a problem like the one here

cbliard:
Got it, you meant in the test. Yes that would work, but there would still be the issue of subsequent updates of an activity not being shown if the update occurred within the same second. It still feels like a bug to me.

Here the update comes from doing 2 custom actions one after the other (unassign a user, then "escalate" the work package which sets some properties on it) and we do not have control over the updated at. The bug could still be worked around by adding a `sleep 1` between the clicks on the custom action buttons. But again, I feel like it's a bug and needs fixing rather than working around it.

klaustopher:
But yeah, definetly sounds like we need a better attribute to discriminate between two actions. 🤔

cbliard:
Having an identifier prevents rerendering when the object did not change. So far it includes only the url (id), the version and updated at. Upon the second action, the same journal will be updated with new attributes and a new updated at value. Another thing to do could be to have some hash of the `details` and `notes` fields. That would be most reliable but would also be a tad more complicated.

aaron-contreras (approved):
🎸

PR #14345, opened 2023-12-06
[#50437] Handle cases when project_folder_id is empty.

ba1ash:
https://community.openproject.org/work_packages/50437

#### Problem:
  If project_folder_id is empty(the background job is slow) and
  project_folder_mode is automatic and
  user tries to open `[openproject_host]/project/[project_identifier]/project_storages/[project_storage_id]/open`
  500 error happens because file_info_query has been called with empty file_id.

### Solution:
  Open project overview with waiting modal in the problematic case.


Kharonus (approved):
LGTM

PR #14351, opened 2023-12-07
Fix time logging issue

klaustopher:
We had several issues:

- To generate the schema, we created a new blank work package, so the theck for `allowed_in_work_package?` was false. Fixed the schema to correctly check `allowed_in_any_work_package?(in_project:)`
- The TimeEntries::Visible scope was never used, instead this method was already defined via the `CostScopes` module. So, I removed the scope
- The `CostScopes` correctly checks the `_own_` permissions on the work_package if it is applicable.

---

Fixes https://community.openproject.org/projects/openproject/work_packages/51403


aaron-contreras (changes requested):
Tried it out with the following steps and it seems I'm not able to see my own logged time:

- Share a work package with Edit rights to a user
- Log in as said user
- Visit the work package
- Log some time

An admin is able to see the logged time
Admin View

But not myself
My point of view

klaustopher:
You need to disable/clear your cache. You can tell by the `spentTime` field not being present. You still have an old version of the schema cached that does not include the field.

aaron-contreras:
Okay, that worked. I guess cache is automatically invalidated on live environments once this is deployed?

aaron-contreras (approved):

klaustopher:
We added the version to the cache key of the representers. This way, the key will be invalidated with a new version

PR #14354, opened 2023-12-07
[#51033] Enable OAuth Client Form submit button and show validation errors in case of invalid submission

akabiru:
#### https://community.openproject.org/work_packages/51033

Screenshot 2023-12-15 at 4 25 57 PM


modules/storages/spec/features/admin_storages_spec.rb
```diff
@@ -201,15 +201,17 @@
...
                                                text: 'Copy these values from Nextcloud Administration / OpenProject.')
 
-            # With null values, submit button should be disabled
+            # With null values, form should render inline errors
             expect(page).to have_css('#oauth_client_client_id', value: '')
             expect(page).to have_css('#oauth_client_client_secret', value: '')
-            expect(find_test_selector('storage-oauth-client-submit-button')).to be_disabled
+            click_button 'Save and continue'
+
+            expect(page).to have_text("Client can't be blank.")
```
Kharonus: 🟡 nit: shouldn't the validation text say: "Client ID can't be blank."?

app/services/oauth_clients/create_service.rb
```diff
@@ -36,10 +36,9 @@ module OAuthClients
   class CreateService < ::BaseServices::Create
     protected
 
-    def before_perform(params, _service_result)
-      OAuthClient.where(integration: params[:integration]).delete_all
-
-      super
+    def after_validate(params, contract_call)
```
Kharonus, line 39: ❓ why changing the hook? what is the difference, in the `before_perform` and the `after_validate`?

Kharonus (approved):
LGTM, could be merged as is, if you want.

There is one question I have, and I think validation text should get amended.

app/services/oauth_clients/create_service.rb
akabiru: Sure, good eye- the difference is the order of operations: if the contract fails validation, we do not want to purge the database of the previous OAuth clients i.e. we should only purge the previous oauth clients, only with a valid contract.

Ref: https://github.com/opf/openproject/blob/dev/app/services/base_services/base_contracted.rb#L59-L68

modules/storages/spec/features/admin_storages_spec.rb
akabiru: Good eye, this is a primer side effect of any form object with an `_id` prefix in the name- I have not be able to find a workaround but can log a bug ticket for investigation. Any ideas @HDinger @jonasjabari ?

akabiru: Solved by overriding the contract validation message: https://github.com/opf/openproject/pull/14354/commits/ca0fc06d369ab3dc3c6d8a83290df1f612ba339b

Kharonus: looks good!

app/services/oauth_clients/create_service.rb
Kharonus: makes perfect sense, question answered

Kharonus (approved):
LGTM, if CI is green feel free to merge.

app/services/oauth_clients/create_service.rb
```diff
@@ -36,10 +36,9 @@ module OAuthClients
...
+      OAuthClient.where(integration: params[:integration]).delete_all if contract_call.success?
```
dominic-braeunlein, line 40: Hi @akabiru, Wieland and I QAed this PR earlier and found something small and not urgent here.

We realised that `if contract_call.success?` should not be needed as `after_validate` is only called when `contract_call.success?` [base_contracted.rb#L63](https://github.com/opf/openproject/blob/a06e519ed789984cdfb2ef80a1d760ffe23e3cb4/app/services/base_services/base_contracted.rb#L63)
Maybe there is anything I am missing, but if not it would be nice to remove this bit for clarity.

akabiru: That's a really nice catch @dominic-braeunlein - I'll update this in dev 👍🏾

akabiru: Done! https://github.com/opf/openproject/pull/14449

PR #14361, opened 2023-12-07
update seeded permissions

ulferts:
Please see the file in the referenced WP for the specification:

https://community.openproject.org/wp/50827

### DIFF
--- a/app/seeders/common.yml
+++ b/app/seeders/common.yml
@@ -280,8 +280,6 @@ project_roles:
       - :view_work_packages
       - :view_calendar
       - :comment_news
-      - :browse_repository
-      - :view_changesets
       - :view_wiki_pages
   - reference: :default_role_anonymous
     t_name: Anonymous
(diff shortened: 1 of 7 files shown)


app/seeders/common.yml
```diff
@@ -280,17 +280,13 @@ project_roles:
       - :view_work_packages
       - :view_calendar
       - :comment_news
-      - :browse_repository
-      - :view_changesets
```
aaron-contreras, line 284: These are being removed from all seeds taking place as of the moment the PR is merged in. Is it okay if the non member and anonymous roles seeded prior to this keep the two removed permissions? Is a migration not needed to remove them from already seeded instances?

aaron-contreras (approved):
Just left a question. Looks good :)

app/seeders/common.yml
ulferts: I would leave them as they are. We cannot know at this point whether the roles have the permissions out of a conscious decision by some user or it is just the result of seeding.

PR #14368, opened 2023-12-08
fix authorization checks with decorated work packages

ulferts:


ulferts:
Hi @aaron-contreras , can you please check if that fixed the problem on your machine as well?

aaron-contreras:
Yes ⭐
Thank you

aaron-contreras (approved):

PR #14370, opened 2023-12-11
Do not display project infos on the index page

klaustopher:
The index page currently still displays information about the project that a user with only shared work packages should not see


ulferts (changes requested):
Good catch @klaustopher .

Please add a test e.g. in spec/features/projects/project_index_spec.rb and this is good to merge.

ulferts (approved):
👍

PR #14372, opened 2023-12-11
Fix `in_project` scope that is used by the members filter

klaustopher:
The filter that was used only returned users, when they were explicit members of the project. Instead the filter now returns all users that have any access to items within the project.

Fixes https://community.openproject.org/projects/openproject/work_packages/51476/activity

### DIFF
--- a/app/models/queries/principals/filters/access_to_anything_in_project_filter.rb
+++ b/app/models/queries/principals/filters/access_to_anything_in_project_filter.rb
@@ -28,7 +28,10 @@
 
 class Queries::Principals::Filters::AccessToAnythingInProjectFilter < Queries::Principals::Filters::PrincipalFilter
   def allowed_values
-    Project.active.pluck(:name, :id)
+    Project
+    .visible(User.current)
+    .active
+    .pluck(:name, :id)
   end
 
   def type
@@ -61,6 +64,6 @@ class Queries::Principals::Filters::AccessToAnythingInProjectFilter < Queries::P
   def member_included_scope
     visible_scope
       .includes(:members)
-      .merge(Member.of_any_project)
+      .merge(Member.where.not(project: nil))
   end
 end
--- a/app/models/queries/principals/filters/access_to_anything_in_project_filter.rb
+++ b/app/models/queries/principals/filters/access_to_anything_in_project_filter.rb
@@ -26,18 +26,41 @@
 # See COPYRIGHT and LICENSE files for more details.
 #++
 
-module Queries::Principals
-  ::Queries::Register.register(PrincipalQuery) do
-    filter Filters::AccessToAnythingInProjectFilter
-    filter Filters::TypeFilter
-    filter Filters::MemberFilter
-    filter Filters::MentionableOnWorkPackageFilter
-    filter Filters::StatusFilter
-    filter Filters::NameFilter
-    filter Filters::AnyNameAttributeFilter
-    filter Filters::TypeaheadFilter
-    filter Filters::IdFilter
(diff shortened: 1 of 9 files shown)


ulferts (changes requested):
I think it is better to have a separate filter for this. The members filter is used at a couple of places.

klaustopher:
Yeah, did exactly that.

app/models/queries/work_packages/filter/shared_with_user_filter.rb
```diff
@@ -70,9 +70,8 @@ def view_shared_work_packages_allowed?
   end
 
   def visible_shared_work_packages
-    WorkPackage.joins("JOIN members ON members.entity_id = work_packages.id")
-               .where(members: { entity_type: 'WorkPackage',
-                                 project: visible_projects })
+    WorkPackage.joins("JOIN members ON members.entity_type = 'WorkPackage' AND members.entity_id = work_packages.id")
+               .where(members: { project: visible_projects })
```
aaron-contreras, line 74: Thank you

lib/api/v3/queries/schemas/project_members_filter_dependency_representer.rb
```diff
@@ -30,7 +30,7 @@ module API
   module V3
     module Queries
       module Schemas
-        class AllPrincipalsFilterDependencyRepresenter <
+        class ProjectMembersFilterDependencyRepresenter <
```
aaron-contreras, line 33: ⭐️

app/models/queries/principals/filters/access_to_anything_in_project_filter.rb
```diff
@@ -0,0 +1,66 @@
...
+# You should have received a copy of the GNU General Public License
+# along with this program; if not, write to the Free Software
+# Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+#
+# See COPYRIGHT and LICENSE files for more details.
+#++
+
+class Queries::Principals::Filters::AccessToAnythingInProjectFilter < Queries::Principals::Filters::PrincipalFilter
+  def allowed_values
+    Project.active.pluck(:name, :id)
```
ulferts: This would need to be `Project.active(User.current).active`. I suspect this was copied over from the `MemberFilter` where this is also wrong.

app/models/queries/principals/filters/access_to_anything_in_project_filter.rb
```diff
@@ -0,0 +1,66 @@
...
+  private
+
+  def visible_scope
+    Principal.visible(User.current)
+  end
+
+  def member_included_scope
+    visible_scope
+      .includes(:members)
+      .merge(Member.of_any_project)
```
ulferts: As this filter is called `AccessToAnythingInProject' this scope would have to be:

```
visible_scope.includes(:members).merge(Member.where.not(project: nil))
```
or
```
visible_scope.includes(:members).merge(Member.where.not(id: Member.global))
```

to get all non global memberships.

ulferts (approved):
Hi @klaustopher ,

except for two small things, it looks good. Feel free to merge after addressing the remarks.

app/models/queries/principals/filters/access_to_anything_in_project_filter.rb
klaustopher: Do you mean `visible`? Because `active` does not take any arguments

ulferts: Yeah, that, sorry for the confusion.

PR #14381, opened 2023-12-11
Fix flaky account activation spec

aaron-contreras:
Uses the page object method that waits until the capabilities request
comes through and renders the "Share" button.


akabiru (approved):
🥇

PR #14384, opened 2023-12-11
[51193] Move remaining work field into core

cbliard:
https://community.openproject.org/wp/51193


cbliard:
The test fails because of a flaky test that was already fixed on `release/13.1` branch in #14383. I've merged the `release/13.1` branch into `dev`.

.rubocop.yml
```diff
@@ -164,6 +164,10 @@ Rails/I18nLocaleAssignment:
   Exclude:
     - 'spec/**/*.rb'
 
+# Do not bother if `let` statements use an index in their name
```
aaron-contreras, line 167: "Do not bother" 😂

modules/backlogs/spec/factories/story_factory.rb
```diff
@@ -28,10 +28,10 @@
...
-    association :priority, factory: :priority
+    priority factory: :priority
     sequence(:subject) { |n| "story#{n}" }
     description { 'story story story' }
-    association :type, factory: :type_feature
-    association :author, factory: :user
+    type factory: :type_feature
+    author factory: :user
   end
 end
```
aaron-contreras, line 37: 👍

.rubocop.yml
```diff
@@ -164,6 +164,10 @@ Rails/I18nLocaleAssignment:
...
+RSpec/IndexedLet:
+  Enabled: false
```
ulferts, line 169: 👍 

I never got the reason for that

modules/backlogs/config/locales/en.yml
```diff
@@ -175,7 +171,7 @@ en:
   rb_label_copy_tasks_open: "Open"
   rb_label_link_to_original: "Include link to original story"
 
-  remaining_hours: "remaining hours"
+  remaining_hours: "remaining work"
 
   required_burn_rate_hours: "required burn rate (hours)"
   required_burn_rate_points: "required burn rate (points)"
```
ulferts, line 177: What is the reason for the translations being in here when the implementation is moved to the core?

modules/backlogs/lib/open_project/backlogs/engine.rb
```diff
@@ -120,9 +120,6 @@ def self.settings
            before: :settings_storage
     end
 
-    # We still override version and project settings views from the core! URH
-    override_core_views!
```
ulferts, line 124: 👍

app/contracts/work_packages/base_contract.rb
```diff
@@ -60,6 +60,10 @@ class BaseContract < ::ModelContract
     attribute :derived_estimated_hours,
               writable: false
 
+    attribute :remaining_hours
+    attribute :derived_remaining_hours,
+              writable: false
```
ulferts, line 65: Is this declaration actually necessary? The most basic use case for adding an attribute to the contract is to make it writable. But here the purpose is the exact opposite.

ulferts (approved):
The change looks good. 

There are minuscule remarks but feel free to merge at will, @cbliard .

modules/backlogs/config/locales/en.yml
cbliard: It was forgotten in a previous commit.

app/contracts/work_packages/base_contract.rb
cbliard: I did not really think about it. I just moved it.
The lines are identical above for `:derived_estimated_hours`. 

At least it is explicit so that people in the future does not ask themself: "`:derived_remaining_hours` is not writable because it is not declared as an attribute, but is it on purpose or has it been forgotten?". Having it explicitly marked as `writable: false` removes any ambiguity about it.

PR #14392, opened 2023-12-12
WP Descendant query also allows single WP access

klaustopher:
The query still limited access to only those WPs where you had access throught the entire project. This change fixes the access query so that the time is correctly displayed.

Fixes https://community.openproject.org/work_packages/51403/activity


cbliard (changes requested):
Can you add a unit test for it?

klaustopher:
There are no Unit tests for any of those scopes

cbliard (approved):

PR #14396, opened 2023-12-12
[chore] added project storages and ancestors

Kharonus:
- project model in API spec now contains `projectStorages` and `ancestors`


machisuji (approved):
👍

PR #14405, opened 2023-12-12
[#51538] Use ensure_connection to open storage from files tab

ba1ash:
https://community.openproject.org/work_packages/51538


docs/api/apiv3/components/schemas/project_storage_model.yml
```diff
@@ -69,14 +69,27 @@ properties:
...
+              A link to open project strorage.
+      openWithConnectionEnsured:
+        allOf:
+          - $ref: './link.yml'
+          - description: |-
+              A link to open project storage with making sure user has access to it.
+
+              # Conditions
 
+              If the storage has not been configured(oauth client is missing, for instance), then the link is empty.
```
Kharonus: instead of `link is empty` it should say `link is null`

Kharonus (approved):
LGTM, I'd vote for being explicit of having a `null` href in the link (in the specification). When a representer has `{ href: nil }` it renders a json to `{"href":null}`.

PR #14407, opened 2023-12-12
Speed up a spec by using shared_let.

ba1ash:
modules/storages/spec/requests/api/v3/project_storages/project_storages_spec.rb
### Before
specs_before

### After
specs_after


aaron-contreras (commented):
Massive!

akabiru (approved):
Woah, amazing! I would love a primer on `shared_let`- haven't seen this use case before 🙌🏾

PR #14414, opened 2023-12-13
improve the notification performance

ulferts:
This improvements has two parts:
* Avoid reapplying the visibility check for notifications.
* Merging the constraints rather than having them in a subselect.

It requires moving the visibility scope of principals as that would otherwise not be applied correctly. But since the Principal's visibility scope does already apply the constraints conditionally, moving does not change the behaviour.

https://community.openproject.org/wp/51622

### DIFF
--- a/app/models/queries/principals/principal_query.rb
+++ b/app/models/queries/principals/principal_query.rb
@@ -32,6 +32,6 @@ class Queries::Principals::PrincipalQuery < Queries::BaseQuery
   end
 
   def default_scope
-    Principal.not_builtin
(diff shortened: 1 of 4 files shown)


klaustopher (approved):

klaustopher:
It seems to be messing with the onboarding tour. Any idea how those changes interfere?

ulferts:
The specs got broken before this PR. @klaustopher already issued a PR for fixing those #14417.

Testing the altered statements against production data indicates, that the new statement is roughly an order of magnitude faster than the old one.

PR #14417, opened 2023-12-13
Fix tests that rely on notifications being pulled all 10 seconds

klaustopher:
In #14413 we added a configurable amount for notification polling and the new default is 60 seconds instead of the old hard coded 10 seconds. This causes some tests to fail because they rely on some behavior that is triggered by the notification polling. Those tests now get a setting to set the notification timeout to the 10 seconds it was before.


ulferts (approved):
Thanks a ton, @klaustopher

PR #14424, opened 2023-12-15
fix slow project visible scope for admins

ulferts:
Short circuit the visible check for admins since having a work package shared with them is irrelevant for the visible calculation. The only condition relevant for them is for the project to be active.

It reduces the runtime of the SQL query on the dataset tested with from roughly 50s to 5ms.

This is a quick fix for https://community.openproject.org/wp/51659

The visible scope is still not as optimized as it could be. For a non admin user it will: 
* Check if the user has a project membership
* Check if the user has no project membership and the project is public
* Check if the user has the permission via a work package membership which involves
  * Checking if the user has a work package share with the :view_work_packages permission
  * Checking if the user has a project membership granting the :view_work_packages permission
  * Checking if the non member role has the :view_work_packages permission, the user has no membership and the project is public

This is done conveniently by joining two available scopes. But it has a lot of overhead (e.g. the check for the non member to have the :view_work_packages permission is redundant with checking if the project is public before).

Since both scopes work on the same data structure (e.g. `members`, `member_roles` and `roles`), it might well be possible to have them use the exact same query, without nesting, and just querying for multiple permissions at the same time (`:view_project` and `:view_work_packages`). This however is not part of the PR but will be done in a subsequent one.

### DIFF
--- a/app/models/project.rb
+++ b/app/models/project.rb
@@ -150,13 +150,13 @@ class Project < ApplicationRecord
   friendly_id :identifier, use: :finders
 
   include ::Scopes::Scoped
-  scopes :allowed_to
+  scopes :allowed_to,
+         :visible
 
   scope :has_module, ->(mod) {
     where(["#{Project.table_name}.id IN (SELECT em.project_id FROM #{EnabledModule.table_name} em WHERE em.name=?)", mod.to_s])
   }
   scope :public_projects, -> { where(public: true) }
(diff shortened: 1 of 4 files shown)


ulferts:
The two failing specs are failing prior to the PR as well.

klaustopher (approved):
Good job! We can definetly improve the checks further in the future. Especially by eliminating the second checks on the project

PR #14425, opened 2023-12-15
[51656] Fix missing translation user wrote in wp activity

cbliard:
https://community.openproject.org/wp/51656

In work package activity, when quoting a message, the text "<username> wrote:" was hard coded. This PR changes it to be translated in the instance default language.


dombesz (approved):
LGTM! 👍

PR #14434, opened 2023-12-18
Copy .ruby-version file for bundle install

cbliard:
As in Gemfile we now use `ruby file: '.ruby-version'`, this file is necessary to run `bundler install`.


machisuji (approved):
👍

PR #14437, opened 2023-12-18
[#51316] Poor performance of user groups when I remove user from group + 500 error

ba1ash:
https://community.openproject.org/work_packages/51316


mereghost (approved):
Seems like a good solution until we have a more robust background job lib.

PR #14443, opened 2023-12-19
Fix of collect_ancestor_attributes to respect overrides in subclasses

virgo47:
Because collect_ancestor_attributes_by goes from the subclass attributes and then combining them with the superclass attributes, merge overrides more specific definitions (from subclass) with the more generic ones. Instead, `reverse_merge!` can be used, even merging in place as the attributes are duped first.

This was discovered during the development of PR #14120.

### DIFF
--- a/app/contracts/base_contract.rb
+++ b/app/contracts/base_contract.rb
@@ -180,9 +180,10 @@ class BaseContract < Disposable::Twin
 
   # Traverse ancestor hierarchy to collect contract information.
   # This allows to define attributes on a common base class of two or more contracts.
+  # Reverse merge is important to use the more specific overrides from subclasses.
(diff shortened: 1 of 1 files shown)


virgo47:
@cbliard Can you please check this `collect_ancestor_attributes` fix we talked about? Nothing seems broken from here. :-)

EDIT: Or, alternatively, we can just close this as an experiment without a merge and I'll reimplement it in #14120 where some new behavior and tests actually depend on it?

virgo47:
I put it into the other branch as well because I needed it there anyway, so no need to merge this.

cbliard:
Great!
Yep, keeping this PR as experiment is :ok_hand:

PR #14448, opened 2023-12-19
Speed up specs by using shared_let.

ba1ash:
Timings on my local machine before: 
specs_before
After:
specs_after


akabiru (approved):
Woah, you're on a roll! 🙌🏾 ⚡

PR #14454, opened 2023-12-20
[#51316] Use RequestStore instead of Thread.current.

ba1ash:
https://community.openproject.org/work_packages/51316

### DIFF
--- a/modules/storages/app/workers/storages/manage_nextcloud_integration_events_job.rb
+++ b/modules/storages/app/workers/storages/manage_nextcloud_integration_events_job.rb
@@ -30,22 +30,23 @@ module Storages
   class ManageNextcloudIntegrationEventsJob < ApplicationJob
     include ManageNextcloudIntegrationJobMixin
 
-    DEBOUNCE_TIME = 5.seconds.freeze
+    THREAD_DEBOUNCE_TIME = 4.seconds.freeze
+    JOB_DEBOUNCE_TIME = 5.seconds.freeze
     THREAD_KEY = :manage_nextcloud_integration_events_job_debounce_happend_at
 
     queue_with_priority :above_normal
 
     def self.debounce
-      last_debounce_happend_at = Thread.current[THREAD_KEY]
-      if last_debounce_happend_at.blank? || Time.current > (last_debounce_happend_at + DEBOUNCE_TIME)
+      last_debounce_happend_at = RequestStore.store[THREAD_KEY]
+      if last_debounce_happend_at.blank? || Time.current > (last_debounce_happend_at + THREAD_DEBOUNCE_TIME)
         count = Delayed::Job
                   .where("handler LIKE ?", "%job_class: #{self}%")
                   .where(locked_at: nil)
-                  .where('run_at <= ?', DEBOUNCE_TIME.from_now)
+                  .where('run_at <= ?', JOB_DEBOUNCE_TIME.from_now)
                   .delete_all
         Rails.logger.info("deleted: #{count} jobs")
-        Thread.current[THREAD_KEY] = Time.current
-        set(wait: DEBOUNCE_TIME).perform_later
+        RequestStore.store[THREAD_KEY] = Time.current
+        set(wait: JOB_DEBOUNCE_TIME).perform_later
       end
     end
 
--- a/modules/storages/app/workers/storages/manage_nextcloud_integration_events_job.rb
+++ b/modules/storages/app/workers/storages/manage_nextcloud_integration_events_job.rb
@@ -30,23 +30,19 @@ module Storages
   class ManageNextcloudIntegrationEventsJob < ApplicationJob
     include ManageNextcloudIntegrationJobMixin
 
-    THREAD_DEBOUNCE_TIME = 4.seconds.freeze
-    JOB_DEBOUNCE_TIME = 5.seconds.freeze
-    THREAD_KEY = :manage_nextcloud_integration_events_job_debounce_happend_at
+    SINGLE_THREAD_DEBOUNCE_TIME = 4.seconds.freeze
+    MULTI_THREAD_DEBOUNCE_TIME = 5.seconds.freeze
+    KEY = :manage_nextcloud_integration_events_job_debounce_happend_at
 
     queue_with_priority :above_normal
 
     def self.debounce
-      last_debounce_happend_at = RequestStore.store[THREAD_KEY]
-      if last_debounce_happend_at.blank? || Time.current > (last_debounce_happend_at + THREAD_DEBOUNCE_TIME)
-        count = Delayed::Job
-                  .where("handler LIKE ?", "%job_class: #{self}%")
-                  .where(locked_at: nil)
-                  .where('run_at <= ?', JOB_DEBOUNCE_TIME.from_now)
-                  .delete_all
-        Rails.logger.info("deleted: #{count} jobs")
-        RequestStore.store[THREAD_KEY] = Time.current
-        set(wait: JOB_DEBOUNCE_TIME).perform_later
+      last_debounce_happend_at = RequestStore.store[KEY]
+      if last_debounce_happend_at.blank? || Time.current > (last_debounce_happend_at + SINGLE_THREAD_DEBOUNCE_TIME)
+        Rails.cache.fetch(KEY, expires_in: MULTI_THREAD_DEBOUNCE_TIME) do
+          set(wait: MULTI_THREAD_DEBOUNCE_TIME).perform_later
+          RequestStore.store[KEY] = Time.current
+        end
       end
     end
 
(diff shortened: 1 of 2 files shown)


modules/storages/app/workers/storages/manage_nextcloud_integration_events_job.rb
```diff
@@ -30,22 +30,23 @@ module Storages
   class ManageNextcloudIntegrationEventsJob < ApplicationJob
     include ManageNextcloudIntegrationJobMixin
 
-    DEBOUNCE_TIME = 5.seconds.freeze
+    THREAD_DEBOUNCE_TIME = 4.seconds.freeze
+    JOB_DEBOUNCE_TIME = 5.seconds.freeze
     THREAD_KEY = :manage_nextcloud_integration_events_job_debounce_happend_at
```
wielinde: I guess `THREAD_KEY` is not the correct naming anymore, now that we do not fiddle around with threads. Maybe `StoreKey`?

modules/storages/app/workers/storages/manage_nextcloud_integration_events_job.rb
```diff
@@ -30,22 +30,23 @@ module Storages
...
 
     queue_with_priority :above_normal
 
     def self.debounce
-      last_debounce_happend_at = Thread.current[THREAD_KEY]
-      if last_debounce_happend_at.blank? || Time.current > (last_debounce_happend_at + DEBOUNCE_TIME)
+      last_debounce_happend_at = RequestStore.store[THREAD_KEY]
+      if last_debounce_happend_at.blank? || Time.current > (last_debounce_happend_at + THREAD_DEBOUNCE_TIME)
```
wielinde: Again `THREAD` terminology. `THREAD_DEBOUNCE_TIME`... better `REQUEST_DEBOUNCE_TIME`?

modules/storages/app/workers/storages/manage_nextcloud_integration_events_job.rb
```diff
@@ -30,22 +30,23 @@ module Storages
...
         count = Delayed::Job
                   .where("handler LIKE ?", "%job_class: #{self}%")
                   .where(locked_at: nil)
-                  .where('run_at <= ?', DEBOUNCE_TIME.from_now)
+                  .where('run_at <= ?', JOB_DEBOUNCE_TIME.from_now)
                   .delete_all
         Rails.logger.info("deleted: #{count} jobs")
-        Thread.current[THREAD_KEY] = Time.current
-        set(wait: DEBOUNCE_TIME).perform_later
+        RequestStore.store[THREAD_KEY] = Time.current
```
wielinde: @ba1ash if we update this value the overall debounce time span will get bigger than 4 seconds. However, we need it to be small than the `JOB_DEBOUNCE_TIME` (5 seconds), right? With this current code, it can get bigger than 5 seconds. So it could lead to skipping jobs that should have run.

wielinde (changes requested):
@ba1ash Please find my inline comments

mereghost:
And to think this is all dealt with by `GoodJob`s concurrency settings out of the box >.<

modules/storages/app/workers/storages/manage_nextcloud_integration_events_job.rb
```diff
@@ -30,22 +30,23 @@ module Storages
...
 
     queue_with_priority :above_normal
 
     def self.debounce
-      last_debounce_happend_at = Thread.current[THREAD_KEY]
-      if last_debounce_happend_at.blank? || Time.current > (last_debounce_happend_at + DEBOUNCE_TIME)
+      last_debounce_happend_at = RequestStore.store[THREAD_KEY]
+      if last_debounce_happend_at.blank? || Time.current > (last_debounce_happend_at + THREAD_DEBOUNCE_TIME)
         count = Delayed::Job
                   .where("handler LIKE ?", "%job_class: #{self}%")
```
machisuji: Alert! This will lead to minor catastrophe in the existing code before this PR already.

In SaaS the delayed jobs of all tenants are in a shared table in the public schema.
This code will delete other tenants' jobs as well.

To fix this we would need to be able to patch this line from the multitenancy plugin to add a `.where("handler LIKE ?", "%namespace: #{Apartment::Tenant.current}%")`.

But I dislike this solution overall. Can we not find something that could make due without having to mess with the delayed jobs table?

machisuji (changes requested):
Having had some more time to think about this I think we can find a better solution that using `RequestStore` as well. One for the debounce to cover all threads. I propose using the cache for that. Using goodjob's concurrency controls may be a better solution once we use goodjob.

Along the lines of

```
time = Time.now.min / 4 # 4-minute windows
Rails.cache.fetch("integrations-during-min-#{time}") do
  Storages::ManageNextcloudIntegrationCronJob.set(wait: 4.minutes).perform_later
end
```

So the first call would enqueue the delayed job to happen in 4 minutes.
Every other call won't do anything since the result is present in the cache.
And this is the same across all threads and all processes.

Calls beyond those 4 minutes will enqueue a new job.

Does that make sense?

edit: ok maybe it was 4 seconds and not 4 minutes, but you get the idea.
edit 2: Rails.cache is namespaced per tenant so this would not require any changes in the multitenancy plugin.

modules/storages/app/workers/storages/manage_nextcloud_integration_events_job.rb
ba1ash: I believe having THREAD terminology is fine. It highlights that debounce time operates in a thread scope. RequestStore basically uses Thread variables as well. The only difference is that it cleans them up at the end of request cycle in the middleware. So, I suggest having the following names: 
```
    THREAD_DEBOUNCE_TIME = 4.seconds.freeze
    JOB_DEBOUNCE_TIME = 5.seconds.freeze
    REQUEST_STORE_KEY = :manage_nextcloud_integration_events_job_debounce_happend_at
```
Will it be good for you?

modules/storages/app/workers/storages/manage_nextcloud_integration_events_job.rb
ba1ash: I think in this case the name is all right. Check comment above, please.

modules/storages/app/workers/storages/manage_nextcloud_integration_events_job.rb
ba1ash: So, it means we have it already for quite some time :(

modules/storages/app/workers/storages/manage_nextcloud_integration_events_job.rb
ba1ash: It would be nice if you could clarify it. I do not get it right away. Let's discuss it when you can.

modules/storages/app/workers/storages/manage_nextcloud_integration_events_job.rb
```diff
@@ -30,22 +30,20 @@ module Storages
...
-        count = Delayed::Job
-                  .where("handler LIKE ?", "%job_class: #{self}%")
-                  .where(locked_at: nil)
-                  .where('run_at <= ?', DEBOUNCE_TIME.from_now)
-                  .delete_all
-        Rails.logger.info("deleted: #{count} jobs")
-        Thread.current[THREAD_KEY] = Time.current
-        set(wait: DEBOUNCE_TIME).perform_later
+      last_current_thread_debounce_happend_at = RequestStore.store[KEY]
+      if last_current_thread_debounce_happend_at.blank? ||
```
machisuji: @ba1ash 

Could we please put these two lines (40 to 42) into a method. And as much as I like descriptive variable names I think in a separate method we could get away with a shorter variable name for it and give the method the descriptive name instead.

modules/storages/spec/workers/storages/manage_nextcloud_integration_events_job_spec.rb
```diff
@@ -39,36 +39,30 @@
...
-      expect(Delayed::Job.pluck(:id)).not_to include(same_handler_within_timeframe1,
-                                                     same_handler_within_timeframe2,
-                                                     same_handler_within_timeframe3)
-      expect(
-        Delayed::Job
-          .where("handler LIKE ?", "%job_class: #{described_class}%")
-          .last
-          .run_at
-      ).to be_within(3.seconds).of(described_class::DEBOUNCE_TIME.from_now)
+        expect(Rails.cache).to have_received(:fetch).once
```
machisuji, line 64: 👍

machisuji (changes requested):
Hi @ba1ash, thanks for adapting the PR. Looks good to me in principle. Could just ask for one change (see comment) to make the code a little bit easier to read.

modules/storages/app/workers/storages/manage_nextcloud_integration_events_job.rb
ba1ash: Yes, sure.

machisuji (approved):
👍

PR #14455, opened 2023-12-20
[51766] Revert renaming of time fields

cbliard:
https://community.openproject.org/wp/51766

The names were renamed as part of epic 40867 "Progress reporting for work package hierarchies" and should have been shipped with 13.2. This epic was delayed to 13.3 and the renaming was already merged into `dev`. This PR reverts the renaming so that 13.2 can be shipped without any field name changes.

Also deleted `api_v3.errors.validation.estimated_hours` and other similar i18n keys. They got removed in dcd2a3b03c in favor of a more generic message more obscure for the end user.


lindenthal:
Hi @cbliard,

What is the problem with shipping the new names in 13.2?

cbliard:
After discussion, Niels decided that 13.2 can be shipped with the renaming, making this PR unneeded.

PR #14463, opened 2023-12-21
Evaluate changes on journal creation only for supported associations

ulferts:
Similar to how it is already done for writing the journal changes later on, evaluating for any changes on associations (e.g. customizable_values)  is only done for associations that are actually owned by the journaled object. For a work package, this excludes e.g. agenda_items, for meetings it will exclude e.g. customizable.

This fixes https://community.openproject.org/wp/51801 since the table not existing is the `meeting_agenda_item_journals` table and this will no longer be called as work packages don't have the association. Work packages are the only type of containers that should exist at this point.

### DIFF
--- a/app/services/journals/create_service.rb
+++ b/app/services/journals/create_service.rb
@@ -229,6 +229,14 @@ module Journals
     def relation_modifications_sql(predecessor)
       relations = []
 
+      for_supported_associations do |association|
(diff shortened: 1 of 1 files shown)


machisuji (approved):
👍 Thanks for the fix!

PR #14472, opened 2023-12-26
[#49527] Allow Shared with users to become assignees

aaron-contreras:
See: https://community.openproject.org/work_packages/49527


spec/support/api/v3/shared_available_principals_examples.rb
```diff
@@ -104,6 +114,11 @@
 
     before { get href }
 
-    it_behaves_like 'unauthorized access'
+    if work_package_scope
+      it_behaves_like 'not found',
+                      I18n.t('api_v3.errors.not_found.work_package')
```
aaron-contreras, line 121: Default behavior for work-package-scoped endpoints is not to reveal the presence of restricted work packages

lib/api/v3/work_packages/available_assignees_api.rb
```diff
@@ -0,0 +1,50 @@
...
+
+require 'api/v3/users/user_collection_representer'
+
+module API
+  module V3
+    module WorkPackages
+      class AvailableAssigneesAPI < ::API::OpenProjectAPI
+        resource :available_assignees do
+          after_validation do
+            authorize_in_work_package(:view_work_packages, work_package: @work_package)
```
ulferts: Why is the permission asked for here `:view_work_packages`? I realise that it is the same permission queried for by the `Projects::AvailableAssigneesAPI` as well. But a user being allowed to see the work packages does not need to see every assignable user. Only users having the `:edit_work_packages` permission need to see it here. For the `Projects::AvailableAssigneesAPI` endpoint, the permission should be `:add_work_packages` by now I think as that endpoint will only be used when creating a new work package now.

docs/api/apiv3/paths/work_package_available_assignees.yml
```diff
@@ -0,0 +1,114 @@
...
+      content:
+        application/hal+json:
+          schema:
+            $ref: "../components/schemas/error_response.yml"
+          examples:
+            response:
+              value:
+                _type: Error
+                errorIdentifier: urn:openproject-org:api:v3:errors:NotFound
+                message: The specified work package does not exist.
```
ulferts: The message actually seems to be ` The work package you are looking for cannot be found or has been deleted.`.

docs/api/apiv3/paths/work_package_available_assignees.yml
```diff
@@ -0,0 +1,114 @@
...
+                errorIdentifier: urn:openproject-org:api:v3:errors:MissingPermission
+                message: You are not allowed to see the assignable users for this
+                  work package.
+      description: |-
+        Returned if the client does not have sufficient permissions.
+
+        **Required permission:** view work packages
+
+        *Note that you will only receive this error, if you are at least allowed to see the corresponding work package.*
+      headers: {}
```
ulferts, line 93: With the way the API is currently structured, this case cannot happen. If the user lacks the `view work packages` permission, they will not see the work package (`/api/v3/work_packages/:id`) which is the precondition to reaching its assignees (`/api/v3/work_packages/:id/available_assignees`).

This might change if the permission is changed to `edit_work_packages`

ulferts (changes requested):
The change is looking good @aaron-contreras with the exception of a permission check which I think can be tightened.

lib/api/v3/work_packages/available_assignees_api.rb
aaron-contreras: You're right. Nice catch!

aaron-contreras:
Feedback addressed. The specs should read a bit better now as well :) @ulferts

klaustopher (approved):
lgtm

PR #14487, opened 2024-01-04
add save option to project list menu

ulferts:
Adds the option to save project queries (lists) for the user themselves. 

### Requirements

* [x] Lists can be persisted under a name chosen by the user by clicking on the "Save" entry in the "More" menu
    * [x] The filters are persisted.
    * Columns and sort order cannot be changed/persisted
    * [x]  For now, a modal is displayed in which the user inserts the selected name. This is to be changed later.
*  [x] Lists can be deleted via an entry in the &quot;more&quot; menu
    *  [x] Before deletion, a confirmation dialog (possibly a browser based one) needs to be confirmed.
    *  [x] Upon deletion of a user having created lists, those lists are deleted automatically
* [x]  Every logged in user can persist lists. By default, those lists will only be visible to the user creating the list.
*  [x] Persisted lists show up in the sidemenu from where they can be opened by clicking on them
    *  [x] Only visible to the user having created the view (Sharing/Public implemented in #51779)
    *  [x] Sort list by name with the static ones on top
    *  [x] Persisted lists are grouped in the sidebar to differentiate them from the static lists.
    *  [x] Whenever a list is opened (user persisted or static), the page header displays the name of the opened list
        *  [x] The static filter &quot;All projects&quot; is the default filter. Whenever the project module is opened, this list is displayed, the sidebar menu is highlighted and the header &quot;All projects&quot; displayed.
* [x]  Besides the user persisted lists the sidebar has a fixed set of static lists:
    *  [x] &quot;Active projects&quot;: All projects that are active. This is the default list for the index page
    *  [x] &quot;My projects&quot;: All active projects in which the current user is member
    *  [x] &quot;Archived projects&quot;: All projects not being active
    *  [x] By status (there are more project statuses but those are not listed):
        *  [x] &quot;On track&quot;: All active projects having the status &quot;On track&quot;
        *  [x] &quot;Off track&quot;: All active projects having the status &quot;Off track&quot;
        *  [x] &quot;At risk&quot;: All active projects having the status &quot;At risk&quot;
    *  [x] The previously existing &quot;Public projects&quot; is to be removed.

* [x] The sidebar only shows group labels if there are items in the group (changed for all submenus) 

### Technical ToDos

* [x] Add query controller that saves
  * [x] Add Project Query Create service
  * [x] Add Project Query Create contract
* [x] Display created Project queries in sidebar
* [x] Delete filters
  * [x] Add Project Query Delete service
  * [x] Add Project Query Delete contract
* [x] Find general structure for query handling 

### Architectural comments

The PR, while also implementing the feature listed above, proposes a structure for all pages that support displaying (e.g. in a table) a collection of models after having chosen the filters and sort order to apply. The pages than could or could not allow for persisting the query. In the long run, other modes of displaying the results like graphs could be added building on the same pattern.

#### Overview

The one view layer consists of multiple components:
* A header
* The generalized filters section
* The table 

The controller layer consists of currently three controllers:
* The one for fetching the query results, in this case the `ProjectsController`
* The one for displaying the sidemenu, in this case the `Projects::MenusController`
* The one for the CRUD actions on the persisted queries, in this case the `Projects::Queries::Controller`

The service layer has the usual suspects of a service for each action of the query (`Create` and `Delete` for now but `Update` will be added once the renaming and saving functionality is added as well). The contracts, apart from the already know contracts also receive a `LoadingContract`. That one is less strict than the contracts for saving. E.g. the name is not marked to necessarily be present.

The model layer consists of the `Queries::Projects::ProjectQuery` that was in place before but is now changed to become persistable.

For state management, the approach to keep the state visible in the url has been chosen. To that end, instead of relying on passing on the params and permitting them whenever a link is rendered in the backend, a Stimulus controller is added that can be called on the view layer to instruct the addition of hidden fields for forms or to append query props to link elements.

#### In depth

In order to administrate the `Queries::Projects::ProjectQuery` objects that by this PR can be persisted a `Projects::QueriesController` is introduced. That controller is responsible for creating and deleting `ProjectQuery` objects. It does so using Services and Contracts so from that controller downwards, the architecture is similar to what by now has been established as a pattern in the application. So far, differentiating between the `ProjectsController` and the `Projects::QueryController` has been a very clean one. They reuse the same functionality where possible and differentiate where necessary.

The `Queries::Projects::ProjectQuery` is turned into a persistable model, thus inheriting from `ApplicationRecord`. Since the query before inherited from `BaseQuery`, the later needed to be turned into a module so that all it's features can still be included into the `ProjectQuery`. So that change had to be applied to all the other `Queries::[Model]::[Model]Query` objects. Since those are not persistable (yet), they also receive the `Queries::UnpersistedQuery` module which adds attributes like `filters`. In the case of `Queries::Projects::ProjectQuery` that is not necessary as the attributes are generated by AR on the basis of DB columns.

Both, the filters as well as the orders (sort order) of the `Queries::Projects::ProjectQuery` are serialized to JSON columns. This is different from the serialization happening for the `Query` where YAML is used. The advantage of the more structured JSON should come in handy in the reoccurring cases where a migration needs to alter filter/order properties e.g. the name of a filter. The same will be used for columns even though that is not yet the case. The orders are not yet sent from the frontend, so for now, every project query will be stored with the order of `lft asc`. This is the default sort order ingrained into the current index page. 

Admittedly, all of this deepens the difference between `Query` and the rest of the query objects. There is a ticket open to fix this associated to the same Epic as the referenced work package.

The static queries of the sidemenu now receive a constant `query_id` value, e.g. `active` and `my`. Doing so hides the implementation details in the newly created `Queries::Projects::Factory` which is called by the query loading module referenced by both the `ProjectsController` as well as the `. Only a reference is thus known in various places of the application instead of the full filter values. Adding or adapting static queries should be quite simple.  

The filters for the Projects::ProjectQuery are no longer sent as JSON. Instead, the client can sent a simpler string following the pattern `...?filters=active = t`. This should make it simpler to understand the filter structure than with the JSON based approach. 

The `ParamsFromQueryController` Stimulus controller has the purpose of making the state of the page, which is kept in the url, available to all the components that together form the view. This is very light weight as the params don't have to be passed around into every component and the params also do not have to be permitted again and again. It remains to be seen, though, whether keeping the state in the url will satisfy all the requirements as it e.g. leads to the params which might contain sensitive information being present in the logs. On the other hand, such sensitive information could also be excluded from the logs. Alternative approaches like backend side session management (`session[:abc] = ...`) however would have other drawbacks. 

The table component now also receives the `query` object. This has the benefit of having a lot of information made available to the component allowing it to be rendered from both the `ProjectsController` as well as the `Projects::QueriesController` without additional instance variable loading. But it definitely leads to loading of data from the view layer.

The menu will now only show categories for which subitems are present. E.g. as long as the user has not persisted a query the "My private project lists" section header will not be displayed.

The PR also contains a rewrite of the project ancestor filter. That filter changed the SQL it generated and the tests were checking for the generated SQL. That has been changed to check for the results.

#### Open steps

A whole bunch of next steps are already described by the Epic this feature belongs to as well as the Epic's other children. Among those are e.g. the support for order and columns so this is not further discussed here.

The deep nesting of `Queries::Projects::ProjectQuery` becomes cumbersome. It seems desirable to have the model specific queries to instead follow the naming schema: `[Model pluralized]::Query` so e.g. `Projects::Query`. The filters, sort by and later on columns would follow suit. Without this, the nesting sometimes becomes four deep which is way to much. It also allows for a stronger cohesion of the models involved. This needs to be carried out for all queries in one go ideally, as the current structure is referenced in multiple parts of the code e.g. the `ParamsToQueryService`. Moving the code will also be safer ones more pages are using the same structure as the structure by than has proven to be stable.

That service has not been used for parsing the query props (i.e. filters). It became too strongly coupled to the API v3 containing e.g. the naming conversion of the v3. It is also bound to the JSON based serialization.

The `Projects::IndexPageHeaderComponent` reuses the `Primer::OpenProject::PageHeader`. That component does not yet support an edit mode which would be helpful in a whole number of scenarios even those outside of the index pages. E.g. the renaming functionality for Meetings could employ such an ability. The component currently is also limited to only have one `with_actions` area. This forces the developer to either wrap the actions in an arbitrary flex box or to use template outlets ( `<%=`) when we don't want to do that. The component needs to be changed so that multiple `with_action` blocks can be defined. That way, the margin could also be ingrained into the base component.

----

https://community.openproject.org/wp/51666


app/helpers/sort_helper.rb
```diff
@@ -284,7 +284,7 @@ def sort_link(column, caption, default_order, html_options = {})
     sort_options = { sort_key => sort_param }
 
     # Don't lose other params.
-    link_to_content_update(h(caption), safe_query_params(%w{filters page per_page expand}).merge(sort_options), html_options)
```
ulferts, line 287: We explicitly want to loose the `page` here. Whenever the sorting changes, the first page should be displayed. The user just changed the sort order so staying on e.g. the 3rd page does not make sense.

oliverguenther:
- [ ] Adding a filter (at least Name and Identifier, project status) and clicking apply results in an error
Bildschirmfoto 2024-01-30 um 13 12 49

oliverguenther (changes requested):
The feature is working nicely, and I only found this issue:

- [x] The export dialog always exports the default query, not the current saved one. This is not the case in current production (but may be broken in dev already)

ulferts:
@oliverguenther , the export feature broke in this PR. It was broken for persisted queries only but now will work also for them.

oliverguenther (approved):

PR #14493, opened 2024-01-05
Bump `view_component` to 3.9.0

aaron-contreras:
* Run `bundle update view_component`

In preparation for the Ruby 3.3 update, view_component 3.9.0 adds support for it.


aaron-contreras:
There are version conflicts with `openproject-primer_view_components` which currently locks `view_component` at 3.7.0.

PR #14498, opened 2024-01-08
Attempt to fix flickering board navigation spec

ulferts:


ulferts:
The spec succeeded five times now.

PR #14507, opened 2024-01-08
[50960] Derive estimated time from summing self and descendants

cbliard:
https://community.openproject.org/wp/50960

### DIFF
--- a/modules/backlogs/spec/contracts/work_packages/base_contract_spec.rb
+++ b/modules/backlogs/spec/contracts/work_packages/base_contract_spec.rb
@@ -163,65 +163,6 @@ RSpec.describe WorkPackages::BaseContract, type: :model do
   end
 
   describe 'version being restricted' do
-    shared_examples_for 'is invalid and notes the error' do
-      it 'is invalid and notes the error' do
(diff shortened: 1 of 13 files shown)


modules/backlogs/spec/contracts/work_packages/base_contract_spec.rb
```diff
@@ -289,15 +220,15 @@
...
+        include_examples 'is valid'
       end
 
       describe 'WITH a version' do
         before do
           work_package.version = version1
         end
 
-        it_behaves_like 'is valid'
+        include_examples 'is valid'
```
aaron-contreras, line 231: 👍 The output reads better this way

aaron-contreras (approved):
Compliments on writing great specs. Made it simple to review!

modules/backlogs/spec/contracts/work_packages/base_contract_spec.rb
cbliard: ~indeed, but I did not touch this part. I'll probably change it in a future commit.~

Sorry I misread, yes I think so too.

PR #14512, opened 2024-01-09
Project visible improvement

klaustopher:
Implements https://community.openproject.org/projects/openproject/work_packages/51706

### DIFF
--- a/app/models/projects/scopes/visible.rb
+++ b/app/models/projects/scopes/visible.rb
@@ -39,9 +39,9 @@ module Projects::Scopes
         # Use a shortcut for admins and anonymous where
         # we don't need to calculate for work package roles which is more expensive
         if user.admin? || user.anonymous?
-          Project.allowed_to(user, :view_project)
+          allowed_to(user, :view_project)
         else
-          Project.where(id: user.members.select(:project_id))
+          active.public_projects.or(active.where(id: user.members.select(:project_id)))
         end
       end
     end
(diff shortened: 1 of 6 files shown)


app/models/projects/scopes/visible.rb
```diff
@@ -39,10 +39,9 @@ def visible(user = User.current)
         # Use a shortcut for admins and anonymous where
         # we don't need to calculate for work package roles which is more expensive
         if user.admin? || user.anonymous?
-          Project.allowed_to(user, :view_project)
+          allowed_to(user, :view_project)
```
ulferts, line 42: As the code below already leaves the allowed_to scope behind, the same could be applied here:

```
if user.admin?
  active
elsif user.anonymous?
  active.public_projects
else
  active.public_projects.or(active.where(id: user.members.select(:project_id)))
end
```

Apart from applying the same abstraction throughout the method, I don't think it makes a difference, though.

ulferts (approved):
On the limited amount of data I had at hand the difference in performance the improved query offered was noticeable. I did not manage to find a query for a user that behaved particularly bad in `dev` so I am not able to attest to the overall improvements. 

But given that the solution turns the query way simpler I have to believe that it will be more performant than before.

app/models/projects/scopes/visible.rb
klaustopher: I was thinkin about the same thing. I sort of like your approach :) But now it's too late as you already merged it :D

PR #14515, opened 2024-01-09
[#52039] File link status displays not found files

Kharonus:
[#52039](https://community.openproject.org/wp/52039) [#52038](https://community.openproject.org/wp/52038)

### Wat?

- replaced "origin_permission" with "origin_status" to cover not found and general error
- updated API URN for all file link status
- changed link "permission" to "status" for file link api model
- updated file link list
  - floating actions are constructed in component
  - all statusses are now depicted in UI
- fixed disabled file link items, no longer clickable

### DIFF
--- a/modules/storages/spec/features/create_file_links_spec.rb
+++ b/modules/storages/spec/features/create_file_links_spec.rb
@@ -38,7 +38,7 @@ RSpec.describe 'Managing file links in work package', :js, :webmock do
   let(:work_package) { create(:work_package, project:, description: 'Initial description') }
 
   let(:oauth_application) { create(:oauth_application) }
-  let(:storage) { create(:nextcloud_storage, oauth_application:) }
+  let(:storage) { create(:nextcloud_storage, name: 'My Storage', oauth_application:) }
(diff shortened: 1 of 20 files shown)


modules/storages/spec/features/create_file_links_spec.rb
```diff
@@ -50,7 +50,7 @@
   let(:sync_service) do
     sync_service = instance_double(Storages::FileLinkSyncService)
     allow(sync_service).to receive(:call) do |file_links|
-      ServiceResult.success(result: file_links.each { |file_link| file_link.origin_permission = :view })
+      ServiceResult.success(result: file_links.each { |file_link| file_link.origin_status = :view_allowed })
```
apfohl, line 53: Wow, didn't know `.each` is returning the array. Feels weird. I mean yeah, for changing the object this is a good method.

apfohl (approved):

modules/storages/spec/features/create_file_links_spec.rb
Kharonus: ruby's array methods do often differ from what you know from other languages.

PR #14525, opened 2024-01-10
Epic/49688 custom set of project attributes grouped in sections

jjabari-op:
https://community.openproject.org/projects/stadt-koeln/work_packages/49688

Includes:
 - https://community.openproject.org/wp/53007


.github/workflows/pullpreview.yml
```diff
@@ -45,7 +45,7 @@ jobs:
       with:
         admins: crohr,HDinger,machisuji,oliverguenther,ulferts,wielinde,cbliard
         compose_files: docker-compose.pullpreview.yml
-        instance_type: large
+        instance_type: xlarge
```
dombesz, line 48: Is this still required?

spec/models/custom_value_spec.rb
```diff
@@ -150,7 +150,7 @@
     end
 
     context "for a string custom field without default value" do
-      shared_let(:custom_field) { create(:project_custom_field, :string) }
+      shared_let(:custom_field) { create(:project_custom_field, :string, projects: [project]) }
```
dombesz, line 153: I know this is not strictly related to the PR, but `shared_let` is supposed to have 1 definition in a spec file and it wasn't designed to be redefined, because the variable should stay the same during running the spec file. Redefining it this way, may also cause unexpected behaviour.

db/migrate/20231031133334_create_project_custom_field_project_mappings.rb
```diff
@@ -0,0 +1,34 @@
+class CreateProjectCustomFieldProjectMappings < ActiveRecord::Migration[7.0]
+  def up
+    create_table :project_custom_field_project_mappings do |t|
```
dombesz, line 3: I think the name `project_custom_fields_projects` name would follow the rails convention of naming tables better.

app/models/project_custom_field_project_mapping.rb
```diff
@@ -0,0 +1,45 @@
...
+# See COPYRIGHT and LICENSE files for more details.
+#++
+
+class ProjectCustomFieldProjectMapping < ApplicationRecord
+  belongs_to :project
+  belongs_to :project_custom_field, class_name: 'ProjectCustomField', foreign_key: 'custom_field_id',
+                                    inverse_of: :project_custom_field_project_mappings
+
+  # # Additionally to the database-level unique constraint, the application-level validation ensures that a
+  # # custom_field is associated with only one section
```
dombesz: At the moment, it is possible to save duplicate associations. Do we still need the uniqueness validator?

app/contracts/project_custom_field_project_mappings/base_contract.rb
```diff
@@ -0,0 +1,51 @@
...
+
+module ProjectCustomFieldProjectMappings
+  class BaseContract < ::ModelContract
+    attribute :project_id
+    attribute :custom_field_id
+
+    validate :validate_has_select_project_custom_fields_permission
+    validate :validate_is_not_required
+
+    def validate_has_select_project_custom_fields_permission
```
dombesz: I think having `validate_` in the method name is redundant. Wdyt think of the following name?
```suggestion
    def user_allowed_to_select_project_custom_fields
```

app/contracts/project_custom_field_project_mappings/base_contract.rb
```diff
@@ -0,0 +1,51 @@
...
+      return if user.allowed_in_project?(:select_project_custom_fields, model.project)
+
+      errors.add :base, :error_unauthorized
+    end
+
+    def validate_is_not_required
```
dombesz:
```suggestion
    def is_not_required
```

app/models/projects/acts_as_customizable_patches.rb
```diff
@@ -0,0 +1,248 @@
...
+        .where(custom_field_section_id: section.id)
+    end
+
+    def sorted_available_custom_fields
+      available_custom_fields
+        .sort_by { |pcf| [pcf.project_custom_field_section.position, pcf.position_in_custom_field_section] }
+    end
+
+    def sorted_available_custom_fields_by_section(section)
+      available_custom_fields_by_section(section)
```
dombesz: This is the only place we use the `available_custom_fields_by_section`. Both methods are just 1 liners, I would rather merge them together.
```suggestion
      available_custom_fields
        .where(custom_field_section_id: section.id)
```

app/models/projects/acts_as_customizable_patches.rb
```diff
@@ -0,0 +1,248 @@
...
+    before_create :reject_section_scoped_validation_for_creation
+    after_create :disable_custom_fields_with_empty_values
+
+    def build_missing_project_custom_field_project_mappings
+      # activate custom fields for this project (via mapping table) if values have been provided for custom_fields but no mapping exists
+      # current shortcommings:
+      # - boolean custom fields are always activated as a nil value is never provided (always true/false)
+      custom_field_ids = project.custom_values
+        .reject { |cv| cv.value.blank? }
+        .reject { |cv| cv.persisted? } # do not reactivate custom fields which have already been used and disabled
```
dombesz: These two could be made in the same interation:
```suggestion
        .reject { |cv| cv.persisted? || cv.value.blank? } # do not reactivate custom fields which have already been used and disabled
```

app/models/projects/acts_as_customizable_patches.rb
```diff
@@ -0,0 +1,248 @@
...
+      # but:
+      # this hook is required as acts_as_customizable build custom values with their default value even if a blank value was provided in the project creation form
+      # `build_missing_project_custom_field_project_mappings` will then activate the custom field although the user explicitly provided a blank value
+      # in order to not patch `acts_as_customizable` further, we simply identify these custom values and deactivate the custom field
+      custom_field_ids = project.custom_values.select { |cv| cv.value.blank? && !cv.required? }.pluck(:custom_field_id)
+
+      project_custom_field_project_mappings.where(custom_field_id: custom_field_ids).destroy_all
+    end
+
+    def query_available_custom_fields_on_global_level
```
dombesz: I would rename this method to `def with_all_available_custom_fields`.

app/components/settings/project_custom_fields/new_form_header_component.rb
```diff
@@ -0,0 +1,40 @@
...
+# See COPYRIGHT and LICENSE files for more details.
+#++
+
+module Settings
+  module ProjectCustomFields
+    class NewFormHeaderComponent < ApplicationComponent
+      include ApplicationHelper
+      include OpPrimer::ComponentHelpers
+
+      def initialize
```
dombesz: This seems to be unused.

app/models/projects/acts_as_customizable_patches.rb
```diff
@@ -0,0 +1,248 @@
...
+
+    def custom_field_values=(values)
+      # overrides acts_as_customizable
+      # we need to query the available custom fields on a global level when updating custom field values
+      # in order to support implicit activation of custom fields when values are provided during an update
+      self._query_available_custom_fields_on_global_level = true # set to false in after_save hook
+      set_custom_field_values_method_from_acts_as_customizable_module(values)
+    end
+
+    # we cannot call super as the code in acts_as_customizable is shipped in a module
```
dombesz: We can remove this method.

app/models/projects/acts_as_customizable_patches.rb
```diff
@@ -0,0 +1,248 @@
...
+      end
+    end
+
+    def custom_field_values=(values)
+      # overrides acts_as_customizable
+      # we need to query the available custom fields on a global level when updating custom field values
+      # in order to support implicit activation of custom fields when values are provided during an update
+      self._query_available_custom_fields_on_global_level = true # set to false in after_save hook
+      set_custom_field_values_method_from_acts_as_customizable_module(values)
+    end
```
dombesz: As I tested it, `super` seems to work, combining it with the renamed `with_all_available_custom_fields` this method could be simplified:

```suggestion
    # we need to query the available custom fields on a global level when updating custom field values
    # in order to support implicit activation of custom fields when values are provided during an update
    def custom_field_values=(values)
      with_all_available_custom_fields { super }
    end
```

app/models/projects/acts_as_customizable_patches.rb
```diff
@@ -0,0 +1,248 @@
...
+    # trying to set a custom field which is not enabled via e.g. custom_field_123="foo"
+    def for_custom_field_accessor(method_symbol)
+      match = /\Acustom_field_(?<id>\d+)=?\z/.match(method_symbol.to_s)
+      if match
+        custom_field = available_custom_fields(global: true).find { |cf| cf.id.to_s == match[:id] }
+        if custom_field
+          yield custom_field
+        end
+      end
+    end
```
dombesz: This method also can be simplified by using the `with_all_available_custom_fields` and calling `super`.

```suggestion
    def for_custom_field_accessor(method_symbol)
      with_all_available_custom_fields { super }
    end
```

app/models/projects/acts_as_customizable_patches.rb
```diff
@@ -0,0 +1,248 @@
...
+        .each { |custom_value| add_custom_value_errors! custom_value }
+    end
+
+    def of_specified_custom_field_section?(custom_value)
+      if _limit_custom_fields_validation_to_section_id.present?
+        custom_field_section_ids[custom_value.custom_field_id] == _limit_custom_fields_validation_to_section_id
+      else
+        true # validate all custom values if no specific section was specified
+      end
+    end
```
dombesz: I think these 3 methods can be simplified:

```suggestion
    def validate_custom_values
      # validate custom values only of a specified section
      # instead of validating ALL custom values like done in acts_as_customizable
      custom_field_section_ids = CustomField
        .where(id: custom_field_values.pluck(:custom_field_id))
        .where(custom_field_section_id: _limit_custom_fields_validation_to_section_id)
        .pluck(:id)

      super(custom_field_section_ids)
    end
```

Then we can define a new optional parameter for the `validate_custom_values` in the `lib_static/plugins/acts_as_customizable/lib/acts_as_customizable.rb`

```ruby
def validate_custom_values(custom_field_ids=[])
  set_default_values! if new_record?

   custom_field_values
    .reject(&:marked_for_destruction?)
    .select { |cv| custom_field_ids.empty? || custom_field_ids.include?(cv.custom_field_id) }
    .select(&:invalid?)
    .each { |custom_value| add_custom_value_errors! custom_value }
end
```
This way we can provide a set of custom_field_ids that needs to be validated, instead of validating all the custom values.
Additionally, the `custom_field_section_ids` did not have any caching, so it was doing N+1 queries. With my suggestion the N+1 queries are also eliminated.

app/models/projects/acts_as_customizable_patches.rb
```diff
@@ -0,0 +1,248 @@
...
+      # which is not nice but helps us to touch acts_as_customizable as little as possible
+      #
+      # additionally we provide the `global` parameter to allow querying the available custom fields on a global level
+      # when we have explicit control over the call of `available_custom_fields`
+      unless global || _query_available_custom_fields_on_global_level
+        custom_fields = custom_fields.where(id: active_custom_field_ids_of_project)
+      end
+
+      custom_fields
+    end
```
dombesz:
```suggestion
    def available_custom_fields(global: false)
      # overrides acts_as_customizable
      # in contrast to acts_as_customizable, custom_fields are enabled per project
      # thus we need to check the project_custom_field_project_mappings
      custom_fields = ProjectCustomField
        .includes(:project_custom_field_section)

      # available_custom_fields is called from within the acts_as_customizable module
      # we don't want to adjust these calls, but need a way to query the available custom fields on a global level in some cases
      # thus we pass in this parameter as an instance flag implicitly here,
      # which is not nice but helps us to touch acts_as_customizable as little as possible
      #
      # additionally we provide the `global` parameter to allow querying the available custom fields on a global level
      # when we have explicit control over the call of `available_custom_fields`
      unless global || new_record? || _query_available_custom_fields_on_global_level
        custom_fields = custom_fields
          .where(id: project_custom_field_project_mappings.select(:custom_field_id))
          .or(ProjectCustomField.required)
      end

      custom_fields
    end
```

app/models/projects/acts_as_customizable_patches.rb
```diff
@@ -0,0 +1,248 @@
...
+      self._query_available_custom_fields_on_global_level = true
+    end
+
+    def set_query_available_custom_fields_to_project_level
+      # reset the query_available_custom_fields_on_global_level after saving
+      # in order not to silently carry this setting in this instance
+      self._query_available_custom_fields_on_global_level = nil
+    end
+
+    def reject_section_scoped_validation_for_creation
```
dombesz, line 86: Perhaps the `Projects::CreateContract` would be a better place for this validation, because this error seems to be tied to the representation of the project custom fields, but not their internal workings.

app/services/project_custom_field_project_mappings/bulk_edit_service.rb
```diff
@@ -0,0 +1,103 @@
...
+#
+# You should have received a copy of the GNU General Public License
+# along with this program; if not, write to the Free Software
+# Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+#
+# See COPYRIGHT and LICENSE files for more details.
+#++
+
+module ProjectCustomFieldProjectMappings
+  class BulkEditService < ::BaseServices::BaseCallable
```
dombesz: To follow the convention of services this should rather be called `BulkUpdateService`

app/controllers/projects/settings/project_custom_fields_controller.rb
```diff
@@ -0,0 +1,118 @@
...
+
+  def enable_all_of_section
+    call = bulk_edit_service.call(action: :enable)
+
+    if call.success?
+      eager_load_project_custom_field_project_mappings # reload mappings
+
+      update_sections_via_turbo_stream # update all sections in order not to mess with stimulus target references
+    else
+      # TODO: handle error
```
dombesz, line 62: There are some todo comments left.

app/controllers/projects/settings/project_custom_fields_controller.rb
```diff
@@ -0,0 +1,118 @@
...
+class Projects::Settings::ProjectCustomFieldsController < Projects::SettingsController
+  include OpTurbo::ComponentStream
+  include Projects::Settings::ProjectCustomFields::ComponentStreams
+
+  menu_item :settings_project_custom_fields
+
+  before_action :eager_load_project_custom_field_sections, only: %i[show toggle enable_all_of_section disable_all_of_section]
+  before_action :eager_load_project_custom_fields, only: %i[show toggle enable_all_of_section disable_all_of_section]
+  before_action :eager_load_project_custom_field_project_mappings,
+                only: %i[show toggle enable_all_of_section disable_all_of_section]
```
dombesz: It seems like we could move the 3 before actions into one.

app/controllers/projects/settings/project_custom_fields_controller.rb
```diff
@@ -0,0 +1,118 @@
...
+      # TODO: handle error
+    end
+
+    respond_with_turbo_streams
+  end
+
+  private
+
+  def eager_load_project_custom_field_sections
+    @project_custom_field_sections = ProjectCustomFieldSection.all.to_a
```
dombesz: Is there a reason we need to explicitly call `to_a`? ActiveRecord seems to be loading a query only once.

app/controllers/projects/settings/project_custom_fields_controller.rb
```diff
@@ -0,0 +1,118 @@
...
+    else
+      render json: {}, status: :unprocessable_entity
+    end
+  end
+
+  def enable_all_of_section
+    call = bulk_edit_service.call(action: :enable)
+
+    if call.success?
+      eager_load_project_custom_field_project_mappings # reload mappings
```
dombesz: Since we can merge the before actions together, we won't have this method anymore. However we could still reload the results by calling:
```suggestion
      @project_custom_field_project_mappings.reload # reload mappings
```

app/forms/custom_fields/inputs/base/autocomplete/multi_value_input.rb
```diff
@@ -0,0 +1,62 @@
...
+  def autocomplete_options
+    {
+      multiple: true,
+      decorated: decorated?,
+      append_to:
+    }
+  end
+
+  def decorated?
+    raise NotImplementedError
```
dombesz, line 48: Is this intentional?

app/forms/custom_fields/inputs/base/autocomplete/single_value_input.rb
```diff
@@ -0,0 +1,50 @@
...
+  def autocomplete_options
+    {
+      multiple: false,
+      decorated: decorated?,
+      append_to:
+    }
+  end
+
+  def decorated?
+    raise NotImplementedError
```
dombesz, line 48: Is this intentional?

app/models/projects/acts_as_customizable_patches.rb
```diff
@@ -0,0 +1,248 @@
...
+      end
+    end
+
+    # overrides acts_as_customizable
+    # we need to query the available custom fields on a global level when
+    # trying to set a custom field which is not enabled via e.g. custom_field_123="foo"
+    def for_custom_field_accessor(method_symbol)
+      match = /\Acustom_field_(?<id>\d+)=?\z/.match(method_symbol.to_s)
+      if match
+        custom_field = available_custom_fields(global: true).find { |cf| cf.id.to_s == match[:id] }
```
dombesz: I get an error related to this line on the browser console while loading the overview page.

image

config/initializers/feature_decisions.rb
```diff
@@ -38,3 +38,5 @@
 #   initializer 'the_engine.feature_decisions' do
 #     OpenProject::FeatureDecisions.add :some_flag
 #   end
+
+OpenProject::FeatureDecisions.add :project_attributes
```
dombesz: Is this in use? Could not find any reference for it.

app/forms/custom_fields/inputs/text.rb
```diff
@@ -0,0 +1,38 @@
...
+#
+# See COPYRIGHT and LICENSE files for more details.
+#++
+
+class CustomFields::Inputs::Text < CustomFields::Inputs::Base::Input
+  form do |custom_value_form|
+    # TODO: rich_text_area not working yet
+    # Uncaught DOMException: Failed to execute 'querySelector' on 'Element': '#project_project[new_custom_field_values_attributes][xyz][value]' is not a valid selector.
+    # --> rich_text_area is not using the configured id, which is not scoped to model via base_config
+    # --> ids with '[' ']' are not valid selectors
```
dombesz: Is this todo still valid? I don't get the error mentioned here while using the rich text input.

modules/overviews/app/controllers/overviews/overviews_controller.rb
```diff
@@ -1,14 +1,84 @@
...
+      update_via_turbo_stream(
+        component: ProjectCustomFields::SidebarComponent.new(
+          project: @project,
+          eager_loaded_project_custom_field_sections:
+        )
+      )
+    end
+
+    def eager_loaded_project_custom_field_sections
+      ProjectCustomFieldSection.all.to_a
```
dombesz: Calling `to_a` explicitly here too.

app/components/projects/settings/project_custom_field_sections/show_component.rb
```diff
@@ -0,0 +1,63 @@
...
+        end
+
+        private
+
+        def wrapper_uniq_by
+          @project_custom_field_section.id
+        end
+
+        def sorted_project_custom_fields
+          @project_custom_fields.sort_by { |pcf| pcf.position_in_custom_field_section }
```
dombesz: I'm wondering if this is necessary? We already seem to have a method that sorts the custom fields in the acts_as_customizable patch. (There are a few other instances too)

https://github.com/opf/openproject/blob/0ea5c3f36848a92c6311ffab6304702d8fbbc6e7/app/models/projects/acts_as_customizable_patches.rb#L166-L169

app/components/settings/project_custom_field_sections/show_component.rb
```diff
@@ -0,0 +1,127 @@
...
+      def draggable_item_config(project_custom_field)
+        {
+          'draggable-id': project_custom_field.id,
+          'draggable-type': 'custom-field',
+          'drop-url': drop_admin_settings_project_custom_field_path(project_custom_field)
+        }
+      end
+
+      def move_actions(menu)
+        # TODO: these methods trigger database queries for each section displayed
```
dombesz: Yes, this will trigger N+1 queries. I have solved a similar issue on the meeting agenda items in this PR #14190 . 
The solution is to pass down the first and last element information to the ShowComponent in order to avoid the N+1 queries.

app/components/settings/project_custom_field_sections/custom_field_row_component.rb
```diff
@@ -0,0 +1,104 @@
...
+      def edit_action_item(menu)
+        menu.with_item(label: t("label_edit"),
+                       href: edit_admin_settings_project_custom_field_path(@project_custom_field),
+                       data: { turbo: "false", qa_selector: "project-custom-field-edit" }) do |item|
+          item.with_leading_visual_icon(icon: :pencil)
+        end
+      end
+
+      def move_actions(menu)
+        # TODO: these methods trigger database queries for each custom field displayed
```
dombesz: Same here: #14190

app/components/settings/project_custom_fields/header_component.rb
```diff
@@ -0,0 +1,41 @@
...
+#++
+
+module Settings
+  module ProjectCustomFields
+    class HeaderComponent < ApplicationComponent
+      include ApplicationHelper
+      include OpPrimer::ComponentHelpers
+      include OpTurbo::Streamable
+
+      def initialize
```
dombesz: Is this necessary?

app/controllers/admin/settings/project_custom_field_sections_controller.rb
```diff
@@ -0,0 +1,102 @@
...
+
+      if @project_custom_field_section.destroy
+        update_sections_via_turbo_stream(project_custom_field_sections: ProjectCustomFieldSection.all)
+      end
+
+      respond_with_turbo_streams
+    end
+
+    def move
+      @project_custom_field_section.move_to = params[:move_to]&.to_sym
```
dombesz: It's not mission critical, but we should leave a todo here to handle errors in this controller.
Probably we also need to move these actions into services in the future.

app/controllers/admin/settings/project_custom_fields_controller.rb
```diff
@@ -0,0 +1,130 @@
...
+      @custom_field.move_to = params[:move_to]&.to_sym
+
+      update_sections_via_turbo_stream(project_custom_field_sections: @project_custom_field_sections)
+
+      respond_with_turbo_streams
+    end
+
+    def drop
+      # prototyopical implementation
+      # needs refactoring via update service
```
dombesz: I agree.

app/controllers/custom_fields_controller.rb
```diff
@@ -36,7 +37,12 @@ class CustomFieldsController < ApplicationController
 
   def index
     # loading wp cfs exclicity to allow for eager loading
-    @custom_fields_by_type = CustomField.all.where.not(type: 'WorkPackageCustomField').group_by { |f| f.class.name }
+    @custom_fields_by_type = CustomField.all
+      .where.not(type: 'WorkPackageCustomField')
+      # ProjectCustomFields now managed in a different UI
+      .tap { |query| query.where.not(type: 'ProjectCustomField') }
```
dombesz: Just a nitpick, type can also take an array of strings
```suggestion
      .where.not(type: ['WorkPackageCustomField', 'ProjectCustomField'])
```

dombesz (commented):
Massive job Jonas, congrats!👏
I have left a few comments, admittedly I didn't spent too much time on the UI elements, but they seem pretty straightforward. I tried to comment all my observations just for the record, so if you think some of them should be addressed separately, that is also fine with me.
I haven't go through the whole PR, still have to double check the migrations and check the specs.

modules/xls_export/spec/models/xls_export/project/exporter/xls_integration_spec.rb
```diff
@@ -33,7 +33,8 @@
 
   context 'with project description containing html' do
     before do
-      project.update(description: "This is an <p>html</p> description.")
+      project.description = "This is an <p>html</p> description."
+      project.save!(validate: false)
```
dombesz:
```suggestion
      project.update_column(:description, "This is an <p>html</p> description.")
```

modules/xls_export/spec/models/xls_export/project/exporter/xls_integration_spec.rb
```diff
@@ -55,27 +56,34 @@
...
         end
 
         expect(sheet.row(1))
-          .to eq [project.id.to_s, project.identifier, project.name, project.description, 'Off track', 'false', *custom_values]
+          .to eq [project.id.to_s, project.identifier, project.name, project.description, 'Off track', 'false',
+                  *custom_values]
+
+        # The column for the project-level-disabled custom field is blank
+        expect(sheet.row(1)[header.index(not_used_string_cf.name)]).to be_nil
+        # TODO: CSV export renders "" instead of nil, why does XLS export render nil?
```
dombesz: This comment seems to be contradictory.

spec/features/admin/project_custom_fields/create_spec.rb
```diff
@@ -0,0 +1,106 @@
...
+# See COPYRIGHT and LICENSE files for more details.
+#++
+
+require 'spec_helper'
+require_relative 'shared_context'
+
+RSpec.describe 'Create project custom fields', :js do
+  include_context 'with seeded project custom fields'
+
+  context 'with unsufficient permissions' do
```
dombesz: Just a typo:
```suggestion
  context 'with insufficient permissions' do
```

spec/features/admin/project_custom_fields/create_spec.rb
```diff
@@ -0,0 +1,106 @@
...
+      latest_custom_field = ProjectCustomField.reorder(created_at: :asc).last
+
+      expect(latest_custom_field.name).to eq('New custom field')
+      expect(latest_custom_field.project_custom_field_section).to eq(section_for_multi_select_fields)
+    end
+
+    it 'prevents creating a new project custom field with an empty name' do
+      click_on('Save')
+
+      # no server side validation shown, html5 validation is used
```
dombesz:
```suggestion
      expect(page).to have_field 'custom_field_name', validation_message: 'Please fill in this field.'
```

spec/features/admin/project_custom_fields/edit_spec.rb
```diff
@@ -0,0 +1,95 @@
...
+# See COPYRIGHT and LICENSE files for more details.
+#++
+
+require 'spec_helper'
+require_relative 'shared_context'
+
+RSpec.describe 'Edit project custom fields', :js do
+  include_context 'with seeded project custom fields'
+
+  context 'with unsufficient permissions' do
```
dombesz:
```suggestion
  context 'with insufficient permissions' do
```

spec/features/admin/project_custom_fields/edit_spec.rb
```diff
@@ -0,0 +1,95 @@
...
+      end
+    end
+
+    it 'prevents saving a project custom field with an empty name' do
+      original_name = boolean_project_custom_field.name
+
+      fill_in('custom_field_name', with: '')
+      click_on('Save')
+
+      # no server side validation shown, html5 validation is used
```
dombesz:
```suggestion
     expect(page).to have_field 'custom_field_name', validation_message: 'Please fill in this field.'
```

spec/features/admin/project_custom_fields/list_spec.rb
```diff
@@ -0,0 +1,276 @@
...
+                                            type: 'ProjectCustomField',
+                                            custom_field_section_id: section_for_multi_select_fields.id
+                                          ))
+      end
+    end
+  end
+
+  # helper methods:
+
+  def within_project_custom_field_section_container(section, &block)
```
dombesz, line 241: I think it would be useful to have a support class for the list, something like `Pages::ProjectCustomFieldListPage`. These classes are usually inherited from the `support/pages/page.rb` and they can hold all the logic that is required to navigate around the page.

spec/features/admin/project_custom_fields/list_spec.rb
```diff
@@ -0,0 +1,276 @@
...
+
+      containers = page.all('.op-project-custom-field-section-container')
+
+      expect(containers[0].text).to include('New section name')
+      expect(containers[1].text).to include(section_for_input_fields.name)
+      expect(containers[2].text).to include(section_for_select_fields.name)
+      expect(containers[3].text).to include(section_for_multi_select_fields.name)
+    end
+
+    describe 'managing project custom fields' do
```
dombesz, line 125: I would also add some tests for the "Move" action buttons.

spec/features/projects/copy_spec.rb
```diff
@@ -140,6 +152,209 @@
...
+            optional_project_custom_field_with_default.id,
+            invisible_field.id
+          )
+
+          expect(copied_project.custom_value_for(invisible_field).typed_value).to eq('foo')
+        end
+      end
+
+      context 'with non-admin user' do
+        # TBD: Not sure if this is the desired behavior, but would be a bit tricky to change
```
dombesz: From my point of view, this is the logical behaviour.

spec/features/projects/copy_spec.rb
```diff
@@ -140,6 +152,209 @@
...
+          project_custom_field.id,
+          optional_project_custom_field.id,
+          optional_project_custom_field_with_default.id
+        )
+
+        # the optional custom field is activated, but set to blank value
+        expect(copied_project.custom_value_for(optional_project_custom_field).typed_value).to eq('')
+      end
+
+      # TBD: Is this intended from a conceptial point of view?
```
dombesz: Best would be to raise this question to product.

spec/features/projects/create_spec.rb
```diff
@@ -133,5 +138,256 @@
...
+          )
+
+          expect(project.custom_value_for(invisible_field).typed_value).to eq('foo')
+        end
+      end
+
+      context 'with a non-admin user' do
+        current_user { create(:user, global_permissions: %i[add_project]) }
+
+        it 'does not show invisible fields in the form and thus not activates the invisible field' do
```
dombesz, line 377: If we would copy the hidden fields, it would possibly expose them to the non-admin user? (Given the non-admin user becomes an admin in the new project.)

spec/permissions/manage_project_custom_values_spec.rb
```diff
@@ -0,0 +1,50 @@
...
+require File.expand_path('../support/permission_specs', __dir__)
+
+RSpec.describe Overviews::OverviewsController, 'manage_project_custom_values permission',
+               type: :controller do
+  include PermissionSpecs
+
+  # render sidebar on project overview page with view_project permission
+  # TODO: prevents calling overviews/overviews#project_custom_fields_sidebar when not having the permission view_project (FAILED - 1)
+  it 'prevents calling overviews/overviews#project_custom_fields_sidebar when not having the permission view_project' do
+    pending 'spec failing, reason unknown'
```
dombesz, line 39: Just putting a comment for the pending spec.

spec/models/projects/exporter/csv_integration_spec.rb
```diff
@@ -60,29 +60,41 @@
 
   describe 'custom field columns selected' do
     let(:query_columns) do
-      %w[name description project_status public] + custom_fields.map(&:column_name)
+      %w[name description project_status public] + global_project_custom_fields.map(&:column_name)
     end
 
     context 'when ee enabled', with_ee: %i[custom_fields_in_projects_list] do
-      it 'renders all those columns' do
+      it 'renders all globally available project custom fields in the header' do
```
dombesz: Shall we test that hidden fields are excluded from the export?

spec/services/project_custom_field_project_mappings/toggle_service_spec.rb
```diff
@@ -0,0 +1,217 @@
...
+             member_with_permissions: {
+               project => %w[
+                 view_work_packages
+                 edit_project
+                 select_project_custom_fields
+               ]
+             })
+    end
+
+    it 'toggles visible, non-required fields' do
```
dombesz, line 125: These tests could be extracted out into shared_examples.

spec/services/project_custom_field_project_mappings/toggle_service_spec.rb
```diff
@@ -0,0 +1,217 @@
...
+             lastname: 'Editor',
+             member_with_permissions: {
+               project => %w[
+                 view_work_packages
+                 edit_project
+               ]
+             })
+    end
+
+    it 'toggles visible, non-required fields' do
```
dombesz:
```suggestion
    it 'does not toggle visible, non-required fields' do
```

.github/workflows/pullpreview.yml
jjabari-op: Good question! @ulferts increased the size, but I don't know if it was just meant for this pullpreview instance...

db/migrate/20231031133334_create_project_custom_field_project_mappings.rb
jjabari-op: True, but I do like the explicitness of `project_custom_field_project_mappings`. @oliverguenther any preference?

app/models/project_custom_field_project_mapping.rb
jjabari-op: I've added a proper validation!

app/models/projects/acts_as_customizable_patches.rb
jjabari-op: I do agree partially, but this is strongly bound to the patching code in this file what makes me want to have it right there as well. Additionally, creating a project through the create contract is technically just one way of doing it. Using the project active record API standalone would then not trigger this important "validation". If you're ok with it, I tend to prefer having it bound to the model directly.

app/controllers/projects/settings/project_custom_fields_controller.rb
jjabari-op: From my understanding it is required in order to force the data to be fetched from the database and to be loaded in memory. I'm doing this in order to avoid N+1 query in view components afterwards.

app/forms/custom_fields/inputs/base/autocomplete/multi_value_input.rb
jjabari-op: Yes as this is a base class

app/forms/custom_fields/inputs/base/autocomplete/single_value_input.rb
jjabari-op: Yes as this is a base class

config/initializers/feature_decisions.rb
jjabari-op: Good point! Forgot to remove this!

app/forms/custom_fields/inputs/text.rb
jjabari-op: Already removed :)

modules/overviews/app/controllers/overviews/overviews_controller.rb
jjabari-op: Same reason, to prevent N+1 queries in the view components

app/components/projects/settings/project_custom_field_sections/show_component.rb
jjabari-op: In this component, eager loaded data is accessed, thus it would not be ideal to go for the model method in this case

app/components/settings/project_custom_field_sections/show_component.rb
jjabari-op: Uh nice! Would like to implement here as well? :)

app/components/settings/project_custom_fields/header_component.rb
jjabari-op: Some part is already cleaned up, but the `Streamable` module is actually required in this case

spec/features/projects/project_custom_fields/overview_page/dialog/inputs_spec.rb
```diff
@@ -0,0 +1,649 @@
...
+          dialog.within_async_content(close_after_yield: true) do
+            field.expect_value(default_value)
+          end
+        end
+      end
+
+      describe 'with boolean CF' do
+        let(:custom_field) { boolean_project_custom_field }
+        let(:default_value) { false }
+        let(:expected_blank_value) { false }
```
dombesz: These values does not seem to be used in the `a custom field checkbox` shared examples.

spec/features/projects/project_custom_fields/overview_page/dialog/update_spec.rb
```diff
@@ -0,0 +1,656 @@
...
+          overview_page.within_custom_field_container(custom_field) do
+            expect(page).to have_no_text(expected_initial_value)
+          end
+        end
+      end
+
+      describe 'with boolean CF' do
+        let(:custom_field) { boolean_project_custom_field }
+        let(:field) { FormFields::Primerized::InputField.new(custom_field) }
+        let(:expected_initial_value) { true }
```
dombesz: This variable is not being used in the expectation.

spec/features/projects/project_custom_fields/overview_page/dialog/validation_spec.rb
```diff
@@ -0,0 +1,520 @@
...
+          custom_field.update!(regexp: '^[A-Z]+$')
+
+          overview_page.open_edit_dialog_for_section(section)
+
+          field.fill_in(with: 'foo')
+
+          dialog.submit
+
+          field.expect_error(I18n.t('activerecord.errors.messages.invalid'))
+        end
```
dombesz: Do you think these test are necessary? The validation functionality is already tested in the custom field unit tests. The only test we need is to ensure that the error messages are shown.

spec/features/projects/project_custom_fields/overview_page/dialog/validation_spec.rb
```diff
@@ -0,0 +1,520 @@
...
+      end
+
+      describe 'with integer CF' do
+        let(:custom_field) { integer_project_custom_field }
+        let(:field) { FormFields::Primerized::InputField.new(custom_field) }
+
+        it_behaves_like 'a custom field input'
+
+        it 'shows an error if the value is too long' do
```
dombesz: Same for these.

spec/features/projects/project_custom_fields/overview_page/sidebar_spec.rb
```diff
@@ -0,0 +1,860 @@
...
+
+      overview_page.visit_page
+
+      overview_page.within_async_loaded_sidebar do
+        expect(page).to have_no_text 'String field enabled for other project'
+      end
+    end
+  end
+
+  describe 'with correct order' do
```
dombesz: With adjusting the specs from the "with correct scoping" we can also ensure the order is correct, and we can remove these specs.

spec/features/projects/project_custom_fields/settings/mapping_spec.rb
```diff
@@ -0,0 +1,448 @@
...
+
+  def expect_unchecked_state
+    expect(page).to have_css('.ToggleSwitch-statusOff')
+  end
+
+  def within_custom_field_section_container(section, &)
+    within("[data-qa-selector='project-custom-field-section-#{section.id}']", &)
+  end
+
+  def within_custom_field_container(custom_field, &)
```
dombesz, line 445: Just as a future improvement we could create a `ProjectCustomFields` support view class.

frontend/src/stimulus/controllers/dynamic/projects/settings/project-custom-fields-mapping-filter.controller.ts
```diff
@@ -0,0 +1,95 @@
...
+        (item as HTMLElement).style.display = 'none';
+      }
+    });
+  }
+
+  resetFilterViaClearButton() {
+    this.showBulkActionContainers();
+
+    this.searchItemTargets.forEach((item) => {
+      (item as HTMLElement).style.display = 'block';
```
dombesz: I was wondering if this could cause problems later on, in case the `display` is going to be changed to another value in the styling. Perhaps it would be better to add and remove a `.hidden` class?

app/views/admin/settings/projects_settings/show.html.erb
```diff
@@ -60,6 +60,7 @@ See COPYRIGHT and LICENSE files for more details.
         </div>
       <% end %>
 
+
```
dombesz, line 63: Just an extra line here.

app/controllers/projects/settings/project_custom_fields_controller.rb
dombesz: I believe that will not prevent any N+1 queries, because when AR loads the results, it will not query the database each time the results are accessed.
However, we should definitely eager load the `custom_fields` association on the query, because it is referenced in the [project_custom_field_sections/show_component](https://github.com/opf/openproject/pull/14525/files#diff-ab7773fd1ebcdc5a97df0b4a20c6afe221b66a9be1c15d3ed51b0e2df5cfc0d5R40), which will indeed cause an N+1.

```suggestion
    @project_custom_field_sections = ProjectCustomFieldSection.includes(:custom_fields)
                                                         .order("custom_fields.position_in_custom_field_section ASC")
                                                         .all
```

Note: in order to not accidentally trigger the N+1 we should remove the `reorder(position_in_custom_field_section: :asc)` from [this line](https://github.com/opf/openproject/pull/14525/files#diff-ab7773fd1ebcdc5a97df0b4a20c6afe221b66a9be1c15d3ed51b0e2df5cfc0d5R40).

app/components/settings/project_custom_field_sections/show_component.rb
```diff
@@ -0,0 +1,127 @@
...
+    class ShowComponent < ApplicationComponent
+      include ApplicationHelper
+      include OpPrimer::ComponentHelpers
+      include OpTurbo::Streamable
+
+      def initialize(project_custom_field_section:)
+        super
+
+        @project_custom_field_section = project_custom_field_section
+        @project_custom_fields = project_custom_field_section.custom_fields.reorder(position_in_custom_field_section: :asc)
```
dombesz: The reorder can be removed, since it is being added to the original query, this way will not trigger an N+1 query.

app/services/projects/copy_service.rb
```diff
@@ -80,6 +80,42 @@ def before_perform(params, service_call)
...
+      # they would not be activated in the target project,
+      # which is desired in pure project creation context (-> as form fields are blank)
+      # but in the copy context we activate them here in order to have the same mapping as seen in the source project:
+      custom_fields_activated_in_source = source.project_custom_fields.pluck(:id)
+      custom_fields_activated_in_target = call.result.project_custom_fields.pluck(:id)
+
+      custom_fields_to_enable = custom_fields_activated_in_source - custom_fields_activated_in_target
+      call.result.project_custom_fields << ProjectCustomField.where(id: custom_fields_to_enable)
+    end
+
```
dombesz: This can be simplified, we need everything activated from the source project in the target project and nothing else.

```suggestion
    def after_perform(call)
      copy_activated_custom_fields(call)

      super
    end

    def copy_activated_custom_fields(call)
       call.result.project_custom_field_ids = source.project_custom_field_ids
    end
```

db/migrate/20240208100316_enable_required_project_custom_fields_in_all_projects.rb
```diff
@@ -0,0 +1,17 @@
...
+  def up
+    required_project_custom_fields = ProjectCustomField.required.find_each.to_a
+
+    Project.includes(:project_custom_field_project_mappings).find_each do |project|
+      required_project_custom_fields.each do |pcf|
+        if project.project_custom_field_project_mappings.pluck(:custom_field_id).exclude?(pcf.id)
+          ProjectCustomFieldProjectMapping.create!(project_id: project.id, custom_field_id: pcf.id)
+        end
+      end
+    end
```
dombesz: This code is not really efficient, it loads up all the required fields and all the projects into objects which can be costly on a large # of projects or custom fields. Additionally it will run N+1 insert queries which can take a long time. I suggest using just the ids instead of objects and the `insert_all!` method to avoid the N+1 inserts.

```suggestion
    required_custom_field_ids = ProjectCustomField.required.ids

    # Gather the custom_field_ids for every project, then add a new mapping
    # of {project_id:, custom_field_id:} for every project that does not have
    # the required required_custom_field_ids activated.
    missing_custom_field_attributes =
      Project
        .includes(:project_custom_field_project_mappings)
        .pluck(:id, 'project_custom_field_project_mappings.custom_field_id')
        .group_by(&:first)
        .transform_values { |values| values.map(&:last) }
        .reduce([]) do |acc, (project_id, custom_field_ids)|

        missing_custom_field_ids = required_custom_field_ids - custom_field_ids

        acc + missing_custom_field_ids.map do |custom_field_id|
          { project_id: , custom_field_id: }
        end
      end

    ProjectCustomFieldProjectMapping.insert_all!(missing_custom_field_attributes)
```

This code will load the ids only, and inserts all the missing mappings in one instert query.

spec/models/projects/customizable_spec.rb
```diff
@@ -27,92 +27,407 @@
...
 
-      context 'with a value set' do
-        let(:custom_value) do
-          build_stubbed(:custom_value,
-                        custom_field:,
-                        value: true)
+      context 'when custom fields are mapped to this project' do
+        before do
+          project.project_custom_fields << [text_custom_field, bool_custom_field]
+          project.reload # TODO: why is this necessary?
```
dombesz: I believe reload is not required anymore.

spec/models/projects/customizable_spec.rb
```diff
@@ -27,92 +27,407 @@
...
+                          text_custom_field_with_default.id => 'overwritten default'
+                        })
+
+        # text_custom_field_with_default is not provided, thus it should not be enabled (in contrast to creation)
+        expect(project.project_custom_field_project_mappings.pluck(:custom_field_id))
+          .to contain_exactly(text_custom_field.id, bool_custom_field.id, text_custom_field_with_default.id)
+      end
+    end
+
+    it 'does re-enable fields with new value which have been disabled in the past' do
```
dombesz: Is this the usecase to ensure that updates coming from the api will accept the disabled custom fields too? I'm wondering if we should raise a validation error instead?

dombesz (commented):
Finished the second round of review, I will have to review the latest changes to wrap up the review completely.

app/components/settings/project_custom_fields/header_component.html.erb
```diff
@@ -0,0 +1,36 @@
...
+  component_wrapper do  
+    render Primer::OpenProject::PageHeader.new do |header|
+      header.with_title(variant: :default) { t('settings.project_attributes.heading') }
+      header.with_description { t('settings.project_attributes.heading_description') }
+      header.with_actions do
+        flex_layout(justify_content: :space_between, align_items: :center) do |action_buttons_container|
+          action_buttons_container.with_column(mr: 2) do
+            render(Primer::Beta::Button.new(
+              tag: :a,
+              href: new_admin_settings_project_custom_field_path(type: "ProjectCustomField"),
```
dombesz, line 12: I think we should disable this button if there are no `@project_custom_field_sections` are present and also validate the presence of the `ProjectCustomField#custom_field_section_id` field. Currently it is possible to create a project custom field without a section id, which is causing an error while rendering the overview page.
<details><summary>Demo of bug</summary>
<p>

attribute missing

</p>
</details>

spec/features/projects/create_spec.rb
jjabari-op: I don't think so as the non-visible fields are only exposed to global admins, not project admins

spec/models/projects/exporter/csv_integration_spec.rb
jjabari-op: Good catch! Have extended specs accordingly

spec/features/admin/project_custom_fields/list_spec.rb
jjabari-op: They are already implemented ;)

frontend/src/stimulus/controllers/dynamic/projects/settings/project-custom-fields-mapping-filter.controller.ts
jjabari-op: Good catch!

spec/models/projects/customizable_spec.rb
jjabari-op: I figured that's the way we enable the implicit control over the activation state via the API without further API adjustments.

app/controllers/concerns/projects/settings/project_custom_fields/component_streams.rb
```diff
@@ -0,0 +1,69 @@
...
+              component: ::Projects::Settings::ProjectCustomFieldSections::IndexComponent.new(
+                project:,
+                project_custom_field_sections:,
+                project_custom_fields_grouped_by_section:,
+                project_custom_field_project_mappings:
+              )
+            )
+          end
+
+          def update_custom_field_row_via_turbo_stream(
```
dombesz: This method does not seem to be used.

dombesz (approved):
Well done @jjabari-op 🎉

jjabari-op:
Thank you for the review, improvements and merge @dombesz

Eric-Guo:
```sql
Migrating to CreateCustomFieldSections (20231123111357)
== 20231123111357 CreateCustomFieldSections: migrating ========================
-- create_table(:custom_field_sections)
  TRANSACTION (0.1ms)  BEGIN
   (4.7ms)  CREATE TABLE "custom_field_sections" ("id" bigserial primary key, "position" integer, "name" character varying, "type" character varying, "created_at" timestamptz(6) NOT NULL, "updated_at" timestamptz(6) NOT NULL)
   -> 0.0051s
-- add_reference(:custom_fields, :custom_field_section)
   (0.3ms)  ALTER TABLE "custom_fields" ADD "custom_field_section_id" bigint
   (1.3ms)  CREATE INDEX "index_custom_fields_on_custom_field_section_id" ON "custom_fields" ("custom_field_section_id")
   -> 0.0046s
-- add_column(:custom_fields, :position_in_custom_field_section, :integer, {:null=>true})
   (0.2ms)  ALTER TABLE "custom_fields" ADD "position_in_custom_field_section" integer
   -> 0.0006s
  CustomFieldSection Load (0.5ms)  SELECT "custom_field_sections".* FROM "custom_field_sections" WHERE "custom_field_sections"."type" = $1 AND ("custom_field_sections"."position" IS NOT NULL) ORDER BY "custom_field_sections"."position" DESC LIMIT $2  [["type", "ProjectCustomFieldSection"], ["LIMIT", 1]]
  ProjectCustomFieldSection Create (0.6ms)  INSERT INTO "custom_field_sections" ("position", "name", "type", "created_at", "updated_at") VALUES ($1, $2, $3, $4, $5) RETURNING "id"  [["position", 1], ["name", "Project attributes"], ["type", "ProjectCustomFieldSection"], ["created_at", "2024-03-25 14:53:54.495725"], ["updated_at", "2024-03-25 14:53:54.495725"]]
  ProjectCustomField Load (0.7ms)  SELECT "custom_fields".* FROM "custom_fields" WHERE "custom_fields"."type" = $1 ORDER BY "custom_fields"."id" ASC LIMIT $2  [["type", "ProjectCustomField"], ["LIMIT", 1000]]
  CustomOption Load (0.6ms)  SELECT "custom_options".* FROM "custom_options" WHERE "custom_options"."custom_field_id" = $1 ORDER BY "custom_options"."position" ASC  [["custom_field_id", 13]]
  CustomField Pluck (0.2ms)  SELECT "custom_fields"."name" FROM "custom_fields" WHERE "custom_fields"."type" = $1 AND "custom_fields"."id" != $2  [["type", "ProjectCustomField"], ["id", 13]]
  ↳ app/models/custom_field.rb:57:in `uniqueness_of_name_with_scope'
  CustomOption Pluck (0.2ms)  SELECT "custom_options"."id" FROM "custom_options" WHERE "custom_options"."custom_field_id" = $1 AND "custom_options"."default_value" = $2 ORDER BY "custom_options"."position" ASC  [["custom_field_id", 13], ["default_value", true]]
  ↳ app/models/custom_field.rb:83:in `default_value'
  ProjectCustomField Load (1.3ms)  SELECT "custom_fields".* FROM "custom_fields" WHERE "custom_fields"."custom_field_section_id" = $1 AND ("custom_fields"."position_in_custom_field_section" IS NOT NULL) ORDER BY "position_in_custom_field_section" DESC LIMIT $2  [["custom_field_section_id", 1], ["LIMIT", 1]]
  ProjectCustomField Update (0.8ms)  UPDATE "custom_fields" SET "updated_at" = $1 WHERE "custom_fields"."id" = $2  [["updated_at", "2024-03-25 14:53:54.548179"], ["id", 13]]
  ProjectCustomField Load (0.4ms)  SELECT "custom_fields".* FROM "custom_fields" WHERE "custom_fields"."custom_field_section_id" = $1 AND ("custom_fields"."position_in_custom_field_section" IS NOT NULL) ORDER BY "position_in_custom_field_section" DESC LIMIT $2  [["custom_field_section_id", 1], ["LIMIT", 1]]
  ProjectCustomField Count (0.3ms)  SELECT COUNT(*) FROM "custom_fields" WHERE "custom_fields"."custom_field_section_id" = $1 AND ("custom_fields"."position_in_custom_field_section" = 1)  [["custom_field_section_id", 1]]
== 20231123111357 CreateCustomFieldSections: migrated (0.0813s) ===============

  ActiveRecord::SchemaMigration Create (0.2ms)  INSERT INTO "schema_migrations" ("version") VALUES ('20231123111357') RETURNING "version"
  TRANSACTION (0.7ms)  COMMIT
```

Nice work, but I found when first doing the CreateCustomFieldSections migrate, existing `ProjectCustomField` (in this case, custom_fields.id==13) the columns custom_field_section_id is not filled, maybe due to custom_field_section is just added.

So I re-run below code and got the custom_field_section_id updated.

```ruby
[2] pry(main)> ProjectCustomField.first
  ProjectCustomField Load (0.6ms)  SELECT "custom_fields".* FROM "custom_fields" WHERE "custom_fields"."type" = $1 ORDER BY "custom_fields"."id" ASC LIMIT $2  [["type", "ProjectCustomField"], ["LIMIT", 1]]
=> #<ProjectCustomField:0x0000000134610960
 id: 13,
 type: "ProjectCustomField",
 field_format: "list",
 regexp: "",
 min_length: 0,
 max_length: 0,
 is_required: false,
 is_for_all: false,
 is_filter: false,
 position: 1,
 searchable: true,
 editable: true,
 visible: true,
 multi_value: false,
 default_value: nil,
 name: "测试项目字段",
 created_at: Thu, 25 Jan 2024 13:53:11.813609000 CST +08:00,
 updated_at: Mon, 25 Mar 2024 22:53:54.548179000 CST +08:00,
 content_right_to_left: false,
 allow_non_open_versions: false,
 custom_field_section_id: nil,
 position_in_custom_field_section: nil>
[3] pry(main)> section = ProjectCustomFieldSection.create!(    
[3] pry(main)*   name: "Project attributes 2"      
[3] pry(main)* )      
  TRANSACTION (0.1ms)  BEGIN
  CustomFieldSection Load (1.2ms)  SELECT "custom_field_sections".* FROM "custom_field_sections" WHERE "custom_field_sections"."type" = $1 AND ("custom_field_sections"."position" IS NOT NULL) ORDER BY "custom_field_sections"."position" DESC LIMIT $2  [["type", "ProjectCustomFieldSection"], ["LIMIT", 1]]
  ProjectCustomFieldSection Create (0.5ms)  INSERT INTO "custom_field_sections" ("position", "name", "type", "created_at", "updated_at") VALUES ($1, $2, $3, $4, $5) RETURNING "id"  [["position", 2], ["name", "Project attributes 2"], ["type", "ProjectCustomFieldSection"], ["created_at", "2024-03-25 14:56:14.169673"], ["updated_at", "2024-03-25 14:56:14.169673"]]
  TRANSACTION (0.9ms)  COMMIT
=> #<ProjectCustomFieldSection:0x0000000169bb6b78
 id: 2,
 position: 2,
 name: "Project attributes 2",
 type: "ProjectCustomFieldSection",
 created_at: Mon, 25 Mar 2024 22:56:14.169673000 CST +08:00,
 updated_at: Mon, 25 Mar 2024 22:56:14.169673000 CST +08:00>
[4] pry(main)> ProjectCustomField.find_each do |project_custom_field|    
[4] pry(main)*   project_custom_field.update!(custom_field_section_id: section.id)      
[4] pry(main)* end      
  ProjectCustomField Load (0.4ms)  SELECT "custom_fields".* FROM "custom_fields" WHERE "custom_fields"."type" = $1 ORDER BY "custom_fields"."id" ASC LIMIT $2  [["type", "ProjectCustomField"], ["LIMIT", 1000]]
  TRANSACTION (3.6ms)  BEGIN
  CustomOption Load (0.7ms)  SELECT "custom_options".* FROM "custom_options" WHERE "custom_options"."custom_field_id" = $1 ORDER BY "custom_options"."position" ASC  [["custom_field_id", 13]]
  CustomField Pluck (0.3ms)  SELECT "custom_fields"."name" FROM "custom_fields" WHERE "custom_fields"."type" = $1 AND "custom_fields"."id" != $2  [["type", "ProjectCustomField"], ["id", 13]]
  CustomOption Pluck (0.2ms)  SELECT "custom_options"."id" FROM "custom_options" WHERE "custom_options"."custom_field_id" = $1 AND "custom_options"."default_value" = $2 ORDER BY "custom_options"."position" ASC  [["custom_field_id", 13], ["default_value", true]]
  ProjectCustomField Load (0.3ms)  SELECT "custom_fields".* FROM "custom_fields" WHERE "custom_fields"."custom_field_section_id" = $1 AND ("custom_fields"."position_in_custom_field_section" IS NOT NULL) ORDER BY "custom_fields"."position_in_custom_field_section" DESC LIMIT $2  [["custom_field_section_id", 2], ["LIMIT", 1]]
  ProjectCustomField Update (0.4ms)  UPDATE "custom_fields" SET "updated_at" = $1, "custom_field_section_id" = $2, "position_in_custom_field_section" = $3 WHERE "custom_fields"."id" = $4  [["updated_at", "2024-03-25 14:56:29.008077"], ["custom_field_section_id", 2], ["position_in_custom_field_section", 1], ["id", 13]]
  ProjectCustomField Load (0.2ms)  SELECT "custom_fields".* FROM "custom_fields" WHERE "custom_fields"."custom_field_section_id" = $1 AND ("custom_fields"."position_in_custom_field_section" IS NOT NULL) ORDER BY "custom_fields"."position_in_custom_field_section" DESC LIMIT $2  [["custom_field_section_id", 2], ["LIMIT", 1]]
  ProjectCustomField Count (0.2ms)  SELECT COUNT(*) FROM "custom_fields" WHERE "custom_fields"."custom_field_section_id" = $1 AND ("custom_fields"."position_in_custom_field_section" = 1)  [["custom_field_section_id", 2]]
  TRANSACTION (1.3ms)  COMMIT
=> nil
```

I try at TH staging server, it will lead 500 error due to custom_field_section_id is NULL. I try running above code to fix and it works.

dombesz:
Thanks for the report @Eric-Guo ! I have filed a bug report on your behalf in our community instance https://community.openproject.org/wp/53728 . We will look into this before the release.

PR #14542, opened 2024-01-11
Move and rename WP share permissions

ulferts:
* The permissions for sharing work packages are moved to the work packages section on the role administration
* The view_shared_work_packages is renamed in the UI

----

https://community.openproject.org/wp/52086


aaron-contreras (approved):
👍

PR #14545, opened 2024-01-11
Bug 52089: fixed WP lockVersion in API for change_work_package_status

virgo47:
This fixes bug https://community.openproject.org/projects/openproject/work_packages/52089/activity

I though about `add_work_package_attachments` permission as well, but if the user has only that permission, he should not be able to reach the form call (everything on WP detail is inactive) and also, attachment upload uses a different endpoint that should not rely on `lockVersion` info.


ulferts (approved):
Merci beaucoup @virgo47 👍

PR #14549, opened 2024-01-12
allow changing configuration at runtime where sensible

machisuji:
A lot of configuration options are not writable at runtime due to historical reasons. I believe a lot of them are fine to change at runtime now. I have gone through all the non-writable definitions and made them writable where it makes sense and doesn't pose a problem (e.g. global variables, caches, initializers etc.).


ulferts (approved):
🙈

PR #14563, opened 2024-01-15
[#51698] added specific css for grid-area containing client ids

Kharonus:
[#51698](https://community.openproject.org/wp/51698)

### Wat?

- added specific handling for descriptions with credentials -> added `break-word` to those grid-areas

### DIFF
--- a/modules/storages/app/components/storages/admin/oauth_application_info_component.html.erb
+++ b/modules/storages/app/components/storages/admin/oauth_application_info_component.html.erb
@@ -5,7 +5,11 @@
       concat(configuration_check_label_for(:openproject_oauth_application_configured))
     end
 
-    grid.with_area(:description, tag: :div, color: :subtle, test_selector: 'storage-openproject-oauth-application-description') do
+    grid.with_area(:description,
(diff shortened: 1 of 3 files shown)


akabiru (commented):
🟠 Nice one! I looked if primer CSS already has these utilities and it seems it does

- https://primer.style/css/storybook/?path=/story/utilities-typography--word-break
- https://github.com/primer/css/blob/207070961eff673ed83c055af6ec8d28f7002b1a/src/utilities/typography.scss#L289

`wb-break-word should do the trick. It is however missing the `overflow: hidden` property but I'm not sure whether it's necessary in the first place as `word-wrap` should already apply- and nothing overflows?

Lmk if you considered this already, happy to chat 🙂 

---

Edit: `wb-break-word` as we need to break on the labels `OAuth Client ID: ` - `wb-break-all` does not respect this

<details> <summary>Click to view Screenshot</summary>

Screenshot 2024-01-16 at 2 38 15 PM

</details>

akabiru (approved):
:shipit:

PR #14564, opened 2024-01-15
Permission management for Sharepoint

mereghost:
### Related WorkPackage: [OP#51711](https://community.openproject.org/projects/openproject/work_packages/51711)

This PR aims to add permission management capabilities to the Sharepoint/OneDrive integration.

Updating the permissions isn't fully supported yet, so it is easier to just drop them all and create new ones.


spec/support/vcr.rb
```diff
@@ -35,6 +35,8 @@
   config.before_record do |i|
     i.response.body.force_encoding('UTF-8')
   end
+
+  config.default_cassette_options = { record: ENV.fetch('VCR_RECORD_MODE', :once).to_sym }
```
mereghost, line 39: I added this so I didn't have to delete the broken cassette every single time.

Kharonus: what to you provide as value, if you want to rerecord? `:once` is the default, so I assume, this is our current mode. If I want a force recreate, what do I put in?

modules/storages/app/common/storages/peripherals/storage_interaction/one_drive/set_permissions_command.rb
```diff
@@ -0,0 +1,156 @@
...
+
+          def item_exists?(item_id)
+            Util.using_admin_token(@storage) { |http| handle_response(http.get(item_path(item_id))) }
+          end
+
+          def get_permissions(path)
+            Util.using_admin_token(@storage) { |http| handle_response(http.get(permissions_path(path))) }
+          end
+
+          def apply_permission_changes(role, permissions, permission_set_id, item_id)
```
Kharonus: 🟡 Can you rethink the naming of the method signature? It differs quite heavily from how it is called in line 63.

role vs. permission
permissions vs. user_ids
permission_id (as on value in the map called permission_ids) vs permission_set_id
path vs item_id

Kharonus: Maybe this starts, with `call(path:)` being named `call(item_id)`? Apparently it is never a path?

modules/storages/spec/common/storages/peripherals/storage_interaction/one_drive/set_permissions_command_spec.rb
```diff
@@ -0,0 +1,161 @@
...
+
+RSpec.describe Storages::Peripherals::StorageInteraction::OneDrive::SetPermissionsCommand, :webmock do
+  let(:storage) do
+    create(:sharepoint_dev_drive_storage,
+           drive_id: 'b!dmVLG22QlE2PSW0AqVB7UOhZ8n7tjkVGkgqLNnuw2ODRDvn3haLiQIhB5UYNdqMy')
+  end
+
+  let(:permissions_command) { described_class.new(storage) }
+
+  let(:folder) do
```
Kharonus, line 42: 🟢 nice idea to build the folder in the spec and delete it afterwards for the VCR recording.

modules/storages/app/common/storages/peripherals/storage_interaction/one_drive/set_permissions_command.rb
```diff
@@ -0,0 +1,156 @@
...
+            @storage = storage
+          end
+
+          def call(path:, permissions:)
+            # This could be broken down into:
+            # 1. Get Permission IDs
+            # 2. Apply Permission Changes
+            # 2.1 Create Permissions
+            # 2.2 Delete Permissions
+            # 2.3 Re-create Permissions
```
Kharonus: 🟡 is that true? from code perspective, I cannot see a point `2.1 Create Permissions`, see line 85

Kharonus (approved):
Looks very good. I only have couple of comprehensive questions.

modules/storages/app/common/storages/peripherals/storage_interaction/one_drive/set_permissions_command.rb
mereghost: For `.call` I'm keeping the same signature as the nextcloud implementation.

modules/storages/app/common/storages/peripherals/storage_interaction/one_drive/set_permissions_command.rb
mereghost: When there are no permissions set, we need to create them. In other words, with a new folder we can't delete permissions because they don't exist and need to create new.

Does this answer your questions or am I being extra dense (probably the case)?

modules/storages/app/common/storages/peripherals/storage_interaction/one_drive/set_permissions_command.rb
mereghost: But to address the `apply_permission_changes` method signature, this amount of args scream `MISSING OBJECT, dumbass!`, but I haven't settled for a good name yet.

modules/storages/app/common/storages/peripherals/storage_interaction/one_drive/set_permissions_command.rb
Kharonus: probably just a misunderstanding of the words of the comment. Looking at the code, I'd write the comment as:

```
# 2. Apply Permission Changes
# 2.1 Delete Permissions (skipped if not existent)
# 2.2 (Re-)Create Permissions
```

modules/storages/app/common/storages/peripherals/storage_interaction/one_drive/set_permissions_command.rb
Kharonus: True, mmhm... Not very satisfied with looking just at the one drive command. But I don't have a better solution here.

Object is probably a good replacement for the signature issue.

PR #14568, opened 2024-01-15
[#51679] Can not add invited users to existing groups

ba1ash:
https://community.openproject.org/work_packages/51679
`NEW USER` section with autocomplete is absent if there are only invited users not present in the group. So, I extended the condition when the section is rendered (invited users are taken into account).
Note: the autocomplete itself includes invited users.

### DIFF
--- a/app/views/groups/_users.html.erb
+++ b/app/views/groups/_users.html.erb
@@ -42,12 +42,10 @@ See COPYRIGHT and LICENSE files for more details.
   </div>
 
   <div class="grid-content medium-6 -visible-overflow">
-    <% users = User
-      .user
-      .active
-      .not_in_group(@group)
-      .limit(1) %>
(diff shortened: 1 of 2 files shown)


app/views/groups/_users.html.erb
```diff
@@ -42,12 +42,10 @@ See COPYRIGHT and LICENSE files for more details.
...
-    <% users = User
-      .user
-      .active
-      .not_in_group(@group)
-      .limit(1) %>
-    <% if users.any? %>
+    <% if User.user
+              .not_in_group(@group)
+              .where(status: [User.statuses[:active], User.statuses[:invited]])
+              .exists? %>
```
Kharonus, line 48: 🟢 The whole condition is written more legible! Thanks!

Kharonus (approved):
LGTM

PR #14574, opened 2024-01-16
[52090] Allow changing status in backlogs with only change_work_package_status permission

cbliard:
https://community.openproject.org/wp/52090

- Allow changing status in backlogs with only `change_work_package_status` permission and disable the other fields
- Disable drag and drop if the user does not have `edit_work_packages` permission
- Hide the "Add new story" menu entry if the user does not have `add_work_packages` permission

### DIFF
--- a/frontend/src/app/features/backlogs/backlogs-page/styles/master_backlog.sass
+++ b/frontend/src/app/features/backlogs/backlogs-page/styles/master_backlog.sass
@@ -199,8 +199,9 @@
 
         .error.icon.icon-bug
           text-align: left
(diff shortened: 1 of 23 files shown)


ulferts (approved):
Looks good with the exception of two small things.
* The "Task board" item is selectable even if the user lacks the permission to "View taskboards". I expect this has been the case all along. But in the scenario here, it becomes more apparent. 

image

* A feature spec where a user with the new permission updates the status would be appreciated to avoid regressions.

Feel free to merge yourself once those topics are addressed.

PR #14580, opened 2024-01-16
Accessibility/51021 test for basic accessibility for the new and edit (primer) file storage pages

akabiru:
https://community.openproject.org/work_packages/51021

### Accessibility

📖 https://www.openproject.org/docs/development/running-tests/#accessibility-tests

- [x] Use [axe-core-rspec](https://github.com/dequelabs/axe-core-gems/blob/develop/packages/axe-core-rspec/README.md) in automated accessibility tests to provide continuous regression testing against common accessibility issues.
- [x] Use [capybara-accessible-selectors](https://github.com/citizensadvice/capybara_accessible_selectors) in [feature tests](https://www.openproject.org/docs/development/running-tests/#feature-tests) to find UI elements using screen-reader compatible selectors. This ensures the page elements used by feature tests are accessible to assistive technologies.

### Code Maintenance

- [x] Decomposes storage components into testable units, breaking away from convoluted tests at the feature level- rather we have components now fully tested and the feature level focused on user behaviour
- [x] Break up the large `admin_storages_spec` into action specific (index,create,edit) specs


modules/storages/spec/features/admin_storages_spec.rb
```diff
@@ -58,9 +58,11 @@
       it 'renders a list of storages' do
         visit admin_settings_storages_path
 
+        expect(page).to be_axe_clean.within '#content'
+
         expect(page).to have_test_selector('storage-name', text: complete_storage.name)
```
Kharonus: 🟡 should we replace selecting on `test-selector` here, too, whenever it is possible? Point is, that this `data-test-selector` is never visible to any screenreader. So, in theory, we could get false-positives here, when we have a not visible element with that html attribute.

modules/storages/spec/features/admin_storages_spec.rb
```diff
@@ -58,9 +58,11 @@
...
         expect(page).to have_test_selector('storage-name', text: incomplete_storage.name)
-        expect(page).to have_css("button[aria-label='Add new storage']", text: 'Storage')
```
Kharonus, line 63: 🟡 should we use `.to have_button` over `.to have_selector :button`?

Kharonus (commented):
Two basic questions, that could change the overall number of lines to be changed. ^^ Feel free, to share your opinion, as those are the first specs we rewrite with those selectors.

modules/storages/spec/features/admin_storages_spec.rb
akabiru: ~Good point, will replace this with localised `test-selector` 👍🏾~

Edit: We had a chat and the feedback is regarding `have_test_selector` - we settled right now that there doesn't seem to be a better way to make the test more "accessibly declarative" save for scoping `within_test_selector` which is functionally the same thing

modules/storages/spec/features/admin_storages_spec.rb
```diff
@@ -107,7 +107,7 @@
         expect(page).to be_axe_clean.within '#content'
 
         # Show Add storage button
-        expect(page).to have_selector :button, 'Storage', aria: { label: 'Add new storage' }
+        expect(page).to have_button 'Storage', aria: { label: 'Add new storage' }
```
aaron-contreras: Didn't know matchers for node attributes could be specified this way just as in views, neat! 🤌

modules/storages/spec/features/admin_storages_spec.rb
aaron-contreras: Would `have_link` not do a better job of what is trying to be accomplished?

akabiru: > Would `have_link` not do a better job of what is trying to be accomplished?

Good idea, it might- but we had a second look and Christophe proposed a really nice rewrite. Will ping once it's up 👍🏾

modules/storages/spec/features/storages/admin/edit_storage_spec.rb
```diff
@@ -0,0 +1,318 @@
...
+        # Automatically managed project folders
+        expect(page).to have_test_selector('storage-managed-project-folders-label',
+                                           text: 'Automatically managed folders')
+
+        expect(page).to have_test_selector('label-managed-project-folders-status', text: 'Active')
+        expect(page).to have_test_selector('storage-automatically-managed-project-folders-description',
+                                           text: 'Let OpenProject create folders per project automatically.')
+      end
+
+      aggregate_failures 'General information' do
```
aaron-contreras, line 112: Yes! 👏

modules/storages/spec/components/storages/admin/new_storage_button_component_spec.rb
```diff
@@ -33,11 +33,10 @@
...
-    with_rendered_component_path(render_inline(described_class.new)) do |_rendered_fragement_path|
-      expect(page).to have_button 'Storage', aria: { label: 'Add new storage' }
+    render_inline(described_class.new)
+    expect(page).to have_button 'Storage', aria: { label: 'Add new storage' }
 
-      expect(page).to have_link 'Nextcloud', href: select_provider_admin_settings_storages_path(provider: 'nextcloud')
-      expect(page).to have_link 'OneDrive/SharePoint', href: select_provider_admin_settings_storages_path(provider: 'one_drive')
-    end
+    expect(page).to have_link 'Nextcloud', href: select_provider_admin_settings_storages_path(provider: 'nextcloud')
+    expect(page).to have_link 'OneDrive/SharePoint', href: select_provider_admin_settings_storages_path(provider: 'one_drive')
```
aaron-contreras, line 40: Unrelated to the change but do we have translations for these?

akabiru: Sure, there are translations- in test we assert against the raw en string for accuracy- i.e. if we just assert against a dynamic value from `I18n` - then if there's a typo in the translations file or the format is incorrect it's relatively difficult to catch

---

See: https://github.com/opf/openproject/blob/dev/modules/storages/app/components/storages/admin/new_storage_button_component.html.erb#L12

Kharonus (approved):
Love it! This is a huge improvement to test coverage and readability of those tests.

PR #14593, opened 2024-01-17
fix[Op-52158]: Drive id message should include `ID` - "Drive ID can't be blank"

akabiru:
#### https://community.openproject.org/wp/52158

Similar to https://github.com/opf/openproject/pull/14354/commits/ca0fc06d369ab3dc3c6d8a83290df1f612ba339b - active model errors strips out the "ID" from the message when the attribute name has an "_id" suffix. This overrides the default message by specifying the relation `drive` label

🐛  _Before_

Screenshot 2024-01-17 at 3 30 00 PM

🔧  _After_

Screenshot 2024-01-17 at 3 43 27 PM

### DIFF
--- a/modules/storages/app/contracts/storages/storages/one_drive_contract.rb
+++ b/modules/storages/app/contracts/storages/storages/one_drive_contract.rb
@@ -33,7 +33,7 @@ module Storages::Storages
     attribute :host
     validates :host, absence: true
     attribute :tenant_id
-    validates :tenant_id, presence: true
+    validates :tenant_id, format: { with: /\A(?:[a-f0-9]{8}-[a-f0-9]{4}-[a-f0-9]{4}-[a-f0-9]{4}-[a-f0-9]{12}|consumers)\z/i }
     attribute :drive_id
     validates :drive_id, presence: true, allow_nil: true
   end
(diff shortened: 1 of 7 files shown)


modules/storages/app/contracts/storages/storages/one_drive_contract.rb
```diff
@@ -35,6 +35,8 @@ class OneDriveContract < ::ModelContract
     attribute :tenant_id
     validates :tenant_id, format: { with: /\A(?:[a-f0-9]{8}-[a-f0-9]{4}-[a-f0-9]{4}-[a-f0-9]{4}-[a-f0-9]{12}|consumers)\z/i }
     attribute :drive_id
-    validates :drive_id, presence: true, allow_nil: true
+    validates :drive_id,
+              presence: { message: I18n.t('activerecord.errors.models.storages/storage.attributes.drive_id.id_blank') },
```
Kharonus: ❓ why do we need a specific localized message for `drive_id` and not for `tenant_id`? according to your screen, the message for tenant contains the id, but for drive it does not (before your fix). How is that?

akabiru: active model errors strips out the "ID" from the message when the attribute name has an "_id" suffix.

akabiru: > ❓ why do we need a specific localized message for `drive_id` and not for `tenant_id`? according to your screen, the message for tenant contains the id, but for drive it does not (before your fix). How is that?

Ah, indeed odd- let me compare! Could do without additional translations 👀

dominic-braeunlein (approved):
LGTM 👍

Kharonus (approved):
Brilliant! even removed localisations, we no longer need.

PR #14598, opened 2024-01-17
[#52181] bad error message on request timeout to remote storage

Kharonus:
- https://community.openproject.org/work_packages/52181
- fixed error handling of httpx responses

### DIFF
--- a/modules/storages/app/common/storages/peripherals/storage_interaction/one_drive/download_link_query.rb
+++ b/modules/storages/app/common/storages/peripherals/storage_interaction/one_drive/download_link_query.rb
@@ -52,11 +52,7 @@ module Storages
                 headers: { 'Authorization' => "Bearer #{token.access_token}" }
               )
 
-              if (300..399).include?(response.status)
-                ServiceResult.success(result: response.headers['Location'])
-              else
-                handle_errors(response)
-              end
+              handle_errors(response)
             end
           end
 
@@ -64,18 +60,20 @@ module Storages
 
           def handle_errors(response)
             case response
-            when 404
+            in { status: 300..399 }
+              ServiceResult.success(result: response.headers['Location'])
+            in { status: 404 }
               ServiceResult.failure(result: :not_found,
-                                    errors: ::Storages::StorageError.new(code: :not_found, data: response.body))
-            when 403
+                                    errors: ::Storages::StorageError.new(code: :not_found, data: response))
+            in { status: 403 }
               ServiceResult.failure(result: :forbidden,
-                                    errors: ::Storages::StorageError.new(code: :forbidden, data: response.body))
-            when 401
+                                    errors: ::Storages::StorageError.new(code: :forbidden, data: response))
+            in { status: 401 }
               ServiceResult.failure(result: :unauthorized,
-                                    errors: ::Storages::StorageError.new(code: :unauthorized, data: response.body))
+                                    errors: ::Storages::StorageError.new(code: :unauthorized, data: response))
             else
               ServiceResult.failure(result: :error,
-                                    errors: ::Storages::StorageError.new(code: :error, data: response.body))
+                                    errors: ::Storages::StorageError.new(code: :error, data: response))
             end
           end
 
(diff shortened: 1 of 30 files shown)


modules/storages/app/common/storages/peripherals/storage_interaction/one_drive/download_link_query.rb
```diff
@@ -52,30 +52,31 @@ def call(user:, file_link:)
...
-                handle_errors(response)
-              end
+              handle_errors(response)
             end
           end
 
           private
 
           def handle_errors(response)
+            json = MultiJson.load(response.body.to_s, symbolize_keys: true)
```
ba1ash: If `response` is a `HTTPX::ErrorResponse` then we will have an issue here because there is no `HTTPX::ErrorResponse#body`.
So, I suggest to postpone json parsing to the moment we are sure body is present. And then json can be received from response like: `response.json(symbolize_names: true)`.

modules/storages/app/common/storages/peripherals/storage_interaction/one_drive/files_query.rb
```diff
@@ -69,13 +69,13 @@ def handle_response(response, map_value)
             json = MultiJson.load(response.body.to_s, symbolize_keys: true)
```
ba1ash: Same issue.

modules/storages/app/common/storages/peripherals/storage_interaction/one_drive/internal/drive_item_query.rb
```diff
@@ -46,16 +46,16 @@ def handle_responses(response)
               json = MultiJson.load(response.body.to_s, symbolize_keys: true)
```
ba1ash: Same issue.

modules/storages/app/common/storages/peripherals/storage_interaction/one_drive/open_storage_query.rb
```diff
@@ -62,16 +62,16 @@ def handle_responses(response)
             json = MultiJson.load(response.body.to_s, symbolize_keys: true)
```
ba1ash: Same issue.

ba1ash (commented):
I like it overall. Pattern matching is neat :smile: 
I have a couple of concerns described in the comments. And I suggest to reproduce the issue with timeout at least for one request in a test case. I would be glad to assist.

modules/storages/app/common/storages/peripherals/storage_interaction/one_drive/download_link_query.rb
Kharonus: very good catch

Kharonus: Can we somehow write a test for it? you know, have a webmock stub, that is converted by httpx into an `ErrorResponse`? I will have a look.

ba1ash: Have a look at this one: https://github.com/opf/openproject/blob/dev/modules/storages/spec/contracts/storages/storages/shared_contract_examples.rb#L180

But actually there is a problem with it, because it raises an error while in reality it does not. I will think about it.

ba1ash: But actually I think we should do it with WebMock: https://github.com/bblimke/webmock?tab=readme-ov-file#raising-timeout-errors
Then whatever http library we use will behave accordingly(raise a timeout error or return an appropriate error object)

Kharonus: I solved it now with mocking the `HTTPX::get` method. This has the advantage, that I can have those tests right next to all the other unit tests of the query.

Inside a VCR context we cannot use webmock and `.to_timeout`

ba1ash (approved):
LGTM! Thanks!

PR #14599, opened 2024-01-17
[#51188] Split "% Complete" into a derived-value and self-value

aaron-contreras:
See: https://community.openproject.org/work_packages/51188

This PR will consist of multiple implementation Work Packages and merged in as a single feature. So far, it starts with the following:

- [ ] https://community.openproject.org/projects/openproject/work_packages/50957
- [ ] https://community.openproject.org/projects/openproject/work_packages/51187


db/migrate/20240116165933_add_derived_done_ratio_to_work_packages.rb
```diff
@@ -0,0 +1,66 @@
...
+
+      direction.down do
+        rollback_from_derived_done_ratio
+      end
+    end
+  end
+
+  def migrate_to_derived_done_ratio
+    non_leaf_work_packages
+      .where.not(done_ratio: nil)
```
cbliard: Not sure if this condition is actually useful: the `done_ratio` field has a non-null constraint and defaults to 0.

db/migrate/20240116165933_add_derived_done_ratio_to_work_packages.rb
```diff
@@ -0,0 +1,66 @@
...
+      .update_all("done_ratio = derived_done_ratio")
+  end
+
+  def non_leaf_work_packages
+    WorkPackage.where.not(id: WorkPackage.leaves)
+  end
+
+  # Sets the SystemUser as the author of the Journal
+  # entries generated by this migration
+  def journal_user = SystemUser.first
```
cbliard: Using `User.system` is better, as it will create the system user if it does not exist, although it should be already created at this point.

app/contracts/work_packages/base_contract.rb
```diff
@@ -56,6 +56,8 @@ class BaseContract < ::ModelContract
               writable: ->(*) {
                 model.leaf? && Setting.work_package_done_ratio == 'field'
               }
+    attribute :derived_done_ratio,
+              writable: false
```
cbliard, line 60: > I prefer having this explicitly marked as false than having to infer that because of the attribute not being listed, it's non-writable.

I totally agree. That was also my reasoning with `:derived_remaining_hours`.

cbliard: And a test in `lib/api/v3/work_packages/schema/work_package_schema_representer.rb` would be nice, to ensure that making it writable by accident would be catched by a unit test.

spec/services/work_packages/update_service_integration_spec.rb
```diff
@@ -413,34 +415,29 @@
         .to be_success
 
       # receives the provided start/finish date
-      expect(work_package.start_date)
-        .to eql(attributes[:start_date])
-      expect(work_package.due_date)
-        .to eql(attributes[:due_date])
+      expect(work_package)
+        .to have_attributes(start_date: attributes[:start_date],
+                            due_date: attributes[:due_date])
```
cbliard, line 420: 🖌️ 
I too like the `have_attributes` matcher. Especially when the test fails, the expectation message feels nicer (even if it is sometimes a bit messy when objects have a lot of instance variables).

cbliard (changes requested):
Nice and clean. Congrats!

I think a few things can be improved:
- the derived % complete should be hidden for leaves work package.
- the derived value should not look like a link, or it should be a proper link, but blue with underline on hover makes it feel like a link and clicking it is deceiving this expectation as nothing happens.

Once done you can merge it.

cbliard:
Another thing I have just seen while working on another work package: the derived % complete is now available in the form configuration of types.

image

It should not be there.

I now understand that that's how the derived remaining work got added by accident. As soon as I found how to exclude it, I'll share the method to do so with you here.

EDIT: it needs to be added to the `EXCLUDED` array in `app/models/type/attributes.rb`.

aaron-contreras:
I've addressed your feedback :) @cbliard

frontend/src/app/shared/components/fields/display/field-types/compound-progress-display-field.module.ts
```diff
@@ -0,0 +1,140 @@
...
+            f: [{ n: 'parent', o: '=', v: [wpID] }],
+          };
+          const href = URI(this.PathHelper.projectWorkPackagesPath(project.identifier as string))
+            .query({ query_props: JSON.stringify(props) })
+            .toString();
+
+          link.href = href;
+        });
+    }
+  }
```
cbliard: The links to view work estimates of work package and its children should show the same query regardless of clinking on the derived work, remaining work or % complete.

The simplest to do so would be to have the code for `addURLToViewWorkPackageChildren` in `EstimatedTimeDisplayField` and `CompoundProgressDisplayField` deduplicated.

cbliard (changes requested):
Thanks for addressing my previous remarks 👍 

There are two things remaining: 
- the code to generate the url is the same as for the work/remaining work and should be factored out.
- the attribute `derived_percentage_done` must not appear in the work package form. To exclude it from the form, add the field name to `EXCLUDED` in `app/models/type/attributes.rb`.

frontend/src/app/shared/components/fields/display/field-types/compound-progress-display-field.module.ts
aaron-contreras: Okay, I interpreted it as clicking on work related links leads to a work-focused query (sans % complete), and clicking on the % complete link leads to a query including the % complete column. Thanks for clarifying

cbliard (approved):
Nice!

PR #14606, opened 2024-01-18
[#51738] Enable HTTPX persistent connections.

ba1ash:
https://community.openproject.org/work_packages/51738

### DIFF
--- a/lib/open_project.rb
+++ b/lib/open_project.rb
@@ -44,4 +44,11 @@ module OpenProject
   def self.logger
     ::OpenProject::Logging::LogDelegator
   end
+
+  def self.httpx
+    HTTPX
+      .plugin(:persistent)
+      .plugin(:basic_auth)
+      .plugin(:webdav)
+  end
 end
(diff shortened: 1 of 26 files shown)


lib/open_project.rb
```diff
@@ -44,4 +44,11 @@ module OpenProject
   def self.logger
     ::OpenProject::Logging::LogDelegator
   end
+
+  def self.httpx
+    HTTPX
+      .plugin(:persistent)
+      .plugin(:basic_auth)
+      .plugin(:webdav)
+  end
```
akabiru, line 53: 🟢 Nice unification 👍🏾  One thing that might change in the future is that not all plugins will be needed in all contexts- and I wonder if there's expense in eager loading them.  But as they say, _"we can cross that bridge when when we get there"_ 😄

akabiru (approved):
Nice one 👍🏾  Just one remark from me, otherwise LGTM

lib/open_project.rb
mereghost: The plugins, in general, just load a couple of files and add a couple of hooks to the session, so shouldn't be an issue.

mereghost (approved):

lib/open_project.rb
ba1ash: Yes, it should not be an issue and plugins code is loaded once we create a session like that for the first time.

PR #14607, opened 2024-01-18
[#52193] PDF export fails with "undefined method `sourcepos'"

as-op:
https://community.openproject.org/work_packages/52193


machisuji (approved):
👍

PR #14645, opened 2024-01-25
fix nested checkboxes style

toy:
Nested lists have a rule for a marker for several levels which has higher specificity than the rule for checkboxes to not have a marker.

| was | now |
| --- | --- |
| Screenshot 2024-01-25 at 13 44 31 | Screenshot 2024-01-25 at 13 44 56 |

https://community.openproject.org/projects/openproject/work_packages/51247/activity


HDinger (approved):
Lgtm 👍

PR #14648, opened 2024-01-25
Implements the OneDrive Sync Service

mereghost:
## What?

This PR introduces the `OneDriveSyncService` as required by the [OP#52394](https://community.openproject.org/projects/openproject/work_packages/52394/activity). The objective here being creating a similar service to the one used by Nextcloud - `NextcloudGroupFolderSyncService`, formerly known as `GroupFolderPropertiesSyncService`.

What it should do?

1. Create a project folders at the File Storage root folder
2. Rename any "misnamed" folder (ensure that the current project name is used)
3. Add and remove **logged in** users so that the permissions match (as close as possible) to the ones in OP
4. Hide any inactive project folders. 

## Why?

This is one of the final pieces needed for the Automatically Managed Project Folders on OneDrive/SharePoint and intends to ensure feature parity between SharePoint and Nextcloud.

## How?

I will go through the order of operations here:

0. Early exit if, somehow, the storage passed has `automatically_managed` set to `false`

1. [Get all existing folders](https://github.com/opf/openproject/blob/914ec5bef6bd44637ebb93964daa7773695432b4/modules/storages/app/services/storages/one_drive_managed_folder_sync_service.rb#L49), their names and IDs: this information is parsed and just the ID/name tuples are returned. If an error occurs here, we need to terminate in an error state as, probably, the credentials are broken or the storage mis-configured.
2. Following this we ensure that the [folders do exist](https://github.com/opf/openproject/blob/914ec5bef6bd44637ebb93964daa7773695432b4/modules/storages/app/services/storages/one_drive_managed_folder_sync_service.rb#L51) by:
   1. [Creating the folders](https://github.com/opf/openproject/blob/914ec5bef6bd44637ebb93964daa7773695432b4/modules/storages/app/services/storages/one_drive_managed_folder_sync_service.rb#L73) if the project doesn't have one associated with it or it cannot be found.
   2. [Renaming the folder](https://github.com/opf/openproject/blob/914ec5bef6bd44637ebb93964daa7773695432b4/modules/storages/app/services/storages/one_drive_managed_folder_sync_service.rb#L76) if needed (i.e. the project changed name)
3. And then we [hide the inactive](https://github.com/opf/openproject/blob/914ec5bef6bd44637ebb93964daa7773695432b4/modules/storages/app/services/storages/one_drive_managed_folder_sync_service.rb#L85) project folders. Hiding a folder just means removing everyone's access from it.

None of these operations warrant an early exit and will be logged by the Rails Logger in case some error happens using the structure:

```json
{ 
    "command" : "Storages::Peripherals::StorageInteraction::OneDrive::SetPermissionsCommand",
    "message" : "Outbound request unauthorized",
    "data" : { "status" : "401", "body" : "response body" }
}
```

From there we can move to setting permissions. On `Sharepoint/OneDrive` there's only 2 levels of access we can grant: `read` and `write`. So, when mapping to `OpenProject` permission system, I only grant `read` permissions if the only file related permission given is `read_file` and `write` if the user has access to `write_files`, this implies the `share_files, delete_files` are also given.

When [applying the new permissions](https://github.com/opf/openproject/blob/914ec5bef6bd44637ebb93964daa7773695432b4/modules/storages/app/services/storages/one_drive_managed_folder_sync_service.rb#L57) we ensure that the **OpenProject Admin** has full write access to the folder and we loop through all the users checking their permissions and creating the hash argument for the `SetPermissionsCommand`. Any errors here will also be logged by the Rails Logger.

### DIFF
--- a/modules/storages/app/common/storages/peripherals/storage_interaction/one_drive/create_folder_command.rb
+++ b/modules/storages/app/common/storages/peripherals/storage_interaction/one_drive/create_folder_command.rb
@@ -2,7 +2,7 @@
 
 #-- copyright
 # OpenProject is an open source project management software.
-# Copyright (C) 2012-2023 the OpenProject GmbH
+# Copyright (C) 2012-2024 the OpenProject GmbH
 #
 # This program is free software; you can redistribute it and/or
 # modify it under the terms of the GNU General Public License version 3.
@@ -54,7 +54,7 @@ module Storages
           private
 
           def handle_response(response)
-            data = ::Storages::StorageErrorData.new(source: self, payload: response.body)
+            data = ::Storages::StorageErrorData.new(source: self.class, payload: response)
 
             case response.status
             when 200..299
--- a/modules/storages/app/common/storages/peripherals/storage_interaction/one_drive/create_folder_command.rb
+++ b/modules/storages/app/common/storages/peripherals/storage_interaction/one_drive/create_folder_command.rb
@@ -56,17 +56,17 @@ module Storages
           def handle_response(response)
             data = ::Storages::StorageErrorData.new(source: self.class, payload: response)
 
-            case response.status
-            when 200..299
+            case response
+            in { status: 200..299 }
               ServiceResult.success(result: file_info_for(MultiJson.load(response.body, symbolize_keys: true)),
                                     message: 'Folder was successfully created.')
-            when 404
+            in { status: 404 }
               ServiceResult.failure(result: :not_found,
                                     errors: ::Storages::StorageError.new(code: :not_found, data:))
-            when 401
+            in { status: 401 }
               ServiceResult.failure(result: :unauthorized,
                                     errors: ::Storages::StorageError.new(code: :unauthorized, data:))
-            when 409
+            in { status: 409 }
               ServiceResult.failure(result: :already_exists,
                                     errors: ::Storages::StorageError.new(code: :conflict, data:))
             else
--- a/modules/storages/app/models/storages/storage.rb
+++ b/modules/storages/app/models/storages/storage.rb
@@ -135,9 +135,29 @@ module Storages
       end
     end
 
+    def automatically_managed?
+      ActiveSupport::Deprecation.warn(
+        '`#automatically_managed?` is deprecated. Use `#automatic_management_enabled?` instead. ' \
+        'NOTE: The new method name better reflects the actual behavior of the storage. ' \
+        "It's not the storage that is automatically managed, rather the Project (Storage) Folder is. " \
+        "A storage only has this feature enabled or disabled."
+      )
+      super
+    end
+
     def automatic_management_enabled?
       !!automatically_managed
     end
+
+    def automatic_management_unspecified?
+      automatically_managed.nil?
+    end
+
+    def automatic_management_enabled=(value)
+      self.automatically_managed = value
+    end
+
+    alias automatic_management_enabled automatically_managed
 
     def configured?
       configuration_checks.values.all?
--- a/modules/storages/app/models/storages/storage.rb
+++ b/modules/storages/app/models/storages/storage.rb
@@ -52,6 +52,8 @@ module Storages
 
     self.inheritance_column = :provider_type
 
+    store_attribute :provider_fields, :automatically_managed, :boolean
+
     has_many :file_links, class_name: 'Storages::FileLink'
     belongs_to :creator, class_name: 'User'
     has_many :project_storages, dependent: :destroy, class_name: 'Storages::ProjectStorage'
@@ -77,6 +79,8 @@ module Storages
     scope :not_enabled_for_project, ->(project) do
       where.not(id: project.project_storages.pluck(:storage_id))
     end
+
+    scope :automatic_management_enabled, -> { where("provider_fields->>'automatically_managed' = 'true'") }
 
     enum health_status: {
       pending: 'pending',
@@ -129,6 +133,10 @@ module Storages
                health_checked_at: Time.now.utc,
                health_reason: nil)
       end
+    end
+
+    def automatic_management_enabled?
+      !!automatically_managed
     end
 
     def configured?
--- a/modules/storages/app/services/storages/one_drive_managed_folder_sync_service.rb
+++ b/modules/storages/app/services/storages/one_drive_managed_folder_sync_service.rb
@@ -33,7 +33,7 @@ module Storages
     using Peripherals::ServiceResultRefinements
 
     DISALLOWED_CHARS = /[\\<>+?:"|\/]/
-    OPG_PERMISSIONS = %i[read_files write_files create_files delete_files share_files].freeze
+    OP_PERMISSIONS = %i[read_files write_files create_files delete_files share_files].freeze
 
     def self.call(storage)
       new(storage).call
@@ -92,11 +92,11 @@ module Storages
     end
 
     def add_user_to_permission_list(permissions, token, project)
-      opg_user_permissions = token.user.all_permissions_for(project)
+      op_user_permissions = token.user.all_permissions_for(project)
 
-      if opg_user_permissions.member?(:write_files)
+      if op_user_permissions.member?(:write_files)
         permissions[:write] << token.origin_user_id
-      elsif opg_user_permissions.member?(:read_files)
+      elsif op_user_permissions.member?(:read_files)
         permissions[:read] << token.origin_user_id
       end
     end
@@ -163,7 +163,7 @@ module Storages
     def project_tokens(project_storage)
       project_tokens = client_tokens_scope.where.not(id: admin_client_tokens_scope).order(:id)
 
-      if project_storage.project.public? && ProjectRole.non_member.permissions.intersect?(OPG_PERMISSIONS)
+      if project_storage.project.public? && ProjectRole.non_member.permissions.intersect?(OP_PERMISSIONS)
         project_tokens
       else
         project_tokens.where(user: project_storage.project.users)
--- a/modules/storages/app/services/storages/one_drive_managed_folder_sync_service.rb
+++ b/modules/storages/app/services/storages/one_drive_managed_folder_sync_service.rb
@@ -1,0 +1,193 @@
+# frozen_string_literal: true
+
+#-- copyright
+# OpenProject is an open source project management software.
+# Copyright (C) 2012-2024 the OpenProject GmbH
+#
+# This program is free software; you can redistribute it and/or
+# modify it under the terms of the GNU General Public License version 3.
+#
+# OpenProject is a fork of ChiliProject, which is a fork of Redmine. The copyright follows:
+# Copyright (C) 2006-2013 Jean-Philippe Lang
+# Copyright (C) 2010-2013 the ChiliProject Team
+#
+# This program is free software; you can redistribute it and/or
+# modify it under the terms of the GNU General Public License
+# as published by the Free Software Foundation; either version 2
+# of the License, or (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program; if not, write to the Free Software
+# Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+#
+# See COPYRIGHT and LICENSE files for more details.
+#++
+
+module Storages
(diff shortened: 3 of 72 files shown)


modules/storages/app/services/storages/one_drive_managed_folder_sync_service.rb
```diff
@@ -0,0 +1,193 @@
...
+                             .on_failure do |service_result|
+          format_and_log_error(service_result.errors, folder: path, context: 'hide_folder')
+        end
+      end
+    end
+
+    def add_user_to_permission_list(permissions, token, project)
+      opg_user_permissions = token.user.all_permissions_for(project)
+
+      if opg_user_permissions.member?(:write_files)
```
wielinde: @mereghost I haven't read the whole PR and just landed here. I wander what `opg` stands for. OpenProject greatness?

mereghost: `OpenProgect`? to be honest I don't remember by frame of mind at the time. `op_user_permissions` probably made me thing of `overpowered user permissions`, so I might have added the `g` there. 😛 

I suck at naming things. 🤣

modules/storages/app/common/storages/peripherals/storage_interaction/one_drive/create_folder_command.rb
```diff
@@ -54,7 +54,7 @@ def call(folder_path:)
           private
 
           def handle_response(response)
-            data = ::Storages::StorageErrorData.new(source: self, payload: response.body)
+            data = ::Storages::StorageErrorData.new(source: self.class, payload: response)
 
             case response.status
```
Kharonus: 🟡 Hmm, this must be based on an outdated state, those handlings should all now use pattern matching

modules/storages/app/models/storages/storage.rb
```diff
@@ -52,6 +52,8 @@ class Storage < ApplicationRecord
 
     self.inheritance_column = :provider_type
 
+    store_attribute :provider_fields, :automatically_managed, :boolean
```
Kharonus, line 55: 🔴 I'd vote to NOT reuse the deprecated provider field name.

instead, we can add the new name to one drive and eventually migrate nextcloud provider fields. At least, that was the approach I chose [here](https://github.com/opf/openproject/pull/14689/files#diff-8ebdbb751fc5dcb76b1f74b8ae5c4e1e4f67f935ac7a085ddc8f1c313b922706)

modules/storages/app/services/storages/one_drive_managed_folder_sync_service.rb
```diff
@@ -0,0 +1,193 @@
...
+    end
+
+    def initialize(storage)
+      @storage = storage
+    end
+
+    def call
+      return unless @storage.automatic_management_enabled?
+
+      rfm = remote_folders_map.on_failure { |failed_result| return failed_result }.result
```
Kharonus: 🟡 very minor: for easier flow when reading code, we should avoid unknown abbreviations like `rfm`. Instead, I'd choose `map` as a variable name.

modules/storages/app/services/storages/one_drive_managed_folder_sync_service.rb
```diff
@@ -0,0 +1,193 @@
...
+                                                  source: self.class, payload: response
+                                                ))
+          format_and_log_error(errors)
+          ServiceResult.failure(result: :error, errors:)
+        end
+      end
+    end
+
+    def filter_folders_from(json)
+      json.fetch(:value, []).each_with_object({}) do |item, hash|
```
Kharonus, line 152: ❓ is `each_with_object` an alias for `reduce`?

modules/storages/app/services/storages/one_drive_managed_folder_sync_service.rb
```diff
@@ -0,0 +1,193 @@
...
+      @storage = storage
+    end
+
+    def call
+      return unless @storage.automatic_management_enabled?
+
+      rfm = remote_folders_map.on_failure { |failed_result| return failed_result }.result
+
+      ensure_folders_exist(rfm).on_success { hide_inactive_folders(rfm) }
+      apply_permission_to_folders
```
Kharonus, line 52: ❓ gimme an explanation :D 

Does this line mean, `apply_permission_to_folders` is executed, even if `ensure_folders_exist` or `hide_inactive_folders` fails?

modules/storages/app/services/storages/one_drive_managed_folder_sync_service.rb
```diff
@@ -0,0 +1,193 @@
...
+    end
+
+    private
+
+    def apply_permission_to_folders
+      active_project_storages_scope.includes(:project).where.not(project_folder_id: nil).find_each do |project_storage|
+        permissions = { read: [], write: admin_client_tokens_scope.pluck(:origin_user_id) }
```
Kharonus, line 59: 🟡 Why having the custom handling for OP instance admins?

Shouldn't those tokens just do fine if running through `add_user_to_permission_list`?

Kharonus (changes requested):
Putting it to "request changes", even with me not seeing a big blocker. 

but there are a few points, I'd at least like to discuss first.

modules/storages/app/services/storages/one_drive_managed_folder_sync_service.rb
mereghost: I just renamed it locally =)

modules/storages/app/services/storages/one_drive_managed_folder_sync_service.rb
mereghost: Not really. `each_with_object` is similar to `reduce` but different in that you explicitly pass in the accumulator object and it only mutates it if you do so.

Example:

In `reduce` (or `inject`) the last expression of the block, becomes the accumulator. So without the call to `acc` there, the next iteration would have been `10` instead of `{ 1 => 1 }`. IIRC there was some memory usage implication, but it might have changed.

```ruby
[1,2,3,4,5,6].reduce({}) { |acc, i| acc[i] = i ** i ;  acc }
```

`each_with_object` iterates through the collection passing the accumulator object back to the block each iteration. This means that if you don't make changes to the accumulator, nothing will happen as it ignores the result of the block, always returning the accumulator object.

```ruby
[1,2,3,4,5,6].each_with_object({}) {|i, acc| acc[i] = i ** i }
```

In general, unless I'm reducing to a single scalar, I prefer `each_with_object` as it is clearer on what is happening.

modules/storages/app/services/storages/one_drive_managed_folder_sync_service.rb
mereghost: This comes from the Nextcloud behaviour, but mainly it is in order to make sure that the admins always have access to the data.

It makes sense, as they might need to check things even not being part of the projects.

modules/storages/app/models/storages/storage.rb
mereghost: Hmmm... that would break a lot of places that check for that specific flag. If we are going to make this change, it should be uniform across all storages.

IF we decide to go down this route, I'd rather have another PR to make the change and required migrations.

@akabiru you dabbled into the renaming of this before, what are your thoughts about this?

modules/storages/app/common/storages/peripherals/storage_interaction/one_drive/create_folder_command.rb
mereghost: Good catch. Those commands where merged in between those changes and missed this. Updating!

modules/storages/app/services/storages/one_drive_managed_folder_sync_service.rb
mereghost: Explanation: Darth Vader told me to.

In all seriousness, reasoning is simple: applying permissions is completely unrelated to folder existing and being hidden.

You will probably ask "Why? How?", right? It maps to the old flowchart I made back when refactoring the Nextcloud version of this code.

`ensure_folders_exist` actually does 2 different things: 

1. create project folders, if they don't have one created yet
2. rename a project folder if some smartass changed it.

The former ends up updating the DB with the `origin_folder_id`, the latter is irrelevant to the API interactions as we don't care about the name, only using IDs, so you can say it is a "cosmetic change".

`hide_inactive_folders` removes all permissions from the folder, effectively hiding it from sight. It only happens to inactive projects with AMPF enabled, if it fails the only downside is that it is not hidden/still accessible.

When it comes to set permissions, if there's no `origin_folder_id`, the project will be ignored (it excludes them on the query). The ones with messed up remote names, are still valid, so we try anyway.

modules/storages/app/models/storages/storage.rb
Kharonus: we do not have a single place, that checks `automatically_managed?` anymore. @akabiru confirmed that to me yesterday. 

regarding migrations, if you take a look at my PR, I'd say, we easily can spare some rework, if we just introduce the new(!) provider field on one drive storage already with final name.

modules/storages/app/services/storages/one_drive_managed_folder_sync_service.rb
Kharonus: Ah, brilliant explanation. then it boils down to a matter of choice. I'd always prefer reduce, as it is know over a variety of languages and hence easy to understand. The fact, that each_with_object does exactly the same, just without the need of returning the accumulator, smells a bit like "meh, I do not want to write that last line." and of course, it is very specific (make `reduce` with an object accumulator), while reduce can have any kind of accumulator.

Yet, I am fine with the usage, from reading the code, it is pretty obvious what it does.

modules/storages/app/services/storages/one_drive_managed_folder_sync_service.rb
Kharonus: me understands

modules/storages/app/services/storages/one_drive_managed_folder_sync_service.rb
Kharonus: Maybe on second thought, I pretty not satisfied with the fact, that `each_with_object` is on design side effect based. While `reduce` explicitly displays in code, what happens with the aggregated object.

modules/storages/app/models/storages/storage.rb
mereghost: We are using an alias right now (that I forgot to move over), which I'm ok with adding here.

A full blown change, like the one you added on your PR, would warrant a migration over all storages, noting that this setting isn't provider specific anymore.

modules/storages/app/services/storages/one_drive_managed_folder_sync_service.rb
mereghost: I just checked and the object allocation difference persists: `reduce` allocates a new object at every interaction, while `each_with_object` reuses the accumulator.

Minor difference but significant specially for array => other collection transformations.

modules/storages/app/models/storages/storage.rb
mereghost: Now that I actually stopped to think about it, this probably should be promoted to a full blown field on the `storages` table. A PG enum field would be perfect with the possible values of `enabled`, `disabled` and `not_configured` or something along those lines.

Then the `Storage` types can implement special behaviour over it.

Kharonus: > A full blown change, like the one you added on your PR, would warrant a migration over all storages, noting that this setting isn't provider specific anymore.

no, why? the change there is only a new provider field for one drive. the public interface of NC and OD storages is the same.

Kharonus: I agree, that this now exceeds the context of `provider_fields`. But, I think this indeed should be another, specific PR.

Very interested in @akabiru opinion here. And then we can agree on a way forward.

modules/storages/app/services/storages/one_drive_managed_folder_sync_service.rb
Kharonus: yes, that's the very issue of (im)mutability. `reduce` is working on immutable objects, while `each_with_object` mutates it.

modules/storages/app/models/storages/storage.rb
akabiru: @mereghost 👋🏾 I agree with the idea to promote `automatic_management` to a PG enum- and that this fits a separate CODE Maintenance work package. I'm happy to spec this out for us, and perhaps it's an item we can pair on during our meet (as you proposed @Kharonus)

Kharonus: as written in the chat, I assume it is save to accept the current implementation of Marcello here? dragging the "delegate" of the provider field to storage base model. And then we opt for migrating it to a column next week.

Kharonus (approved):

PR #14656, opened 2024-01-26
[52384] Fix project copy failing when there is a child having dates and work estimates

cbliard:
https://community.openproject.org/projects/openproject/work_packages/52384


spec/support/table_helpers/column.rb
```diff
@@ -52,6 +52,8 @@ def self.for(header)
         Duration.new(header:, attribute: :remaining_hours)
       when /derived remaining work/i
         Duration.new(header:, attribute: :derived_remaining_hours)
+      when /end date/i
+        Generic.new(header:, attribute: :due_date)
```
machisuji, line 56: 🐝

machisuji (approved):
👍

PR #14680, opened 2024-01-31
[52252] Exclude derived_remaining_time from work package form

cbliard:
https://community.openproject.org/work_packages/52252

`Derived Remaining Time` attribute was accidentaly added to the work package form when it was introduced in #13401. This PR removes it (actually, it hides it).

It includes the number of `EXCLUDED` attributes in the cache key for the work package form attributes. This ensures that the cache is invalidated when the number of excluded attributes changes. Without it, one may still see the same attribute groups.

### DIFF
--- a/app/models/type/attributes.rb
+++ b/app/models/type/attributes.rb
@@ -35,6 +35,7 @@ module Type::Attributes
                 derived_start_date
                 derived_due_date
                 derived_estimated_time
+                derived_remaining_time
                 ignore_non_working_days
                 duration
                 description
@@ -80,6 +81,7 @@ module Type::Attributes
 
       OpenProject::Cache.fetch('all_work_package_form_attributes',
                                *wp_cf_cache_parts,
+                               EXCLUDED.length,
                                merge_date) do
         calculate_all_work_package_form_attributes(merge_date)
       end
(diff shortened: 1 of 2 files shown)


app/models/type/attributes.rb
```diff
@@ -80,6 +81,7 @@ def all_work_package_form_attributes(merge_date: false)
 
       OpenProject::Cache.fetch('all_work_package_form_attributes',
                                *wp_cf_cache_parts,
+                               EXCLUDED.count,
```
aaron-contreras: https://github.com/fastruby/fast-ruby?tab=readme-ov-file#arraylength-vs-arraysize-vs-arraycount-code

```suggestion
                               EXCLUDED.length,
```

or if feeling idiomatic:

```suggestion
                               EXCLUDED.size,
```

cbliard: Well, I'll change it, but it's not really impactful here.

aaron-contreras: I've been looking to add cops for some of these fastruby findings. Marked it to start as of now :)

cbliard: Yeah I have a love/hate relation with fastruby advices.
Sometimes it really help, like advising `list << item` instead of `list += item` in a loop in some code being used a lot. That's a huge saving.
And also it helps knowing the most efficient ways to write faster ruby code.
And sometimes it's just annoying because it keeps suggesting low-level enhancements that are largely neglectable compared to a bad N+1 issue.

PR #14689, opened 2024-01-31
[#52525] Change storage configuration form

Kharonus:
[#52525](https://community.openproject.org/work_packages/52525)

### WAT?

- Added the AMPF switch to the general information section when creating one drive storages
- switch is enabled per default
  - `PROVIDER_FIELD_DEFAULTS` enabled for all storage types


mereghost:
I'm too lazy to re-type here, but the whole thread over [the sync service PR](https://github.com/opf/openproject/pull/14648#discussion_r1474101523) applies here.

TL;DR: After some back and forth I think we should move the `automatically_managed_enabled` provider field to a proper field on the `storages` table as it is not provider based, but a feature of OpenProject.

Storage types than can implement special behaviour over it, but keeping the main API in `Storages::Storage`

Kharonus:
In general we should decide on either way in your PR, and then I adapt this PR here according to it.

Kharonus:
Updated accordingly to merged changes from @mereghost to provider fields.

modules/storages/app/models/storages/one_drive_storage.rb
```diff
@@ -54,5 +58,13 @@ def uri
     def connect_src
       %w[https://*.sharepoint.com https://*.up.1drv.com]
     end
+
+    def automatic_management_unspecified?
+      automatic_management_enabled.nil?
+    end
```
akabiru: 🟠 Should be Ok to move this method up the chain?

akabiru (approved):
Nice one 👍🏾  One remark

modules/storages/app/models/storages/one_drive_storage.rb
Kharonus: True, will do in a bit

PR #14696, opened 2024-02-01
Fix autocompleter query for WorkPackages relation setup so it only returns visible columns

klaustopher:
We discovered, that when entering a work package ID or even a title in the auto completer for i.e. setting the work package's parent, it would also list work packages that you do not have access to. It does not disclose any information about the work package itself apart from its ID, but it would then allow assigning that work package as a parent and if for example using time tracking this would edit certain information (i.e. start- and end-date and duration) on that parent.

This PR
- [x] identifies why the not visible WorkPackages are even shown
- [x] fix the issue and not display those WorkPackages anymore
- [x] throw an error when trying to assign a work package as a parent, when you do not really have access to it 

---

Fixes https://community.openproject.org/work_packages/52556

### DIFF
--- a/app/models/query/results.rb
+++ b/app/models/query/results.rb
@@ -50,8 +50,8 @@ class ::Query::Results
 
   def sorted_work_packages_matching_the_filters_today
     sorted_work_packages
-      .visible
       .merge(filtered_work_packages.merge(filter_merges))
+      .visible
   end
 
   # For filtering on historic data, this returns the work packages
--- a/spec/requests/api/v3/work_packages/available_relation_candidates_resource_spec.rb
+++ b/spec/requests/api/v3/work_packages/available_relation_candidates_resource_spec.rb
@@ -62,7 +62,7 @@ RSpec.describe API::V3::WorkPackages::AvailableRelationCandidatesAPI do
     request
     JSON.parse last_response.body
   end
-  let(:subjects) { work_packages.pluck("subject") }
+  let(:subjects) { work_packages.pluck("id") }
 
   def work_packages
     result["_embedded"]["elements"]
@@ -82,7 +82,7 @@ RSpec.describe API::V3::WorkPackages::AvailableRelationCandidatesAPI do
           with_settings: { cross_project_work_package_relations: false } do
     describe "relation candidates for wp1 (in hierarchy)" do
       it "returns an empty list" do
-        expect(subjects).to contain_exactly("WP 1.2.1")
+        expect(subjects).to contain_exactly(wp1_2_1.id)
       end
     end
 
@@ -90,7 +90,7 @@ RSpec.describe API::V3::WorkPackages::AvailableRelationCandidatesAPI do
       let(:href) { "/api/v3/work_packages/#{wp2.id}/available_relation_candidates?query=WP" }
 
       it "returns WP 2.1 and 2.2" do
-        expect(subjects).to contain_exactly("WP 2.1", "WP 2.2")
+        expect(subjects).to contain_exactly(wp2_1.id, wp2_2.id)
       end
     end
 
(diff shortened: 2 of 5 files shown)


spec/requests/api/v3/work_packages/available_relation_candidates_resource_spec.rb
```diff
@@ -197,10 +197,17 @@ def work_packages
       end
 
       it 'does not return work packages from that project' do
-        # TODO: Remove and fix before merge
-        pending "Temporary disable"
```
klaustopher, line 201: This test was broken in 13.1 and we disabled it for now, it sort of was testing exactly for this behavior. But because `subjects` was just filled with `[nil, nil]` entries we have temporarily disabled the tests.

I have changed the entire test file to work with the IDs instead of the names, this would have given us a hint what the `nil` values were and we might have figured this out sooner

app/models/query/results.rb
```diff
@@ -50,8 +50,8 @@ def work_packages
 
   def sorted_work_packages_matching_the_filters_today
     sorted_work_packages
-      .visible
       .merge(filtered_work_packages.merge(filter_merges))
+      .visible
```
aaron-contreras, line 54: Interesting... `merge` was overriding filter conditions set by `visible`?

spec/requests/api/v3/work_packages/available_relation_candidates_resource_spec.rb
```diff
@@ -82,31 +82,31 @@ def work_packages
           with_settings: { cross_project_work_package_relations: false } do
     describe "relation candidates for wp1 (in hierarchy)" do
       it "returns an empty list" do
```
aaron-contreras:
```suggestion
      it "returns WP 1.2.1" do
```

spec/requests/api/v3/work_packages/available_relation_candidates_resource_spec.rb
aaron-contreras: Nice.

aaron-contreras (approved):
Great find ⭐️

Just a test description update to match the expectation.

app/models/query/results.rb
klaustopher: It does when you add a where condition for the same field. Before the move to WP based permissions, the `visible` scope added a where on the `project_id` field. The move to WP based permissions moved the `where` condition to `id`. But the `filter_merges` query already has a bunch of where conditions on the `id` field, so it overrides those when merging.

In the [docs for #merge](https://api.rubyonrails.org/classes/ActiveRecord/SpawnMethods.html#method-i-merge) it warns about this:

> For conditions that exist in both relations, those from other will take precedence. To find the intersection of two relations, use `QueryMethods#and`.


```ruby
[1] pry(main)> WorkPackage.where(id: [1,2,3]).merge(WorkPackage.where(id: 5)).to_sql
=> "SELECT \"work_packages\".* FROM \"work_packages\" WHERE \"work_packages\".\"id\" = 5"

[2] pry(main)> WorkPackage.where(id: [1,2,3]).merge(WorkPackage.where(project_id: 1)).to_sql
=> "SELECT \"work_packages\".* FROM \"work_packages\" WHERE \"work_packages\".\"id\" IN (1, 2, 3) AND \"work_packages\".\"project_id\" = 1"

[3] pry(main)> WorkPackage.where(id: [1,2,3]).and(WorkPackage.where(id: 5)).to_sql
=> "SELECT \"work_packages\".* FROM \"work_packages\" WHERE \"work_packages\".\"id\" IN (1, 2, 3) AND \"work_packages\".\"id\" = 5"
```

aaron-contreras: Great debugging. Thank you

machisuji (approved):
👍

PR #14701, opened 2024-02-01
fix and cleanup backup rake tasks

toy:
`ActiveRecord::Base.configurations` [does not return a hash since rails 6](https://github.com/rails/rails/pull/33637), switch to using `ActiveRecord::Base.connection_db_config` and cleanup code

### DIFF
--- a/lib/tasks/backup.rake
+++ b/lib/tasks/backup.rake
@@ -44,13 +44,12 @@ namespace :backup do
                         "--file=#{args[:path_to_backup]}",
                         '--format=custom',
                         '--no-owner']
-        pg_dump_call << "--host=#{config['host']}" if config['host']
-        pg_dump_call << "--port=#{config['port']}" if config['port']
-        user = config.values_at('user', 'username').compact.first
-        pg_dump_call << "--username=#{user}" if user
-        pg_dump_call << config['database'].to_s
+        pg_dump_call << "--host=#{config[:host]}" if config[:host]
+        pg_dump_call << "--port=#{config[:port]}" if config[:port]
+        pg_dump_call << "--username=#{config[:user]}" if config[:user]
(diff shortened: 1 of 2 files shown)


machisuji:
@toy Thanks for fixing the rake tasks. Looks good in principle. But may I suggest re-using the logic from the [backup job](https://github.com/opf/openproject/blob/dev/app/workers/backup_job.rb#L262-L290) for the postgres configuration including password? This uses [env vars](https://github.com/opf/openproject/blob/dev/app/workers/backup_job.rb#L242) which is less of a hassle than creating a temp file too. And this can be both reused for pg_dump and pg_restore saving us a bunch of lines.

toy:
@machisuji Is passing password using environment variable instead of a file not a concern?

machisuji:
> @machisuji Is passing password using environment variable instead of a file not a concern?

I would argue that env is more secure than the file in this case. The variable will only be set for the spawned process and there will be no trace of it otherwise. In the text file approach it will literally be in the tmp directory for everyone to read.

toy:
File will have 600 permissions, so only readable to this user and root. Postgres documentation [suggests to use a temp file](https://www.postgresql.org/docs/current/libpq-envars.html#:~:text=PGPASSWORD), as env can be visible to other users in some systems/configurations

machisuji:
As far as I know this concern referred to in the postgres docs is outdated. There was a time where you could see other processes' env vars via `ps axe` but unless your system hasn't been updated in the past 5 or so years that should not be the case any longer.

But to be honest, it doesn't really matter what we do here as this is a simple rake task for local development.
So no need to worry too much about these kinds of details here.

I was more concerned with the verbosity and repetition of the code. But again, this is just a development rake task.
So what's important here is that you fixed it. Thanks for the fix. I'll merge it as it is.

PR #14705, opened 2024-02-02
Use `parallel_split_test` for CI feature spec runs

aaron-contreras:
The `parallel_split_test` gem allows parallelizing on an example-by-example basis instead of a file-by-file basis. This removes the un-evenness in overall group runtime because we get a much more even split among all 32-cores on the CI.

**Notes:**
I decided not running unit specs with `parallel_split_test` as the
penalty for executing `before(:all)` hooks before every example is way too
noticeable in the overall execution time.

This however, is a negligible amount in the overall execution time for
feature specs. Being able to not pay a penalty for having a really long
feature spec file is a plus as this can help better balance the total
CI time.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -282,6 +282,7 @@ group :test do
   gem 'shoulda-matchers', '~> 6.0', require: nil
 
   gem 'parallel_tests', '~> 4.0'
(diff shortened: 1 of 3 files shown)


aaron-contreras:
Execution time was found to be about the same.

PR #14706, opened 2024-02-02
[#52526] update project storage form to handle automatic management enabled

Kharonus:
[#52526](https://community.openproject.org/wp/52526)

### WAT?

- show the decision of automatically managed project folders when adding a project storage of a one drive storage
  - only if the storage was configured as `automatical_management_enabled`


akabiru (approved):
:shipit:

PR #14440, opened 2023-12-18
Code maintenance/46647 bump rails to 7.1

dombesz:
https://community.openproject.org/wp/46647

PR #14711, opened 2024-02-05
Rewrite allowed_to scope for permitted work packages using CTE

klaustopher:
The query to retrieve allowed work packages is a bottleneck, when a project has a very very high amount of work packages. This was created when we added the permission checks for work packages, because a part of the subquery also returned all work packages the user had access to via a project and that could potentially yield a very high amount. 

Since we recently switched to Rails 7.1. (#14440) we got access to the `with` function to quickly introduce CTEs into queries. This PR takes advantage of this and selects the permitted projects and single work packages via a CTE and then simply queries that CTE.
This model could also be used in the future for other queries, as we can for example do filtering in a CTE and then query that result. It could simplify some of our query construction. Interesting to investigate for the future.

---
Total number of work packages: 2.192.975
Number of work packages in project: 1.152.981
Number of work packages in parent project: 20.010

| Query | Runtime **OLD** | Runtime **NEW** |
|-------|-----------------|-----------------|
| `COUNT(DISTINCT)` | 4190.5ms | 434.8ms |
| `SELECT id LIMIT 100`| 4325.4ms |  45.5ms |
| `SELECT` with all related things | 2755.2ms | 151.5ms |

---

Closes https://community.openproject.org/work_packages/50850
Closes https://community.openproject.org/work_packages/52022
Closes https://community.openproject.org/work_packages/52156

### DIFF
--- a/app/models/work_packages/scopes/allowed_to.rb
+++ b/app/models/work_packages/scopes/allowed_to.rb
@@ -46,11 +46,12 @@ module WorkPackages::Scopes
           where(project_id: Project.allowed_to(user, permissions))
         else
           allowed_via_wp_membership = allowed_to_member_relation(user, permissions).select(arel_table[:id]).arel
-          allowed_via_project_membership = unscoped.where(project_id: Project.unscoped.allowed_to(user, permissions).select(:id))
-            .select(arel_table[:id])
-            .arel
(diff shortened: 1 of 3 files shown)


ulferts (approved):

ulferts:
The failing specs are unrelated and will be fixed separately.

PR #14714, opened 2024-02-05
Update copyright years

dombesz:
Update some leftover occurrences of copyright year from 2023 to 2024.


aaron-contreras (approved):
Thank you!

I think these weren't picked up by Christophe's script since they were being worked on at about the same time as that PR was merged which did the New Year's update. It probably resolved to taking the latest merged branch as the changeset that was valid.

PR #14716, opened 2024-02-05
[52012] Remove colour in share modal is not correct

bsatarnejad:
https://community.openproject.org/projects/openproject/work_packages/52012/activity?query_id=2865


aaron-contreras (approved):
👍

PR #14718, opened 2024-02-05
[#52119] Make renamed progress attributes searchable with old names

aaron-contreras:
Allows alternative matches for both the filters autocompleter and the draggable autocompleter used in the work package view configuration form.

See: https://community.openproject.org/work_packages/52119


frontend/src/app/shared/components/work-packages/alternative-search.service.ts
```diff
@@ -0,0 +1,43 @@
...
+
+  public searchFunction = (term:string, currentItem:QueryFilterResource):boolean => {
+    const lowercaseSearchTerm = term.toLowerCase();
+    const lowercaseCurrentItemName = currentItem.name.toLowerCase();
+
+    const alternativeMatch = Object
+      .keys(this.alternativeNames)
+      .some((alternativeName) => {
+        return alternativeName.toLowerCase().indexOf(lowercaseSearchTerm) > -1
+          && currentItem.name === this.alternativeNames[alternativeName];
```
cbliard: `alternativeName.toLowerCase().includes(lowercaseSearchTerm)` reads better than `indexOf() > -1`.

frontend/src/app/shared/components/work-packages/alternative-search.service.ts
```diff
@@ -0,0 +1,43 @@
...
+      });
+
+    return (
+      lowercaseCurrentItemName.indexOf(lowercaseSearchTerm) > -1
```
cbliard: same

config/locales/js-en.yml
```diff
@@ -1043,13 +1043,17 @@ en:
         createdAt: "Created on"
         description: "Description"
         date: "Date"
+        done_ratio: "% Complete"
+        done_ratio_alternative: "Progress"
```
cbliard: As you named the other 2 fields after their new name, can you also name `done_ratio[_alternative]` as `percent_complete[_alternative]`?

cbliard: Also I'm not sure if mixing cases is ok there. Shouldn't it be snakeCase like others?

cbliard (approved):
Nice! I especially like the expressiveness of the tests.

I only have minor remarks. Feel free to merge at your will.

config/locales/js-en.yml
aaron-contreras: I saw snake case and camel case being used alternatively so didn't think we enforce a style. I do know camel is the commonly accepted one so I'll do so :)

cbliard: Yeah, I'm not familiar with frontend so I suppose camelCase is the way to go there.

PR #14721, opened 2024-02-05
Feature/50266 conditionally remove shares of removed project members

toy:
Script to generate possible members (first run `bin/rails backup:database:create'[dump/initial]'` to create snapsot, then `bin/rails runner PATH_TO_FILE` to generate):
```ruby
Kernel.system(OpenProject::PostgresEnvironment.pg_env, *%W[
  pg_restore
  --clean
  --no-owner
  --single-transaction
  --dbname=#{OpenProject::PostgresEnvironment.pg_env['PGDATABASE']}
  dump/initial
]) if File.exist?('dump/initial')

include FactoryBot::Syntax::Methods

admin = User.find_by(login: 'admin')

project = create(:project)
project_role_direct = create(:project_role, name: 'direct')
project_role_inherited = create(:project_role, name: 'inherited')
view_work_package_role = WorkPackageRole.find_by(name: 'Work package viewer') || create(:view_work_package_role)
comment_work_package_role = WorkPackageRole.find_by(name: 'Work package commenter') || create(:comment_work_package_role)
edit_work_package_role = WorkPackageRole.find_by(name: 'Work package editor') || create(:edit_work_package_role)

name_sequence = ('AA'..).to_enum

[true, false].repeated_permutation(4) do |direct_project_role, inherited_project_role, direct_shared_wps, inherited_shared_wps|
  next unless direct_project_role || inherited_project_role || direct_shared_wps || inherited_shared_wps

  name = name_sequence.next

  user = create(:user, firstname: "#{name}", lastname: [
    direct_project_role ? 'DirRole' : nil,
    inherited_project_role ? 'GroupRole' : nil,
    direct_shared_wps ? 'DirectShared' : nil,
    inherited_shared_wps ? 'GroupShared' : nil,
  ].compact.join(' '))

  if inherited_project_role || inherited_shared_wps
    group = create(:group, lastname: "G_#{name}", members: [user])
  end

  if direct_project_role
    create(:member, principal: user, project:, roles: [project_role_direct])
  end

  if inherited_project_role
    create(:member, principal: group, project:, roles: [project_role_inherited])
  end

  if direct_shared_wps || inherited_shared_wps
    work_package_both = create(:work_package, project:)
  end

  if direct_shared_wps
    work_package_direct = create(:work_package, project:)

    create(:member, principal: user, project:, roles: [view_work_package_role], entity: work_package_direct)
    create(:member, principal: user, project:, roles: [edit_work_package_role], entity: work_package_both)
  end

  if inherited_shared_wps
    work_package_inherited = create(:work_package, project:)

    create(:member, principal: group, project:, roles: [comment_work_package_role], entity: work_package_inherited)
    create(:member, principal: group, project:, roles: [edit_work_package_role], entity: work_package_both)
  end

  if inherited_project_role || inherited_shared_wps
    Groups::CreateInheritedRolesService
      .new(group, current_user: create(:admin))
      .call(user_ids: group.user_ids,
            send_notifications: false,
            project_ids: [project.id])
  end
end
```

------

https://community.openproject.org/wp/50266

### DIFF
--- a/app/components/members/delete_member_dialog_component.html.erb
+++ b/app/components/members/delete_member_dialog_component.html.erb
@@ -33,17 +33,17 @@ See COPYRIGHT and LICENSE files for more details.
   <% if can_delete_roles? && can_delete_shares? %>
     <% if can_delete? %>
       <% dialog.with_body do %>
-        <%= render(Primer::Beta::Text.new(tag: "p")) { scoped_t(:will_remove_the_users_role_html, shared_work_packages_link:) } %>
-        <%= render(Primer::Beta::Text.new(tag: "p")) do %>
+        <%= paragraph { scoped_t(:will_remove_the_users_role_html, shared_work_packages_link:) } %>
+        <%= paragraph do %>
           <%= scoped_t(:would_you_like_to_remove_the_shares_too) %>
           <%= scoped_t(:will_not_affect_inherited_shares) if inherited_shared_work_packages_count? %>
         <% end %>
       <% end %>
     <% else %>
       <% dialog.with_body do %>
-        <%= render(Primer::Beta::Text.new(tag: "p")) { scoped_t(:can_remove_direct_but_not_shared_role) } %>
-        <%= render(Primer::Beta::Text.new(tag: "p")) { scoped_t(:also_user_has_shares_html, shared_work_packages_link:) } %>
-        <%= render(Primer::Beta::Text.new(tag: "p")) do %>
+        <%= paragraph { scoped_t(:can_remove_direct_but_not_shared_role) } %>
+        <%= paragraph { scoped_t(:also_user_has_shares_html, shared_work_packages_link:) } %>
+        <%= paragraph do %>
           <%= scoped_t(:remove_just_user_or_shares_too) %>
           <%= scoped_t(:will_not_affect_inherited_shares) if inherited_shared_work_packages_count? %>
         <% end %>
@@ -51,34 +51,34 @@ See COPYRIGHT and LICENSE files for more details.
     <% end %>
 
     <% dialog.with_footer do %>
-      <%= render(Primer::Beta::Button.new(data: { close_dialog_id: id })) { t(:button_cancel) } %>
-      <%= render(Primer::Beta::Button.new(scheme: :danger, tag: :a, href: delete_url(project: true), data: { method: :delete })) { t(:button_remove_member) } %>
-      <%= render(Primer::Beta::Button.new(scheme: :danger, tag: :a, href: delete_url(project: true, work_package_shares_role_id: "all"), data: { method: :delete })) { t(:button_remove_member_and_shares) } %>
+      <%= button(data: { close_dialog_id: id }) { t(:button_cancel) } %>
+      <%= button(scheme: :danger, tag: :a, href: delete_url(project: true), data: { method: :delete }) { t(:button_remove_member) } %>
+      <%= button(scheme: :danger, tag: :a, href: delete_url(project: true, work_package_shares_role_id: "all"), data: { method: :delete }) { t(:button_remove_member_and_shares) } %>
     <% end %>
   <% else %>
     <% if can_delete_roles? %>
       <% if can_delete? %>
         <% dialog.with_body do %>
-          <%= render(Primer::Beta::Text.new(tag: "p")) { scoped_t(:will_remove_all_access_priveleges) } %>
+          <%= paragraph { scoped_t(:will_remove_all_access_priveleges) } %>
         <% end %>
       <% else %>
         <% dialog.with_body do %>
-          <%= render(Primer::Beta::Text.new(tag: "p")) { scoped_t(:can_remove_direct_but_not_shared_role) } %>
+          <%= paragraph { scoped_t(:can_remove_direct_but_not_shared_role) } %>
         <% end %>
       <% end %>
 
       <% dialog.with_footer do %>
-        <%= render(Primer::Beta::Button.new(data: { close_dialog_id: id })) { t(:button_cancel) } %>
-        <%= render(Primer::Beta::Button.new(scheme: :danger, tag: :a, href: delete_url(project: true), data: { method: :delete })) { t(:button_remove) } %>
+        <%= button(data: { close_dialog_id: id }) { t(:button_cancel) } %>
+        <%= button(scheme: :danger, tag: :a, href: delete_url(project: true), data: { method: :delete }) { t(:button_remove) } %>
       <% end %>
     <% else %>
       <% dialog.with_body do %>
-        <%= render(Primer::Beta::Text.new(tag: "p")) { scoped_t(:cannot_delete_group_member) } %>
-        <%= render(Primer::Beta::Text.new(tag: "p")) { scoped_t(:cannot_delete_group_member_note_html, administration_settings_link:) } %>
+        <%= paragraph { scoped_t(:cannot_delete_group_member) } %>
+        <%= paragraph { scoped_t(:cannot_delete_group_member_note_html, administration_settings_link:) } %>
       <% end %>
 
       <% dialog.with_footer do %>
-        <%= render(Primer::Beta::Button.new(data: { close_dialog_id: id })) { t(:button_cancel) } %>
+        <%= button(data: { close_dialog_id: id }) { t(:button_cancel) } %>
       <% end %>
     <% end %>
   <% end %>
--- a/app/components/members/delete_member_dialog_component.html.erb
+++ b/app/components/members/delete_member_dialog_component.html.erb
@@ -1,0 +1,85 @@
+<%#-- copyright
+OpenProject is an open source project management software.
+Copyright (C) 2012-2024 the OpenProject GmbH
+
+This program is free software; you can redistribute it and/or
+modify it under the terms of the GNU General Public License version 3.
+
+OpenProject is a fork of ChiliProject, which is a fork of Redmine. The copyright follows:
+Copyright (C) 2006-2013 Jean-Philippe Lang
+Copyright (C) 2010-2013 the ChiliProject Team
+
+This program is free software; you can redistribute it and/or
+modify it under the terms of the GNU General Public License
+as published by the Free Software Foundation; either version 2
+of the License, or (at your option) any later version.
+
+This program is distributed in the hope that it will be useful,
+but WITHOUT ANY WARRANTY; without even the implied warranty of
+MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+GNU General Public License for more details.
+
+You should have received a copy of the GNU General Public License
+along with this program; if not, write to the Free Software
+Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+
+See COPYRIGHT and LICENSE files for more details.
+
+++#%>
+
+<%= render(Primer::Alpha::Dialog.new(id:, title: scoped_t(:title), size: :large)) do |dialog| %>
+  <% dialog.with_header(variant: :large) %>
+
+  <% if can_delete_roles? && can_delete_shares? %>
+    <% if can_delete? %>
+      <% dialog.with_body do %>
+        <%= render(Primer::Beta::Text.new(tag: "p")) { scoped_t(:will_remove_the_users_role_html, shared_work_packages_link:) } %>
+        <%= render(Primer::Beta::Text.new(tag: "p")) do %>
+          <%= scoped_t(:would_you_like_to_remove_the_shares_too) %>
+          <%= scoped_t(:will_not_affect_inherited_shares) if inherited_shared_work_packages_count? %>
+        <% end %>
+      <% end %>
+    <% else %>
+      <% dialog.with_body do %>
+        <%= render(Primer::Beta::Text.new(tag: "p")) { scoped_t(:can_remove_direct_but_not_shared_role) } %>
+        <%= render(Primer::Beta::Text.new(tag: "p")) { scoped_t(:also_user_has_shares_html, shared_work_packages_link:) } %>
+        <%= render(Primer::Beta::Text.new(tag: "p")) do %>
+          <%= scoped_t(:remove_just_user_or_shares_too) %>
+          <%= scoped_t(:will_not_affect_inherited_shares) if inherited_shared_work_packages_count? %>
+        <% end %>
+      <% end %>
+    <% end %>
+
+    <% dialog.with_footer do %>
+      <%= render(Primer::Beta::Button.new(data: { close_dialog_id: id })) { t(:button_cancel) } %>
+      <%= render(Primer::Beta::Button.new(scheme: :danger, tag: :a, href: delete_url(project: true), data: { method: :delete })) { t(:button_remove_member) } %>
+      <%= render(Primer::Beta::Button.new(scheme: :danger, tag: :a, href: delete_url(project: true, work_package_shares_role_id: "all"), data: { method: :delete })) { t(:button_remove_member_and_shares) } %>
+    <% end %>
+  <% else %>
+    <% if can_delete_roles? %>
+      <% if can_delete? %>
+        <% dialog.with_body do %>
+          <%= render(Primer::Beta::Text.new(tag: "p")) { scoped_t(:will_remove_all_access_priveleges) } %>
+        <% end %>
+      <% else %>
+        <% dialog.with_body do %>
+          <%= render(Primer::Beta::Text.new(tag: "p")) { scoped_t(:can_remove_direct_but_not_shared_role) } %>
+        <% end %>
+      <% end %>
+
+      <% dialog.with_footer do %>
+        <%= render(Primer::Beta::Button.new(data: { close_dialog_id: id })) { t(:button_cancel) } %>
+        <%= render(Primer::Beta::Button.new(scheme: :danger, tag: :a, href: delete_url(project: true), data: { method: :delete })) { t(:button_remove) } %>
+      <% end %>
+    <% else %>
+      <% dialog.with_body do %>
+        <%= render(Primer::Beta::Text.new(tag: "p")) { scoped_t(:cannot_delete_group_member) } %>
+        <%= render(Primer::Beta::Text.new(tag: "p")) { scoped_t(:cannot_delete_group_member_note_html, administration_settings_link:) } %>
+      <% end %>
+
+      <% dialog.with_footer do %>
+        <%= render(Primer::Beta::Button.new(data: { close_dialog_id: id })) { t(:button_cancel) } %>
+      <% end %>
+    <% end %>
+  <% end %>
+<% end %>
--- a/app/components/members/delete_member_dialog_component.html.erb
+++ b/app/components/members/delete_member_dialog_component.html.erb
@@ -36,7 +36,7 @@ See COPYRIGHT and LICENSE files for more details.
       <% if can_delete? %>
         <%= paragraph do %>
           <%= scoped_t(:will_remove_the_users_role) %>
-          <%= scoped_t(:however_shares_with_user_html, shared_work_packages_link:) %>
+          <%= scoped_t(:however_shares_with_user_html, shared_work_packages_link:, count: shared_work_packages_count) %>
         <% end %>
         <%= paragraph do %>
           <%= scoped_t(:would_you_like_to_remove_the_shares_too) %>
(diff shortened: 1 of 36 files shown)


app/components/members/row_component.rb
```diff
@@ -152,6 +152,11 @@ def edit_link
       )
     end
 
+    def edit_icon
+      render(Primer::Beta::Octicon.new(icon: :pencil))
+      # render(Primer::Beta::IconButton.new(size: :small, icon: :pencil, "aria-label": "Edit roles"))
+    end
+
```
aaron-contreras: Compliment on your "table of contents" ordering of methods

app/components/members/delete_dialog_component.html.erb
```diff
@@ -40,7 +40,7 @@ See COPYRIGHT and LICENSE files for more details.
     <% dialog.with_footer do %>
       <%= render(Primer::Beta::Button.new(data: { close_dialog_id: id })) { "Cancel" } %>
       <%= render(Primer::Beta::Button.new(scheme: :danger, tag: :a, href: delete_url(delete_member: true), data: { method: :delete })) { "Keep only shared" } %>
-      <%= render(Primer::Beta::Button.new(scheme: :danger, tag: :a, href: delete_url(delete_member: true, delete_shared_work_packages: true), data: { method: :delete })) { "Remove all" } %>
+      <%= render(Primer::Beta::Button.new(scheme: :danger, tag: :a, href: delete_url(delete_member: true, delete_work_package_shares: true), data: { method: :delete })) { "Remove all" } %>
```
aaron-contreras: Thanks, this is no longer misleading.

toy: Noted while fixing specs that this reads wrong

app/components/members/delete_member_dialog_component.html.erb
```diff
@@ -0,0 +1,85 @@
...
+      <% end %>
+    <% end %>
+
+    <% dialog.with_footer do %>
+      <%= render(Primer::Beta::Button.new(data: { close_dialog_id: id })) { t(:button_cancel) } %>
+      <%= render(Primer::Beta::Button.new(scheme: :danger, tag: :a, href: delete_url(project: true), data: { method: :delete })) { t(:button_remove_member) } %>
+      <%= render(Primer::Beta::Button.new(scheme: :danger, tag: :a, href: delete_url(project: true, work_package_shares_role_id: "all"), data: { method: :delete })) { t(:button_remove_member_and_shares) } %>
+    <% end %>
+  <% else %>
+    <% if can_delete_roles? %>
```
ulferts: This can be turned into a single `elsif` to avoid one layer of else-if

app/components/members/delete_member_dialog_component.html.erb
```diff
@@ -0,0 +1,85 @@
...
+      <% end %>
+
+      <% dialog.with_footer do %>
+        <%= render(Primer::Beta::Button.new(data: { close_dialog_id: id })) { t(:button_cancel) } %>
+        <%= render(Primer::Beta::Button.new(scheme: :danger, tag: :a, href: delete_url(project: true), data: { method: :delete })) { t(:button_remove) } %>
+      <% end %>
+    <% else %>
+      <% dialog.with_body do %>
+        <%= render(Primer::Beta::Text.new(tag: "p")) { scoped_t(:cannot_delete_group_member) } %>
+        <%= render(Primer::Beta::Text.new(tag: "p")) { scoped_t(:cannot_delete_group_member_note_html, administration_settings_link:) } %>
```
ulferts: In case the user is no admin, referencing the admin group settings will lead to a link those users are not allowed to follow. I'd rather only show that link if the user is an admin and can actually do something about the group membership.

app/components/members/delete_member_dialog_component.html.erb
```diff
@@ -0,0 +1,85 @@
...
+
+++#%>
+
+<%= render(Primer::Alpha::Dialog.new(id:, title: scoped_t(:title), size: :large)) do |dialog| %>
+  <% dialog.with_header(variant: :large) %>
+
+  <% if can_delete_roles? && can_delete_shares? %>
+    <% if can_delete? %>
+      <% dialog.with_body do %>
+        <%= render(Primer::Beta::Text.new(tag: "p")) { scoped_t(:will_remove_the_users_role_html, shared_work_packages_link:) } %>
```
ulferts: I like the usage of `scoped_t` here. Definitely improves readability.

ulferts: The `if` statement can be moved inside the `dialog.with_body` as both cases will just change the content of the body. That would make it a bit easier to understand.

app/components/members/delete_work_package_shares_dialog_component.html.erb
```diff
@@ -0,0 +1,112 @@
...
+
+      <% dialog.with_footer do %>
+        <%= render(Primer::Beta::Button.new(data: { close_dialog_id: id })) { t(:button_cancel) } %>
+      <% end %>
+    <% end %>
+  <% else %>
+    <% if direct_shared_work_packages_count? %>
+      <% if inherited_shared_work_packages_count? %>
+        <% dialog.with_body do %>
+          <%= render(Primer::Beta::Text.new(tag: "p")) do %>
```
ulferts: You could extract the rather repetitive `render(Primer::Beta::Text.new(tag: "p"))` into a method like `render_paragraph` or `render_text` or `primer_text`

app/components/members/delete_work_package_shares_dialog_component.html.erb
```diff
@@ -0,0 +1,112 @@
...
+        <%= render(Primer::Beta::Text.new(tag: "p")) { scoped_t(:cannot_remove_inherited_with_role, shared_role_name:) } %>
+        <%= render(Primer::Beta::Text.new(tag: "p")) { scoped_t(:cannot_remove_inherited_note_html, administration_settings_link:) } %>
+      <% end %>
+
+      <% dialog.with_footer do %>
+        <%= render(Primer::Beta::Button.new(data: { close_dialog_id: id })) { t(:button_cancel) } %>
+      <% end %>
+    <% end %>
+  <% else %>
+    <% if direct_shared_work_packages_count? %>
```
ulferts: Here, you could again avoid one nesting when using `elsif` instead.

app/components/members/delete_work_package_shares_dialog_component.html.erb
```diff
@@ -0,0 +1,112 @@
...
+        <%= render(Primer::Beta::Text.new(tag: "p")) { scoped_t(:cannot_remove_inherited, shared_role_name:) } %>
+        <%= render(Primer::Beta::Text.new(tag: "p")) { scoped_t(:cannot_remove_inherited_note_html, administration_settings_link:) } %>
+      <% end %>
+
+      <% dialog.with_footer do %>
+        <%= render(Primer::Beta::Button.new(data: { close_dialog_id: id })) { t(:button_cancel) } %>
+      <% end %>
+    <% end %>
+  <% end %>
+<% end %>
```
ulferts, line 107: These deeply nested conditions are hard to follow and I don't have a recipe I am sure will work to improve the situation. But what I would try is avoiding nested `if` in favor of `elsif` wherever the full content of the modal (body + footer) differentiates between the cases. That would require repeating some of the conditions but I believe that this is actually beneficial for readability. Next thing I am rather unsure about but might be worth a try is trying to label the different variants with something like `only_direct_shares?` and `direct_and_inherited_shares?`. With that in place I would revisit whether it makes sense to deliberately accept repetitions in favor of having the different variants stand out more clearly (and being able to introduce more states) by having both the body as well as the footer completely mentioned within each state. To avoid repetition, it might then be necessary to extract repeating parts into the methods.

But I really don't know, if that is a good change and would have to try it out in order to access whether it improves he situation.

config/locales/en.yml
```diff
@@ -345,6 +344,29 @@ en:
...
+      also_user_has_shares_html: "Also, %{shared_work_packages_link} have been shared with this user."
+      remove_just_user_or_shares_too: "Do you want to remove just the user as a direct member (and keep the shares) or remove the work package shares too?"
+      will_remove_all_access_priveleges: "Deleting this member will remove all access privileges of the user to the project. The user will still exist as part of the instance."
+      cannot_delete_group_member: "You cannot delete this member because they belong to a group that is itself a member of this project."
+      cannot_delete_group_member_note_html: "You can either remove the group as a member of the project or this specific member from the group in the %{administration_settings_link}."
+    delete_work_package_shares_dialog:
+      title: "Revoke work package shares"
+      shared_with_this_user_html: "%{all_shared_work_packages_link} have been shared with this user."
+      shared_with_this_group_html: "%{all_shared_work_packages_link} have been shared with this group."
+      shared_with_permission_html: "Only %{shared_work_packages_link} have been shared with %{shared_role_name} permissions."
```
ulferts: These always have the plural "have" in them. In case only "1 work package" is shared, that reads oddly.

config/locales/en.yml
```diff
@@ -333,7 +333,6 @@ en:
       no_results_content_text: Add a member to the project
     invite_by_mail: "Send invite to %{mail}"
     send_invite_to: "Send invite to"
-    no_modify_on_shared: "You currently cannot modify or remove shared memberships through the member page. Use the sharing modal instead."
```
ulferts, line 349: Bonus points for removing this key

app/services/members/delete_service.rb
```diff
@@ -29,6 +29,14 @@
...
   include Members::Concerns::CleanedUp
 
+  def destroy(object)
+    if object.member_roles.where.not(inherited_from: nil).empty?
+      super
+    else
+      object.member_roles.where(inherited_from: nil).destroy_all
+    end
+  end
+
```
ulferts, line 39: In case the member is in fact not completely deleted, I would not send out a notification (`MEMBER_DESTROYED`). The member is still there after all.

spec/services/members/delete_by_principal_service_spec.rb
```diff
@@ -0,0 +1,214 @@
...
+      end
+      let!(:member_c) do
+        create(:member,
+               project:,
+               principal:,
+               member_roles: [build(:member_role, role:), build(:member_role, role:, inherited_from: 123)],
+               entity: work_package_c)
+      end
+      let!(:member_d) do
+        create(:member,
```
ulferts, line 168: Are these `create` calls necessary or would a `build`/`build_stubbed` suffice?

ulferts (changes requested):
@toy, I overall like the PR. The structure looks good and the code is well written. 

As was to be expected when tackling a complex topic, there are a couple of issues in the PR: 

* If the user has the permission "Share work packages" but lacks the permission to "Manage members", the actions in the members list will still show the "delete action". Clicking on it will lead to an error.

* There is an edge case when: 
  * User A has a WP shared with "Comment" role
  * Group G has same WP shared with "View" role
  * Going to the "Comment" filter in the members list and choosing to revoke shares
The user can click on "Revoke all" but it will be the same as removing only the "Comment" shares. The ones of the roles are not removed. This might be an edge case but it feels weird:
image
We would need to discuss if we want to do something about it.

* No feature specs have been added. There are probably too many cases to cover in feature specs, but covering the main ones should be done. I did not check how well that part of the application was already covered by feature specs so take that statement with a grain of salt. 

* When removing a group, the modal says: "Deleting this member will remove all access privileges of the user to the project. The user will still exist as part of the instance." - speaking of a user but it is a group. No mentioning of how it will affect the group's members, the users. Ideally some text would explain that their access privileges are removed as well.

* A topic of personal preference probably, but I would prefer a bit more space between the header and the body of the dialog. In total, the dialogs as designed look a bit more lofty:
image

compared to ones implemented:
image

Which is due to the space between body and header but also between paragraphs in the body. I see in the implementation that the defaults were chosen so I don't blame this PR for it. But I guess that defining the layout of such confirmation dialogs a bit more would be fruitful /cc @HDinger 

* When removing project memberships not via the project members table but rather via the API, no confirmation dialog at all is displayed and only the project memberships are removed whereas the shares remain. Ideally, we would reuse the modals and have the user decide whether to keep/delete the shares as well.  

* The flash messages always report that the "member" was removed from the project. This is also true if only shares have been removed or when the member wasn't completely removed. This is a bit surprising in some scenarios. I would not want to change the way we display the messages for now but the content could be improved.

toy:
> If the user has the permission "Share work packages" but lacks the permission to "Manage members", the actions in the members list will still show the "delete action". Clicking on it will lead to an error.

Good catch, I first switched to checking access to `destroy_by_principal` action, but as it is needed also for deleting work package shares, it is not valid to use for permission to delete member.
See if you agree with https://github.com/opf/openproject/pull/14721/commits/80af7b8eeb43b50a227d6a157e204d6801d81cae

toy:
> * There is an edge case when:
>   
>   * User A has a WP shared with "Comment" role
>   * Group G has same WP shared with "View" role
>   * Going to the "Comment" filter in the members list and choosing to revoke shares
>     The user can click on "Revoke all" but it will be the same as removing only the "Comment" shares. The ones of the roles are not removed. This might be an edge case but it feels weird:
> 
> image We would need to discuss if we want to do something about it.

I already discussed it with @marcalcobe and he suggested to treat it as an edge case.

toy:
> * No feature specs have been added. There are probably too many cases to cover in feature specs, but covering the main ones should be done. I did not check how well that part of the application was already covered by feature specs so take that statement with a grain of salt.

There were already some feature tests for deleting member that go through action menu and dialog, there are integration tests for controller that check all main parameters and services are completely covered. But tell me if you think that there should be feature tests for revoking work packages and/or revoking work packages alongside member.

toy:
> * When removing a group, the modal says: "Deleting this member will remove all access privileges of the user to the project. The user will still exist as part of the instance." - speaking of a user but it is a group. No mentioning of how it will affect the group's members, the users. Ideally some text would explain that their access privileges are removed as well.

Good catch, I handled the user vs group in revoke dialog, but not here. I will ask Marc for texts

toy:
> * A topic of personal preference probably, but I would prefer a bit more space between the header and the body of the dialog. In total, the dialogs as designed look a bit more lofty:
> 
> image
> compared to ones implemented: image
> 
> Which is due to the space between body and header but also between paragraphs in the body. I see in the implementation that the defaults were chosen so I don't blame this PR for it. But I guess that defining the layout of such confirmation dialogs a bit more would be fruitful /cc @HDinger

I'm trying to follow using components as is without additional styling. Maybe @marcalcobe also can chime in

toy:
> * When removing project memberships not via the project members table but rather via the API, no confirmation dialog at all is displayed and only the project memberships are removed whereas the shares remain. Ideally, we would reuse the modals and have the user decide whether to keep/delete the shares as well.

I didn't get what you mean by API. Wouldn't API mean not Web UI?

toy:
> * The flash messages always report that the "member" was removed from the project. This is also true if only shares have been removed or when the member wasn't completely removed. This is a bit surprising in some scenarios. I would not want to change the way we display the messages for now but the content could be improved.

The texts of notifications was the thing I didn't like, but I though we agreed to leave it as is for now. @marcalcobe

spec/services/members/delete_by_principal_service_spec.rb
toy: In this case yes, as the service uses scopes to find what to delete and it should be covered in tests

app/components/members/delete_work_package_shares_dialog_component.html.erb
toy: I was thinking about doing that

app/components/members/delete_member_dialog_component.html.erb
```diff
@@ -0,0 +1,85 @@
...
+          <%= scoped_t(:remove_just_user_or_shares_too) %>
+          <%= scoped_t(:will_not_affect_inherited_shares) if inherited_shared_work_packages_count? %>
+        <% end %>
+      <% end %>
+    <% end %>
+
+    <% dialog.with_footer do %>
+      <%= render(Primer::Beta::Button.new(data: { close_dialog_id: id })) { t(:button_cancel) } %>
+      <%= render(Primer::Beta::Button.new(scheme: :danger, tag: :a, href: delete_url(project: true), data: { method: :delete })) { t(:button_remove_member) } %>
+      <%= render(Primer::Beta::Button.new(scheme: :danger, tag: :a, href: delete_url(project: true, work_package_shares_role_id: "all"), data: { method: :delete })) { t(:button_remove_member_and_shares) } %>
```
HDinger: image

Primer recommends to use the `danger` buttons sparingly. So I am wondering whether we can avoid the doubled buttons here. Maybe with a checkbox before stating something like "remove shares as well". Opinions on that @marcalcobe @psatyal ?

toy: We discussed using checkboxes, and even combining editing roles and deleting into one dialog with lots of checkboxes, but went back to buttons

marcalcobe:
> > * When removing a group, the modal says: "Deleting this member will remove all access privileges of the user to the project. The user will still exist as part of the instance." - speaking of a user but it is a group. No mentioning of how it will affect the group's members, the users. Ideally some text would explain that their access privileges are removed as well.
> 
> Good catch, I handled the user vs group in revoke dialog, but not here. I will ask Marc for texts

I've added some mockups for this case when trying to remove a group: https://www.figma.com/file/lLDYzQW7evoBA4lha5PwXO/Project-members?type=design&node-id=1136-13208&mode=design&t=Nxyc5ejAZ3jBQXUf-4

app/components/members/delete_member_dialog_component.html.erb
marcalcobe: We discussed this again in the Frontend daily this morning with all devs and agreed that we would need to modify this design again to try to follow as much as we can Primer. Together with @psatyal we created the mockups in this page: https://www.figma.com/file/lLDYzQW7evoBA4lha5PwXO/Project-members?type=design&node-id=1135-29390&mode=design

I know this is a change again on how the modal looks like and it should had been taken in consideration earlier, I'm sorry for that and I hope this doesn't cost to much extra time to implement.

marcalcobe:
> > * A topic of personal preference probably, but I would prefer a bit more space between the header and the body of the dialog. In total, the dialogs as designed look a bit more lofty:
> > 
> > image
> > compared to ones implemented: image
> > Which is due to the space between body and header but also between paragraphs in the body. I see in the implementation that the defaults were chosen so I don't blame this PR for it. But I guess that defining the layout of such confirmation dialogs a bit more would be fruitful /cc @HDinger
> 
> I'm trying to follow using components as is without additional styling. Maybe @marcalcobe also can chime in

About this I would suggest we use the component as it comes from Primer to not complicate the implementation.

app/components/members/delete_work_package_shares_dialog_component.html.erb
toy: I've reduced nesting (but didn't remove it completely) and it looks cleaner, so have a look

app/services/members/delete_service.rb
toy: Tell me if you have a better idea how to handle it then ff49068bde55501a9644ee6c94577c9097d603e3

app/services/members/delete_service.rb
```diff
@@ -44,7 +46,7 @@ def after_perform(service_call)
       member = call.result
 
       cleanup_for_group(member)
-      send_notification(member)
+      send_notification(member) if @send_notification
```
toy: use `.destroyed?`

app/components/members/delete_member_dialog_component.html.erb
```diff
@@ -0,0 +1,97 @@
...
+      <% else %>
+        <%= paragraph { scoped_t(:cannot_delete_inherited_membership_note_non_admin) } %>
+      <% end %>
+    <% end %>
+
+    <% dialog.with_footer do %>
+      <%= cancel_button %>
+    <% end %>
+  <% end %>
+<% end %>
```
ulferts, line 97: I like it @toy. It still requires heavy thinking but the combination of the case statement with the reduction due to `paragraph` at al. makes reading it simpler. 👏

ulferts:
> > * When removing project memberships not via the project members table but rather via the API, no confirmation dialog at all is displayed and only the project memberships are removed whereas the shares remain. Ideally, we would reuse the modals and have the user decide whether to keep/delete the shares as well.
> 
> I didn't get what you mean by API. Wouldn't API mean not Web UI?

For whatever reason, I definitely used the wrong term. What I meant to say is that there is no confirmation dialog when removing the project memberships from the administration (Administration -> Users and permissions -> "A User" -> Project tab):

image

toy:
> > > * When removing project memberships not via the project members table but rather via the API, no confirmation dialog at all is displayed and only the project memberships are removed whereas the shares remain. Ideally, we would reuse the modals and have the user decide whether to keep/delete the shares as well.
> > 
> > 
> > I didn't get what you mean by API. Wouldn't API mean not Web UI?
> 
> For whatever reason, I definitely used the wrong term. What I meant to say is that there is no confirmation dialog when removing the project memberships from the administration (Administration -> Users and permissions -> "A User" -> Project tab):
> 
> image

Ah yes, I agree that it should show the dialog, though I suggest doing it separately

ulferts (changes requested):
Hey @toy ,

thanks for addressing the issues I raised. Looks almost good now. Almost, since when retesting I found a weird behaviour which I think is a regression:

Every member row, regardless of it having a work package shared or not and regardless of it being a user (and not a group) has the option now to "Revoke work package shares", when clicking on it the following is displayed:

image

Again, this was shown after clicking on the action item for a user.

toy:
> Every member row, regardless of it having a work package shared or not and regardless of it being a user (and not a group) has the option now to "Revoke work package shares", when clicking on it the following is displayed:

Good catch, indeed I broke it in the last commit

ulferts (approved):
🎉 Thanks for pushing through.

PR #14724, opened 2024-02-06
[52149] Clean menu structure on project lists page

bsatarnejad:
- [x] Move show Gantt and overall activity buttons to more menu button
- [x] Remove configure from more menu entries
- [x] Fix failing specs

https://community.openproject.org/projects/openproject/work_packages/52149/activity


app/components/projects/index_page_header_component.html.erb
```diff
@@ -73,6 +52,26 @@
           item.with_leading_visual_icon(icon: 'op-file-download')
         end
 
+        menu.with_item(
+          tag: :a,
+          label: t(:label_overall_activity),
+          href: activities_path
+        )
+
+        if !gantt_portfolio_project_ids.empty?
```
aaron-contreras: There's a method to test the opposite of empty

```suggestion
        if gantt_portfolio_project_ids.any?
```

spec/support/pages/projects/index.rb
```diff
@@ -239,7 +242,7 @@ def open_filters
       def click_more_menu_item(item)
         page.find('[data-test-selector="project-more-dropdown-menu"]').click
 
-        find('.ActionListItem', text: item).click
+        page.find('.ActionListItem', text: item, exact_text: true).click
```
aaron-contreras, line 245: 👍

spec/support/pages/projects/index.rb
```diff
@@ -122,9 +122,12 @@ def expect_no_project_create_button
...
 
-      def expect_gantt_button(disabled: false)
-        expect(page).to have_css("button#{disabled ? '[disabled]' : ''}",
-                                 text: 'Open as Gantt view')
+      def expect_gantt_menu_entry(visible: false)
+        if visible
+          expect(page).to have_css('#projects-index-open-as-gantt', text: 'Open as Gantt view')
+        else
+          expect(page).to have_no_css('#projects-index-open-as-gantt', text: 'Open as Gantt view')
+        end
```
aaron-contreras: You can use `have_(no)_link` and not rely on the id anymore :) 
```suggestion
      def expect_gantt_menu_entry(visible: false)
        if visible
          expect(page).to have_link('Open as Gantt view')
        else
          expect(page).to have_no_link('Open as Gantt view')
        end
```

aaron-contreras (commented):
Very nice ⭐

I noticed that "Configure View" no longer links somewhere. Is that being worked on in another work package? Other than that, the rest looks good!

bsatarnejad:
> Very nice ⭐
> 
> I noticed that "Configure View" no longer links somewhere. Is that being worked on in another work package? Other than that, the rest looks good!

Hi @aaron-contreras Thanks a lot for your great remarks!
Yeah that's right, the actions linked to some of the new entries like 'rename', 'configure view', 'save as', ... are implemented in other WPs.

aaron-contreras (approved):

PR #14725, opened 2024-02-06
Get rid of type enums

klaustopher:
Rails 7.1 now validate that enums are backed by a database column. If our application is run in multi-tenancy mode, like it is on our hosted offering, this check fails because Rails does not properly detect the column if it is defined as an enum internally.

We have pjut a band-aid on the problem by fixing the column to map to a string in the rails application in #14713 but we have discussed again if we really need the database level security offered by enums. The general decision was that the only sanctioned interface to the database is via the rails app and the rails app ensures that only allowed values are put in the DB.

To mimick the current behavior, we have added contraints to the columns, but in the future we should revisit if we **REALLY** need that level of security on the DB side or if we can live without such strict constraints, as we do not use them in other areas of the database that would be even more critical.

### DIFF
--- a/db/migrate/20240206085104_remove_enums.rb
+++ b/db/migrate/20240206085104_remove_enums.rb
@@ -1,0 +1,73 @@
+class RemoveEnums < ActiveRecord::Migration[7.1]
+  def up
+    change_column(:storages, :health_status, :string, default: 'pending')
+    add_check_constraint(:storages,
+                         'health_status IN (\'pending\', \'healthy\', \'unhealthy\')',
+                         name: 'storages_health_status_check')
+
+    change_column(:last_project_folders, :mode, :string, default: 'inactive')
+    add_check_constraint(:last_project_folders,
+                         'mode IN (\'inactive\', \'manual\', \'automatic\')',
+                         name: 'last_project_folders_mode_check')
+
+    change_column(:project_storages, :project_folder_mode, :string)
+    add_check_constraint(:project_storages,
+                         'project_folder_mode IN (\'inactive\', \'manual\', \'automatic\')',
+                         name: 'project_storages_project_folder_mode_check')
+
+    change_column(:delayed_job_statuses, :status, :string, default: 'in_queue')
+    add_check_constraint(:delayed_job_statuses,
+                         'status IS NULL OR status IN (\'in_queue\', \'error\', \'in_process\', \'success\', \'failure\', \'cancelled\')',
+                         name: 'delayed_job_statuses_status_check')
+
+    execute <<~SQL.squish
+      DROP TYPE public.delayed_job_status RESTRICT;
+      DROP TYPE public.project_folder_modes RESTRICT;
+      DROP TYPE public.storage_health_statuses RESTRICT;
+    SQL
+  end
+
+  def down
+    execute <<~SQL.squish
+      CREATE TYPE public.delayed_job_status AS ENUM (
+          'in_queue',
+          'error',
+          'in_process',
+          'success',
+          'failure',
+          'cancelled'
+      );
+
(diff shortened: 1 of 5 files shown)


db/migrate/20240206085104_remove_enums.rb
```diff
@@ -0,0 +1,73 @@
...
+                         'project_folder_mode IN (\'inactive\', \'manual\', \'automatic\')',
+                         name: 'project_storages_project_folder_mode_check')
+
+    change_column(:delayed_job_statuses, :status, :string, default: 'in_queue')
+    add_check_constraint(:delayed_job_statuses,
+                         'status IS NULL OR status IN (\'in_queue\', \'error\', \'in_process\', \'success\', \'failure\', \'cancelled\')',
+                         name: 'delayed_job_statuses_status_check')
+
+    execute <<~SQL.squish
+      DROP TYPE public.delayed_job_status RESTRICT;
```
machisuji: Is there a reason for the explicit schema prefix (`public.`) here?

The problem is that this will break the migration in the m10y migrations where each schema needs to have the type.

db/migrate/20240206085104_remove_enums.rb
```diff
@@ -0,0 +1,73 @@
...
+      DROP TYPE public.project_folder_modes RESTRICT;
+      DROP TYPE public.storage_health_statuses RESTRICT;
+    SQL
+  end
+
+  def down
+    execute <<~SQL.squish
+      CREATE TYPE public.delayed_job_status AS ENUM (
```
machisuji: Same issue here. We should not qualify the schema so this works in the m10y context.

machisuji (changes requested):
Thanks for addressing this! Do have one remark, though!

db/migrate/20240206085104_remove_enums.rb
klaustopher: I just copy&pasted from the current `structure.sql` no other reason. Will remove

machisuji: The search path in m10y is `<tenant schema>, pgcatalog`. So just creating the type in public won't be enough. I'm not sure if we could only just create it in `pgcatalog`. But as it is right now, each schema gets their own type.

db/migrate/20240206085104_remove_enums.rb
```diff
@@ -0,0 +1,73 @@
...
+
+      CREATE TYPE public.storage_health_statuses AS ENUM (
+          'pending',
+          'healthy',
+          'unhealthy'
+      );
+    SQL
+
+    remove_check_constraint(:storages, name: 'storages_health_status_check')
+    change_column(:storages, :health_status, :storage_health_statuses, default: nil,
```
Kharonus, line 55: ❓ `default: nil` necessary here for any reason? because the default should be same as in the `up`, just that in the `down` we need to select the default from the enum type, which usually cannot be nil, right?

klaustopher: Yes, we need to switch to `NULL` and then in the next step switch it back to the enum type again, see next line. Otherwise the migration fails because it cannot convert the existing default (`string`) to the ENUM

Kharonus: So this means, on a down migration we loose all values? or can it convert the string values, just not the default?

klaustopher: the `using: 'health_status::storage_health_statuses'` makes sure that all values are converted. it just does not work for the default. Maybe there is a way with typecasting in the `CHANGE COLUMN` call for the default. But the migration is run in a transaction, so this is just the easier way :)

machisuji (approved):
👍

oliverguenther (approved):

PR #14726, opened 2024-02-06
Housekeeping: amend user invitation case for members controller

machisuji:
Adds missing spec for #14717.

### DIFF
--- a/spec/controllers/members_controller_spec.rb
+++ b/spec/controllers/members_controller_spec.rb
@@ -175,6 +175,52 @@ RSpec.describe MembersController do
           expect(user4).to be_member_of(project)
         end
       end
+
+      context 'with yet-to-be-invited emails' do
+        let(:emails) { ['h.wurst@openproject.com', 'l.lustig@openproject.com'] }
+        let(:params) do
+          {
+            project_id: project.id,
+            member: {
+              role_ids: [role.id],
+              user_ids: [emails.first] + [user2.id, user3.id] + [emails.last]
+            }
+          }
+        end
+
+        let(:invited_users) { User.where(mail: emails).to_a }
+        let(:users) { invited_users + [user2, user3] }
+        let(:original_member_count) { Member.count }
+
+        before do
+          original_member_count
+
+          allow(Mails::InvitationJob).to receive(:perform_later)
+
+          post :create, params:
+        end
+
+        it 'redirects to the members list' do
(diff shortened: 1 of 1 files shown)


spec/controllers/members_controller_spec.rb
```diff
@@ -175,6 +175,52 @@
...
+          expect(users).to all be_member_of(project)
+
+          expect(Member.count).to eq (original_member_count + users.size)
+        end
+
+        it 'invites new users' do
+          expect(invited_users.size).to be 2
+
+          invited_users.each do |user|
+            expect(Mails::InvitationJob)
```
machisuji: I've tried to check via `ActionMailer::Base.deliveries` and also to check for enqueued delayed jobs, but both is empty here for some reason. If anyone knows why, please let me know. I don't. So I just test it like this instead.

aaron-contreras: You want to call `perform_enqueued_jobs` before the expectation.

spec/controllers/members_controller_spec.rb
```diff
@@ -175,6 +175,52 @@
...
+        it 'adds members' do
+          expect(users.size).to eq 4 # 2 emails, 2 existing users
+          expect(users).to all be_member_of(project)
+
+          expect(Member.count).to eq (original_member_count + users.size)
+        end
+
+        it 'invites new users' do
+          expect(invited_users.size).to be 2
+
```
aaron-contreras:
```suggestion
          expect(invited_users.size).to be 2

          perform_enqueued_jobs

```

aaron-contreras (approved):
Thanks for this 👍  It's fine either way but if you'd like to test on the actual mail delivery I annotated how you can do so.

spec/controllers/members_controller_spec.rb
machisuji: @aaron-contreras Thanks, that would help. 😄

machisuji:
> Thanks for this 👍 It's fine either way but if you'd like to test on the actual mail delivery I annotated how you can do so.

Thanks a lot! I keep forgetting this and didn't seem to see it in the other specs I looked up either. Anyway, I'll adapt the PR to simply use that and then check the deliveries.

spec/controllers/members_controller_spec.rb
machisuji: Note to self: `perform_enqueued_jobs`

PR #14733, opened 2024-02-07
extract PostgresEnvironment to share between BackupJob and backup rake tasks

toy:
Use same code in BackupJob and backup rake tasks to rely on environment variables to pass database config for `pg_dump` and `pg_restore` in rake tasks. It does look like concern about preferring temp files over environment variables for security reasons is outdated and environment variables were used for BackupJob already for several years.
Followup of #14701


machisuji (approved):
Good stuff! Thanks! 👍

PR #14736, opened 2024-02-07
[52149] Clean menu structure on project lists page (change order of entries, use correct icons)

bsatarnejad:
https://community.openproject.org/work_packages/52149/activity#activity-12


ulferts (changes requested):
This is now looking way better, thanks @bsatarnejad.

With the removal of the tooltip, a [method became unused](https://github.com/opf/openproject/blob/dev/app/components/projects/index_page_header_component.rb#L78-L87). Removing the method will lead to two i18n keys becoming unused so they should then be deleted as well.

ulferts (approved):

PR #14745, opened 2024-02-08
Remove public permission adding when checking in Role

klaustopher:
Since #14010 we are always adding public permissions to ProjectRoles, this way we don't have to manually add them every time we check for a permission.

Implements https://community.openproject.org/projects/openproject/work_packages/50769


dombesz (approved):
LGTM! 👍

PR #14746, opened 2024-02-08
[#52673] replaced index scope

Kharonus:
[52673](https://community.openproject.org/wp/52673)

- only project related memberships are shown in the table


akabiru (approved):
Nice catch 👍🏾

PR #14754, opened 2024-02-09
[#52755] store project folder id after sync in historic data

Kharonus:
[#52755](https://community.openproject.org/work_packages/52755)

### WAT???

- on successful AMPF creation -> store folder id in `last_project_folder`
- enabled AMPF deletion for all storages
- fixed error formatting in OD-sync service according to NC-sync service (to handle timeouts etc.)


ba1ash (commented):
Could you please add appropriate checks to tests? Apart from that it looks good to me :+1:

modules/storages/app/models/storages/project_storage.rb
```diff
@@ -61,7 +61,10 @@ def automatic_management_possible?
   end
 
   def project_folder_path
-    "#{storage.group_folder}/#{project.name.tr('/', '|')} (#{project.id})/"
+    return "#{storage.group_folder}/#{project.name.tr('/', '|')} (#{project.id})/" if storage.provider_type_nextcloud?
+    return project_folder_id if storage.provider_type_one_drive?
+
+    raise 'Unknown Storage'
```
mereghost, line 67: 🟢 WDYT of creating an Exception with this name and raise it? 

🍏 Also, I've been playing with the idea of who should know about how we should format project folder paths. It is clearly storage dependent, but I'm not sure I'd move this to the storage itself. Dilemmas... >.<

modules/storages/app/services/storages/one_drive_managed_folder_sync_service.rb
```diff
@@ -126,6 +126,12 @@ def create_folder(project_storage, folder_path)
         .call(storage: @storage, folder_path:)
         .match(on_failure: ->(error) { format_and_log_error(error, folder_path:) },
                on_success: ->(folder_info) do
+                 last_project_folder = ::Storages::LastProjectFolder
+                                         .find_by(
+                                           project_storage_id: project_storage.id,
+                                           mode: project_storage.project_folder_mode
+                                         )
+                 last_project_folder.update(origin_folder_id: folder_info.id)
                  project_storage.update(project_folder_id: folder_info.id)
```
mereghost: 🟡 This should be an atomic operation (meaning both writes should succeed) so we don't end up with weird results.

I'd chain the calls if we don't care about one of the operations (`project_storage.update and last_project_folder.update`) or wrap this in a transaction if we want to make sure that both succeed and fail together.

modules/storages/app/services/storages/project_storages/delete_service.rb
```diff
@@ -33,7 +33,7 @@ module Storages::ProjectStorages
   # by the model before_destroy hook.
   class DeleteService < ::BaseServices::Delete
     def before_perform(*)
-      delete_project_folder if model.storage.is_a?(Storages::NextcloudStorage)
+      delete_project_folder if model.project_folder_automatic?
```
mereghost, line 36: 💚

mereghost (approved):
LGTM a couple of points noted, that while could lead to some strange situations I don't see the need to block the merge.

modules/storages/app/models/storages/project_storage.rb
Kharonus: yeah, I feel you. I basically did not try to solve that question, and simply replaced the code in the current place.

modules/storages/app/services/storages/one_drive_managed_folder_sync_service.rb
Kharonus: so, question: option a or b? chain two separate calls? or introduce transaction for writing this to db?

ba1ash: :+1: I suppose to be correct we should use transaction around both operations. I will add it.

cbliard (approved):
Nothing to say about the code, and from the user point of view, it fixes the issue and works as expected.

modules/storages/app/services/storages/one_drive_managed_folder_sync_service.rb
ba1ash: @mereghost @Kharonus it has been done. Have a look, please.

cbliard:
About the failing test, it has been fixed in [another commit](https://github.com/opf/openproject/pull/14766/commits/4074f6150660d92a44db2a76e8fbb6e1ddd85414).

PR #14759, opened 2024-02-09
fix time entries endpoint to return ongoing time entries while lacking log own time permission

machisuji:
This might actually be exactly the problem in [#50025](https://community.openproject.org/projects/openproject/work_packages/50025/activity).

The steps to reproduce are the following:

1. Start logging some time for oneself
2. Remove "log own time" permission from role in project
3. Try logging time again

The user won't see that there is an ongoing time entry already.

I initially hoped to include the fix in 13.3.0 but that won't be enough time anymore, now.
This PR still needs tests and QA. Let's put this into 13.3.1 then.

I will continue with this and confirm it when I'm back on Wednesday.

### DIFF
--- a/modules/costs/spec/models/queries/time_entries/time_entry_query_integration_spec.rb
+++ b/modules/costs/spec/models/queries/time_entries/time_entry_query_integration_spec.rb
@@ -55,6 +55,14 @@ RSpec.describe Queries::TimeEntries::TimeEntryQuery, 'integration' do
       it 'only returns the users own time entries' do
         expect(subject).to contain_exactly(user_timer)
       end
+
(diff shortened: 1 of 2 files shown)


oliverguenther (approved):

modules/costs/spec/models/queries/time_entries/time_entry_query_integration_spec.rb
```diff
@@ -55,6 +55,14 @@
       it 'only returns the users own time entries' do
         expect(subject).to contain_exactly(user_timer)
       end
+
+      context 'when user has log_time permission' do
```
machisuji, line 59: <3

PR #14768, opened 2024-02-12
[#40749] Consistently calculate % Complete in Work Package hierarchies

aaron-contreras:
See: https://community.openproject.org/work_packages/40749

Current status towards 14.0 (Feature freeze on Wednesday April 10)

- Work packages UI:
    - ✅ display unset % Complete as "-" instead of "0%" in activity
- Popover:
    - ✅ work-based mode
    - ✅ Fix jumpiness of popover with the help of Henriette
    - ✅ Fix updating the angular part when receiving a reply from the hotwire part with the help of Oliver
    - ✅ Display the % complete as simple text in popover
    - ✅ Make progress popover work on work package creation form
    - ✅ status-based mode
    - ✅ Extract locale strings from modal (Aaron)
    - ✅ Only display statuses that are valid transitions from the current status (Aaron)
    - ✅ Display migration information on the popover when only % Complete is set ([figma link](https://www.figma.com/file/znrocQvFNLb5jXRXNo7oeD/Progress-estimation?type=design&node-id=1467-18326&mode=design&t=YqkYf8DLuYt6DrHA-4))
    - 🔜 Preview computed values in popover on changes (Aaron / Attila - not launch critical)
    - ✅ See if jumpiness of the popover dialog can be fixed (between display of popover and loading of content, the size expands) (Aaron - use fixed sizes - 500px)
    - 🔜  See if possible to have units displayed in the popover (h, %) (Jens - not launch critical)
    - ✅  When inline creating a work package in table view, after setting a value for work/remaining work via the modal, it doesn't reflect on the display field as the date field does. (Oliver)
    - ✅ Add Learn More link to modal linking to documentation page (Aaron)
- Business logic:
    - ✅ Implement business logic described by specs
    - ✅ Total % Complete computed from total work and total remaining work
    - ✅ Statuses always have a % Complete, 0 by default
    - ✅ Unsetting remaining work while setting work produces an error
    - ✅ Unsetting work while setting remaining work produces an error
    - ✅ Setting Remaining work will set Work to the same value if unset
        - This limits issues with the backlogs module where the remaining work is set without explicitly setting the work. As a side effect, increasing the remaining work when on the backlogs task board can result in an error if it exceeds the work.
    - ✅ Unset % Complete when Work is 0h
- Migration:
    - ✅ Migrate disabled mode to work-based mode as described by spec
    - ✅ Migrate data for work packages in work-based mode
    - ✅ Add journals with changes (System user) with precise cause explanation
    - ✅ Migrate data for work packages in status-based mode
    - ✅ Recompute totals after migration
    - 🔜 Make sure concurrent modification of a work package while the migration job is performing is safe:
      - if a user changes work package data during that job execution, the changes from the job should be in a different journal. The changes from system must be different from the changes from user. updated_at and lock_version should be cached to check it, and then retried if different from the start. maybe lock too. See Jens comments [here](https://github.com/opf/openproject/pull/14768#discussion_r1555912837) and [here](https://github.com/opf/openproject/pull/14768#discussion_r1555918151).
- Administration:
    - ✅ Updating the default % complete of a status should start a job to update all % Complete in instance
    - ✅ Changing from work-based to status-based should start a job to update all % Complete in instance (Christophe)
    - ✅ Create meaningful journal entry for status changes in administration (System user)
    - ✅ Remove "Update work packages with done ratio" button in statuses page in admin
    - ~~Handle status % complete update through~~ [~~statuses API~~](https://www.openproject.org/docs/api/endpoints/statuses/) Since the API does not provide any write-endpoints, an update can not be carried out via the API
    - ✅ Remove "Disabled" mode from Administration UI
    - 🔜 (skippable) Implement the [new Primer UI](https://www.figma.com/file/znrocQvFNLb5jXRXNo7oeD/Progress-estimation?type=design&node-id=1245-33663&mode=design&t=kwtZPf5RHZqf9iUj-4)
- Misc:
    - ✅ Make it impossible to update % Complete by sending emails
    - ✅ No bulk editing of % Complete anymore
    - ✅ Setting status by commit message should not update % Complete anymore


app/services/work_packages/set_attributes_service.rb
```diff
@@ -281,6 +282,35 @@ def update_duration
...
+  # - +estimated_hours+
+  #
+  # Unless both +remaining_hours+ and +estimated_hours+ are set, +done_ratio+ will be
+  # considered 0.
+  def update_done_ratio
+    return if WorkPackage.use_status_for_done_ratio?
+    return unless work_package.remaining_hours_changed? || work_package.estimated_hours_changed?
+
+    work_package.done_ratio = if done_ratio_dependent_attribute_unset?
+                                0
```
cbliard: Eventually, when both work and remaining work are unset, the % Complete will need to be unset too.
That also means we'll have to add a new migration to remove the non-null constraint.

spec/services/work_packages/set_attributes_service_spec.rb
```diff
@@ -145,6 +145,68 @@
     it_behaves_like 'service call'
   end
 
+  context 'for done_ratio' do
```
cbliard: I find it strange to have the implementation and the spec in 2 different commits. Didn't you write them at the same time?

spec/services/work_packages/set_attributes_service_spec.rb
```diff
@@ -145,6 +145,68 @@
...
+          { remaining_hours: nil, done_ratio: 0 }
+        end
+
+        it_behaves_like 'service call'
+      end
+
+      context 'when both estimated_hours and remaining_hours are unset' do
+        let(:call_attributes) { { estimated_hours: nil, remaining_hours: nil } }
+        let(:expected_attributes) do
+          { estimated_hours: nil, remaining_hours: nil, done_ratio: 0 }
```
cbliard: like said in the other comment, expected values will eventually be `{ estimated_hours: nil, remaining_hours: nil, done_ratio: nil }`.
But that's ok, we are moving forward step by step.

spec/factories/work_package_factory.rb
```diff
@@ -125,5 +125,12 @@
                                     updated_at: work_package.journals.maximum(:updated_at))
       end
     end
+
+    callback(:after_build) do |work_package, _evaluator|
+      if work_package.estimated_hours.present? && work_package.remaining_hours.present?
+        work_package.done_ratio = (work_package.estimated_hours - work_package.remaining_hours) \
+                                  / work_package.estimated_hours * 100
+      end
+    end
```
cbliard: Nice 👍

docs/installation-and-operations/configuration/incoming-emails/README.md
```diff
@@ -184,17 +184,18 @@ The subject of the work package that shall be created is derived from the subjec
...
+| Key             | Description                                           | Example                          |
+|-----------------|-------------------------------------------------------|----------------------------------|
+| Project         | sets the project. Use the project identifier          | Project:test\_project            |
+| Assignee        | sets the assignee. Use the email or login of the user | Assignee:test.nutzer@example.org |
+| Type            | sets the type                                         | type:Milestone                   |
+| Version         | sets the version                                      | version:v4.1.0                   |
+| Start date      | sets the start date                                   | start date:2015-02-28            |
+| Due date        | sets the finish date                                  |                                  |
+| Estimated hours | sets the estimated hours. Use a number                | Estimated hours:10.5             |
+| Remaining hours | sets the remaining hours. Use a number                | Remaining hours:2.5              |
```
cbliard, line 195: That's where I do not know if we should use the old names or the new names (and have the old ones still working but undocumented). I suppose using the old names does not hurt, and it can still be revisited later.

cbliard: And the table looks nice now that you have added some padding to it.

spec/models/query/results_sums_integration_spec.rb
```diff
@@ -130,7 +130,7 @@
     create(:float_wp_custom_field)
   end
   let(:type) do
-    create(:type).tap do |t|
+    create(:type) do |t|
```
cbliard, line 130: Is it this kind of things that triggered you to create the Rubocop pull request?

cbliard (commented):
It looks great so far! Nice tests.

docs/installation-and-operations/configuration/incoming-emails/README.md
aaron-contreras: I second guessed on the names too but I thought it could be something to address when we ensure the same terms are used throughout the code base (solely work, remaining_work, percent_complete)

spec/models/query/results_sums_integration_spec.rb
aaron-contreras: Yes :)

spec/services/work_packages/set_attributes_service_spec.rb
aaron-contreras: I don't recall why. I probably didn't finish late one afternoon, committed, then continued with missing specs the next day. I'll squash them

spec/services/work_packages/set_attributes_service_spec.rb
cbliard: done

app/services/work_packages/set_attributes_service.rb
cbliard: done

lindenthal:
Hi @cbliard and @aaron-contreras,

gr**E**aT progr**E**ss!

grafik

In the modal the unit is unclear. It is not a big UX problem since it is shown in the table. And we will address the multi-unit support in 14.1. 

But if you have an idea how to fix this towards the multi-unit approach please let me know.


Best
Niels

app/workers/work_packages/update_progress_job.rb
```diff
@@ -0,0 +1,257 @@
...
+      CREATE UNLOGGED TABLE temp_wp_progress_values
+      AS SELECT
+        id,
+        status_id,
+        estimated_hours,
+        remaining_hours,
+        done_ratio,
+        NULL::double precision AS total_work,
+        NULL::double precision AS total_remaining_work,
+        NULL::integer AS total_p_complete
```
ulferts, line 80: I'd include the `updated_at` or `lock_version` here. Otherwise, a user might loose their input in case they update the work package between the time the `temp_...` table is filled and then written back into the `work_packages` table. When writing back, the sql could bail on rows that have a higher `updated_at`/`lock_version` than the one in the `temp_...` table.

app/workers/work_packages/update_progress_job.rb
```diff
@@ -0,0 +1,257 @@
...
+      RETURNING work_packages.id
+    SQL
+    results.column_values(0)
+  end
+
+  def create_journals_for_updated_work_packages(updated_work_package_ids)
+    WorkPackage.where(id: updated_work_package_ids).find_each do |work_package|
+      Journals::CreateService.new(work_package, system_user)
+        .call(cause: { type: "system_update", feature: "progress_calculation_changed" })
+    end
```
ulferts, line 247: Here, it could happen (don't know how likely it is), that the job runs, updates the values in the `work _packages` table but hasn't completed the journal creation yet, that a user comes along and updates the work package. The updates to the work/done_ratio would be attributed to the user.

The larger the dataset, the more likely something like this is to happen. That being said, it might be something we are willing to accept. 

A way around that might be to do the updating of the work packages one at a time (potentially in combination with aquiring [advisory locks as done by the services](https://github.com/opf/openproject/blob/dev/app/services/shared/service_context.rb#L44)). Admittedly, that would be slower to execute.

ulferts (commented):
I only took a look at the update job, @cbliard . 

There is an edge case (as described below) but I think we could live with that one being there.

app/workers/work_packages/update_progress_job.rb
cbliard: Indeed, this is a possibility.

It should then be made more robust.
I'll add it on the TODO.

app/workers/work_packages/update_progress_job.rb
cbliard: That's a valid point. But why would the modification be attributed to the user? The change is in a transaction (one created by the job) and the transaction will only be commited at the end of the job, so if a user changes something, this user will not see the not-yet-commited changes of the job, right?

Maybe I'm wrong and the job does not crete a transaction on its own.
I just checked and it seems the job does not create any transaction. I need to change that then.

Not completely sure of how this could be fully prevented. I'll add it as a TODO.

app/components/work_packages/progress/status_based/modal_body_component.rb
```diff
@@ -0,0 +1,46 @@
...
+# Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+#
+# See COPYRIGHT and LICENSE files for more details.
+# ++
+
+module WorkPackages
+  module Progress
+    module StatusBased
+      # rubocop:disable OpenProject/AddPreviewForViewComponent
+      class ModalBodyComponent < BaseModalComponent
```
klaustopher: just an FYI: You can also add the disable comment at the end of the line containing the offense. Has the advantage that you don't need to enable it again:

```suggestion
      class ModalBodyComponent < BaseModalComponent # rubocop:disable OpenProject/AddPreviewForViewComponent
```

app/controllers/statuses_controller.rb
```diff
@@ -63,6 +63,7 @@ def update
     if @status.update(permitted_params.status)
       flash[:notice] = I18n.t(:notice_successful_update)
       redirect_to action: "index"
+      apply_status_p_complete_change
```
klaustopher:
```suggestion
      apply_status_p_complete_change
      redirect_to action: "index"
```

The `render` or `redirect` call should be the last thing in a controller action.

app/controllers/work_packages/progress_controller.rb
```diff
@@ -0,0 +1,138 @@
...
+  def modal_class
+    if WorkPackage.use_status_for_done_ratio?
+      WorkPackages::Progress::StatusBased::ModalBodyComponent
+    else
+      WorkPackages::Progress::WorkBased::ModalBodyComponent
+    end
+  end
+
+  def set_work_package
+    @work_package = WorkPackage.find(params[:work_package_id])
```
klaustopher: Eventhough the services validate the permissions via the contracts, we should make sure that we always only allow access to `visible` resources.

```suggestion
    @work_package = WorkPackage.visible.find(params[:work_package_id])
```

app/models/changeset.rb
```diff
@@ -231,15 +231,18 @@ def fix_work_package(work_package)
...
-    OpenProject::Hook.call_hook(:model_changeset_scan_commit_for_issue_ids_pre_issue_update,
-                                changeset: self, issue: work_package)
-    if !work_package.save(validate: false) && logger
+    call = WorkPackages::UpdateService
+           .new(model: work_package,
+                user: user || User.anonymous)
+           .call(status:,
+                 journal_notes: I18n.t(:text_status_changed_by_changeset,
+                                       value: text_tag,
+                                       locale: Setting.default_language))
```
klaustopher, line 240: Not for right now, but we should maybe add a ticket that we replace those notes by a cause so we can properly show a translated message

app/services/work_packages/copy_service.rb
```diff
@@ -77,8 +77,7 @@ def copied_attributes(work_package, override)
     attributes = work_package
                    .attributes
                    .slice(*writable_work_package_attributes(work_package))
-                   .merge("parent_id" => work_package.parent_id,
-                          "custom_field_values" => work_package.custom_value_attributes)
```
klaustopher, line 81: `parent_id` is removed here because it is already part of the sliced params?

```ruby
WorkPackages::CopyContract.new(WorkPackage.new, User.first).writable_attributes.include?("parent_id")
#=> true
```

app/workers/work_packages/update_progress_job.rb
```diff
@@ -0,0 +1,257 @@
...
+        unset_all_percent_complete_values if previous_mode == "disabled"
+        fix_remaining_work_set_with_100p_complete
+        fix_remaining_work_exceeding_work
+        fix_only_work_being_set
+        fix_only_remaining_work_being_set
+        derive_unset_remaining_work_from_work_and_p_complete
+        derive_unset_work_from_remaining_work_and_p_complete
+        derive_p_complete_from_work_and_remaining_work
+      end
+      if current_mode == "status"
```
klaustopher:
```suggestion
      elsif current_mode == "status"
```

I don't think we change `current_mode` in the first if block, so the second block can never be executed. This way we will not even need to look at the second if block.

lib_static/plugins/acts_as_journalized/lib/journal_changes.rb
```diff
@@ -31,70 +31,88 @@ def get_changes
...
-      )
-    end
+    changes = [
+      get_cause_changes,
+      get_data_changes,
+      get_attachments_changes,
+      get_custom_fields_changes,
+      get_file_links_changes,
+      get_agenda_items_changes
+    ].compact
```
klaustopher, line 41: Very good: 👏 👏

progress-technical-debt.md
```diff
@@ -0,0 +1,14 @@
...
+* Rename "derived" fields as "total" fields in spec files
+* Remove the leading attribute name in validation messages in popover for work and remaining work
+  * How to: see https://www.bigbinary.com/blog/rails-6-allows-to-override-the-activemodel-errors-full_message-format-at-the-model-level-and-at-the-attribute-level?ref=writesoftwarewell.com
+  * Link to discussion with Parimal: https://matrix.to/#/!abggwvHnVGESiAdLKR:openproject.org/$9ommjHQiGj4lTFpBKOCzcUQpp-leaC5-b7jgisTy9VA?via=openproject.org
+* Harmonize messages for validation of negative work values for work and remaining work
+  * "Work must be >= 0": one we created ourself in https://github.com/opf/openproject/commit/15f52bcdc6cc50c30e14c213133cab4d59715347
+  * "Remaining work must be greater than or equal to 0" is the standard rails message (builtin. no translation needed)
+* In `app/services/work_packages/update_ancestors_service.rb`, `WorkPackage::UpdateAncestorsService#derive_attributes` is updating both `derived_estimated_hours` and `derived_remaining_hours` even when only one needs to be updated. This is because the tests depend on this behavior. (not sure that's worth modifying)
+*
```
klaustopher: I think these should be Code Maintenance tickets

app/components/work_packages/progress/status_based/modal_body_component.rb
cbliard: @aaron-contreras this FYI is for you ;)
Thanks! I'll commit it.

app/controllers/statuses_controller.rb
cbliard: You're totally right. I don't know why I did this.

app/controllers/work_packages/progress_controller.rb
cbliard: Good catch 👍

cbliard: @aaron-contreras

app/models/changeset.rb
cbliard: @ulferts could you address this one?

app/services/work_packages/copy_service.rb
cbliard: Yes, that's something we realized with Jens in a pairing session.
We did not understand why this was here as it seemed it would be picked by the `slice` call above. So we looked at tests, tried with, and without, and as expected the parent is carried over, so we removed it.

progress-technical-debt.md
cbliard: yes, you're right. I'll create one.

app/components/work_packages/progress/status_based/modal_body_component.rb
aaron-contreras: TIL :)

app/workers/work_packages/update_progress_job.rb
cbliard: You're right. I used a case statement instead. Reads better IMO.
8c5c2094919

app/components/work_packages/progress/status_based/modal_body_component.html.erb
```diff
@@ -0,0 +1,36 @@
...
+      <% if should_display_migration_warning? %>
+        <% modal_body.with_row(mt: 3) do |_migration_warning| %>
+          <%= render(Primer::Alpha::Banner.new) { t("work_package.progress.modal.migration_warning_text") } %>
+        <% end %>
+      <% end %>
+
+      <% modal_body.with_row(mt: 3) do |_tooltip| %>
+        <%= render(Primer::Beta::Text.new(font_weight: :semibold)) { t("work_package.progress.label_note") } %>
+        <%= render(Primer::Beta::Text.new) { t("work_package.progress.modal.status_based_help_text") } %>
+        <%= render(Primer::Beta::Link.new(href: "#")) { t(:label_learn_more) } %>
```
aaron-contreras: Do we have a URL already for this?

app/components/work_packages/progress/work_based/modal_body_component.html.erb
```diff
@@ -0,0 +1,36 @@
...
+      <% if should_display_migration_warning? %>
+        <% modal_body.with_row(mt: 3) do |_migration_warning| %>
+          <%= render(Primer::Alpha::Banner.new) { t("work_package.progress.modal.migration_warning_text") } %>
+        <% end %>
+      <% end %>
+
+      <% modal_body.with_row(mt: 3) do |_tooltip| %>
+        <%= render(Primer::Beta::Text.new(font_weight: :semibold)) { t("work_package.progress.label_note") } %>
+        <%= render(Primer::Beta::Text.new) { t("work_package.progress.modal.work_based_help_text") } %>
+        <%= render(Primer::Beta::Link.new(href: "#")) { t(:label_learn_more) } %>
```
aaron-contreras: Same for this

app/components/work_packages/progress/status_based/modal_body_component.html.erb
cbliard: @aaron-contreras as suggested by Oliver, let's remove the links, and add them back once the documentation path is clear.

progress-technical-debt.md
cbliard: ticket created: https://community.openproject.org/projects/openproject/work_packages/54050/activity

app/components/work_packages/progress/status_based/modal_body_component.html.erb
cbliard: @aaron-contreras as suggested by Parimal, it should link to https://www.openproject.org/docs/user-guide/time-and-costs/progress-tracking/.

app/components/work_packages/progress/work_based/modal_body_component.html.erb
cbliard: Same, https://www.openproject.org/docs/user-guide/time-and-costs/progress-tracking/

aaron-contreras: Done!

app/components/work_packages/progress/status_based/modal_body_component.html.erb
aaron-contreras: Done!

ulferts (approved):
Admittedly I did not do a fully fledged review. But the reading and the tests I did all looked positive.

So from my point of view, this can be merged. Thanks @cbliard and @aaron-contreras for bringing this alive.

PR #14769, opened 2024-02-12
[#41299] Updated API spec for placeholder users

Kharonus:
[#41299](https://community.openproject.org/wp/41299)

### WAT??

- added inheritance hirarchy for principal resource models
- added placeholder user paths
- revisited other principal specifications

### Thoughts and stuff

Hey guys, the PR is rather big, with a lot of text. The reason for that is, that placeholder users were not documented at all, yet. And also affect the documentation about users and groups.

I hope, the changes are understandable, yet, there are some open questions:
- there is a group filter for placeholder users, and apparently I can even add it to a group via API. But I wasn't able to do so via UI. What is our intention here?
- There is an internal status set for placeholder users. Yet, it is never returned by the API, leading to a weird situation, that when you delete placeholder users you still get the data in the collection response. You cannot see a difference, as the status is not rendered.


Kharonus:
> there is a group filter for placeholder users, and apparently I can even add it to a group via API. But I wasn't able to do so via UI. What is our intention here?

it is not possible to add placeholder users to a group via API or UI. The filters are probably superfluous, but I didn't remove them in this PR.

> there is a group filter for placeholder users, and apparently I can even add it to a group via API. But I wasn't able to do so via UI. What is our intention here?

I added the status property to the model of placeholder users in the the API response.

lib/api/v3/placeholder_users/placeholder_user_representer.rb
```diff
@@ -55,10 +55,21 @@ class PlaceholderUserRepresenter < ::API::V3::Principals::PrincipalRepresenter
...
+                 setter: ->(fragment:, represented:, **) { represented.status = User.statuses[fragment.to_sym] },
+                 exec_context: :decorator,
+                 render_nil: true,
+                 cache_if: -> { current_user_can_manage? }
+
         def _type
           'PlaceholderUser'
         end
 
+        def current_user_can_see_date_properties?
```
apfohl, line 69: Where is this used? Maybe obsolete?

apfohl (approved):
Very good work. :+1:

lib/api/v3/placeholder_users/placeholder_user_representer.rb
Kharonus: This is used in the `principal_representer` for the `createdAt` and `updatedAt` properties. I had to add some inheritance here, as the permissions for the dates are different from placeholder users and users.

so, for a user it is `admin_or_self`, for a placeholder user it is `user_can_manage` (global permission `:manage_placeholder_user`)

PR #14777, opened 2024-02-13
[52127] Project list: Truncate long text fields and disable expand action

bsatarnejad:
- [x] Limit displaying (project description, project status description, project custom fields/attributes of the type long text) inside the table to one line, truncate with ellipses (...) when necessary
- [x] Add "Expand" link at the end of the truncated text
- [x] Clicking "Expand" will display the full contents of the long text field in a modal
- [x] Disable the expand/collapse action that displays project description
- [x] Remove the last column that contains the action
- [x] Write a test for the changes

https://community.openproject.org/projects/openproject/work_packages/52127/activity


app/components/projects/row_component.rb
```diff
@@ -117,6 +117,14 @@ def status_explanation
       end
     end
 
+    def description
+      return nil unless user_can_view_project?
+
+      if project.description
+        content_tag :div, helpers.format_text(project.description), class: 'wiki'
```
ulferts: I think the `class: 'wiki'` part should be removed. The fact that the other methods in this class also use it is a relic of the past we should fix.

frontend/src/global_styles/content/_table.sass
```diff
@@ -381,3 +381,10 @@ thead.-sticky th
 
   p:last-child
     margin-bottom: 0
+
+.long-text-truncation
+  display: flex
```
ulferts, line 386: This breaks the layout in some cases where multi line custom fields are now displayed in one line with each of them shortened:
image

ulferts (changes requested):
Hi @bsatarnejad ,

it is good to see , that we can remove a bunch of code by this change. However, there are a couple of issues with this PR:

------

* [x] The text of the status description, descriptionText not displayed in only a single line.
image

With tables and images possible being part of the text, this might become quite hard to achieve. I believe there are two options: 
* Remove all tags from the output/render the text without tags
* Limit the height to a single line
Of the two options, I would go with the first because it promises to be simpler to implement and does not run into edge cases like having only an image or a table in the text which might lead to the whole text being removed in case the first option is taken.

------

* [x] The "Expand" is not on the same height as the shortened text:
image

------

* [x] The "Expand" is not always displayed:
image

For this, the height would also have to be taken into account when calculating whether to add the "Expand" element. This problem is of course strongly coupled to the first mentioned problem and might be affected by the solution to it.

------

* [x] Formatting (newlines) not applied in modal:
image

When clicking on "Expand" becomes:

image

Additionally, the modal does not have a header. The need for a header on the modal is not specified but I would think that the user would have an easier time understanding the behaviour if the modal would have a header consisting of the project name and the attribute. But ultimately, that is for product to decide, so better check with @psatyal.

bsatarnejad:
I'm closing this PR because this feature is implemented in another PR.

PR #14781, opened 2024-02-13
Implements the CopyTemplateFolderCommand for OneDrive/Sharepoint

mereghost:
~~Relax... take a deep breath... Now imagine an incredible description for this PR.~~

### What?

This PR intends to mirror the behaviour of `Nextcloud::CopyTemplateFolder` .

### How?

Copy template folder basically copies a folder while renaming it. It is only applied on **AMPF** `ProjectStorage`.

Creating some robust tests required some minor changes that made it necessary to do some retouches on some of the already existing cassettes, which triggered some changes as to make tests good and robust enough to handle cassette changes.

Note that the `OneDrive` request return a 202 and a progress `url` . Said URL has the new `itemId` for the copied data.

### What's really important?

Your mental health. :+1:

But on this PR, mainly the test cases and the `CopyTemplateFolder` logic.


modules/storages/app/common/storages/peripherals/storage_interaction/nextcloud/copy_template_folder_command.rb
```diff
@@ -55,73 +63,65 @@ def validate_inputs(source_path, destination_path)
...
-          Util.error(:conflict, 'Destination folder already exists.')
-        in { status: 401 }
-          Util.error(:unauthorized, "unauthorized (validate_destination)")
-        in { status: 404 }
-          ServiceResult.success(result: urls)
-        else
-          Util.error(:unknown, "Unexpected response (validate_destination): #{response.code}", response)
-        end
+    def remote_folder_does_not_exist?(destination_url)
+      response = OpenProject.httpx.basic_auth(@username, @password).head(destination_url)
```
Kharonus: 🔴 you did remove `@username` and `@password` from 35f. Or do I simply not see, where those variables are written now? Should this be replaced by `@storage.username`?

modules/storages/app/common/storages/peripherals/storage_interaction/one_drive/copy_template_folder_command.rb
```diff
@@ -0,0 +1,91 @@
...
+#
+# See COPYRIGHT and LICENSE files for more details.
+#++
+
+module Storages
+  module Peripherals
+    module StorageInteraction
+      module OneDrive
+        class CopyTemplateFolderCommand
+          def self.call(storage:, source_path:, destination_path:)
```
Kharonus, line 36: 🟡 rename `source_path` to `source_location`? because the usage of the term path looks weird in line 85.

modules/storages/app/common/storages/peripherals/storage_interaction/one_drive/create_folder_command.rb
```diff
@@ -33,24 +33,31 @@ module Peripherals
     module StorageInteraction
       module OneDrive
         class CreateFolderCommand
-          def self.call(storage:, folder_path:)
-            new(storage).call(folder_path:)
+          using ServiceResultRefinements
+
+          def self.call(storage:, folder_path:, parent_location: nil)
```
Kharonus: 🔴 this interface is now no longer similar to the nextcloud command.

I understand, why this is needed: OneDrive cannot derive the parent location from the path of the folder to be created. But the solution would be to just add this to the interface of the nextcloud command.

modules/storages/app/models/storages/project_storage.rb
```diff
@@ -62,11 +62,22 @@ def automatic_management_possible?
 
   def project_folder_path
```
Kharonus: ❓ should we rename that to `automatically_managed_project_folder_path`?

because it is only true for those, not for manual PFs.

Kharonus (changes requested):
I have some thoughts on it.

modules/storages/app/common/storages/peripherals/storage_interaction/nextcloud/copy_template_folder_command.rb
mereghost: Oh, good catch. >.<

modules/storages/app/common/storages/peripherals/storage_interaction/one_drive/create_folder_command.rb
mereghost: It isn't the same, but it was there just so we could add subfolders while setting up tests. The parameter being optional makes the difference invisible to the rest of the application.

I could simplify the test setup to not use subfolders, or create them by hand too or add a special method that can deal with that. WDYT would be the best approach here?

modules/storages/app/common/storages/peripherals/storage_interaction/one_drive/copy_template_folder_command.rb
mereghost: I kept the same signature on `.call`. I could rename the params on `#call` so that it makes more sense.

```ruby
def self.call(storage:, source_path:, destination_path:)
  new(storage).call(source_location: source_path, folder_name: destination_path)
end
```

WDYT?

modules/storages/app/models/storages/project_storage.rb
mereghost: Gooooooood question. `fancy_folder_path_mcfancypaths` sounds a bit long. xD We use this only on the sync jobs and the dependent services.

So we could rename it with little impact overall. Let's say someone invokes this outside of the scope of an AMPF, what would be the most innocuous return?

modules/storages/app/common/storages/peripherals/storage_interaction/one_drive/copy_template_folder_command.rb
Kharonus: I'm fine with either, why I labeled it as yellow :D

personally I'd probably even change `.call` for both, according to the perception, that those commands actually use a "location", not specifically a path.

modules/storages/app/models/storages/project_storage.rb
Kharonus: one option would be to return just that: the path it would(!) have if AMPF is enabled. But then a clear naming would be helpful.

Or we make it sensitive to the project folder mode, returning the real manual folder and `nil` on `inactive`. But there is no use case, so I'd hesitate to put the extra effort

Or, we check for AMPF inside there and fail loudly. That would kind of forbid using this method outside of the correct context.

I tend to option 1 or 3, a bit more leaning towards 1.

modules/storages/app/common/storages/peripherals/storage_interaction/one_drive/create_folder_command.rb
Kharonus: My first intention would be to adapt NC command to also eat this parameter. Maybe even making it non optional. Then you would read the signature as: "you need the storage, where you wanna create - a location identifier - and a parent context, given as a location identifier". For OneDrive we need it all to create any folder anywhere. (useful also for the soon to be implemented feature of creating folders from the location picker).

For nextcloud we basically can ignore the last argument.
Or even better interpret it like:

```ruby
CreateFolderCommand.call(storage:, folder_location: '/Folder', parent_location: '/Parent')
```
is similar in the resulting execution to
```ruby
CreateFolderCommand.call(storage:, folder_location: '/Parent/Folder', parent_location: nil)
CreateFolderCommand.call(storage:, folder_location: '/Parent/Folder', parent_location: '/')
CreateFolderCommand.call(storage:, folder_location: '/Parent/Folder', parent_location: '')
```

Kharonus: You understand my sentiment here?

mereghost: A thing that occurred me is that I could also have it only on `#call`, omitting the main interface. We keep the flexibility for tests and don't mess with the "compatibility".

modules/storages/app/models/storages/project_storage.rb
mereghost: I like 1 more than the others, but what to name the method? xD

modules/storages/app/common/storages/peripherals/storage_interaction/one_drive/copy_template_folder_command.rb
mereghost: I'll leave `.call` as is as it is the interface. We need to standardize a lot of stuff in a bigger code maintenance task, we can do that later.

modules/storages/app/common/storages/peripherals/storage_interaction/one_drive/create_folder_command.rb
Kharonus: Yeah, then lets do that. This PR should no be the place for more wide range refactoring. Instead, I'd really like to keep the interface in common, so that this thingies do not evolve into sleepers.

modules/storages/app/models/storages/project_storage.rb
Kharonus: no big fan of my long name, are ye? :D

so, what about just `managed_project_folder_path`? skipped one word, as a dev you still know that this is a path the system is involved into.

Kharonus (approved):
LGTM, I think we can continue with that.

PR #14788, opened 2024-02-14
Fix seeding on BIM edition

aaron-contreras:
Standardizes the Types and Priorities used in the SharedWorkPackageSeeder so seeding is agnostic of specific records present on either the standard or BIM edition of OpenProject.

### DIFF
--- a/app/seeders/source/seed_data.rb
+++ b/app/seeders/source/seed_data.rb
@@ -43,15 +43,25 @@ class Source::SeedData
     registry[reference] = record
   end
 
-  def find_reference(reference, default: :__unset__)
+  # Finds and returns the value associated with the given reference.
+  #
+  # @param reference [Symbol] The reference to search for.
+  # @param fallbacks [Array<Symbol>] Optional fallback references to search for if the primary reference is not found.
+  # @param default [Object] The default value to return if no reference or fallbacks are found.
+  # @return [Object, nil] The value associated with the reference, or nil if the reference is nil.
+  # @raise [ArgumentError] If no reference or fallbacks are found and no default value is provided.
+  def find_reference(reference, *fallbacks, default: :__unset__)
     return if reference.nil?
 
-    registry.fetch(reference) do
-      if default == :__unset__
-        raise ArgumentError, "Nothing registered with reference #{reference.inspect}"
-      end
-
+    existing_ref = [reference, *fallbacks].find { |ref| registry.key?(ref) }
+    if existing_ref
+      registry[existing_ref]
+    elsif default != :__unset__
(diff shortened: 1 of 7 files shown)


akabiru (commented):
Hey @aaron-contreras , thanks for looking into this- unfortunately there was still the same `ArgumentError` failure on my end. @wielinde and I looked into a somewhat separate approach in https://github.com/opf/openproject/pull/14807 that sets up fallbacks.

However, I see you added tests that could otherwise be useful, should we harmonise the two PRs?

aaron-contreras:
@akabiru I've ran the exact reproduction steps on my machine and there is no `ArgumentError` being raised by doing a `db:seed:replant`. If semantically the Phase type is equivalent to the Epic on the standard edition, I'm all for it but all types and priorities are present on both `standard.yml` and `bim.yml`. Defaulting back to BIM's is okay if so.

I'm curious what you're experiencing as I'm following the exact reproduction steps we found together

akabiru (approved):
@aaron-contreras I must have failed to checkout properly into this branch, as I've just attempted to reproduce the failure and it's already fixed. 😄 Yeah, let's go ahead with this one. We can link it to https://community.openproject.org/wp/52906

aaron-contreras:
We can add a commit to seed with phases instead of epics using the default trick on your PR if that would be the semantically correct type. Let me know :)

akabiru:
> We can add a commit to seed with phases instead of epics using the default trick on your PR if that would be the semantically correct type. Let me know :)

Semantically, we discussed this to be more fitting. We also need to back port the fix to 13.3.x so I would update the base branch

---

ref: https://github.com/opf/openproject/pull/14807/commits/46f9f6a3a3b3ca07fc8d424cdccde3a500c6e31a

akabiru (approved):
:shipit:

app/seeders/source/seed_data.rb
```diff
@@ -43,15 +43,25 @@ def store_reference(reference, record)
...
-
+    existing_ref = [reference, *fallbacks].find { |ref| registry.key?(ref) }
+    if existing_ref
+      registry[existing_ref]
+    elsif default != :__unset__
       default
+    else
+      references = [reference, *fallbacks].map(&:inspect)
+      message = "Nothing registered with #{'reference'.pluralize(references.count)} #{references.to_sentence(locale: false)}"
+      raise ArgumentError, message
```
aaron-contreras, line 64: ⭐

cbliard: yeah, maybe a bit overkill but I like nice messages

PR #14792, opened 2024-02-15
[Op#52605]: Add Loading state to Project Storage OAuth Grant Access nudge modal

akabiru:
### https://community.openproject.org/wp/52605

On "Login" click, transition to loading state; render back on success (failure state to be handle in [#Bug/52798](https://community.openproject.org/wp/52798))

- [x] Update the "Login" button to a form button to preseve semantic A11ly focus & tab benefits. The previous implementaion with a "button link" breaks accessibility as ScreenReader can't tab to it. Also buttons have the advantage of executing via KYBD `Enter` or `Spacebar`
- [x]  On modal load, set focus on the "Login" button as the default action. The user can exit easily via `ESC` or tab back to the two close actions: Close button at the top OR the cancel "I will do it later" button.
- [x] Define loading state as an [`aria-live`](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Live_Regions) region to enable screen readers to announce loading status 

Accessibility References:

_Key A11ly AC:_ Convey when login has started, while it's in progress and when it's completed or failed

* https://primer.style/ui-patterns/loading#accessibility
* https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/button_role
* https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Live_Regions

https://github.com/opf/openproject/assets/17295175/3d8ea32d-92a9-4090-b761-0b96f62e621c

https://github.com/opf/openproject/assets/17295175/6fa46438-dd42-4b51-aa70-e7fcd2fc9424

### DIFF
--- a/frontend/src/stimulus/controllers/dynamic/storages/oauth-access-grant-nudge-modal.controller.ts
+++ b/frontend/src/stimulus/controllers/dynamic/storages/oauth-access-grant-nudge-modal.controller.ts
@@ -42,6 +42,7 @@ export default class OAuthAccessGrantNudgeModalController extends Controller<HTM
 
   static values = {
     closeButtonLabel: String,
+    loadingScreenReaderMessage: String,
   };
 
   declare readonly requestAccessFormTarget:HTMLFormElement;
@@ -53,6 +54,7 @@ export default class OAuthAccessGrantNudgeModalController extends Controller<HTM
   declare readonly hasRequestAccessButtonTarget:boolean;
 
   declare readonly closeButtonLabelValue:string;
+  declare readonly loadingScreenReaderMessageValue:string;
 
   connect() {
     this.element.showModal();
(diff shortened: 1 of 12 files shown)


ba1ash (changes requested):
I played with it locally and that is what I found:
-  I faced a translation problem problem. 


   I assume it is due to https://github.com/opf/openproject/blob/57ec904bbd9182d1b0617b2e6184862a852fdb7c/modules/storages/app/components/storages/admin/oauth_access_grant_nudge_modal_component.rb#L35 being resolved too early. Probably it should be a component method.
- In Chromium I face `Refused to send form data to 'https://openproject.local/projects/30/settings/project_storages/87/oauth_access_grant?utf8=%E2%9C%93' because it violates the following Content Security Policy directive: "form-action 'self' openproject-development.s3.amazonaws.com openproject-development.s3.eu-west-1.amazonaws.com".` after hitting `Login`.
  See warning in https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/form-action. Can we use `window.location.href =` to send user to do login?
- Somehow `Do it later` and cross sign does not work in the initial modal. Is that intentional? I click and nothing happens. See video, please.
- In Chromium I see a modal jump from the top to the center of the page. Probably, we should try to avoid it. See video, please.
-  I wonder if we should initialize nudge modal on storage edit as well. Wdyt?


https://github.com/opf/openproject/assets/11717656/a459558c-33a5-4e83-a2ca-031231ea10f9

ba1ash (approved):
Great work, Kaburu! :+1:

PR #14807, opened 2024-02-19
[#52906] Seeding the BIM edition in a development environment fails

akabiru:
#### https://community.openproject.org/work_packages/52906

In the bim edition `OPENPROJECT_EDITION=bim`, only select types are available. For instance EPIC types are not used in the BIM edition, rather those would be PHASEs.

This PR sets up default fallbacks relevant for the BIM edition.

#### Testing

(1) Enable the BIM edition

```
#.env
OPENPROJECT_EDITION=bim
```

(2) Seed or replant your database

```sh
$ docker compose exec backend bin/rails db:seed:replant
```


akabiru:
Let's go with the first fix in https://github.com/opf/openproject/pull/14788#pullrequestreview-1888426131

PR #14838, opened 2024-02-22
[52127] project list truncate long text fields and disable expand action

bsatarnejad:
- [x] Limit displaying (project description, project status description, project custom fields/attributes of the type long text) inside the table to one line, truncate with ellipses (...) when necessary
- [x]  Add "Expand" link at the end of the truncated text
- [x]  Clicking "Expand" will display the full contents of the long text field in a modal
- [x]  Disable the expand/collapse action that displays project description
- [x]  Remove the last column that contains the action
- [x]  Write a test for the changes

https://community.openproject.org/projects/openproject/work_packages/52127/activity


frontend/src/stimulus/controllers/dynamic/project.controller.ts
```diff
@@ -32,43 +32,17 @@
...
-      clickedProjectRow.classList.toggle('-no-highlighting');
-      clickedProjectRow.classList.toggle('-expanded');
-      projectDescriptionRow.classList.toggle('-expanded');
-
-      this.setAriaLive(projectDescriptionRow);
-    }
-
-    if (otherTrigger) {
-      otherTrigger.focus();
-    }
```
ulferts, line 64: :+1: for removing this part as it is no longer needed.

app/components/projects/row_component.html.erb
```diff
@@ -54,31 +52,5 @@ See COPYRIGHT and LICENSE files for more details.
...
-      data-project-target="descriptionRow"
-      data-project-id="<%= project.id %>">
-    <td></td>
-    <td colspan="<%= columns.length%>" class="project--hierarchy">
-      <div class="description-container wiki">
-        <%= helpers.format_text(helpers.short_project_description(project), project: project) %>
-      </div>
-    </td>
-  </tr>
-<% end %>
```
ulferts, line 84: Way cleaner with all this extra behaviour gone 😃

app/components/projects/row_component.rb
```diff
@@ -57,8 +58,14 @@ def custom_field_column(column)
...
 
-      if cf.field_format == 'text'
-        custom_value.html_safe # rubocop:disable Rails/OutputSafety
+      if cf.field_format == 'text' && custom_value.present?
+
+        concat content_tag :p, Nokogiri::HTML(custom_value).text, class: 'project-long-text', id: "#{project.id}-cf-#{cf.id}"
+        render(Primer::Alpha::Dialog.new(id: "dialog-#{project.id}-cf-#{cf.id}", title: cf.name)) do |component|
+          component.with_show_button(scheme: :link, display: (includes_table(custom_value) ? :block : :none)) { I18n.t('js.label_expand') }
+          component.with_body { helpers.format_text(custom_value.html_safe) } # rubocop:disable Rails/OutputSafety
+          component.with_header(show_divider: true)
```
ulferts: We haven't used the dividers in any of the other dialogs so far. There isn't really a specification in the ticket but I would remove it. Again, there was not specification but just for my taste, I would add some margins to the text and maybe even use a larger dialog to start with:

 
image

It of course depends on how long we expect the texts to become but I fear that they can become longer. On the subject of missing specification, it would be good to also get @psatyal 's opinion.

app/components/projects/row_component.rb
```diff
@@ -112,8 +119,26 @@ def project_status
...
+    def description
+      return nil unless user_can_view_project?
+
+      if project.description.present?
+        concat content_tag :p, Nokogiri::HTML(project.description).text, class: 'project-long-text', id: "#{project.id}-description"
+        render(Primer::Alpha::Dialog.new(id: "dialog-#{project.id}-description", title: I18n.t('activerecord.attributes.project.description'))) do |component|
+          component.with_show_button(scheme: :link, display: (includes_table(project.description) ? :block : :none)) { I18n.t('js.label_expand') }
+          component.with_body { helpers.format_text(project.description) }
+          component.with_header(show_divider: true)
+        end
```
ulferts: The code for displaying the full text in a dialog is repeated three times in this component. Given that the same functionality will be required for displaying the project attributes on the overview page as well, let's extract the functionality into a component that can be reused. This doesn't even have to be project specific but rather a component able to display the full text of a single attribute.

app/components/projects/row_component.rb
```diff
@@ -57,8 +58,14 @@ def custom_field_column(column)
       cf = custom_field(column)
       custom_value = project.formatted_custom_value_for(cf)
 
-      if cf.field_format == 'text'
-        custom_value.html_safe # rubocop:disable Rails/OutputSafety
+      if cf.field_format == 'text' && custom_value.present?
+
+        concat content_tag :p, Nokogiri::HTML(custom_value).text, class: 'project-long-text', id: "#{project.id}-cf-#{cf.id}"
```
ulferts: In the case of a macro being the only content of the field, e.g. an embedded work package table no content is currently displayed. So the user has no possibility to then click on the link to display the table in the Dialog. Granted, I don't know how important doing that actually is. But we have a bug report stating that to be a problem of the former implementation (related to project description) so at least someone used it. 

Would it be possible to, similar like the check for the table, check if there is a macro in there in display the expand button same as done for the tables in this case and some placeholder text within the table?

frontend/src/stimulus/controllers/dynamic/project.controller.ts
```diff
@@ -32,43 +32,17 @@
...
+  connect():void {
+    const longTexts = document.querySelectorAll('.project-long-text');
+    longTexts.forEach((e) => {
+      const isEllipssed= this.isEllipsisActive(e as HTMLElement);
+        if (isEllipssed) {
+          document.querySelectorAll(`[data-show-dialog-id=dialog-${e.id}]`)[0].classList.remove('d-none');
+        }
+    });
```
ulferts: Instead of relying on classes which led me to think about looking for some specific styling you could also use [Targets from Stimulus](https://stimulus.hotwired.dev/reference/targets) for both the text as well as for the button. In combination with moving the controllers domain into the table cell itself, this would make the code for referencing the text and the button very simple.

In other words:
* replace the `ProjectsController` by a controller that is active only within a single table cell, e.g. the project description, status description, ... That means that you would end up with an instance of the controller per cell to be checked.
* have targets on the text as well as on the button.

spec/features/projects/projects_index_spec.rb
```diff
@@ -202,6 +202,21 @@ def expect_projects_in_order(*projects)
...
+          description: 'I am a nice project with a very long long long long long long long long long description'
+        )
+
+        login_as(admin)
+        projects_page.visit!
+        tr_project_development = page.first('tr', text: 'Development project')
+        expect(tr_project_development).to have_css('button', text: 'Expand')
+        tr_project_development.find('button', text: 'Expand').click
+        expect(page).to have_css('.Overlay-body', text: development_project.description)
+      end
```
ulferts: I'd add a case to this test for the other types of fields as well: status explanation and long text custom fields. I'd also add a negative case test in which you check that the "Expand" button is not displayed on shorter text.

ulferts (changes requested):
Hi Behri, we are almost there yet with the desired behaviour. To me, there are a number of open questions related to the displaying of the fields in the modal but that is lack of specification IMO.

There actually is a specification for the project attributes. There, the modal looks like this:

image

I assume that is the expected design for such modals throughout the application (/cc @psatyal )

app/components/projects/row_component.rb
```diff
@@ -27,6 +27,7 @@
 #
 # See COPYRIGHT and LICENSE files for more details.
 #++
+require 'nokogiri'
```
ulferts: This is no longer required here since it has been moved into a component

app/components/projects/row_component.rb
```diff
@@ -163,17 +167,21 @@ def additional_css_class(column)
...
       end
     end
 
     def user_can_view_project?
       User.current.allowed_in_project?(:view_project, project)
     end
+
+    def show_expand_button(text)
+      text.to_s.include?('figure') || text.to_s.include?('macro')
+    end
```
ulferts: This has also been moved into the component now, right?

app/components/projects/table_component.html.erb
```diff
@@ -76,17 +76,7 @@ See COPYRIGHT and LICENSE files for more details.
...
-                   data-application-target="dynamic"
-                   data-params-from-query-all-anchors-value="true"
-                   data-params-from-query-allowed-value='["query_id", "per_page", "page", "sortBy", "filters"]'>
-                <% if params[:expand] == 'all' %>
-                  <%= link_to t(:button_collapse_all) %>
-                <% else %>
-                  <%= link_to t(:button_expand_all), { params: { expand: 'all' } } %>
-                <% end %>
-              </div>
+            <%= render(Primer::Beta::Octicon.new(:gear, size: :small)) %>
```
ulferts: I'd remove the gear icon for now since it is non functional. We'll readd it later when we can trigger a modal by it.

frontend/src/stimulus/controllers/dynamic/attribute.controller.ts
```diff
@@ -0,0 +1,52 @@
...
+  static targets = [
+    'descriptionText',
+    'expandButton',
+  ];
+
+  declare readonly descriptionTextTarget:HTMLParagraphElement;
+  declare readonly expandButtonTarget:HTMLButtonElement;
+
+  connect():void {
+  if (this.isEllipssed(this.descriptionTextTarget)) {
```
ulferts: The indentation is off.

frontend/src/stimulus/controllers/dynamic/attribute.controller.ts
```diff
@@ -0,0 +1,52 @@
...
+export default class AttributeController extends Controller {
+  static targets = [
+    'descriptionText',
+    'expandButton',
+  ];
+
+  declare readonly descriptionTextTarget:HTMLParagraphElement;
+  declare readonly expandButtonTarget:HTMLButtonElement;
+
+  connect():void {
```
ulferts: Just for reference, as I don't think it is better given that the controller is really simple, you could also have used the `descriptionTextTargetConnected` [callback](https://stimulus.hotwired.dev/reference/targets#connected-and-disconnected-callbacks)

ulferts (approved):
This is really good now, @bsatarnejad. the code has become quite simple and is easy to read. Well done.

I have some nitpicking remarks but once you addressed those, feel free to merge the PR.

I think that the display of the "Preview not available" could still be improved by some different styling (we would have to talk to Product about that) and having the "Expand" button right aligned. But this can be done in a separate PR.

frontend/src/stimulus/controllers/dynamic/attribute.controller.ts
bsatarnejad: It's perfect point! Thanks!

PR #14843, opened 2024-02-23
Feature/51670 allow columns to be changed and persisted

ulferts:
**Acceptance criteria**

* [x] The user can choose the columns to displays within a "Configure view" modal
    * [x] The modal can be triggered from the more menu
    * [x] All currently available columns can be selected and their order changed
    * [x] The "name" column is always displayed. It cannot be deselected
    * [x] The `DraggableAutocompleteComponent` is used for this, which is also used for the column selection on the WP page. The component is restyled a bit. It receives additional labels. This change is reflected in both places. 
* [x] When persisting lists by a logged in user, the selected columns are persisted as well including the order the columns are in.
    * [x] When opening the persisted list, the table shows the selected columns in the order the user selected them to be in when persisting.
* [x] In the administration (admin/settings/projects), adapt the setting for selecting the columns displayed in a project list to the same UI used within the modal for selecting columns for a specific list.
  * This allows to also select the order the columns are in by default
  * Deletion of the 'Name' column is to be prevented.
* [x] In the administration (admin/settings/work_packages), adapt the setting for selecting the columns displayed in a work package list to the same UI used within the modal for selecting columns for a specific list.
  * This allows to also select the order the columns are in by default
  * Deletion of the 'Id' and the 'Subject' column is to be prevented.
* [x] When exporting the list of projects (PDF and XLS), only the selected columns are exported
* [x] Already persisted queries receive the currently active default columns.
* [x] Document form dialogs in lookbook.

**Notes**

The PR introduces the concepts of `selects` to be rendered to a query. This follows the same pattern as `filter` and `order` registration. The name `select` was chosen over e.g. `column` for two reasons: 
* On the layer of the query, a model, the projection/selection of the (SQL) query closer than the UI perspective of it potentially being a column.
* On the UI layer, it will not always be a column that is mapped to it. E.g. when rendering graphs, the same selects can be used.

The `selects` are for now rather passive. They only signal which selects/columns can be chosen. In a subsequent PR, the plan is to have them called when fetching the result of the query so that only the necessary selects are carried out.

On the work package query (`Query`) the columns are also renamed to selects. There is no functional change there yet, but it should make it simpler to harmonize the `Query` object with the rest of the [Model]Query implementation later on.

The `DraggableAutocompleter` component is used within the frontend to select the columns. That component has been turned into a WebComponent to achieve this. It was only used inside the column administration of the work package list configuration modal. It is now also applied to the configuration modal of project lists as well as to the administration pages of both project and work packages. There, it currently looks a bit off, as it has a vertical layout. But it is better than having some longer list of checkboxes. The administration layout is specified to change anyway and then those configuration settings would have their own page making the weirdness disappear.

-----

https://community.openproject.org/wp/51670


app/models/queries/projects/selects/custom_field.rb
```diff
@@ -1,6 +1,6 @@
-#-- copyright
+# -- copyright
```
klaustopher, line 1: I guess Git/GitHub did mess something up detecting a rename of the file here, right?

app/views/projects/_project_export_modal.html.erb
```diff
@@ -31,7 +31,7 @@ See COPYRIGHT and LICENSE files for more details.
     data-controller="params-from-query"
     data-application-target="dynamic"
     data-params-from-query-all-anchors-value="true"
-    data-params-from-query-allowed-value='["query_id", "filters", "sortBy"]'>
+    data-params-from-query-allowed-value='["query_id", "filters", "sortBy", "columns"]'>
```
klaustopher, line 34: Already stumbled upon this a few times, but why are we not consistently using `selects` instead of `columns`. I guess it's for keeping the current saved links, etc intact? 

Maybe we can document somewhere, where the boundary between calling it `columns` and `selects` is and why this was chosen.

db/migrate/20240222155909_rename_columns_on_project_queries.rb
```diff
@@ -0,0 +1,21 @@
+class RenameColumnsOnProjectQueries < ActiveRecord::Migration[7.1]
+  def change
+    rename_column :project_queries, :columns, :selects
```
klaustopher, line 3: Just to note, this will change a column that is actively used by code before the migration runs and can cause errors. Just to be aware of this, that running the migration needs the application to be shut down before they are executed.

klaustopher (approved):
Looked through the code thoroughly and checked the branch out. Looks good and sensible. I have made some comments inline.

One thing I also realized: Do we know if people out there have maybe added custom columns in plugins they have written? If so, by introducing this PR we will be breaking that functionality and require them to change the base classes for their work package columns, etc. Is this something we want to do in 13.4 release?

HDinger (commented):
Hi @ulferts 

looks good overall, I just have some minor remarks which aim for harmonisation with [a PR](https://github.com/opf/openproject/pull/14921) I just created recently (and is not merged yet).

* Could you please add a number in front of the file name (such as "01-dialogs.md.erb")? Thus we avoid, that the page moves below the "Deprecated" folder and thus seems less important.
* I am wondering whether we really want to show the Dialog preview as a separate example in the lookbook. My suggestion would be to hide that from the left side menu similar to how I did it [here](https://github.com/opf/openproject/pull/14921/files#diff-12013a99c4badaa4a56c4ec8daa82b52d34b4e2020654010d9b7d8d77b363259R3). If you want to show details (like the source or the params tab), you could also show them in the md.erb file , like [this](https://github.com/opf/openproject/pull/14921/files#diff-fa6df4bb7f6f25bab1f6eff399dad72966d1d09eb84b7a0319527c6761e00f08R5). In my eyes, this is better since we can avoid a third place where Dialogs are presented (especially since there is no indication that the Dialog Preview in the OpenProject folder is only about forms inside a dialog). What do you think?

ulferts:
@HDinger , thanks for looking into it.

I chose the magic number of 05 now for no particular reason other than I wanted to have some space left before. We could also just start with more spaces and employ three digit numbers here. 

The preview part is now hidden. I like that change. 

When looking into the referenced PR, I was wondering if what is described there isn't a special case of a Dialog and should be moved to the Dialog page. Because in the end, more examples for dialogs will end up in that page anyway. I plan to add one on confirmation dialogs once I start to work on the next feature where I am going to need one.

PR #14844, opened 2024-02-23
debounce requests to recalculate budget after updating count

toy:
Checked due to flaky test in modules/budgets/spec/features/budgets/update_budget_spec.rb. Problem was with the race of two requests (update to 0 then to value) triggered by fill_in in test. Debouncing will not only ensure test doesn't fail randomly, but also remove unnecessary requests in production.

Why simple operation of multiplying count by item cost requires a call to backend?


ulferts (approved):
Good improvement, @toy . I meant to add a debouncer myself but then forgot about it. Good that you did the cleanup here 👍

PR #14845, opened 2024-02-23
Introduce ManagedFolderIdentifier (and a dash of other stuff)

mereghost:
## What?

`ProjectStorage` has the need to identify the folder name of an AMPF enabled project and its location, which can differ from the name/path. To make matters worse both are provider dependent.

This PR aims to address this.

## How?

I introduced the `ManagedFolderIdentifier` namespace, with implementations named after the 2 storage providers we have.

This also triggered a rework of the `Registry` keys.

## Who cares?

Me! :D I was finding the keys with `thing` as the main component kinda weird since we expanded the use of the `Registry` to more than commands and queries.

Should you care? SURE! I'd love to hear your thoughts on it.


ba1ash (approved):
Looks good to me! Well done @mereghost :muscle:

PR #14846, opened 2024-02-23
[Op#53021] Define Gitlab CE local docker compose setup with TLS

akabiru:
## GitLab CE Service

Within `docker/dev/gitlab` a compose file is provided for running local Gitlab instance with TLS support. The provides
a production like environment for testing the OpenProject GitLab integration against a community edition gitlab instance
accessible on `https://gitlab.local`.

> NOTE: Configure [TLS Support](https://www.openproject.org/docs/development/development-environment-docker/#tls-support) first before starting the GitLab service

See [Install GitLab using Docker Compose](https://docs.gitlab.com/ee/install/docker.html#install-gitlab-using-docker-compose)
official GitLab documentation.

### Running the GitLab Instance

Start up the docker compose service for gitlab as follows:

```shell
docker compose --project-directory docker/dev/gitlab up -d
```

### Initial password

Once the GitLab service is started and running, you can access the initial `root` user password as follows:

```shell
docker compose --project-directory docker/dev/gitlab exec -it gitlab grep 'Password:' /etc/gitlab/initial_root_password
```

Screenshot 2024-02-27 at 6 12 53 PM

### DIFF
--- a/docker/dev/gitlab/docker-compose.yml
+++ b/docker/dev/gitlab/docker-compose.yml
@@ -1,0 +1,37 @@
+version: "3.9"
+
+services:
+  gitlab:
+    image: gitlab/gitlab-ce:latest
+    restart: no
+    volumes:
+      - gitlab-etc:/etc/gitlab
+      - gitlab-logs:/var/log/gitlab
+      - gitlab-opt:/var/opt/gitlab
+      # Linux
+      - /etc/ssl/certs:/etc/ssl/certs:ro
+      - /usr/local/share/ca-certificates:/usr/local/share/ca-certificates:ro
+      # In case of MacOS, you need to mount the certs from the host machine
+      # having them previously bundled via `sudo update-ca-certificates`
+      #
+      # - ~/.step/certs:/etc/ssl/certs
+      # - ~/.step/certs:/usr/local/share/ca-certificates
+    networks:
+      - external
+    extra_hosts:
+      - "openproject.local:host-gateway"
+    labels:
+      - "traefik.enable=true"
+      - "traefik.http.routers.gitlab.rule=Host(`gitlab.local`)"
+      - "traefik.http.routers.gitlab.entrypoints=websecure"
+      - "traefik.http.services.gitlab.loadbalancer.server.port=80"
+
+volumes:
+  gitlab-etc:
+  gitlab-logs:
+  gitlab-opt:
+
+networks:
+  external:
+    name: gateway
+    external: true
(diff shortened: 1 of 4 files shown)


docs/development/development-environment-docker/README.md
```diff
@@ -394,6 +394,33 @@ docker compose --project-directory docker/dev/tls down
 docker compose --project-directory docker/dev/tls up -d
 ```
 
+## GitLab CE Service
+
+Within `docker/dev/gitlab` a compose file is provided for running local Gitlab instance with TLS support. The provides
```
akabiru:
```suggestion
Within `docker/dev/gitlab` a compose file is provided for running local Gitlab instance with TLS support. This provides
```

docs/development/development-environment-docker/README.md
```diff
@@ -394,6 +394,33 @@ docker compose --project-directory docker/dev/tls down
...
+a production like environment for testing the OpenProject GitLab integration against a community edition gitlab instance
```
akabiru:
```suggestion
a production like environment for testing the OpenProject GitLab integration against a community edition GitLab instance
```

docker/dev/gitlab/docker-compose.yml
```diff
@@ -0,0 +1,37 @@
...
+      # - ~/.step/certs:/usr/local/share/ca-certificates
+    networks:
+      - external
+    extra_hosts:
+      - "openproject.local:host-gateway"
+    labels:
+      - "traefik.enable=true"
+      - "traefik.http.routers.gitlab.rule=Host(`gitlab.local`)"
+      - "traefik.http.routers.gitlab.entrypoints=websecure"
+      - "traefik.http.services.gitlab.loadbalancer.server.port=80"
```
mereghost, line 27: Stupid question: why we need a new service and loadBalancer defined?

mereghost (approved):
Seems good. Just one question as I don't know the details of the gitlab ce container.

docker/dev/gitlab/docker-compose.yml
akabiru: @apfohl has a better handle on this but my (very basic) understanding (see https://github.com/opf/openproject/pull/14846#discussion_r1509092502) ~is a dedicated load balancer service for `gitlab` is able to designate traffic to one or more gitlab instances- in dev we'll mainly just use one instance. I reckon it's also better to have split traffic between other sservices~

apfohl: It's way easier then you might think. In traefik routing, the loadbalancer stage is always involved. And it is also the point where you can configure to which port the routing should go. Because the container is exposing three ports it will use the lowest by default which is 22. Therefore it's necessary to tell the loadbalancer stage to use port 80. ;)

apfohl: And it's not a separate service.

PR #14853, opened 2024-02-26
Make open project running under Ruby 3.3

Eric-Guo:
Just fix [Anonymous block parameter](https://rubyreferences.github.io/rubychanges/3.3.html#anonymous-parameters-forwarding-inside-blocks-are-disallowed) is also used within block (SyntaxError)

### DIFF
--- a/app/components/op_primer/component_helpers.rb
+++ b/app/components/op_primer/component_helpers.rb
@@ -28,31 +28,31 @@
 
 module OpPrimer
   module ComponentHelpers
-    def flex_layout(**, &)
-      render(Primer::OpenProject::FlexLayout.new(**), &)
+    def flex_layout(**, &block)
+      render(Primer::OpenProject::FlexLayout.new(**), &block)
     end
 
-    def grid_layout(css_class, **, &)
-      render(Primer::OpenProject::GridLayout.new(css_class:, **), &)
+    def grid_layout(css_class, **, &block)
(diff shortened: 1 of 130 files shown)


klaustopher:
Hey @Eric-Guo, looking into the issue a bit more, there has been a discussion in the [Rubocop repo](https://github.com/rubocop/rubocop/issues/12571) and the current errors that are shown are due to a bug in Ruby 3.3.0 and this will be fixed in 3.3.1. See https://bugs.ruby-lang.org/issues/20090. Issues like those are why we have not upgraded the Ruby version for OpenProject yet.

As soon as 3.3.1 is out, we will re-evaluate if we can switch over to the new version.

Could I ask what your specific use case is to force OP to run under 3.3?

Eric-Guo:
Not yet, the reason to change is just because I using Mac and brew to install ruby and ruby version is 3.3, without this PR, I can not start up the OP in my local env. 😃

ulferts:
Hi @Eric-Guo,

most people on a Mac developing OpenProject [use something like rbenv](https://www.openproject.org/docs/development/development-environment-osx/#install-ruby) for managing the ruby version. Would that be an option for you in the meantime?

Regards
Jens

Eric-Guo:
Yeah, but I'm still prefer one ruby version for all (10+ rails) projects I maintained, anyway I'm glad to known openproject know this issue better than me.

PR #14856, opened 2024-02-26
[52920] Wrapping of URLs in lists in dynamic meetings

bsatarnejad:
https://community.openproject.org/projects/openproject/work_packages/52920/activity


dombesz (approved):
LGTM! 👍I believe the base branch needs to be changed to `dev` before merging, as the release is already closed.
Note: In case we are going to have a `13.3.2`, then we have to cherry pick this fix to the release branch.

PR #14863, opened 2024-02-26
Fix #43520 Search bar doesn't have focus state and the first element on the list seams always selected

toy:
| was | now |
| --- | --- |
| image | image |
| image | image |

Related to #8225


marcalcobe (approved):
From a Design perspective things look good!

HDinger (approved):

PR #14864, opened 2024-02-26
[#52292] OpenProject behind prefix some assests still loaded from webroot

dominic-braeunlein:
https://community.openproject.org/work_packages/52292

Remove `/` from `frontend_asset_path` to allow for url prefix to take effect.


machisuji (changes requested):
Thanks for the fix! Looks good to me. However, do you think we could add a spec for this?

dominic-braeunlein:
@machisuji Good point. I will add a test

ba1ash:
@machisuji have an extra look, pls.

machisuji (approved):
👍

PR #14877, opened 2024-02-28
Fixes for shared filter

klaustopher:
- You could not properly use the *Shared with me* filter because if you did not have the `:view_shared_work_packages` on the project. This is fixed by checking if you are using the `me` value and if so, it will remove those checks.
- Found another thing that we tried to load a list of shares when there are no work packages to be displayed, and this request failed with a  `400` error

Fixes https://community.openproject.org/work_packages/53071


ulferts (changes requested):
The filter is now available as desired and also filtering for the me value works as expected. 

Where the PR fails is that it still offers more values (all project members or shared with users) and selecting any of those other values will lead to an error:

image

The error is even displayed if not the 'me' value but rather the '[first name] [last name]' value of the current user is selected.

We would have to adapt the selection to only allow the 'me' value (and the users own name) in such a case. That would require to use a different filtering for the principals. It may be possible to dynamically change the filters on the principal but I am not sure of this.

klaustopher:
Shouldn't that be handled by the `available_values`? How would we go about presenting this to the frontend?

app/models/queries/work_packages/filter/shared_with_me_filter.rb
```diff
@@ -0,0 +1,67 @@
...
+    I18n.t('query_fields.shared_with_me')
+  end
+
+  private
+
+  def shared_work_packages
+    WorkPackage
+      .joins("JOIN members ON members.entity_type = 'WorkPackage' AND members.entity_id = work_packages.id")
+      .where(members: { user_id: User.current.id })
+      .select('work_packages.id')
```
ulferts: The whole SQL could be done within the members table:

```
Member
  .where(entity_type: WorkPackage.to_s)
  .where(principal: User.current)
  .select(:entity_id)
```

Just as it is already done but other than what I thought initially, I would not limit this subselect to the query's `#project`, because subprojects might be desired as filter results as well.

app/models/queries/work_packages/filter/shared_with_me_filter.rb
```diff
@@ -0,0 +1,67 @@
...
+# Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+#
+# See COPYRIGHT and LICENSE files for more details.
+# ++
+
+class Queries::WorkPackages::Filter::SharedWithMeFilter < Queries::WorkPackages::Filter::WorkPackageFilter
+  include Queries::Filters::Shared::BooleanFilter
+
+  def available?
+    User.current.members.of_any_entity.any?
```
ulferts: This should differentiate between the two cases of the filter (via it's query) being executed inside a project or globally. A work package filter has access to the `#project` the query is in so that can be used here.

ulferts (changes requested):
The structure of the PR looks good.

When adding the filter in the frontend, it is added in a weird way:

image

The filter looks to be in the "false" state and the results are grayed out. The user needs to first toggle the filter to have it become active and the filter results updated. The filter should be added in the "true" state and be applied right away.

klaustopher:
> The filter looks to be in the "false" state and the results are grayed out. The user needs to first toggle the filter to have it become active and the filter results updated. The filter should be added in the "true" state and be applied right away.

How can this be achieved with the filter logic?

ulferts (approved):
👍 nice change to the boolean filter. Feels better that way.

PR #14879, opened 2024-02-28
Implementation/50915 send an email with the information and how to fix it

apfohl:
https://community.openproject.org/projects/openproject/work_packages/50915/activity

This PR introduces sending of notification mails for the storages health status.

- The storage jobs are now sending two kind of events:
  - `STORAGE_TURNED_UNHEALTHY`
  - `STORAGE_TURNED_HEALTHY`
- Events are received via storage engine and passed to the new `HealthService`
- `HealthService` handles send logic and triggers new `StoragesMailer`


config/environments/development.rb
```diff
@@ -109,6 +109,7 @@
 
   # Send mails to browser window
   config.action_mailer.delivery_method = :letter_opener
```
ba1ash: This line seems to be obsolete in favor of the next one.
```suggestion
```

config/environments/development.rb
```diff
@@ -109,6 +109,7 @@
...
+  config.action_mailer.delivery_method = File.exist?('/.dockerenv') ? :letter_opener_web : :letter_opener
```
ba1ash: Why not to set it to `letter_opener_web` for docker based dev env and not docker based? Then it would be simpler and dev experience will be same.

modules/storages/app/services/storages/health/health_service.rb
```diff
@@ -0,0 +1,76 @@
...
+#
+# You should have received a copy of the GNU General Public License
+# along with this program; if not, write to the Free Software
+# Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+#
+# See COPYRIGHT and LICENSE files for more details.
+#++
+
+module Storages
+  module Health
```
ba1ash: I suggest removing this namespace and have just 'Storages::HealthService`

modules/storages/app/workers/storages/health_status_mailer_job.rb
```diff
@@ -0,0 +1,53 @@
...
+
+module Storages
+  class HealthStatusMailerJob < ApplicationJob
+    def perform(admins:, storage:)
+      if storage.health_healthy?
+        return
+      end
+
+      admins.each do |admin|
+        Storages::StoragesMailer.notify_unhealthy(admin, storage).deliver_now
```
ba1ash: Probably, should be `deliver_later`. What happens to other deliveries if one `deliver_now` fails with Timeout, for instance?

modules/storages/app/workers/storages/health_status_mailer_job.rb
```diff
@@ -0,0 +1,53 @@
...
+# You should have received a copy of the GNU General Public License
+# along with this program; if not, write to the Free Software
+# Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+#
+# See COPYRIGHT and LICENSE files for more details.
+#++
+
+module Storages
+  class HealthStatusMailerJob < ApplicationJob
+    def perform(admins:, storage:)
```
ba1ash: I assume it is not correct to pass predefined set of admins as a parameter to the job, because job scheduling can happen on one day and execution on the next one, but during this time a set of admins can easily change(user has been promoted to admin, admin has been removed from the system etc.). Due to that I suggest query admins at the job runtime.

modules/storages/app/workers/storages/health_status_mailer_job.rb
```diff
@@ -0,0 +1,53 @@
...
+      if storage.health_healthy?
```
ba1ash: It should be check that storage is actually present in the database here. Because it can be removed after the job has been scheduled.

modules/storages/app/services/storages/health/health_service.rb
```diff
@@ -0,0 +1,76 @@
...
+        admin_users.each do |admin|
+          ::Storages::StoragesMailer.notify_healthy(admin, @storage, reason).deliver_later if was_unhealthy
+        end
+      end
+
+      def unhealthy(reason:)
+        last_reason = @storage.health_reason
+        @storage.mark_as_unhealthy(reason:)
+
+        if @storage.health_reason != last_reason
```
ba1ash: It seems like we have similar logic for checking whether health_reason is same or not inside `Storages::Storage` model. Therefore it seems like a duplication. Should we try to do it in one place?

modules/storages/app/services/storages/health/health_service.rb
```diff
@@ -0,0 +1,76 @@
...
+      end
+
+      def healthy
+        was_unhealthy = @storage.health_unhealthy?
+        reason = @storage.health_reason
+
+        @storage.mark_as_healthy
+
+        admin_users.each do |admin|
+          ::Storages::StoragesMailer.notify_healthy(admin, @storage, reason).deliver_later if was_unhealthy
```
ba1ash: We have a check inside `Storages::Storage#mark_as_healthy` if it was healthy before. For me it seems like a duplication. Does not it?

ba1ash (changes requested):
- `/letter_opener` page needs inline scripts execution to be allowed by CSP. I beleive this leads to some UI functions being broken.
letter_opener_content_security

config/environments/development.rb
apfohl: I don't exactly get what you want. Right now it's `:letter_opener_web` for Docker and the normal `letter_opener` for normal dev setup.

modules/storages/app/services/storages/health/health_service.rb
apfohl: Then I would have to move this file one level up... I don't like this.

modules/storages/app/workers/storages/health_status_mailer_job.rb
apfohl: good point

HDinger (approved):
LGTM 👍

modules/storages/app/workers/storages/health_status_mailer_job.rb
apfohl: I also added a reload for the object, to see if it actually still has an error.

modules/storages/app/services/storages/health/health_service.rb
apfohl: The `HealthService` itself is now in charge of setting the state on the storage. I know that `mark_as_healthy` is doing a check, but it is too transparent from the outside to see, if it was unhealthy before. I need this to determine if I need to send mails or not. 

We could think about something like a `finalize` method to call a block if it has changed or something like that. But for now I think this would just make it more complicated.

My favorite thing would be, to extract this logic from the model completely, because it's not the right this to handle this invariant inside a data class, which main purpose should be writing consistent data to the database an obtaining it again.

modules/storages/app/services/storages/health/health_service.rb
apfohl: Yep, but it's private. I don't see a need to change the API for the `Storage` model here. It already has too much logic in it.

config/environments/development.rb
ba1ash: My point is to have only one thing `:letter_opener_web` and do not bother differentiating development environments.

apfohl: Shall I make it the only default? Someone will find this unexpected....

ba1ash: True. I do not think it is a big problem. I am under the impression that emails are rarely touched. So, I would just change it and announce then it has been changed. But let's have it like that for now. Could we have a ticket to get rid of `letter_opener` later then?

modules/storages/app/workers/storages/health_status_mailer_job.rb
```diff
@@ -0,0 +1,62 @@
...
+# along with this program; if not, write to the Free Software
+# Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+#
+# See COPYRIGHT and LICENSE files for more details.
+#++
+
+module Storages
+  class HealthStatusMailerJob < ApplicationJob
+    def perform(storage:)
+      return unless Storages::Storage.exists?(storage.id)
```
ba1ash, line 43: While testing I see the following error:

image

I scheduled the job, then removed storage, then started worker. 

So to prevent error properly we should pass `storage_id` and find storage at the beginning of `perform`.
The second option would be get back to previous implementation and accept raising of `ActiveJob::DeserializationError` which then removes job from the que and no emails are sent.

modules/storages/app/services/storages/health/health_service.rb
apfohl: Health logic moved to `HelthService`.

Gemfile.lock
```diff
@@ -428,8 +428,7 @@ GEM
       descendants_tracker (~> 0.0.1)
     color_conversion (0.1.1)
     colored2 (4.0.0)
-    commonmarker (1.0.4)
-      rb_sys (~> 0.9)
+    commonmarker (1.0.4-x86_64-linux)
```
ba1ash: I think we should not have linux specific version of the gem here. Let's just keep 1.0.4.

Gemfile.lock
```diff
@@ -747,8 +749,7 @@ GEM
     net-smtp (0.4.0.1)
       net-protocol
     nio4r (2.7.0)
-    nokogiri (1.16.3)
-      mini_portile2 (~> 2.8.2)
+    nokogiri (1.16.3-x86_64-linux)
```
ba1ash: same here.

Gemfile.lock
```diff
@@ -1155,7 +1155,7 @@ GEM
     zeitwerk (2.6.13)
 
 PLATFORMS
-  ruby
+  x86_64-linux
```
ba1ash: Let's get back to ruby.

modules/storages/app/services/storages/health_service.rb
```diff
@@ -0,0 +1,100 @@
...
+        ::Storages::StoragesMailer.notify_unhealthy(admin, @storage).deliver_later
+      end
+    end
+
+    def admin_users
+      User.where(admin: true)
+          .where.not(mail: [nil, ""])
+    end
+
+    def schedule_mail_job(storage)
```
ba1ash, line 90: I do not see where this method is called. Then it is not clear how `::Storages::HealthStatusMailerJob` gets its way to the queue.

modules/storages/app/services/storages/health_service.rb
```diff
@@ -0,0 +1,100 @@
...
+      ::Storages::HealthStatusMailerJob.schedule(storage:)
+    end
+
+    def mail_job_exists?
```
ba1ash: This method seems to be obsolete due to GoodJob being merged to dev branch. It seems to be unused as well.

Gemfile.lock
apfohl: Oh, when did that happen....

modules/storages/app/services/storages/health_service.rb
apfohl: Oh damn, something went wrong. Some code is missing.... need to check.

apfohl: @ba1ash Can you tell me how I should act here.

mereghost: You can probably rely on GoodJob concurrency settings. Take a look at `ManageNextcloudIntegrationJob.good_job_control_concurrency_with` to handle that.

PR #14880, opened 2024-02-28
[#32940] Unsetting a user value in bulk is not possible

dombesz:
https://community.openproject.org/work_packages/32940

### DIFF
--- a/spec/features/work_packages/bulk/update_work_package_spec.rb
+++ b/spec/features/work_packages/bulk/update_work_package_spec.rb
@@ -276,8 +276,8 @@ RSpec.describe 'Bulk update work packages through Rails view', :js, :with_cuprit
 
         it 'clears the chosen values' do
           # Required fields should not have a 'none' option
-          expect(page).to have_no_select(required_boolean_cf.name, options: ['none'])
-          expect(page).to have_no_select(required_list_cf.name, options: ['none'])
+          expect(page).to have_no_select(required_boolean_cf.name, with_options: ['none'])
+          expect(page).to have_no_select(required_list_cf.name, with_options: ['none'])
 
           # Unset any non-required fields
           select 'none', from: boolean_cf.name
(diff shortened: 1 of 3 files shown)


spec/features/work_packages/bulk/update_work_package_spec.rb
```diff
@@ -204,6 +204,101 @@
...
+          end
+
+          refresh
+          wait_for_reload
+        end
+
+        it 'clears the chosen values' do
+          # Required fields should not have a 'none' option
+          expect(page).to have_no_select(required_boolean_cf.name, options: ['none'])
+          expect(page).to have_no_select(required_list_cf.name, options: ['none'])
```
aaron-contreras: These expectations could lead to false positives. The `options` attribute allows you to specify **exact** matches. Meaning it's trying to ensure there is no required select field with _solely_ the 'none' option (which there never will be) even if this behavior later changes to explicitly be that.

This is visible if you try:
```ruby
expect(page).to have_select(required_boolean_cf.name, options: ['none'])
```

you'll see the failure output:
```
     Failure/Error: expect(page).to have_select(required_boolean_cf.name, options: ['none'])
       expected to find visible select box "Required Bool CF" that is not disabled with options ["none"] but there were no matches. Also found "(No change)\nyes\nno", which matched the selector but not all filters. Expected options ["none"] found ["(No change)", "yes", "no"]
```

A more precise matcher for this case would be `with_options` which tests against *partial* matches. Meaning it will assert there are no required select fields with a 'none' option included in their option set.

If we then try:
```ruby
expect(page).to have_select(required_boolean_cf.name, with_options: ['none'])
```

you'll see:
```
     Failure/Error: expect(page).to have_select(required_boolean_cf.name, with_options: ['none'])
       expected to find select box "Required Bool CF" that is not disabled with at least options ["none"] but there were no matches
```

I'd suggest the `with_options` matcher instead to ensure no false positives :)

```suggestion
          expect(page).to have_no_select(required_boolean_cf.name, with_options: ['none'])
          expect(page).to have_no_select(required_list_cf.name, with_options: ['none'])
```

aaron-contreras (approved):
Other than a possible false positive in the specs down the line, this is great! Merge once addressed 🏎️

aaron-contreras:
Nice!

PR #14889, opened 2024-02-29
Fix for multi-value CF "is (AND)" and "is not" filtering

virgo47:
This should fix https://community.openproject.org/projects/openproject/work_packages/53198/activity
"Multi-value custom fields filtering with "is (AND)" and "is not" is affected by values in other CFs"

### DIFF
--- a/app/models/queries/filters/shared/custom_fields/list_optional.rb
+++ b/app/models/queries/filters/shared/custom_fields/list_optional.rb
@@ -59,6 +59,7 @@ module Queries::Filters::Shared
 
         operator_strategy.sql_for_customized(
           values_replaced,
+          custom_field.id,
           Arel.sql(customized_model.name),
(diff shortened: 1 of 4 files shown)


ulferts:
Hi @virgo47, and thanks for providing yet another fix. The change looks good but to safeguard your work, a spec for it would be needed. 

Please add one to the [appropriate file](https://github.com/opf/openproject/blob/dev/spec/models/query/results_custom_field_filter_integration_spec.rb)

virgo47:
Hi @ulferts, I expected that :-) but I wanted to see whether the rest of the tests don't show some inconsistency. I was already working on a separate file, hope that is fine too, it goes through the browser and both of the tests would show a failure of either "is (AND)" or "is not". I couldn't locate the fix you linked, but if it is necessary, I can redo the tests there.

ulferts:
@virgo47 works for me as well. Thanks again.

PR #14892, opened 2024-03-01
Introduces the CopyProjectFolderJob

mereghost:
## What?

Due to the need of polling for OneDrive we have split the copy and file link setup process from the main `ProjectCopy` service.

## How?

I introduced a new job and a couple of new services, deprecating the already existing ones.

More details, tonight at OpenProject News. 🤣


modules/storages/app/common/storages/peripherals/storage_interaction/one_drive/copy_template_folder_command.rb
```diff
@@ -62,9 +62,7 @@ def call(source_location:, destination_name:)
           def handle_response(response)
             case response
             in { status: 202 }
-              id = extract_id_from_url(response.headers[:location])
-
-              ServiceResult.success(result: { id:, url: response.headers[:location] })
+              ServiceResult.success(result: { id: nil, url: response.headers[:location] })
```
Kharonus: 🟡 can we skip the property `id` if it is always `nil` when using the copy command of OneDrive?

modules/storages/app/services/projects/copy/storage_project_folders_dependent_service.rb
```diff
@@ -46,6 +46,12 @@ def source_count
     def copy_dependency(*)
       return unless state.copied_project_storages
 
+      GoodJob::Batches.enqueue(on_success: NotifyCopyCompletedJob, project_storages: state.copied_project_storages) do
+        state.copied_project_storages.each do |project_storages|
+          ::Storages::CopyProjectFoldersJob.enqueue(project_storages[:source], project_storages[:target])
+        end
+      end
```
Kharonus: 🔴 Did we talk about GoodJob not being part of 13.4 today in the daily? If yes, we cannot use it to enqueue batches. Am I missing something from the communication?

modules/storages/app/services/projects/copy/storage_project_folders_dependent_service.rb
Kharonus, line 1: 🟡 Should most of the code get removed? As we have this service now as a "non-dependent" service? With the dependent service being "detached" from `/projects/copy_service.rb`, this is now dead code, isn't it?

Same for file links dependent service

modules/storages/app/services/storages/file_links/copy_file_links_service.rb
```diff
@@ -0,0 +1,111 @@
...
+    class CopyFileLinksService
+      def self.call(source:, target:, user:, work_packages_map:)
+        new(source:, target:, user:, work_packages_map:).call
+      end
+
+      def initialize(source:, target:, user:, work_packages_map:)
+        @source = source
+        @target = target
+        @user = user
+        @work_packages_map = work_packages_map.to_h { |key, value| [key.to_i, value.to_i] }
```
Kharonus, line 42: ❓ huh? this seems not the change the map. What is the matter of the iteration here?

modules/storages/app/services/storages/file_links/copy_file_links_service.rb
```diff
@@ -0,0 +1,111 @@
...
+      end
+
+      def create_unmanaged_file_links(source_file_links)
+        source_file_links.find_each do |source_file_link|
+          attributes = source_file_link.dup.attributes
+
+          attributes['creator_id'] = @user.id
+          attributes['container_id'] = @work_packages_map[source_file_link.container_id]
+
+          # TODO: Do something when this fails
```
Kharonus: 🟡 resolve todo? 😁 

This is not part of current implementation, I know. What should we do?

`CopyFileLinksService` seems to not return a `ServiceResult` right now. Should it return one and log the included failures from the caller?

Or simply log the failures when the creation fails.

modules/storages/app/common/storages/peripherals/storage_interaction/one_drive/copy_template_folder_command.rb
Kharonus: Actually, I'd retract my change request here. I like it to be very explicit about NOT having an id here.

Kharonus (changes requested):
At least one thing me thinks is preventing a merge right now. The others are open questions und uncertain issues.

modules/storages/app/common/storages/peripherals/storage_interaction/one_drive/copy_template_folder_command.rb
mereghost: Yeah, that was the idea. Be explicit about the nil id. =)

modules/storages/app/services/projects/copy/storage_project_folders_dependent_service.rb
mereghost: Oh, this is enqueue as many jobs as there are project storages. Batches would give us a way to check for completion of all jobs and also re-organize the main `CopyProjectJob` around it.

modules/storages/app/services/projects/copy/storage_project_folders_dependent_service.rb
mereghost: Yes! I forgot to remove those! Good catch.

modules/storages/app/services/storages/file_links/copy_file_links_service.rb
mereghost: This transform all the keys and values to integers. 

Due to limitations of `ActiveJob` hashes can only be serialized if the keys are symbols or strings which is kinda inconvenient for our usage.

modules/storages/app/services/storages/file_links/copy_file_links_service.rb
mereghost: That was part of the feedback I wanted. I think for now I'll just log the errors.

We need to decide on what to return when some object does multiple operations, some of which might fail but not be catastrophic. A success with errors feels about right, but I'm no sure. Same goes of for having a result on a failure, feels correct but... idk.

modules/storages/app/services/projects/copy/storage_project_folders_dependent_service.rb
Kharonus: So there is no problem with using `GoodJob::Batches.enqueue` before our delayed jobs are refactored to GoodJob?

mereghost: In theory not, enqueue is part of `ActiveJob`s API, but this file just got deleted as it isn't needed anymore. XD

modules/storages/app/services/storages/file_links/copy_file_links_service.rb
```diff
@@ -0,0 +1,111 @@
...
+        @source = source
+        @target = target
+        @user = user
+        @work_packages_map = work_packages_map.to_h { |key, value| [key.to_i, value.to_i] }
+      end
+
+      def call
+        source_file_links = FileLink
+          .includes(:creator)
+          .where(container_id: @work_packages_map.keys, container_type: "WorkPackage")
```
akabiru, line 48: 🟢 guard clause `return ServiceResult.success if @work_packages_map.keys.empty?`

modules/storages/app/workers/storages/copy_project_folders_job.rb
```diff
@@ -0,0 +1,83 @@
...
+      FileLinks::CopyFileLinksService.call(source:, target:, user:, work_packages_map: work_package_map)
+    end
+
+    private
+
+    def prepare_polling(result, source)
+      return if result[:id]
+      return unless result[:url] # Inactive project storages can return id: nil here.
+
+      Thread.current[job_id] = result[:url]
```
akabiru: 🟠 `RequestStore` thread safer?

modules/storages/app/workers/storages/copy_project_folders_job.rb
```diff
@@ -0,0 +1,83 @@
...
+  class CopyProjectFoldersJob < ApplicationJob
+    retry_on Errors::PollingRequired, wait: 3, attempts: (Rails.env.test? ? 3 : :unlimited)
+    discard_on HTTPX::HTTPError
+
+    def perform(user_id:, source_id:, target_id:, work_package_map:)
+      target = ProjectStorage.find(target_id)
+      source = ProjectStorage.find(source_id)
+      user = User.find(user_id)
+
+      # TODO: Do Something when this fails
```
akabiru: 🟢 perhaps log and retry polling with exponential backoff? In the perhaps rare case that there s throttling or a network issue?

akabiru (dismissed):
Looks good to me 👍🏾

modules/storages/app/workers/storages/copy_project_folders_job.rb
mereghost: In this case it shouldn't be an issue as it isn´t running on the context of a puma thread, just on the background worker.

But I could make the change as I don't have any qualms with RequestStore. :P

modules/storages/app/services/storages/file_links/copy_file_links_service.rb
mereghost: There will be a call to `find_each` which will immediately return in case of an empty result.

In all cases it calls for a query or iterates over the empty array, so no extra handling should be necessary or distracting in this case.

mereghost:
PR #15054 includes this one and supersedes it. Closing for now.

PR #14893, opened 2024-03-01
fix link in document added mail text part

machisuji:
WP [#53211](https://community.openproject.org/projects/openproject/work_packages/53211/activity)


aaron-contreras (approved):
Very nice, thanks for the spec!

PR #14895, opened 2024-03-01
Show modified warning and save as button in project list

ulferts:
**Acceptance criteria**

* [x] Changing a persisted or static list:
  * [x] An information message is displayed next to the page header that the list was manipulated
  * [x] A "Save as" button appears in the top left toolbar
* [x] Pressing on the "Save as" button opens an input field for the name of the new list (same as clicking on the "Save as" entry in the menu)
* [x] All selections of the current list are saved (filters, columns sort order) 
* [x] Clarify what to do about the button styles overwritten from Primer
* [x] Check acceptance of altered message

**Notes**

The acceptance criteria deviate from what is specified in the work package as the code does not yet support sharing of lists and the storage of filters and sort order. The PRs implementing that will have to provide the functionality at the placed now added to the code later.

---------------

https://community.openproject.org/wp/52152

### DIFF
--- a/app/components/projects/index_page_header_component.html.erb
+++ b/app/components/projects/index_page_header_component.html.erb
@@ -3,6 +3,33 @@
     <% header.with_title(data: { 'test-selector': 'project-query-name'}) { page_title } %>
 
     <% header.with_actions do %>
+      <% if query_saveable? %>
+        <%= render(
+              Primer::Beta::Text.new(tag: :span,
+                                     mr: BUTTON_MARGIN_RIGHT,
+                                     color: :muted)) do
+            t('lists.can_be_saved_as')
+          end
+        %>
+
+        <%= render(
+              Primer::Beta::Button.new(scheme: :invisible,
+                                       size: :medium,
+                                       mr: BUTTON_MARGIN_RIGHT,
+                                       tag: :a,
+                                       href: new_projects_query_path,
+                                       data: {
+                                         controller: "params-from-query",
+                                         'application-target': "dynamic",
+                                         'params-from-query-allowed-value': '["filters"]'
+                                       },
+                                       classes: 'Button--invisibleOP')) do |button|
+          button.with_leading_visual_icon(icon: :'op-save')
+
+          t('button_save_as')
+        end %>
+      <% end %>
+
       <%= render(Primer::Alpha::ActionMenu.new) do |menu|
         menu.with_show_button(icon: 'op-kebab-vertical', 'aria-label': t(:label_more), data: { 'test-selector': 'project-more-dropdown-menu' })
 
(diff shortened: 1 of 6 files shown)


aaron-contreras (approved):
Code looks great 👍 - I love how simple it becomes when relying on the full page reload.

Regarding annotations for this specific PR, I did read the note on sorts not being persisted yet in the scope of this Work Package but it'd seem a bit confusing for users that changes in sort orders trigger the "Save as" warning as if it actually would be persisted - yes these are not persisted after triggering the form for giving the query a name but one could theoretically still sort while the form is active and be led to believe it would be saved. As long as this is communicated, I think it's fine :)

---

Noticed a couple of awkward behaviors while playing around with it. I do not believe they stem from this PR but I'm jotting them down here. You could prioritize them better as your understanding of the project list roadmap is a lot more thorough than mine and what the current bug list is for it:

1. If I sort by a certain column on a persisted query once or twice with a nice pause in between each sort, I get no "Save as" prompt and the `query_id` prop remains in the URL. However, if I do the aforementioned steps quickly enough in a consecutive manner, the `query_id` prop is lost in the URL and I'm prompted to save the unnamed query. Here's a recording demonstrating it (I start sorting slowly with a nice pause in between and then speed up, you'll notice the `query_id` prop gone when I do speed up):

https://github.com/opf/openproject/assets/61627014/48c77432-748f-4320-a93a-d5e988585ad3

The same case if I hit the "Apply" button quickly enough on the Filters section, regardless if it's a persisted query or not, I end up with an empty set of filters in the query props:

https://github.com/opf/openproject/assets/61627014/489192ec-ff07-4a50-b8e9-1799a186e493

My initial guess on the above behavior is that it's probably rooted in some javascript not being yet loaded which fetches the props currently on the URL or from the present elements on the DOM. I bring up this behavior as it can potentially cause a user to lose a very elaborate filter/status combination before they have a chance to persist it.

2. Sorting by "latest_activity_at" causes a server error as there's a missing join with the "activities" table.
image

aaron-contreras:
There is one merge conflict to resolve but it looks simple enough :)

ulferts:
Thanks for the review, @aaron-contreras.

> Sorting by "latest_activity_at" causes a server error as there's a missing join with the "activities" table.

There is a bug open for this: https://community.openproject.org/wp/53315

> However, if I do the aforementioned steps quickly enough in a consecutive manner, the query_id prop is lost in the URL and I'm prompted to save the unnamed query

I opened a bug for it: https://community.openproject.org/wp/53329

> sorts not being persisted yet in the scope of this Work Package but it'd seem a bit confusing for users that changes in sort orders trigger the "Save as" warning as if it actually would be persisted

Admittedly, I would just accept that shortcoming for this version of OP as the next one will provide the functionality. But I'll take another look to see if the message can be prevented to show up easily.

PR #14897, opened 2024-03-01
Bump gems

aaron-contreras:


aaron-contreras:
Do you happen to know if we rely on `ruby2_keywords` in some way? I don't think so but maybe you know if we do @ulferts

PR #14910, opened 2024-03-04
Use insert_all in  Workflows bulk update service

dombesz:
https://community.openproject.org/wp/53305
Replacing an old TODO.


aaron-contreras (approved):
🏎️

PR #14923, opened 2024-03-05
Feature/53022 use primer modal for project list deletion

ulferts:
**Acceptance criteria**

By the feature
* [x] The confirmation modal shown before the deletion of a persisted project list is Primer based

Documentation
* [x] Document confirmation modals in general and deletion confirmation modals briefly

Improvements
* [x] Move project list export UI completely into a component. That one is not yet generalized but could evolve into being used whenever we export a list.

---------

https://community.openproject.org/wp/53022


app/components/projects/delete_list_modal_component.html.erb
```diff
@@ -0,0 +1,16 @@
+<%= render(Primer::Alpha::Dialog.new(title: t(:'projects.lists.delete_modal.title'),
+                                     size: :large,
+                                     id: MODAL_ID,
+                                     data: { 'test-selector': MODAL_ID })) do |d| %>
+  <% d.with_header(variant: :large, mb: 2) %>
+  <% d.with_body { t(:'projects.lists.delete_modal.text') } %>
+
+  <%= render(Primer::Alpha::Dialog::Footer.new) do %>
```
aaron-contreras: I believe rendering the footer this way messes with the slot mechanism for the Primer Component as I don't see the expected body text in the dialog. Using `d.with_footer` ensures the slot is properly registered.

By inspecting the component's source code, it seems registering anything other than slots overrides the component's body.

```suggestion
  <% d.with_footer do %>
```

aaron-contreras: image

aaron-contreras (changes requested):
Very clean!

I've annotated inline the reason/fix for the missing body text on the delete confirmation dialog. Merge once addressed :)

app/components/projects/delete_list_modal_component.html.erb
ulferts: Thanks for catching this, @aaron-contreras. I use the render before also for the body and then found it better to use the slots but apparently wasn't thorough enough when making the change.

PR #14927, opened 2024-03-06
Fix Gem::InvalidSpecificationException in rake generated plugin

top4ek:
Fixes Gem::InvalidSpecificationException while loading empty plugin.


oliverguenther (approved):
Good catch, thanks @top4ek !

PR #14929, opened 2024-03-06
No need cells form view file now

Eric-Guo:
This file accidentally re-introduced at bbcdcbcc53b71ed0b4281d1351a6adafb2e407b7, but it already replace by role_form_component.html.erb at pull request #12655


ulferts (commented):
🙇 thanks @Eric-Guo

PR #14936, opened 2024-03-06
Link error to current appsignal transaction if available

cbliard:
Using `send_error` will always create a new transaction with a random request id, which is different from the real request id of the current web request and being present in log files.

Using `set_error` instead will link the error to the current transaction, adding a lot of extra context information to the error in the AppSignal web app like the correct request_id.


oliverguenther (approved):
So much debugging for such a simple solution 😅

PR #14941, opened 2024-03-07
Feature/53330 disable profile fields for ldap user

virgo47:
The UI fix was straightforward (as shown on [WP 53330](https://community.openproject.org/projects/openproject/work_packages/53330/activity)), but it also required a branch in the backend, because for LDAP user the Profile does not send any `user` hash. But the same backend call is used also for the Settings page. Please, suggest changes.

I had some strange resolution problems on the user edit page, so I also added one explict `::` prefix, I hope it's OK.

### DIFF
--- a/app/models/permitted_params.rb
+++ b/app/models/permitted_params.rb
@@ -187,8 +187,13 @@ class PermittedParams
   end
 
   def user(additional_params = [])
-    permitted_params = params.require(:user).permit(*self.class.permitted_attributes[:user] + additional_params)
-    permitted_params.merge(custom_field_values(:user))
+    if params[:user].present?
+      permitted_params = params.require(:user).permit(*self.class.permitted_attributes[:user] + additional_params)
+      permitted_params.merge(custom_field_values(:user))
+    else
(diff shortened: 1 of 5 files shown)


oliverguenther:
Thanks @virgo47 , LGTM 👍 Could you add a (feature) test for it, ideally with an LDAP user that has a custom field and update that to ensure it's still working

virgo47:
I'll look into it on Monday. Thanks for the feedback and tips.

virgo47:
Hi @oliverguenther, I added the test into `my_spec.rb` which already tests these pages. After adding the user CF, I had a minor hiccup though. Originally I wanted to specify the `let!` for the user CF inside `context 'LDAP user'`, but this caused the new test (on line 439) to fail if it didn't run after the other included test (for Settings page) in the same context. This puzzles me a bit, but moving the `let!` to the top-level scope helped and didn't harm the other tests. Do you know what's the matter?

virgo47:
Hi @oliverguenther - sorry for the impatience, anything new with this one?

oliverguenther (approved):
Looks good to me 👍

PR #14945, opened 2024-03-07
enforce double quotes

toy:
I care about consistency and as there are thousands of rubocop violations, why not enable this to fix things with every commit?


aaron-contreras:
Team's consensus seems to be all for it! 👍

toy:
I'd also want to hear from @ulferts, as he added the comment about not enabling it

ulferts (approved):

PR #14954, opened 2024-03-08
[#53368] add authentication for storage interaction

Kharonus:
[#53368](https://community.openproject.org/wp/53368)

### WAT?

- Added the authentication class to storage interaction namespace
- added unit test that shows the usage and test the auth behaviour of all storage providers


modules/storages/app/common/storages/peripherals/storage_interaction/authentication.rb
```diff
@@ -0,0 +1,69 @@
...
+            raise "Invalid authentication strategy '#{strategy}'"
+          end
+        end
+
+        # Checks for the current authorization state of a user on a specific file storage.
+        # Returns one of three results:
+        # - :connected If a valid user token is available
+        # - :failed_authorization If a user token is available, which is invalid and not refreshable
+        # - :error If an unexpected error occurred
+        def self.authorization_state(storage:, user:)
```
Kharonus, line 55: ignore this method, I simply didn't want to cut this out into a separate commit. It is needed later, when authorization state is extracted from the connection manager

modules/storages/app/common/storages/peripherals/oauth_configurations/nextcloud_configuration.rb
```diff
@@ -32,13 +32,18 @@ module Storages
   module Peripherals
     module OAuthConfigurations
       class NextcloudConfiguration < ConfigurationInterface
+        Util = ::Storages::Peripherals::StorageInteraction::Nextcloud::Util
```
mereghost: 🟢 The reference here could probably omit the `::Storages::Periphereals` part.

modules/storages/app/common/storages/peripherals/storage_interaction/authentication_strategies/error.rb
```diff
@@ -0,0 +1,44 @@
...
+# Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+#
+# See COPYRIGHT and LICENSE files for more details.
+#++
+
+module Storages
+  module Peripherals
+    module StorageInteraction
+      module AuthenticationStrategies
+        class Error
```
mereghost: 🟡 Since this is basically a factory for `ServiceResult`s we could make it into a simple lambda:

```ruby
module X...
  FailureBuilder = ->(code:, log_message:, data:) { same_implementation } 
end
```

and use `FailureBuilder.call(...)` on the code. `create` seems so tied to `ActiveRecord` that I thought you were persisting the data XD

modules/storages/app/common/storages/peripherals/storage_interaction/authentication_strategies/o_auth_client_credentials.rb
```diff
@@ -0,0 +1,82 @@
...
+                                    client_id: config.client_id,
+                                    client_secret: config.client_secret,
+                                    scope: config.scope,
+                                    http_options:,
+                                    &)
+            else
+              log_message = 'Cannot authenticate storage with client credential oauth flow. Storage not configured.'
+              data = ::Storages::StorageErrorData.new(source: self, payload: storage)
+              Error.create(code: :error, log_message:, data:)
+            end
```
mereghost: 🟡 WDYT we make it into a guard clause?

```ruby
return build_error(bunch:, of:, data:) unless config.complete?
```

modules/storages/app/common/storages/peripherals/storage_interaction/authentication_strategies/o_auth_client_credentials.rb
```diff
@@ -0,0 +1,82 @@
...
+                                .with(http_options)
+            rescue HTTPX::HTTPError => e
+              data = ::Storages::StorageErrorData.new(source: self, payload: e.response.json)
+              return Error.create(code: :unauthorized,
+                                  log_message: 'Error while fetching OAuth access token.',
+                                  data:)
+            end
+
+            yield http
+          end
```
mereghost: 🟡 Any reason we have the yield outside of the block?

You will return in case the first part raises, so the `begin...end` and just yield after building the httpx session. The rescue then can move to the outer scope.

It might catch other non-caught HTTPX errors, but that can be great if the messaging is improved (likely by using the exception error message). OR it can be refined into just rescuing 401/403s.

modules/storages/app/common/storages/peripherals/storage_interaction/authentication_strategies/o_auth_configuration.rb
```diff
@@ -0,0 +1,55 @@
...
+                         issuer: nil,
+                         scope: nil)
+            @client_id = client_id
+            @client_secret = client_secret
+            @issuer = issuer
+            @scope = scope
+          end
+
+          def complete?
+            [@client_id, @client_secret, @issuer].all?(&:present?)
```
mereghost: 🟡 WDYT of leveraging [`ActiveModel::Validations`](https://api.rubyonrails.org/classes/ActiveModel/Validations.html) to check this and call `valid?` on the code, making it more... seamless and idiomatic.

modules/storages/app/common/storages/peripherals/storage_interaction/authentication_strategies/o_auth_user_token.rb
```diff
@@ -0,0 +1,117 @@
...
+  module Peripherals
+    module StorageInteraction
+      module AuthenticationStrategies
+        class OAuthUserToken
+          def self.strategy
+            Strategy.new(:oauth_user_token)
+          end
+
+          def initialize(user)
+            super()
```
mereghost: 🔴 Unnecessary call to super()

modules/storages/app/common/storages/peripherals/storage_interaction/authentication_strategies/o_auth_user_token.rb
```diff
@@ -0,0 +1,117 @@
...
+            @user = user
+          end
+
+          # rubocop:disable Metrics/AbcSize
+          def call(storage:, http_options: {}, &)
+            config = storage.oauth_configuration
+            current_token = OAuthClientToken.find_by(user_id: @user, oauth_client_id: config.oauth_client.id)
+            if current_token.nil?
+              data = ::Storages::StorageErrorData.new(source: self)
```
mereghost: 🟡 All uses of data are the same, we could just initialize it directly at the instance level `@data = Storage::ErrorData.new(source: self)`.

Also Is your intention of adding the whole object to the `ErrorData` or just a reference to the class name? On a log this would look like `<StorageErrorData#12313 @source: <SomeClass#1234551>>`

modules/storages/spec/common/storages/peripherals/storage_interaction/network_error_spec.rb
```diff
@@ -0,0 +1,90 @@
...
+  let(:storage) { create(:sharepoint_dev_drive_storage, oauth_client_token_user: user) }
+  let(:request_url) { 'https://my.timeout.org/' }
+  let(:auth_strategy) do
+    Storages::Peripherals::StorageInteraction::AuthenticationStrategies::OAuthUserToken.strategy.with_user(user)
+  end
+
+  context 'if a timeout happens' do
+    before do
+      request = HTTPX::Request.new(:get, request_url)
+      httpx_double = class_double(HTTPX, get: HTTPX::ErrorResponse.new(request, 'Timeout happens', {}))
```
mereghost: 🔴 You can actually mock a timeout with `WebMock`

```ruby
stub_request(:get, 'https://timeout-as-a-service.com/').to_timeout

**Review by mereghost — CHANGES_REQUESTED** [review]
From what I see you have the usage be somewhat like so:

```ruby
CreateTemplateFolderCommand.call(auth: Strategy.new(:oauth_client_credentials), storage: ....)

# or on the inside of an object
class RmRfCommand
  def call(strategy: :basic_auth, storage:, location:)
     Authentication[strategy].call do |authed_http_session|
       do_awesome_requests
     end
  end
end
```

modules/storages/spec/common/storages/peripherals/storage_interaction/network_error_spec.rb
Kharonus: True, that spec was part of the query tests before, hence webmock was no option (interfering with vcr). 

I can change that now.

modules/storages/app/common/storages/peripherals/storage_interaction/authentication_strategies/o_auth_user_token.rb
Kharonus: Same question as in the configuration classes? linter wants super calls, should we disable linter locally or call super? Or should we disable linter globally?

modules/storages/app/common/storages/peripherals/oauth_configurations/nextcloud_configuration.rb
Kharonus: removed

modules/storages/app/common/storages/peripherals/storage_interaction/authentication_strategies/error.rb
Kharonus: Sure, I have no dips on this as a class. will change it to a lambda.

modules/storages/app/common/storages/peripherals/storage_interaction/authentication_strategies/o_auth_user_token.rb
Kharonus: Yeah, we need to make this discussion:

there are some advantages and disadvantages for both, IMHO.

- putting an object makes a written error log less readable
- putting an object makes debugging simpler, as the object contains state, as the internal values

wdyt?

modules/storages/app/common/storages/peripherals/storage_interaction/authentication_strategies/o_auth_configuration.rb
Kharonus: What would be part of leveraging the configuration objects to an active model? Pros and cons?

modules/storages/app/common/storages/peripherals/storage_interaction/authentication_strategies/o_auth_client_credentials.rb
Kharonus: point for the rescue here is actually to raise on every status. e.g. GRAPH API fails under certain circumstances with a 400, not a 401, when refreshing tokens (misconfigured or corrupted data).

My understanding was, that when moving the yield inside the rescue, that we will catch exceptions also from within our queries, and we might not want that. Instead we probably want the queries to make there own error handling. Or do you think different?

modules/storages/app/common/storages/peripherals/storage_interaction/authentication_strategies/o_auth_client_credentials.rb
Kharonus: done

modules/storages/app/common/storages/peripherals/storage_interaction/authentication_strategies/o_auth_user_token.rb
mereghost: You know what is weird? The fact that the Rubymine complains but lefthook doesn't. >.< Maybe it is ignoring something.

modules/storages/app/common/storages/peripherals/storage_interaction/authentication_strategies/o_auth_configuration.rb
mereghost: the validations module just add the validation macros to an object. I'm usually more on the `Dry::Validation` side of having a validation contract et al, but for now adding the ActiveModel can help with clarity and error messaging.

```ruby
class OAuthConfiguration
  include ActiveModel::Validations

  validate :presence, of: :client_id, :client_secret, :issuer
end
```

Does basically the same using RoR idioms of `validity.

modules/storages/app/common/storages/peripherals/storage_interaction/authentication_strategies/o_auth_user_token.rb
mereghost: Discussed on Element and reached an agreement. I'd add a link to it... if Element would cooperate. :yells-at-element:

modules/storages/app/common/storages/peripherals/storage_interaction/authentication_strategies/o_auth_client_credentials.rb
Kharonus: discussed: we will leave it outside the block for now, due to the mentioned reasons.

modules/storages/app/common/storages/peripherals/storage_interaction/authentication_strategies/o_auth_user_token.rb
Kharonus: discussed: prefer disabling linter rule over writing unnecessary super

modules/storages/app/common/storages/peripherals/storage_interaction/authentication_strategies/o_auth_user_token.rb
Kharonus: discussed: use `self.class` as we value logging > debugging.

modules/storages/app/common/storages/peripherals/storage_interaction/authentication_strategies/o_auth_configuration.rb
Kharonus: great one, will use it.

modules/storages/spec/common/storages/peripherals/storage_interaction/network_error_spec.rb
Kharonus: Update: I will refactor that test with the first query that is moved to new auth strategies. Right now, I don't see a good sense in this whole test :/

I wanted to cover the behaviour of httpx to return network errors in a different object. But the handling of it happens inside the queries. So, I would suggest to add a network unit test with webmock, that runs on top of an example query, which is considered a placeholder for all other queries, too.

mereghost (approved):

PR #14958, opened 2024-03-08
Make done_ratio nil by default

cbliard:
This is a work in progress to find broken tests

### DIFF
--- a/config/initializers/new_framework_defaults_7_0.rb
+++ b/config/initializers/new_framework_defaults_7_0.rb
@@ -90,7 +90,7 @@
 # This default means that all columns will be referenced in INSERT queries
 # regardless of whether they have a default or not.
 # Previous versions had true. Rails 7.0+ default is false.
-# Rails.application.config.active_record.partial_inserts = false
+Rails.application.config.active_record.partial_inserts = false
 
 # https://guides.rubyonrails.org/configuring.html#config-action-controller-raise-on-open-redirects
 # Protect from open redirect attacks in `redirect_back_or_to` and `redirect_to`.
--- a/spec/services/work_packages/set_attributes_service_spec.rb
+++ b/spec/services/work_packages/set_attributes_service_spec.rb
@@ -30,9 +30,9 @@ require "spec_helper"
 
 RSpec.describe WorkPackages::SetAttributesService,
                type: :model do
-  shared_let(:status_no_pct_complete) { create(:status, default_done_ratio: nil, name: 'no % complete') }
-  shared_let(:status_50_pct_complete) { create(:status, default_done_ratio: 50, name: '50% complete') }
-  shared_let(:status_70_pct_complete) { create(:status, default_done_ratio: 70, name: '70% complete') }
+  shared_let(:status_no_pct_complete) { create(:status, default_done_ratio: nil, name: "no % complete") }
+  shared_let(:status_50_pct_complete) { create(:status, default_done_ratio: 50, name: "50% complete") }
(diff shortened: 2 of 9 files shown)


config/initializers/new_framework_defaults_7_0.rb
```diff
@@ -90,7 +90,7 @@
 # This default means that all columns will be referenced in INSERT queries
 # regardless of whether they have a default or not.
 # Previous versions had true. Rails 7.0+ default is false.
-# Rails.application.config.active_record.partial_inserts = false
+Rails.application.config.active_record.partial_inserts = false
```
aaron-contreras, line 93: I see this is relevant for changing column defaults in Migrations now and I'd like to understand why. Is it that with partial inserts, if done_ratio was not part of the work package record initial record attributes being set, there would be a clash with the non-null constraint? And with partial_inserts disabled, all columns are part of the insert statement and defaults are fetched from the table metadata and added to the insert statement and avoid the non-null constraint clash?

aaron-contreras (approved):
Neat! Thanks for the spec updates :)

config/initializers/new_framework_defaults_7_0.rb
cbliard: That was [recommended](https://github.com/ankane/strong_migrations?tab=readme-ov-file#changing-the-default-value-of-a-column) in strong_migrations gem which is always my go-to reference when updating database schema. They recommended to use `config.active_record.partial_writes = false` because "Rails < 7 enables partial writes by default, which can cause incorrect values to be inserted when changing the default value of a column.".

That was not very helpful so I wanted to know more and read https://blog.kiprosh.com/rails-7-introduces-partial-inserts-config-for-activerecord/ which helped with understanding the issue.

With partial insert, any default value is not inserted so that the database takes care of inserting its default. For instance when doing `WorkPackage.create(done_ratio: 0)`, Rails would not actually insert done_ratio into the query, and the value will be set to `0` by the database itself.

So far so good.

But what happens if the database default value is changed and the app is not restarted (strong_migrations thinks from a zero-downtime migration point of view). Let's say we changed the default value of `WorkPackage.done_ratio` from 0 to 50. From a non-restarted app we then do `WorkPackage.create(done_ratio: 0)`. Rails sees that it is the default value (it read the database info on startup) and as it is smart it removes it from the sql insert query. So the value is backfilled by the database, and we end up with the value `50`. Oups.

It's to prevent this kind of issues that `config.active_record.partial_inserts` should be set to `false`.

aaron-contreras: Excellent example - this clears up the underlying mechanics going on!

cbliard:
@aaron-contreras can you please review the latest commit? I tried adding all tests scenarios from the work package. The unimplemented ones are skipped for now.

spec/services/work_packages/set_attributes_service_spec.rb
```diff
@@ -209,91 +202,175 @@
...
-      context "when both estimated_hours and remaining_hours are unset" do
-        let(:call_attributes) { { estimated_hours: nil, remaining_hours: nil } }
-        let(:expected_attributes) do
-          { estimated_hours: nil, remaining_hours: nil, done_ratio: nil }
+        context "when remaining work is changed to a value greater than work", skip: "TODO: not implemented yet" do
+          let(:call_attributes) { { remaining_hours: 200.0 } }
+          let(:expected_attributes) { "error" }
+
+          # open question: should it be capped or produce an error?
+          # I would opt for the error which would then be displayed in the popover
```
aaron-contreras, line 327: I'd opt for an error too

PR #14963, opened 2024-03-11
[#53345] Synchronize only configured storages.

ba1ash:
https://community.openproject.org/work_packages/53345

It solves the problem when background job tries to "synchronize" not configured storage and raises a run time exception.


mereghost (approved):
LGTM. That would kill another source of issues here.

PR #14965, opened 2024-03-11
[#53047] 13.4 Documentation: project lists changes  https://community…

MayaBerd:
….openproject.org/work_packages/53047

[#53047] 13.4 Documentation: project lists changes

https://community.openproject.org/work_packages/53047


docs/user-guide/projects/project-lists/README.md
```diff
@@ -41,16 +41,26 @@ You can use this projects overview to **create a multi project status dashboard*
...
 
 ![new subproject project list](new-subproject-project-list.png)
 
 You can choose the **columns displayed by default** in the [System settings](../../../system-admin-guide/system-settings/project-system-settings) in the Administration. 
 
+You can add the columns, as well as define the order of the columns by using the **Configure view** modal. Navigate to it via the menu in the far right corner (three dots) and click **Configure view**.
+
+![Configure view of project lists in OpenProject](configure-view-project-list.png)
+
+A window will open, allowing you to manage and reorder columns. Don't forget to click **Apply** to see the changes.
```
marcalcobe: I wouldn't use the term window here but dialog or modal as it is not opening a new window.

Also I would change the see to save in this sentence "...click **Apply** to see"

docs/user-guide/projects/project-lists/README.md
```diff
@@ -75,7 +85,7 @@ Projects can be filtered in OpenProject. The default view will list all currentl
 
 To adjust the project lists view use the **Filters** button, select the filtering criteria you require and click the blue **Apply** button.
 
-If you want to save this filtered list use the menu in the far right corner (three dots) and click **Save**.
+If you want to save this filtered list use the menu in the far right corner (three dots) and click **Save as**. Alternatively you can also use **Save as link** next to the warning message.
```
marcalcobe: I would change the order of this sentence (and screenshots) and do some small changes:

"If you want to save this filtered list use the link **Save as** next to the information message in the page header or alternatively click on the menu (three dots) and click **Save as**.

marcalcobe: Another thing to consider here is that this is not always **Save as**. There is two cases:

1. The modified a list that can be saved after the changes, then the link will be **Save** not **Save as**. And in the more icon button there will be both actions **Save** and **Save as**.
2. The user has modified a list that can't be saved with the changes (eg. a default project list like "Active projects"). Then there is only the option to **Save as**.

marcalcobe (changes requested):
@MayaBerd I've added some comments

docs/user-guide/projects/project-lists/README.md
MayaBerd: @marcalcobe good point and important to document. However, changing persisted own lists was moved out of 13.4. (https://community.openproject.org/projects/openproject/work_packages/52144/)
I will add this once the feature is implemented

PR #14966, opened 2024-03-11
fix Users resolution in lib/open_project/ui/extensible_tabs.rb

toy:
It goes to lookbook/previews/open_project/users/avatar_component_preview.rb, introduced in #14885


cbliard:
@toy Can you retarget you PR for release/13.4 branch?

cbliard:
And while your fix is correct, it would be nice to add a feature test which simply opens the user edit page. It's surprising that we have none at the moment.

toy:
After some searching this looks like the only problem with `Users` resolution

cbliard (approved):
Thanks! I'll merge it like this. If you have time for adding a feature test for it that would be nice.

PR #15015, opened 2024-03-18
Start correcting single quotes

toy:
This will probably create some conflicts with current PRs, but there is no way around it.

Finding the most changed files using:
```shell
git log --name-only --since="5 years" --pretty=format: Gemfile '*.rb' | grep -v '^$' | sort | uniq -c | sort -g | tail -n 10
```

### DIFF
--- a/.git-blame-ignore-revs
+++ b/.git-blame-ignore-revs
@@ -5,3 +5,5 @@
 21a696ef9b170e14ad2daf53364a4c2113822c2f
 # Update copyright information for 2023
 c795874f7f281297bbd3bad2fdb58b24cb4ce624
+# switch to double quotes in most changed files with a lot of quotes
+5c72ea0046a6b5230bf456f55a296ed6fd579535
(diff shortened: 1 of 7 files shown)


toy:
@cbliard I thought we should limit it to something smaller, like `/[a-z0-9.\-_]/`?

cbliard:
@toy the warning was about having overlapping character classes in the character ranges. I only removed the duplicates to fix this warning. I did not question if `[a-zA-Z0-9_.\-]` (same as `[\w.\-]`) is actually relevant or if it should be limited to a shorter range. I assumed the range had been chosen for a reason and that there are valid uses of all the characters in the range.

And searching in history, c41ea1ed48b "Fix restricting the rev only to lowercase tags" changed `/[a-z0-9\.\-_]+/` to `/[\w0-9\.\-_]+/`. That's where the duplication first appeared, so I think `[\w.\-]` is ok.

toy:
@cbliard Thanks for checking history

cbliard (approved):
We are finally moving forward with this. Thanks for taking care of it!

PR #15025, opened 2024-03-18
Bug/53391 project custom fields and project description no longer allows macros

ulferts:
The PR will change the support for macros in project description as well as the status explanation. Both now support macros on the project settings form. This is in line with the project overview page where those two fields support macros as well. 

Custom fields no longer support macros (via the toolbar). The project settings and also on the overview page, macros had never been working (albeit briefly by the merging of the project attributes PR) but the custom field administration supported it for the default value. The last is no longer the case. This affects all custom fields.

Having macros for status explanation and description is poorly supported in the project list (see [#41338](https://community.openproject.org/wp/41338)). The PR only improves the stability and the design of the error message when this happens. 

-----

https://community.openproject.org/wp/53391
https://community.openproject.org/wp/53701

### DIFF
--- a/app/components/open_project/common/attribute_component.rb
+++ b/app/components/open_project/common/attribute_component.rb
@@ -76,6 +76,14 @@ module OpenProject
 
       private
 
+      def first_paragraph_content
+        return unless first_paragraph_ast
+
+        first_paragraph_ast
+          .inner_html
+          .html_safe # rubocop:disable Rails/OutputSafety
+      end
+
       def first_paragraph
         @first_paragraph ||= if body_children.any?
                                body_children
@@ -85,6 +93,13 @@ module OpenProject
                              else
                                ""
                              end
+      end
+
+      def first_paragraph_ast
+        @first_paragraph_ast ||= text_ast
+                                 .xpath("html/body")
+                                 .children
+                                 .first
       end
 
       def text_ast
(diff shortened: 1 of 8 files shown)


dombesz:
I have a few observations, not necessarily related to this PR, but they might be:
1. Navigating away from the overview page, will render an error on some macros:

<details><summary>[Demo]</summary>
<p>

Weird error message


</p>
</details> 

2. The default value editor of the long text project attributes has the dropdowns cut off:
<details><summary>[Demo]</summary>
<p>

image

image

Same for the custom fields too:
image


image

</p>
</details> 
3. The project overview sidebar attributes have the dropdowns hidden behind the toolbar of the editor:
<details><summary>Demo</summary>
<p>


image


</p>
</details>

app/components/open_project/common/attribute_component.rb
```diff
@@ -88,7 +103,9 @@ def body_children
       end
 
       def multi_type?
-        first_paragraph.include?("figure") || first_paragraph.include?("macro")
+        @multi_type ||= (description.present? && first_paragraph_ast.nil?) ||
+          %w[opce-macro-embedded-table figure macro].include?(first_paragraph_ast.name) ||
+          first_paragraph_ast.css("figure, macro, .op-uc-toc--list, .opce-macro-embedded-table")&.any?
```
dombesz: In [this PR](https://github.com/opf/openproject/pull/15130/files) I discovered that in some cases we double formatted the output of project custom fields and hence sometimes we had `macro` and other times`opce-macro-embedded-table` in the output.
The double rendering happened on the project list page only and [it's fixed now](https://github.com/opf/openproject/pull/15130/commits/9dc9f81ee09b7104e5b5a5b1405b1b0117af416a). This might also affect the issue related to the [#41338](https://community.openproject.org/wp/41338) mentioned in the PR description.

dombesz (approved):
I just left a few observations, otherwise LGTM!👍

app/components/open_project/common/attribute_component.rb
ulferts: Good catch @dombesz and also a good improvement to the PR you did there.

PR #15028, opened 2024-03-18
[#53429] Project storage menu links does not include prefix.

ba1ash:
https://community.openproject.org/work_packages/53429


mereghost (approved):
LGTM

akabiru (commented):
🟢 _Nit_ I wonder if it's more consistent to define a `url_helpers` method in `ApplicationForm` similar to [`ApplicationComponent#url_helpers`](https://github.com/opf/openproject/blob/dev/app/components/application_component.rb#L40-L42) to DRY things up a bit as it also seems like an easy miss. Perhaps even a mixin from StaticRouting

[Primer ApplicationForm is just `Primer::Forms::Base`](https://github.com/opf/primer_view_components/blob/main/app/forms/application_form.rb)

mereghost:
Hmmmm good question @akabiru. How often are we using it? Would this change a bunch of the current components? If so I'd make it into a Code Maintenance WP making sure we get all references.

[random unsolicited rant]
I personally hate when pieces of code far from the transport/delivery layer having to know about `url_helpers`. In this case it is ok as those are components but I've messed with jobs that know how to build application urls. 😠 
[/random unsolicited rant]

akabiru:
> code far from the transport/delivery layer having to

By my reckoning we're not using it too much- I guess here I'm looking out for the probability of repeating the same "mistake". 😄 I can see myself or someone new to the team defaulting to `Rails.application.routes` again. Perhaps the larger question is whether `Rails.application.routes` should be then one that is prefix aware and not a completely separate module. The ol' convention vs configuration approach

This can be a code maintenance discussion for sure

PR #15040, opened 2024-03-19
Flaky schedule date alerts notifications job spec

toy:
Starting with 18:15 utc(?) there is an extra Notifications::WorkflowJob created for Kathmandu time zone test which fails the change count test.

Am I correcting the right thing?

### DIFF
--- a/spec/workers/notifications/schedule_date_alerts_notifications_job_spec.rb
+++ b/spec/workers/notifications/schedule_date_alerts_notifications_job_spec.rb
@@ -100,7 +100,7 @@ RSpec.describe Notifications::ScheduleDateAlertsNotificationsJob, type: :job, wi
                           .last
           expect_job(j, user)
         end
-      end.to change(GoodJob::Job, :count).by 1
+      end.to change { GoodJob::Job.where(job_class: "Notifications::CreateDateAlertsNotificationsJob").count }.by 1
     end
   end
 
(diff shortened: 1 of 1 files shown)


spec/workers/notifications/schedule_date_alerts_notifications_job_spec.rb
```diff
@@ -86,104 +86,95 @@ def deserialize_job(job)
...
+    it "creates the job for the user" do
       expect do
         run_job(timezone:, scheduled_at:, local_time:) do
           j = GoodJob::Job.where(job_class: "Notifications::CreateDateAlertsNotificationsJob")
                           .order(created_at: :desc)
                           .last
           expect_job(j, user)
         end
-      end.to change(GoodJob::Job, :count).by 1
+      end.to change { GoodJob::Job.where(job_class: "Notifications::CreateDateAlertsNotificationsJob").count }.by 1
```
ulferts: It would be a tiny bit more robust if the job class is taken from the constant `Notifications::CreateDateAlertsNotificationsJob` instead of the string.

ulferts (approved):
The change itself is good. 

Extra points for changing the string into the constant. But merge at will since that is minor.

spec/workers/notifications/schedule_date_alerts_notifications_job_spec.rb
toy: Good point

PR #15044, opened 2024-03-19
[#53394] OAuth flow causes loss of already selected option while adding a storage to a project.

ba1ash:
https://community.openproject.org/work_packages/53394

The idea is to have currently selected **project_folder_mode** option in a query parameter and to make backend controller respect the query parameter. Then after user is away to grant oauth access return url contains **project_folder_mode** and it will be chosen on user return.


akabiru (approved):
Nice one ✨  :shipit:

PR #15054, opened 2024-03-20
Introduces the use of `GoodJob::Batch` for CopyProjectJob

mereghost:
## What is this?

It all started when we decided to add `OneDrive/Sharepoint` as part of the File Storages offering… ~~It was a rainy night at a dive bar in downtown Ber~~

Supporting **OneDrive/Sharepoint** (*OD* for brevity sake) has presented us with new challenges, from what was the actual concept of a `FileStorage` to implementation details such as permission handling et al. It works so differently than our other use case (**Nextcloud**) that we had to adapt our codebase and software behavior to make it conform to both.

This is one of those places:  copying a project or using it as a template. This PR aims to implement this feature for **OD**.

## Why do we need this?

**OD**, when it comes to copying folders is an asynchronous call. We ask for the copy, they gives an URL to check the status of the copy job and at the end handles us the new `resource_id`. Compared to the synchronous nature of **Nextcloud**, this was quite the change.

So the code would need to be bent and reshaped to our needs: we'd need to poll the **OD** provided URL until it is done, which of course presents its own challenges. 

During this time we had `GoodJob`<sup>Pavel</sup> merged giving us the ability to run jobs in batches, which fits this type of work as a glove.

### Part 1: Batches 101

To make a long topic short: `GoodJob::Batch` coordinates jobs enqueued under it. It also makes possible to share some state between the jobs, by the use or `Batch#properties` and to add other jobs as callbacks to batch events (such as finished, discarded etc).

To enqueue a new `Batch` :

```ruby
GoodJob::Batch.enqueue(on_finish: CallBackJob, some_key: 'this becomes part of the batch properties') do
  MyCoolJob.perform_later(my_awesome: job_arguments)
end

# or alternatively the slow build

batch = GoodJob::Batch.new
batch.on_finish: CallBackJob
batch.properties[:some_key] = 'this becomes part of the batch properties'
batch.add { MyCoolJob.perform_later(my_awesome: job_arguments) }
batch.save
batch.enqueue
```

The batch properties are serialized by `ActiveJob::Serializers` so anything that's safe for a job, is safe for it.

From inside a job, if you have included `GoodJob::ActiveJobExtensions::Batches` you can access the batch by simply calling `batch`.

```ruby
class SomeJob < ActiveJob
  include GoodJob::ActiveJobExtensions::Batches
  
  def perform
    batch.properties[:some_key]
    batch.properties[:featuring] = 'David Bowie'
    batch.save
    
    batch.enqueue { OtherCoolJob.perform(:under_pressure) }
  end
end
```

But that's not all! Batches can be updated in runtime, adding new jobs or properties to it as demonstrated above. Cool, huh?

### Part 2: The current state

Here is a small list of all the pieces that already exist that needed some tuning or revamps.

1. `Project::CopyService`: This is were the bulk of the action happens. `Project::CopyService` uses the structure of the `DependentServices` to handle all its intricacies. 
1. `CopyProjectJob`: This is where we ensure that everything non-copying is done, like making sure we respect the user localization and sending e-mails . It also is the job that gets polled by the UI for copy completion. 
1. `StoragesDependentService`, `StorageProjectFoldersDependentService`, `FileLinksDependentService`: this is where most of the work for `Storages` happen, making sure that the particularities of each storage and project storage mode are respected.

These pieces worked together to provide the the entirety of the *Copy Folder* feature taking into account the Nextcloud integration.

### Part 3: What changed (aka the cool part):warning:

First thing was to break most of the behavior for copying project folders from `CopyService`. This created 2 services: `ProjectStorages::CopyProjectFoldersService` and  `FileLinks::CopyFileLinksService`. Those 2 new services take over the old `Dependent` services but provide basically the same functionality.

But `CopyService` uses the `.copy_dependencies` method as a base to tell the UI what can be copied, so here it comes the first jury-rig of the PR `.copyable_dependencies` was overridden and the entries relating to `FileLinks` and `ProjectFolders` added by hand (aka copy-pasted) from the now deprecated services.

But now we have to deal with "**maybe** polling", so a new background job comes in: `CopyProjectFoldersJob`. The main point of this job is to retry itself if a `PollingRequired` exception is raised. This will happen if the result of the `CopyFolderCommand` indicates that it requires polling.

So, now we need a way to coordinate the `CopyProjectJob` with one or more `CopyProjectFoldersJob`, that's where the `GoodJob::Batch` comes in. Most of the changes will affect the `CopyProjectJob`. I'll add some snippets, but please check out the code.

#### `CopyProjectJob` Batch edition

After cleaning up a bit the instance variables (it had like 11 *million* of them), we hook it to the batch system. We also want to enqueue the `CopyProjectFoldersJob` from within it, adding them to the batch. But how? Luckily `CopyProject` keeps a state with a lot of information about what has transpired internally so that other `Dependent` services can rely on previous data.

So right after the copy process concludes, we tap into that state (conveniently called `#state`), and look for the key that we are interested in: `copied_project_storages`. This carries the source and copied `ProjectStorage` pairs, so looping through this list we add a new `CopyProjectFoldersJob` for each copied `ProjectStorage`, passing along some extra needed information.

Now that we have the batch, we can rely on it to store the polling URLs and the state of the polling process without the need for error prone `Thread.current` values.

We also only want to e-mail the user once the entire process completes, so another job was born `SendCopyProjectStatusEmailJob`. This one is added as a callback that is called once the entire batch finishes (successfully or not). Sounds easy enough, but it needs a bunch of information (errors, copied project name etc) so we push all this info to the `Batch#properties` making easy this new job easy to access the necessary info (or any other jobs in the batch).

### Ok, cool. Which are the gotchas here?

Since our API doesn't know how to poll for `Batch` completion the user will be redirected once the `CopyProjectJob` is done even if the `CopyProjectFoldersJob` are enqueued/running.  The completion email, will only be sent once everything is done.

The amount of change would not be trivial, so I left this as a future improvement (future can be tomorrow :rofl:).  `Storages` aren't also writing to the `errors` right now as we still need to decide on how to deal with `command`s that might error successfully (or some operations might fail, but overall be considered a success).


modules/storages/app/workers/storages/copy_project_folders_job.rb
```diff
@@ -30,13 +30,15 @@
 
 module Storages
   class CopyProjectFoldersJob < ApplicationJob
-    retry_on Errors::PollingRequired, wait: 3, attempts: (Rails.env.test? ? 3 : :unlimited)
+    include GoodJob::ActiveJobExtensions::Batches
+
+    retry_on Errors::PollingRequired, wait: 3, attempts: :unlimited
```
mereghost: What would be a good backoff algo here?

mereghost:
Well... this dev rebae was intense. 

Guy sweating

app/workers/send_copy_project_status_email_job.rb
```diff
@@ -0,0 +1,67 @@
...
+# along with this program; if not, write to the Free Software
+# Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+#
+# See COPYRIGHT and LICENSE files for more details.
+#++
+
+class SendCopyProjectStatusEmailJob < ApplicationJob
+  # Job is to be used as a callback to the CopyProjectJob batch
+
+  def perform(batch, _args)
```
akabiru, line 34: Ok this is very nice 👍🏾

modules/storages/app/common/storages/peripherals/storage_interaction/nextcloud/copy_template_folder_command.rb
```diff
@@ -26,102 +26,109 @@
...
+              @storage.uri,
+              "remote.php/dav/files",
+              escaped_username,
+              Util.escape_path(destination_path)
+            )
+
+            { source_url:, destination_url: }
+          end
+
+          def remote_folder_does_not_exist?(destination_url)
```
akabiru: 🟢 _nitpick_ Seems less of a predicate method

```suggestion
          def remote_folder_does_not_exist!(destination_url)
```

modules/storages/app/common/storages/peripherals/storage_interaction/nextcloud/copy_template_folder_command.rb
```diff
@@ -26,102 +26,109 @@
...
+            else
+              Util.error(:unknown, "Unexpected response (validate_destination): #{response.code}", response)
+            end
+          end
+
+          def copy_folder(source_url:, destination_url:)
+            response = OpenProject
+              .httpx
+              .basic_auth(@storage.username, @storage.password)
+              .request("COPY", source_url, headers: { "Destination" => destination_url, "Depth" => "infinity" })
```
akabiru, line 109: Q: I presume this is a HTTP COPY method type, that nextcloud knows how to handle? "Depth" => "Infinity"? 😄

modules/storages/app/common/storages/peripherals/storage_interaction/result_data/copy_template_folder.rb
```diff
@@ -0,0 +1,11 @@
+module Storages
+  module Peripherals
+    module StorageInteraction
+      module ResultData
+        CopyTemplateFolder = Data.define(:id, :polling_url, :requires_polling) do
```
akabiru, line 5: 🟢 Good to see this in the wild, I'm still stuck on `Struct` and haven't caught up on why `Data.define` is the better option 😁

modules/storages/app/services/storages/file_links/copy_file_links_service.rb
```diff
@@ -0,0 +1,134 @@
...
+        OpenProject.logger.error "STORAGE CLASS: #{@source.storage.class}, #{@target.storage.class}"
+
+        source_file_links = FileLink
+          .includes(:creator)
+          .where(storage: @source.storage,
+                 container_id: @work_packages_map.keys,
+                 container_type: "WorkPackage")
+
+        with_locale_for(@user) do
+          if @source.project_folder_mode == "automatic"
```
akabiru: 🟢 
```suggestion
          if @source.project_folder_automatic?
```

modules/storages/app/services/storages/file_links/copy_file_links_service.rb
```diff
@@ -0,0 +1,134 @@
...
+
+      def initialize(source:, target:, user:, work_packages_map:)
+        @source = source
+        @target = target
+        @user = user
+        @work_packages_map = work_packages_map.to_h { |key, value| [key.to_i, value.to_i] }
+      end
+
+      def call
+        OpenProject.logger.error "STORAGE CLASS: #{@source.storage.class}, #{@target.storage.class}"
```
akabiru: 🟠 Q: Why is the log level an `error` here? Should it not be `info`?
```suggestion
        OpenProject.logger.info "STORAGE CLASS: #{@source.storage.class}, #{@target.storage.class}"
```

modules/storages/app/workers/storages/copy_project_folders_job.rb
akabiru: 🟠 I tend to prefer the [exponential backoff algo](https://edgeapi.rubyonrails.org/classes/ActiveJob/Exceptions/ClassMethods.html#method-i-retry_on), as opposed to a fixed wait time. In case the third party API has an outage, an exponential wait time is more permissive from my POV.

(2) I also worry about "unlimited" running jobs, they can pile up and cause inconvenient load. In which case I find setting a limit of when we can safely "give up" and IF needed implement functionality for user initiated retries.

Proposal: retry for max ~15hours with exponential backoff

```suggestion
    retry_on Errors::PollingRequired, wait: :exponentially_longer, attempts: :15
```

-----

_Calculations on exponential backoff_

https://edgeapi.rubyonrails.org/classes/ActiveJob/Exceptions/ClassMethods.html#method-i-retry_on

```ruby
[12] pry(main)> jitter = 0.15 # default is 15%
[13] pry(main)> (1..15).each.with_index(1) do |executions, index|
  wait_time = ((executions**4) + (Kernel.rand * (executions**4) * jitter)) + 2
  puts "#{index} - #{wait_time} - #{ActiveSupport::Duration.build(wait_time).inspect}"
end
1 - 3.029356351145444 - 3.029356351 seconds
2 - 18.531756452471637 - 18.531756452 seconds
3 - 94.93286506765344 - 1 minute and 34.932865068 seconds
4 - 260.5615664632436 - 4 minutes and 20.56156646300002 seconds
5 - 687.2875056190111 - 11 minutes and 27.287505618999944 seconds
6 - 1304.5205741702346 - 21 minutes and 44.52057416999992 seconds
7 - 2664.6549223569446 - 44 minutes and 24.654922357000032 seconds
8 - 4693.939057922666 - 1 hour, 18 minutes, and 13.939057923000291 seconds
9 - 7067.282289407836 - 1 hour, 57 minutes, and 47.282289408 seconds
10 - 10819.430883151394 - 3 hours and 19.43088315100067 seconds
11 - 16283.021883556074 - 4 hours, 31 minutes, and 23.02188355600083 seconds
12 - 22716.493685539815 - 6 hours, 18 minutes, and 36.493685540001024 seconds
13 - 29218.06802800422 - 8 hours, 6 minutes, and 58.06802800399964 seconds
14 - 42323.63218544297 - 11 hours, 45 minutes, and 23.63218544299889 seconds
15 - 51947.17299507795 - 14 hours, 25 minutes, and 47.172995077999076 seconds
=> 1..15
```

modules/storages/app/workers/storages/copy_project_folders_job.rb
```diff
@@ -0,0 +1,103 @@
...
+#
+# See COPYRIGHT and LICENSE files for more details.
+#++
+
+module Storages
+  class CopyProjectFoldersJob < ApplicationJob
+    include GoodJob::ActiveJobExtensions::Batches
+
+    retry_on Errors::PollingRequired, wait: 3, attempts: :unlimited
+    discard_on HTTPX::HTTPError
```
akabiru, line 38: 🟢  Log error?

```suggestion
    discard_on HTTPX::HTTPError do |_job, error|
      ::OpenProject.logger.error error
    end
```

modules/storages/spec/common/storages/peripherals/storage_interaction/nextcloud/copy_template_folder_command_spec.rb
```diff
@@ -115,8 +115,7 @@
         result = subject.call(source_path:, destination_path:)
 
         expect(result).to be_success
-        expect(result.result[:id]).to eq('349')
-        expect(result.result[:url]).to eq(destination_url)
```
akabiru, line 119: 🟢  Q: anything changed with `destination_url`? i.e. why are we not asserting the expected result anymore?

modules/storages/spec/services/storages/project_storages/copy_project_folders_service_spec.rb
```diff
@@ -0,0 +1,93 @@
...
+  let(:result_data) { Storages::Peripherals::StorageInteraction::ResultData::CopyTemplateFolder.new(nil, nil, false) }
+
+  subject(:service) { described_class }
+
+  context "with automatically managed project folders" do
+    let(:source) { create(:project_storage, :as_automatically_managed, storage:) }
+
+    it 'if polling is required, returns a nil id and an url' do
+      Storages::Peripherals::Registry
+        .stub("#{source.storage.short_provider_type}.commands.copy_template_folder",
```
akabiru, line 47: 🟢 Q: Keen to better understand this `.stub` syntax - will reach out for a primer :)

akabiru (approved):
✨ Amazing feat @mereghost 👏🏾 and as always, thank you for the detailed PR description! 👍🏾 

---

✅ Code looks great to me- added Qs and discussion points
📆 Will reach out to you for some smoke testing

modules/storages/app/common/storages/peripherals/storage_interaction/nextcloud/copy_template_folder_command.rb
mereghost: Good catch! I'll change the name of the method to be longer but better descriptive.

modules/storages/app/common/storages/peripherals/storage_interaction/nextcloud/copy_template_folder_command.rb
mereghost: Yes! `COPY` is a WebDAV extension method. `Depth => Infinity` is to make sure that we get all files from all subfolders (so, infinite file tree depth).

app/workers/send_copy_project_status_email_job.rb
mereghost: `Batches` are awesome! :D

modules/storages/app/common/storages/peripherals/storage_interaction/result_data/copy_template_folder.rb
mereghost: It depends a bit on usage. Data is treated as value object (immutable etc), while Struct as, well, another object.

There are some implications to memory allocation etc, but the main one is usage. =)

modules/storages/app/services/storages/file_links/copy_file_links_service.rb
mereghost: Oooops, that was a debugging statement I left there xD I'll remove it.

modules/storages/app/workers/storages/copy_project_folders_job.rb
mereghost: GoodJob should already log discarded jobs on the worker logs. :)

modules/storages/spec/services/storages/project_storages/copy_project_folders_service_spec.rb
mereghost: `stub` (defined in `dry/container/stub`) allow us to override the registered value with something else.

In this case I'm just playing pretend that I called the command and got the result I was interested. It is great to test the current code without having to worry about the whole storage communication.

modules/storages/spec/common/storages/peripherals/storage_interaction/nextcloud/copy_template_folder_command_spec.rb
mereghost: Oh, it changed place and therefore is now tested elsewhere.

modules/storages/app/workers/storages/copy_project_folders_job.rb
mereghost: It looks great but I'd like to cap the max time. In this case we should probably cap it at around one hour. 

Why? This is the time that MS is taking copying the files/folders around. It should be pretty fast so maybe, a slower - and capped - growth rate might be better. So maybe reducing the growth factor a bit and increasing the number of tries could be better, idk.

mereghost: I've played around with different factors:

A factor of 4 gives us 15 retries in ~2 days
A factor of 3 gives us 30 retries in ~2 days
A factor of 2 gives us 80 retries in ~2 days

I'm tending towards a factor closer to 2, so we check more often and fail around the same time.

PS: I do think that 2 days is too much btw. ~12 hours should be the cap until we get yelled at.

mereghost: Code that I used to verify this:

```ruby
backoff_lambda = ->(jitter, factor, i) do
  (i**factor) + (Kernel.rand * (i**factor) * jitter) + 2
end.curry

presentation_lambda = ->(list) do
  list.each_with_index do |duration, index|
    puts "Execution: #{index + 1} - Wait Time: #{ActiveSupport::Duration.build(duration).inspect} - Total Elapsed time: #{ActiveSupport::Duration.build(list[0..index].sum).inspect}"
  end
end

factor4 = backoff_lambda.(0, 4)
factor3 = backoff_lambda.(0, 3)
factor2 = backoff_lambda.(0, 2)

puts "Factor 4"
presentation_lambda[1.upto(15).map(&factor4)]

puts "Factor 3"
presentation_lambda[1.upto(30).map(&factor3)]

puts "Factor 2"
presentation_lambda[1.upto(80).map(&factor2)]
```

Kharonus: Well, there are already a bunch of thoughts made on this topic. ;)

My 2 cent? 
- I love the idea to cap the retries and doing exponential backoff
- can we yell, when the limit is reached and last execution was unsuccessful, too? Or is the "yelling" printing an error somewhere?

cap of 12h or 24h should be fine, IMHO. More does not really make sense.

mereghost: When the job exhausts the number or retries it gets discarded (which should be logged). We could add the information about it to the email that gets send in the end.

PR #15087, opened 2024-03-22
Disable httpx persistent connection plugin.

ba1ash:
Due to it does not work properly with at least some nextlcoud installations.


machisuji (approved):
👍

PR #15111, opened 2024-03-26
Feature/53577 handle no active project attributes in project overview sidebar

jjabari-op:
https://community.openproject.org/projects/openproject/work_packages/53577


jjabari-op:
The implementation in this PR is currently just an idea how we could approach this topic quite easily. Thoughts?

I'm not too happy with the count query introduced in the meta tag rendering. I guess this query would be triggered in a lot of cases when it's actually not required for further rendering. It's only required for the project overview page. Maybe we can catch this with a condition checking for the current path.

dombesz (approved):

PR #15120, opened 2024-03-26
Bug/52798 add meaningful flash error message when user cancels oauth flow on onedrivesharepoint

akabiru:


mereghost (approved):

Kharonus:
⚠️ Usually we only change the basic `en.yml`. The other locales are created from crowdin. Are you sure, changing them here won't create any conflicts?

akabiru:
> ⚠️ Usually we only change the basic `en.yml`. The other locales are created from crowdin. Are you sure, changing them here won't create any conflicts?

Good shout. I understood Crowdin should overwrite this either way, let me update to only en.yml 👍🏾

Kharonus (approved):
LGTM

PR #15124, opened 2024-03-26
[fix] login into storage from project storage form

Kharonus:
This is very much related to:  [#53397](https://community.openproject.org/wp/53394). Without that fix, the whole bug was not fixed for me. @ba1ash , please recheck this code and maybe tell me, if I missed something.

### Steps to reproduce

- add a project storage of a storage, you do not have login data
- select folder mode `manual`
- login

#### What happened?

- folder mode was set to inactive

### What now happens?

- folder mode is reset to manual


ba1ash (approved):
Yes, that is perfectly makes sense!
Well caught! Thanks :heart:

PR #15125, opened 2024-03-26
Fix chrome reported issue

Eric-Guo:
Chrome 123 issue panel always reported: A form field element has neither an id nor a name attribute. 

image


oliverguenther (approved):
Good catch, thanks @Eric-Guo  👍

PR #15129, opened 2024-03-27
Bump to primer_view_components 0.28.1 and update PageHeaders in Core

bsatarnejad:
With version 0.26.0 of our primer_view_components, the definition of the `Primer::OpenProject::PageHeader` component has changed. So we have to update those places where it is already used.

**Places to update:**
- [x] Members page
- [x] Projects list page
- [x] Admin Project Attributes pages
- [x] Quarantined attachments page
- [x] Virus scanning page
- [x] Storages pages

https://community.openproject.org/work_packages/52746/activity

a related bugfix: https://community.openproject.org/projects/openproject/work_packages/53663/activity


Gemfile
```diff
@@ -383,4 +383,4 @@ end
 
 gem "openproject-octicons", "~>19.8.0"
 gem "openproject-octicons_helper", "~>19.8.0"
-gem "openproject-primer_view_components", "~>0.23.0"
```
HDinger, line 386: You need to commit the `Gemfile.lock` file as well.

app/components/members/index_page_header_component.html.erb
```diff
@@ -1,39 +1,27 @@
...
-              )
-            ) do |button|
-              button.with_leading_visual_icon(icon: :plus)
-              t('activerecord.models.member')
-            end
-          end
-        end
+    header.with_action_icon_button(mobile_icon: "filter",
+                                   scheme: :default,
+                                   icon: 'filter',
```
HDinger: Please use quotation marks consitently. We recently decided for the doubled `"`.

app/components/projects/index_page_header_component.html.erb
```diff
@@ -1,115 +1,108 @@
...
-              Primer::Beta::Text.new(tag: :span,
-                                     mr: BUTTON_MARGIN_RIGHT,
-                                     color: :muted)) do
-            t('lists.can_be_saved_as')
-          end
-        %>
+    <% if query_saveable? %>
+      <% header.with_action_text(tag: :span,
+                                    mr: BUTTON_MARGIN_RIGHT,
+                                    color: :muted) { t('lists.can_be_saved_as') } %>
```
HDinger: All three passed stystem_arguments are matching the default. So they can be removed

app/components/projects/index_page_header_component.html.erb
```diff
@@ -1,115 +1,108 @@
...
-                                         controller: "params-from-query",
-                                         'application-target': "dynamic",
-                                         'params-from-query-allowed-value': '["filters", "columns"]'
-                                       },
-                                       classes: 'Button--invisibleOP')) do |button|
-          button.with_leading_visual_icon(icon: :'op-save')
-
-          t('button_save_as')
-        end %>
+      <% header.with_action_button(mobile_icon: "op-save", mobile_label: t('button_save_as'),
```
HDinger: Please align the arguments correctly

app/components/projects/index_page_header_component.html.erb
```diff
@@ -1,115 +1,108 @@
...
+                                   scheme: :invisible,
+                                   size: :medium,
+                                   mr: BUTTON_MARGIN_RIGHT,
```
HDinger: Can be removed

app/components/projects/index_page_header_component.html.erb
```diff
@@ -1,115 +1,108 @@
...
+                                   tag: :a,
+                                   href: new_projects_query_path,
+                                   data: {
+                                     controller: "params-from-query",
+                                     'application-target': "dynamic",
+                                     'params-from-query-allowed-value': '["filters", "columns"]'
+                                   },
+                                   classes: 'Button--invisibleOP') do |button|%>
+        <% button.with_leading_visual_icon(icon: "op-save") %>
+        t('button_save_as')
```
HDinger: The brackets are missing `<% %>`
Bildschirmfoto 2024-04-02 um 08 54 29

app/components/projects/index_page_header_component.html.erb
```diff
@@ -1,115 +1,108 @@
...
-            href: gantt_portfolio_query_link,
-            id: 'projects-index-open-as-gantt',
-            content_arguments: { target: '_blank' }
-          ) do |item|
-            item.with_leading_visual_icon(icon: 'op-view-timeline')
-          end
-        end
+    <% end %>
 
+    <%= header.with_action_menu(menu_arguments: { anchor_align: :end }, button_arguments: { icon: "op-kebab-vertical", "aria-label": t(:label_more), data: { 'test-selector': 'project-more-dropdown-menu' } })  do |menu, button|
```
HDinger: Please align the arguments correctly.

app/components/projects/index_page_header_component.rb
```diff
@@ -69,12 +69,15 @@ def gantt_portfolio_project_ids
   def page_title
     query.name || t(:label_project_plural)
   end
-
```
HDinger: I guess this was an accident?

app/components/projects/index_page_header_component.rb
```diff
@@ -69,12 +69,15 @@ def gantt_portfolio_project_ids
...
   def query_saveable?
     current_user.logged? && query.name.blank?
   end
 
   def show_state?
     state == :show
   end
+
```
HDinger, line 79: Please remove also the definition of `BUTTON_MARGIN_RIGHT`

app/components/projects/index_page_header_component.rb
```diff
@@ -69,12 +69,15 @@ def gantt_portfolio_project_ids
...
+  def breadcrumb_items
+    [{ href: projects_path, text: t("en.types.edit.projects") }, page_title]
```
HDinger: The string definition is not correct

app/components/settings/project_custom_fields/edit_form_header_component.rb
```diff
@@ -34,6 +34,11 @@ def initialize(custom_field:)
 
         @custom_field = custom_field
       end
+
+      def breadcrumbs_items
+        [{ href: admin_settings_project_custom_fields_path, text: t("label_administration") },
```
HDinger: I think this definition is not correct. Since this is an Edit page, I expect a three parts breadcrumb:

Bildschirmfoto 2024-04-02 um 09 04 41

app/components/settings/project_custom_fields/header_component.rb
```diff
@@ -32,6 +32,10 @@ class HeaderComponent < ApplicationComponent
       include ApplicationHelper
       include OpPrimer::ComponentHelpers
       include OpTurbo::Streamable
+      def breadcrumbs_items
```
HDinger, line 36: new line missing before the method definition

app/components/settings/project_custom_fields/header_component.html.erb
```diff
@@ -4,30 +4,29 @@
...
-              tag: :a,
-              href: new_admin_settings_project_custom_field_path(type: "ProjectCustomField"),
-              scheme: :primary,
-              data: { turbo: "false", qa_selector: "new-project-custom-field-button" }
-            )) do |button|
-              button.with_leading_visual_icon(icon: :plus)
-              t('settings.project_attributes.label_new_attribute')
-            end
+      header.with_breadcrumbs(breadcrumbs_items)
+      flex_layout(justify_content: :space_between, align_items: :center) do |action_buttons_container|
```
HDinger: Remove the flex_layout. It is not needed any more. Further, the second button is now shown at all

app/components/settings/project_custom_fields/header_component.rb
```diff
@@ -32,6 +32,10 @@ class HeaderComponent < ApplicationComponent
...
+        [{ href: admin_index_path, text: t("label_administration") },
```
HDinger, line 37: Clicking that link results in a crash of the page, as it violates the content secruity policy.

app/controllers/admin/attachments/quarantined_attachments_controller.rb
```diff
@@ -51,12 +51,8 @@ def destroy
         redirect_to action: :index
       end
 
-      def default_breadcrumb
-        t("antivirus_scan.quarantined_attachments.title")
-      end
-
       def show_local_breadcrumb
-        true
+        false
```
HDinger, line 57: The whole method can be removed as `false` is the default.

app/controllers/admin/attachments/quarantined_attachments_controller.rb
```diff
@@ -78,6 +74,13 @@ def find_attachment
       rescue ActiveRecord::RecordNotFound
         render_404
       end
+
+      def breadcrumb_items
+        [{ href: admin_index_path, text: t("label_administration") },
+         { href: admin_settings_attachments_path, text: t("attributes.attachments") },
```
HDinger: I am not sure, whether we want the "Attachments" as an extra element. For the project attributes we did not add the intermediate step. I will double check with the product team

app/controllers/admin/settings/project_custom_fields_controller.rb
```diff
@@ -39,16 +39,8 @@ class ProjectCustomFieldsController < ::Admin::SettingsController
...
-      if action_name == "index"
-        t("label_project_attribute_plural")
-      else
-        ActionController::Base.helpers.link_to(t("label_project_attribute_plural"), admin_settings_project_custom_fields_path)
-      end
-    end
-
     def show_local_breadcrumb
-      true
+      false
```
HDinger, line 43: Method can be removed as well

app/controllers/admin/settings/virus_scanning_settings_controller.rb
```diff
@@ -33,10 +33,9 @@ class VirusScanningSettingsController < ::Admin::SettingsController
     before_action :require_ee
     before_action :check_clamav, only: %i[update], if: -> { scan_enabled? }
 
-    def default_breadcrumb
-      t("settings.antivirus.title")
+    def show_local_breadcrumb
+      false
```
HDinger, line 37: Same as above

app/views/admin/attachments/quarantined_attachments/index.html.erb
```diff
@@ -3,6 +3,7 @@
 <%=
   render(Primer::OpenProject::PageHeader.new) do |header|
     header.with_title(variant: :large) { t('antivirus_scan.quarantined_attachments.title') }
```
HDinger: Please remove the `:large` variant, as we want the default one

frontend/package.json
```diff
@@ -98,7 +98,7 @@
     "@ngneat/content-loader": "^7.0.0",
     "@ngx-formly/core": "^6.1.4",
     "@openproject/octicons-angular": "^19.8.0",
-    "@openproject/primer-view-components": "^0.23.0",
```
HDinger, line 101: Please commit the `package-lock.json`

modules/storages/app/controllers/storages/admin/storages_controller.rb
```diff
@@ -208,7 +208,7 @@ def default_breadcrumb
   # See: default_breadcrum above
   # Defines whether to show breadcrumbs on the page or not.
   def show_local_breadcrumb
-    true
+    false
```
HDinger, line 202: Whole method can be removed

frontend/src/global_styles/openproject/_primer-adjustments.sass
```diff
@@ -49,7 +49,7 @@ ul.tabnav-tabs
   margin-left: 0
 
 /* Remove margin-left: 2rem from Breadcrumbs */
-#breadcrumb
+#breadcrumb,page-header
```
HDinger: A bit further down in that file there is that block 

```
/* TODO: The actions within the PageHeader are currently not aligned correctly.
   The pageHeader itself already uses center alignment but the actions themselves default to 'normal'.
   Will be added to our primer repository */
.PageHeader-actions
  align-items: center
```

Could you please remove that, as it is now part of the PageHeader and thus not needed any more

app/components/projects/index_page_header_component.html.erb
HDinger: Pressing the button, currently leads to an error.

HDinger (changes requested):
Hi @bsatarnejad 

I now that this is still in a Draft state, but I had a look already as I had some time at hand.

app/controllers/admin/attachments/quarantined_attachments_controller.rb
HDinger: I discussed this with the product team and we agreed, that we want the menu header as an element in the breadcrumb (like it is here). So could you please add that to the project attributes and file storages pages?

app/components/settings/project_custom_fields/header_component.rb
HDinger: Might be related to https://community.openproject.org/projects/openproject/work_packages/53813/activity

HDinger (approved):

app/components/projects/index_page_header_component.rb
```diff
@@ -31,14 +31,13 @@
 class Projects::IndexPageHeaderComponent < ApplicationComponent
   include OpPrimer::ComponentHelpers
   include Primer::FetchOrFallbackHelper
+  include Menus::ProjectsHelper
```
ulferts, line 34: I couldn't find the reason for including this module. Which method within it is referenced?

HDinger: The menu_items are created in there. We currently need those to create the correct breadcrumb. It is used [here](https://github.com/opf/openproject/pull/15129/files/1a5581b54caace87c57f93a856a290f895ac63f8#diff-74b240ebb55cde576590dc2a1c56a6eee5ae87b809524abfe663ade8c9796cb3R90).

app/components/members/index_page_header_component.html.erb
```diff
@@ -1,39 +1,27 @@
...
+    header.with_action_button(scheme: :primary,
+                              mobile_icon: :plus,
+                              mobile_label: t('activerecord.models.member'),
+                              size: :medium,
+                              aria: { label: I18n.t(:button_add_member) },
+                              title: I18n.t(:button_add_member),
+                              id: "add-member-button",
+                              data: add_button_data_attributes) do |button|
+      button.with_leading_visual_icon(icon: :plus)
+      t('activerecord.models.member')
```
ulferts, line 15: The order of the filter and the "+Member" buttons are switched here. With the discussion being unfinished, I would leave the order as it is.

app/components/projects/index_page_header_component.html.erb
```diff
@@ -1,115 +1,112 @@
...
+                                 data: {
+                                   controller: "params-from-query",
+                                   'application-target': "dynamic",
+                                   'params-from-query-allowed-value': '["filters", "columns"]'
+                                 }) do %>
+        <% render(Primer::Beta::Octicon.new(icon: "op-save",
+                                            align_self: :center,
+                                            "aria-label": I18n.t("button_save_as"),
+                                            mr: 1)
+            ) + content_tag(:span, t("button_save_as")) %>
```
ulferts: It would read easier if one could say something like:

```

header.with_action_link(**options) do |link|
  link.with_leading_visual_icon('op-save')
  t('button_save_as')
end  

```
or have the text and the icon be part of the `with_action_link` method. But that does not seem to be along the line of thinking of how Primer components are build.

What also strikes me as weird is that a link is used and not a button. This seems to violate the statement of primer: "Links navigate you to a new place or new content. Do not use links for triggering an action. Instead, use [buttons](https://primer.style/components/button), which are designed to activate a feature." ([source](https://primer.style/components/link#usage))

app/components/projects/index_page_header_component.html.erb
```diff
@@ -1,115 +1,112 @@
...
-            label: t('projects.index.open_as_gantt'),
-            href: gantt_portfolio_query_link,
-            id: 'projects-index-open-as-gantt',
-            content_arguments: { target: '_blank' }
-          ) do |item|
-            item.with_leading_visual_icon(icon: 'op-view-timeline')
-          end
-        end
-
+    <%= header.with_action_menu(menu_arguments: {
```
ulferts: The `=` part of the text interpolation (`<%=`) is no longer necessary I think. In fact, one could remove a whole lot of text interpolation tags and just write the code in one block now thanks to the work done on the slots. Doesn't have to happen in this PR.

app/components/projects/index_page_header_component.html.erb
```diff
@@ -1,115 +1,112 @@
...
+                                  anchor_align: :end
+                                },
+                                button_arguments: {
+                                  icon: "op-kebab-vertical",
+                                  "aria-label": t(:label_more),
+                                  data: { "test-selector": "project-more-dropdown-menu" }
+                                })  do |menu, button|
```
ulferts: The `button` argument is never used - should be removable.

app/components/projects/index_page_header_component.rb
```diff
@@ -77,4 +76,25 @@ def query_saveable?
...
+      { href: projects_path, text: t(:label_project_plural) },
+      current_breadcrumb_element
+    ]
+  end
+
+  def current_breadcrumb_element
+    return page_title if query.name.blank?
+
+    current_object = first_level_menu_items.find do |section|
+      section.children.find { |menu_query| menu_query.title == query.name }.present?
```
ulferts: This is currently only a problem in the rare case where a user chooses the name of the static queries in the "Project status" section for their own private list but then, the breadcrumb can be wrong:

image

Because of the name clash, the first section is assumed to be the correct one when in fact the user clicked on the one in the "Project status" section.

When using 
```
 section.children.any? { |menu_query| menu_query.selected }
```

it becomes easier to read and shouldn't suffer from the aforementioned problem

app/components/settings/project_custom_fields/header_component.rb
ulferts: Problem still exists

app/controllers/projects/settings/project_custom_fields_controller.rb
```diff
@@ -106,4 +106,11 @@ def bulk_edit_service
         project_custom_field_section: @project_custom_field_section
       )
   end
+
+  def breadcrumb_items
+    [{ href: project_overview_path(@project.id), text: @project.name },
+     { href: project_settings_general_path(@project.id), text: I18n.t("label_project_settings") },
+     t("settings.project_attributes.heading")]
+  end
+  helper_method :breadcrumb_items
```
ulferts: It would be cleaner if those methods wouldn't be in the controller but rather in a component. For some of the pages no dedicated header component exists yet, so they will need to be created in those cases. Alternatively, the code should be placed inside the view or in the corresponding helper.

ulferts (changes requested):
The look of it is nice. I in particular like the mobile behaviour.

There are a couple of remarks on the code that need to be addressed before merging. Of those, the one about the CSP is I guess the most troublesome. I don't think it originates in this PR but it is something we need to fix before the release.

app/components/projects/index_page_header_component.html.erb
HDinger: > header.with_action_link(**options) do |link|
>  link.with_leading_visual_icon('op-save')
>  t('button_save_as')
> end  

The link component unfortunately does not support leading visual (although they mention it in the docs, it is not implemented).

The product team and I discussed the decision for the link for quite some time. Of course, the invisible button would be the best solution. However, Primer has this weird behavior  of changing the font color of that buttons depending on the existance of an icon. The issue I raised for that in Primer was closed witht the comment that it "is working as expected" 🤷  Since GitHub and even Primer itself is very fuzzy when it comes to their usage of Buttons vs Links, we decided to go with the Link at this place.

PR #15134, opened 2024-03-27
[#53577] Use sidebar_enabled method on the overview page instead of project meta tags

dombesz:
https://community.openproject.org/work_packages/53577

Based on the comment https://github.com/opf/openproject/pull/15111#issuecomment-2019428730

Optimize the logic to not introduce an extra query for all the project related pages when calculating whether the Overview page sidebar should be enabled or not.

### DIFF
--- a/modules/overviews/app/controllers/overviews/overviews_controller.rb
+++ b/modules/overviews/app/controllers/overviews/overviews_controller.rb
@@ -2,8 +2,8 @@ module ::Overviews
   class OverviewsController < ::Grids::BaseInProjectController
     include OpTurbo::ComponentStream
 
-    before_action :authorize
     before_action :jump_to_project_menu_item
+    before_action :set_sidebar_enabled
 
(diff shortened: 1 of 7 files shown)


modules/overviews/app/controllers/overviews/overviews_controller.rb
```diff
@@ -2,8 +2,8 @@ module ::Overviews
   class OverviewsController < ::Grids::BaseInProjectController
     include OpTurbo::ComponentStream
 
-    before_action :authorize
```
dombesz, line 5: @jonasjabari the `::Grids::BaseInProjectController` already has this before action, calling it here will result in two code executions.

https://github.com/opf/openproject/blob/17b5a0c94cf6ee0ebdbbd019c33353902337eee2/modules/grids/app/controllers/grids/base_in_project_controller.rb#L4

jjabari-op (approved):

PR #15140, opened 2024-03-27
[53768] Fix impressum translation missing

cbliard:
https://community.openproject.org/wp/53768


aaron-contreras (approved):
Unrelated failure on CI. Thanks for the fix!

PR #15147, opened 2024-03-28
Fix css typo

baldarn:
Tried to compile the css and this was shown as warning. I think this was a typo here: https://github.com/opf/openproject/commit/e1c9266eaeb26655c9f23d885e24e7ef54dbe66d


oliverguenther (approved):
Good catch, thanks @baldarn 👍

PR #15149, opened 2024-03-28
Write a documentation to set up `keycloak` integration with `OpenProject` via docker `TLS`

SagarGi:
### Description
- As a first phase this PR adds a documentation for setting up `keycloak` with `OpenProject` via docker `TLS`.
- The next phase will be documenting the keycloak with `nextcloud` for testing SSO (Single sign on) as a part of (OpenProject Integration with nextcloud which needs TBD). 

### Related WorkPackage:
https://community.openproject.org/work_packages/53590/activity


docs/development/development-environment-docker/README.md
```diff
@@ -420,6 +420,41 @@ Should you need to reset your root password, execute the following command:
...
+docker compose --project-directory docker/dev/keycloak up -d
+```
+
+Once the keycloak service is started and running, you can access the keycloak instance on `https://keycloak.local` 
+and login with initial username and password as `admin`.
+
+Keycloak being an OpenID connect provider, we need to setup an OIDC integration for OpenProject.
+[Setup OIDC (keycloak) integration for OpenProject](https://www.openproject.org/docs/installation-and-operations/misc/custom-openid-connect-providers/#keycloak)
+
+Once the above setup is completed, In the root `docker-compose.override.yml` file, uncomment all the environment in `backend` service for keycloak and set the values according to configuration done in keycloak for OpenProject Integration.
```
nabim777, line 445: In the future, there may need to include some uncommented code in the backend environment. So, it would be good to have a list of environment variables that need to be uncommented for Keycloak.

ba1ash (approved):
Thanks @SagarGi 
It looks and works fine. Let's merge it.

PR #15150, opened 2024-03-28
[fix] remove caching httpx sessions for reuse

Kharonus:
### Reasoning

- reused session caused under certain circumstances http requests to fail
- reusing sessions was mandatory for persistent connections, but those are not used anymore

### Example test execution:

- run on `dev` before this fix:

```
rspec './modules/storages/spec/common/storages/peripherals/registry_spec.rb[1:2:1]' './modules/storages/spec/requests/api/v3/file_links/file_links_api_spec.rb[1:4:1]' --seed 8189
```

### DIFF
--- a/lib/open_project.rb
+++ b/lib/open_project.rb
@@ -26,17 +26,17 @@
 # See COPYRIGHT and LICENSE files for more details.
 #++
 
-require 'redmine/menu_manager'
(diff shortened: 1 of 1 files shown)


machisuji (approved):
👍

PR #15164, opened 2024-04-03
[#53718] added access management form section

Kharonus:
[#53718](https://community.openproject.org/wp/53718)

### WAT?

- Added new access management section after general information
  - only for one drive storage configuration
  - "Access Management" is supposed to be a new domain term, replacing (or better covering) project folder configuration for storages (not project storages) -> Access management describes more generally, that in the AMPF decision case, the OP instance takes care to guarantee the user access on the file storage
- removed previous project folder checkbox from general information
- adapted turbo streams
- added new small controller for access management


modules/storages/app/controllers/storages/admin/storages_controller.rb
```diff
@@ -85,7 +85,7 @@ def upsale; end
 
   def select_provider
     @object = Storages::Storage.new(provider_type: @provider_type)
-    service_result = ::Storages::Storages::SetProviderFieldsAttributesService
+    service_result = ::Storages::Storages::SetAttributesService
```
akabiru, line 88: 🟢 Q: Is this no longer needed for Nextcloud default `automatically_managed = true` setting? FWIW, testing went Ok just seeking clarification.

akabiru (changes requested):
Coming along nicely. 👍🏾 I've noticed the following some odd behaviors that I'd like to clarify with you.

🟠 (1) Access management is _still_ editable? I understood this should NOT be possible?
🔴 (2) Upon editing the Access management state, the OAuthClient is purged from the DB for some reason. _Here's a Video Demo_

```ruby
[7] pry(main)> od.oauth_client
  OAuthClient Load (1.0ms)  SELECT "oauth_clients".* FROM "oauth_clients" WHERE "oauth_clients"."integration_id" = $1 AND "oauth_clients"."integration_type" = $2 LIMIT $3  [["integration_id", 2], ["integration_type", "Storages::Storage"], ["LIMIT", 1]]
=> #<OAuthClient:0x0000ffff94dfa518
 id: 5,
 client_id: "123",
 client_secret: "123",
 integration_type: "Storages::Storage",
 integration_id: 2,
 created_at: Thu, 04 Apr 2024 10:47:58.263655000 UTC +00:00,
 updated_at: Thu, 04 Apr 2024 10:47:58.263655000 UTC +00:00>
[8] pry(main)> od.reload
  Storages::OneDriveStorage Load (1.6ms)  SELECT "storages".* FROM "storages" WHERE "storages"."provider_type" = $1 AND "storages"."id" = $2 LIMIT $3  [["provider_type", "Storages::OneDriveStorage"], ["id", 2], ["LIMIT", 1]]
=> #<Storages::OneDriveStorage:0x0000ffff950d6f20
 id: 2,
 provider_type: "Storages::OneDriveStorage",
 name: "Eric's OneDrive Sandbox",
 host: nil,
 creator_id: 4,
 created_at: Tue, 26 Mar 2024 15:36:17.112900000 UTC +00:00,
 updated_at: Thu, 04 Apr 2024 10:36:26.459426000 UTC +00:00,
 provider_fields:
  {"drive_id"=>"b!FeO...",
   "tenant_id"=>"e3cc...",
   "automatically_managed"=>true},
 health_status: "pending",
 health_changed_at: Tue, 26 Mar 2024 15:36:17.086243000 UTC +00:00,
 health_reason: nil,
 health_checked_at: Tue, 26 Mar 2024 15:36:17.086243000 UTC +00:00>
[9] pry(main)> od.reload.oauth_client
  Storages::OneDriveStorage Load (0.9ms)  SELECT "storages".* FROM "storages" WHERE "storages"."provider_type" = $1 AND "storages"."id" = $2 LIMIT $3  [["provider_type", "Storages::OneDriveStorage"], ["id", 2], ["LIMIT", 1]]
  OAuthClient Load (0.6ms)  SELECT "oauth_clients".* FROM "oauth_clients" WHERE "oauth_clients"."integration_id" = $1 AND "oauth_clients"."integration_type" = $2 LIMIT $3  [["integration_id", 2], ["integration_type", "Storages::Storage"], ["LIMIT", 1]]
=> nil
```

https://github.com/opf/openproject/assets/17295175/bec5ebbc-3cda-44a0-b14c-c32dfcd97c22

🔴 (3) One Drive Delete flow breaks with `undefined method `client_id' for nil:NilClass`

https://github.com/opf/openproject/assets/17295175/be44b81f-c18c-424b-a928-9c03a9ab2802

```ruby
[Application Trace](https://openproject.local/admin/settings/storages/2#) | [Framework Trace](https://openproject.local/admin/settings/storages/2#) | [Full Trace](https://openproject.local/admin/settings/storages/2#)
modules/storages/app/common/storages/peripherals/oauth_configurations/one_drive_configuration.rb:71:in `basic_rack_oauth_client'
modules/storages/app/common/storages/peripherals/storage_interaction/one_drive/util.rb:77:in `using_admin_token'
modules/storages/app/common/storages/peripherals/storage_interaction/one_drive/delete_folder_command.rb:46:in `call'
modules/storages/app/common/storages/peripherals/storage_interaction/one_drive/delete_folder_command.rb:37:in `call'
modules/storages/app/services/storages/project_storages/delete_service.rb:63:in `delete_project_folder'
modules/storages/app/services/storages/project_storages/delete_service.rb:36:in `before_perform'
```

modules/storages/app/controllers/storages/admin/storages_controller.rb
Kharonus: Nextcloud setup does not need to set the provider fields in the `select_provider` action (which is the first one executed when the multistep form is loaded, i.e. "empty general info"). The provider field defaults are inserted into the model before entering the last form step, with the method called `prepare_storage_for_automatic_management_form` in the `oauth_clients_controller`.

Kharonus: It was needed before, as OneDrive setup had the checkbox for AMPF in the general info step.

Kharonus:
(1) Restricting edit should be solved in https://community.openproject.org/wp/53938

Will check for the other two things.

modules/storages/app/controllers/storages/admin/storages_controller.rb
akabiru: Ah, got you. Cheers 👍🏾

Kharonus:
I fixed (2) by adding a `update` stream action. First I was unaware, that there is actually a difference between create and update here. But there is, exactly in the use case you described in (2).

I cannot reproduce (3). Every storage I had I could delete. :/ no idea, why you storage has no oauth client.

modules/storages/app/controllers/storages/admin/access_management_controller.rb
```diff
@@ -0,0 +1,115 @@
...
+    service_result.on_success do
+      respond_to { |format| format.turbo_stream }
+    end
+
+    service_result.on_failure do
+      respond_to do |format|
+        format.turbo_stream { render :edit }
+      end
+    end
+  end
```
akabiru, line 81: 🟠 One issue I'm not certain about is why we have different behavior on `#update` & `#create` whilst we call the same update service

akabiru (approved):
One point of clarification, otherwise looks good! :shipit:

modules/storages/app/controllers/storages/admin/access_management_controller.rb
Kharonus: The difference is in the related turbo-stream update.

on a create, it should close this form section and open a oauth client section with a freshly build oauth client.
on an update, it should only close this form section.
the respective turbo stream files define that behavior.

so, the pencil icon actually leads to the edit action, followed by the update action (convention). I tried to deduplicate the code a bit, but then I have to fiddle around a bit with conventional behaviour. And I failed on that. 😬

akabiru: Got it, thanks. I think we can revisit this flow separately as I find it rather odd that rendering the OAuth form already purges the previous state. It's particularly odd when one clicks on "Cancel" within the OAuth form and finds the values already purged. I'd expect us to replace on submit.

This is naturally a separate topic from this PR 🔖

PR #15166, opened 2024-04-04
Bug/53706 reordering project attributes is popping back on render

jjabari-op:
https://community.openproject.org/projects/openproject/work_packages/53706


jjabari-op:
@dombesz do you have an idea why the CI build step fails?

dombesz (approved):
LGMT 👍

dombesz:
> @dombesz do you have an idea why the CI build step fails?

@jjabari-op according to the logs, the field `position_in_custom_field_section` does not exist when the migration `CreateProjectCustomFieldProjectMappings` is ran:

```shell
PG::UndefinedColumn: ERROR:  column "position_in_custom_field_section" does not exist
LINE 1: ...ields" WHERE "custom_fields"."type" = $1 ORDER BY "position_...
                                                             ^
/app/db/migrate/20231031133334_create_project_custom_field_project_mappings.rb:23:in `create_default_mapping'
/app/db/migrate/20231031133334_create_project_custom_field_project_mappings.rb:12:in `up'
```

This happens, because calling the `ProjectCustomField` in the line `20231031133334_create_project_custom_field_project_mappings.rb:23:` will trigger the default scope, but the `position_in_custom_field_section` does not exist yet, because it is added in a latter migration. `/app/db/migrate/20231123111357_create_custom_field_sections.rb`

Usually the issue is solved by defining a custom class for the table inside the migration. I provided a fix for it 1adf0804f6d5

dombesz:
@jonasjabari Seems like the default scope ordering on the `ProjectCustomField` class is causing issues on merged queries. I would suggest adding the default order on the `ProjectCustomFieldSection#custom_fields` relation instead. I provided a commit for my suggestion.

jjabari-op:
Thank you @dombesz

PR #15174, opened 2024-04-05
Feature/53580 check if and how project activity logs should be adapted

dombesz:
https://community.openproject.org/wp/53580

<details><summary>Additionally displaying a "The changes were retracted" message, when empty journals are displayed. [Click to demo]</summary>
<p>

Apr-08-2024 14-02-28

</p>
</details>

### DIFF
--- a/app/components/activities/item_component.html.erb
+++ b/app/components/activities/item_component.html.erb
@@ -67,5 +67,11 @@ See COPYRIGHT and LICENSE files for more details.
         <li class="op-activity-list--item-detail"><%= link_to "Details", @event.event_url %></li>
       <% end -%>
     </ul>
+  <% elsif noop? -%>
+    <ul class="op-activity-list--item-details">
+      <li class="op-activity-list--item-detail">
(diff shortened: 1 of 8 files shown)


ulferts (changes requested):
The functionality is fine and I also like the solution of having the changes_retracted displayed in case there is no change left. 

What I struggle with is the way how and the place where the change is modified. For one, it is very specific to Project but it is executed for every change. And it is executed within the fetcher after all the changes have already been processed. If we ever want to send out notifications, we would have to reimplement the behaviour. 

I have no solution for the project specific code affecting all the other models as well but I  think placing the code in the [`get_changes` method](https://github.com/opf/openproject/blob/dev/lib_static/plugins/acts_as_journalized/lib/journal_changes.rb) would be a more stable place.

dombesz:
Thank you for the insight @ulferts , I moved the logic into the `JournalChanges` which seems to be a more appropriate place for this.

oliverguenther (approved):

PR #15185, opened 2024-04-08
Hide storage link if user has no ':read_files' permission

apfohl:
This PR implements a switch where the link next to the storage heading is hidden, when a user does not have the `:read_files` permission.

It is done by hiding the two links inside the response of the `ProjectStorages` response of the HAL API.

### DIFF
--- a/frontend/src/app/features/work-packages/components/wp-single-view-tabs/files-tab/op-files-tab.component.ts
+++ b/frontend/src/app/features/work-packages/components/wp-single-view-tabs/files-tab/op-files-tab.component.ts
@@ -37,9 +37,7 @@ import { map } from 'rxjs/operators';
 
 import { WorkPackageResource } from 'core-app/features/hal/resources/work-package-resource';
 import { I18nService } from 'core-app/core/i18n/i18n.service';
-import { HookService } from 'core-app/features/plugins/hook-service';
 import { CurrentUserService } from 'core-app/core/current-user/current-user.service';
-import { StoragesResourceService } from 'core-app/core/state/storages/storages.service';
 import { HalResource } from 'core-app/features/hal/resources/hal-resource';
 import { ProjectStoragesResourceService } from 'core-app/core/state/project-storages/project-storages.service';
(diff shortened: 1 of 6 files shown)


frontend/src/app/features/work-packages/components/wp-single-view-tabs/files-tab/op-files-tab.component.ts
Kharonus, line 1: 👍🏽 for cleaning

frontend/src/app/shared/components/storages/storage/storage.component.html
```diff
@@ -6,12 +6,12 @@
       class="op-tab-content--header-icon spot-icon spot-icon_{{icon.storageHeader((storage | async)?._links.type.href)}} op-files-tab--icon op-files-tab--icon_primary"
     ></span>
     <span class="op-tab-content--header-text" [textContent]="(storage | async)?.name"></span>
-    <a
+    <a *ngIf="projectStorage._links.openWithConnectionEnsured || projectStorage._links.open"
```
Kharonus: 🟡 Does it make sense to extract that to a getter in the controller? so that we do not have a 90% duplicate to line 13?

like 

```typescript
public get openStorageLink(): {
  return this.projectStorage._links.openWithConnectionEnsured.href || this.projectStorage._links.open.href
}
```

it then can be used in both places.

modules/storages/lib/api/v3/project_storages/project_storage_representer.rb
```diff
@@ -47,10 +47,14 @@ class ProjectStorageRepresenter < ::API::Decorators::Single
     end
 
     link :open do
+      next unless can_read_files
```
Kharonus: 🔴 this is currently not correct, as it would hide this link also for project storages with inactive or manual project folders.

I'd rephrase the helper method to `show_open_storage_links`, and include the check for only `automatic_management_enabled?`

Kharonus (changes requested):

Kharonus (approved):
LGTM

PR #15186, opened 2024-04-08
[Op#52449]: Add toggle to unsubscribe/subscribe admins health notifications for a storage

akabiru:
#### https://community.openproject.org/wp/52449

Supersedes: https://github.com/opf/openproject/commit/eceb8ce0ff1c0a8c6e71a83d68f5a892f878a56c

**There are two notification events to manage**

1.  Immediate notifications on storage health/unhealthy events managed via the `HealthService`
2. Daily health status mailer job that run at ~2 a.m via the `HealthStatusMailerJob` - For this one, **we run the notifications check inline** so that admins who have since updated their notifications preferences will still get notifications if subscribed at the time the job runs. 

https://github.com/opf/openproject/assets/17295175/74bc8307-a591-4559-b625-5696c8f11b3c

<details> <summary>Mobile Screenshot</summary>

Screenshot 2024-04-09 at 5 39 06 PM

</details>


modules/storages/app/controllers/storages/admin/storages_controller.rb
```diff
@@ -162,6 +165,20 @@ def update
     end
   end
 
+  def change_health_email_notifications
+    return head :forbidden unless %w[1 0].include?(permitted_storage_params[:health_notifications_enabled])
```
ba1ash: I think `forbidden` is semantically not correct here. What about 204: `head :no_content` or 400: `head :bad_request`?

modules/storages/app/controllers/storages/admin/storages_controller.rb
```diff
@@ -162,6 +165,20 @@ def update
     end
   end
 
+  def change_health_email_notifications
```
ba1ash: Why invent new name here? Let it be `def change_health_notifications_enabled`? Then it is not misleading about what it does.

modules/storages/app/controllers/storages/admin/storages_controller.rb
```diff
@@ -162,6 +165,20 @@ def update
...
+    return head :forbidden unless %w[1 0].include?(permitted_storage_params[:health_notifications_enabled])
+
+    if @storage.update(health_notifications_enabled: permitted_storage_params[:health_notifications_enabled])
+      update_via_turbo_stream(component: Storages::Admin::Sidebar::HealthNotificationsComponent.new(storage: @storage.reload))
```
ba1ash: Why are we reloading `@storage` here? It seems to be a non necessary request to me. We are under condition about successful update. Therefore `@stprage.health_notifications_enabled` should be up to date already.

modules/storages/app/models/storages/storage.rb
```diff
@@ -53,6 +53,7 @@ class Storage < ApplicationRecord
     self.inheritance_column = :provider_type
 
     store_attribute :provider_fields, :automatically_managed, :boolean
+    store_attribute :provider_fields, :health_notifications_enabled, :boolean, default: true
```
ba1ash, line 56: I remember problems with this default values. They only work for new records as far as I remember. But I do not see it as a problem in this case. I think it means that existing storages should enable notification explicitly from the UI.

modules/storages/app/workers/storages/health_status_mailer_job.rb
```diff
@@ -40,14 +40,15 @@ class HealthStatusMailerJob < ApplicationJob
     )
 
     def perform(storage:)
-      return unless Storages::Storage.exists?(storage.id)
+      return unless ::Storages::Storage.exists?(storage.id)
```
ba1ash: It is not related to your changes. But this line seems to be obsolete to me. Because ActiveJob will already raise an error if there is no storage with such id. It can be just removed(then error will be raised) or rewritten as(no errors):
```ruby 
    def perform(storage_id:)
      storage = Storages::Storage.find_by(id: storage_id)
      return unless storage
```
The error will be:
```
ActiveJob::DeserializationError: Error while trying to deserialize arguments: Couldn't find Storages::NextcloudStorage with 'id'=2 [WHERE \"storages\".\"provider_type\" = $1]
```

cc @apfohl

modules/storages/app/workers/storages/health_status_mailer_job.rb
```diff
@@ -40,14 +40,15 @@ class HealthStatusMailerJob < ApplicationJob
...
 
       storage.reload
```
ba1ash:
```suggestion
```
The reload is unnecessary, because the storage has just been loaded from the database. It is not related to your changes as well. But you could just remove it I think.

ba1ash (commented):
I tested locally. It works fine to me. Good job :+1: 
I have a couple of comments. Some of them related to your changes and others are not. Kindly have a look.

modules/storages/app/components/storages/admin/sidebar/health_notifications_component.html.erb
```diff
@@ -0,0 +1,36 @@
+<%=
```
apfohl, line 30: You missed the copy right here.

modules/storages/app/controllers/storages/admin/storages_controller.rb
akabiru: Good spot, yeah `bad_request` is the more semantic option

modules/storages/app/controllers/storages/admin/storages_controller.rb
akabiru: Yeah, I went through a few iterations with the name and forgot to update. `change_health_notifications` sounds more fitting to me. Your call 👍🏾

modules/storages/app/models/storages/storage.rb
akabiru: Ah, good spot- I think we can base it off `automatically_managed` as we have the AC that it should be on by default for all storages with automatic management.

Thoughts on this instead?

```ruby
store_attribute :provider_fields, :health_notifications_enabled, :boolean

def health_notifications_enabled
  super.presence || automatic_management_enabled?
end
```

modules/storages/app/components/storages/admin/sidebar/health_notifications_component.rb
```diff
@@ -0,0 +1,61 @@
...
+#++
+
+module Storages
+  module Admin
+    class Sidebar::HealthNotificationsComponent < ApplicationComponent # rubocop:disable OpenProject/AddPreviewForViewComponent
+      include ApplicationHelper
+      include OpTurbo::Streamable
+      include OpPrimer::ComponentHelpers
+
+      attr_reader :storage
```
apfohl: Why do you need to make the storage accessible to the outside from this component class? The view does not need it.

modules/storages/app/workers/storages/health_status_mailer_job.rb
akabiru: I agree, fyi- it should be a global id as the full storage is passed. Either way, it should fail if it doesn't exist. But perhaps @apfohl can chime in in case we're missing sth.

The top level namespacing was necessary for unit testing, we didn't have these before which is why it wasn't an issue. Running the `health_status_mailer_job_spec` should reproduce it.

apfohl: @ba1ash But then you get an error that needs to be either handled or ignored. Leaving this out would feel a bit incomplete. The purpose to have it in the code is also documentation what happens during `perform`.

I would leave it in for that reason.

modules/storages/app/workers/storages/health_status_mailer_job.rb
apfohl: Why was the storage just loaded from the database, when the job got dispatched 12h ago?

apfohl (changes requested):

modules/storages/app/controllers/storages/admin/storages_controller.rb
akabiru: yeah, we should be able to do without it 👍🏾

modules/storages/app/components/storages/admin/sidebar/health_notifications_component.rb
akabiru: It's just a convenience instance reader, in other storage components we have `alias_method :storage :model` - this way we don't have to access via the instance var.

modules/storages/app/workers/storages/health_status_mailer_job.rb
ba1ash: I do not see how this line helps. If storage is absent in the database then it we never reach it. If storage is present then we do need it. So, I added an explicit `discard_on ActiveJob::DeserializationError`.

modules/storages/app/workers/storages/health_status_mailer_job.rb
ba1ash: When we use ActiveRecord instance as a parameter then https://github.com/rails/globalid is used to serialize this parameter. Then in the database there is something like: `gid://app/Storage/1`. Then when the parameter is deserialized it has been fetched from the database and modal instance has been created with actual data. So, because of that behavior we do need to reload at the beginning of `perform` method.

ba1ash (approved):

PR #15191, opened 2024-04-08
[#51793] Split existing project administration settings into multiple pages

dombesz:
https://community.openproject.org/work_packages/51793


jjabari-op (changes requested):
Hi @dombesz! Thank you for taking care of this!

I found two things:

- The spacing between the heading (e.g. "New Project") and the content below seems to be too big.
- If I see it correctly, we don't have specs for the project list settings page?

dombesz:
Thanks Jonas,
1. The content adjustment is done.
2. Yes, there are no mockups for this change.
image
image

PR #15197, opened 2024-04-09
Adds an explanation message under the Permission Header

mereghost:
Related Work Package: [OP#52571](https://community.openproject.org/projects/openproject/work_packages/52571)

Very naive solution for the requirements.

Directly from /admin/roles/5/edit

image


wielinde:
I like that approach. Just find my tiny comment above and it is ready to merge.

apfohl (approved):

mereghost:
@wielinde wrote:
> I like that approach. Just find my tiny comment above and it is ready to merge.

404 tiny comment not found. xD

PR #15199, opened 2024-04-09
Fix crowdin workflow

toy:
It wasn't working because `BRANCH` environment variable was not set


cbliard (approved):
That was it! `BRANCH` was simply not defined. That's what I suspected at some point, but as it was used before and worked, I discarded this possibility (a bit too quickly). I could have know if I had inspected the output as I added some `echo` to debug that.

The `if` reads better in multiline 👍 

Thanks for fixing it!

PR #15200, opened 2024-04-09
ugly hack to make member popup menu visible in firefox

toy:
Firefox doesn't yet support popover attribute by default, so popup menu is clipped by overflow

Followup for #14721

### DIFF
--- a/app/components/members/table_component.rb
+++ b/app/components/members/table_component.rb
@@ -101,6 +101,10 @@ module Members
       { caption: }
     end
 
+    def container_class
+      "generic-table--container_visible-overflow"
+    end
+
     ##
     # Adjusts the order so that users are joined to support
     # sorting by their attributes
--- a/app/components/table_component.html.erb
+++ b/app/components/table_component.html.erb
@@ -27,7 +27,7 @@ See COPYRIGHT and LICENSE files for more details.
 
 ++#%>
 
-<div class="generic-table--container" data-test-selector="<%= test_selector %>">
+<div class="generic-table--container <%= container_class %>" data-test-selector="<%= test_selector %>">
   <div class="generic-table--results-container">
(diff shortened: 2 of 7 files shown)


HDinger (commented):
Hi @toy 

what do I have to do to reproduce the original issue here?

toy:
@HDinger Go to members page of a project and have there only one member displayed

HDinger (changes requested):
We have a class `generic-table--container_visible-overflow` which can be set on the `generic-table--container` to achieve this. Could you please also change that on the project list as well?

toy:
@HDinger There is a problem with `. generic-table--container_visible-overflow` also changing overflow for `.generic-table--results-container` from `auto` to `visible`, so it becomes not possible to scroll the table horizontally if the window is not wide enough
That was added in https://github.com/opf/openproject/commit/eacbc7528d26c78c6d97573126cc3c1184a04cca, so part of https://github.com/opf/openproject/pull/11970 which is about standardising date pickers
I also can't understand why is `overflow: visible` needed for project lists, there is a different menu and it is force shown using a `padding-bottom: $spot-spacing-10`

HDinger (approved):
Hi @toy 

I changed the `.generic-table--container_visible-overflow` to only set `overflow: visible` ( as the name indicates). The other rules for the cost_types were obsolete by now as the datepicker is now rendered as a direct child of the body anyway. So I removed the class from there as well. 
With regards to the project list: I agree, that the hack is not needed there at all due to the big padding below. I removed that one as well. The big padding looks super weird in my opinion, but I will talk about this with the product team and we will tackle that separately. 
Last but not least, I rebase to the release/14.0 branch and changed the target. Once the tests are green, this is good to merge for me.

PR #15208, opened 2024-04-10
[#54032] added adapter for user related auth strategy picking

Kharonus:
[#54032](https://community.openproject.org/wp/54032)

### WAT?

- added two lambdas in storage adapters that serves authentication strategies in the context of user-bound and user-less authentication
- this way we can instantiate strategies independent of the storage context (IoC)


modules/storages/app/common/storages/peripherals/one_drive_registry.rb
```diff
@@ -58,6 +58,11 @@ module Peripherals
       namespace("models") do
         register(:managed_folder_identifier, ManagedFolderIdentifier::OneDrive)
       end
+
+      namespace("authentication") do
+        register(:userless, StorageInteraction::AuthenticationStrategies::OneDriveStrategies::UserLess)
```
Kharonus, line 63: we still have the chance now to name this `user_less`. At least my spell checker in the IDE marks that as wrong xD

mereghost: WDYT of `application` and `user`.

modules/storages/app/common/storages/peripherals/storage_interaction/authentication_strategies/nextcloud_strategies.rb
```diff
@@ -0,0 +1,49 @@
...
+#
+# See COPYRIGHT and LICENSE files for more details.
+#++
+
+module Storages
+  module Peripherals
+    module StorageInteraction
+      module AuthenticationStrategies
+        module NextcloudStrategies
+          UserLess = -> do
```
mereghost, line 36: IIRC Lambdas will check for arity. You could add an `|_|` there so that it won't explode if a user is ever passed.

modules/storages/app/common/storages/peripherals/one_drive_registry.rb
Kharonus: I do not have a strong take on that. I was kind of hyped by the words `user-less` and `user-bound`, but there is no deeper reasoning behind.

I'd be fine with `application` and `user`, too.

mereghost: We can keep it as is as renaming it later on is easy. =)

apfohl (approved):

modules/storages/app/common/storages/peripherals/storage_interaction/authentication_strategies/nextcloud_strategies.rb
Kharonus: After rethinking I'd argue, that we should add that. Instead, it should throw when this is called with a method param. It is in the end a programmatical error and should get fixed.

PR #15212, opened 2024-04-10
optimize images in docs using `image_optim -R docs`

toy:
Applying my creation starting with docs
Reduce size by 7.48% (9.5M)


as-op:
Hi @toy, thank you for the PR.

I regularly already optimize PNGs with `oxipng -o 6 -i 0 --strip safe` (and GIFs with gifsicle, and so on). 
Problem is, if these tools are run again, they optimize each run, so I created a wrapper script that implements a minimum of of size reduction of 1KB and a percentage threshold, e.g min. 4% for PNG. There is a loss in many of those runs, especially with GIFs.

So I rather would not merge your PR and opt for closing it. 

PS: There are so many files in this PR, Githubs PR Files Tab is crashing for me 😀

toy:
`image_optim` has ability to cache, so no waisted time if cache can be shared somehow, alternative is to optimise only newly added images

as-op:
Good idea, so far I've done optimizations locally, but they probably belong in a Github Action, where shared caching could be implemented.
Let's have a chat sometime to create one.

PR #15214, opened 2024-04-10
fix member role creation conflict with user invitation rate limiting

toy:
invite_via_members? in openproject-multitenancy plugin is expecting `user_ids` to be an array https://appsignal.com/openproject-gmbh/sites/62a6d833d2a5e482c1ef825d/exceptions/incidents/888


cbliard (approved):
All good 👍 
I just changed it a little to use `f.field_name(:user_ids, multiple: true)` instead of `f.field_name(:user_ids) + "[]"`.

toy:
Thanks, that is where `multiple: true` can be specified, I was looking for it in `hidden_field_tag`

PR #15221, opened 2024-04-11
[#54016] Undefined method status for HTTPX::ErrorResponse

ba1ash:
https://community.openproject.org/work_packages/54016

### DIFF
--- a/modules/storages/spec/common/storages/peripherals/storage_interaction/one_drive/set_permissions_command_spec.rb
+++ b/modules/storages/spec/common/storages/peripherals/storage_interaction/one_drive/set_permissions_command_spec.rb
@@ -143,17 +143,14 @@ RSpec.describe Storages::Peripherals::StorageInteraction::OneDrive::SetPermissio
 
     context "when there is a timeout" do
       it "logs a warnig and does not raise NoMethodError", vcr: "one_drive/set_permissions_delete_permission_read" do
+        stub_request_with_timeout(:post, /invite$/)
         allow(OpenProject.logger).to receive(:warn)
-        request_mock = HTTPX::Request.new("POST", "http://example.com/")
-        error_response_mock = HTTPX::ErrorResponse.new(request_mock, HTTPX::ConnectTimeoutError.new(60, "timed out while waiting on select"), {})
-        allow_any_instance_of(HTTPX::Session).to receive(:post).with(any_args).and_call_original
-        allow_any_instance_of(HTTPX::Session).to receive(:post).with(/\/invite$/, anything).and_return(error_response_mock)
 
         permissions_command.call(path:, permissions: { read: ["d6e00f6d-1ae7-43e6-b0af-15d99a56d4ce"] })
 
         expect(OpenProject.logger)
           .to have_received(:warn)
-                .with(command: Storages::Peripherals::StorageInteraction::OneDrive::SetPermissionsCommand,
+                .with(command: described_class,
                       message: nil,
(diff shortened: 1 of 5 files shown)


modules/storages/spec/common/storages/peripherals/storage_interaction/one_drive/set_permissions_command_spec.rb
```diff
@@ -140,6 +140,25 @@
...
+        permissions_command.call(path:, permissions: { read: ["d6e00f6d-1ae7-43e6-b0af-15d99a56d4ce"] })
+
+        expect(OpenProject.logger)
+          .to have_received(:warn)
+                .with(command: Storages::Peripherals::StorageInteraction::OneDrive::SetPermissionsCommand,
+                      message: nil,
+                      data: { body: "/usr/local/bundle/gems/httpx-1.2.3/lib/httpx/response.rb:260:in `full_message': timed out while waiting on select (HTTPX::ConnectTimeoutError)\n",
+                              status: nil }).once
+      end
+    end
```
Kharonus, line 158: 🟡 I did the same once, using mock objects, so that this test can live in the command's unit test (because you cannot use `webmock.to_timeout` in an RSpec block where vcr is enabled).

An alternative could be to add this command specific timeout test to `modules/storages/spec/common/storages/peripherals/storage_interaction/network_error_spec.rb`. But we will have the question, if we need to test timeout handling on each and every query/command.

modules/storages/spec/services/storages/one_drive_managed_folder_sync_service_spec.rb
```diff
@@ -256,6 +255,21 @@
                         message: nil,
                         data: { status: 400, body: /drive id/ })
         end
+
+        it "does not break in case of timeout", vcr: "one_drive/sync_service_root_read_failure" do
```
Kharonus, line 260: ❓ shouldn't you remove the `vcr` attribute here? No cassette is recorded, is it?

Kharonus (changes requested):
I have two minor issues. Let's talk about those.

modules/storages/spec/services/storages/one_drive_managed_folder_sync_service_spec.rb
ba1ash: There is a request to get token prior to the problematic one. So, I reused existing vcr cassette because of that.

modules/storages/spec/common/storages/peripherals/storage_interaction/one_drive/set_permissions_command_spec.rb
ba1ash: I would not create a specific file for network erorrs and put this tests in command/request specific test files.

modules/storages/spec/services/storages/one_drive_managed_folder_sync_service_spec.rb
Kharonus: That makes sense. 👍🏽

modules/storages/spec/common/storages/peripherals/storage_interaction/one_drive/set_permissions_command_spec.rb
Kharonus: Using another file (or another block) allows you to use `webmock.to_timeout`, which makes the whole test setup be a one liner.

Maybe there is a way, to have webmock working with VCR in the same file? Can you check that?

ba1ash: But then we have to make prior requests stubbing with webmock as well and it is done here with VCR.
I've checked and have not found a way to make webmock and vcr working together.

Kharonus (approved):
I love it, thanks.

PR #15225, opened 2024-04-11
Documentation gitlab add local network settings

tiroessler:
Add a hint in the Documentation for users who might run GitLab in a local network.

add missing backslash in example docker command for ruining clamav


dombesz (approved):
Thanks @tiroessler ! I think the target branch should be set to 14.0. I rebased the branch against 14.0 and removed commits related to dev.

PR #15226, opened 2024-04-11
[#53799] Overview page crashes if there have been project custom fiel…

dombesz:
Addresses both tickets: https://community.openproject.org/wp/53728 and https://community.openproject.org/wp/53799 .
## The problem

Backfilling the data in the migration `20231123111357_create_custom_field_sections.rb` is unreliable due to using `acts_as_list` callbacks during migrations. Sometimes it does not fill the `custom_field_section_id`.

## The solution
A better approach is to use plain SQL for data backfilling, avoiding eventual problems of the newly migrated field not being backfilled. This approach should also be a lot faster, because we are executing a single update query instead of loading all the models into memory, then updating them one by one.

**Additionally:** The `position_in_custom_field_section` is also being filled in order to have an initial value for the `acts_as_list` plugin. The calculation of the current row's position happens in an SQL subquery as the [acts_as_list](https://github.com/brendon/acts_as_list?tab=readme-ov-file#adding-acts_as_list-to-an-existing-model) gem recommends.

## The expectation 

The expectation is to have all the `ProjectCustomField#custom_field_section_id` set to the newly created `ProjectCustomFieldSection` and to have an incremental position set in the `position_in_custom_field_section` field **only** for the `custom_fields` with the `type='ProjectCustomField'`.

### DIFF
--- a/db/migrate/20231123111357_create_custom_field_sections.rb
+++ b/db/migrate/20231123111357_create_custom_field_sections.rb
@@ -28,9 +28,19 @@ class CreateCustomFieldSections < ActiveRecord::Migration[7.0]
       name: "Project attributes"
     )
 
-    # trigger acts_as_list callbacks via updating each record instead of bulk update
-    ProjectCustomField.find_each do |project_custom_field|
-      project_custom_field.update!(custom_field_section_id: section.id)
-    end
+    ActiveRecord::Base.connection.execute <<~SQL.squish
+       UPDATE "custom_fields"
+       SET
+         "position_in_custom_field_section" = "mapping"."new_position",
+         "custom_field_section_id" = #{section.id}
+       FROM (
+         SELECT
+           id,
+           ROW_NUMBER() OVER (ORDER BY updated_at) AS new_position
+         FROM "custom_fields"
+         WHERE "custom_fields"."type" = 'ProjectCustomField'
+       ) AS "mapping"
+       WHERE "custom_fields"."id" = "mapping"."id";
+     SQL
   end
 end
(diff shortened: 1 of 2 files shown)


db/migrate/20231123111357_create_custom_field_sections.rb
```diff
@@ -28,9 +28,19 @@ def create_and_assign_default_section
       name: "Project attributes"
     )
 
-    # trigger acts_as_list callbacks via updating each record instead of bulk update
-    ProjectCustomField.find_each do |project_custom_field|
-      project_custom_field.update!(custom_field_section_id: section.id)
-    end
+    ActiveRecord::Base.connection.execute <<~SQL.squish
```
machisuji: We have the `execute_sql` helper which allows for passing sanitized parameters. I don't think there is anything to worry about here in particular, but it's still a good habit to just use this precisely so that we don't have to think about potential issues with this. Besides this is otherwise surely going to turn up in some security analysis as a potential risk as well anyway and this can be avoided this way too.

So please use that instead like seen in the example in the source:

```
execute_sql "select id from users where mail = :email", email: params[:email]
```

db/migrate/20231123111357_create_custom_field_sections.rb
```diff
@@ -28,9 +28,19 @@ def create_and_assign_default_section
       name: "Project attributes"
```
machisuji: I think it would be a good idea to get rid off the last piece of model code in this migration too and create the section via plain SQL too. This is probably a super simple query and as opposed to this model code it will keep working even if the model should disappear one day.

I know this is outside of the actual scope of the PR but we might as well do it while we're at it.

machisuji (changes requested):
Thanks for the fix and improvement to the migration, @dombesz! Looks good to me in principle. Still, I have 3 suggestions.
Two are annotated directly. The last one would be to **add a test for this migration**.

dombesz:
Thank you @machisuji for the feedback, I made the changes you have requested.

machisuji (approved):
👍

PR #15227, opened 2024-04-12
[#54131] Render help text for missing values only on work-based mode

aaron-contreras:
See: https://community.openproject.org/work_packages/54131


HDinger (changes requested):
Lgtm 👍  There is a test still reyling on the existance of the banner. Once that is fixed, feel free to merge

PR #15232, opened 2024-04-12
bug/storage-health-status-mails-are-send-for-every-failed-request

apfohl:
Fixed concurrency key for HealthStatusMailerJob by adding `.id` to the storage. Before the storage object was stringifyed and lead to a different key every time. Which in turn caused GoodJob to enqueue a job for every failed request. Now the storage ID is used and the key stays the same.


Kharonus (approved):

akabiru:
⚠️ @apfohl should this be based on `release/14.0`?

PR #15233, opened 2024-04-12
[#53993] Use authentication in delete_folder_command

Kharonus:
[#53993](https://community.openproject.org/work_packages/53993)

### WAT?

- added caching for oauth_client_credentials
- reworked delete folder command
- adapted unit tests


modules/storages/app/common/storages/peripherals/storage_interaction/authentication_strategies/failures.rb
```diff
@@ -37,6 +37,20 @@ module Failures
...
+            payload =
+              case response
+              in { content_type: { mime_type: "application/json" } }
+                response.json
+              in { content_type: { mime_type: "text/xml" } }
+                response.xml
+              else
+                response.body.to_s
+              end
+
```
mereghost, line 51: 🟢 Since this just extract the payload from the response, I wonder if naming it to `ErrorDataPayload` or `PayloadExtractor` would make its intent clearer.


I'm not dying on this hill tho.

modules/storages/app/common/storages/peripherals/storage_interaction/authentication_strategies/oauth_client_credentials.rb
```diff
@@ -37,41 +37,57 @@ def self.strategy
...
             return build_failure(storage) unless config.valid?
 
-            create_http_and_yield(issuer: config.issuer,
-                                  client_id: config.client_id,
-                                  client_secret: config.client_secret,
-                                  scope: config.scope,
-                                  http_options:,
-                                  &)
-          end
+            access_token = Rails.cache.read("storage.#{storage.id}.access_token")
```
mereghost, line 45: 🟡 Could this be replaced by `cache.fetch`?

```ruby
access_token = Rails.cache.fetch('stinking.key') do
  handle_missing_token_and_return_it
end
```

modules/storages/app/common/storages/peripherals/storage_interaction/nextcloud/internal/delete_entity_command.rb
```diff
@@ -28,37 +28,76 @@
...
+    end
+
+    private
+
+    def origin_user_id(auth_strategy)
+      case auth_strategy.key
+      when :basic_auth
+        ServiceResult.success(result: @storage.username)
+      when :oauth_user_token
+        origin_user_id = OAuthClientToken.find_by(user_id: user, oauth_client_id: @storage.oauth_client.id)&.origin_user_id
```
mereghost:
```suggestion
        origin_user_id = OAuthClientToken.find_by(user_id: user, oauth_client: @storage.oauth_client)&.origin_user_id
```

modules/storages/app/common/storages/peripherals/storage_interaction/nextcloud/internal/delete_entity_command.rb
```diff
@@ -28,37 +28,76 @@
...
+        origin_user_id.present? ?
+          ServiceResult.success(result: origin_user_id) :
+          failure(code: :error,
+                  payload: nil,
+                  log_message: "No origin user ID or user token found. Cannot execute query without user context.")
```
mereghost: 🟡 As this is a multi-line statement, I think it would be clearer as normal if, rather than a ternary.

mereghost (approved):
Seems solid. Just a bunch of questions that shouldn't block the merge at al. =)

modules/storages/app/common/storages/peripherals/storage_interaction/authentication_strategies/failures.rb
Kharonus: I struggled with the naming, too. But it does do return the whole error data object, after extracting the payload from resource.

modules/storages/app/common/storages/peripherals/storage_interaction/authentication_strategies/oauth_client_credentials.rb
Kharonus: no, as the I cannot figure any block that returns only the token.

with httpx you create an http session, and while doing so, you get an access token attached to the session. Which is why I extract the token later to write the cache.

modules/storages/app/common/storages/peripherals/storage_interaction/nextcloud/internal/delete_entity_command.rb
Kharonus: sad eric is sad :D he really likes ternaries, because they have less lines ;) 

But I get the motivation, and I understand not everyone is a ternary junkie

PR #15236, opened 2024-04-12
Fix/transaction for update progress job

ulferts:
Wraps both jobs changing progress & work first and then creating journals for it in a transaction to avoid inconsistent states.

### DIFF
--- a/spec/migrations/update_progress_calculation_spec.rb
+++ b/spec/migrations/update_progress_calculation_spec.rb
@@ -692,17 +692,24 @@ RSpec.describe UpdateProgressCalculation, type: :model do
   end
 
   describe "error during job execution" do
+    let_work_packages(<<~TABLE)
+      subject     | work | remaining work | % complete | ∑ work | ∑ remaining work | ∑ % complete
+      wp working  |      |             4h |        60% |    10h |               4h |          60%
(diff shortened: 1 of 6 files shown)


spec/migrations/update_progress_calculation_spec.rb
```diff
@@ -686,4 +690,46 @@ def expect_migrates(from:, to:)
       end
     end
   end
+
+  describe "error during job execution" do
```
cbliard, line 694: To match the flow of reading, I would put things in a more natural order:
1. `let_work_packages` first
2. then the before setting behavior
3. then set behavior of "wp working" first as it is the first created in the `let_work_packages` table
4. then set behavior of "wp breaking" second as it is the second created.

It would feel easier and more natural to me.

The right code range is https://github.com/opf/openproject/pull/15236/files#diff-dfee84290130c56c5027b2721fae4092d20e4c9fe44ced22c9783f89e6ef359dR695-R720

cbliard: BTW, the other one in `spec/workers/work_packages/apply_statuses_p_complete_job_spec.rb` is already ok.

cbliard (approved):
All good!
Just a minor remark about the way a test is organized. Feel free to ignore.

cbliard (approved):

PR #15244, opened 2024-04-15
Unset old total p complete values in journals

ulferts:
Unsets derived_done_ratio values in both the work_package as well as the work_package_journals briefly to null before the newly calculated values are inserted into the work package table. This also has the effect that now, a journal is only created for cases where a derived_done_ratio value is set. Before, it was also created in case the value was changed from 0 to null.

----

https://community.openproject.org/wp/54141

### DIFF
--- a/app/workers/work_packages/update_progress_job.rb
+++ b/app/workers/work_packages/update_progress_job.rb
@@ -53,6 +53,8 @@ class WorkPackages::UpdateProgressJob < ApplicationJob
 
       update_totals
 
+      unset_total_p_complete
+
       copy_progress_values_to_work_packages_and_update_journals
     end
   end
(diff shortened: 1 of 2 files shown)


ulferts:
@cbliard I still need to run this changed migration against a larger database to see the effect but wanted to give you the chance to already look over the changes.

So please refuse merging until I have the results available which I will trigger in the evening.

ulferts:
The results came back quicker than anticipated. Instead of taking multiple hours, the migration job now returned after 6 minutes. I cannot proved this to be a representable dataset but I would imagine that most of the time values are empty so I would expect it to to be representable.

cbliard (approved):
Nice! That's a great performance improvement.

PR #15248, opened 2024-04-15
[#54061] Rails.application.routes don't respects  OPENPROJECT_RAILS__RELATIVE__URL__ROOT

ba1ash:
https://community.openproject.org/work_packages/54061

- Solution
    Move the config up in application.rb to be called early enough.
    The config option has been renamed due to its more genela nature than being in`config.action_controller.*` scope.

- How to test?
    1. Set OPENPROJECT_RAILS__RELATIVE__URL__ROOT=/something
    2. Open up rails console before/after changes and call `Rails.application.routes.url_helpers.root_path`.
    3. Prefix should not be there before and should be there after.

Check discussion in https://github.com/rails/rails/issues/42243 for more details if you'd like.

### DIFF
--- a/config/application.rb
+++ b/config/application.rb
@@ -124,6 +124,14 @@ module OpenProject
     # That directory contains configurations and patches to rails core functionality.
     config.autoload_once_paths << Rails.root.join('lib_static').to_s
 
+    # Configure the relative url root to be whatever the configuration is set to.
+    # This allows for setting the root either via config file or via environment variable.
+    # It must be called early enough. In our case in should be called earlier
+    # than `config.exceptions_app = routes`. Otherwise Rails.application.routes.url_helpers
+    # will not have configured prefix.
+    # Read https://github.com/rails/rails/issues/42243 for some details.
+    config.relative_url_root = OpenProject::Configuration['rails_relative_url_root']
+
     # Use our own error rendering for prettier error pages
     config.exceptions_app = routes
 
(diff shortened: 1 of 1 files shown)


akabiru (commented):
🚧 Should we then remove all references to `OpenProject::StaticRouting` and use `Rails.application.routes.url_helpers` to complete this change?

ba1ash:
> 🚧 Should we then remove all references to `OpenProject::StaticRouting` and use `Rails.application.routes.url_helpers` to complete this change?

I am not sure that we should do it now. There are quite a number of usages. By accepting this PR we will prevent further bugs if someone decides to use `Rails.application.routes.url_helpers`. And as a follow up we in future we should consider getting rid of `StaticRouting`.
I suggest to create a follow up CODE MAINTENANCE and plan for near future. WDYT @akabiru ?

mereghost:
@ba1ash & @akabiru would be worth to bring this to the Dev Weekly this Tuesday (16.04)?

akabiru:
> @ba1ash & @akabiru would be worth to bring this to the Dev Weekly this Tuesday (16.04)?

Sure @mereghost , we already started a discussion in [#devs](https://matrix.to/#/!xcCvdAzoRtrTNmmFlD:openproject.org/$syYcbQ4cIva4ggzByjs8a5P1e51ThW5JVECRh0NKQEU?via=openproject.org)

akabiru (approved):
:shipit:

PR #15249, opened 2024-04-15
Allow shared users to be set as accountable/responsible as well

klaustopher:
The `responsible` field still used the project specific autocompleter. This was moved to the work package when the package was persisted. I found that the `available_responsibles` API endpoint returns the same data as the `available_assignees` endpoint. Since it is not used in any places, I have also removed the endpoint

Fixes https://community.openproject.org/projects/openproject/work_packages/51551


docs/release-notes/14-0-0/README.md
```diff
@@ -71,6 +71,10 @@ Since the done_ratio is now a read only value, derived from work and remaining w
 
 For more information, see [#40749](https://community.openproject.org/work_packages/40749)
 
+### Removed `available_responsibles` from the API
+
+The `available_responsibles` endpoint has been removed from the API. This endpoint was used to retrieve a list of users that could be set as the **responsbile** for a work package. This information has been identical to the results by the  `available_assignees` endpoint. When you are using the `available_responsibles` endpoint in your application, you should switch to using the `available_assignees` endpoint instead.
```
ulferts: Typo responsbile -> responsible

lib/api/v3/projects/projects_api.rb
```diff
@@ -67,7 +67,6 @@ class ProjectsAPI < ::API::OpenProjectAPI
             mount ::API::V3::Projects::UpdateFormAPI
 
             mount API::V3::Projects::AvailableAssigneesAPI
-            mount API::V3::Projects::AvailableResponsiblesAPI
```
klaustopher, line 70: Whooops, sorry for overlooking and thanks for fixing. I did not restart my server after removing the class. I wonder why the tests did not fail when autoloading? 🤔

ulferts: Yeah, sorry for stepping over the reviewer boundaries here but it was so straight forward that it seemed weird to note it down instead.

What was weird is that the CI didn't run on your commit [d15d38c](https://github.com/opf/openproject/pull/15249/commits/d15d38cc261a78bdbad2b6bb2daf3c9470bf1e14) before otherwise it would have informed you already.

ulferts (approved):

PR #15252, opened 2024-04-15
[40147] Snackbar/toast informing users of new notifications has the wrong icon

bsatarnejad:
https://community.openproject.org/work_packages/40147/activity

Add an arbitrary icon to the toast component, and make the icon in info toast blue, as discussed with Parimal.

### DIFF
--- a/frontend/src/global_styles/content/_toast.sass
+++ b/frontend/src/global_styles/content/_toast.sass
@@ -46,7 +46,7 @@ $nm-color-warning-border: #ef9e56
 $nm-color-warning-icon: #ef9e56
 $nm-color-warning-background: #ffe6c6
 
-$nm-color-info-icon: #333
+$nm-color-info-icon: #155282
 
 $nm-color-border: #8fc0db
 $nm-color-background: #e3f5ff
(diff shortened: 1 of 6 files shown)


frontend/src/global_styles/content/_toast.sass
```diff
@@ -145,31 +144,38 @@ $nm-upload-box-padding: rem-calc(15) rem-calc(25)
...
-    &::before
+    &:not(.toast-icon)::before
       @include messages-icon
       @extend %nm-icon-warning
 
   &.-info
     @extend %info-placeholder
 
     &::before
+      color: $nm-color-info-icon
```
HDinger, line 176: Similar to how you extracted the color here, you have to do the same thing for the other types (error, success, warning) as well. Otherwise, the icon will be black there:

Bildschirmfoto 2024-04-15 um 15 09 18

HDinger (changes requested):
hi @bsatarnejad 

The defined case works fine, I only have a small remark. 👍

HDinger (approved):

PR #15255, opened 2024-04-15
Attempt to use arm64 runners for building docker images

crohr:


crohr:
Need to fix conflict, but at least the builds are fine. I tested the resulting all-in-one variant on amd64 and it works fine. Would be nice to test the slim variant as well.

PR #15259, opened 2024-04-15
[54179] Total Remaining work of parent should be updated when children are closed in status-based mode

cbliard:
https://community.openproject.org/wp/54179

With this change, the total values are not discarded anymore when the total is 0h. Before this change, having total remaining work == 0h would unset it, which leads to not having total % complete either.

This is now fixed.

As a side-effect, the total % complete is now set for leaf work packages, which is consistent with how the update progress job migration behaves.


aaron-contreras (approved):
👍

PR #15270, opened 2024-04-16
[54193] Inconsistent UX with commas for time unit when inputting remaining work or work

bsatarnejad:
Convert remaining hours string value to the right hour format,
for example: 
passed value: 1h30 -> converted value: 1.5 
passed value: 1,50 -> converted value: 1.5 

https://community.openproject.org/projects/openproject/work_packages/54193/activity


cbliard (approved):
👍

PR #15271, opened 2024-04-16
[#54296] PDF export gantt chart

as-op:
Use prawn to export Gantt chart in background worker

*   [x] Subject of work package in first column
*   [x] &quot;Zoom&quot;
    *   [x] Day, month or quarter based view
    *   [x] Adjustable column sizes
*   [x] Split chart on pages depending on PDF &quot;paper&quot; size
*   [x] Vector based shapes in PDF
*   [x] Work package type defines bar color
*   [x] Work package type &quot;milestone&quot; is a diamond shape
*   [x] Chart header repeats on every page
*   [x] work package sorting
*   [x] Dependency lines between work packages (follows-precedes relationships)
*   [x] work package hierarchies with index numbers
*   [x] work package groupings
*   [x] Refactor humongous app/models/work_package/pdf_export/gantt.rb
*   [x] Remove DEBUG UI from Export modal
*   [x] Add UI Options described in https://community.openproject.org/projects/openproject/work_packages/54297

https://community.openproject.org/work_packages/54296


frontend/src/app/shared/components/modals/export-modal/wp-table-export.modal.html
```diff
@@ -25,6 +25,30 @@
...
+            <option value="11">Zoom: 11 (Quarter, 24px)</option>
+          </select>
+          <select class="debug-export-values" name="paper_size">
+            <option selected value="EXECUTIVE">Page size: EXECUTIVE</option>
+            <option value="A4">Page size: A4</option>
+            <option value="A3">Page size: A3</option>
+            <option value="A2">Page size: A2</option>
+            <option value="A1">Page size: A1</option>
+          </select>
+      </div>
```
toy: Second closing div tag is missing

app/models/work_package/pdf_export/gantt.rb
```diff
@@ -0,0 +1,1146 @@
...
+  class GantDataLineFragment
+    attr_accessor :left, :right, :top, :bottom
+
+    def initialize(left, right, top, bottom)
+      @left = left
+      @right = right
+      @top = top
+      @bottom = bottom
+    end
+  end
```
toy: Unused class

app/models/work_package/pdf_export/gantt.rb
```diff
@@ -0,0 +1,1146 @@
...
+      @right = left + width
+      @top = top
+      @bottom = top + height
+      @width = width
+      @height = height
+      @work_package = work_package
+      @columns = columns
+      @color = color
+    end
+  end
```
toy: Is immutable, so can be replaced by Data class:
```suggestion
  GanttDataShape = Data.define(:type, :left, :width, :top, :height, :work_package, :columns, :color) do
    def right = left + width
    def bottom = top + height
  end
```

app/models/work_package/pdf_export/gantt.rb
```diff
@@ -0,0 +1,1146 @@
...
+      @width = width
+      @left = left
+      @right = left + width
+      @padding_h = padding_h
+      @padding_v = padding_v
+      @top = top
+      @height = height
+      @bottom = top + height
+    end
+  end
```
toy: Immutable classes can be created in fewer lines using Data class, those that needs mutation using Struct class. Both also support keyword arguments if needed and can be combined with value omission, which will be better when there are a lot of values.

```ruby
Point = Data.define(:x, :y)
x = 1
y = 2
Point.new(x:, y:)
```

app/models/work_package/pdf_export/gantt.rb
```diff
@@ -0,0 +1,1146 @@
...
+      width_per_day = @column_width.to_f / test_date.day
+      day_in_month = end_date.day
+      @column_width - (day_in_month * width_per_day)
+    end
+
+    def wp_on_month?(work_package, date)
+      start_date = work_package.start_date || work_package.due_date
+      end_date = work_package.due_date || Time.zone.today
+      Range.new(Date.new(start_date.year, start_date.month, 1), Date.new(end_date.year, end_date.month, -1))
+           .include?(date)
```
toy: Range can be created using literal `..`, also unless there is a reason not to, it is better to use quicker `cover?` method which compares value with range start and end instead of `include?` which for such case will build the list of values in range and check if the checked value is among them:
```suggestion
      (Date.new(start_date.year, start_date.month, 1)..Date.new(end_date.year, end_date.month, -1))
           .cover?(date)
```

app/models/work_package/pdf_export/gantt.rb
```diff
@@ -0,0 +1,1146 @@
...
+#   - GanttBuilderQuarters: This class is a subclass of GanttBuilder and is responsible for building the Gantt chart with quarters as the zoom level.
+# - GanttPainter: This class is responsible for painting the Gantt chart. It paints the grid, header row, lines, bars, milestones, and rows.
+# - Data classes: These classes are used to store the data that is used to build the Gantt chart.
+#
+# 1. Build the data classes, do the layout, measuring, etc.
+# 2. Paint the Gantt chart into the PDF
+
+module WorkPackage::PDFExport::Gantt
+  def write_work_packages_gantt!(work_packages, _)
+    wps = work_packages.reject { |work_package| work_package.start_date.nil? && work_package.due_date.nil? }
```
toy:
```suggestion
    wps = work_packages.select { |work_package| work_package.start_date || work_package.due_date }
```

toy (commented):
Just few suggestions, didn't look through all the code

frontend/src/app/shared/components/modals/export-modal/wp-table-export.modal.html
as-op: Thanks!

app/models/work_package/pdf_export/gantt.rb
as-op: Thanks a lot! Converted all to Data or Struct. Much much nicer.

app/models/work_package/pdf_export/gantt.rb
as-op: Thanks for the insight!

oliverguenther (changes requested):
I like the sublevel we enter when clicking Gantt chart. 👍 It seems to be lacking a feature test for it however, which I'd like to see given the custom behavior added here.

I couldn't find any issues with the actual PDF export of the Gantt view, that appears to be behaving correctly.

as-op:
Hi @oliverguenther, thanks for review. You caught me :D. As there weren't any tests I broke with my UI changes and the deprecated usage of Angular, I thought there weren't any at all and skipped frontend tests. 
(Backend gantt PDF export itself is fully spec'd though).

I will add the UI tests in a separate PR, so the export can already be QA'd.

PR #15272, opened 2024-04-16
[54290] Superflous drop-down arrow on %Complete input modal

bsatarnejad:
https://community.openproject.org/projects/openproject/work_packages/54290/activity

### DIFF
--- a/frontend/src/global_styles/openproject/_forms.sass
+++ b/frontend/src/global_styles/openproject/_forms.sass
@@ -157,7 +157,8 @@ select
   border-radius: 0
   border: $input-border
 
-  @if $select-arrow
-    background: transparent url(image-triangle($select-arrow-color)) right 10px center no-repeat
-    background-size: 8px 8px
-    padding-right: rem-calc(18px) + $form-padding
+  select:not(.FormControl-select)
+    @if $select-arrow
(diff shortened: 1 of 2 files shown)


cbliard (commented):
I'm not sure which down arrow is preferred. There is the one from Primer which is double-sided, and the one from us, which is a drop-down arrow.

I would opt for keeping the Primer one, but honestly I do not know. I asked Henriette in the work package comments.

If wanting to enable the Primer one, this snippet in `frontend/src/global_styles/openproject/_forms.sass` works:

on line 160, replace this:
```sass
select
  // ...

  @if $select-arrow
    background: transparent url(image-triangle($select-arrow-color)) right 10px center no-repeat
    background-size: 8px 8px
    padding-right: rem-calc(18px) + $form-padding
```

with this:
```sass
select
  // ...

select:not(.FormControl-select)
  @if $select-arrow
    background: transparent url(image-triangle($select-arrow-color)) right 10px center no-repeat
    background-size: 8px 8px
    padding-right: rem-calc(18px) + $form-padding
```

HDinger (changes requested):
I agree with @cbliard . We definetly want to keep the native arrows of the Primer select fields. So instead oh overwriting those, please hide our cursom arrow on the primer select fields as sugessted above.

HDinger (approved):

PR #15278, opened 2024-04-16
[#54293] Reenable HTTPX persistent connections.

ba1ash:
https://community.openproject.org/work_packages/54293

### DIFF
--- a/lib/open_project.rb
+++ b/lib/open_project.rb
@@ -47,20 +47,23 @@ module OpenProject
   end
 
   def self.httpx
-    session = HTTPX
-                .plugin(:oauth)
-                .plugin(:basic_auth)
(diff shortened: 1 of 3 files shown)


machisuji:
@ba1ash We had the issue before that requests didn't time out that should've timed out. You have a reproduction script in the work package description. Is it possible to add a spec for this that doesn't take too long?

machisuji (approved):
👍

machisuji:
> @ba1ash We had the issue before that requests didn't time out that should've timed out. You have a reproduction script in the work package description. Is it possible to add a spec for this that doesn't take too long?

Discussed this with Pavel. Will test it on edge once deployed instead.

PR #15279, opened 2024-04-16
Applied retry_on for ManageNextcloudIntegrationJob

apfohl:
https://community.openproject.org/projects/openproject/work_packages/54046/activity

## WAT?

To have the `ManageNextcloudINtegrationJob` retry failing outbound requests, we introduced raising a custom error (`::Storages::Errors::IntegrationJobError`) in case of a failure response. This error is then caught by the `retry_on` configuration of the job. The health status of the storage is only updated after the 5th atempt failed.

### DIFF
--- a/modules/storages/app/common/storages/errors.rb
+++ b/modules/storages/app/common/storages/errors.rb
@@ -48,6 +48,16 @@ module Storages
       end
     end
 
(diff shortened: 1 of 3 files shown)


ba1ash (approved):
LGTM!

PR #15287, opened 2024-04-17
[54299] Total values are unset for work packages without any children

cbliard:
See https://community.openproject.org/wp/54299

When setting work or remaining work on a work package, it's odd to see "Total work set to __h" along with "Work set to __h" when this work package does not have any children. By not setting totals for "leaf" work packages, we avoid having it displayed in the activity stream.

Thus the update progress job (migration job) has been updated accordingly, as well as the `UpdateAncestorsService`.

### DIFF
--- a/app/services/work_packages/update_ancestors/loader.rb
+++ b/app/services/work_packages/update_ancestors/loader.rb
@@ -46,17 +46,6 @@ class WorkPackages::UpdateAncestors::Loader
     @descendants[queried_work_package]
   end
 
-  def leaves_of(queried_work_package)
-    @leaves ||= Hash.new do |hash, wp|
-      hash[wp] = replaced_related_of(wp, :leaves) do |leaf|
-        # Mimic work package by implementing the closed? interface
-        leaf.send(:"closed?=", leaf.is_closed)
-      end
-    end
-
-    @leaves[queried_work_package]
-  end
-
   def children_of(queried_work_package)
     @children ||= Hash.new do |hash, wp|
       hash[wp] = descendants_of(wp).select { |d| d.parent_id == wp.id }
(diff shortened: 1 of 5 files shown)


app/services/work_packages/update_ancestors/loader.rb
```diff
@@ -46,17 +46,6 @@ def descendants_of(queried_work_package)
...
-  def leaves_of(queried_work_package)
-    @leaves ||= Hash.new do |hash, wp|
-      hash[wp] = replaced_related_of(wp, :leaves) do |leaf|
-        # Mimic work package by implementing the closed? interface
-        leaf.send(:"closed?=", leaf.is_closed)
-      end
-    end
-
-    @leaves[queried_work_package]
-  end
```
ulferts, line 58: 👍 for noticing this being removable

app/services/work_packages/update_ancestors_service.rb
```diff
@@ -149,22 +147,23 @@ def derive_ignore_non_working_days(ancestor, loader)
   end
 
   def derive_total_estimated_and_remaining_hours(work_package, loader)
-    descendants = loader.descendants_of(work_package)
-
-    work_package.derived_estimated_hours = total(all_estimated_hours([work_package] + descendants))
-    work_package.derived_remaining_hours = total(all_remaining_hours([work_package] + descendants))
+    work_package.derived_estimated_hours = derive_total(work_package, :estimated_hours, loader)
+    work_package.derived_remaining_hours = derive_total(work_package, :remaining_hours, loader)
+    # total(all_remaining_hours([work_package] + descendants))
```
ulferts: leftover I guess

ulferts (approved):
Change looks good. The runtime of the migration job also didn't really change on the dataset I tested with so that makes me hopeful that the same will be true on SaaS.

There is one leftover in the code. Feel free to merge yourself after removing it.

PR #15288, opened 2024-04-17
Allow configuration of sorting for Project Lists

klaustopher:
Implements https://community.openproject.org/work_packages/51671


app/components/projects/configure_view_modal_component.html.erb
```diff
@@ -29,11 +33,20 @@
                                               dragAreaLabel: I18n.t(:'queries.configure_view.columns.drag_area_label'),
                                               appendToComponent: true
                                             }%>
+            </div>
+          <% end %>
+        <% end %>
+        <% tab_panel.with_tab(id: "tab-selects") do |tab| %>
+          <% tab.with_text { I18n.t("label_sort") }%>
+          <% tab.with_panel do %>
+            <%= render(Queries::SortByComponent.new(query: query, selectable_columns:)) %>
```
ulferts, line 42: 👍 for creating a separate component for this.

app/components/projects/configure_view_modal_component.rb
```diff
@@ -30,11 +30,17 @@
...
+  QUERY_FORM_ID = "op-project-list-configure-query-form"
   COLUMN_HTML_NAME = "columns"
 
   options :query
 
+  def selectable_columns
+    @selectable_columns ||= [
+      { id: :lft, name: I18n.t(:label_project_hierarchy) }
+    ] + helpers.projects_columns_options
+  end
```
ulferts, line 42: I think this method would be better placed in the `SortByComponent` since that component is the only one using the method. It could then even become a private method.

app/components/projects/configure_view_modal_component.html.erb
```diff
@@ -29,11 +33,20 @@
                                               dragAreaLabel: I18n.t(:'queries.configure_view.columns.drag_area_label'),
                                               appendToComponent: true
                                             }%>
```
ulferts, line 35: Now that the `SortByComponent` is in, it would make sense to extract the part about the columns into its own component that could then also be reused in the administration. There wouldn't be too much template code in it but a number of helper methods could be moved into it.

app/components/queries/sort_by_component.rb
```diff
@@ -0,0 +1,53 @@
...
+# See COPYRIGHT and LICENSE files for more details.
+# ++
+
+class Queries::SortByComponent < ApplicationComponent
+  options :query
+  options :selectable_columns
+
+  def current_orders
+    JSON.dump(query.orders.map { |order| [order.attribute, order.direction] })
+  end
```
ulferts, line 37: This code serializes the query's order so that it can be placed in the form. I guess using the `params-from-query` controller will not work since the values would be overwritten. And I am still unsure about whether the js based approach turns out to be a good one (currently also discussing with @toy). 

But if we go this way, then I'd think extracting the serialization into a shared helper right away would make sense as I would then expect the functionality to be reused time and again.

app/components/queries/sort_by_component.rb
```diff
@@ -0,0 +1,53 @@
...
+
+  def available_orders
+    @available_orders ||= begin
+      all_order_keys = ::Queries::Register.orders[query.class]&.map(&:key)
+
+      # Keys from the order can be symbols, strings or regexes
+      selectable_columns.select do |column_option|
+        all_order_keys.any? { |order_key| order_key === column_option[:id] }
+      end
```
ulferts, line 54: I would move the code finding the available orders into the `Queries::Orders::AvailableOrders` as that reduces the interface exposed and only keep the part about the interception between the orders and the columns in here.

frontend/src/stimulus/controllers/dynamic/sort-by-config.controller.ts
```diff
@@ -0,0 +1,184 @@
...
+    'inputRow',
+    'inputRowContainer',
+  ];
+
+  declare readonly sortByFieldTarget:HTMLInputElement;
+  declare readonly inputRowTargets:HTMLElement[];
+  declare readonly inputRowContainerTarget:HTMLElement;
+
+  connect():void {
+    this.inputRowTargets.forEach((row) => {
```
ulferts, line 54: You should be able to also use `inputRowTargetConnected`. Not necessary, though.

app/components/queries/sort_by_component.rb
ulferts: Looking at the JS code which does the same, do we need to have this code in here at all? We could let the JS set the value upon initialization.

lookbook/previews/queries/sort_by_component_preview.rb
```diff
@@ -0,0 +1,17 @@
+module Queries
+  class SortByComponentPreview < Lookbook::Preview
+    def default
+      query = ::Queries::Projects::ProjectQuery.new
+      query.order(lft: :asc)
+      query.order(created_at: :desc)
+
+      render SortByComponent.new(
```
ulferts, line 8: This throws an exception since the constant is looked up in the wrong namespace.

lookbook/previews/queries/sort_by_component_preview.rb
```diff
@@ -0,0 +1,17 @@
...
+        query:,
+        selectable_columns: [
+          { id: :lft, name: I18n.t(:label_project_hierarchy) },
+          { id: :created_at, name: I18n.t("attributes.created_at") }
+        ]
+      )
+    end
```
ulferts, line 15: Bonus points for creating a preview 💯

ulferts (commented):
The code structure is heading into the right direction. Unfortunately I found a larger number of issues that need to be worked on before merging. 

From a user perspective I struggled with removing a sorting. It took me a while to understand that I could set it to the blank value again. This is how I see it specified as well but it felt akward.

* [x] Brush up design of component by adding descriptive text, placeholder and margin:

Specification: 
image

https://www.figma.com/file/YCCMdJWkrtP9YSmf49Od0i/Project-lists?type=design&node-id=1830-33701&mode=design&t=F9bkDN6kccdsRZxi-0

Implementation:
image

* [x] Persist sorting when saving a list. Currently the sorting falls back to the "Project hierarchy"

* [ ] Add feature spec that covers both the changing of unpersisted lists as well as saving the sort order upon persisting a list and then querying for it again e.g. in: https://github.com/opf/openproject/blob/dev/spec/features/projects/persisted_lists_spec.rb and https://github.com/opf/openproject/blob/dev/spec/features/projects/projects_index_spec.rb

* [x] Sorting by "Project hierarchy" is currently mutually exclusive with sorting by any other column. The reason for this is the inability to sort by a second/third argument on each level. That is why the projects table upon clicking on the hierarchy sorting trigger resets the sorting to only work with the "Project hierarchy". 

* [x] Sorting by "Project hierarchy" only makes sense in a single direction. When having selected that option, the "Asc"/"Desc" toggle should be removed I think.

* [x] When exporting a project list (e.g. to PDF) the sort order should be applied as well.

app/components/queries/sort_by_component.rb
klaustopher: This will not work as for example the `lft` virtual column is not mentioned in the selects, so it would not work there.

klaustopher:
Bildschirmfoto 2024-05-07 um 19 31 01


I don't know how you got the screenshot you posted, but there is definetly spacing, so I don't know what commit you tested. I'll add the headlines. But the way how the dropdowns are styled are how the primer component looks. There is no chevron indicator, and I don't think we should apply a special styling here...

lookbook/previews/queries/sort_by_component_preview.rb
klaustopher: Bildschirmfoto 2024-05-07 um 19 33 05

Are you sure you opened the most recent commit from this branch? Lookbook works fine

app/components/queries/sort_by_component.rb
klaustopher: This serializes the queries orders to the exact format how they need to be added to the query string. Can we PLEASE PLEASE PLEASE, not throw a reformatting of the query string in this PR, as we did in other places. We can do that in a follow up and change the serialization format ... But I don't know why we should ...

app/components/projects/configure_view_modal_component.html.erb
klaustopher: Isn't this PR about implementing the sorting functionality? Extracting the column stuff is a perfect candidate for a Code maintenance follow up ticket. I'd like not to extend the range of this PR as it already is way to much for what it does.

app/components/projects/configure_view_modal_component.rb
klaustopher: No, it cannot be part of the `SortByComponent` as this component does not know that it is used in a project query context. So we need to pass in the selectable columns and this place is the right place, as it is the last component in the hierarchy that knows that it is dealing with a ProjectQuery

frontend/src/stimulus/controllers/dynamic/sort-by-config.controller.ts
klaustopher: Why would I want to do this. As we are dealing with multiple targets here, the TargetConnected method would fire 3 times and thus this code would  be executed 3 times.

klaustopher:
> Sorting by "Project hierarchy" is currently mutually exclusive with sorting by any other column. The reason for this is the inability to sort by a second/third argument on each level. That is why the projects table upon clicking on the hierarchy sorting trigger resets the sorting to only work with the "Project hierarchy".

> Sorting by "Project hierarchy" only makes sense in a single direction. When having selected that option, the "Asc"/"Desc" toggle should be removed I think.


This is a bit annoying as we are already doing special things like adding the `lft` as a sortable column, because it is nowhere defeined. Now we also need to treat it specially and remove the order buttons. Could we make the sorting by `lft` somehow explicit, and define inside of the `Order` class what directions it supports and then forward this info to the frontend? Also, in the WorkPackages Angular view, it's not even possible to add the hierarchy sorting in the Sort By dialog

klaustopher:
>  When exporting a project list (e.g. to PDF) the sort order should be applied as well.

The `sortBy` params are correctly forwarded to the export, here's an example:

Opening 

```
http://localhost:3000/projects?columns=name+created_at+public+cf_13&sortBy=[["cf_13","desc"],["name","desc"]]
```

I see that the export link correctly points to

```
http://localhost:3000/projects.csv?columns=name+created_at+public+cf_13&sortBy=[["cf_13","desc"],["name","desc"]]
```

So that looks correct. I added a debug output to the `Projects::ExportJob` and when I output the query there, the `orders` param gets converted into a hash and it has the wrong order in there:

```
worker  | ****************************************************************************************************
worker  | #<Queries::Projects::Orders::NameOrder:0x000000032cf5af08>
worker  | #<Queries::Projects::Orders::CustomFieldOrder:0x000000032cf5a8a0>
worker  | ****************************************************************************************************
```

So, something goes wrong when the `sortBy` params are converted to the order for the export job. I'll take a look what's going wrong internally there. But that should be a problem that has already been in the code for exporting code.


---

Also, looking at the job params that get stored in the database, we already have a wrong order:

```js
{
  "job_id": "2277295d-025e-4383-8dfe-e436e8c0022e",
  "locale": "en",
  "priority": 7,
  "timezone": "UTC",
  "arguments": [
    {
      "user": {
        "_aj_globalid": "gid://open-project/User/4"
      },
      "query": {
        "user": {
          "_aj_globalid": "gid://open-project/User/4"
        },
        "orders": {
          "name": {
            "value": "desc",
            "_aj_serialized": "ActiveJob::Serializers::SymbolSerializer"
          },
          "cf_13": {
            "value": "desc",
            "_aj_serialized": "ActiveJob::Serializers::SymbolSerializer"
          },
          "_aj_symbol_keys": [
          ]
        },
        "filters": [
          {
            "name": {
              "value": "active",
              "_aj_serialized": "ActiveJob::Serializers::SymbolSerializer"
            },
            "values": [
              "t"
            ],
            "operator": "=",
            "_aj_symbol_keys": [
              "name",
              "operator",
              "values"
            ]
          }
        ],
        "selects": [
          {
            "value": "name",
            "_aj_serialized": "ActiveJob::Serializers::SymbolSerializer"
          },
          {
            "value": "created_at",
            "_aj_serialized": "ActiveJob::Serializers::SymbolSerializer"
          },
          {
            "value": "public",
            "_aj_serialized": "ActiveJob::Serializers::SymbolSerializer"
          },
          {
            "value": "cf_13",
            "_aj_serialized": "ActiveJob::Serializers::SymbolSerializer"
          }
        ],
        "group_by": null,
        "_aj_symbol_keys": [
          "filters",
          "orders",
          "group_by",
          "selects",
          "user"
        ]
      },
      "export": {
        "_aj_globalid": "gid://open-project/Projects::Export/7"
      },
      "mime_type": {
        "value": "csv",
        "_aj_serialized": "ActiveJob::Serializers::SymbolSerializer"
      },
      "_aj_ruby2_keywords": [
        "export",
        "user",
        "mime_type",
        "query"
      ]
    }
  ],
  "job_class": "Projects::ExportJob",
  "executions": 0,
  "queue_name": "default",
  "enqueued_at": "2024-05-08T06:41:44.632141000Z",
  "scheduled_at": null,
  "provider_job_id": null,
  "exception_executions": {
  }
}
```

frontend/src/stimulus/controllers/dynamic/sort-by-config.controller.ts
ulferts: My suggestion was to use the `inputRowTargetConnected` not with `this.inputRowTargets.forEach` but rather with only the single row that is connected when the element is called. In either case, it was just a suggestion.

lookbook/previews/queries/sort_by_component_preview.rb
ulferts: Whatever the reason, it is working fine now for me as well.

ulferts (approved):
All is fine now functionality wise. I'd still prefer some changes to the code but yes, we can also do it in subsequent PRs.

ulferts:
Good job on digging up the reason for the export problems, @klaustopher

PR #15292, opened 2024-04-17
[54309] Display total work even when set to 0h

cbliard:
See https://community.openproject.org/wp/54309

And renamed some files and descriptions from "estimated time" to "work".

image


ulferts (approved):
👍

PR #15296, opened 2024-04-17
[#53704] Visible=false project attribute values are deleted when a non-admins edit the attributes

dombesz:
https://community.openproject.org/work_packages/53704
### The problem

When a custom field is hidden (`visible=false`) for non-admins (usually for project admins), it is cleared out while the project custom fields are updated from the sidebar. This happens because the hidden field is not submitted with the form, and the `acts_as_customizable` plugin will clear all the fields that are not submitted.

### The solution

In order to fix the issue, we should also include the hidden fields in the backend when assigning the updated custom fields. To make this happen, we have to return the hidden fields too from the `available_custom_fields` method, when the `_query_available_custom_fields_on_global_level` is set to true. This flag signals to the plugin, that we want to request the available_custom_fields globally as opposed to requesting the project enabled custom fields.

Due to the complexity of the acts_as_customizable plugin and the plugin patch, the fix causes [1 test to fail](https://github.com/opf/openproject/pull/15296/files#diff-fe00deb197cdde8955e039ce8b85ba36c57733622ad8753f4553379d57cc0e86R465). The test ensures, we do not activate hidden fields when they are being submitted by non-admin users. This contradicts the bug we are fixing here, because we have to assign the hidden field on the backend in order to not lose its value and its activated state.
This is not a critical issue, because it does not manifest on the frontend, and it will not allow setting any value for the hidden field either. I will provide [a fix](https://github.com/opf/openproject/pull/15296/commits/6701d98f2db1b0f4f253c42040ef6d76a8c1ac31) for the failing testcase in the next minor release.

### DIFF
--- a/app/models/projects/acts_as_customizable_patches.rb
+++ b/app/models/projects/acts_as_customizable_patches.rb
@@ -49,8 +49,8 @@ module Projects::ActsAsCustomizablePatches
     before_create :reject_section_scoped_validation_for_creation
     before_create :build_missing_project_custom_field_project_mappings
 
+    after_create :disable_custom_fields_with_empty_values
     after_save :reset_section_scoped_validation, :set_query_available_custom_fields_to_project_level
-    before_commit :disable_custom_fields_with_empty_values, on: :create
 
     def build_missing_project_custom_field_project_mappings
       # activate custom fields for this project (via mapping table) if values have been provided for custom_fields but no mapping exists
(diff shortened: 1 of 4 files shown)


app/models/projects/acts_as_customizable_patches.rb
```diff
@@ -118,10 +118,21 @@ def available_custom_fields(global: false)
...
         .order("custom_field_sections.position", :position_in_custom_field_section)
 
+      # Do not hide the invisble fields when accessing via the _query_available_custom_fields_on_global_level
+      # flag. Due to the internal working of the acts_as_customizable plugin, when a project admin updates
+      # the custom fields, it will clear out all the hidden fields that are not visible for them.
+      # This happens because the `#ensure_custom_values_complete` will gather all the `custom_field_values`
+      # and assigns them to the custom_fields association. If the `custom_field_values` do not contain the
+      # hidden fields, they will be cleared from the association. The `custom_field_values` will contain the
+      # hidden fields, only if they are returned from this method. Hence we should not hide them,
+      # when accessed with the the _query_available_custom_fields_on_global_level flag on.
```
ulferts: typo "the the"

ulferts (approved):
There is a tiny typo but apart from this, the change looks good, @dombesz. Merge at will.

PR #15313, opened 2024-04-18
[#54178] Activate Storage module when a user has "Manage file storages in project" permission

ba1ash:
https://community.openproject.org/work_packages/54178

#### Solution

The solution is to make `manage_storages_in_project`  not `storages` module permission anymore. Now if `storages` module is disabled it does not take over `manage_storages_in_project`.


config/initializers/permissions.rb
```diff
@@ -177,6 +177,14 @@
                      permissible_on: :global,
                      require: :loggedin,
                      grant_to_admin: true
+
+      map.permission :manage_storages_in_project,
```
Kharonus: 🔴 This line is the one mentioned in daily.

It is preferrable IMHO, to keep storages related code inside the module code tree. So, we need a way to add those permissions from inside the storages module code. You said this is possible, @ba1ash ?

Kharonus (changes requested):

config/initializers/permissions.rb
ba1ash: Yes, that is possible. Have a look, pls.

Kharonus (approved):
LGTM, failing test possibly flaky? I triggered a rerun.

PR #15316, opened 2024-04-18
[54305] Mobile issues with progess reporting modal

bsatarnejad:
- [x] Drop modal is full-screen when the screen width is smaller than 1000px, so we want to make it possible for drop modal component to be not full-screen on screens smaller than 1000px.
- [x] Force progress modal elements to be in one column below each other, when it is in mobile mode.

https://community.openproject.org/projects/openproject/work_packages/54305/activity


app/components/work_packages/progress/work_based/modal_body_component.sass
```diff
@@ -3,3 +3,7 @@
     background-color: unset !important
     border: none !important
     box-shadow: none !important
+
+  @media screen and (max-width: $breakpoint-sm)
```
HDinger: You are repeating the same code block here. Please use either a class or (even better) just use the [system_argumens](https://primer.style/view-components/lookbook/pages/system_arguments) for that.

frontend/src/app/spot/styles/sass/components/drop-modal.sass
```diff
@@ -35,6 +34,9 @@
       width: calc(100vw - (2 * #{$spot-spacing-1}))
       height: calc(#{var(--app-height)} - (#{$spot-spacing-3_5} + #{$spot-spacing-1}))
 
+      &_not-full-screen
```
HDinger, line 37: We just recently decided that on mobile, the modals should be centered. For the drop-modal this was not relevant until now as they always span the whole height. With that new option, we have to take care of that. Currently the progress modal is shown at the bottom.

HDinger (changes requested):
Looks good overall, I have some minor remarks. 👍

HDinger:
I think there should be more space between the fields. Currently it is not clear which field belongs to which label 
Bildschirmfoto 2024-04-19 um 07 49 41

frontend/src/global_styles/openproject/_primer-adjustments.sass
```diff
@@ -64,3 +64,12 @@ page-header
 .Overlay
   &-body_autocomplete_height
     min-height: 300px
+
+// For some specific cases like progress form in small screens,
+// we need to change the direction of elements to column,
+// therefore elements are placed below each other
+.horizontal-form-group
```
HDinger: The sense of this class is to align the group vertically on mobile. I'd like that to be reflected in the class name. Further, please try to stick to our BEM convention, e.g `.Form_mobile-vertical` or something similar.

frontend/src/global_styles/openproject/_primer-adjustments.sass
```diff
@@ -64,3 +64,12 @@ page-header
...
+  @media screen and (max-width: $breakpoint-sm)
+    .FormControl-horizontalGroup
+      flex-direction: column
+      row-gap: 16px
```
HDinger: Please use a `rem` based value here.

app/components/work_packages/progress/work_based/modal_body_component.html.erb
```diff
@@ -7,7 +7,7 @@
         data: { "application-target": "dynamic",
                 controller: "work-packages--progress--focus-field" }) do |f| %>
     <%= flex_layout do |modal_body| %>
-      <% modal_body.with_row do |_fields| %>
+      <% modal_body.with_row(classes: "horizontal-form-group") do |_fields| %>
         <%= render(WorkPackages::ProgressForm.new(f, work_package:, mode:, focused_field:)) %>
```
HDinger, line 11: It feels weird, that this class is on the body as it actually belongs to the form. I know, that the Primer:: Group doesn't apply the system_arguments but maybe there is still an alternative way. Maybe you could add a parameter `mobile_alignment` (or whatever) and inside the component render a div around the form which is only responsible to display the class if the parameter is set ?

HDinger (changes requested):

HDinger (approved):

PR #15326, opened 2024-04-19
Feature/52144 changing a persisted list

toy:
[OP#52144](https://community.openproject.org/projects/openproject/work_packages/52144)

### DIFF
--- a/app/controllers/projects/queries_controller.rb
+++ b/app/controllers/projects/queries_controller.rb
@@ -32,7 +32,7 @@ class Projects::QueriesController < ApplicationController
   # No need for a more specific authorization check. That is carried out in the contracts.
   before_action :require_login
   before_action :find_query, only: %i[update destroy]
-  before_action :load_query_or_deny_access, only: %i[new]
+  before_action :build_query_or_deny_access, only: %i[new create]
 
   current_menu_item [:new, :create] do
     :projects
@@ -46,7 +46,7 @@ class Projects::QueriesController < ApplicationController
 
   def create
     call = Queries::Projects::ProjectQueries::CreateService
-             .new(user: current_user)
+             .new(from: @query, user: current_user)
              .call(permitted_query_params)
 
     if call.success?
--- a/app/controllers/projects/queries_controller.rb
+++ b/app/controllers/projects/queries_controller.rb
@@ -50,8 +50,12 @@ class Projects::QueriesController < ApplicationController
              .call(permitted_query_params)
 
     if call.success?
+      flash[:notice] = I18n.t("lists.create.success")
+
       redirect_to projects_path(query_id: call.result.id)
     else
+      flash[:error] = I18n.t("lists.create.failure", errors: call.errors.full_messages.join("\n"))
+
       render template: "/projects/index",
              layout: "global",
              locals: { query: call.result, state: :edit }
@@ -64,8 +68,12 @@ class Projects::QueriesController < ApplicationController
              .call(permitted_query_params)
 
     if call.success?
+      flash[:notice] = I18n.t("lists.update.success")
+
       redirect_to projects_path(query_id: call.result.id)
     else
+      flash[:error] = I18n.t("lists.update.failure", errors: call.errors.full_messages.join("\n"))
+
       render template: "/projects/index",
              layout: "global",
              locals: { query: call.result, state: :edit }
--- a/app/controllers/projects/queries_controller.rb
+++ b/app/controllers/projects/queries_controller.rb
@@ -31,7 +31,7 @@ class Projects::QueriesController < ApplicationController
 
   # No need for a more specific authorization check. That is carried out in the contracts.
   before_action :require_login
-  before_action :find_query, only: :destroy
+  before_action :find_query, only: %i[update destroy]
   before_action :load_query_or_deny_access, only: %i[new]
 
   current_menu_item [:new, :create] do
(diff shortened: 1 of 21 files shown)


spec/services/queries/projects/project_queries/create_service_spec.rb
```diff
@@ -33,4 +33,47 @@
...
+      let(:instance) { build(:project_query).where("public", "=", "t").order(name: :desc) }
+
+      it "returns the instance" do
+        expect(result).to eq(instance)
+      end
+
+      it "keeps instance value for attribute not passed in params" do
+        expect(result).to have_attributes(
+          orders: [having_attributes(attribute: :name, direction: :desc)]
+        )
```
ulferts, line 53: I wasn't aware of the `have_attributes` matcher. Reads quite nicely. TIL

config/locales/en.yml
```diff
@@ -339,7 +339,14 @@ Project attributes and sections are defined in the <a href=%{admin_settings_url}
...
   lists:
-    can_be_saved_as: "The modifications can only be saved in a new list."
+    create:
+      success: "The list has been saved"
+      failure: "The list cannot be saved: %{errors}"
+    update:
+      success: "The modified list has been saved"
+      failure: "The modified list cannot be saved: %{errors}"
+    can_be_saved: "List modified:"
+    can_be_saved_as: "There are modifications but this view cannot be edited:"
```
ulferts: You are correct in that the figma files state this string. But they should not do so any more since @marcalcobe and I agreed to use "The modifications can only be saved in a new list."

spec/models/queries/projects/factory_spec.rb
```diff
@@ -869,7 +869,7 @@
 
     it "is ordered by lft asc" do
       expect(find.orders.map { |order| [order.attribute, order.direction] })
-        .to eq([["lft", :asc]])
+        .to eq([%i[lft asc]])
     end
```
ulferts, line 1051: The changes done around duplicating an existing query are not reflected in these specs.

app/controllers/projects/queries_controller.rb
```diff
@@ -46,12 +46,34 @@ def new
 
   def create
     call = Queries::Projects::ProjectQueries::CreateService
-             .new(user: current_user)
+             .new(from: @query, user: current_user)
              .call(permitted_query_params)
```
ulferts, line 50: In other places of the application, we relied on `CopyServices` for doing operations like this. I think that would have the potential to keep the factory more focused of course at the expanse of a higher variety in the controller.

ulferts (changes requested):
Very solid work, @toy. The behaviour itself is implemented very thoroughly and the only thing I can complain about behaviour wise is something I only spotted because of some previous bugfix.

If a list employs a custom field as a filter or a column, is persisted, and then the custom field is deleted, a valid subset is generated to prevent errors from being displayed. The same scenario (which can only happen after sharing is implemented) is one where an admin only filter or column is part of a list which is then called by a non admin user. In that case, because the filters/columns are changed, the notification "List modified" is displayed although the user did not change anything. I guess the best approach would be to avoid the query to be flagged as dirty in case an invalid filter/column/order is removed.

app/controllers/projects/queries_controller.rb
toy: I found usage of `CopyService`s only in api endpoints dedicated to copying, not in controllers. Also `BaseServices::Copy` is copying instance without changes. It is possible to first change an instance and then copy it, or even add overriding attributes in `…::ProjectQueries::CopyService`, but to me "creating from template" looks more straightforward.

config/locales/en.yml
toy: Updated few more texts and used colon instead of dot

app/controllers/projects/queries_controller.rb
ulferts: If I understand the need for doing all this correctly is that when you use "Save as" you might not have all the parameters sent along with the request but rather just an id of an existing query from which you want to deviate from. 

With the current approach I see the responsibility for all this split between the Controller (or the QueryLoading module), the factory and the create service. This is what I don't believe to be a good split especially since the parameter processing happens in two places. And the creation of the new query now also somewhat happens in the two places since the new query is instantiated in the factory and then passed to the CreateService to be overwritten and persisted. 

I would still think that having the responsibility of parameter processing in the service would be the cleanest solution but it seems you are right that most CopyServices don't take override parameters. Others do which makes it even weirder.

Would it be an option to place the parameters of the query one saves from (as identified by the query_id param) in the params passed into the create service? So that the whole parameters from the params sent with the request and also the ones taken from the source query are combined in `permitted_query_params` when they are passed to the create service?

toy: > Would it be an option to place the parameters of the query one saves from (as identified by the query_id param) in the params passed into the create service? So that the whole parameters from the params sent with the request and also the ones taken from the source query are combined in permitted_query_params when they are passed to the create service?

I've tried this at first, but didn't find a decent way to get attributes from the query in serialised form

toy:
> If a list employs a custom field as a filter or a column, is persisted, and then the custom field is deleted, a valid subset is generated to prevent errors from being displayed. The same scenario (which can only happen after sharing is implemented) is one where an admin only filter or column is part of a list which is then called by a non admin user. In that case, because the filters/columns are changed, the notification "List modified" is displayed although the user did not change anything. I guess the best approach would be to avoid the query to be flagged as dirty in case an invalid filter/column/order is removed.

See 5fb2ce0c55e94601370e95560136a2e7cd43e827 and b94ccb204ecaed8f68021fe7c9b6a7b53586a16c

spec/models/queries/projects/factory_spec.rb
toy: Indeed, it is already checked in feature tests, but should be in factory spec too, so added in de51c5a63e3a6ef79027b535427a193a31427e06, also some fixed for existing tests in 2cc9f9cc46dbb996b95471c7c4202754417ca407 and b70fb78997802ae03f7418898b69dfd421a36050

app/controllers/projects/queries_controller.rb
ulferts: After looking into the code, I understand that struggle. 

Let's work with the current solution for now and create a code maintenance ticket to turn the copy services to take override parameters.

ulferts (approved):
Once again, solid work @toy

PR #15328, opened 2024-04-19
[15328 ] Use better messages for progress calculation adjustments

cbliard:
See https://community.openproject.org/wp/54328

It has a better wording and links to the blog post explaining the changes in progress calculations.

### DIFF
--- a/app/workers/work_packages/update_progress_job.rb
+++ b/app/workers/work_packages/update_progress_job.rb
@@ -30,31 +30,16 @@
 class WorkPackages::UpdateProgressJob < ApplicationJob
   queue_with_priority :default
 
+  attr_reader :current_mode, :previous_mode
+
   def perform(current_mode:, previous_mode:)
+    @current_mode = current_mode
+    @previous_mode = previous_mode
+
     with_temporary_progress_table do
-      case current_mode
-      when "field"
-        unset_all_percent_complete_values if previous_mode == "disabled"
-        fix_remaining_work_set_with_100p_complete
-        fix_remaining_work_exceeding_work
-        fix_only_work_being_set
-        fix_only_remaining_work_being_set
-        derive_unset_remaining_work_from_work_and_p_complete
-        derive_unset_work_from_remaining_work_and_p_complete
-        derive_p_complete_from_work_and_remaining_work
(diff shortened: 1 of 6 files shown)


app/workers/work_packages/update_progress_job.rb
```diff
@@ -290,10 +296,19 @@ def copy_progress_values_to_work_packages
   end
 
   def create_journals_for_updated_work_packages(updated_work_package_ids)
+    cause = { type: "system_update", feature: system_update_explanation }
```
ulferts, line 299: Just my preference but I don't see the benefit of having it assigned to a variable here over having it be defined inside the `call` method call

lib/open_project/journal_formatter/cause.rb
```diff
@@ -73,8 +73,20 @@ def cause_description(cause, html)
     end
   end
 
-  def system_update_message(cause)
-    I18n.t("journals.cause_descriptions.system_update.#{cause['feature']}")
+  def system_update_message(cause, html)
+    feature = cause["feature"]
+    feature = "progress_calculation_adjusted" if feature == "progress_calculation_changed"
```
ulferts, line 78: Is this "progress_calculation_changed" in here to have it displayed on instances like the community where the job has already run?

ulferts (commented):
LGTM, @cbliard. I have one question but I already assume that I know the answer. 

The CI currently complains but with that fixed, feel free to merge.

ulferts (approved):
I already meant to approve before

app/workers/work_packages/update_progress_job.rb
cbliard: It's to avoid creating a new hash every time in the loop. It saves some memory.
Maybe a premature optimization from my side.

lib/open_project/journal_formatter/cause.rb
cbliard: Yes it is. And on stage and edge too.

cbliard:
> The CI currently complains but with that fixed, feel free to merge.

Yes, I did a last minute change and forgot to rerun the tests. It should be fixed now.

PR #15336, opened 2024-04-22
Code maintenance/54117 refactor storage view components to use the component streams that are more declarative at the controller level

akabiru:
https://community.openproject.org/work_packages/54117


akabiru:
Scrapping this, will restart. 🧹

PR #15337, opened 2024-04-22
Feature/53771 add % complete to section estimates and time

ulferts:
### TODOs

* [x] move `percentage_done` into `:estimates_and_time` attribute group
* [x] avoid showing the `:estimates_and_time` attribute group for milestone types
* [x] Rename `:estimates_and_time` to 'Estimates and progress'
  * This last point was done half-baked. It would have been way cleaner to rename the key as well but that would require a migration renaming the already stored keys in the database which was refrained from given the short time before the release. 

-------------

https://community.openproject.org/work_packages/53771

### DIFF
--- a/app/models/type/attribute_groups.rb
+++ b/app/models/type/attribute_groups.rb
@@ -41,11 +41,11 @@ module Type::AttributeGroups
     # May be extended by plugins
     mattr_accessor :default_group_map do
       {
-        author: :people,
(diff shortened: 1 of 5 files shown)


ulferts:
There is currently a flickering feature spec that is reported as failing on this PR.

oliverguenther (approved):

PR #15338, opened 2024-04-22
[#54349] Dragging agenda items to the top not possible on long agendas

mrmir:
https://community.openproject.org/work_packages/54349

I don't know if the way I added the method and called it inside connect() is correct, but this seems to work as needed.


dombesz (approved):
LGTM 👍
This is not related to the PR, but we also have a [Generic drag and drop stimulus controller](https://github.com/opf/openproject/blob/dev/frontend/src/stimulus/controllers/dynamic/generic-drag-and-drop.controller.ts), that we probably have to refactor at some point in order to avoid code duplication.

Merging, because the failing spec is not related.

PR #15339, opened 2024-04-22
[#53112] 14.0 Documentation: Remove member and revoke shared work pac…

MayaBerd:
…kages  https://community.openproject.org/work_packages/53112

[#53112] 14.0 Documentation: Remove member and revoke shared work packages

https://community.openproject.org/work_packages/53112


docs/user-guide/members/README.md
```diff
@@ -74,19 +74,35 @@ Find out [here](../../getting-started/invite-members/#add-existing-users) how to
 
 To change the role of a member within a project, select the corresponding project and open the Members module.
 
-To edit an existing member of a project, click the **Edit icon** in the list next to the member on the right. Choose the new role and press the blue **Change button** so save your changes.
+To edit an existing member of a project, click the **Icon with three dots** in the list next to the member on the right and select **Manage roles**. Choose the new role and press the green **Change ** button so save your changes.
```
marcalcobe: Hi @MayaBerd, I would call the icon **More icon** instead of using the term three dots.

marcalcobe: Also maybe change the sentence "Choose the new role... " as the user can also remove existing roles from this manage action. Something like "Add and remove roles and..." maybe?

docs/user-guide/members/README.md
```diff
@@ -74,19 +74,35 @@ Find out [here](../../getting-started/invite-members/#add-existing-users) how to
...
 
 ![Edit project members in OpenProject](edit-project-member.png)
 
 
 
 ## Remove members
 
-To remove members from a project, [select the project](../../getting-started/projects/#open-an-existing-project) for which you want to remove the members. In the project menu on the left, select the **Members** module. In the members list, click the delete icon at the right end of the row with the corresponding member name.
+To remove members from a project, [select the project](../../getting-started/projects/#open-an-existing-project) for which you want to remove the members. In the project menu on the left, select the **Members** module. In the members list, click the three dot icon at the right end of the row with the corresponding member name and select **Remove member**.
```
marcalcobe: In the edit members section we use the sentence:

`To change the role of a member within a project, select the corresponding project and open the Members module.`

Maybe we should use the same style also? I would also change the three dot icon to more icon.

marcalcobe: Maybe we need to also add a note to specify that this actions are only doable with the correct permisions.

docs/user-guide/members/README.md
```diff
@@ -74,19 +74,35 @@ Find out [here](../../getting-started/invite-members/#add-existing-users) how to
...
 
 ![Remove project members in OpenProject](delete-project-member.png)
 
+You will be asked to confirm your decision. 
 
+![Confirm removing a user from a project in OpenProject](confirm-user-deletion.png)
+
+If the project member you are removing has shared work packages, you will also be asked whether these sharing rights also need to be removed.
+
+![Confirm removing a user with shared work packages from a project in OpenProject](openproject_user_guide_confirm_user_removal_with_shares.png)
```
marcalcobe, line 95: I think we need to add a note to explain the case where users are part of a group and therefore the individual role of the user will be removed but the group role will still be there. You can find the mockups to understand better this case in this [Figma file](https://www.figma.com/file/lLDYzQW7evoBA4lha5PwXO/Project-members?type=design&node-id=1142-37273&mode=design&t=l3L5FYnETbyYTSb7-4).

I saw that we are already doing this disclaimer for revoking group work package shares, maybe we can reuse that note, something like this:

`Note: the remove action will only affect the member individual roles. All those roles obtained via a group will not be removed. To remove those group roles you can either remove the member from the group or remove the entire group from the project.`

docs/user-guide/members/README.md
```diff
@@ -74,19 +74,35 @@ Find out [here](../../getting-started/invite-members/#add-existing-users) how to
...
+
+## Revoke sharing privileges
+
+If a work package has been [shared](../work-packages/share-work-packages), you may need to revoke sharing privileges at a later stage in the project. To do that select the **three dot icon** at the right end of the row with the corresponding member name and select **Revoke work package shares**. You can also choose the **View shared work packages** option to see the list of all shared work packages.
```
marcalcobe: Here also changing the three dot term for more icon

marcalcobe: I would change the end of the paragraph to "...list of all work packages shared with the user."

docs/user-guide/members/README.md
```diff
@@ -74,19 +74,35 @@ Find out [here](../../getting-started/invite-members/#add-existing-users) how to
...
+
+![Remove sharing privileges in OpenProject](openproject_user_guide_members_remove_work_package_shares.png)
+
+> Note: it is not possible to revoke sharing rights of project members, if they are a part of a project only as members of a group. In this case you can either remove a single member from the group (and then remove the rights), or revoke the shared privileges from the entire group.
```
marcalcobe: I find this a bit confusing. If this disclaimer is about revoking work package shares I will phrase it a bit different:

`Note: the revoking action will only affect the individually work packages shares. All those shares with the user via a group will not be revoked. To revoke those group shares you can either remove the member from the group or revoke the privileges from the entire group.`

With that we could delete the sentence bellow.

docs/user-guide/work-packages/share-work-packages/README.md
```diff
@@ -70,7 +70,11 @@ Once a work package has been shared, you will see the number of shared users in
 
 ## Remove sharing privileges
 
-You can also remove the user from the list by clicking on **Remove** next to the user name. Please note that this will not remove a user entirely, but only revoke the viewing and/or editing rights for the work package. User account will remain intact. If you need to [delete a user](../../../system-admin-guide/users-permissions/users/#delete-users), please do that in system administration or contact your administrator.
+You can also remove the user from the list by clicking on **Remove** next to the user name. Please note that this will not remove a user entirely, but only revoke the viewing and/or editing rights for the work package. User account will remain intact.
```
marcalcobe: There is also a "Comment" role so maybe we could change the second sentence to "...only revoke the rights for this work package."

marcalcobe (commented):
Hi @MayaBerd, I added my review points to the PR. Hope it helps.

PR #15341, opened 2024-04-22
[54370] Prevent error when saving unset progress modal

cbliard:
See https://community.openproject.org/wp/54370

A 500 is produced when being on a work package creation form, opening the progress modal, and clicking save without setting any values.


ulferts (approved):
👍

PR #15342, opened 2024-04-22
[#53996] use auth strategies in create folder commands

Kharonus:
[#53996](https://community.openproject.org/wp/53996)

### WAT?

- the `provider.commands.create_folder` commands were implemented very differently
- unified the interface (calling parameters, return values)
- used new authentication method
- rerecorded vcr cassettes for new login (that were many \*sigh\*)
- added new tests for create and delete folder
  - now a single test covers all provider types to enforce common interface, see spec for details
- replaced usage in production code


modules/storages/spec/common/storages/peripherals/storage_interaction/create_folder_command_spec.rb
Kharonus, line 1: This is the new spec format, where both provider type commands are tested against the same shared examples.

modules/storages/app/common/storages/peripherals/storage_interaction/nextcloud/create_folder_command.rb
```diff
@@ -32,52 +32,96 @@ module Storages::Peripherals::StorageInteraction::Nextcloud
...
-                     )
-                   )
+    def call(auth_strategy:, folder_name:, parent_location:)
+      origin_user_id = Util.origin_user_id(caller: self.class, storage: @storage, auth_strategy:)
+      if origin_user_id.failure?
+        return origin_user_id
+      end
+
+      folder_path = Util.join_uri_path(parent_location, folder_name)
+      base_url = Util.join_uri_path(@storage.uri, "remote.php/dav/files", CGI.escapeURIComponent(origin_user_id.result))
```
mereghost: 🟡 Wouldn't be better to have the `base_url` without the `origin_user_id` defined as method as we do with the `OneDrive` adapter?

modules/storages/app/common/storages/peripherals/storage_interaction/nextcloud/create_folder_command.rb
```diff
@@ -32,52 +32,96 @@ module Storages::Peripherals::StorageInteraction::Nextcloud
...
+      path_prefix = URI.parse(base_url).path
+      request_url = Util.join_uri_path(base_url, Util.escape_path(folder_path))
+
+      Auth[auth_strategy].call(storage: @storage) do |http|
+        result = handle_response(http.mkcol(request_url))
+        return result if result.failure?
+
+        handle_response(http.propfind(request_url, requested_properties)).map(&storage_file(path_prefix))
```
mereghost: 🟡 I may be alone on this, but I find link chains hard to parse/understand, specially with the weird `map(&method(param))` in the end.

`map` takes a block, why not use one so you can remove the inner lambda from the method?

```ruby
service_result.map { |result| storage_file(result, path_prefix) }
```

looks clearer and conveys more intent as you see that it is working over the `result` and needs one extra parameter.

modules/storages/app/common/storages/peripherals/storage_interaction/nextcloud/util.rb
```diff
@@ -80,7 +80,46 @@ def token(user:, configuration:, &)
...
-      Nokogiri::XML(response.body).xpath("//s:message").text
+      response.xml.xpath("//s:message").text
+    end
+
+    def origin_user_id(caller:, storage:, auth_strategy:)
+      case auth_strategy.key
+      when :basic_auth
+        ServiceResult.success(result: storage.username)
+      when :oauth_user_token
+        origin_user_id = OAuthClientToken.find_by(user_id: auth_strategy.user, oauth_client: storage.oauth_client)&.origin_user_id
```
mereghost: 🟡 Since we expect just one record to return you could use [`pick`](https://api.rubyonrails.org/classes/ActiveRecord/Calculations.html#method-i-pick)

```suggestion
        origin_user_id = OAuthClientToken.where(user_id: auth_strategy.user, oauth_client: storage.oauth_client).pick(:origin_user_id)
```

modules/storages/app/common/storages/peripherals/storage_interaction/one_drive/create_folder_command.rb
```diff
@@ -35,31 +35,34 @@ module OneDrive
         class CreateFolderCommand
           using ServiceResultRefinements
 
-          def self.call(storage:, folder_path:)
-            new(storage).call(folder_path:)
+          Auth = ::Storages::Peripherals::StorageInteraction::Authentication
```
mereghost: 🟡 Do we need this?

If we declare the namespacing without a shorthand (`Namespace::OtherNamespace` - which I'd dare say is one of the sad antipatterns pushed by RoR), Ruby should be able to infer that `Authentication` would refer to the one under `StorageInteraction` as this is in the path of name resolution.

modules/storages/app/common/storages/peripherals/storage_interaction/one_drive/create_folder_command.rb
```diff
@@ -97,15 +100,15 @@ def file_info_for(json_file)
...
             {
-              name: folder_path,
+              name: folder_name,
               folder: {},
               "@microsoft.graph.conflictBehavior" => "fail"
             }.to_json
           end
 
-          def base_uri = "/v1.0/drives/#{@storage.drive_id}"
+          def base_uri = Util.join_uri_path(@storage.uri, "v1.0", "drives", @storage.drive_id)
```
mereghost: 🟡 Do we need to use `File.join` (knowing it will break on Windows) at all? We know how to build the URI and hardly it has any data coming from a user, so not even that would break it.

modules/storages/app/services/storages/one_drive_managed_folder_sync_service.rb
```diff
@@ -183,6 +189,10 @@ def client_tokens_scope
       OAuthClientToken.where(oauth_client: @storage.oauth_client)
     end
 
+    def auth_strategy
+      Peripherals::StorageInteraction::AuthenticationStrategies::OAuthClientCredentials.strategy
```
mereghost: 🟡 WDYT about using the `authentication.userbound|userless` Registry entries?

modules/storages/spec/common/storages/peripherals/storage_interaction/create_folder_command_spec.rb
mereghost: 🟡 As I mentioned to you, I'd still keep the specs separated, but using the same shared examples.

Maybe we should move the shared examples to a support `adapter_examples` (or better named) folder.

modules/storages/spec/common/storages/peripherals/storage_interaction/create_folder_command_spec.rb
```diff
@@ -0,0 +1,197 @@
...
+    it "returns a failure" do
+      result = described_class.call(storage:, auth_strategy:, folder_name:, parent_location:)
+
+      result.match(
+        on_failure: ->(error) do
+          expect(error.code).to eq(:not_found)
+          expect(error.data.source).to eq(error_source)
+        end,
+        on_success: ->(result) { fail "Expected failure, got #{result}" }
+      )
```
mereghost: 🟡 IMO this style offers little in terms of clarity.

The suggestion below feels more idiomatic and is shorter. =)

```suggestion
      expect(result).to be_failure
   
      error = result.error
      expect(error.code).to eq(:not_found)
      expect(error.data.source).to eq(error_source)
```

mereghost (commented):
Left a bunch of comments so we can have an organized discussion. =)

modules/storages/app/common/storages/peripherals/storage_interaction/one_drive/create_folder_command.rb
Kharonus: The main reason for using it is to harden the usage against trailing slashes in the data. Is Windows really an issue here? Windows is no target platform for OP.

modules/storages/app/common/storages/peripherals/storage_interaction/one_drive/create_folder_command.rb
Kharonus: done

modules/storages/app/common/storages/peripherals/storage_interaction/one_drive/create_folder_command.rb
mereghost: It is more about the signal vs noise ratio here. The URI is fixed and has no trailing slash, and if configured correctly it won´t have one or the entire integration won't work.

This is a issue on Nextcloud as there's user define path parts (Folder names). As for file names, OD don't even allow slashes (of either type).

mereghost (approved):
Looks great! :D

Thanks for fixing the mess I created.

PR #15379, opened 2024-04-25
Update Ruby to 3.3.1

klaustopher:
Updating to Ruby 3.3.1

- [x] Make cache on GitHub actions work with ruby version updates
- [x] Ruby Parser gem does not support 3.3.1 yet, but as we are not using the fixed syntax, it's not a problem. See https://github.com/whitequark/parser/issues/963

## Optional
- [x] Enable YJIT because it can give real performance improvements

---

Closes https://community.openproject.org/work_packages/54458


klaustopher:
I am a bit confused here. The cache includes the `.ruby-version` file as a cache key, but it does not invalidate the cache. That's why the runs like https://github.com/opf/openproject/actions/runs/8828307074/job/24237045450 failed. When I completely disable the cache, it works nicely.

@machisuji or @crohr do you maybe have an idea why that's the case?

AaronDewes:
> I am a bit confused here. The cache includes the `.ruby-version` file as a cache key, but it does not invalidate the cache. That's why the runs like https://github.com/opf/openproject/actions/runs/8828307074/job/24237045450 failed. When I completely disable the cache, it works nicely.
> 
> @machisuji or @crohr do you maybe have an idea why that's the case?

> `restore-keys` - An ordered list of prefix-matched keys to use for restoring stale cache if no cache hit occurred for key.

You have that set. So if no previous cache with the same ruby version and Gemfile.lock exists, any other one will be used.

Fix would be to has the ruby version and lockfile separately and add the version hash to restore-keys too.

klaustopher:
> Fix would be to has the ruby version and lockfile separately and add the version hash to restore-keys too.

Isn't that the case with `key: gem-${{ hashFiles('Gemfile.lock', '.ruby-version') }}`? IMHO this should add the content of the `.ruby-version` file to the cache key and if we change the content of that file, the cache key should be a new one and the cache lookup should miss?

.github/workflows/test-core.yml
```diff
@@ -30,45 +30,45 @@ jobs:
...
+#   - name: Restore CI image from cache
+#     if: steps.cache_docker.outputs.cache-hit == 'true'
+#     run: docker load -i cache/docker/image.tar
+#   - name: Cache GEM
+#     uses: runs-on/cache@v4
+#     with:
+#       path: cache/bundle
+#       key: gem-${{ hashFiles('Gemfile.lock', '.ruby-version') }}
+#       restore-keys: |
+#         gem-
```
AaronDewes:
```suggestion
#       key: gem-${{ hashFiles('.ruby-version') }}-${{ hashFiles('Gemfile.lock') }}
#       restore-keys: |
#         gem-${{ hashFiles('.ruby-version') }}
```

AaronDewes:
Vielleicht kann ich es auf Deutsch besser erklären als auf Englisch: `restore-keys` ist eine Liste von Prefixen für den `key`. Wenn du also als `key` den Wert `gem-${{ hashFiles('Gemfile.lock', '.ruby-version') }}` hast, dann überprüft es erst, ob es ein Cache mit dem Key `gem-${{ hashFiles('Gemfile.lock', '.ruby-version') }}` schon gibt. Wenn ja, nimmt es immer den. Wenn nein, kann es den ja nicht nehmen.

Aber weil `gem-` ohne sonst etwas in den restore-keys steht, nimmt es dann einfach irgendein Cache, dessen key auch mit `gem-` beginnt. Und das kann eine andere Ruby-Version sein.

klaustopher:
Ahhh! Thank you, that makes a lot of sense!

klaustopher:
Parser gem was updated, so we can go ahead and merge this

oliverguenther (approved):

.github/workflows/test-core.yml
```diff
@@ -44,9 +44,9 @@ jobs:
       uses: runs-on/cache@v4
       with:
         path: cache/bundle
-        key: gem-${{ hashFiles('Gemfile.lock') }}
+        key: gem-${{ hashFiles('.ruby-version') }}-${{ hashFiles('Gemfile.lock') }}
         restore-keys: |
-          gem-
+          gem-${{ hashFiles('.ruby-version') }}
```
toy: Just to suggest that it is a prefix:
```suggestion
          gem-${{ hashFiles('.ruby-version') }}-
```

PR #15381, opened 2024-04-25
Add "no permission" message to file links using OneDrive/SharePoint 

mereghost:
## What and Why

Those are in the work package linked above, so I'll omit for brevity.

## How?

The tooltip is controlled by the `file-link-list-item` component and renders the related tooltip based on the [status.href](https://github.com/opf/openproject/blob/bd45359d0c8067fea53348a1f67132d9dc49878a/frontend/src/app/shared/components/storages/file-link-list-item/file-link-list-item.component.ts#194) property the file link that is returned by our API under `/work_packages/:id/file_links`.

This endpoint, calls the [FileLinkService](https://github.com/opf/openproject/blob/3b2121f7333f0ce7f3515b29c760c0580c31a245/modules/storages/lib/api/v3/file_links/work_packages_file_links_api.rb#L50). This service loads the FileLinks from the database and also calls the [FilesInfoQuery](https://github.com/opf/openproject/blob/905a5dd7f679af62468702bd95684c35be898ccc/modules/storages/app/services/storages/file_link_sync_service.rb#L57) for the appropriate storage.

From there it updates some metadata and eventually, looking at the [status_code](https://github.com/opf/openproject/blob/905a5dd7f679af62468702bd95684c35be898ccc/modules/storages/app/services/storages/file_link_sync_service.rb#L88) of the `StorageFileInfo`, it sets the ephemeral `origin_status` ephemeral attribute on the file link. Of our main interest is the `view_not_allowed` value that's set when the status_code is 403.

And here is where the issue lies. *OneDrive* behaves slightly different than *Nextcloud*: the latter returns a 403 saying the file exists but the user has no access to it, the former returns a 404 as the user can't even see the file.

So, to solve this we do a [second request](https://github.com/opf/openproject/blob/5c824eaa676f482f14d9fb8b471282e74fa4f42b/modules/storages/app/common/storages/peripherals/storage_interaction/one_drive/file_info_query.rb#L65) if we get a 404 from OneDrive, using the application account. If the application can see it, we construct the `StorageFileInfo` with a 403, otherwise we return the good old 404.

PS: `FilesInfoQuery` on OneDrive calls `FileInfoQuery`, so the fix is implemented on it so they behave the same as Nextcloud.


modules/storages/app/common/storages/peripherals/storage_interaction/one_drive/file_info_query.rb
```diff
@@ -34,58 +34,80 @@ module StorageInteraction
...
-                                                     data: StorageErrorData.new(source: self.class),
-                                                     log_message: "File ID can not be nil")
+                errors: StorageError.new(code: :error,
+                                         data: @error_data, log_message: "File ID can not be nil")
               )
             end
 
-            Auth[auth_strategy].call(storage: @storage) do |http|
-              @delegate.call(http:, drive_item_id: file_id, fields: FIELDS).map(&storage_file_infos)
+            user_bound_result = Authentication[auth_strategy].call(storage: @storage) do |http|
```
Kharonus: 🟡 shouldn't we differentiate here? right now you assume, that the first result is a `user_bound_result`. But one can directly call the `file_info_query` with a user-less strategy. I'd rather rename it to something like `file_info_result`. And later in the check combine the conditions `if sr.result == :not_found && auth_strategy.user.present?`.

modules/storages/app/common/storages/peripherals/storage_interaction/one_drive/file_info_query.rb
```diff
@@ -34,58 +34,80 @@ module StorageInteraction
...
+              mime_type: Util.mime_type(json),
+              size: json[:size],
+              owner_name: json.dig(:createdBy, :user, :displayName),
+              owner_id: json.dig(:createdBy, :user, :id),
+              trashed: false,
+              permissions: nil,
+              location: Util.extract_location(json[:parentReference], json[:name])
+            )
+
+            add_file_change_info(info, json)
```
Kharonus: 🟡 is this deconstruction really necessary, just to please the linter?? I was at least 30 seconds looking at it and asked myself "Why?". If the linter rule is violated here, but there is a good reason to violate it (as in constructing this object), we should disable it locally, but not forcefully deconstruct our methods.

modules/storages/app/common/storages/peripherals/storage_interaction/one_drive/files_info_query.rb
```diff
@@ -44,19 +44,11 @@ def initialize(storage)
...
           def call(auth_strategy:, file_ids:)
-            if file_ids.nil?
-              return ServiceResult.failure(
-                result: :error,
-                errors: ::Storages::StorageError.new(code: :error, log_message: "File IDs can not be nil")
-              )
-            end
-
-            result = file_ids.map do |file_id|
+            result = Array(file_ids).map do |file_id|
```
Kharonus, line 54: 🔴 this now has a changed behavior: if `file_ids` is nil, it now returns success with an empty result. Why the change?

modules/storages/app/models/storages/storage_error_data.rb
```diff
@@ -29,12 +29,7 @@
...
 
 module Storages
-  class StorageErrorData
-    attr_reader :source, :payload
-
-    def initialize(source:, payload: nil)
-      @source = source
-      @payload = payload
-    end
+  StorageErrorData = Data.define(:source, :payload) do
```
Kharonus, line 32: 🟢 nice!

Kharonus (changes requested):
me being nitpicky ;)

modules/storages/app/common/storages/peripherals/storage_interaction/one_drive/files_info_query.rb
mereghost: Oh, crap. I forgot to revert that. xD

modules/storages/app/common/storages/peripherals/storage_interaction/one_drive/file_info_query.rb
mereghost: Makes sense, I'll make the change

modules/storages/app/common/storages/peripherals/storage_interaction/one_drive/file_info_query.rb
mereghost: It started as I was trying to figure if it was worth the effort of building a better way to transform the JSON, but this was a leftover. =/

Kharonus (approved):
LGTM, now for kabiru to dive into that :P

akabiru (approved):
:shipit: ✨

PR #15383, opened 2024-04-25
Feature/54464 snappier progress popover

ulferts:
### TODO

* [x] reduced the debounce time to 100ms
* [x] ~~applied `data-turbo-permanent` attribute on focused field in the progress popover and prevented a DOM element with such an attribute from being updated. This prevents inputs from being eaten up. 
Using an attribute is in line with how Turbo does it so once the Turbo-bug on js triggered frame updates is fixed, it should still work out of the box.~~
That path was abandoned again as idiomorph has a flag to be passed in when morphing `ignoreActiveValue` which, while documentation is sparse, seems to provide the behaviour we want out of the box. 
* [x] removed the duplicate event listener on `turbo:before-frame-render` from setup.ts. It is only necessary for js triggered updates in the preview controller and not on a general level.
* [x] replaced morphdom by idiomorph as that library is used by Turbo anyway.

-----

https://community.openproject.org/wp/54464


frontend/src/stimulus/controllers/dynamic/work-packages/progress/preview-progress.controller.ts
```diff
@@ -48,15 +48,15 @@ export default class PreviewProgressController extends Controller {
...
+    this.debouncedPreview = debounce((event:Event) => { void this.preview(event); }, 100);
     // TODO: Ideally morphing in this single controller should not be necessary.
     // Turbo supports morphing, by adding the <turbo-frame refresh="morph"> attribute.
     // However, it has a bug, and it doesn't morphs when reloading the frame via javascript.
     // See https://github.com/hotwired/turbo/issues/1161 . Once the issue is solved, we can remove
     // this code and just use <turbo-frame refresh="morph"> instead.
     this.frameMorphRenderer = (event:CustomEvent<TurboBeforeFrameRenderEventDetail>) => {
       event.detail.render = (currentElement:HTMLElement, newElement:HTMLElement) => {
-        morphdom(currentElement, newElement, { childrenOnly: true });
+        Idiomorph.morph(currentElement, newElement, { ignoreActiveValue: true });
```
aaron-contreras, line 59: Well well well... Seems they've thought of this already 😆

aaron-contreras (approved):
Very nice @ulferts !

PR #15394, opened 2024-04-26
fix: Openproject / Open Project -> OpenProject

AaronDewes:
This fixes the spelling of OpenProject in a few places

### DIFF
--- a/config/locales/crowdin/af.yml
+++ b/config/locales/crowdin/af.yml
@@ -3573,7 +3573,7 @@ af:
   link: link
   plugin_openproject_auth_plugins:
     name: "OpenProject Auth Plugins"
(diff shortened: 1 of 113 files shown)


ulferts (changes requested):
Hi @AaronDewes thanks for the PR. In a few places, the string "Openproject" got simply removed by it. I guess the search/replace when wrong there.

AaronDewes:
Sorry :man_facepalming:, I'll fix that.

ulferts (approved):
Works now, @AaronDewes.

What will happen though is that the translations in /config/locales/crowdin will be overwritten from the [translation work happening on Crowdin](https://crowdin.com/project/openproject). But the change in the English locale will be reflected on that platform so it can be hoped for that the translators there will pick up the change as well.

PR #15399, opened 2024-04-26
Clean up application loading paths

dombesz:
- Remove unnecessary path eager loading from OpenProject::Reporting engine


oliverguenther (approved):
@dombesz The PR is still marked as a draft, I guess that's by accident?

PR #15400, opened 2024-04-27
[#54459] Add `AvailableProjectAttributesFilter` to project queries to display the mapped projects list in projects attribute settings view

akabiru:
https://community.openproject.org/work_packages/54459

### Wat?

Project Attributes are `ProjectCustomField`s, when they are activated within a project a mapping is created in the join table `ProjectCustomFieldProjectMapping`

### How

Add a new filter that runs an inclusion search against `ProjectCustomFieldProjectMapping` join table, returning a list of projects where the given _custom_field_ids_ values match

Example
```ruby
query = Queries::Projects::ProjectQuery.new(name: "test") do |query|
  query.where(:available_project_attributes, '=', [13, 14])
end
query.results
```


app/models/project.rb
```diff
@@ -171,6 +171,10 @@ class Project < ApplicationRecord
   scope :with_member, ->(user = User.current) { where(id: user.memberships.select(:project_id)) }
   scope :without_member, ->(user = User.current) { where.not(id: user.memberships.select(:project_id)) }
 
+  scope :with_available_custom_fields, ->(custom_field_ids) do
```
akabiru: No unit test exists yet- would like to investigate using `Queries::Filters::Base#scope` to build the query entirely without touching the model.

At present, I'm not sure if that's at all useful apart from reducing the footprint of this work.

app/models/queries/filters/strategies/list_contains.rb
```diff
@@ -0,0 +1,34 @@
...
+#
+# See COPYRIGHT and LICENSE files for more details.
+#++
+
+module Queries::Filters::Strategies
+  class ListContains < List
+    self.supported_operators = ["~"]
+    self.default_operator = "~"
+  end
+end
```
akabiru: Oddly enough, this did not exist or I'm missing some context- so please shout if I should be approaching this differently.

app/models/project.rb
mereghost: I think it is fine if we intend to use this scope outside of the queries, otherwise it would be better to try and implement it as part of the filter itself.

ulferts: I think generally doing it in scopes as long as it is not too specific (whatever that means) is a good approach to increase discoverability. So even if we don't use a scope now outside of a feature, it might later on be used because somebody discovered its existence.

What I try to do is move as much as possible from the models out into separate files for the sake of structure. To that end, there are the [scope files](https://github.com/opf/openproject/tree/dev/app/models/projects/scopes). In the end, they are just modules. Even for more complicated scopes, that tend to require extracting code into several methods, all concerns for a scope then end up in the same file. And that is then mirrored by having a separate test file for it.

Regarding the actual AR produced here: It is definitely not wrong and will do the job. It however runs the risk of producing unexpected results in case the scope is used in some ways. We don't do it for projects currently but for work packages we have highly optimized the loading done in the API, which is roughly like this for projects (pseudocode):

```
ids = Project.with_filters_applied.limit(x).pluck(:id)
actual_models = Project.with_eager_loading.where(id: ids)
```

With the joins used in the way this is done here, the first line will not as expectedly return the ids of the first x work packages but rather will build the product of Project id times project_custom_field_project_mappings and limit this to x. So for my local data:

```
Project.with_available_custom_fields(ids).limit(10).pluck(:id)
=> [1, 1, 1, 1, 2, 2, 2, 2, 3, 3]
```

As I said, it will currently not cause any problems but if used in a filter context, it is safer to rely on subqueries.

All that hopefully changes if ever I get to finish https://community.openproject.org/wp/53379.

app/models/queries/filters/strategies/list_contains.rb
ulferts: If you use the `List` strategy, it will have to operators: `=` and `!`. In the context of the list, the `=` operator maps to [Queries::Operators::EqualsOr which is actually Queries::Operators::Equals](https://github.com/opf/openproject/blob/dev/app/models/queries/operators/equals.rb#L45-L46). That operator already matches on a subset which I think is what you are trying to achieve.

That is also present in the frontend where there is a distinction between "is (or)" and "is (and)":
image

app/models/queries/projects/filters/available_project_attributes_filter.rb
```diff
@@ -0,0 +1,56 @@
...
+  def self.key
+    :available_project_attributes
+  end
+
+  def type
+    :list_contains
+  end
+
+  def allowed_values
+    @allowed_values ||= ProjectCustomFieldProjectMapping.pluck(:custom_field_id).map { |id| [id, id.to_s] }
```
ulferts: The `allowed_values` method for a very long time consisted of a somewhat useless second value instead of only the id. That was because the API did not make use of the second value. But with the move to Hotwire, I recon that this is about to change. The second value is the label to be displayed. So if ever the filter is to be used in the UI, it should receive a human readable label, e.g. `ProjectCustomFieldProjectMapping.includes(:project_custom_field).pluck(:custom_field_id, :name)`

spec/support/queries/filters/shared_filter_examples.rb
```diff
@@ -370,6 +370,54 @@
...
+      it "is the same as handwriting the query" do
+        handwritten_scope_sql = <<-SQL.squish
+          SELECT "projects".* FROM "projects"
+            INNER JOIN "project_custom_field_project_mappings" ON "project_custom_field_project_mappings"."project_id" = "projects"."id"
+            WHERE "project_custom_field_project_mappings"."custom_field_id" IN (#{values.join(', ')})
+        SQL
+
+        expect(instance.scope.to_sql).to eql handwritten_scope_sql
+      end
+    end
```
ulferts: I have in the past written those SQL based expectations myself. Nowadays I consider them too brittle as e.g. the order might change between AR versions without the behaviour changing. 

The spec here is also specific for the `project_custom_field_project_mappings`

ulferts (changes requested):
Nicely worked through the whole mass of code that are the queries. It is quite something to stomach. 

I added some remarks. Not all of them are actually must haves so let's talk when you are unsure.

app/models/queries/filters/strategies/list_contains.rb
Kharonus: I stumbled over the exact same thing. It is true, what Jens says, we are using `=` as a `contains` operator on lists. You could understand it as the values are an array of values that can match, concatenated with `OR`. So, a project matches the filter, if it has ANY of the project custom fields assigned, that where passed into the filter as an array.

🔴 -> the filter type should be standard `:list`

There is on difference to e.g. the `user_action_filter`, where a project must match ALL the values given to the filter, so basically an `is (AND)`. This would be filter type `:list_all`.

Kharonus: Update: I do understand that correctly, yes? This filter is supposed to be `is (OR)`?

app/models/queries/projects/filters/available_project_attributes_filter.rb
```diff
@@ -0,0 +1,56 @@
...
+# along with this program; if not, write to the Free Software
+# Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+#
+# See COPYRIGHT and LICENSE files for more details.
+# ++
+#
+
+class Queries::Projects::Filters::AvailableProjectAttributesFilter < Queries::Projects::Filters::ProjectFilter
+  def self.key
+    :available_project_attributes
```
Kharonus, line 34: 🟡 The documentation of this filter in `docs/api/apiv3/paths/projects.yml` is missing. The name is okay, IMHO.

Kharonus (changes requested):

app/models/queries/filters/strategies/list_contains.rb
ulferts: I thinks that would be consistent which is why I would use it, yes.

app/models/queries/projects/filters/available_project_attributes_filter.rb
akabiru: Thanks! Turns out it's the other way round: the first value is the label to be displayed? As the unit tests were able to inform me.

https://github.com/opf/openproject/blob/dev/app/models/queries/filters/strategies/list.rb#L54-L56

spec/support/queries/filters/shared_filter_examples.rb
akabiru: Agree they could be a bit brittle, the value I see is the integration level validation at the query level. Perhaps Ok to retain this convention see how it plays out?

app/models/projects/scopes/available_custom_fields.rb
```diff
@@ -0,0 +1,53 @@
...
+        subquery = project_custom_fields_project_mapping_subquery(custom_field_ids:)
+        where.not(id: subquery)
+      end
+
+      private
+
+      def project_custom_fields_project_mapping_subquery(custom_field_ids:)
+        ProjectCustomFieldProjectMapping.select(:project_id)
+                          .where(custom_field_id: custom_field_ids)
+                          .distinct
```
ulferts: Not an issue, but what is the `distinct` for? As this is used in a subquery, duplicate results will not matter. It might (and here I am guessing) be slower to perform the distinct in the subselect and might (again guessing) prevent indices from being used.

ulferts (approved):
Looks good to me, @akabiru. 

I have one remark but that is based on guesswork so feel free to ignore it.

spec/support/queries/filters/shared_filter_examples.rb
ulferts: Fine with me.

app/models/queries/projects/filters/available_project_attributes_filter.rb
ulferts: Ah, thanks for the clarification.

app/models/projects/scopes/available_custom_fields.rb
akabiru: The reason ~is~ _was_ to remove the duplicate resulting `project_id`s so that the subquery covers a wider scope but I now see that assumption is incorrect as there is no "limit" within the subquery. Will remove it, thanks for catching that! 👍🏾 

> (and here I am guessing) be slower to perform the distinct in the subselect and might (again guessing) prevent indices from being used.

Interesting, I didn't know this was a possibility 🎓

akabiru: > The reason ~is~ _was_ to remove the duplicate resulting `project_id`s so that the subquery covers a wider scope but I now see that assumption is incorrect as there is no "limit" within the subquery. Will remove it, thanks for catching that! 👍🏾
> 
> > (and here I am guessing) be slower to perform the distinct in the subselect and might (again guessing) prevent indices from being used.
> 
> Interesting, I didn't know this was a possibility 🎓

The indexes are still used AFAICT, just the perf load from distinct application

```ruby
[6] pry(main)> ProjectCustomFieldProjectMapping.select(:project_id).where(custom_field_id: [14,13]).pluck(:project_id)
  ProjectCustomFieldProjectMapping Pluck (0.6ms)  SELECT "project_custom_field_project_mappings"."project_id" FROM "project_custom_field_project_mappings" WHERE "project_custom_field_project_mappings"."custom_field_id" IN ($1, $2)  [["custom_field_id", 14], ["custom_field_id", 13]]
=> [3, 4, 4, 9, 9, 3]
[9] pry(main)> ProjectCustomFieldProjectMapping.select(:project_id).where(custom_field_id: [14,13]).explain(:analyze)
  ProjectCustomFieldProjectMapping Load (0.5ms)  SELECT "project_custom_field_project_mappings"."project_id" FROM "project_custom_field_project_mappings" WHERE "project_custom_field_project_mappings"."custom_field_id" IN ($1, $2)  [["custom_field_id", 14], ["custom_field_id", 13]]
=> EXPLAIN (ANALYZE) SELECT "project_custom_field_project_mappings"."project_id" FROM "project_custom_field_project_mappings" WHERE "project_custom_field_project_mappings"."custom_field_id" IN ($1, $2) [["custom_field_id", 14], ["custom_field_id", 13]]
                                                                          QUERY PLAN
--------------------------------------------------------------------------------------------------------------------------------------------------------------
 Bitmap Heap Scan on project_custom_field_project_mappings  (cost=8.40..19.08 rows=12 width=8) (actual time=0.011..0.012 rows=6 loops=1)
   Recheck Cond: (custom_field_id = ANY ('{14,13}'::bigint[]))
   Heap Blocks: exact=1
   ->  Bitmap Index Scan on index_project_cf_project_mappings_on_custom_field_id  (cost=0.00..8.39 rows=12 width=0) (actual time=0.007..0.007 rows=7 loops=1)
         Index Cond: (custom_field_id = ANY ('{14,13}'::bigint[]))
 Planning Time: 0.067 ms
 Execution Time: 0.036 ms
(7 rows)
[10] pry(main)> ProjectCustomFieldProjectMapping.select(:project_id).where(custom_field_id: [14,13]).distinct.explain(:analyze)
  ProjectCustomFieldProjectMapping Load (2.6ms)  SELECT DISTINCT "project_custom_field_project_mappings"."project_id" FROM "project_custom_field_project_mappings" WHERE "project_custom_field_project_mappings"."custom_field_id" IN ($1, $2)  [["custom_field_id", 14], ["custom_field_id", 13]]
=> EXPLAIN (ANALYZE) SELECT DISTINCT "project_custom_field_project_mappings"."project_id" FROM "project_custom_field_project_mappings" WHERE "project_custom_field_project_mappings"."custom_field_id" IN ($1, $2) [["custom_field_id", 14], ["custom_field_id", 13]]
                                                                                QUERY PLAN
--------------------------------------------------------------------------------------------------------------------------------------------------------------------------
 Unique  (cost=19.30..19.36 rows=12 width=8) (actual time=0.202..0.205 rows=3 loops=1)
   ->  Sort  (cost=19.30..19.33 rows=12 width=8) (actual time=0.201..0.201 rows=6 loops=1)
         Sort Key: project_id
         Sort Method: quicksort  Memory: 25kB
         ->  Bitmap Heap Scan on project_custom_field_project_mappings  (cost=8.40..19.08 rows=12 width=8) (actual time=0.013..0.162 rows=6 loops=1)
               Recheck Cond: (custom_field_id = ANY ('{14,13}'::bigint[]))
               Heap Blocks: exact=1
               ->  Bitmap Index Scan on index_project_cf_project_mappings_on_custom_field_id  (cost=0.00..8.39 rows=12 width=0) (actual time=0.008..0.008 rows=7 loops=1)
                     Index Cond: (custom_field_id = ANY ('{14,13}'::bigint[]))
 Planning Time: 0.078 ms
 Execution Time: 0.231 ms
(11 rows)
```

PR #15407, opened 2024-04-29
[#54355] Use authentication in FolderFilesFileIdsDeepQuery

Kharonus:
[#54355](https://community.openproject.org/work_packages/54355)

### WAT?

- harmonized interface between nextcloud and one_drive
- path key of nextcloud folders no longer end with trailing slash
- added vcr cassettes for nextcloud
- added shared test examples for both provider types

### DIFF
--- a/modules/storages/app/common/storages/peripherals/storage_interaction/nextcloud/folder_files_file_ids_deep_query.rb
+++ b/modules/storages/app/common/storages/peripherals/storage_interaction/nextcloud/folder_files_file_ids_deep_query.rb
@@ -28,19 +28,44 @@
 # See COPYRIGHT and LICENSE files for more details.
 #++
 
-module Storages::Peripherals::StorageInteraction::Nextcloud
-  class FolderFilesFileIdsDeepQuery
-    def self.call(storage:, folder:)
-      ::Storages::Peripherals::Registry
-        .resolve("nextcloud.queries.propfind")
-        .call(
-          storage:,
-          depth: "infinity",
-          path: folder.path,
-          # nc:acl-list is only required to avoid https://community.openproject.org/wp/49628. See comment #4.
-          props: %w[oc:fileid nc:acl-list]
-        ).map do |obj|
-        obj.transform_values { |value| Storages::StorageFileInfo.from_id(value["fileid"]) }
+module Storages
+  module Peripherals
+    module StorageInteraction
+      module Nextcloud
+        class FolderFilesFileIdsDeepQuery
+          def self.call(storage:, auth_strategy:, folder:)
+            new(storage).call(auth_strategy:, folder:)
+          end
+
+          def initialize(storage)
+            @storage = storage
+            @propfind_query = Internal::PropfindQuery.new(storage)
+          end
+
+          def call(auth_strategy:, folder:)
+            origin_user_id = Util.origin_user_id(caller: self.class, storage: @storage, auth_strategy:)
+            if origin_user_id.failure?
+              return origin_user_id
+            end
+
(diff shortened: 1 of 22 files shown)


modules/storages/app/common/storages/peripherals/storage_interaction/nextcloud/internal/propfind_query_legacy.rb
Kharonus, line 1: This is the old internal propfind query, as it would be out of scope to change `file_ids_query`, too, which consumes this internal query.

modules/storages/app/common/storages/peripherals/storage_interaction/nextcloud/folder_files_file_ids_deep_query.rb
```diff
@@ -28,19 +28,44 @@
...
-          path: folder.path,
-          # nc:acl-list is only required to avoid https://community.openproject.org/wp/49628. See comment #4.
-          props: %w[oc:fileid nc:acl-list]
-        ).map do |obj|
-        obj.transform_values { |value| Storages::StorageFileInfo.from_id(value["fileid"]) }
+module Storages
+  module Peripherals
+    module StorageInteraction
+      module Nextcloud
+        class FolderFilesFileIdsDeepQuery
```
mereghost, line 35: 🟢 The name of query is... weird, but I don't have a good suggestion to what to change it to. It returns a dictionary of full remote file path and ids, and operates recursively given a start point.

The recursive part is kind of a detail since we don't specify a maximum depth. `FilePathToIdMapQuery`? `WeirdThingWeNeedQuery`?

modules/storages/app/common/storages/peripherals/storage_interaction/nextcloud/folder_files_file_ids_deep_query.rb
```diff
@@ -28,19 +28,44 @@
...
+          def initialize(storage)
+            @storage = storage
+            @propfind_query = Internal::PropfindQuery.new(storage)
+          end
+
+          def call(auth_strategy:, folder:)
+            origin_user_id = Util.origin_user_id(caller: self.class, storage: @storage, auth_strategy:)
+            if origin_user_id.failure?
+              return origin_user_id
+            end
```
mereghost: 🟡 Is using the `ServiceResult#on_failure` feasible? 

```ruby
origin_user_id = Util.origin_user_id.on_failure {|result| return result }

# origin_user_id can be used as it will always be a successful result.
```

mereghost (approved):
Looks good. Just left a couple of nitpicks, but shouldn't block a merge.

modules/storages/app/common/storages/peripherals/storage_interaction/nextcloud/folder_files_file_ids_deep_query.rb
Kharonus: `FilePathToIdMapQuery` doesn't sound too weird. Should I just issue a small renaming PR after this one?

mereghost: Sure, why not.

modules/storages/app/common/storages/peripherals/storage_interaction/nextcloud/folder_files_file_ids_deep_query.rb
Kharonus: sure thing, I'm just not yet used to use it. ;) changed

PR #15410, opened 2024-04-29
Attempt to fix flaky specs where the toast might disappear before the page finishes loading

dombesz:
Let's consider the following scenario:

We are saving a form by clicking the submit button. The page redirects, the toast message is shown, but the page still loads some requests in the background. Capybara will not consider the page submission with the "click_on" command done, unless the page is loaded. Hence it will not go to the next expectation unless the page is loaded. If the page takes too long to load, the toast message might already disappear by the time capybara runs the next expectation for the toast's presence. This will make the expectation fail.

Attempt to fix flaky spec on [./modules/budgets/spec/features/budgets/add_budget_spec.rb:59](https://github.com/opf/openproject/actions/runs/8873891332/job/24360406463?pr=15401) 

Some specs got broken alongside while applying the fix. Those cases were prone to toast timing and fixed those too.

### DIFF
--- a/modules/boards/spec/features/support/board_index_page.rb
+++ b/modules/boards/spec/features/support/board_index_page.rb
@@ -72,6 +72,8 @@ module Pages
       new_board_page.set_board_type action
       new_board_page.click_on_submit
 
+      new_board_page.expect_and_dismiss_toaster(message: "Successful creation.")
+
       if expect_empty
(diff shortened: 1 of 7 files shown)


ulferts (approved):
Only time will tell in the end but the reasoning sound compelling, @dombesz. Once again thanks for digging into this.

PR #15280, opened 2024-04-16
Bug/52098 misleading error message when ifc upload file size exceeds `Setting.attachment_max_size`

akabiru:
#### https://community.openproject.org/work_packages/52098

_Before_

image

_After_

Screenshot 2024-04-16 at 6 18 34 PM

PR #15411, opened 2024-04-29
fix[Op#52098]: Invalidate form validity when direct upload is too large

akabiru:
### https://community.openproject.org/wp/52098

Follows https://github.com/opf/openproject/pull/15280

On direct upload, client side JQuery overrides the form submission to the _"direct"_ storage provider. This PR intercepts this submission by attaching form validity via [`HTMLObjectElement.setCustomValidity()`](https://developer.mozilla.org/en-US/docs/Web/API/HTMLObjectElement/setCustomValidity)

See: https://developer.mozilla.org/en-US/docs/Web/API/HTMLObjectElement/setCustomValidity

- [x] add feature test
- [x] add request test
- [ ] Fix failing test

https://github.com/opf/openproject/assets/17295175/812831a9-44b7-4cce-955d-b7e4886e8c43


Kharonus (approved):
I love it, simple and direct solution for this bug.

Eventually we will get rid of the jQuery here, maybe when we rework that whole form to primer.

PR #15416, opened 2024-04-29
Rubocop check

toy:
Enable rubocop check using reviewdog action that will comment on all PRs. It was silently failing, as not all dependencies were enabled.

~To not overwhelm with failures a todo file for all failures is included plus a separate file for disable directives (it is not generated automatically). The goal is to start checking all new changes and gradually empty the todo files.~

#15557 as an example of how many violations could there be, so removed todos (#15561 with the branch containing them)

### DIFF
--- a/.github/workflows/rubocop-core.yml
+++ b/.github/workflows/rubocop-core.yml
@@ -18,9 +18,10 @@ jobs:
           # fetch head commit of base branch
           git fetch --deepen 1 origin "${{ github.ref }}"
       - uses: ruby/setup-ruby@v1
-      - uses: opf/action-rubocop@v2
+      - uses: opf/action-rubocop@master
         with:
           github_token: ${{ secrets.github_token }}
           rubocop_version: gemfile
           rubocop_extensions: rubocop-inflector:gemfile rubocop-performance:gemfile rubocop-rails:gemfile rubocop-rspec:gemfile
           reporter: github-pr-check
+          only_changed: true
--- a/.github/workflows/rubocop-core.yml
+++ b/.github/workflows/rubocop-core.yml
@@ -16,5 +16,5 @@ jobs:
         with:
           github_token: ${{ secrets.github_token }}
           rubocop_version: gemfile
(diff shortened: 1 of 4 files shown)


cbliard:
Why is a todo file needed if it runs only on modified files?

toy:
@cbliard Mostly to not force resolving bigger issues (like complexity) even when doing small changes right away. I agree that as reviewdog will even reduce comments only to changed lines, it may be not really needed.

cbliard:
What I do not like when working with todo like this is that the autocorrect "randomly" does not work as some files are excluded from some rules. I do not really see how it can help us write better code if everything is hidden.

Besides, I do not understand why some are disabled. It was generated automatically but still. For instance this part:
```
FactoryBot/AssociationStyle:
  Exclude:
    - 'modules/backlogs/spec/factories/impediment_factory.rb'
    - 'modules/backlogs/spec/factories/task_factory.rb'
    - 'modules/budgets/spec/factories/budget_factory.rb'
    - ...
```

I ran `rubocop --only FactoryBot/AssociationStyle modules/backlogs/spec/factories/impediment_factory.rb`  just to see what the problem is, but it emits no warning. I thought it was a pending cop so I added `--enable-pending-cops` but same result. And then while writing this message I realized I have to remove the line from the todo first 🙃 and then I could see the issues.

I wonder how we will manage the todo file in the future.

Meanwhile, I think it's very important to have rubocop enabled again and I am really grateful that you invested some time in it. I just have some fears about that todo file. I wonder what other people think about it.

toy:
I agree, just worried that if reintroducing rubocop will create too much additional work, there will be a desire to get rid of it. Maybe a good option is to remove the todo list and add it separately if the checker will become problematic

toy:
I checked how bad will it look if I for example [autocorrect a bunch of quotes](https://github.com/opf/openproject/pull/15557) and it doesn't look to bad to fix the found violations, so unless there will be another opinion I'm going to remove todos

cbliard:
Great! I'm all for removing the todo files, and then adding them later if we feel it's needed.

cbliard (approved):
Sounds great!

PR #15428, opened 2024-04-30
need to wait for network idle before editing value

toy:
Minimal fix, please tell if there is a better way


cbliard (approved):
Thanks for having a look at it

PR #15434, opened 2024-04-30
[54500] Refresh filters cache before building the query in job

cbliard:
https://community.openproject.org/wp/54500

If not refreshed, the only known and usable custom fields for filters and groupbys are the ones that were present when the server was started and a query was first built, because some classes are created on the fly for the custom fields. Queries using custom fields added after will fail to be found.

Even worse, in a multi-tenant environment, a cached custom field class could be found and returned, but could have been generated from the custom fields of another tenant, because the cache is global.

Calling `CostQuery::Cache.check` will check if the cache is outdated by checking the number of custom fields and the last updated timestamp of the custom fields. If the cache is outdated, the custom field `GroupBy` and `Filter` classes will be deleted and recreated.

But it would still be problematic in a multithreaded environment.

### DIFF
--- a/modules/reporting/app/workers/cost_query/export_job.rb
+++ b/modules/reporting/app/workers/cost_query/export_job.rb
@@ -18,6 +18,7 @@ class CostQuery::ExportJob < Exports::ExportJob
   private
 
   def prepare!
+    CostQuery::Cache.check
     self.query = build_query(query)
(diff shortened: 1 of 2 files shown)


machisuji (approved):
👍 Looking good, thanks for the fix and spec!

PR #15441, opened 2024-05-01
Bump request_store, revert to avoid Fiber.

Eric-Guo:
https://github.com/steveklabnik/request_store/issues/96


ba1ash (approved):
Thanks @Eric-Guo :+1:

PR #15449, opened 2024-05-02
[#54558] openDesk setup script fails

Kharonus:
[#54558](https://community.openproject.org/work_packages/54558)

### WAT?

- hardened prod code to not fall into an exception
- added test case

### INFO

The issue occurred due to the fact that the created storage in the script is incomplete (missing oauth client configuration). A code change in 14.0 did not take into account, that this code can run on incomplete storages. The code was improved, so that this case no longer creates exceptions.


mereghost (approved):
Looks good. :shipit:

PR #15450, opened 2024-05-02
[50954] Record progress in multiple units

cbliard:
https://community.openproject.org/wp/50954


aaron-contreras:
eslint is complaining about the `chronic_duration.js` file. This is taken from Gitlab's implementation and is meant to be kept inline with their `chronic_duration` gem. There's no npm package for `chronic_duration.js` as it's directly part of the gitlab source code. Would vendoring the file prevent eslint from trying to check it or is there a config to exclude this file from eslint. Probably has been done before in the core repo but unsure of how to handle it @oliverguenther

aaron-contreras:
@cbliard 'eqeqeq' eslint issue addressed more globally in https://github.com/opf/openproject/pull/15560

cbliard (commented):
I played a little with it. It responds really well. Good job!
3 things I wonder about, and probably @psatyal can help us on it:

- what do we do with text longer than the allowed size? Currently it's clipping. Do we extend the default width?
- what about the precision? When entering 1m, it can't represent it accurately with only 2 decimals. It results to "1m 12s". Not a big deal, but it's kinda surprising to enter "2h 12m" and end up with "2h 12m 0.0000000001s", or enter "2h 13m" and end up with "2h 13m 12.00000000001s". Should we switch from floating points holding nb of hours to integer holding nb of seconds?
- when omitting the "m", or before entering it, the value is interpreted as hours. It would be better if interpreted as minutes. This is especially visible when entering a value for the first time: with w -> rw, and typing each character one by one, it gives this:
    - work -> remaining work (when 100%)
    - 2 -> 2h
    - 2h -> 2h
    - 2h1 -> 3h (would have been expecting 2h 1m)
    - 2h15 -> 17h (would have been expecting 2h 15m)
    - 2h15m -> 2h 15m

app/controllers/admin/settings/working_days_and_hours_settings_controller.rb
```diff
@@ -64,6 +65,11 @@ def non_working_days_params
       non_working_days.to_h.values
     end
 
+    def duration_configuration_params
+      params.require(:settings)
+            .permit(:hours_per_day, :days_per_week, :days_per_month)
+    end
+
```
cbliard: The fact that `#non_working_days_params` and `#duration_configuration_params` are both doing `params.require` with different arguments leads to a "Unpermitted params" everytime which feels like an error on the console:
```
D, [2024-05-24T11:21:49.662215 #70386] DEBUG -- : Unpermitted parameters: :hours_per_day, :days_per_week, :days_per_month, :working_days. Context: { controller: Admin::Settings::WorkingDaysAndHoursSettingsController, action: update, request: #<ActionDispatch::Request:0x000000033fcac820>, params: {"utf8"=>"✓", "_method"=>"patch", "authenticity_token"=>"[FILTERED]", "settings"=>{"hours_per_day"=>"abcd", "days_per_week"=>"50", "days_per_month"=>"200.5", "working_days"=>["", "1", "2", "3", "4", "5"]}, "Non-working day"=>"", "controller"=>"admin/settings/working_days_and_hours_settings", "action"=>"update"} }
D, [2024-05-24T11:21:49.662389 #70386] DEBUG -- : Unpermitted parameter: :working_days. Context: { controller: Admin::Settings::WorkingDaysAndHoursSettingsController, action: update, request: #<ActionDispatch::Request:0x000000033fcac820>, params: {"utf8"=>"✓", "_method"=>"patch", "authenticity_token"=>"[FILTERED]", "settings"=>{"hours_per_day"=>"abcd", "days_per_week"=>"50", "days_per_month"=>"200.5", "working_days"=>["", "1", "2", "3", "4", "5"]}, "Non-working day"=>"", "controller"=>"admin/settings/working_days_and_hours_settings", "action"=>"update"} }
```

Actually, the permitted params are given by `settings_params` from the super class which is doing `permitted_params.settings.to_h` which uses settings definitions to get all allowed keys. The keys `:hours_per_day, :days_per_week, :days_per_month` are already part of the `settings` being returned in the parent method. So it should not be needed to call `duration_configuration_params` at all.

Yep, I just tried and it works.

There is still one logged error line, but this one was already there before.

app/contracts/settings/working_days_and_hours_params_contract.rb
```diff
@@ -0,0 +1,96 @@
...
+
+module Settings
+  class WorkingDaysAndHoursParamsContract < ::ParamsContract
+    include RequiresAdminGuard
+
+    validate :working_days_are_present
+    validate :hours_per_day_are_present
+    validate :days_per_week_are_present
+    validate :days_per_month_are_present
+    validate :days_per_week_and_days_per_month_are_consistent
```
cbliard: It does not validate enough IMHO:
- It should not be possible to set more than 24h per day, 7 days per week, or 31 days per month. It should raise an error.
- It should not be possible to enter negative values
- It should not be possible to enter zero values or text. Using text is "blocked" client side, but not checked server side. When doing so with a curl request, the value is interpreted as 0 due to some `to_i` or `to_f` here and there.
  - When hours per day value is 0, it leads to division by zero errors when converting time

app/views/admin/settings/working_days_and_hours_settings/show.html.erb
```diff
@@ -27,22 +27,51 @@ See COPYRIGHT and LICENSE files for more details.
...
       class: 'op-working-days-admin-settings'
     ) do %>
 
+  <section class="form--section"
+           data-application-target="dynamic"
+           data-controller="admin--working-days-and-hours">
+    <fieldset class="form--fieldset">
+      <div class="form--field">
+        <%= setting_text_field :hours_per_day,
+                               type: :number %>
```
cbliard: This field and the 2 others are larger than what was designed in figma. In the mockups they are only half width. You can achieve it by adding `container_class: "-wide"` argument. The css classes are defined in `frontend/src/global_styles/content/_forms.sass:579-598`

app/views/admin/settings/working_days_and_hours_settings/show.html.erb
```diff
@@ -27,22 +27,51 @@ See COPYRIGHT and LICENSE files for more details.
...
+    </fieldset>
+  </section>
+
+  <div class="op-toast -warning">
+    <div class="op-toast--content">
+      <p>
+        <%= t("working_days.warning").html_safe %>
+      </p>
+    </div>
+  </div>
```
cbliard: I feel like this warning should belong to the next section. Currently it looks like it belongs to the wrong one.
Maybe @psatyal can confirm.

app/views/layouts/base.html.erb
```diff
@@ -156,28 +156,28 @@ See COPYRIGHT and LICENSE files for more details.
   // execute these lines to prevent flickering when page loads:
 
   (function($) {
-    var wrapper = $('#wrapper');
-    var savedMainMenuWidth = window.OpenProject.guardedLocalStorage("openProject-mainMenuWidth");
-    var mainMenuCollapsed =  window.OpenProject.guardedLocalStorage("openProject-mainMenuCollapsed");
+  var wrapper = $('#wrapper');
+  var savedMainMenuWidth = window.OpenProject.guardedLocalStorage("openProject-mainMenuWidth");
+  var mainMenuCollapsed = window.OpenProject.guardedLocalStorage("openProject-mainMenuCollapsed");
```
cbliard: It looks like the indentation was all messed up in this function. Not sure why.

cbliard (commented):
I can't approve nor request changes on this PR because I created it 😅 

The multiple units display is really nice, and the way it's used in the popover is smooth and does not get in the way. That's nice!

I have some tiny remarks here and there.

I also have the feeling that the precision is annoying us. For instance:
* entering "2h 13m" leads to 7980 seconds, converted to 2.216666667 hours, rounded to 2.22 (2 decimals). This value 2.22 is the one being saved in database.
* When frontend is rendering it, it has 2 ways of rendering it:
  * the modal is rendering it through `DurationConverter.output`, which for 2.22 will render "2h 13m 12s"
  * the angular part is using a different route: the 2.22 is converted in the representer using `datetime_formatter.format_duration_from_hours`, so 2.22 is converted to iso duration "PT2H13M12S". This duration is then handled in the timezone service which converts it to hours with `Number(moment.duration(durationString).asHours().toFixed(2))` which is 2.22, converted to seconds which gives 7992.000000000001, and then it is converted to chronic duration which gives "2h 13m 12.000000000001s"

So the rounding is always false, and sometimes more false than others. I think the second one can be fixed by avoiding converting iso duration to hours and then to second: by converting iso duration directly to seconds, we could avoid precision loss.

Another thing we could do is to avoid displaying seconds entirely by making the hours to seconds conversion remove them. Something like `(hours * 3600 + 15).to_i / 60 * 60`. For 2.22 hours, it would give 7980 which converts to "2h 13m". No seconds, no errors.

WDYT?

app/views/layouts/base.html.erb
aaron-contreras: Hmmm, I think it was a Rubymine "autocorrect-on-save".

aaron-contreras: Fixed.

app/views/admin/settings/working_days_and_hours_settings/show.html.erb
aaron-contreras: You're right. Played around with it and like it more this way:

Banner position demo

aaron-contreras: Thoughts @psatyal ?

aaron-contreras: This is the commit https://github.com/opf/openproject/pull/15450/commits/adfa1bd7f6863d2f953f25f95cbe9597b514e04d

in case we agree this is where it should be

cbliard: Seen with Parimal during today's frontend meeting:
- the text has been slightly reworded
- the warning is moved on top of the section, so the warning can be read before doing the changes

image

He also asked if it's possible when clicking "Apply changes" to display the warning message about changed days only if there are actually changed days. I'll do another review to specifically ask for this so we can track it more easily.

app/views/admin/settings/working_days_and_hours_settings/show.html.erb
```diff
@@ -27,37 +27,66 @@ See COPYRIGHT and LICENSE files for more details.
...
+        <%= t("working_days.info") %>
       </p>
 
       <div class="form--field op-working-days-admin-settings--day-selectors" id="setting_working_days">
         <%= setting_multiselect :working_days,
-
-              I18n.t('date.day_names').rotate.zip(WeekDay::DAY_RANGE),
-              direction: :horizontal %>
+                                I18n.t('date.day_names').rotate.zip(WeekDay::DAY_RANGE),
+                                direction: :horizontal %>
```
cbliard, line 88: When clicking "Apply changes", there is this "Change working days" information dialog that always opens itself to warn about changing working days.

image


Parimal would like it to be displayed only if the working days have been actually changed.

Would it be possible?

cbliard (commented):
An additional request from Parimal while discussing this screen during today's frontend meeting

cbliard (commented):
I can't approve because I created the PR, but I approve it 👍

app/views/admin/settings/working_days_and_hours_settings/show.html.erb
cbliard: I'll create a bug work package for that one.

PR #15451, opened 2024-05-02
Implement deploy status checks for github integration

machisuji:
WP [55425](https://community.openproject.org/projects/openproject/work_packages/55425/activity)

This introduces a new state for pull requests called `'deployed'`. 
A cron job checks the deploy status of every pull request regularly and updates the state accordingly once deployed.
It will also add a comment to each relevant work package to let users know where the pull request was deployed to.

image

To make this work an `introspection` permission is introduced and the APIv3 root endpoint amended with the deployed core SHA that is running.

Lastly, next to the `deploy_targets` table another table `github_deploy_status_checks` is created used to make sure the deploy status is only checked as often as necessary and not more.

**To do**
- [x] implement deploy status job doing the actual work
- [x] add `deployed` state in frontend
- [x] add UI to allow configuring deploy targets
- [x] add tests 

**Remarks**
- it would be nice if it was the github user who comments on the work packages with the deploy status updates, however, outside of the pull request webhook request cycle we simply don't know who the github user is and I didn't want to add any more code than necessary for the time being

### DIFF
--- a/modules/github_integration/app/models/deploy_target.rb
+++ b/modules/github_integration/app/models/deploy_target.rb
@@ -34,7 +34,7 @@ class DeployTarget < ApplicationRecord
   has_many :deploy_status_checks, dependent: :destroy
 
   validates_presence_of :host
-
+  validates_presence_of :type
   # this is very much specific to the only type of target we support for now, OpenProject
   validates_presence_of :api_key
 end
--- a/modules/github_integration/app/models/deploy_target.rb
+++ b/modules/github_integration/app/models/deploy_target.rb
@@ -34,7 +34,10 @@ class DeployTarget < ApplicationRecord
   has_many :deploy_status_checks, dependent: :destroy
 
   validates_presence_of :host
+  validates_uniqueness_of :host
+
   validates_presence_of :type
+
   # this is very much specific to the only type of target we support for now, OpenProject
   validates_presence_of :api_key
 end
--- a/modules/github_integration/app/models/deploy_target.rb
+++ b/modules/github_integration/app/models/deploy_target.rb
@@ -26,23 +26,15 @@
 # See COPYRIGHT and LICENSE files for more details.
 #++
 
-en:
-  button_add_deploy_target: Add deploy target
-  label_deploy_target: Deploy target
-  label_deploy_target_new: New deploy target
-  label_deploy_target_plural: Deploy targets
-  label_github_integration: GitHub Integration
-  notice_deploy_target_created: Deploy target created
-  notice_deploy_target_destroyed: Deploy target deleted
-  plugin_openproject_github_integration:
-    name: "OpenProject GitHub Integration"
-    description: "Integrates OpenProject and GitHub for a better workflow"
+class DeployTarget < ApplicationRecord
+  self.inheritance_column = nil
 
-  project_module_github: "GitHub"
-  permission_show_github_content: "Show GitHub content"
-  permission_introspection: Read running OpenProject core version and build SHA
-  text_deploy_target_type_info: >
-    So far we only support OpenProject itself.
-  text_deploy_target_api_key_info: >
-    An OpenProject [API key](docs_url)
-    belonging to a user who has the global introspection permission.
+  store_accessor :options, :api_key
+
+  has_many :deploy_status_checks, dependent: :destroy
+
+  validates_presence_of :host
+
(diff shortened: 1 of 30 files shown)


modules/github_integration/app/models/deploy_target.rb
```diff
@@ -0,0 +1,40 @@
...
+
+class DeployTarget < ApplicationRecord
+  self.inheritance_column = nil
+
+  store_accessor :options, :api_key
+
+  has_many :deploy_status_checks, dependent: :destroy
+
+  validates_presence_of :host
+
```
dombesz, line 40: Since the migration has `t.text :type, null: false` I think we should add a presence validator for `type` too.
```suggestion
  validates_presence_of :type
```

modules/github_integration/app/workers/cron/check_deploy_status_job.rb
```diff
@@ -0,0 +1,183 @@
...
+        )
+      end
+    end
+
+    def delete_status_check(pull_request, deploy_target)
+      # we use `find` and delete it this way to also cover
+      # not-yet-persisted records
+      check = pull_request
+        .deploy_status_checks
+        .find { |c| c.deploy_target == deploy_target }
```
dombesz: Since we can have multiple `deploy_status_checks` with the same `deploy_target`, I would use `select` instead of `find` to remove all of them.
```suggestion
        .select { |c| c.deploy_target == deploy_target }
```

modules/github_integration/config/routes.rb
```diff
@@ -0,0 +1,31 @@
...
+#
+# You should have received a copy of the GNU General Public License
+# along with this program; if not, write to the Free Software
+# Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+#
+# See COPYRIGHT and LICENSE files for more details.
+#++
+
+Rails.application.routes.draw do
+  resources :deploy_targets, only: %i[index new edit update create destroy]
```
dombesz: Just a nitpick, but we don't seem to have edit and update actions.
```suggestion
  resources :deploy_targets, only: %i[index new create destroy]
```

modules/github_integration/app/models/deploy_target.rb
```diff
@@ -0,0 +1,40 @@
...
+#++
+
+class DeployTarget < ApplicationRecord
+  self.inheritance_column = nil
+
+  store_accessor :options, :api_key
+
+  has_many :deploy_status_checks, dependent: :destroy
+
+  validates_presence_of :host
```
dombesz, line 36: Since we have a [unique index here](https://github.com/opf/openproject/pull/15451/files#diff-19c6210cb5d9070ea7e81240ef4088c0544d4eda3b34e24cec25002504e5f880R11), should we also have a uniqueness validator?

```suggestion
  validates_presence_of :host
  validates_uniqueness_of :host
```

modules/github_integration/app/workers/cron/check_deploy_status_job.rb
```diff
@@ -0,0 +1,183 @@
...
+    def commit_contains?(base_commit_sha, merge_commit_sha)
+      data = compare_commits base_commit_sha, merge_commit_sha
+
+      return false if data.nil?
+
+      status = data["status"].presence
+      ahead_by = data["ahead_by"].presence
+      behind_by = data["behind_by"].presence
+
+      status == "behind" && ahead_by == 0 && (behind_by.present? && behind_by > 0)
```
dombesz, line 160: Just a minor suggestion, when the `base_commit_sha` and the `merge_commit_sha` are the same, github returns "identical", probably we should return true from this method when that is the case.

modules/github_integration/app/models/deploy_target.rb
dombesz: In a future version probably we should validate the host format, ie: to not allow the protocol to be present.

dombesz (approved):
Well done @machisuji , nice feature! 👏
I left a few minor comments, but nothing important that can block the approval. Feel free to address them now or later and please merge at will.

modules/github_integration/app/workers/cron/check_deploy_status_job.rb
machisuji: Good shout! 👍

PR #15452, opened 2024-05-02
[chore] Add proper api spec schema for Available_AssigneesModel

sbu-WBT:
I took attachments_model.yml as a template; hopefully everything's alright


ulferts (approved):
Looks good @sbu-WBT. Thanks for taking the time.

It would be nice to also have an example but it is definitely better to get the specification correct than to have examples.

sbu-WBT:
Do you know in which collection is currently an example? I will add it next time

PR #15460, opened 2024-05-03
[#53672] Hide attachments in Files tab

ba1ash:
https://community.openproject.org/work_packages/53672


db/migrate/20240424160513_add_hide_attachments_to_projects.rb
```diff
@@ -26,17 +26,10 @@
...
-    include ApplicationHelper
-
-    def breadcrumb_items
-      [{ href: admin_index_path, text: t("label_administration") },
-       { href: admin_settings_attachments_path, text: t("attributes.attachments") },
-       t("antivirus_scan.quarantined_attachments.title")]
-    end
+class AddHideAttachmentsToProjects < ActiveRecord::Migration[7.1]
+  def change
+    add_column :projects, :hide_attachments, :boolean, default: false
```
Kharonus: 🟡 So, we do not have something like a settings table for projects? Feels odd to put this boolean into the project model. What were your thoughts, @ulferts ?

lib/api/v3/work_packages/schema/work_package_schema_representer.rb
```diff
@@ -91,6 +91,11 @@ def initialize(schema, self_link:, **context)
             { href: @base_schema_link } if @base_schema_link
           end
 
+          link :attachments do
+            next if represented.work_package.hide_attachments?
+            { href: nil }
```
Kharonus, line 96: 🟡 is setting a `nil` link here better then setting a schema property?

frontend/src/app/features/work-packages/components/wp-tabs/services/wp-tabs/wp-files-count.function.ts
```diff
@@ -43,7 +43,7 @@ export function workPackageFilesCount(
 
   return combineLatest(
     [
-      attachmentService.collection(workPackage.$links.attachments.href || ''),
+      workPackage.$links.attachments ? attachmentService.collection(workPackage.$links.attachments.href || '') : new BehaviorSubject([]),
```
Kharonus: 🟡 why a BehaviorSubject? And, please lets extract that into a variable to keep the combineLatest readable.

Kharonus (changes requested):

lib/api/v3/work_packages/schema/work_package_schema_representer.rb
Kharonus: 🟢 I just read your commit message explaining it. Makes sense, I guess.

db/migrate/20240424160513_add_hide_attachments_to_projects.rb
ba1ash: Yes, it seem like we do not have project settings table or something similar. In the current database structure projects table seems to be right place for this boolean. I do not see it as a problem at all. If you have a better place for it feel free to suggest it.

lib/api/v3/work_packages/schema/work_package_schema_representer.rb
ba1ash: Yes, I think both options are not ideal, but using `schema` caused issues.

frontend/src/app/features/work-packages/components/wp-tabs/services/wp-tabs/wp-files-count.function.ts
ba1ash: Extracted.
I assume I can't use plain empty array with combineLatest and I need an observable to emit it. So, I used BehaviorSubject to do it. I am not proficient with RxJs and there can be complete misunderstanding from my side. So, feel free to correct me.

db/migrate/20240424160513_add_hide_attachments_to_projects.rb
Kharonus: I talked a lot with Wieland, Oliver and Jens, and we agreed to use a `settings` column, which is of type jsonb. Accessible with `store_attribute`.

I also updated the behavior of the `work_package.hide_attachments?` method. Only fall back to global setting, if setting is unset.

frontend/src/app/features/work-packages/components/wp-tabs/services/wp-tabs/wp-files-count.function.ts
Kharonus: usually you use `of([])` to create a simple observable of an empty array. I updated this line.

Kharonus (approved):

PR #15461, opened 2024-05-03
[54646] Do not set totals of non-parent work packages on status changes

cbliard:
https://community.openproject.org/wp/54646

Childless work packages should not have any totals being set. This PR adds a migration to fix it, and modifies a background job which was mistakingly computing and setting totals for all work packages, regardless of the work package being a parent or not.

### DIFF
--- a/app/workers/work_packages/progress/job.rb
+++ b/app/workers/work_packages/progress/job.rb
@@ -26,37 +26,17 @@
 # See COPYRIGHT and LICENSE files for more details.
 #++
 
-class Settings::UpdateService < BaseServices::BaseContracted
-  def initialize(user:)
-    super(user:,
-          contract_class: Settings::UpdateContract)
-  end
+# Base class for jobs updating work packages progress values.
+#
+# Handles concurrency so that only one progress job is performed at a time.
+class WorkPackages::Progress::Job < ApplicationJob
+  include GoodJob::ActiveJobExtensions::Concurrency
+  include WorkPackages::Progress::SqlCommands
 
-  def persist(call)
-    params.each do |name, value|
-      set_setting_value(name, value)
-    end
-    call
-  end
+  queue_with_priority :default
 
-  private
-
-  def set_setting_value(name, value)
-    old_value = Setting[name]
-    new_value = derive_value(value)
-    Setting[name] = new_value
-    if name == :work_package_done_ratio && old_value != "status" && new_value == "status"
-      WorkPackages::Progress::ApplyStatusesPCompleteJob.perform_later(cause_type: "progress_mode_changed_to_status_based")
-    end
-  end
-
-  def derive_value(value)
-    case value
-    when Array, Hash
-      # remove blank values in array, hash settings
-      value.compact_blank!
-    else
-      value.strip
-    end
-  end
+  good_job_control_concurrency_with(
+    perform_limit: 1,
+    key: -> { "WorkPackagesProgressJob" }
+  )
 end
--- a/app/workers/work_packages/progress/migrate_remove_totals_from_childless_work_packages_job.rb
+++ b/app/workers/work_packages/progress/migrate_remove_totals_from_childless_work_packages_job.rb
@@ -44,14 +44,12 @@ class WorkPackages::Progress::MigrateRemoveTotalsFromChildlessWorkPackagesJob < 
       SET derived_estimated_hours = NULL,
           derived_remaining_hours = NULL,
           derived_done_ratio = NULL
-      FROM (
-        SELECT wp_tree.ancestor_id AS id,
-               MAX(generations) AS generations
-        FROM work_package_hierarchies wp_tree
-        GROUP BY wp_tree.ancestor_id
-      ) hierarchy
-      WHERE work_packages.id = hierarchy.id
-      AND hierarchy.generations = 0
+      WHERE work_packages.id IN (
+        SELECT ancestor_id AS id
+        FROM work_package_hierarchies
+        GROUP BY id
+        HAVING MAX(generations) = 0
+      )
       AND (
         work_packages.derived_estimated_hours IS NOT NULL
         OR work_packages.derived_remaining_hours IS NOT NULL
--- a/app/workers/work_packages/progress/migrate_remove_totals_from_childless_work_packages_job.rb
+++ b/app/workers/work_packages/progress/migrate_remove_totals_from_childless_work_packages_job.rb
@@ -1,0 +1,65 @@
+#-- copyright
+# OpenProject is an open source project management software.
+# Copyright (C) 2012-2024 the OpenProject GmbH
+#
+# This program is free software; you can redistribute it and/or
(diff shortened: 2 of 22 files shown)


alandryz83:
Sent from my iPhoneOn May 6, 2024, at 4:39 AM, github-actions[bot] ***@***.***> wrote:﻿


  
    
      
      
          1 Warning
      
     
  
  
    
      ⚠️
      This PR has migration-related changes on a release branch. Ping @opf/operations
    
  


  Generated by 🚫 Danger

lib/open_project/static/links.rb
```diff
@@ -177,6 +177,10 @@ def static_links
               href: "https://www.openproject.org/docs/release-notes/",
               label: :label_release_notes
             },
+            release_notes_14_0_1: {
+              href: "https://www.openproject.org/docs/release-notes/14-0-1/",
+              label: "Release notes for OpenProject 14.0.1"
+            },
```
ulferts, line 183: Why is this added? And if it is added for a reason, It would need i18n. So does the `blog_article_progress_changes` in line 172.

spec/workers/work_packages/progress/migrate_remove_totals_from_childless_work_packages_job_spec.rb
```diff
@@ -0,0 +1,149 @@
...
+            child4    |        |                  |
+        TABLE
+      )
+      work_packages.each do |work_package|
+        expect(work_package.journals.count).to eq 1
+      end
+    end
+    # rubocop:enable RSpec/ExampleLength
+
+    it "does not even try to recompute totals (not that job's business)" do
```
ulferts: To me, the description would be clearer if the sentence was completed "does not even try to recompute totals for parent wps (not that job's business)"

lib/open_project/static/links.rb
ulferts: I now see that the label is not really used. How about removing it then? And maybe have a comment in here on where it is used?

app/workers/work_packages/progress/migrate_remove_totals_from_childless_work_packages_job.rb
```diff
@@ -0,0 +1,65 @@
...
+    results = execute(<<~SQL.squish)
+      UPDATE work_packages
+      SET derived_estimated_hours = NULL,
+          derived_remaining_hours = NULL,
+          derived_done_ratio = NULL
+      FROM (
+        SELECT wp_tree.ancestor_id AS id,
+               MAX(generations) AS generations
+        FROM work_package_hierarchies wp_tree
+          LEFT JOIN work_packages ON wp_tree.descendant_id = work_packages.id
```
ulferts: I think this line is removable.

app/workers/work_packages/progress/migrate_remove_totals_from_childless_work_packages_job.rb
```diff
@@ -0,0 +1,65 @@
...
+        GROUP BY wp_tree.ancestor_id
```
ulferts: You COULD also move the `hierarchy.generations = 0` condition into here by a 
```
HAVING MAX(generations) = 0
```
condition in the subselect. That way, all conditions for the hierarchy would be in one place and the conditions on the work packages in another. That means you could rename the `hierarchy` alias to something like `childless_work_packages`. Or, you COULD also phrase the whole statement like
```
UPDATE 
work_packages
SET ...
WHERE
work_packages.id IN ([subselect for childless work packages])
AND [other conditions]
``` 

I cannot really decide what reads better, so change if at all after your own preferences.

app/workers/work_packages/progress/job.rb
```diff
@@ -0,0 +1,42 @@
...
+# Base class for jobs updating work packages progress values.
+#
+# Handles concurrency so that only one progress job is performed at a time.
+class WorkPackages::Progress::Job < ApplicationJob
+  include GoodJob::ActiveJobExtensions::Concurrency
+  include WorkPackages::Progress::SqlCommands
+
+  queue_with_priority :default
+
+  good_job_control_concurrency_with(
```
ulferts, line 38: I did not know this capability. I am glad that I do now.

ulferts (approved):
I only have minor things so feel free to merge once you are contend. Not everything I noted down needs to be addressed.

Related to this PR (but it was that way before) I stumbled over the phrasing here:

image

I think 

> Status % Complete changed % complete value for status 'In progress' changed from 30% to 40%

can be simplified to

> Status % Complete changed for 'In progress' from 30% to 40%

lib/open_project/static/links.rb
cbliard: Yes, it's not used. I saw some other links having a non-translated non-used label for links. I added it so that the "interface" of the links is respected. Should I remove it?

cbliard:
> I only have minor things so feel free to merge once you are contend. Not everything I noted down needs to be addressed.
> 
> Related to this PR (but it was that way before) I stumbled over the phrasing here:
> 
> image
> I think
> 
> > Status % Complete changed % complete value for status 'In progress' changed from 30% to 40%
> 
> can be simplified to
> 
> > Status % Complete changed for 'In progress' from 30% to 40%

What do you think about Jens' proposed improvement about the journal text @psatyal ?

app/workers/work_packages/progress/migrate_remove_totals_from_childless_work_packages_job.rb
cbliard: I ended up using 

```sql
UPDATE work_packages
SET derived_estimated_hours = NULL,
    derived_remaining_hours = NULL,
    derived_done_ratio = NULL
WHERE work_packages.id IN (
  SELECT ancestor_id AS id
  FROM work_package_hierarchies
  GROUP BY id
  HAVING MAX(generations) = 0
)
AND (
  work_packages.derived_estimated_hours IS NOT NULL
  OR work_packages.derived_remaining_hours IS NOT NULL
  OR work_packages.derived_done_ratio IS NOT NULL
)
RETURNING work_packages.id
```
which seems to read better.

spec/workers/work_packages/progress/migrate_remove_totals_from_childless_work_packages_job_spec.rb
cbliard: Good attention to details 👍

lib/open_project/static/links.rb
ulferts: I don't see value in having it other than maybe making it clearer what the string actually is and for that, I would use a comment. So yes, I would remove it.

cbliard:
> I think
> 
> > Status % Complete changed % complete value for status 'In progress' changed from 30% to 40%
> 
> can be simplified to
> 
> > Status % Complete changed for 'In progress' from 30% to 40%

@ulferts This will be updated in wp 49409 https://community.openproject.org/journals/610615/diff/description?activity_page=work_packages%2F49409

So the text will be "Status 'In progress' % complete value changed from 30% to 40%".

PR #15465, opened 2024-05-03
Make JobStatus API GoodJob::Batch aware

mereghost:
## The Issue

Since we added the usage of `GoodJob::Batch` on the `CopyProjectJob`, it initiates the redirect before the actual batch has ended.

## The fix

We have to alter the `status` property returned by the `API::V3::JobStatus` to `in_process` until the batch finishes.

### DIFF
--- a/modules/job_status/spec/lib/api/v3/job_status/job_status_representer_spec.rb
+++ b/modules/job_status/spec/lib/api/v3/job_status/job_status_representer_spec.rb
@@ -1,0 +1,101 @@
+# frozen_string_literal: true
+
+#-- copyright
+# OpenProject is an open source project management software.
+# Copyright (C) 2012-2024 the OpenProject GmbH
+#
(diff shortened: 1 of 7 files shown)


modules/job_status/spec/lib/api/v3/job_status/job_status_representer_spec.rb
```diff
@@ -0,0 +1,101 @@
...
+          # noop
+        end
+        status = JobStatus::Status.order(:created_at).last
+
+        status_json = described_class.new(status, current_user: user).to_json
+        expect(status_json).to be_json_eql("failure".to_json).at_path("status")
+      end
+    end
+  end
+end
```
akabiru, line 101: Chef's kiss on the tests 🤌

akabiru (approved):
🤌 ✨

PR #15479, opened 2024-05-06
fix visibility scope of labor items

ulferts:
Fixes the visible scope of `LaborBudgetItem`. It was broken in multiple places:

* Unless a project was provided, no scope (nil) was returned. In effect, that lead to the visible scope being inactive most of the time.
* If a project was returned, the scope failed since it tried to find a non existing `projects_id` (instead of `project_id`) column on budgets.
* When trying to actually run the scope that was supposed to be returned, that failed since there is no `project_id` column on the labor_budget_items table but only on the budgets table.

The second flaw suggests, that no project was ever provided as that would have lead to hard exceptions. The PR thus removes the second parameter which also brings the visibility scope more in line with those on other models.

Fixes

https://community.openproject.org/wp/45834


dombesz (approved):
The change looks good, thanks @ulferts! 👏
However, I am not entirely sure if the bugfix works as expected. I tried to reproduce the issue and the Budget list page seems to correctly display 0 for the "Planned budget" if the user has the permissions from the ticket. The details page still shows the amounts of the "Planned budget" and to my understanding it should show 0.

image
image
image

Please feel free to merge it, if this is the expected behaviour.

ulferts:
@dombesz you were right, that part of the information wasn't hidden properly. ~That was not part of the originally described bug but it makes sense to fix the issues together.~ Edit: I was wrong, it was described as well and I overlooked it.

I refactored a bit to make the code clearer but the only changes apart from what you already reviewed are in #2e0e39aacf6cfb90816227cc701579e09c07d16e which was changed again in [2e0e39aacf6cfb90816227cc701579e09c07d16e](https://github.com/opf/openproject/pull/15479/commits/b0527dce646725686155303daa16563bd43ec765#diff-2c3401257092403593fb3392051e86eefab0c5f20d8229a65e0cfbfe34d9f5f5R64-R66) for a simpler implementation. The rest is just moving code around.

dombesz (approved):
Nice job extracting the code into view components!👏
Everything works as expected, I have one question only.

When the user is allowed to `view_own_hourly_rate?`, and has some time logged, the "Available" column shows the exact amount spent by the user with a negative sign.

From my point of view this is okay since we are not showing anything that the user is not supposed to see. But is this the expected behaviour?

image

PR #15493, opened 2024-05-07
Retry jobs that failed because they were not defined (yet)

cbliard:
The worker may fail to load some jobs because the job class does not exist at the time of execution. This can happen on upgrades when the worker is still running the previous version while a migration is enqueuing jobs defined in the new version. The worker fails deserializing them because the job class can not be loaded.

This commit adds an initializer to retry all jobs failing due to NameError on restart so that they can be picked up once the migration is over and the worker gets restarted.

### DIFF
--- a/config/initializers/retry_failed_jobs_with_name_error.rb
+++ b/config/initializers/retry_failed_jobs_with_name_error.rb
@@ -47,5 +47,7 @@ Rails.application.configure do
       rescue StandardError => e
         Rails.logger.error("Failed to enqueue job for retry #{job.display_name} (job id: #{job.id}): #{e.message}")
       end
+  rescue LoadError
+    # Ignore LoadError that happens when nulldb://db database adapter is used
   end
 end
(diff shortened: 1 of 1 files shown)


config/initializers/retry_failed_jobs_with_name_error.rb
```diff
@@ -0,0 +1,51 @@
...
+    # in the new version.
+    #
+    # Once the migration is over and the worker gets restarted, the job will be
+    # retried thanks to this piece of code below.
+    GoodJob::Job
+      .discarded
+      .where("error LIKE ?", "NameError: uninitialized constant %Job")
+      .find_each do |job|
+        job.retry_job
+        Rails.logger.info("Successfully retry job #{job.display_name} (job id: #{job.id})")
```
mrmir: This is just me poking around, but does 'retried' work better here?

cbliard: True, I hesitated to put "retried", but actually, it only enqueues it for retry. It does not retry yet. Maybe "enqueued for retry" would sound better.

cbliard: Fixed. Thanks!

oliverguenther (approved):

PR #15494, opened 2024-05-07
Bundler 2.5.9 -> 2.5.10

klaustopher:
Update Bundler in our files to 2.5.10, it's not completely required, but there was a bug before that potentially URLs with Basic Auth would be dumped into the lockfile. Not really something that is problematic for us, but let's be safe: https://github.com/rubygems/rubygems/blob/master/bundler/CHANGELOG.md#2510-may-3-2024


dombesz (approved):
Thanks Klaus!👍

[This comment](https://github.com/opf/openproject/blob/dc2d337ee87197ff282b1bcdac41148dc92d5fb6/script/github_pr_errors#L286) also contains the bundler version, probably it would be good to update it too.

https://github.com/opf/openproject/blob/dc2d337ee87197ff282b1bcdac41148dc92d5fb6/script/github_pr_errors#L286

PR #15495, opened 2024-05-07
[#53669] Change default view for meetings module to upcoming invitations

mrmir:
https://community.openproject.org/work_packages/53669

### DIFF
--- a/modules/meeting/app/controllers/meetings_controller.rb
+++ b/modules/meeting/app/controllers/meetings_controller.rb
@@ -368,7 +368,7 @@ class MeetingsController < ApplicationController
   end
 
   def determine_date_range
-    @days = 31 # Setting.activity_days_default.to_i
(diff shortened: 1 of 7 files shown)


aaron-contreras (approved):
Very nice. There's only one little caveat (I checked though and it is already a bug on the dev branch so it wasn't introduced by this PR).

On "light mode" the menu item is properly highlighted:
Menu item highlighted

On "High contrast mode" the menu items do not highlight when selected:
image

The acceptance criteria on the work package specifies that the menu entry should be highlighted (but again, probably was written up from a "Light" mode and not a "High contrast mode"). If you could fix this, great :) but it was already a bug present on the current dev branch.

mrmir:
I poked around for a bit but couldn't fix it quickly, so I'll create a separate bug ticket for this issue. It seems to be happening everywhere, and not just in meetings, so I think that makes more sense to do anyway.

PR #15498, opened 2024-05-07
Code maintenance/53729 remove project custom field mapping and the projects aac patch

dombesz:
https://community.openproject.org/wp/53729


oliverguenther (approved):
Thanks for the cleanup 👍 

As already discussed with Attila, some logic is now moved back into acts_as_customizable, even though the only overriding use-case is the project attributes. In the future, this will make moving something like the work package type groups into a similar section/mapping implementation like with projects easier.

PR #15515, opened 2024-05-10
Possilbe having email like 12775518238@qq.com

Eric-Guo:


app/controllers/concerns/member_helper.rb
```diff
@@ -59,7 +59,7 @@ def user_ids_for_new_members(member_params)
 
   def invite_new_users(user_ids, send_notification: true)
     user_ids.filter_map do |id|
-      if id.to_i == 0 && id.present? # we've got an email - invite that user
+      if id.present? && (id.to_i == 0 || EmailValidator.valid?(id)) # we've got an email - invite that user
```
dombesz, line 62: I believe this could be simplified:
```suggestion
      if EmailValidator.valid?(id) # we've got an email - invite that user
```

dombesz (commented):
Thanks @Eric-Guo for the contribution. 👏
Usually we like to have tests included with PRs, san you please add a test for this fix?

app/controllers/concerns/member_helper.rb
Eric-Guo: Yes, but I guess the original implement is for performance reason as the logic is used in batch import also.

But I'm OK if member think the simplified version is OK.

Eric-Guo:
@dombesz Guess I can reuse one so I don't need to write new one. 😎 

Any way the test case is added.

spec/controllers/members_controller_spec.rb
```diff
@@ -164,7 +164,7 @@
     end
 
     context "with yet-to-be-invited emails" do
-      let(:emails) { ["h.wurst@openproject.com", "l.lustig@openproject.com"] }
+      let(:emails) { ["h.wurst@openproject.com", "12775518238@qq.com"] }
```
dombesz: Thanks @Eric-Guo , the test looks good 👍, however I would prefer either using the `@openproject.com` domain in the test or a domain that is reserved for testing such as `@example.com`.

dombesz (changes requested):

dombesz (approved):
Thanks @Eric-Guo !👍

PR #15521, opened 2024-05-13
Downgrade prawn 2.4.0 due to Embedded Font problem in Chrome (windows only) and Acrobat

Eric-Guo:
[Upstream issue](https://github.com/prawnpdf/prawn/issues/1346#issuecomment-2036433712)

Below is the problem PDF file:

image


ulferts:
@Eric-Guo thanks for the PR.

@as-op can you please look into this?

Eric-Guo:
I can confirm Acrobat Pro DC 2023.001 ~ Acrobat Pro DC 2024.002 all having missing fonts issue. some smart PDF reader, like PDF Expert / FoxReader will resolve missing fonts problem automatically. So not every PDF reader having problem.

as-op:
Thank you @Eric-Guo for bringing this to our attention!

I tried with Acrobat Reader 2024.002.20287, but that is not affected. As there is a reproducible case in the linked upstream issue and the confirmation by @Eric-Guo we should downgrade.

Our dependency md-to-pdf is still on prawn 2.4.0 anyhow, so the downgrade should be safe.
I verified this PR, it successfully downgrades to unaffected gem versions.

```
% bundle exec gem list --all --exact prawn ttfunk pdf-core

*** LOCAL GEMS ***

prawn (2.4.0)

*** LOCAL GEMS ***

ttfunk (1.7.0)

*** LOCAL GEMS ***

pdf-core (0.9.0)

```

PR #15523, opened 2024-05-13
Only allow precompiled assets to be injected when running from CI

crohr:
https://community.openproject.org/work_packages/54993/activity


machisuji (approved):
Thanks! 👍

PR #15530, opened 2024-05-13
[51447] Use same headings for permissions report and role forms

cbliard:
https://community.openproject.org/wp/51447

The pages are now showing the same headings (and using the same code to render them)

Before:

image

After:

image


Kharonus (approved):
LGTM

PR #15533, opened 2024-05-13
[#51745] Consolidate translations strings for work package sharing feature

dombesz:
https://community.openproject.org/work_packages/51745

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -2891,7 +2891,7 @@ Project attributes and sections are defined in the <a href=%{admin_settings_url}
     member_of_group: "Assignee's group"
     name_or_identifier: "Name or identifier"
     only_subproject_id: "Only subproject"
-    shared_with_user: "Shared with user"
+    shared_with_user: "Shared with users"
     shared_with_me: "Shared with me"
     subproject_id: "Including subproject"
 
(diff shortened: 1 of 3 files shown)


config/locales/js-en.yml
```diff
@@ -1123,7 +1123,7 @@ en:
       sharing:
         share: "Share"
         title: "Share work package"
-        show_all_users: "Show all shared users"
+        show_all_users: "Show all users with whom the work packages has been shared with"
```
dombesz: According to the docs, the meaning of the translation was incorrect. Here's my proposed version.
cc: @psatyal

config/locales/en.yml
```diff
@@ -357,6 +357,7 @@ Project attributes and sections are defined in the <a href=%{admin_settings_url}
       project_roles: "Project roles"
       wp_shares: "Work package shares"
       groups: "Groups"
+    no_modify_on_shared: "You currently cannot modify or remove shared memberships through the member page. Use the sharing modal instead."
```
dombesz, line 360: This translation was completely missing and it is not being used. Added it for future use.

cbliard (approved):
LGTM.
I only fixed one typo in `query_fields.shared_with_user`.

dombesz:
Thanks @cbliard, I am not sure if that is a typo. According to the specification it should be "users".

image

cbliard:
@dombesz well, I called it a typo because I assumed that the person having added that string made a typo when entering it.
I [fixed it](https://github.com/opf/openproject/pull/15533/commits/925f69797afe7e735e8a2e4d25fc60b41474e088) to match the specification so it's "users" as expected.

dombesz:
@cbliard you are right, I overlooked that one, thanks!

PR #15539, opened 2024-05-13
Exclude text custom fields from sorting

klaustopher:
As discussed in https://community.openproject.org/projects/openproject/work_packages/55015/activity we want to exclude long texts from custom field ordering

---

Also includes review feedback that the empty option is hardly selectable. So we use the `-` that we also use in WorkPackage configuríng


ulferts (approved):
👍

PR #15559, opened 2024-05-14
rename persisted project query

toy:
[OP#51673](https://community.openproject.org/projects/openproject/work_packages/51673)

Includes rework of projects menu, with two goals - properly marking list selected from both project and project queries controller for all actions and fixing the bug with marking menu item if there was query_id param even on unrelated pages (work_packages persisted lists) or marking the «Active projects» when there were no params.

Also split huge persisting queries spec into multiple smaller ones.

### DIFF
--- a/app/forms/queries/projects/form.rb
+++ b/app/forms/queries/projects/form.rb
@@ -26,7 +26,7 @@
 # See COPYRIGHT and LICENSE files for more details.
 # ++
 
-class Queries::Projects::Create < ApplicationForm
+class Queries::Projects::Form < ApplicationForm
   include OpenProject::StaticRouting::UrlHelpers
 
   form do |query_form|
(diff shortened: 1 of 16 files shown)


app/forms/queries/projects/form.rb
```diff
@@ -52,8 +52,13 @@ class Queries::Projects::Create < ApplicationForm
...
         data: { "params-from-query-target": "anchor" },
-        href: OpenProject::StaticRouting::StaticUrlHelpers.new.projects_path
+        href: projects_path(query_id: @query)
       )
     end
   end
+
+  def initialize(query:)
+    super()
+    @query = query
```
toy, line 62: I didn't find a way to get model (query) from the ApplicationForm, so simpler and probably cleaner to just pass it

klaustopher (approved):
looks good to me. A lot of refactorings but looks reasonable

PR #15574, opened 2024-05-15
[#55100] Meeting sections do not show "Move up" and "Move to top" action

dombesz:
https://community.openproject.org/work_packages/55100

### DIFF
--- a/modules/meeting/app/controllers/meeting_sections_controller.rb
+++ b/modules/meeting/app/controllers/meeting_sections_controller.rb
@@ -53,7 +53,8 @@ class MeetingSectionsController < ApplicationController
       add_section_via_turbo_stream
       update_section_header_via_turbo_stream(state: :edit)
       # update the section header of the previously last section in order to ensure the action menu move options are updated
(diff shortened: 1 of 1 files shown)


HDinger (changes requested):
It works correctly for the section that was just moved. However, the section that is now newly at the top (respectivly bottom) does not show the correct items.

agenda_items_menu_after_drag

jjabari-op:
@HDinger thanks for double checking! it should be fixed now :)

HDinger (commented):
Yes, the described issue works correctly. 👍  While testing, I just found one other thing: When creating a new section, the previuosly last section is not correctly updated. It still shows only the arrows up, but should have all options:

Bildschirmfoto 2024-05-16 um 11 59 33

I don't know if that is an easy fix. If not, feel free to extract that into a separate ticket.

jjabari-op:
@HDinger Good eye! That's an easy fix actually, give me a sec ;)

jjabari-op:
@HDinger Could you have another look? Thank you!

HDinger (approved):
Looks good, thanks 👍  Good to merge, once the tests are green

PR #15582, opened 2024-05-16
Add example for IMAP folder

suntorytimed:
Adding an example for IMAP folders for successfully and unsuccessfully handled incoming mails. This should make it easier to configure the incoming mail support.


oliverguenther (approved):
Thanks for the suggestion, and the corrected string 👍

PR #15593, opened 2024-05-17
Fix ERB error

Eric-Guo:
introduced at #15585


klaustopher:
Already fixed. Thank you

PR #15602, opened 2024-05-17
Fix/limit schedule reminder mail jobs

ulferts:
Applies the same pattern as done for the `ScheduleDateAlertsNotificationsJob` in #15573 to the `ScheduleReminderMailsJob`. It now also takes the time between the two cron_at values to be the boundaries in which the job operates.

The similarities of the two jobs are now extracted into a shared module. 

In order to avoid a single job being retried indefinitely a [`retry_on` limit is added](https://github.com/opf/openproject/compare/release/14.1...fix/limit_schedule_reminder_mail_jobs?expand=1#diff-0fd1663714342168a70bbbd898374fb1a2609f3ad6767f36485275133dc44eeaR47-R49)

### DIFF
--- a/app/models/users/scopes/having_reminder_mail_to_send.rb
+++ b/app/models/users/scopes/having_reminder_mail_to_send.rb
@@ -31,13 +31,13 @@ module Users::Scopes
     extend ActiveSupport::Concern
 
     class_methods do
-      # Returns all users for which a reminder mails should be sent now. A user
+      # Returns all users for which a reminder mails should be sent. A user
       # will be included if:
       #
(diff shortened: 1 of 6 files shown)


ulferts:
@mereghost , this is the PR that will add the limits you asked for in #15573. I tested the behaviour locally and it seems to work as expected. The job is retried twice (with a 5 minute backoff) and then discarded. 5 minutes after that, the next job is scheduled according to the cron tab.

mereghost (approved):
Nice. I'll apply the same patterns on other jobs that have the concurrency controls set.

PR #15603, opened 2024-05-17
Makes sure jobs are eventually discarded when concurrency happens

mereghost:
##### ⚠️ Related Work Package [OP#54984](https://community.openproject.org/projects/openproject/work_packages/54984)

This replicates the workaround to forever retrying jobs that @ulferts applied in #15602 to other jobs using concurrency controls.

### DIFF
--- a/app/workers/notifications/create_date_alerts_notifications_job.rb
+++ b/app/workers/notifications/create_date_alerts_notifications_job.rb
@@ -30,14 +30,7 @@ module Notifications
   class CreateDateAlertsNotificationsJob < ApplicationJob
     include GoodJob::ActiveJobExtensions::Concurrency
 
-    good_job_control_concurrency_with(
-      enqueue_limit: 1,
-      perform_limit: 1
-    )
-
-    retry_on GoodJob::ActiveJobExtensions::Concurrency::ConcurrencyExceededError,
-             wait: 5.minutes,
-             attempts: 3
+    good_job_control_concurrency_with total_limit: 1
 
     def perform(user)
       return unless EnterpriseToken.allows_to?(:date_alerts)
--- a/app/workers/notifications/create_date_alerts_notifications_job.rb
+++ b/app/workers/notifications/create_date_alerts_notifications_job.rb
@@ -28,6 +28,17 @@
 
 module Notifications
   class CreateDateAlertsNotificationsJob < ApplicationJob
+    include GoodJob::ActiveJobExtensions::Concurrency
+
+    good_job_control_concurrency_with(
+      enqueue_limit: 1,
+      perform_limit: 1
+    )
+
+    retry_on GoodJob::ActiveJobExtensions::Concurrency::ConcurrencyExceededError,
+             wait: 5.minutes,
+             attempts: 3
+
     def perform(user)
       return unless EnterpriseToken.allows_to?(:date_alerts)
 
--- a/app/workers/notifications/create_date_alerts_notifications_job.rb
+++ b/app/workers/notifications/create_date_alerts_notifications_job.rb
@@ -30,7 +30,10 @@ module Notifications
   class CreateDateAlertsNotificationsJob < ApplicationJob
     include GoodJob::ActiveJobExtensions::Concurrency
 
-    good_job_control_concurrency_with total_limit: 1
+    good_job_control_concurrency_with(
+      total_limit: 1,
+      key: -> { "#{self.class.name}-#{arguments.last}" }
+    )
 
     def perform(user)
       return unless EnterpriseToken.allows_to?(:date_alerts)
--- a/app/workers/work_packages/apply_working_days_change_job.rb
+++ b/app/workers/work_packages/apply_working_days_change_job.rb
@@ -34,6 +34,10 @@ class WorkPackages::ApplyWorkingDaysChangeJob < ApplicationJob
     total_limit: 1
   )
 
+  retry_on GoodJob::ActiveJobExtensions::Concurrency::ConcurrencyExceededError,
+           wait: 5.minutes,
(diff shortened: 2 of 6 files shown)


app/workers/work_packages/apply_working_days_change_job.rb
```diff
@@ -34,6 +34,10 @@ class WorkPackages::ApplyWorkingDaysChangeJob < ApplicationJob
     total_limit: 1
   )
 
+  retry_on GoodJob::ActiveJobExtensions::Concurrency::ConcurrencyExceededError,
+           wait: 5.minutes,
+           attempts: 3
+
```
ulferts: With a `total_limit` of 1, the ConcurrencyExceededError should never happen I believe. Because it would only be thrown for a queued job that tries to be executed. But it shouldn't even be possible to queue a second job.

app/workers/work_packages/progress/job.rb
```diff
@@ -39,4 +39,8 @@ class WorkPackages::Progress::Job < ApplicationJob
     perform_limit: 1,
     key: -> { "WorkPackagesProgressJob" }
   )
+
+  retry_on GoodJob::ActiveJobExtensions::Concurrency::ConcurrencyExceededError,
+           wait: 5.minutes,
+           attempts: 3
```
ulferts: I wouldn't want to loose the data change that the job carries out. If anything, I would set the `retry_on` for this job to `:unlimited` and [overwrite the default of 5](https://api.rubyonrails.org/classes/ActiveJob/Exceptions/ClassMethods.html#method-i-retry_on).

app/workers/notifications/create_date_alerts_notifications_job.rb
```diff
@@ -28,6 +28,17 @@
...
+    include GoodJob::ActiveJobExtensions::Concurrency
+
+    good_job_control_concurrency_with(
+      enqueue_limit: 1,
+      perform_limit: 1
+    )
+
+    retry_on GoodJob::ActiveJobExtensions::Concurrency::ConcurrencyExceededError,
+             wait: 5.minutes,
+             attempts: 3
```
ulferts: For this job, I think a `total_limit` of 1 makes sense. With that in place, we don't need the `retry_on` (at least for the `ConcurrencyExceededError`)

ulferts (changes requested):
Hi Marcello, good, that you went through the jobs to apply concurrency limitations. I fear, though, that I found a number of things where I would handle things differently. 

Apart from the comments within the code, I also think that it would make sense to apply a `total_limit` of 1 to the `Mails::ReminderJob`.

app/workers/work_packages/progress/job.rb
mereghost: And keep the 5 minute interval retry?

mereghost:
> Hi Marcello, good, that you went through the jobs to apply concurrency limitations. I fear, though, that I found a number of things where I would handle things differently.

That was exactly why I asked for your review, as you have more context on the details of the utility of those jobs. =)

app/workers/work_packages/progress/job.rb
```diff
@@ -39,4 +39,8 @@ class WorkPackages::Progress::Job < ApplicationJob
     perform_limit: 1,
     key: -> { "WorkPackagesProgressJob" }
   )
+
+  retry_on GoodJob::ActiveJobExtensions::Concurrency::ConcurrencyExceededError,
+           wait: 5.minutes,
+           attempts: :unlimited
```
ulferts, line 45: I would change this to apply to every almost every error. So it would have to be something like

`retry_on StandardError, wait: 5.minutes, attempts: :unlimited`

This has less to do with Concurrency concerns and more with the necessity of this job to be executed. On the other hand, at some point it just doesn't make sense to try again. But we don't have a pattern yet on how to handle this (e.g. by sending out a notification to an admin). So in the meantime, I would probably do 

`retry_on StandardError, wait: 5.minutes, attempts: 3`

ulferts (changes requested):
Hey @mereghost, please understand the comment as a starting point of a discussion since I really don't know which solution would be best.

app/workers/work_packages/progress/job.rb
mereghost: Long train of thought here (correct any misunderstandings please):

1. We have a `perform_limit` of 1, meaning 1 running and infinite queued.
2. We have a fixed concurrency `key`.
3. This is the base job of 3 others, that will share the same key meaning the others will wait until the current processing job is done.
4. The process of waiting will raise a concurrency error.
5. We need these jobs to run, no matter why, no matter how long it takes for it finally 
run.

Due to the shared key, we might run on the concurrency error frequently (I have no clue on how many times per <SEGMENT OF TIME> this jobs are run), so a retry on that should take care of the exponential backoff that  `GoodJob` adds by default.

Retrying on `StandardError` seems... off. It is too generic and might keep enqueuing a job that has bad arguments or, let's say the code is broken due to the mystical influence of Production DBs.

If we could narrow down a bit, let's say `ActiveRecord` errors, `PG::Errors` or so, I think it would be more viable.

(Also a place for admins see which jobs have failed and maybe retry some of them would be awesome 😛)

ulferts: Ok, I can follow your reasoning with the `ConcurrencyExceededError`, @mereghost, so lets add the statement just you like you proposed initially.

As for the other errors, I currently just don't know which ones to expect. But without any other specification AFAIK, the jobs will just be retried 5 times so that might work for the time being anyways.

ulferts (approved):
I read up on your explanation @mereghost and I can get behind it. 

The PR requires rebasing but can then be merged.

PR #15609, opened 2024-05-19
Allow ProjectQuery to be marked as public

klaustopher:
This implements the portion of making project lists public. 

- [x] Add DB migration to add a `public` flag
- [x] Add `public` and `private` scopes to the `ProjectQuery` model
- [x] Add a global `manage_public_project_queries` permission
- [x] Add a **temporary** link to the `...` menu to change public status of a query
- [x] Add a section to the projects sidebar, to show all the public queries
- [x] Change loading of the queries from `where(user:)` to `public.or(private)`
- [x] When a user changes a public query and they do not have the permission to change them offer **save as** instead of **save** option
- [x] Add tests for contract, service, new scopes
- [x] Add tests for behaviors

It is intended to be a self contained spike that is not relying on the modal or anything, hence why we are **temporarily** adding the menu items.

---

Implements https://community.openproject.org/work_packages/55159

### DIFF
--- a/app/components/projects/index_page_header_component.html.erb
+++ b/app/components/projects/index_page_header_component.html.erb
@@ -81,6 +81,28 @@
         end
 
         if query.persisted?
+          if can_publish?
+            if query.public?
+              menu.with_item(
+                label: t(:button_unpublish),
+                scheme: :danger,
+                href: unpublish_projects_query_path(query),
+                content_arguments: { data: { method: :post } }
+              ) do |item|
+                item.with_leading_visual_icon(icon: 'eye-closed')
+              end
+            else
+              menu.with_item(
+                label: t(:button_publish),
+                scheme: :default,
+                href: publish_projects_query_path(query),
+                content_arguments: { data: { method: :post } }
+              ) do |item|
+                item.with_leading_visual_icon(icon: 'eye')
+              end
+            end
+          end
+
           menu.with_item(
             label: t(:button_delete),
             scheme: :danger,
@@ -92,7 +114,6 @@
       end
     end
   %>
-
   <%= render(Projects::ConfigureViewModalComponent.new(query:)) %>
   <%= render(Projects::DeleteListModalComponent.new(query:)) if query.persisted? %>
   <%= render(Projects::ExportListModalComponent.new(query:)) %>
--- a/app/components/projects/index_page_header_component.html.erb
+++ b/app/components/projects/index_page_header_component.html.erb
@@ -81,7 +81,7 @@
         end
 
         if query.persisted?
-          if can_publish? && OpenProject::FeatureDecisions.project_list_sharing_active?
+          if can_publish?
             if query.public?
               menu.with_item(
                 label: t(:button_unpublish),
--- a/app/components/projects/index_page_header_component.html.erb
+++ b/app/components/projects/index_page_header_component.html.erb
@@ -81,6 +81,7 @@
         end
 
         if query.persisted?
+          # TODO: Remove section when the sharing modal is implemented (https://community.openproject.org/projects/openproject/work_packages/55163)
           if can_publish?
             if query.public?
               menu.with_item(
--- a/app/components/projects/index_page_header_component.html.erb
+++ b/app/components/projects/index_page_header_component.html.erb
@@ -81,7 +81,7 @@
         end
 
         if query.persisted?
-          if can_publish?
+          if can_publish? && OpenProject::FeatureDecisions.project_list_sharing_active?
             if query.public?
               menu.with_item(
                 label: t(:button_unpublish),
--- a/app/components/projects/index_page_header_component.rb
+++ b/app/components/projects/index_page_header_component.rb
@@ -70,7 +70,17 @@ class Projects::IndexPageHeaderComponent < ApplicationComponent
 
   def can_save_as? = may_save_as? && query.changed?
 
-  def can_save? = can_save_as? && query.persisted? && query.user == current_user
+  def can_save?
+    return false unless current_user.logged?
+    return false unless query.persisted?
+    return false unless query.changed?
+
+    if query.public?
+      current_user.allowed_globally?(:manage_public_project_queries)
+    else
+      query.user == current_user
+    end
+  end
 
   def can_rename? = may_save_as? && query.persisted? && query.user == current_user && !query.changed?
 
--- a/app/components/projects/index_page_header_component.rb
+++ b/app/components/projects/index_page_header_component.rb
@@ -82,7 +82,17 @@ class Projects::IndexPageHeaderComponent < ApplicationComponent
     end
   end
 
-  alias can_rename? can_save?
+  def can_rename?
+    return false unless current_user.logged?
+    return false unless query.persisted?
+    return false if query.changed?
+
+    if query.public?
+      current_user.allowed_globally?(:manage_public_project_queries)
+    else
+      query.user == current_user
+    end
+  end
 
   def can_publish?
     OpenProject::FeatureDecisions.project_list_sharing_active? &&
--- a/app/components/projects/index_page_header_component.rb
+++ b/app/components/projects/index_page_header_component.rb
@@ -82,7 +82,7 @@ class Projects::IndexPageHeaderComponent < ApplicationComponent
     end
   end
 
-  def can_rename? = may_save_as? && query.persisted? && query.user == current_user && !query.changed?
+  alias can_rename? can_save?
 
   def can_publish?
     OpenProject::FeatureDecisions.project_list_sharing_active? &&
--- a/app/components/projects/index_page_header_component.rb
+++ b/app/components/projects/index_page_header_component.rb
@@ -85,7 +85,9 @@ class Projects::IndexPageHeaderComponent < ApplicationComponent
   def can_rename? = may_save_as? && query.persisted? && query.user == current_user && !query.changed?
 
   def can_publish?
-    current_user.allowed_globally?(:manage_public_project_queries) && query.persisted?
+    OpenProject::FeatureDecisions.project_list_sharing_active? &&
+    current_user.allowed_globally?(:manage_public_project_queries) &&
+    query.persisted?
   end
 
   def show_state?
--- a/app/components/projects/index_page_header_component.rb
+++ b/app/components/projects/index_page_header_component.rb
@@ -74,6 +74,10 @@ class Projects::IndexPageHeaderComponent < ApplicationComponent
 
   def can_rename? = may_save_as? && query.persisted? && query.user == current_user && !query.changed?
 
+  def can_publish?
+    current_user.allowed_globally?(:manage_public_project_queries) && query.persisted?
+  end
+
   def show_state?
     state == :show
   end
--- a/app/components/projects/row_component.rb
+++ b/app/components/projects/row_component.rb
@@ -46,19 +46,19 @@ module Projects
 
     def favored
       render(Primer::Beta::IconButton.new(
-        icon: currently_favored? ? "star-fill" : "star",
-        scheme: :invisible,
-        mobile_icon: currently_favored? ? "star-fill" : "star",
-        size: :medium,
-        tag: :a,
-        tooltip_direction: :e,
-        href: helpers.build_favorite_path(project, format: :html),
-        data: { method: currently_favored? ? :delete : :post },
-        classes: currently_favored? ? "op-primer--star-icon " : "op-project-row-component--favorite",
-        label: currently_favored? ? I18n.t(:button_unfavorite) : I18n.t(:button_favorite),
-        aria: { label: currently_favored? ? I18n.t(:button_unfavorite) : I18n.t(:button_favorite) },
-        test_selector: 'project-list-favorite-button'
-      ))
+               icon: currently_favored? ? "star-fill" : "star",
+               scheme: :invisible,
+               mobile_icon: currently_favored? ? "star-fill" : "star",
+               size: :medium,
+               tag: :a,
+               tooltip_direction: :e,
+               href: helpers.build_favorite_path(project, format: :html),
+               data: { method: currently_favored? ? :delete : :post },
+               classes: currently_favored? ? "op-primer--star-icon " : "op-project-row-component--favorite",
+               label: currently_favored? ? I18n.t(:button_unfavorite) : I18n.t(:button_favorite),
+               aria: { label: currently_favored? ? I18n.t(:button_unfavorite) : I18n.t(:button_favorite) },
+               test_selector: "project-list-favorite-button"
+             ))
     end
 
     def currently_favored?
@@ -197,7 +197,7 @@ module Projects
     def additional_css_class(column)
       if column.attribute == :name
         "project--hierarchy #{project.archived? ? 'archived' : ''}"
-      elsif [:status_explanation, :description].include?(column.attribute)
+      elsif %i[status_explanation description].include?(column.attribute)
         "project-long-text-container"
       elsif custom_field_column?(column)
         cf = column.custom_field
@@ -254,7 +254,7 @@ module Projects
         href: helpers.build_favorite_path(project, format: :html),
         data: { method: :post },
         label: I18n.t(:button_favorite),
-        aria: { label: I18n.t(:button_favorite) },
+        aria: { label: I18n.t(:button_favorite) }
       }
     end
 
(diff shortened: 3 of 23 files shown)


app/components/projects/index_page_header_component.html.erb
```diff
@@ -72,6 +72,28 @@
         end
 
         if query.persisted?
+          if can_publish? && OpenProject::FeatureDecisions.project_list_sharing_active?
```
toy: Minor - I think `OpenProject::FeatureDecisions.project_list_sharing_active?` should better be inside `can_publish?` if that method is used anywhere else

app/components/projects/row_component.rb
```diff
@@ -46,19 +46,19 @@ def hierarchy
 
     def favored
```
toy, line 47: Are all changes in this file style/rubocop only?

app/components/projects/table_component.rb
```diff
@@ -64,11 +64,10 @@ def build_sort_header(column, options)
     # We don't return the project row
     # but the [project, level] array from the helper
     def rows
```
toy, line 66: Also only style/rubocop changes?

app/contracts/queries/projects/project_queries/base_contract.rb
```diff
@@ -41,17 +41,36 @@ def self.model
...
+    validate :user_is_logged_in
+    validate :allowed_to_modify_private_query
+    validate :allowed_to_modify_public_query
 
+    protected
 
-    def user_is_current_user_and_logged_in
-      unless user.logged? && user == model.user
+    def user_is_logged_in
+      if !user.logged?
```
toy: Just a remark, no need to change - using unless with two conditions can be confusing, but with one seem better than `if !`

app/contracts/queries/projects/project_queries/base_contract.rb
```diff
@@ -41,17 +41,36 @@ def self.model
...
         errors.add :base, :error_unauthorized
       end
     end
 
+    def allowed_to_modify_private_query
+      return if model.public?
+
+      if model.user != user
+        errors.add :base, :can_only_be_modified_by_owner
+      end
```
toy, line 63: Wouldn't two guards read cleaner:
```suggestion
      return if model.public?
      return if model.user == user

      errors.add :base, :can_only_be_modified_by_owner
```

app/contracts/queries/projects/project_queries/base_contract.rb
```diff
@@ -41,17 +41,36 @@ def self.model
...
+    end
+
+    def allowed_to_modify_public_query
+      return unless model.public?
+
+      unless user.allowed_globally?(:manage_public_project_queries)
+        errors.add :base, :need_permission_to_modify_public_query
+      end
```
toy, line 71: =||=
```suggestion
      return unless model.public?
      return if user.allowed_globally?(:manage_public_project_queries)

      errors.add :base, :need_permission_to_modify_public_query
```

app/controllers/projects/queries_controller.rb
```diff
@@ -31,13 +31,17 @@ class Projects::QueriesController < ApplicationController
...
+  before_action :find_query, only: %i[show update destroy publish unpublish]
   before_action :build_query_or_deny_access, only: %i[new create]
 
   current_menu_item [:new, :create] do
     :projects
   end
 
+  def show
+    redirect_to projects_path(query_id: @query.id)
+  end
```
toy, line 43: Is the action just to redirect really needed?

app/controllers/projects/queries_controller.rb
```diff
@@ -89,7 +89,23 @@ def destroy
...
+
+      render template: "/projects/index",
+             layout: "global",
+             locals: { query: modified_query, state: :edit }
+    end
+  end
+
   def find_query
-    @query = Queries::Projects::ProjectQuery.find(params[:id])
+    @query = Queries::Projects::ProjectQuery.visible(user: current_user).find(params[:id])
```
toy: Good catch, so before this it is possible to change other user lists?

app/controllers/projects/query_loading.rb
```diff
@@ -30,10 +30,10 @@ module QueryLoading
     private
 
     def load_query(duplicate:)
```
toy, line 32: Only style/rubocop changes

app/helpers/menus/projects_helper.rb
```diff
@@ -64,9 +66,18 @@ def static_status_filters
       ]
     end
 
+    def public_filters
+      ::Queries::Projects::ProjectQuery
+        .public_lists
+        .order(:name)
+        .map do |query|
+        query_menu_item(query)
+      end
```
toy:
```suggestion
        .map { |query| query_menu_item(query) }
```

app/models/queries/projects/project_query.rb
```diff
@@ -36,6 +36,17 @@ class Queries::Projects::ProjectQuery < ApplicationRecord
...
   serialize :selects, coder: Queries::Serialization::Selects.new(self)
 
+  scope :public_lists, -> { where(public: true) }
+  scope :private_lists, ->(user: User.current) { where(public: false, user:) }
+
+  scope :visible, ->(user: User.current) {
+                    public_lists.or(private_lists(user:))
+                  }
+
+  def visible?(user: User.current)
```
toy: Other methods with that name use positional argument

config/initializers/permissions.rb
```diff
@@ -178,6 +178,14 @@
                      permissible_on: :global,
                      require: :loggedin,
                      grant_to_admin: true
+
+      map.permission :manage_public_project_queries,
+                     {
+                       "projects/queries": %i[publish unpublish]
```
toy, line 184: Feels a bit odd that permission gives ability to not only unpublish, but also publish lists

db/migrate/20240519123921_add_public_to_project_queries.rb
```diff
@@ -0,0 +1,6 @@
+class AddPublicToProjectQueries < ActiveRecord::Migration[7.1]
+  def change
+    add_column :project_queries, :public, :boolean, default: false, null: false
+    add_index :project_queries, :public
```
toy, line 4: Do we know if there are instances with enough projects to make this slow (presence of default value) and is it a problem?

spec/contracts/queries/projects/project_queries/shared_contract_examples.rb
```diff
@@ -55,7 +55,51 @@
...
+      it_behaves_like "contract is invalid", base: :can_only_be_modified_by_owner
+    end
+
+    context "if the list is public and the editing user has the permission" do
+      let(:query_user) { build_stubbed(:user) }
+
+      before do
+        query.change_by_system do
+          query.public = true
+        end
```
toy, line 67: Is `change_by_system` really needed, seems not necessary in this shared spec context

spec/controllers/projects/queries_controller_spec.rb
```diff
@@ -157,7 +174,9 @@
 
       before do
         allow(controller).to receive(:permitted_query_params).and_return(query_params)
-        allow(Queries::Projects::ProjectQuery).to receive(:find).with(query_id).and_return(query)
+        scope = instance_double(ActiveRecord::Relation)
+        allow(Queries::Projects::ProjectQuery).to receive(:visible).and_return(scope)
+        allow(scope).to receive(:find).with(query_id).and_return(query)
```
toy, line 214: Maybe easier to use `receive_message_chain` in this case

app/components/projects/index_page_header_component.html.erb
klaustopher: It's not but it makes sense to put it there

app/components/projects/row_component.rb
klaustopher: Yeah, I made a change in the file and rubocop fixing runs on save

app/components/projects/table_component.rb
klaustopher: yep, format on save

app/contracts/queries/projects/project_queries/base_contract.rb
klaustopher: I basically followed the style that is used in the other methods in the class. They usuall go with "auxillary" clauses as guards and the "main" check as an if/unless clause

app/contracts/queries/projects/project_queries/base_contract.rb
klaustopher: Same as above. I think it's fine to have some border conditions checked in guard clauses but the "main" check in an if. But that's a question of personal style. No hard feelings on any solution

app/controllers/projects/queries_controller.rb
klaustopher: I had a few cases, where the browser encountered an error on the POST to `/queries/1` and that then showed up as the URL in the browser, when refreshing I was met with a 404 error. Personally, I want to surprise the user the least, and if they refresh they should land on a page that makes sense and that what this redirect does

app/controllers/projects/queries_controller.rb
klaustopher: They were found, but when the service was called to update them, the contract prevented it because it made the `query.user == current_user` check

app/models/queries/projects/project_query.rb
klaustopher: Yeah, makes sense to be consistent here.

config/initializers/permissions.rb
klaustopher: Why does this feel odd? The specification says that you need a global permission to make lists public.

db/migrate/20240519123921_add_public_to_project_queries.rb
klaustopher: @ulferts Can you maybe chime in? I don't think that there are instances where we have too many project queries, that adding a default will slow us down significantly. We can always go ahead and split up the migration and add a background job to add the default value everywhere. But I suppose that this is a non issue.

Also, it seems like [Postgres 11](https://www.depesz.com/2018/04/04/waiting-for-postgresql-11-fast-alter-table-add-column-with-a-non-null-default/) changed the way how new columns with default values are added so it does not change every row in the table, but has a more efficient way to insert.

spec/contracts/queries/projects/project_queries/shared_contract_examples.rb
klaustopher: Yes, it is needed, because according to the contract, the `public` attribute is not writable except for the `PublishContract`. The same is true for the `user` attribute. That's why we need to set it this way, otherwise the contract will fail because we are trying to write to a field that is not writable.

spec/controllers/projects/queries_controller_spec.rb
klaustopher: Again, I am following the pattern that is used in the rest of the files related. Also, `receive_message_chain` has a SUPER WEIRD syntax when it comes to validating arguments. The call would look like:

```ruby
allow(Queries::Projects::ProjectQuery).to \
  receive_message_chain(:visible, :find)
    .with(no_args) # args for visible
    .with(query_id) # args for find
    .and_return(query)
```

I find this a lot more confusing than the way how it is now.

---

Also, our rubocop rules do not allow `receive_message_chain`. Also for good reasons, as it does not verify that the methods are even defined, so it's easy to mess things up

app/components/projects/index_page_header_component.html.erb
toy: * if in future

app/components/projects/row_component.rb
toy: It is confusing, as it leads to trying to find any change in logic to no avail

app/controllers/projects/queries_controller.rb
toy: I assume you mean resubmitting the url from the url bar, as refresh of post POST by browsers should still be a POST request ([discussion about it](https://github.com/whatwg/html/issues/6600)). Not sure it will reduce user surprise, as changes that were in post data will be gone

config/initializers/permissions.rb
toy: Disregard, I misunderstood the logic

spec/contracts/queries/projects/project_queries/shared_contract_examples.rb
toy: But why not `query.update_attribute!(:public, true)` or alike?

klaustopher: Because in the tests we do not persist the query. So everything that is set up on the query outside the `change_by_system` method is part of the `changes` and will be evaluated by the contract

toy: I think I got what you mean after checking the whole contract inheritance chain

app/components/projects/index_page_header_component.html.erb
```diff
@@ -81,6 +81,28 @@
         end
 
         if query.persisted?
+          if can_publish?
```
toy, line 85: Maybe it is better to introduce a separate `can_delete?` and wrap deletion menu with it instead of `query.persisted?` and put publish menu items outside as `can_publish?` already checks for `query.persisted?`

klaustopher: The entire menu items around sharing will be removed when we introduce [the sharing modal](https://community.openproject.org/work_packages/55163/activity). So this is only temporary and will be refactored in the end

klaustopher: I added a TODO in the code to mark that

toy: It would still be nice to create `can_delete?` method instead of using `query.persisted?`, but is no related to this PR

app/components/projects/index_page_header_component.html.erb
```diff
@@ -81,6 +81,29 @@
...
+                scheme: :danger,
+                href: unpublish_projects_query_path(query),
+                content_arguments: { data: { method: :post } }
+              ) do |item|
+                item.with_leading_visual_icon(icon: 'eye-closed')
+              end
+            else
+              menu.with_item(
+                label: t(:button_publish),
+                scheme: :default,
```
toy, line 98: Unpublishing should probably also use scheme danger, but not important if it is all gone with the modal

app/components/projects/index_page_header_component.rb
```diff
@@ -70,9 +70,35 @@ def may_save_as? = current_user.logged?
...
+    return false if query.changed?
+
+    if query.public?
+      current_user.allowed_globally?(:manage_public_project_queries)
+    else
+      query.user == current_user
+    end
+  end
+
+  def can_publish?
```
toy, line 97: Nitpick - probably better name would be something like `can_manage_publishing?`, as otherwise it is not obvious that it is also for unpublishing

spec/models/queries/projects/project_query_spec.rb
```diff
@@ -387,4 +387,66 @@
...
+        expect(described_class.private_lists(user: private_query.user)).to contain_exactly(private_query)
+      end
+    end
+
+    describe ".visible" do
+      it "returns public and private queries owned by the user" do
+        public_query = create(:project_query, public: true)
+        public_query_other_user = create(:project_query, public: true)
+        private_query = create(:project_query, public: false)
+        create(:project_query, public: false)
```
toy, line 417: Nitpick - it may be easier to understand if all instances for all scopes tests are created together at the top, also second public query is not really needed in tests, as it is not different from the first one

app/services/queries/projects/project_queries/publish_service.rb
```diff
@@ -0,0 +1,49 @@
...
+#
+# You should have received a copy of the GNU General Public License
+# along with this program; if not, write to the Free Software
+# Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+#
+# See COPYRIGHT and LICENSE files for more details.
+#++
+
+module Queries::Projects::ProjectQueries
+  class PublishService < BaseServices::Update
```
toy, line 30: Same point as for `can_publish?`, may be better to make clear in the name that unpublishing can also be done by this service

toy (approved):
Just few more nitpicks.
I'm still not convinced about `show` action, but the rest looks fine, so I leave it up to you

spec/models/queries/projects/project_query_spec.rb
klaustopher: Yes, it is different as it belongs to another user. So we are testing that the owner is irrelevant when queries are marked as public

toy: Both don't have user explicitly set, and neither `public_query.user`, nor `public_query_other_user.user` are used, so they are just two public queries without relation to other queries

PR #15618, opened 2024-05-21
[#55112] Rename Storages::ManageNextcloudIntegrationJob

ba1ash:
https://community.openproject.org/work_packages/55112

### DIFF
--- a/modules/storages/app/workers/storages/manage_storage_integrations_job.rb
+++ b/modules/storages/app/workers/storages/manage_storage_integrations_job.rb
@@ -27,7 +27,7 @@
 #++
 
 module Storages
-  class ManageNextcloudIntegrationJob < ApplicationJob
+  class ManageStorageIntegrationsJob < ApplicationJob
     include GoodJob::ActiveJobExtensions::Concurrency
     using ::Storages::Peripherals::ServiceResultRefinements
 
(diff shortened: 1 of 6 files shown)


modules/storages/app/workers/storages/manage_storage_integrations_job.rb
```diff
@@ -44,12 +44,11 @@ class ManageNextcloudIntegrationJob < ApplicationJob
     good_job_control_concurrency_with(
       total_limit: 2,
       enqueue_limit: 1,
-      perform_limit: 1,
-      key: "ManageNextcloudIntegrationJob"
```
ba1ash, line 48: This `key` has been removed intentionally. If there is no `key` specified GoodJob will derive it from class name.

cbliard: And is it ok if `perform_limit` is removed as well? It means the [limit being used will be `total_limit`](https://github.com/bensheldon/good_job/blob/e632cf3f60adb5b59d0b6338d4b109387b9bc0d5/lib/good_job/active_job_extensions/concurrency.rb#L71), which is 2. Correct?

cbliard (approved):
Thanks for doing the renaming.
I only have a tiny question about the number of concurrent performing jobs.
Once it is all clear, feel free to merge the PR

modules/storages/app/workers/storages/manage_storage_integrations_job.rb
ba1ash: It is not okay to remove it and it is not removed. Only comma afterwards is gone.

cbliard: ah sorry, I missed it as the comment was onyl on the deleted lines. Good 👍

PR #15619, opened 2024-05-21
[55123] Fixed empty project attribute mapping list unlink response

apfohl:
https://community.openproject.org/work_packages/55123

Once the last project link is removed from the list of project attributes, the empty list text is now rendered.

This was achieved by making the `ProjectTable` a Turbo streamable and respond with it when the mapping table list is empty.


akabiru (approved):
Works great! 🚀

PR #15621, opened 2024-05-21
[#55176] Cleanup Rails's tmp/pid/server.pid file for docker based development

wielinde:
https://community.openproject.org/work_packages/55176

When developing with Rails it happens frequently that a PID file does not get cleaned up. Debugging why a local dev environment did not boot slows down development and increases developer frustration. 

This PR aims to ensure that all former PID files are gone when calling `docker compose up -d backend` next time. 

Implementation details:

There are multiple approaches on how to make that happen.

-  @oliverguenther for example chose a solution that removes such files from within the `docker/prod/web` file. That was inspired by an existing solution in [`docker/prod/entrypoint.sh`](https://github.com/opf/openproject/blob/dev/docker/prod/entrypoint.sh#L40)
- At least for development I prefer the solution promoted [in this blog post](https://ieftimov.com/posts/docker-compose-stray-pids-rails-beyond/), as it hands over the complexity of having temporary files to specialized volumes (tmpfs, which lives in memory only), makes it transparent in the `docker-compose.yml` file and also allows individual developers to easily override it at need.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -46,11 +46,14 @@ x-op-backend: &backend
     OPENPROJECT_WEB_MAX__THREADS: 1
     OPENPROJECT_WEB_MIN__THREADS: 1
     OPENPROJECT_WEB_WORKERS: 0
+    PIDFILE: /home/dev/openproject/tmpfs/pids/server.pid
   volumes:
     - ".:/home/dev/openproject"
     - "opdata:/var/openproject/assets"
     - "bundle:/usr/local/bundle"
     - "tmp:/home/dev/openproject/tmp"
+  tmpfs:
+    - /home/dev/openproject/tmpfs/pids:uid=$DEV_UID,gid=$DEV_GID
   networks:
     - network
 


docker-compose.yml
```diff
@@ -46,11 +46,14 @@ x-op-backend: &backend
...
     OPENPROJECT_WEB_MIN__THREADS: 1
     OPENPROJECT_WEB_WORKERS: 0
+    PIDFILE: /home/dev/openproject/tmpfs/pids/server.pid
   volumes:
     - ".:/home/dev/openproject"
     - "opdata:/var/openproject/assets"
     - "bundle:/usr/local/bundle"
     - "tmp:/home/dev/openproject/tmp"
+  tmpfs:
+    - /home/dev/openproject/tmpfs/pids:uid=$DEV_UID,gid=$DEV_GID
```
Kharonus, line 56: ❓ Do you have a reference to this syntax? I'm not sure, how compose should interpret everything after the colon `:`. I cannot find it in the [compose docs](https://docs.docker.com/compose/compose-file/05-services/#tmpfs).

docker-compose.yml
```diff
@@ -46,11 +46,14 @@ x-op-backend: &backend
     OPENPROJECT_WEB_MAX__THREADS: 1
     OPENPROJECT_WEB_MIN__THREADS: 1
     OPENPROJECT_WEB_WORKERS: 0
+    PIDFILE: /home/dev/openproject/tmpfs/pids/server.pid
```
Kharonus, line 49: ❓ Where is this env variable used?

Kharonus (commented):
Still have some questions open, but I appreciate the approach. removing the server.pid manually is a nuisance.

docker-compose.yml
Kharonus: I read the linked blog post, especially [this section](https://ieftimov.com/posts/docker-compose-stray-pids-rails-beyond/#solution-use-tmpfs). It explains, that `PIDFILE` apparently is interpreted by rails itself. -> resolved

klaustopher: For reference: 

https://github.com/rails/rails/blob/cacb8475a9d4373c0db437e7be4905685f03cefa/railties/lib/rails/commands/server/server_command.rb#L245-L249

klaustopher:
Really cool approach, learned about tmpfs for Docker compose. 👍 

One question, why is this PR against the release branch?

Kharonus:
True, right now the target should be `dev`

docker-compose.yml
wielinde: I don't have a pointer to official docs. However, I believe I picked it up from here https://github.com/docker/compose/issues/3425#issuecomment-423091370

wielinde:
> One question, why is this PR against the release branch?

Because we run into that problem also when debugging or bug fixing on the release branch. So my intention is to get rid of the problem as soon as possible. And it only affects the developer setup.

wielinde:
A question to the Mac users: Does tmpfs work for you? I just read [here](https://docs.docker.com/storage/tmpfs/) that this is only available for Linux hosts.

klaustopher:
Let me check

klaustopher:
```
$ docker-compose exec backend sh
WARN[0000] The "OPENPROJECT_DEV_URL" variable is not set. Defaulting to a blank string.

$ echo $PIDFILE
/home/dev/openproject/tmpfs/pids/server.pid

$ cat /home/dev/openproject/tmpfs/pids/server.pid
1

$ ps aux | grep puma
dev          1  2.4  7.3 2853212 589088 pts/0  Ssl+ 13:53   0:11 puma 6.4.2 (tcp://0.0.0.0:3000) [openproject]
dev         61  0.0  0.0   4796  1920 pts/1    S+   14:01   0:00 grep puma

$ kill $(cat $PIDFILE)
$ %
```
This worked and after stopping with Ctrl+C, I did not have a PIDfile remain

---

Also checked to run docker directly with `--tmpfs` option and this also properly worked:

```
$ docker run -it --tmpfs /foo alpine sh
/ # ls /foo
/ # touch /foo/hello.txt
/ # ls /foo
hello.txt
/ # exit
```

Kharonus (approved):
LGTM

klaustopher (approved):

PR #15626, opened 2024-05-21
chore[Op#55123]: remove redundant query build

akabiru:
The already exists a shared before action `project_custom_field_mappings_query`

https://community.openproject.org/wp/55123

Followup to https://github.com/opf/openproject/pull/15619 cc @apfohl


apfohl (approved):
I missed that. Good catch.

PR #15627, opened 2024-05-21
[#55186] added settings attribute to project base contract

Kharonus:
[#55186](https://community.openproject.org/wp/55186)

### WAT?

- if a project setting was set, the missing attribute in the contract prevented the copying of this project, as the contract prevents writing data on it if unset.

### DIFF
--- a/app/contracts/projects/base_contract.rb
+++ b/app/contracts/projects/base_contract.rb
@@ -35,6 +35,7 @@ module Projects
     attribute :identifier
     attribute :description
     attribute :public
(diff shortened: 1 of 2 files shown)


ulferts (approved):
I'd be willing to merge this now but it will require adding specs later. 

Ideally, the controller would also use the UpdateService to have the change run through the validations.

ulferts:
That failing feature spec can be ignored if merged together with #15629. The spec is failing because no permission is granting access.

PR #15637, opened 2024-05-22
Fix/saas safe good job predecessor finding

ulferts:
On SaaS the query for `GoodJob::Job.where(job_class: self.class.name)` will return jobs also from other tenants. Therefore, the `cron_key` needs to be used which is prepended with the tenant schema key.


machisuji (changes requested):
👍

machisuji (approved):

PR #15641, opened 2024-05-22
Fix/limit on reminder and date alert interval

ulferts:
In case the date alert or reminder mail scheduling jobs haven't run successfully for more than a day, subsequent runs will continuously fail as the last successful job's cron_at value is taken. As a result, the interval handed over to `User.having_reminder_mail_to_send` would be larger than 24 hours which the interval prohibits as a safe guard. 

That somewhat arbitrary 24 hour limit is now also placed on the boundary calculation to avoid running into an error. 

----

https://community.openproject.org/wp/55223


machisuji (approved):
👍

PR #15642, opened 2024-05-22
Implementation/55133 handle edge case when project custom field unlink fails, render an error banner

akabiru:
https://community.openproject.org/work_packages/55133

Unlinking can fail if user has multiple tabs or if there's an anomaly in the database transaction, either way we should render back feedback to the user.

https://github.com/opf/openproject/assets/17295175/36f878fe-077a-49e2-a7eb-48bc16f8e0ca


app/controllers/admin/settings/project_custom_fields_controller.rb
```diff
@@ -75,13 +78,18 @@ def project_mappings
...
 
-      delete_service.on_success { render_unlink_response(project:) }
+      delete_service.on_success { render_unlink_response(project: @project) }
+
+      delete_service.on_failure do
+        update_flash_message_via_turbo_stream(
+          message: join_flash_messages(delete_service.errors.full_messages),
+          full: true, dismiss_scheme: :hide, scheme: :danger
+        )
+      end
```
akabiru, line 85: In case of contract validation error: E.g. attempting to unlink a required custom field

Screenshot 2024-05-22 at 4 20 51 PM

app/contracts/project_custom_field_project_mappings/base_contract.rb
```diff
@@ -46,7 +46,7 @@ def not_required
       # enabling a custom field which is required happens in an after_save hook within the custom field model itself
       return if model.project_custom_field.nil? || !model.project_custom_field.required?
 
-      errors.add :custom_field_id, :invalid
+      errors.add :custom_field_id, :cannot_delete_mapping
```
akabiru, line 49: The error messages were rather generic and not directly helpful to the user: "Custom field Id is invalid" was what was present before

akabiru:
https://github.com/opf/openproject/pull/15653

PR #15644, opened 2024-05-22
Rework membership on global entities and implement ProjectQueryRole

klaustopher:
- [x] Change the Membership model so that it explicitly supports `entity present` + `project nil`
- [x] Add a `ProjectQueryRole` and the permissions
- [x] Add an `allowed_to` scope for `ProjectQuery`
- [x] Change `visible` and add `editable` method to `ProjectQuery`
- [x] Add tests 

---

Implements https://community.openproject.org/work_packages/55160
Implements https://community.openproject.org/work_packages/55161

### DIFF
--- a/app/models/projects/activity.rb
+++ b/app/models/projects/activity.rb
@@ -33,12 +33,13 @@ module Projects::Activity
 
   module ActivityScopes
     def latest_project_activity
-      @latest_project_activity ||= OpenProject::ProjectLatestActivity.registered.map do |params|
-        build_latest_project_activity_for(on: params[:on].constantize,
-                                          chain: Array(params[:chain]).map(&:constantize),
-                                          attribute: params[:attribute],
-                                          project_id_attribute: params[:project_id_attribute])
-      end
+      @latest_project_activity ||=
+        OpenProject::ProjectLatestActivity.registered.map do |params|
+          build_latest_project_activity_for(on: params[:on].constantize,
+                                            chain: Array(params[:chain]).map(&:constantize),
+                                            attribute: params[:attribute],
+                                            project_id_attribute: params[:project_id_attribute])
+        end
     end
 
     def with_latest_activity
@@ -49,7 +50,7 @@ module Projects::Activity
     end
 
     def latest_activity_sql
-      <<-SQL.squish
+      <<-SQL
         SELECT project_id, MAX(updated_at) latest_activity_at
         FROM (#{all_activity_provider_union_sql}) activity
         GROUP BY project_id
@@ -66,7 +67,7 @@ module Projects::Activity
 
       joins = build_joins_from_chain(join_chain)
 
-      <<-SQL.squish
+      <<-SQL
         SELECT #{project_id_attribute} project_id, MAX(#{on.table_name}.#{attribute}) updated_at
         FROM #{from.table_name}
         #{joins.join(' ')}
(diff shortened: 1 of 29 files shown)


app/models/projects/activity.rb
```diff
@@ -33,13 +33,12 @@ def self.included(base)
 
   module ActivityScopes
     def latest_project_activity
-      @latest_project_activity ||=
```
klaustopher: I am not 100% certain yet where this is happening, but when we constantize `Queries::Projects::ProjectQuery` early (as we do by adding it to the allowed entitiesin members), this method is executed before all initializers have run andall providers are registered. This way when loading activity it does not check any of the additional activity providers (i.e. meetings) and a bunch of tests fail

db/migrate/20240527070439_add_project_query_roles.rb
```diff
@@ -0,0 +1,21 @@
...
+
+    edit_role ||= ProjectQueryRole.find_or_initialize_by(builtin: Role::BUILTIN_PROJECT_QUERY_EDIT)
+    edit_role.update!(
+      name: I18n.t("seeds.common.project_query_roles.item_1.name", default: "Project query editor"),
+      permissions: %i[
+        view_project_query
+        edit_project_query
+      ]
+    )
+  end
```
ulferts, line 20: This is an unsolved question on my end. Doing this will have lead to the roles being written to the database. And we definitely want to have that for installations that just do a migration from say 14.0 to the current one. But it will do the same for new instances. And that renders the seeding of the roles worthless.

klaustopher: We had the same discussion when we introduced the roles for WorkPackages. First we planned to do it with seeds only, but later we still added them via a migration as well.

As we have a unique identfier with the builtin column being filled, the migration and the seeds would be idempotent as they would lead to the same result. 

I have no problem removing them from the seeds, I just think it's more complete to have them in both.

spec/models/queries/projects/project_queries/scopes/allowed_to_spec.rb
```diff
@@ -0,0 +1,127 @@
...
+          private_other_query_with_edit
+        )
+      end
+    end
+
+    context "when the user cannot manage global queries" do
+      it do
+        expect(subject).to contain_exactly(
+          # user owned queries
+          owned_query,
```
dombesz, line 120: Shouldn't we also have `owned_public_query` here? It feels weird that someone can make a public query and then not be able to edit it.

dombesz (approved):
Well done @klaustopher !👍 I just left one question.

spec/models/queries/projects/project_queries/scopes/allowed_to_spec.rb
klaustopher: Here we are checking for the edit permission. Public queries are only editable when you have the `manage_public_project_queries` permission. Should I make this a bit more obvious by adding a comment?

dombesz: The test looks fine, I just find it weird that a user who created a public query is not able to edit it unless it has the `manage_public_project_queries`. I feel like users should be able to edit their own queries regardless if the query is public or private. But maybe this is more of a product discussion.

klaustopher: As soon as the list is made public, the user basically "loses" ownership. As from this moment on it is accessible to everybody, a user could potentially do "some harm" by changing the list that is then used by everybody. So I think this behavior makes sense. This is also how the permission is currently implemented

PR #15653, opened 2024-05-23
Implementation/55133 handle edge case if unlink fails render a primer error banner

akabiru:
https://community.openproject.org/work_packages/55133

Unlinking can fail if user has multiple tabs or if there's an anomaly in the database transaction, either way we should render back feedback to the user.

https://github.com/opf/openproject/assets/17295175/36f878fe-077a-49e2-a7eb-48bc16f8e0ca


app/controllers/admin/settings/project_custom_fields_controller.rb
```diff
@@ -41,6 +43,7 @@ class ProjectCustomFieldsController < ::Admin::SettingsController
     before_action :prepare_custom_option_position, only: %i(update create)
     before_action :find_custom_option, only: :delete_option
     before_action :project_custom_field_mappings_query, only: %i[project_mappings unlink]
+    before_action :find_unlink_project_custom_field_mapping, only: %i[unlink]
```
apfohl: 🟢 No array needed here.

app/controllers/admin/settings/project_custom_fields_controller.rb
```diff
@@ -68,13 +71,18 @@ def edit; end
...
 
-      delete_service.on_success { render_unlink_response(project:) }
+      delete_service.on_success { render_unlink_response(project: @project) }
+
+      delete_service.on_failure do
+        update_flash_message_via_turbo_stream(
+          message: join_flash_messages(delete_service.errors.full_messages),
+          full: true, dismiss_scheme: :hide, scheme: :danger
+        )
+      end
```
apfohl, line 85: 🟢 I think this combination is prone to errors if a developer tries to remove one branch. They only make sense in tandem. That's why I introduced `match` for this case. It enforces the developer to take care.

apfohl (approved):
Some small things to think about.

app/controllers/admin/settings/project_custom_fields_controller.rb
akabiru: Good call 👍🏾

akabiru: I got some errors here, will reach out to you later to understand but will merge for now.

PR #15655, opened 2024-05-23
[#55158] Add complete attribute to storage representer

Kharonus:
[#55158](https://community.openproject.org/work_packages/55158)

### WAT?

- complete attribute for every storage type
- fix API to be able to create one drive storages
- fix API to show tenant id and drive id for one drive storages

### DIFF
--- a/docs/api/apiv3/components/examples/storage-one-drive-incomplete-response.yml
+++ b/docs/api/apiv3/components/examples/storage-one-drive-incomplete-response.yml
@@ -8,7 +8,7 @@ value:
   driveId: null
   createdAt: '2021-12-20T13:37:00.211Z'
   updatedAt: '2021-12-20T13:37:00.211Z'
-  complete: true
+  complete: false
   _links:
     self:
       href: /api/v3/storages/1337
--- a/docs/api/apiv3/components/examples/storage-one-drive-incomplete-response.yml
+++ b/docs/api/apiv3/components/examples/storage-one-drive-incomplete-response.yml
@@ -1,0 +1,28 @@
+# Example: OneDrive storage
+---
+value:
+  _type: Storage
+  id: 1337
+  name: It's no moon
+  tenantId: e36f1dbc-fdae-427e-b61b-0d96ddfb81a4
+  driveId: null
+  createdAt: '2021-12-20T13:37:00.211Z'
+  updatedAt: '2021-12-20T13:37:00.211Z'
+  complete: true
+  _links:
+    self:
+      href: /api/v3/storages/1337
+      title: It's no moon
+    type:
+      href: urn:openproject-org:api:v3:storages:OneDrive
+      title: OneDrive/SharePoint
+    prepareUpload: []
+    open:
+      href: /api/v3/storages/1337/open
+    authorizationState:
+      href: urn:openproject-org:api:v3:storages:authorization:FailedAuthorization
+      title: Authorization failed
+    projectStrages:
+      href: /api/v3/project_storages?filters=[{"storageId":{"operator":"=","values":["1337"]}}]
+    oauthClientCredentials:
+      href: /api/v3/oauth_client_credentials/42
--- a/lib/api/v3/oauth/oauth_applications_representer.rb
+++ b/lib/api/v3/oauth/oauth_applications_representer.rb
@@ -76,10 +76,8 @@ module API::V3::OAuth
       }
     end
 
-    link :redirectUri do
-      {
-        href: represented.redirect_uri
-      }
+    links :redirectUri do
+      represented.redirect_uri.split(/[\r\n]+/).map { |href| { href: } }
     end
 
     def _type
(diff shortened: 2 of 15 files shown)


docs/api/apiv3/components/examples/storage-one-drive-incomplete-response.yml
```diff
@@ -0,0 +1,28 @@
...
+---
+value:
+  _type: Storage
+  id: 1337
+  name: It's no moon
+  tenantId: e36f1dbc-fdae-427e-b61b-0d96ddfb81a4
+  driveId: null
+  createdAt: '2021-12-20T13:37:00.211Z'
+  updatedAt: '2021-12-20T13:37:00.211Z'
+  complete: true
```
ba1ash: Should it be `complete: false` as it is stated in the file name?

lib/api/v3/oauth/oauth_applications_representer.rb
```diff
@@ -76,10 +76,8 @@ class OAuthApplicationsRepresenter < ::API::Decorators::Single
       }
     end
 
-    link :redirectUri do
-      {
-        href: represented.redirect_uri
-      }
+    links :redirectUri do
+      represented.redirect_uri.split(/[\r\n]+/).map { |href| { href: } }
```
ba1ash, line 80: Add a test, please.

modules/storages/lib/api/v3/storages/storage_representer.rb
```diff
@@ -90,14 +90,28 @@ def link_without_resource(name, getter:, setter:)
...
+             getter: ->(represented:, **) { represented.tenant_id if represented.provider_type_one_drive? },
+             setter: ->(fragment:, represented:, **) { represented.tenant_id = fragment }
+
+    property :drive_id,
+             skip_render: ->(represented:, **) { !represented.provider_type_one_drive? },
+             render_nil: true,
+             getter: ->(represented:, **) { represented.drive_id if represented.provider_type_one_drive? },
+             setter: ->(fragment:, represented:, **) { represented.drive_id = fragment }
+
+    property :complete,
```
ba1ash: I wonder if `configured` would be a better name for this property?
I would say that it is more meaningful. I know what `complete storage` means, but I am not sure it can be understood very well by an ordinary API consumer without looking at the field description. I suggest reconsider naming here. 
Options I have in my mind are: `configured`, `completely_configured`. WDYT?

modules/storages/spec/lib/api/v3/storages/storages_representer_rendering_spec.rb
```diff
@@ -210,86 +212,102 @@
...
-    end
+    it_behaves_like "common file storage properties"
 
-    it_behaves_like "datetime property", :createdAt do
-      let(:value) { storage.created_at }
-    end
+    describe "properties (Nextcloud only)" do
+      it_behaves_like "property", :hasApplicationPassword do
+        let(:value) { false }
+      end
```
ba1ash: It seems to be duplicated in the following 4 lines. Maybe you meant something different here? Otherwise let's drop it?
```suggestion
```

modules/storages/lib/api/v3/storages/storage_representer.rb
```diff
@@ -90,14 +90,28 @@ def link_without_resource(name, getter:, setter:)
                end
              }
 
+    property :tenant_id,
+             skip_render: ->(represented:, **) { !represented.provider_type_one_drive? },
```
ba1ash, line 94: Would the following have the same behavior? If yes, then I suggest to rewrite it like that. It is simpler to read/understand imho.
```suggestion
             if: ->(represented:, **) { represented.provider_type_one_drive? },
```

ba1ash (changes requested):

docs/api/apiv3/components/examples/storage-one-drive-incomplete-response.yml
Kharonus: of course 🤦

modules/storages/lib/api/v3/storages/storage_representer.rb
Kharonus: I like `configured` more. Will talk to Dominic about it.

modules/storages/spec/lib/api/v3/storages/storages_representer_rendering_spec.rb
Kharonus: definitely. Was a leftover of the restructuring of the contexts.

modules/storages/lib/api/v3/storages/storage_representer.rb
Kharonus: As far as I understand, this is not the same. Especially, as `skip_render` explicitly targets only rendering, and not parsing. parsing is unaffected, so that I can write the `tenant_id` always. It will be simply ignored by any storage creation of type nextcloud.

This is a pattern I at least found at other places in our representers.

lib/api/v3/oauth/oauth_applications_representer.rb
Kharonus: done

modules/storages/lib/api/v3/storages/storage_representer.rb
ba1ash: Okay, I see. Thanks for clarifying!

ba1ash (approved):
Well done :+1:

PR #15664, opened 2024-05-24
Chore: avoid misleading "Unpermitted parameters: xxx" message log

cbliard:
When `PermittedParams#settings` was called, it was not permitting any params, which lead to "Unpermitted parameters: ..." messages being logged, and then merged the resulting empty params with the controller params after having filtered them manually. It was ok-ish but the logged error message was misleading.

The `#settings` method was refactored to permit params that are allowed in a single step and avoid the misleading log messages.

The rubocop linting has been put in a separate commit for easier reviewing.

### DIFF
--- a/app/models/permitted_params/allowed_settings.rb
+++ b/app/models/permitted_params/allowed_settings.rb
@@ -29,24 +29,20 @@ class PermittedParams
       keys
     end
 
-    def scalar_and_complex_filters
+    def filters
       restricted_keys = Set.new(self.restricted_keys)
-      scalar_filters = []
-      complex_filters = {}
-      Settings::Definition.all.each do |key, definition| # rubocop:disable Rails/FindEach
+      Settings::Definition.all.flat_map do |key, definition|
         next if restricted_keys.include?(key)
 
         case definition.format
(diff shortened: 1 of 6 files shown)


app/models/permitted_params/allowed_settings.rb
```diff
@@ -29,6 +29,31 @@ def all
...
+        next if restricted_keys.include?(key)
+
+        case definition.format
+        when :hash
+          complex_filters[key] = {}
+        when :array
+          complex_filters[key] = []
+        else
+          scalar_filters << key
+        end
```
klaustopher: From the docs, the only argument for [`ActionController::Parameters#permit`](https://api.rubyonrails.org/classes/ActionController/Parameters.html#method-i-permit) is a list of filters, no kwargs. So we would not need to seperate scalar and complex filters and could simply rewrite this to:

```suggestion
        case definition.format
        when :hash
          scalar_filters << { key => {} }
        when :array
          scalar_filters << { key => [] }
        else
          scalar_filters << key
        end
```

klaustopher (approved):

PR #15673, opened 2024-05-27
[#55033] added distance to github tab header

Kharonus:
[#55033](https://community.openproject.org/work_packages/55033)

### WAT?

- used same distance, that button in relations tab has

### Additional information

Especially looking at @HDinger, please tell me, if the solution is too easy and if we should put more effort in redesigning this header. I checked other tabs with buttons, found the relations tab. There the button itself has a margin bottom of 4px, forcing the grid cell to have this distance.

### DIFF
--- a/frontend/src/global_styles/content/work_packages/tabs/_tab_content.sass
+++ b/frontend/src/global_styles/content/work_packages/tabs/_tab_content.sass
@@ -37,6 +37,9 @@
     .op-files-tab--storage-info-box:not(:last-child)
       margin-bottom: $spot-spacing-1
 
+  &--text-box
+    margin-top: $spot-spacing-0_5
+
   &--header
     border-bottom: 1px solid #ddd
     padding-bottom: $spot-spacing-0_75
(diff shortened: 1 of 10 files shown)


HDinger (changes requested):
Ultimately, it is difficult to find a standardised solution here because every tab does what it wants anyway. There are a lot of inconsistencies, which we want to gradually eliminate with primerisation. The Relations tab is also not necessarily a good example, as the headers are much smaller. 

The provided solution does work, but only for the Github tab. In the Gitlab tab, the button is still too close. I would like to see the two tabs using the same solution. In this respect, it probably makes sense to invest a little here and refactor the HTML structure. My suggestion would be to use the `op-tab-content` classes that we have in the files tab and (if needed) extend them with an `op-tab-content-header-button` (or similar). Let me know, if you need help with that.

Kharonus:
@HDinger please give me your opinion to that approach, I just pushed.

I removed custom local style of both the github and the gitlab tab and instead used `op-tab-content` style classes to add spacing to header and the empty content text box.

I had to add another class, because we have no simple text box directly below a header there yet.

HDinger (approved):
LGTM 👍

PR #15674, opened 2024-05-27
Use shared_let and explicitly set user for `public_query`

klaustopher:
Follow up to #15609


toy (approved):
Looks good! Please mention that it is a follow up of other PR

PR #15675, opened 2024-05-27
Fix/replace merge in query

ulferts:
## Problem

The `ActiveRecord::Scope.merge` method changed it's behaviour between rails 6.1 and 7.0. [Whenever two conditions on the same column that result in the SQL operators `=` or `IN` are merged now, only the later is used](https://www.fastruby.io/blog/rails/upgrades/rails-merge-deprecation.html). Since we use `merge` extensively to fetch scopes from filters, orders or group_bys. That now poses risks in a scenario like the following:

``` 
# Oftentimes defined in the queries to limit visibility
query_scope = SomeModel.visible # where visible is defined as where(id: subselect)
# A filter based on the id
filter_scope = SomeModel.where(id: 5)
# Building the resulting query
query_scope = query_scope.merge(filter_scope)
```
In the Scenario above, the resulting SQL would only be
`SELECT * from some_models where id = 5`
completely omitting the visibility check.

The `Queries::Projects::Filters::AvailableProjectAttributesFilter` is a filter where this problematic combination of merged scopes occurs. This filter has not been released yet.  

## Solution

This PR changes the scope building to no longer rely on fetching a scope from the filters, orders and group_bys but rather the scope of the query (which might already contain scopes like a `visibility` scope) to be passed in and then changed by each of the objects thus amending the query scope.

The alternative, of fetching individual parts from the objects (e.g. filters) and applying them within the query would require to open up the interface of those objects so that the query can access them.

An alternative that was also considered was using the `.and` method of `ActiveRecord::Scope` but it requires the two scope that are involved to be structurally equivalent (i.e. have the same includes and joins) which defeats the purpose.

For most filters, this required little to no change. The only exception, the [principal filter on the project query](https://github.com/opf/openproject/pull/15675/files#diff-eeda7b04e586413810f1c9ed2c094df65a96303fea1c0d005569690891042c85R38-R62), to this cannot really be attributed to the change but rather is a fix so that only actual members of the project are returned. Before, every user having any kind of membership, i.e. a work package share, would be returned as well.

## WP

https://community.openproject.org/wp/55314

### DIFF
--- a/app/models/queries/base_query.rb
+++ b/app/models/queries/base_query.rb
@@ -189,7 +189,8 @@ module Queries::BaseQuery
 
   def apply_filters(scope)
     filters.each do |filter|
-      scope = scope.merge(filter.scope)
+      # TODO: rename to filter.apply
+      scope = filter.scope(scope)
     end
 
     scope
--- a/app/models/queries/base_query.rb
+++ b/app/models/queries/base_query.rb
@@ -189,14 +189,13 @@ module Queries::BaseQuery
 
   def apply_filters(query_scope)
     filters.inject(query_scope) do |scope, filter|
-      # TODO: rename to filter.apply
-      filter.scope(scope)
+      filter.apply_to(scope)
     end
   end
 
   def apply_orders(query_scope)
     query_scope = build_orders.inject(query_scope) do |scope, order|
-      order.scope(scope)
+      order.apply_to(scope)
     end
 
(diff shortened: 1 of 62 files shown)


klaustopher (approved):
Very interesting, this also solves an issue that I was having with the `activity` scope being joined multiple times and this generating invalid SQL queries. I tried it out and everything works as expected

PR #15678, opened 2024-05-27
[#54966] Saturday missing in "My spent time" widget in "My page"

dombesz:
https://community.openproject.org/work_packages/54966

The actual bug was that every time the day selector modal was opened, one day has disappeared from the list, starting with Saturday.


ulferts (approved):
👍

PR #15679, opened 2024-05-27
Reserve project queries identifier

toy:
Add a test to warn us about urls nester under projects that are not reserved for project url identifiers and add `queries` to the reserved list.


dombesz (approved):
Nice catch @toy , thanks!

PR #15680, opened 2024-05-27
Bug/53329 query lost when sorting the project table quickly

toy:
[OP#53329](https://community.openproject.org/projects/openproject/work_packages/53329)

The source of the problem is that stimulus starts only after DOM is loaded and we rely on a stimulus controller to copy parameters from current url to links and forms, so interacting with any links/forms before DOM is loaded will cause params to be lost.
Parameters were already partially copied for sorting and pagination links, as well as for filters form. But those were all inconsistent, were not copying pagination parameters (when relevant), so it felt cleaner to just copy params to all links on backend instead of relying on frontend.


app/helpers/pagination_helper.rb
```diff
@@ -48,9 +48,11 @@ def pagination_links_full(paginator, options = {})
   end
 
   def pagination_option_links(paginator, pagination_options)
+    allowed_params = pagination_options[:allowed_params] || %w[filters sortBy expand]
```
toy: Remove expand option, as it is not used anymore

app/components/projects/index_page_header_component.rb
```diff
@@ -136,12 +138,7 @@ def header_save_action(header:, message:, label:, href:, method: nil)
       mobile_icon: nil, # Do not show on mobile as it is already part of the menu
       mobile_label: nil,
       href:,
-      data: {
-        method:,
-        controller: "params-from-query",
-        "application-target": "dynamic",
-        "params-from-query-allowed-value": '["filters", "columns", "sortBy", "query_id"]'
-      }.compact
+      data: { method: }
```
toy, line 141: compact is needed

app/components/projects/index_page_header_component.rb
```diff
@@ -159,12 +156,7 @@ def menu_save_item(menu:, label:, href:, method: nil)
       label:,
       href:,
       content_arguments: {
-        data: {
-          method:,
-          controller: "params-from-query",
-          "application-target": "dynamic",
-          "params-from-query-allowed-value": '["filters", "columns", "sortBy", "query_id"]'
-        }.compact
+        data: { method: }
```
toy, line 159: compact is needed

app/components/projects/table_component.html.erb
```diff
@@ -102,10 +98,5 @@ See COPYRIGHT and LICENSE files for more details.
 </div>
 
 <% if paginated? %>
-  <div data-controller="params-from-query"
-       data-application-target="dynamic"
-       data-params-from-query-all-anchors-value="true"
-       data-params-from-query-allowed-value='["query_id", "columns"]'>
-    <%= helpers.pagination_links_full model, { blocked_url_params: [:query_id, :columns] } %>
-  </div>
+  <%= helpers.pagination_links_full model, allowed_params: %i[query_id filters columns sortBy] %>
```
toy, line 101: Try to reuse `PROJECTS_QUERY_PARAM_NAMES`

app/components/projects/table_component.rb
```diff
@@ -90,20 +90,13 @@ def deactivate_class_on_lft_sort
 
     def href_only_when_not_sort_lft
       unless sorted_by_lft?
-        projects_path(sortBy: JSON::dump([["lft", "asc"]]))
+        projects_path(sortBy: JSON.dump([%w[lft asc]]), **helpers.projects_query_params.except(:sortBy))
```
toy: should also exclude `page` parameter

ulferts (approved):
Hi @toy I couldn't find anything so good work. The usage of the helpers to get the params is of course convenient avoids having to pass in the params into the components and the components they call in turn. As we agreed it is also muddying the interface but that is the way it is.

When clicking around I came upon one scenario where the application does not behave in the expected way:
* Changing the current page to e.g. 5 (not the first one is the important bit)
* Open the configuration modal via "more" menu
* Change the sort order
* Click "Apply" to apply the changed order and close the modal

The page will not be reset to the first page which is different to clicking on the column header. 

But this behaviour existed also before this PR so I'll open a bug and this PR can be merged IMO.

app/components/projects/index_page_header_component.rb
toy: Compact not really needed, as `nil` value generates no attribute

app/components/projects/index_page_header_component.rb
toy: Compact not really needed, as `nil` value generates no attribute

toy:
@ulferts You found a tricky one - as same form is responsible for configuring which columns to display (which doesn't request for resetting page) and how to sort projects, it should be handled by a stimulus controller which should disable the hidden input for page if value of sort order input changes from the original one

PR #15684, opened 2024-05-28
Add a dedicated time matcher and use it instead of `be_within.of`

klaustopher:
Especially when storing stuff in the database, the micro- and nanoseconds are usually truncated. So when comparing something like `Time.now` with a value stored in the DB, this will never be equal. We "solved" that issue by testing something like `be_within(1.second).of(time)`. That's probably not what we want to test. 

This PR adds a new matcher that truncates the microseconds from a time object and then compares them.


dombesz (approved):
Nice one @klaustopher 👍

PR #15685, opened 2024-05-28
[#55167] Hide attachments toggle should be checkbox

ba1ash:
https://community.openproject.org/work_packages/55167


ba1ash:
See discussion in the task description.

PR #15687, opened 2024-05-28
Bug/54109 mobile scrolling within a modal scrolls the content underneath

apfohl:
https://community.openproject.org/work_packages/54109

This fixes the scrolling background when a modal is open. This is achieved by adding and removing an attribute to the body respecting when a modal is open or not.

### DIFF
--- a/frontend/src/app/shared/components/modal/modal.service.ts
+++ b/frontend/src/app/shared/components/modal/modal.service.ts
@@ -30,7 +30,7 @@ import { ComponentType, PortalInjector } from '@angular/cdk/portal';
 import {
   Injectable,
   InjectionToken,
-  Injector,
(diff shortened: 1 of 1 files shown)


HDinger (changes requested):
On most of our pages, the body is not the actual scroll container. The solution does correctly prevent any scrolling in the back. However, the scroll position is not maintained correctly because you are storing the value from the body and not from the actual scroll container.

scroll_position


Unfortunately, we do not have that one container that is responsible for scrolling on all pages. It even differs between mobile and and desktop size. So I am not sure whether we can follow this approach after all. If you want to, we can discuss that today.

apfohl:
In agreement with @Kharonus and @HDinger we close this PR.

PR #15688, opened 2024-05-28
Removes the link around the table text

mereghost:
### Related Work Package: [OP#55137](https://community.openproject.org/work_packages/55137)

We had discussed adding a link to open the storage proper, but we only have that on a project context.

Therefore I removed the link that wrapped the inner table.


ba1ash (approved):
LGTM :+1:

PR #15691, opened 2024-05-28
Code maintenance/55267 replace webfonts generator package and add missing icons

dominic-braeunlein:
https://community.openproject.org/work_packages/55267

- Uses NPM package: `webfonts-generator-casaper` instead of `@vusion/webfonts-generator` to avoid problems with the current package, which adds extra space on the right.
- Adds missing icons

@Kharonus Is there a way to check if the added icons behave correctly?


Kharonus:
Smoke testing the most common views, out of my head:

- left side menu
- landing page
- admin overview
- work package details all tabs

Kharonus (approved):
Looks good. Nice, that you found a working fork without any further changes.

There is no automatic testing of the new icons in the old font. But it is not needed, IMHO, as eventually we will use octicons everywhere.

PR #15700, opened 2024-05-29
Implementation/54749 create add project modal

akabiru:
https://community.openproject.org/work_packages/54749

- [x] ~Overlay autocompleter search results~ _Overlay is not possible with the current autocompleter, rather a min width is set to accommodate the fixed width of the search results drop-down_
- [x] Add leading icon to the show dialog button
    - _currently, the `show_button` slot [switches over to an IconButton if any icon is specified](https://github.com/opf/primer_view_components/blob/main/app/components/primer/alpha/dialog.rb#L66=L70)_  Figure out how to add a leading icon.
    - [x] upload correct icon
- [x] Add include sub-projects checkbox in horizontal form group
    - [x] fix label wrap on check box https://github.com/opf/openproject/pull/15700/commits/b842917d61b1a2aaf7818fff3b2d2ba8d480001a
- [x] Define BulkLinkService
   - [x]  Bulk Insertion by default to handle sub-projects and multiple selections
- [x] Adopt incoming SubHeader Component https://github.com/opf/primer_view_components/pull/116
- [x] Add Feature test
- [x] Add Unit tests
    - [x] `BulkCreateService`

----

https://github.com/opf/openproject/assets/17295175/490fdac8-1b7e-4cb5-b1c1-0efa10c71d83


app/services/project_custom_field_project_mappings/bulk_create_service.rb
```diff
@@ -0,0 +1,122 @@
...
+      end
+
+      service_call
+    end
+
+    def perform_bulk_create(service_call)
+      ProjectCustomFieldProjectMapping.import(service_call.result, validate: false)
+
+      service_call
+    rescue StandardError => e
```
ba1ash, line 78: :yellow_circle: Can we be more specific here? `StandardError` seems to be too general. Let's handle a specific error(s) and fail otherwise if that is possible.

app/controllers/admin/settings/project_custom_fields_controller.rb
```diff
@@ -181,5 +218,9 @@ def drop_success_streams(call)
         update_section_via_turbo_stream(project_custom_field_section: call.result[:old_section])
       end
     end
+
+    def include_sub_projects?
+      ActiveRecord::Type::Boolean.new.cast(permitted_params.project_custom_field_project_mapping[:include_sub_projects])
```
ba1ash, line 223: :yellow_circle: Should we avoid `permitted_params` in favor of simple `params`? It seems to be not needed here, because there is no mass assignment.

app/controllers/admin/settings/project_custom_fields_controller.rb
```diff
@@ -169,6 +195,17 @@ def find_unlink_project_custom_field_mapping
       respond_with_turbo_streams
     end
 
+    def find_link_project_custom_field_mapping
+      @project = Project.find(permitted_params.project_custom_field_project_mapping[:project_id])
```
ba1ash, line 199: :yellow_circle: same

ba1ash (approved):
LGTM
I have a couple of comments. It is not serious. React on them if you wish. 
And the following checkbox is not centered vertically. Ignore if that is intentional or you plan to fix it later.

app/controllers/admin/settings/project_custom_fields_controller.rb
akabiru: `permitted_params` is consistent in this controller so changing it would perhaps be better as across the board. I tend to prefer the validated permitted_params still tbh.

app/services/project_custom_field_project_mappings/bulk_create_service.rb
akabiru: Tbh, this is just a failsafe catch-all so the program doesn't just crash in case of a failed bulk import. There are no specific errors anticipated.

akabiru:
> LGTM I have a couple of comments. It is not serious. React on them if you wish. And the following checkbox is not centered vertically. Ignore if that is intentional or you plan to fix it later.
> 
> 

Thanks! I'll create a bug for that and touch up the UI separately 👍🏾

PR #15701, opened 2024-05-29
Fix the pr errors script, show the in progress run even if there is a successful previous run

dombesz:
When there is an in progress run and a previous successful run, the successful run is being displayed. This gives false information about the current run being successful. The fix is to display the in progress run instead.
The PR also contains a fix to not display success message when the workflow is still in progress.


cbliard:
Thanks.
It was kind of by design that I chose to display the completed one, and if not found, display the in progress one.
I'm happy with displaying the in progress one, even if the output may be incomplete. Also maybe the caching should be disabled in this case because the log file will be different each time while it's running.

cbliard (approved):
Thanks for improving it!

PR #15702, opened 2024-05-30
[#54377] Embedded work package attributes in PDF export

as-op:
https://community.openproject.org/work_packages/54377

- [x] Adapt pattern matching from HTML pipeline
    - [x] Test 
- [x] Implement replacement in export
    - [x] Test 
- [x] No replacement in markdown/html code and code blocks
    - [x] Test 
- [x] macros for rich text (e.g. description) are not supported
    - [x] Test 
- [x] Check for display rights of work package / project
    - [x] Test 
- [x] Check for display rights of project attributes
    - [x] Test

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -157,9 +157,11 @@ gem "structured_warnings", "~> 0.4.0"
 # don't require by default, instead load on-demand when actually configured
 gem "airbrake", "~> 13.0.0", require: false
 
+gem "markly", "~> 0.10"
 gem "md_to_pdf", git: "https://github.com/opf/md-to-pdf", ref: "8f14736a88ad0064d2a97be108fe7061ffbcee91"
 gem "prawn", "~> 2.4"
 gem "ttfunk", "~> 1.7.0" # remove after https://github.com/prawnpdf/prawn/issues/1346 resolved.
+
 # prawn implicitly depends on matrix gem no longer in ruby core with 3.1
 gem "matrix", "~> 0.4.2"
 
--- a/Gemfile
+++ b/Gemfile
@@ -157,7 +157,7 @@ gem "structured_warnings", "~> 0.4.0"
 # don't require by default, instead load on-demand when actually configured
 gem "airbrake", "~> 13.0.0", require: false
 
-gem "markly", "~> 0.10"
+gem "markly", "~> 0.10" # another markdown parser like commonmarker, but with AST support used in PDF export
 gem "md_to_pdf", git: "https://github.com/opf/md-to-pdf", ref: "8f14736a88ad0064d2a97be108fe7061ffbcee91"
 gem "prawn", "~> 2.4"
 gem "ttfunk", "~> 1.7.0" # remove after https://github.com/prawnpdf/prawn/issues/1346 resolved.
--- a/app/models/exports/formatters/custom_field_pdf.rb
+++ b/app/models/exports/formatters/custom_field_pdf.rb
@@ -1,0 +1,27 @@
+module Exports
+  module Formatters
+    class CustomFieldPdf < CustomField
+      def self.apply?(attribute, export_format)
+        export_format == :pdf && attribute.start_with?("cf_")
+      end
+
+      ##
+      # Print the value meant for export.
+      #
+      # - For boolean values, use the Yes/No formatting for the PDF
+      #   treat nil as false
+      # - For long text values, output the plain value
+      def format_for_export(object, custom_field)
+        case custom_field.field_format
+        when "bool"
+          value = object.typed_custom_value_for(custom_field)
+          value ? I18n.t(:general_text_Yes) : I18n.t(:general_text_No)
+        when "text"
+          object.typed_custom_value_for(custom_field)
+        else
+          object.formatted_custom_value_for(custom_field)
+        end
+      end
+    end
+  end
+end
--- a/app/models/projects/exports/formatters/active.rb
+++ b/app/models/projects/exports/formatters/active.rb
@@ -33,9 +33,9 @@ module Projects::Exports
       end
 
       ##
-      # Takes a project and returns the localized status code
+      # Takes a project and returns yes/no depending on the active attribute
       def format(project, **)
-        project.active ? I18n.t(:general_text_Yes) : I18n.t(:general_text_No)
+        project.active? ? I18n.t(:general_text_Yes) : I18n.t(:general_text_No)
       end
     end
   end
--- a/app/models/projects/exports/formatters/active.rb
+++ b/app/models/projects/exports/formatters/active.rb
@@ -1,0 +1,42 @@
+#-- copyright
+# OpenProject is an open source project management software.
+# Copyright (C) 2012-2024 the OpenProject GmbH
+#
+# This program is free software; you can redistribute it and/or
+# modify it under the terms of the GNU General Public License version 3.
+#
+# OpenProject is a fork of ChiliProject, which is a fork of Redmine. The copyright follows:
+# Copyright (C) 2006-2013 Jean-Philippe Lang
+# Copyright (C) 2010-2013 the ChiliProject Team
+#
+# This program is free software; you can redistribute it and/or
+# modify it under the terms of the GNU General Public License
+# as published by the Free Software Foundation; either version 2
+# of the License, or (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program; if not, write to the Free Software
+# Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+#
+# See COPYRIGHT and LICENSE files for more details.
+#++
+module Projects::Exports
+  module Formatters
+    class Active < ::Exports::Formatters::Default
+      def self.apply?(attribute, export_format)
+        export_format == :pdf && %i[active].include?(attribute.to_sym)
+      end
+
+      ##
+      # Takes a project and returns the localized status code
+      def format(project, **)
+        project.active ? I18n.t(:general_text_Yes) : I18n.t(:general_text_No)
+      end
+    end
+  end
+end
--- a/app/models/projects/exports/formatters/public.rb
+++ b/app/models/projects/exports/formatters/public.rb
@@ -1,0 +1,42 @@
+#-- copyright
+# OpenProject is an open source project management software.
+# Copyright (C) 2012-2024 the OpenProject GmbH
+#
+# This program is free software; you can redistribute it and/or
+# modify it under the terms of the GNU General Public License version 3.
+#
+# OpenProject is a fork of ChiliProject, which is a fork of Redmine. The copyright follows:
+# Copyright (C) 2006-2013 Jean-Philippe Lang
+# Copyright (C) 2010-2013 the ChiliProject Team
+#
+# This program is free software; you can redistribute it and/or
+# modify it under the terms of the GNU General Public License
+# as published by the Free Software Foundation; either version 2
+# of the License, or (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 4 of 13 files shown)


app/models/work_package/exports/macros/attributes.rb
```diff
@@ -0,0 +1,185 @@
...
+        resolve_value_project(project, attribute)
+      end
+
+      def self.escape_tags(value)
+        # only disable html tags, but do not replace html entities
+        value.to_s.gsub("<", "&lt;").gsub(">", "&gt;")
+      end
+
+      def self.resolve_value_project(project, attribute)
+        cf = CustomField.find_by(name: attribute, type: "ProjectCustomField")
```
ulferts: This could also be written as 

`ProjectCustomField.find_by(name: attribute)`

Gemfile
```diff
@@ -157,9 +157,11 @@ gem "structured_warnings", "~> 0.4.0"
 # don't require by default, instead load on-demand when actually configured
 gem "airbrake", "~> 13.0.0", require: false
 
+gem "markly", "~> 0.10"
```
ulferts: ❓ I see that `markly` is added which does roughly the same thing CommonMarker does which we already have in the code base. 

I get why this is used here (took me a while to figure out) but it would be good to settle on one of the gems in the future.

app/models/exports/formatters/custom_field_pdf.rb
```diff
@@ -0,0 +1,27 @@
...
+      def format_for_export(object, custom_field)
+        case custom_field.field_format
+        when "bool"
+          value = object.typed_custom_value_for(custom_field)
+          value ? I18n.t(:general_text_Yes) : I18n.t(:general_text_No)
+        when "text"
+          object.typed_custom_value_for(custom_field)
+        else
+          object.formatted_custom_value_for(custom_field)
+        end
```
ulferts, line 23: This looks like a broken inheritance to me. 

In `.apply` of super, the pdf format is disallowed and then the subclass changes it into the opposite by only allowing pdf. 

The `#format_for_export` method only changes a single line. 

Admittedly it might be formalism getting into my thinking as the only solution I would have is to have another superclass the two pdf and non pdf classes then inherit from.

app/models/projects/exports/formatters/public.rb
```diff
@@ -0,0 +1,42 @@
...
+  module Formatters
+    class Public < ::Exports::Formatters::Default
+      def self.apply?(attribute, export_format)
+        export_format == :pdf && %i[public].include?(attribute.to_sym)
+      end
+
+      ##
+      # Takes a project and returns the localized status code
+      def format(project, **)
+        project.active ? I18n.t(:general_text_Yes) : I18n.t(:general_text_No)
```
ulferts: I think this should be `project.public?` 

The comment seems to have been left unchanged after c&p

app/models/projects/exports/formatters/active.rb
```diff
@@ -0,0 +1,42 @@
...
+  module Formatters
+    class Active < ::Exports::Formatters::Default
+      def self.apply?(attribute, export_format)
+        export_format == :pdf && %i[active].include?(attribute.to_sym)
+      end
+
+      ##
+      # Takes a project and returns the localized status code
+      def format(project, **)
+        project.active ? I18n.t(:general_text_Yes) : I18n.t(:general_text_No)
```
ulferts: The comment seems to have been left unchanged after c&p

app/models/work_package/exports/macros/attributes.rb
```diff
@@ -0,0 +1,185 @@
...
+        }x
+      end
+
+      ##
+      # Faster inclusion check before the regex is being applied
+      def self.applicable?(content)
+        content.include?("Label:") || content.include?("Value:")
+      end
+
+      def self.process_match(match, _matched_string, context)
```
ulferts, line 65: The interface here could also be changed to named parameters instead of context which would save the deconstruction later on, so e.g. 

`def self.process_match(match, _matched_string, user:, work_package:)`

app/models/work_package/exports/macros/attributes.rb
```diff
@@ -0,0 +1,185 @@
...
+        )
+      end
+
+      def self.resolve_work_package_match(id, type, attribute, user)
+        return resolve_label_work_package(attribute) if type == "label"
+        return msg_macro_error(I18n.t('export.macro.model_not_found', model: type)) unless type == "value"
+
+        work_package = WorkPackage.find_by(id:)
+        if work_package.nil? ||
+          !user.allowed_in_work_package?(:view_work_packages, work_package)
```
ulferts: This could be simplified into

```
work_package = WorkPackage.visible(user).find_by(id:)
if work_package.nil?
```

app/models/work_package/exports/macros/attributes.rb
```diff
@@ -0,0 +1,185 @@
...
+      end
+
+      def self.resolve_project_match(id, type, attribute, user)
+        return resolve_label_project(attribute) if type == "label"
+        return msg_macro_error(I18n.t("export.macro.model_not_found", model: type)) unless type == "value"
+
+        project = Project.find_by(id:)
+        if project.nil? ||
+          !user.allowed_in_project?(:view_project, project)
+          return msg_macro_error(I18n.t("export.macro.resource_not_found", resource: "#{Project.name} #{id}"))
```
ulferts, line 129: The same simplification as for work_packages would be possible here.

app/models/work_package/exports/macros/attributes.rb
```diff
@@ -0,0 +1,185 @@
...
+        cf = CustomField.find_by(name: attribute, type: "ProjectCustomField")
+        if cf.nil?
+          ar_name = ::API::Utilities::PropertyNameConverter.to_ar_name(attribute.to_sym, context: project)
+        else
+          ar_name = "cf_#{cf.id}"
+          # currently we do not support embedding rich text fields: long text custom fields
+          return msg_macro_error_rich_text if cf.formattable?
+
+          # Is the user allowed to see this custom field/"project attribute"?
+          if project.available_custom_fields.find { |pcf| pcf.id == cf.id }.nil?
```
ulferts: If you are loading the project's available_custom_fields anyway, this could also be used to fetch it, right?

This would then change the method to:

```
return msg_macro_error_rich_text if DISABLED_PROJECT_RICH_TEXT_FIELDS.include?(ar_name.to_sym)

cf = project.available_custom_fields.find { |pcf| pcf.name == attribute }

return msg_macro_error_rich_text if cf&.formattable?

ar_name = if cf.nil?
            ::API::Utilities::PropertyNameConverter.to_ar_name(attribute.to_sym, context: project)
          else 
            "cf_#{cf.id}"
          end  

format_attribute_value(ar_name, Project, project)
```

There is a slight difference here compared to the other implementation which is that custom fields that exist but not in the project's context would more quickly fall back into the attempt of dealing with it via the static attributes. But I guess the problem of having a custom field which's name clashes with the built in attributes is an unsolved on in any case

app/models/work_package/exports/macros/attributes.rb
```diff
@@ -0,0 +1,185 @@
...
+          # TODO: Are there access restrictions on work_package custom fields?
+        end
+
+        # currently we do not support embedding rich text field: workPackageValue:1234:description
+        if DISABLED_WORK_PACKAGE_RICH_TEXT_FIELDS.include?(ar_name.to_sym)
+          return msg_macro_error_rich_text
+
+        end
+
+        format_attribute_value(ar_name, WorkPackage, work_package)
```
ulferts: This could be rewritten just like the method for Project. It would then also use the `work_package.available_custom_fields` which would answer the TODO.

ulferts: If it turns out that both methods can be changed in the way proposed here, this probably also means that the code can be reused so maybe we end up only having a single method which just takes an additional parameter for either `WorkPackage` or `Project`.

app/models/work_package/pdf_export/markdown_field.rb
```diff
@@ -37,7 +38,52 @@ def write_markdown_field!(work_package, markdown, label)
...
+  end
+
+  private
+
+  def apply_markdown_field_macros(markdown, work_package)
+    apply_macros(markdown, work_package, WorkPackage::Exports::Macros::Attributes)
+  end
+
+  def apply_macros(markdown, work_package, formatter)
+    return markdown unless formatter.applicable?(markdown)
```
ulferts, line 52: The formatter in effect seems to be static since the method is only called from a single place if I am not mistaken. So it could then just become a local variable (or be placed into a method)

app/models/work_package/pdf_export/markdown_field.rb
```diff
@@ -37,7 +38,52 @@ def write_markdown_field!(work_package, markdown, label)
...
+        node.string_content = apply_macro_html(node.string_content, work_package, formatter)
+      elsif node.type == :text
+        node.string_content = apply_macro_text(node.string_content, work_package, formatter)
+      end
+    end
+    document.to_markdown
+  end
+
+  def apply_macro_text(text, work_package, formatter)
+    return text unless formatter.applicable?(text)
```
ulferts, line 66: This has already been checked in the calling method and this method here is private. Admittedly it is in a module so an including class could change this but currently that is not the case and somewhat unlikely I think.

app/models/work_package/pdf_export/markdown_field.rb
```diff
@@ -37,7 +38,52 @@ def write_markdown_field!(work_package, markdown, label)
...
+
+    text.gsub!(formatter.regexp) do |matched_string|
+      matchdata = Regexp.last_match
+      formatter.process_match(matchdata, matched_string, { user: User.current, work_package: })
+    end
+  end
+
+  def apply_macro_html(html, work_package, formatter)
+    return html unless formatter.applicable?(html)
```
ulferts, line 75: Same as above, this has been checked before in the callstack.

app/models/work_package/exports/macros/attributes.rb
```diff
@@ -0,0 +1,185 @@
...
+      end
+
+      def self.resolve_value_project(project, attribute)
+        cf = CustomField.find_by(name: attribute, type: "ProjectCustomField")
+        if cf.nil?
+          ar_name = ::API::Utilities::PropertyNameConverter.to_ar_name(attribute.to_sym, context: project)
+        else
+          ar_name = "cf_#{cf.id}"
+          # currently we do not support embedding rich text fields: long text custom fields
+          return msg_macro_error_rich_text if cf.formattable?
```
ulferts: If the field is formattable but disabled in the project, the error returned will still be

`[Rich text embedding currently not supported in export]`

when one would expect to get a message that the 'resource not found' error.

app/models/work_package/exports/macros/attributes.rb
```diff
@@ -0,0 +1,185 @@
...
+#++
+module WorkPackage::Exports
+  module Macros
+    # OpenProject attribute macros syntax
+    # Examples:
+    #   workPackageLabel:1234:subject # Outputs work package label attribute "Subject" + help text
+    #   workPackageValue:1234:subject # Outputs the actual subject of #1234
+    #
+    #   projectLabel:statusExplanation # Outputs current project label attribute "Status description" + help text
+    #   projectValue:statusExplanation # Outputs current project value for "Status description"
```
ulferts: The macro for the project also supports a [reference to a different project](https://www.openproject.org/docs/user-guide/wysiwyg/#available-attributes-for-projects). Currently, the PDF export does not support it.

Gemfile
ulferts: A comment explaining where it is used would be nice.

ulferts (changes requested):
Hey @as-op please find my remarks next to the code. 

Admittedly I skipped the specs since I wanted to provide feedback more quickly.

Gemfile
as-op: Commonmarker gem removed support fo the markdown abstract syntax tree some time ago.
I needed that for the attribute parsing in the pdf export, so  use the markly gem, which md_to_pdf already depends on it, too. So here just changed that it is a direct dependency now.

We could remove the commonmarker gem in the future and only use markly, but that was out of scope every time this topic came up.

I'll add a comment.

app/models/work_package/pdf_export/markdown_field.rb
as-op: It's also called in apply_macro_html_node, so this check is useful to skip costly html parsing.

In apply_macro_html() it is useless, indeed.

PR #15704, opened 2024-05-30
Generate packages with GitHub Actions

crohr:
Hosted builders provided by [Packager.io](https://packager.io) will soon be deprecated. The expectation is that users handle the build in their CI (e.g. GitHub Actions) using the official [`pkgr` docker images](https://github.com/orgs/pkgr/packages?repo_name=pkgr), and then push the resulting packages into the Packager.io backend.

Switching to GitHub Actions allows:

- better control over the packaging process, notably the assignment of the package version
- unlimited concurrency
- better observability / status checks right into GitHub

Packages pushed by this new workflow end up on https://go.packager.io/u/opf/openproject/channels/feature/package-via-github-actions. This backend supports more recent distributions.

The legacy Packager.io backend will be available for a few weeks / months so that users can transition easily (i.e. packages will be built and pushed in two places for a while) on their next OpenProject update.


machisuji (approved):
👍

PR #15710, opened 2024-05-31
[#55381] The available_projects api endpoint displays custom fields incorrectly

dombesz:
https://community.openproject.org/work_packages/55381

Previously if the first project didn't had any custom fields enabled the ProjectEagerLoadingWrapper returned an empty available_custom_fields section for all the projects. The fix consists of returning the available custom fields for each project apart.


ulferts (approved):
Change is fine, @dombesz.

Can you please reset the commit on release/14.1 so that it can become part of the next patch release?

Once done, merge at will.

PR #15715, opened 2024-05-31
[55318] Convert pt to pt-BR in packaged installations

cbliard:
https://community.openproject.org/wp/55318

If the value "pt" (Portuguese) was selected in the installation wizard, the value is saved in `/etc/openproject/installer.dat`. This value is then used in the postinstall script to set `OPENPROJECT_DEFAULT_LANGUAGE`. But since work package #53374, pt has been split into pt-BR and pt-PT and the app will refuse to start with value pt.

Changing the saved value to pt-BR when it's pt fixes the issue.


machisuji (approved):
👍

PR #15731, opened 2024-06-03
[#52670] rework main menu resizer

Kharonus:
[#52670](https://community.openproject.org/work_packages/52670)

### WAT?

- fixed typings in resizer component
- fixed linting issues
- changed CSS for resizer to be:
  - drag handle now has 1rem width
  - drag handle is now fixed position over right side of the menu itself
  - changed hover behavior of toggle to replacing the icon
  - changed hover color to accent

### DIFF
--- a/frontend/src/app/shared/components/grids/grid/grid.component.html
+++ b/frontend/src/app/shared/components/grids/grid/grid.component.html
@@ -35,17 +35,17 @@
                    [ndcDynamicOutputs]="widgetComponentOutput(area)">
       </ndc-dynamic>
     </div>
-    <resizer *ngIf="resize.isResizable"
+    <op-resizer *ngIf="resize.isResizable"
              [customHandler]="true"
-             (end)="resize.end(area)"
-             (start)="resize.start(area)"
+             (resizeFinished)="resize.end(area)"
+             (resizeStarted)="resize.start(area)"
              (move)="resize.moving()">
       <div class="grid--resizer">
         <i class="icon-resizer-bottom-right hidden-for-mobile"
            aria-hidden="true">
         </i>
       </div>
-    </resizer>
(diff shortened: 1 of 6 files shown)


Kharonus:
Not really a primer bug, but something style related nontheless. Please have a look, @HDinger, if you are satisfied with the changes I suggest. (e.g. not displaying the chevron left next to the vertical lines, but instead replacing them + using accent color). 

I think, that is the more accepted behavior, yet, if we really want to keep the vertical lines and add the chevron left to it, I can change it of course.

HDinger (changes requested):
Thanks @Kharonus ! This is so much better than before 🙇 
As discussed, please use the Octicons instead of the spot-icons and then just toggle the visiblity on hover. The rest is looking fine 👍

HDinger (commented):
Changes look good to me 👍 
I am just wondering whether we want to keep the color as it was before (a light grey) or stay with white. Maybe we can ask product later. Other than that it is good to merge 👍

Kharonus:
Adapted the changes from meeting:
- removed hover accent color for the toggle
- added border in accent color for whole resize bar

HDinger (commented):
I have to be honest, that last change look super weird to me. I'd like to check with product whether that is really what they had in mind. 
Further, from what I understood, the icon should be in light grey and change their color to a darker variant on hover. Maybe we have to bring these quesions up in tomorrows meeting.

Kharonus:
Talked to @marcalcobe with the results of the last (squashed) commit:

- reduced width of hover border
- toggler icon changes color on hover to main menu border color
- hover border uses main menu border color

HDinger (approved):
Awesome 👍 Thanks for the patience 🙇

PR #15735, opened 2024-06-03
[#55275] New section option not i18n-ed

dombesz:
https://community.openproject.org/work_packages/55275

### DIFF
--- a/modules/meeting/app/components/meeting_agenda_items/new_button_component.html.erb
+++ b/modules/meeting/app/components/meeting_agenda_items/new_button_component.html.erb
@@ -23,7 +23,7 @@
       )
       unless @meeting_section
         component.with_item(
-          label: "Section",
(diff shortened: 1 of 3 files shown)


modules/meeting/app/components/meeting_agenda_items/new_button_component.html.erb
```diff
@@ -23,7 +23,7 @@
       )
       unless @meeting_section
         component.with_item(
-          label: "Section",
+          label: t("activerecord.models.meeting_section", count: 1),
```
klaustopher: Why is the `count: 1` added here? The translation key you added does not have the `one`/`other` subkeys, so the count does not do anything here. 

```suggestion
          label: t("activerecord.models.meeting_section"),
```

klaustopher (approved):

PR #15741, opened 2024-06-04
Typo week, but it should be month as setting days per month.

Eric-Guo:


dombesz (approved):
Thank you @Eric-Guo !

PR #15744, opened 2024-06-04
[#55431] Display an inline error when duration input is not a duration  

aaron-contreras:
Strings such as "Hello world" aren't durations and hence, should not
be parsed or continue through to the services. An early validation at
the controller level guards against these.

See: https://community.openproject.org/work_packages/55431

### DIFF
--- a/app/controllers/work_packages/progress_controller.rb
+++ b/app/controllers/work_packages/progress_controller.rb
@@ -131,13 +131,14 @@ class WorkPackages::ProgressController < ApplicationController
   def work_package_params
     params.require(:work_package)
           .permit(allowed_params).tap do |wp_params|
-      if wp_params["estimated_hours"].present?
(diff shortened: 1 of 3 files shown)


cbliard (commented):
Nice.

I wonder if that's possible to make it a bit more accurate

For instance, in this example, the values were initially blank. When setting "qqq" as value for Work, Remaining work got 0h. I think it would be more correct to keep it empty.

image

Seemingly, with initial values being work=6h, remaining work=6h, and so % complete=0%, when setting remaining work to qqq, the % complete changes to 100%. I think it would be more correct if it was unset.

image

I'll see if I can make that happen.

cbliard:
But still it behaved already like this in the previous version, so I see no reason to hold merging. That's something we can improve in another PR.

cbliard (approved):
Good job 👍

PR #15753, opened 2024-06-05
allow changing of direct login provider at runtime

machisuji:
because why would you not


machisuji:
Closed in favour of #15754

PR #15759, opened 2024-06-05
allow changing of direct login provider at runtime

machisuji:
because why would you not?

Also fixes the `with_config` spec helper to also stub Settings to pin their values just like they are outside of tests.


ba1ash (approved):
LGTM

PR #15760, opened 2024-06-05
[55467] Warn about potential data loss when changing progress modes

cbliard:
https://community.openproject.org/wp/55467

Display a warning banner when the mode is being changed.

image

### DIFF
--- a/frontend/src/stimulus/controllers/dynamic/admin/work-packages-settings.controller.ts
+++ b/frontend/src/stimulus/controllers/dynamic/admin/work-packages-settings.controller.ts
@@ -1,0 +1,69 @@
+/*
+ * -- copyright
+ * OpenProject is an open source project management software.
+ * Copyright (C) 2023 the OpenProject GmbH
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU General Public License version 3.
+ *
+ * OpenProject is a fork of ChiliProject, which is a fork of Redmine. The copyright follows:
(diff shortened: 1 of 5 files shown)


frontend/src/stimulus/controllers/dynamic/admin/work-packages-settings.controller.ts
```diff
@@ -0,0 +1,69 @@
...
+      this.warningToastTarget.hidden = true;
+    }
+  }
+
+  getWarningMessageHtml():string {
+    const newMode = this.progressCalculationModeSelectTarget.value;
+    return I18n.t(
+      `js.admin.work_packages_settings.warning_progress_calculation_mode_change_from_${this.initialMode}_to_${newMode}_html`,
+      { defaultValue: '' },
+    );
```
dombesz, line 67: I see what you did here, it's a tricky one! 🙂
I hope it's not going to cause some unexpected performance issues, by looking for non-existing translation strings.

dombesz (approved):
Nice job @cbliard !🍹

frontend/src/stimulus/controllers/dynamic/admin/work-packages-settings.controller.ts
cbliard: Yep, the initial version had a lot more `if`s. This version is more straightforward.
Hopefully even if there are some performance issues, they should be limited as the user is only changing a select value from one value to another.

PR #15761, opened 2024-06-05
[#55429] Only display "working/non-working days changed warning" when they actually change

aaron-contreras:
See: https://community.openproject.org/work_packages/55429


cbliard (approved):
Nice job @aaron-contreras ! 🪴 

I noticed a bug but it was there before your changes: when adding a non working day and removing it, then clicking apply, the popup text is wrong. I'll file a bug for it.

PR #15765, opened 2024-06-06
Add week support of Gantt PDF export.

Eric-Guo:
Example of generated PDF file (first page)

[PLM__2024-06-06_14-43.pdf](https://github.com/user-attachments/files/15604974/PLM__2024-06-06_14-43.pdf)


Eric-Guo:
@as-op previous implement has cross year bug, now should be fine.

as-op (approved):
Work as expected for all my test projects, thank you, @Eric-Guo!

PR #15768, opened 2024-06-06
[#53987] 🇺🇦 Use different Ukraine supporting logos.

ba1ash:
https://community.openproject.org/work_packages/53987


lindenthal:
Thanks @ba1ash

PR #15771, opened 2024-06-06
Do not trigger development seeders if installation is obviously old

cbliard:
`DevelopmentData::ProjectsSeeder` creates 4 projects trying to be useful for development, but it's possible that we do not need them and delete them. If that's the case, we do not want
`DevelopmentData::ProjectsSeeder` to try to create them again.

This commit adds an additional check: if there are no projects older than 1 hour, that means that the instance is recent and the dev projects should be seeded. Otherwise, they should not.

It should prevent us from having errors in the future when we delete the projects and the seeder tries to create them again.

### DIFF
--- a/app/seeders/development_data/projects_seeder.rb
+++ b/app/seeders/development_data/projects_seeder.rb
@@ -47,7 +47,13 @@ module DevelopmentData
     end
 
     def applicable?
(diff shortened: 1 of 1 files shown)


Kharonus (approved):
LGTM

PR #15772, opened 2024-06-06
[55431] Fix error message when duration input is invalid

cbliard:
[Bug: Display an error when duration input is invalid (#55431)](https://community.openproject.org/projects/openproject/work_packages/55431)

Passing invalid input like "abcdef" will now display the error as "is not a valid duration." instead of "is not a number.".


aaron-contreras (approved):
Very nice!

PR #15782, opened 2024-06-07
[#55507] Make it possible to activate custom fields on multiple projects at the same time

akabiru:
https://community.openproject.org/work_packages/55507

https://github.com/opf/openproject/assets/17295175/306fc256-643c-4161-bc1c-6bf5aeb53ff4

### DIFF
--- a/app/controllers/admin/settings/project_custom_fields_controller.rb
+++ b/app/controllers/admin/settings/project_custom_fields_controller.rb
@@ -43,7 +43,7 @@ module Admin::Settings
     before_action :prepare_custom_option_position, only: %i(update create)
     before_action :find_custom_option, only: :delete_option
     before_action :project_custom_field_mappings_query, only: %i[project_mappings unlink]
-    before_action :find_link_project_custom_field_mapping, only: :link
+    before_action :find_custom_field_projects_to_link, only: :link
     before_action :find_unlink_project_custom_field_mapping, only: :unlink
     # rubocop:enable Rails/LexicallyScopedActionFilter
 
@@ -75,7 +75,7 @@ module Admin::Settings
 
     def link
       create_service = ProjectCustomFieldProjectMappings::BulkCreateService
-                         .new(user: current_user, project: @project, project_custom_field: @custom_field,
+                         .new(user: current_user, projects: @projects, project_custom_field: @custom_field,
                               include_sub_projects: include_sub_projects?)
                          .call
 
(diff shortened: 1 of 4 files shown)


app/controllers/admin/settings/project_custom_fields_controller.rb
```diff
@@ -75,15 +75,15 @@ def project_mappings
 
     def link
       create_service = ProjectCustomFieldProjectMappings::BulkCreateService
-                         .new(user: current_user, project: @project, project_custom_field: @custom_field,
+                         .new(user: current_user, projects: @projects, project_custom_field: @custom_field,
```
ba1ash, line 78: What do you think about passing project_ids to the service and moving all logic about finding and initializing relevant projects there?
In current implementation this logic is distributed between controller and service. I suggest to avoid that.

ba1ash: And do we need to react with `Not found` error in case there are none projects with given ids? Then we could simplify logic a bit more.

akabiru: > What do you think about passing project_ids to the service and moving all logic about finding and initializing relevant projects there? In current implementation this logic is distributed between controller and service. I suggest to avoid that.

I reckon one disadvantage to that approach is that we can't fail early- and return a 404 if some invalid data is passed. The controller is nicely updating the flash component via turbo in this case- which makes a case for keeping it in the controller. WDYT?

akabiru: > And do we need to react with `Not found` error in case there are none projects with given ids? Then we could simplify logic a bit more.

Not necessarily, just handling a corner case. Given it's just a one liner guard clause, I reckoned that was acceptable?

app/services/project_custom_field_project_mappings/bulk_create_service.rb
```diff
@@ -72,23 +74,25 @@ def validate_contract(service_call, project_ids)
...
     def perform_bulk_create(service_call)
-      ProjectCustomFieldProjectMapping.import(service_call.result, validate: false)
+      ProjectCustomFieldProjectMapping.insert_all(
+        service_call.result.map { |model| model.attributes.slice("project_id", "custom_field_id") }
+      )
 
       service_call
-    rescue StandardError => e
-      service_call.success = false
-      service_call.errors = e.message
```
akabiru, line 80: btw, I got rid of this @ba1ash - should put a smile on your face 😄

ba1ash: Yep :)

app/controllers/admin/settings/project_custom_fields_controller.rb
ba1ash: Yes, I see. My point if it is really needed to handle this corner case. If it is very unlikely to happen then maybe just respond with 204 no content and do nothing. Let's have a chat.

ba1ash (approved):
LGTM!

PR #15783, opened 2024-06-07
Task/55493 14 documentation transform remove action from share modals to an iconbutton

MayaBerd:


psatyal (approved):
Perfect, thanks, @MayaBerd!

PR #15784, opened 2024-06-07
[#50954] Preview chronic durations only on blur

aaron-contreras:
See: https://community.openproject.org/work_packages/50954
Relevant comment: https://community.openproject.org/work_packages/50954/activity#activity-40


cbliard (changes requested):
I tried the pull preview. It nicely formats the input on blur, but I noted the following issues:

---

**format jump**

When editing a field, the other one is show with durations, but as soon as it is focused, it switches back to plain hours number without any units. It was better before when it kept the multiple units format.

To reproduce:
- set work to "10" (remaining work is set to "1d 2h")
- focus remaining work
- => it sets work to "1d 2h" (ok), but sets back remaining work to "10". It should keep the calculated "1d 2h" for edition.

---

**Stuck focus**

The error field is being focused, and when selecting another field, it is focused again, making it almost impossible to edit another field.

To reproduce:
- work based
- set work to 10
- set remaining work to 7
- set work empty (error "required when remaining work is set")
- => try unsetting remaining work (focus issue)

another path:
- set work to 10
- set remaining work to 7
- unset remaining work (empty)
- => try unsetting work (focus issue)

We should remove this "focus errored field" thing as it was useful before, but now that it updates dynamically on every input change, it's more annoying than useful.

---

**calculated values as input**

The calculated values in the progress popover should be ignored by the backend. Only the values supplied by the user should be used. Currently, the calculated values may be used when typing fast, leading to unexpected results:

to reproduce:
- work-based
- set work to "4" and wait until remaining work is calculated to be "4h"
- still in the work field, type "2" and immediately press enter.
- => work is set to 42h, while remaining work is set to 4h, but the user never set remaining work to 4h. It should be calculated and be 42h like work.

(I noticed this bug by accident while trying to reproduce another one)

We can have a look at them together if you want.

cbliard:
I addressed the "calculated values as input" concern in "[Only use user-supplied values when updating progress](https://github.com/opf/openproject/pull/15784/commits/7a588db68a970cda2b58569864786ec8f2be4b3a)" commit.

app/components/work_packages/progress/base_modal_component.rb
```diff
@@ -40,21 +40,32 @@ class BaseModalComponent < ApplicationComponent
...
       include ApplicationHelper
       include Turbo::FramesHelper
       include OpPrimer::ComponentHelpers
       include OpenProject::StaticRouting::UrlHelpers
 
-      attr_reader :work_package, :mode, :focused_field, :touched_field_map
+      attr_reader :work_package,
+                  :mode,
+                  :focused_field,
+                  :format_durations,
```
cbliard: It looks like `format_durations` is now always true, and so is not needed anymore and can be removed. Am I right?

cbliard (approved):
All good 🥂

app/components/work_packages/progress/base_modal_component.rb
cbliard: I removed it

PR #15795, opened 2024-06-10
translate deploy comment, allow to be made by manually set github user

machisuji:
A bug in the deploy target feature was noticed. The comment made is not translated.
This PR fixes it and while we're at it, it also allows the github user to be configured (for now) by hand so that the comments can come from is rather than the system user.


dombesz (approved):
Nice job @machisuji !👍

PR #15801, opened 2024-06-10
Bug/53772 the label for spent time is still visible after deactiving the module time and costs active modules

ulferts:
### Goal

For admin users, permissions were granted regardless of whether the module a permission belongs to is active or not. Now, this is always checked. 

This comes with the caveat that for some scenarios, e.g. displaying the menu items "Boards" and "Meetings", a project needs to exist to have the menu items show up. Since the permission check is against `allowed_in_any_project?` this seems to be correct however. In most real live scenarios, there should always be a project anyway. In tests, this might not always be the case which is why some needed to be amended in the scope of this PR.

### WP

https://community.openproject.org/wp/53772

### DIFF
--- a/spec/controllers/news_controller_spec.rb
+++ b/spec/controllers/news_controller_spec.rb
@@ -36,7 +36,7 @@ RSpec.describe NewsController do
   let(:news) { create(:news) }
 
   shared_let(:project) { create(:project) }
-  current_user { create(:admin) }
+  shared_current_user { create(:admin) }
 
   describe "#index" do
     it "renders index" do
--- a/spec/controllers/news_controller_spec.rb
+++ b/spec/controllers/news_controller_spec.rb
@@ -33,15 +33,10 @@ RSpec.describe NewsController do
 
   include BecomeMember
 
-  let(:user) do
-    create(:admin)
-  end
-  let(:project) { create(:project) }
   let(:news) { create(:news) }
 
-  before do
-    allow(User).to receive(:current).and_return user
-  end
+  shared_let(:project) { create(:project) }
+  current_user { create(:admin) }
 
   describe "#index" do
     it "renders index" do
@@ -101,7 +96,7 @@ RSpec.describe NewsController do
   describe "#create" do
     context "with news_added notifications" do
       it "persists a news item" do
-        become_member(project, user)
+        become_member(project, current_user)
 
         post :create,
              params: {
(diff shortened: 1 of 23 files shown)


spec/controllers/news_controller_spec.rb
```diff
@@ -33,15 +33,10 @@
...
-    create(:admin)
-  end
-  let(:project) { create(:project) }
   let(:news) { create(:news) }
 
-  before do
-    allow(User).to receive(:current).and_return user
-  end
+  shared_let(:project) { create(:project) }
+  current_user { create(:admin) }
```
cbliard: There is a `shared_current_user` available to prevent recreating the same user for each test.

spec/features/projects/global_menu_item_spec.rb
```diff
@@ -32,10 +32,9 @@
 require "spec_helper"
 
 RSpec.describe "Projects global menu item", :js, :with_cuprite do
-  let(:user) { create(:user) }
+  current_user { create(:user) }
```
cbliard: In this case `shared_current_user` won't work because it is changed later (well, maybe it would work. I did not try). What I generally do in this case is this:
```
shared_let(:user) { create(:user) }
shared_let(:admin) { create(:admin) }
current_user { user }

context "with an admin user" do
  current_user { admin }
end
```

cbliard (changes requested):
@ulferts It works well. The Time spent field is no longer displayed in the work package details view when the "Time and Costs" module is disabled. I like the way the new tests for `permissible_on?` are written.

Meanwhile when the "Time and Costs" module is disabled, it is still possible to display the column "Spent time" in the work package table view. Is it intended? I feel like this column should be removed.

spec/controllers/news_controller_spec.rb
ulferts: Thanks for the pointer, didn't know that one.

ulferts:
> Meanwhile when the "Time and Costs" module is disabled, it is still possible to display the column "Spent time" in the work package table view. Is it intended? I feel like this column should be removed.

I share the sentiment that the column should be removed. Unfortunately, the column/select implementation used for work packages does not yet support checking for availability. So I don't see an easy way of hiding the column in case the module is inactive. We can get back to this once the work package query uses the same select implementation as e.g. project does. I'll add it as a TODO to a ticket we have for this.

cbliard (approved):

PR #15802, opened 2024-06-10
Only cache the client credential token if requests where successful

mereghost:
## What?

We cache the client credential token for most of its duration, which is fine unless the storage is miss-configured and we get a failure. Then the possibly broken token will potentially continue failing until the cache expires.

This PR intends to avoid that, by only caching tokens after a successful response.

## How?

The caching of the token was moved to after the yielding of the HTTP client. There's also a check for a `:forbidden` result (from a 403) that evicts an existing token from the cache.

Not all Commands/Queries return 403s but introducing this was left out of this PR

### DIFF
--- a/modules/storages/app/common/storages/peripherals/storage_interaction/authentication_strategies/oauth_client_credentials.rb
+++ b/modules/storages/app/common/storages/peripherals/storage_interaction/authentication_strategies/oauth_client_credentials.rb
@@ -37,56 +37,57 @@ module Storages
             Strategy.new(:oauth_client_credentials)
           end
 
-          # rubocop:disable Metrics/AbcSize
           def call(storage:, http_options: {})
             config = storage.oauth_configuration.to_httpx_oauth_config
-
             return build_failure(storage) unless config.valid?
 
-            access_token = Rails.cache.read("storage.#{storage.id}.httpx_access_token")
+            token_cache_key = cache_key(storage)
+            access_token = Rails.cache.read(token_cache_key)
 
-            http_result = if access_token.present?
-                            http_with_current_token(access_token:, http_options:)
-                          else
-                            http_with_new_token(issuer: config.issuer,
-                                                client_id: config.client_id,
-                                                client_secret: config.client_secret,
-                                                scope: config.scope,
-                                                http_options:)
-                          end
-
-            return http_result if http_result.failure?
-
-            http = http_result.result
+            # In ruby 3.4 this can become `return it`
+            http = build_http_session(access_token, config, http_options).on_failure { return _1 }.result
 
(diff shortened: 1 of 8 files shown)


modules/storages/spec/common/storages/peripherals/storage_interaction/authentication_strategies/oauth_client_credentials_spec.rb
```diff
@@ -0,0 +1,122 @@
...
+    before do
+      stub_request(:post, "https://login.microsoftonline.com/4d44bf36-9b56-45c0-8807-bbf386dd047f/oauth2/v2.0/token")
+        .and_return(status: 200, body: token_json, headers: { content_type: "application/json" })
+
+      stub_request(:get, "https://graph.microsoft.com/v1.0/drives/b!dmVLG22QlE2PSW0AqVB7UOhZ8n7tjkVGkgqLNnuw2OBb-brzKzZAR4DYT1k9KPXs/root")
+        .and_return(status: 403)
+    end
+
+    it "does not cache the result" do
+      expect(Rails.cache.fetch("storage.#{storage.id}.httpx_access_token")).to be_nil
```
ba1ash: I suggest extracting `"storage.#{storage.id}.httpx_access_token"` into a let variable.

modules/storages/app/common/storages/peripherals/storage_interaction/authentication_strategies/oauth_client_credentials.rb
```diff
@@ -59,12 +59,16 @@ def call(storage:, http_options: {})
...
             http = http_result.result
 
-            if access_token.nil?
+            operation_result = yield http
+
+            if access_token.nil? && operation_result.success?
               token = http.instance_variable_get(:@options).oauth_session.access_token
               Rails.cache.write("storage.#{storage.id}.httpx_access_token", token, expires_in: 50.minutes)
+            elsif operation_result.failure? && operation_result.result == :forbidden
+              Rails.cache.delete("storage.#{storage.id}.httpx_access_token")
```
ba1ash: Let's extract cache key construction to a method or save it to a local variable to have it defined just in one place.
It will help to avoid stupid error when accidentally cache key got changed only in one place, but not in the other.

ba1ash (approved):
It looks good to me overall, I have two minor comments.

modules/storages/app/common/storages/peripherals/storage_interaction/authentication_strategies/oauth_client_credentials.rb
mereghost: Good point.

PR #15818, opened 2024-06-11
[#55529] encode collection key correctly

Kharonus:
[#55529](https://community.openproject.org/work_packages/55529)


ba1ash (approved):

wielinde:
Hey @Kharonus @ba1ash, I had a look into this PR which supposedly shall fix a certain issue with ampersands in URLs that are passed as parameters. Did I miss anything or does this fix do not have a specific test for the original issue? If so, why did you decide against providing one? Is this due to the fact that is front-end logic and thus harder to test?

PR #15820, opened 2024-06-11
[#55523] make add projects dialog mobile friendly

judithroth:
https://community.openproject.org/work_packages/55523

I also tried centering the checkbox and its label on desktop with flexbox / flex grid, but then the actual checkbox and its label did no longer align (even though they both were more centered than before). If anyone knows a better solution, please let me know :)

### DIFF
--- a/app/forms/projects/custom_fields/custom_field_mapping_form.rb
+++ b/app/forms/projects/custom_fields/custom_field_mapping_form.rb
@@ -47,8 +47,8 @@ module Projects::CustomFields
           name: :include_sub_projects,
           label: I18n.t("projects.settings.project_custom_fields.new_project_mapping_form.include_sub_projects"),
           checked: false,
-          class: "my-sm-2",
-          label_arguments: { class: "no-wrap my-sm-2" }
+          my: [0, 2],
+          label_arguments: { class: "no-wrap", my: [0, 2] }
         )
       end
     end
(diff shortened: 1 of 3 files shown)


judithroth:
On mobile:
Screenshot from 2024-06-12 09-34-42

On desktop:
Screenshot from 2024-06-12 09-37-17

judithroth:
@HDinger maybe you want to take a look as well :)

app/forms/projects/custom_fields/custom_field_mapping_form.rb
```diff
@@ -43,11 +43,12 @@ class CustomFieldMappingForm < ApplicationForm
           }
         )
 
-        f_group.check_box(
+        group.check_box(
           name: :include_sub_projects,
           label: I18n.t("projects.settings.project_custom_fields.new_project_mapping_form.include_sub_projects"),
           checked: false,
-          label_arguments: { class: "no-wrap" }
+          class: "my-sm-2",
```
HDinger: Instead of passing the utitlity class directly, please use the [system arguments ](https://primer.style/view-components/lookbook/pages/system_arguments/)for that. Thus we ensure that will also work if they for example rename the class. The margin is a responsive atttribute so you should be able to do something like this:

```
group.check_box(
          name: :include_sub_projects,
          label: I18n.t("projects.settings.project_custom_fields.new_project_mapping_form.include_sub_projects"),
          checked: false,
          label_arguments: { class: "no-wrap" }
          my: [2 ,  0]
          label_arguments: { class: "no-wrap", my: [2, 0] }

HDinger (changes requested):
Although the solution with the margin works, I am still wondering why the alignment via flex did not work. Maybe we can shortly chat about that @judithroth ?

HDinger (approved):
There is still a tiny misalignment between the checkbox and the autocompleter. As discussed, this is fine to merge and we will tackle that issue separatley. 👍

akabiru (approved):
:shipit:

PR #15821, opened 2024-06-11
[#55621] Presenter field on agenda item becomes tiny if the current presenter is removed and a new one is searched for

mrmir:
https://community.openproject.org/work_packages/55621


mrmir:
@HDinger 200px is perfect and looks better than 150px and 250px to me

oliverguenther (approved):

PR #15827, opened 2024-06-12
Fix global menu item too height in Safari 17

Eric-Guo:
Before:

iShot_2024-06-11_22 56 17

After:

iShot_2024-06-12_10 48 17

### DIFF
--- a/frontend/src/global_styles/common/menu/menu.sass
+++ b/frontend/src/global_styles/common/menu/menu.sass
@@ -13,8 +13,7 @@
       align-items: center
 
     &-action
-      display: flex
-      align-items: center
+      display: block
       height: 32px
       line-height: 32px
       text-decoration: none


HDinger (changes requested):
Hi @Eric-Guo 

unfortunately, I cannot reproduce the original bug. This is how it looks for me with Safari 17.1

Bildschirmfoto 2024-06-12 um 09 46 45

I see that they are slightly off, but not as bad as in your case. However, applying your fix only makes that worse. 
Bildschirmfoto 2024-06-12 um 09 48 22

Looking at your screenshot, I see that you also have an own menu entry and that the icon for WorkPackages is not the same. Is it possible that you have added own styles or classes that interfer here?

Eric-Guo:
I not add own styles or class, what I add just below in Engine:

```
      menu :top_menu,
           :th_keyin,
           '/th_keyin',
           icon: 'id-badge',
           caption: :label_th_keyin
```

My Safari version is 17.5 (19618.2.12.11.6)

Eric-Guo:
But from your screen shot, seems the global menu icon is not changed, so probably, the commit 4f2b8d39c26ef2ab45606db2c0d6e64d24f97313 is not yet deploy to your QA server yet? @HDinger

HDinger (changes requested):
> But from your screen shot, seems the global menu icon is not changed, so probably, the commit [4f2b8d3](https://github.com/opf/openproject/commit/4f2b8d39c26ef2ab45606db2c0d6e64d24f97313) is not yet deploy to your QA server yet? @HDinger

Ah. yes, you are right. My bad 🙈 

However, using `display: list-item` is reserved for actual lists, although due ti accessibility. The surrounding `<li>` already has that property. For me, locally it works wih `display: block`. Could you please check whether that works for you as well?

Eric-Guo:
Yes, the block works [very well too](https://plm-ppp.thape.com.cn/), just change it. @HDinger

HDinger (approved):
Lgtm 👍 Thanks @Eric-Guo ! Good to merge once the tests are green

PR #15841, opened 2024-06-13
[WIP] Reorganizes the StorageInteraction code plus other improvements

mereghost:
# THIS IS AN EXPERIMENT

## What?

We have a bunch of namespaces under `Storages` but some don't bring much clarity on what is their intended purpose or we gained new insight on the necessities for different service providers. 

## Why?

All storage supporting code lives under `Peripherals` (under the `common/storages/` folder) and has become kind of that practical drawer where you just throw anything you wanna "hide" (under the guise of organizing) and it is slowing becoming a mess.

## How?

First I'm introducing the Adapters namespace (under `common/storage/adapters`) where the code that is tailored goes (`adapters/nextcloud`, `adapters/one_drive`).

Each adapter structure is identical (in an ideal world) with the following namespaces / folders:

- **Commands**: all commands (requests that makes changes to the remote service) go here
- **Queries**: all queries (requests that only fetch remote information) go here
- **Contracts**: contracts to be used with the `Storage` models
- **Support**: any support code required for that service provider

The Adapters namespace will also have supporting code to it - Authentication Strategies, Specific Errors and also Objects defining the shape the results (success/failure) for all the required `Commands` and `Queries`.  
 
The `adapters` folder/namespace also houses all the pieces used by the adapters (authentication, value objects etc)


modules/storages/app/common/storages/adapters/nextcloud/commands/copy_template_folder.rb
```diff
@@ -27,11 +29,12 @@
...
-    module StorageInteraction
-      module Nextcloud
-        class CopyTemplateFolderCommand
-          using ServiceResultRefinements
+  module Adapters
+    module Nextcloud
+      module Commands
+        class CopyTemplateFolder
+          using Peripherals::ServiceResultRefinements
+          Util = Peripherals::StorageInteraction::Nextcloud::Util
```
mereghost, line 38: This would eventually be removed when `Util` gets moved (and maybe split) into `support`

modules/storages/app/common/storages/adapters/nextcloud/managed_folder_identifier.rb
```diff
@@ -0,0 +1,50 @@
...
+# along with this program; if not, write to the Free Software
+# Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+#
+# See COPYRIGHT and LICENSE files for more details.
+#++
+
+module Storages
+  module Adapters
+    module Nextcloud
+      class ManagedFolderIdentifier
```
mereghost, line 34: This isn't exactly support (as in supporting code) but represents a concept that is storage dependent.

Should it go under a special namespace?

modules/storages/app/common/storages/adapters/nextcloud/queries/files.rb
```diff
@@ -0,0 +1,245 @@
...
+#
+# See COPYRIGHT and LICENSE files for more details.
+#++
+
+module Storages
+  module Adapters
+    module Nextcloud
+      module Queries
+        class Files
+          Auth = Peripherals::StorageInteraction::Authentication
```
mereghost, line 36: `Authentication` will probably live under `Adapters::Support::Authentication`

modules/storages/app/common/storages/peripherals/nextcloud_registry.rb
```diff
@@ -37,7 +37,7 @@ module Peripherals
         register(:file_ids, StorageInteraction::Nextcloud::FileIdsQuery)
         register(:file_info, StorageInteraction::Nextcloud::FileInfoQuery)
         register(:files_info, StorageInteraction::Nextcloud::FilesInfoQuery)
-        register(:files, StorageInteraction::Nextcloud::FilesQuery)
+        register(:files, Adapters::Nextcloud::Queries::Files)
```
mereghost, line 41: `Registries` should be moved to their adapter folders (maybe we could add a way to auto-register? 🤔)

mereghost:
Ok, I have an idea of the amount of work involved. I'll sketch a code maintenance to do this.

PR #15842, opened 2024-06-13
[#55502] Sorting suggested to be possible when it isn't

judithroth:
Only marking the table component as not sortable was not sufficient since then it would raise in ProjectsHelper#projects_sort_header_tag, which is used when rendering the component.


akabiru (approved):
Tested and looks great! :shipit:

PR #15847, opened 2024-06-13
[#55651] Missing custom action strategy for custom fields with field format link (URL)

dombesz:
https://community.openproject.org/work_packages/55651


aaron-contreras (approved):
Very nice!

PR #15850, opened 2024-06-13
[#45896] Generate PDF document from a work package description 

as-op:
* [x] Refactor for better resue
* [x] Reuse logo handling from PDF report
* [x] Use md-to-pdf to generate the document from a work package description 
* [x] Trigger with entry in work package drop down menu
* [x] Hide behind feature flag `generate_pdf_from_work_package`
* [x] Dialog
   * [X] turbo/primer dialog
   * [x] use dialog parameters
* [x] Specs

https://community.openproject.org/work_packages/45896


lindenthal:
Hi @as-op,
Is the disclaimer in the description not correct? Are there topics to be specified?
Best
Niels

as-op:
Hi @lindenthal! No, that was a deprecated warning. I removed it.

app/components/work_packages/exports/generate/modal_dialog_component.rb
```diff
@@ -0,0 +1,127 @@
...
+            { value: "ms", label: "Melayu" },
+            { value: "nl", label: "Nederlands" },
+            { value: "no", label: "Norsk" },
+            { value: "pl", label: "Polski" },
+            { value: "pt", label: "Português" },
+            { value: "ru", label: "Русский" },
+            { value: "sk", label: "Slovenčina" },
+            { value: "sv", label: "Svenska" }
+          ].sort_by { |item| item[:label] }
+          [{ value: "", label: "Off", default: true }].concat(supported_languages)
```
EinLama, line 109: I think this method would read better if we added an empty line between return value and the bulky hash definition above it. Makes it clear that they are separate instructions.

app/components/work_packages/exports/generate/modal_dialog_component.rb
```diff
@@ -0,0 +1,127 @@
...
+              caption: I18n.t("pdf_generator.dialog.paper_size.caption"),
+              options: paper_size_options
+            }
+          ]
+        end
+
+        def hyphenation_options
+          # This is a list of languages that are supported by the hyphenation library
+          # https://rubygems.org/gems/text-hyphen
+          # The labels are the language names in the language itself (NOT to be put I18n)
```
EinLama, line 78: Great comment! 👍🏻

app/models/work_package/pdf_export/common/attachments.rb
```diff
@@ -51,4 +51,27 @@ def delete_all_resized_images
...
+
+    local_file = attachment_image_local_file(attachment)
+    return nil if local_file.nil?
+
+    resize_image(local_file.path)
+  end
+
+  def attachment_by_api_content_src(work_package, src)
+    # find attachment by api-path
+    work_package.attachments.detect { |a| api_url_helpers.attachment_content(a.id) == src }
```
EinLama: Since the comment above already uses the verb "find", consider using the alias `find` instead of `detect`.

app/models/work_package/pdf_export/common/logo.rb
```diff
@@ -0,0 +1,23 @@
+module WorkPackage::PDFExport::Common::Logo
+  def logo_image
+    image_obj, image_info = pdf.build_image_object(logo_image_filename)
+    [image_obj, image_info]
+  end
+
+  def logo_image_filename
+    image_file = custom_logo_image_filename
+    image_file = Rails.root.join("app/assets/images/logo_openproject.png") if image_file.nil?
+    image_file
```
EinLama: This could be shortened to: 

```suggestion
    custom_logo_image_filename || Rails.root.join("app/assets/images/logo_openproject.png")
```

app/models/work_package/pdf_export/common/logo.rb
```diff
@@ -0,0 +1,23 @@
+module WorkPackage::PDFExport::Common::Logo
```
EinLama, line 31: Not commenting in every file, but most of these new files lack the copyright notice, while others got them. I think when in doubt, we should include these in every file.

EinLama: Additionally, `frozen_string_literal: true` is a good idea, unless you explicitly want unfrozen strings in these files.

config/routes.rb
```diff
@@ -574,6 +574,9 @@
 
     get "hover_card" => "work_packages/hover_card#show", on: :member
 
+    get "generate_pdf_dialog" => "work_packages#generate_pdf_dialog", on: :member
+    get "generate_pdf" => "work_packages#generate_pdf", on: :member
```
EinLama: Why is this a GET call? Should creating a file not be POST? 🤔

app/models/work_package/pdf_export/generator/contracts.yml
```diff
@@ -0,0 +1,244 @@
+page:
+  font: 'NotoSans'
```
EinLama, line 2: This PR adds OpenSans as font, but NotoSans is used here instead (and SpaceMono further below). Is this an accident or a conscious decision?

spec/models/work_package/pdf_export/document_generator_spec.rb
```diff
@@ -0,0 +1,90 @@
...
+      it "contains correct data" do
+        expected_result = [
+          "honorificabilitudinitatibus honorificabilitudinitatibus honorificabilitudinitatibus " \
+          "honorificabili ­ tudinitatibus honorificabilitudinitatibus honorificabilitudinitatibus",
+          export_time_formatted,
+          "Page 1 of 1"
+        ]
+        result = pdf
+        expect(result.join(" ")).to eq(expected_result.join(" "))
+      end
```
EinLama, line 88: That's a nice spec! Very easy to read.

spec/models/work_package/pdf_export/document_generator_spec.rb
```diff
@@ -0,0 +1,90 @@
...
+  let(:export_time_formatted) { format_time(export_time, include_date: true) }
+  let(:export_pdf) do
+    Timecop.freeze(export_time) do
+      export.export!
+    end
+  end
+
+  subject(:pdf) do
+    content = export_pdf.content
+    # File.binwrite("WorkPackageDocumentGenerator-test-preview.pdf", content)
```
EinLama, line 49: A debug leftover? Could be useful to keep, but a short explanation in another comment above might be helpful for future use.

app/models/work_package/pdf_export/common/common.rb
```diff
@@ -26,7 +26,7 @@
 # See COPYRIGHT and LICENSE files for more details.
 #++
 
-module WorkPackage::PDFExport::Common
+module WorkPackage::PDFExport::Common::Common
```
EinLama, line 29: `Common::Common` 😄 That's a wonky name. Don't have anything better to offer though.

EinLama (changes requested):
Added a couple of questions and smaller notes.

I tried generating some PDFs and it works and looks great! Great job :) Couldn't see any performance issues due to creating it synchronously, but didn't try it with a giant work package.

app/components/work_packages/exports/generate/modal_dialog_component.rb
as-op: True, I also pre-sorted the array, so sorting is not necessary at runtime

app/models/work_package/pdf_export/common/attachments.rb
as-op: Done.

app/models/work_package/pdf_export/common/logo.rb
as-op: Thanks! Much nicer.

app/models/work_package/pdf_export/common/logo.rb
as-op: Done, included copyright and string freezing to every export file

app/models/work_package/pdf_export/generator/contracts.yml
as-op: Very good catch. For now only NotoSans is used, so I removed OpenSans that was only added for a spike.

spec/models/work_package/pdf_export/document_generator_spec.rb
as-op: It is quite helpful for debugging to actually see the PDF, so I left that in there but added a comment.

app/models/work_package/pdf_export/common/common.rb
as-op: Wonky indeed 😁

config/routes.rb
as-op: The other exports are GET, too and I wanted to have that consistent. 
So it’s enough to add .pdf to the end of your URL to start/download the export

e.g. https://community.openproject.org/work_packages/45896.pdf
or even on a query https://community.openproject.org/projects/openproject/work_packages.pdf?query_id=5307

Arguably it's less useful for `https://community.openproject.org/work_packages/45896/generate_pdf`, but I'm not aware of good reasons that a file download must be issued via a POST. But I don't have strong feelings about this, so I would be happy to hear your reasoning!

EinLama: I figured from a REST point of view, this route is creating an export, thus POST. The description of a work_package might change in-between exporting two PDFs, using POST prevents the browser from implicitly caching an old version of the file.

Both of these are rather weak arguments, though. And I was not aware that the other export already uses GET. I agree that consistency is a good reason to keep it that way.

app/components/work_packages/exports/generate/modal_dialog_component.rb
EinLama: nice, good catch

config/routes.rb
as-op: I have found a good argument against GET myself. 

For now passing the form values as query parameters is fine. In the future the form will probably contain file uploads, e.g. for a custom logo image. So this would need to be POST, so I will introduce this endpoint already as POST.

Feels more natural for a form endpoint anyhow. Fixed. Thanks 😀

EinLama (approved):
Nice work! 🎉

PR #15851, opened 2024-06-13
[#55523] Add Dialog not mobile friendly: don't center align on mobile

judithroth:
Before (mobile / desktop):
Screenshot from 2024-06-13 17-18-57
Screenshot from 2024-06-13 17-19-18

After (mobile / desktop):
Screenshot from 2024-06-13 17-18-33
Screenshot from 2024-06-13 17-21-36

### DIFF
--- a/app/components/settings/project_custom_fields/project_custom_field_mapping/new_project_mapping_component.html.erb
+++ b/app/components/settings/project_custom_fields/project_custom_field_mapping/new_project_mapping_component.html.erb
@@ -27,7 +27,7 @@
       ) do |form|
         concat(render(Primer::Alpha::Dialog::Body.new(
           id: dialog_body_id, test_selector: dialog_body_id, aria: { label: title },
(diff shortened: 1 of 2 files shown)


HDinger (changes requested):
That class name is kind of complicated. Nobody will understand why there is this limitation because normally you want an element in all screens center aligned. The only reason we do not want that here, is because we change the flex-direction with the other class. In that case, this class should then also be responsible for the correct alignment.
So I'd suggest something like this:

```sass
// For some specific cases like progress form in small screens,
// we need to change the direction of elements to column,
// therefore elements are placed below each other
.FormControl-horizontalGroup--sm-vertical
  @media screen and (max-width: $breakpoint-sm)
    .FormControl-horizontalGroup
      flex-direction: column
      row-gap: 1rem
      // Enforce the alignment due to lower specificity
      align-items: normal !important

.FormControl-horizontalGroup--center-aligned
  .FormControl-horizontalGroup
    align-items: center
```

HDinger (approved):

PR #9878, opened 2021-11-20
Work around bad hostname detection during postfix package install.

drmoose:
This pull request addresses a failure that occurs on `docker build -f docker/prod/Dockerfile .` if the docker host system is on a network that does not set a search domain.

## Bug

Trying to `apt-get install postfix` fails in some environments because its postinstall script [tries to detect a value for `myhostname` in my.cnf](https://salsa.debian.org/postfix-team/postfix-dev/-/blob/debian/buster-updates/debian/postfix.postinst#L40) using a very curious algorithm.

1. It reads `hostname --fqdn`, which because this is a docker build is always a randomly-generated container id.
2. It skips reading `/etc/hostname` because `hostname` returned something non-blank.
3. It tries checking `/etc/resolv.conf` for a search domain to append to the hostname it found, because it wants the hostname to have dots in it.

`/etc/resolv.conf` gets copied from the host by docker, which means the result of the dockerfile depends on what your `/etc/resolv.conf` contained when you ran `docker build`, and on systems that aren't on a network that has a search domain, results in an install failure with errors like these:

```
Running newaliases
newaliases: warning: valid_hostname: misplaced delimiter: 416a192f42bf..
newaliases: fatal: file /etc/postfix/main.cf: parameter myhostname: bad parameter value: 416a192f42bf..
dpkg: error processing package postfix (--configure):
 installed postfix package post-installation script subprocess returned error exit status 75
```

This was previously reported as [Ticket 31899](https://community.openproject.org/projects/openproject/work_packages/31899/) which was closed inactive a year ago because the OP disappeared. The last question asked on that ticket was:

> Hello, any reason to rebuild the image instead of pulling it from Docker Hub?

I don't know what the OP's reason was, but mine is because my group has an internal fork of openproject that we rebase and `docker build` whenever there's a new release.

## Fix

Because `postfix.postinst` honors `/etc/hostname` only if it is unable to call `/bin/hostname`, this fix works by temporarily moving `/bin/hostname` so that call fails, and supplying a fixed string as `/etc/hostname`.

## Side Effect

Currently the openproject on the docker hub embeds your CI environment's domain in the postfix configuration:

```
$ docker run --rm -it openproject/community:12 grep ^myhostname /etc/postfix/main.cf
-----> Setting PGVERSION=13 PGBIN=/usr/lib/postgresql/13/bin PGCONF_FILE=/etc/postgresql/13/main/postgresql.conf
[...]

PR #15853, opened 2024-06-13
apt-get update before attempting to install postfix

drmoose:
Moves the `apt-get update` command above the `apt-get install` on line 13-14.

My previous PR #9878 added a docker build workaround for systems without dots in their fqdn as reported by /etc/resolv.conf.

More recent versions of openproject appear to have changed when `apt-get update` is run, and so the `apt-get install` I introduced previously to correct that bug no longer works.

### DIFF
--- a/docker/prod/setup/postinstall-onprem.sh
+++ b/docker/prod/setup/postinstall-onprem.sh
@@ -1,6 +1,7 @@
 #!/bin/bash
 set -eox pipefail
 
+apt-get update -qq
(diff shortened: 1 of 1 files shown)


Kharonus (approved):
I see no issue here, thanks for the fix. Apparently the added `apt-get update -qq` was simply inserted in the wrong line, maybe not seeing that there was another `apt-get install` for the postfix.

PR #15859, opened 2024-06-14
[#55529] fixed interface of file info query

Kharonus:
[#55529](https://community.openproject.org/work_packages/55529)

- location data is returned escaped
- id data is always a string
- additional encoding in frontend removed


dominic-braeunlein (approved):
LGTM

PR #15867, opened 2024-06-17
Fix for 500 of /projects URL which sorted by latest_activity_at

Eric-Guo:
I'm not sure this fix is right or not in implementation, but at least it resolve the 500 error at [THAPE staging server](https://plm-ppp.thape.com.cn/projects?query_id=my&sortBy=%5B%5B%22latest_activity_at%22%2C%22asc%22%5D%5D)

image

### DIFF
--- a/app/models/current.rb
+++ b/app/models/current.rb
@@ -26,45 +26,6 @@
 # See COPYRIGHT and LICENSE files for more details.
 #++
 
-class Queries::Projects::ProjectQuery < ApplicationRecord
-  include Queries::BaseQuery
-  include Queries::Serialization::Hash
-
-  belongs_to :user
-
-  serialize :filters, coder: Queries::Serialization::Filters.new(self)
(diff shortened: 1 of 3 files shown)


klaustopher:
Could you write a proper bug report for the behavior you are seeing and when you encounter that error? Sorting by the last activity does not show a 500 here.

klaustopher:
You could also add a failing spec, that we could start working off

Eric-Guo:
OP#55783 [created](https://community.openproject.org/wp/55783).

Eric-Guo:
@klaustopher I also add a spec, the implementation is correct, but the key problem is I have to introduce `Class Current < ActiveSupport::CurrentAttributes` to store the project model.

klaustopher:
@Eric-Guo Thanks for the failing spec, can you try #15883 and see if this fixes your problem?

PR #15871, opened 2024-06-17
Makes sure Commands/Queries handle 403 responses

mereghost:
## What?

This tries to make sure that all commands handle 403s correctly. 

## How?

By handling 403s! Surprising, huh? But in all seriousness there are some standardization of the way we treat the errors across different commands as well as some code cleanup where needed.

### DIFF
--- a/modules/storages/app/common/storages/peripherals/storage_interaction/one_drive/util.rb
+++ b/modules/storages/app/common/storages/peripherals/storage_interaction/one_drive/util.rb
@@ -69,10 +69,12 @@ module Storages
                 )
             end
 
-            def storage_error(response:, code:, source:)
-              data = StorageErrorData.new(source:, payload: response.json(symbolize_keys: true))
+            def storage_error(response:, code:, source:, log_message: nil)
+              # Some errors, like timeouts, aren't json responses so we need to adapt
+              payload = response.respond_to?(:json) ? response.json(symbolize_keys: true) : response.to_s
+              data = StorageErrorData.new(source:, payload:)
 
-              StorageError.new(code:, data:)
+              StorageError.new(code:, data:, log_message:)
             end
 
             def join_uri_path(uri, *)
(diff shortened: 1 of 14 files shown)


modules/storages/app/common/storages/peripherals/storage_interaction/one_drive/util.rb
```diff
@@ -69,10 +69,12 @@ def using_user_token(storage, user, &)
                 )
             end
 
-            def storage_error(response:, code:, source:)
-              data = StorageErrorData.new(source:, payload: response.json(symbolize_keys: true))
+            def storage_error(response:, code:, source:, log_message: nil)
+              # Some errors, like timeouts, aren't json responses so we need to adapt
+              payload = response.respond_to?(:json) ? response.json(symbolize_keys: true) : response.to_s
```
Kharonus, line 74: 🟡 rather then using `respond_to?(:json)`, can we use `response.content_type.mime_type`? in the nextcloud util we have such a matching: 

```ruby
            def error_data_from_response(caller:, response:)
              payload =
                case response.content_type.mime_type
                when "application/json"
                  response.json
                when "text/xml", "application/xml"
                  response.xml
                else
                  response.body.to_s
                end

              StorageErrorData.new(source: caller, payload:)
            end
```

Kharonus (changes requested):
There is only one issue I want to raise

modules/storages/app/common/storages/peripherals/storage_interaction/one_drive/util.rb
mereghost: It doesn't seem to be the case as we have 2 things in play here: 

1. a `Response` that happens to be an error to us
2. an `HTTPXErrorResponse` which happens when the request wasn't able to be made at all

So the first one we could parse the mime_type (in theory should just be json as this is only for the MS API) but on the second one not really.

Kharonus (approved):

modules/storages/app/common/storages/peripherals/storage_interaction/one_drive/util.rb
Kharonus: I think, the current approach is alright. In the end, it is just a subjective point of view.

PR #15873, opened 2024-06-17
[chore] Add regression tests for storage folders

Kharonus:
### WAT?
- locations with ampersands are now specifically tested in the query unit tests
- file info query tests refactored to use shared examples for all storage providers
- files query of one drive now return escaped location data, too


modules/storages/app/common/storages/peripherals/storage_interaction/nextcloud/file_info_query.rb
Kharonus, line 1: Wow, github really cannot display new indentions :/

mereghost (approved):
:shipit:

modules/storages/app/common/storages/peripherals/storage_interaction/nextcloud/file_info_query.rb
```diff
@@ -28,107 +28,122 @@
...
-        case response_object.ocs.data.statuscode
-        when 200..299
-          ServiceResult.success(result: response_object)
-        when 403
-          Util.error(:forbidden, "Access to storage file forbidden!", error_data)
-        when 404
-          Util.error(:not_found, "Storage file not found!", error_data)
-        else
-          Util.error(:error, "Outbound request failed!", error_data)
+module Storages
```
ba1ash, line 31: It could have been written like that to have the same meaning you want and not changing indentation for whole file:
```
module Storages; module Peripherals; module StorageInteraction; module Nextcloud
  class FileInfoQuery
  ...
  end
end; end; end; end
```

Kharonus: right, sadly the autoformatter I use has no rules to cope with that ;)

PR #15887, opened 2024-06-18
[#55771] Warn user that the type is not available when moving a work package

aaron-contreras:
Displays a warning toast letting the user know to activate the current types of the work packages under move in the target project and making the "auto-reassign" behavior known to them so they can better take a decision when performing moves.

See: https://community.openproject.org/work_packages/55771

### DIFF
--- a/app/controllers/work_packages/moves_controller.rb
+++ b/app/controllers/work_packages/moves_controller.rb
@@ -117,10 +117,24 @@ class WorkPackages::MovesController < ApplicationController
     @target_project ||= @project
     @types = @target_project.types.order(:position)
     @target_type = @types.find { |t| t.id.to_s == params[:new_type_id].to_s }
-    @unavailable_type_in_target_project = @types.exclude?(@target_type)
+    @unavailable_type_in_target_project = set_unavailable_type_in_target_project
     @available_versions = @target_project.assignable_versions
     @available_statuses = Workflow.available_statuses(@project)
     @notes = params[:notes] || ""
+  end
+
+  def set_unavailable_type_in_target_project
+    if @target_project == @project
+      false
+    elsif @target_type.nil?
+      Type.where(id: @work_packages.select(:type_id))
+          .select("distinct id")
+          .pluck(:id)
+          .difference(@types.pluck(:id))
+          .any?
+    else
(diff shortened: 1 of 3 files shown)


dombesz (commented):
The warning seems to be coming on, even if I don't change anything on the move form yet.
<details><summary>Bug Demo</summary>
<p>

Jun-20-2024 10-16-32


</p>
</details> 


<details><summary>Just an observation,the toast seems to have no space on top. Please ignore it if that's the intended look.</summary>
<p>

image


</p>
</details>

aaron-contreras:
> The warning seems to be coming on, even if I don't change anything on the move form yet.

I see this as being correct though, it's warning you if you haven't done anything because not doing anything will cause the type to be automatically re-assigned.

> Just an observation,the toast seems to have no space on top. Please ignore it if that's the intended look.

There were no visuals for it but I wanted to make it seem tied to the field above and not just floating around 😅 . Do you have another idea?

app/controllers/work_packages/moves_controller.rb
```diff
@@ -114,8 +115,9 @@ def prepare_for_work_package_move
     @allowed_projects = WorkPackage.allowed_target_projects_on_move(current_user)
     @target_project = @allowed_projects.detect { |p| p.id.to_s == params[:new_project_id].to_s } if params[:new_project_id]
     @target_project ||= @project
-    @types = @target_project.types
+    @types = @target_project.types.order(:position)
     @target_type = @types.find { |t| t.id.to_s == params[:new_type_id].to_s }
+    @unavailable_type_in_target_project = @types.exclude?(@target_type)
```
dombesz: If the `@target_project == @project` then the `@unavailable_type_in_target_project` should be `false`.

Furthermore, when I move a WP with type task, that is enabled in the target project too while leaving the `Type` on `(No change)`, the `params[:new_type_id]` will not be set, which in turn makes the warning to be shown. I believe that is incorrect.

aaron-contreras: Nice catch. I understand now what you mean

Addressed in latest commit!

dombesz (approved):
Nice job Aaron! 👍

PR #15893, opened 2024-06-19
Handle errors when Copying Project Folders when the Storage is broken

mereghost:
## What?

I hadn't handled the issue of trying to copy a project folder on a broken storage. This PR aims to fix that.

## How?

Well... by handling the service result failure. =/

This PR also:

1. Adds the `auth_strategy` to `CopyTemplateFolder`
2. Standardizes the error responses for `CopyTemplateFolder`

### DIFF
--- a/modules/storages/app/common/storages/peripherals/storage_interaction/nextcloud/copy_template_folder_command.rb
+++ b/modules/storages/app/common/storages/peripherals/storage_interaction/nextcloud/copy_template_folder_command.rb
@@ -33,8 +33,8 @@ module Storages
         class CopyTemplateFolderCommand
           using ServiceResultRefinements
 
-          def self.call(storage:, source_path:, destination_path:)
-            new(storage).call(source_path:, destination_path:)
+          def self.call(auth_strategy:, storage:, source_path:, destination_path:)
+            new(storage).call(auth_strategy:, source_path:, destination_path:)
           end
 
           def initialize(storage)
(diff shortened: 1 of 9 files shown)


modules/storages/app/common/storages/peripherals/storage_interaction/nextcloud/copy_template_folder_command.rb
```diff
@@ -33,28 +33,27 @@ module Nextcloud
         class CopyTemplateFolderCommand
           using ServiceResultRefinements
 
-          def self.call(storage:, source_path:, destination_path:)
-            new(storage).call(source_path:, destination_path:)
+          def self.call(auth_strategy:, storage:, source_path:, destination_path:)
```
Kharonus, line 36: 🟢 good thing I saw that now, was going to start with that command almost myself.

modules/storages/app/workers/storages/copy_project_folders_job.rb
```diff
@@ -42,6 +42,7 @@ def perform(source:, target:, work_packages_map:)
       user = batch.properties[:user]
 
       project_folder_result = results_from_polling || initiate_copy(target)
+      project_folder_result.on_failure { |failed| return log_failure(failed) }
```
Kharonus, line 45: ❓ so, this is the actual fix, the rest is boyscouting, right? Or did I miss another piece of changes?

Kharonus (approved):
LGTM, thanks for the auth rework of those commands. Just a question left.

modules/storages/app/workers/storages/copy_project_folders_job.rb
mereghost: Yep, this is the entire fix, the rest was just a case of "Ok, since I'm here already..."

PR #15894, opened 2024-06-19
[#55382] add connection validation section

Kharonus:
[#55382](https://community.openproject.org/work_packages/55382)

### WAT?

- added connection validation section to sidebar
  - allows manual request of a health check for this storage
  - checks for tenant id, drive id, oauth credentials, AMPF content
- added new controller for connection validation
- show connection validation only for one drive storages


Gemfile
```diff
@@ -210,6 +210,7 @@ gem "validate_url"
 
 # Storages support code
 gem "dry-container"
+gem "dry-monads"
```
akabiru, line 213: Sweet 👍🏾

modules/storages/app/components/storages/admin/sidebar/validation_result_component.html.erb
```diff
@@ -0,0 +1,57 @@
...
+
+<%=
+  flex_layout do |container|
+    container.with_row do
+      render(Primer::Beta::Text.new(font_weight: :bold, test_selector: "validation-result--subtitle")) do
+        I18n.t("storages.health.connection_validation.subtitle")
+      end
+    end
+
+    if @result.type != :none
```
akabiru: 🟢 _nit:_ could we encapsulate this check in a named method so we can add meaning to it? Ex. `@result.connection_validation_exists?`

modules/storages/app/controllers/storages/admin/connection_validation_controller.rb
```diff
@@ -0,0 +1,62 @@
...
+# along with this program; if not, write to the Free Software
+# Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+#
+# See COPYRIGHT and LICENSE files for more details.
+#++
+
+module Storages
+  module Admin
+    class ConnectionValidationController < ApplicationController
+      include OpTurbo::ComponentStream
```
akabiru, line 34: 🟠 Because we've mixed in these helpers, we can simplify the `#validate_connection` action via the following helpers:

- _update_via_turbo_stream_ - replaces the need for `modules/storages/app/views/storages/admin/connection_validation/validate_connection.turbo_stream.erb` as it can take in a component with `OpTurbo::Streamable` properties which basically removes the need for defining our own `OpTurbo::FrameComponent` as it already generates a unique `wrapper_id` based on the component name
- _respond_with_turbo_streams_ - returns all component streams that were added to the controller i.e. we can have multiple Ex. https://github.com/opf/openproject/blob/dev/app/controllers/admin/settings/project_custom_fields_controller.rb#L150-L163

----

Thereby, connection_validation_controller could be:

```git
--- a/modules/storages/app/controllers/storages/admin/connection_validation_controller.rb
+++ b/modules/storages/app/controllers/storages/admin/connection_validation_controller.rb
@@ -45,10 +45,9 @@ module Storages
         @result = Peripherals::OneDriveConnectionValidator
                     .new(storage: @storage)
                     .validate
+        update_via_turbo_stream(component: Storages::Admin::Sidebar::ValidationResultComponent.new(result: @result))

-        respond_to do |format|
-          format.turbo_stream
-        end
+        respond_with_turbo_streams
       end

       private
```

coupled by making ValidationResultComponent an OpTurbo::Streamable component

```git
--- a/modules/storages/app/components/storages/admin/sidebar/validation_result_component.rb
+++ b/modules/storages/app/components/storages/admin/sidebar/validation_result_component.rb
@@ -33,6 +33,7 @@ module Storages
     module Sidebar
       class ValidationResultComponent < ApplicationComponent # rubocop:disable OpenProject/AddPreviewForViewComponent
+        include OpTurbo::Streamable

         def initialize(result:)
           super(result)
```

wrap ValidationResultComponent with `component_wrapper` so the component id is set

```git
--- a/modules/storages/app/components/storages/admin/sidebar/validation_result_component.html.erb
+++ b/modules/storages/app/components/storages/admin/sidebar/validation_result_component.html.erb
@@ -28,28 +28,30 @@ See COPYRIGHT and LICENSE files for more details.
 ++#%>

 <%=
-  flex_layout do |container|
-    container.with_row do
-      render(Primer::Beta::Text.new(font_weight: :bold, test_selector: "validation-result--subtitle")) do
-        I18n.t("storages.health.connection_validation.subtitle")
+  component_wrapper do
+    flex_layout do |container|
+      container.with_row do
+        render(Primer::Beta::Text.new(font_weight: :bold, test_selector: "validation-result--subtitle")) do
+          I18n.t("storages.health.connection_validation.subtitle")
+        end
       end
-    end

-    if @result.type != :none
-      container.with_row(mt: 2) do
-        status = status_indicator
+      if @result.type != :none
+        container.with_row(mt: 2) do
+          status = status_indicator

-        concat(render(Primer::Beta::Text.new(pr: 2, test_selector: "validation-result--timestamp")) do
-          I18n.t('storages.health.checked', datetime: helpers.format_time(@result.timestamp))
-        end)
-        concat(render(Primer::Beta::Label.new(scheme: status[:scheme])) { status[:label] })
+          concat(render(Primer::Beta::Text.new(pr: 2, test_selector: "validation-result--timestamp")) do
+            I18n.t('storages.health.checked', datetime: helpers.format_time(@result.timestamp))
+          end)
+          concat(render(Primer::Beta::Label.new(scheme: status[:scheme])) { status[:label] })
+        end
       end
-    end

-    if @result.description.present?
-      container.with_row(mt: 2) do
-        render(Primer::Beta::Text.new(color: :muted, test_selector: "validation-result--description")) do
-          @result.description
+      if @result.description.present?
+        container.with_row(mt: 2) do
+          render(Primer::Beta::Text.new(color: :muted, test_selector: "validation-result--description")) do
+            @result.description
+          end
         end
       end
     end
```

updating the rendering process, removing the turbo frame component wrapper

```git
--- a/modules/storages/app/components/storages/admin/sidebar/health_status_component.html.erb
+++ b/modules/storages/app/components/storages/admin/sidebar/health_status_component.html.erb
@@ -36,9 +36,7 @@ See COPYRIGHT and LICENSE files for more details.

       if @storage.provider_type_one_drive?
         health_status_container.with_row(mt: 2) do
-          render(OpTurbo::FrameComponent.new(id: :connection_validation_result)) do
-            render(Storages::Admin::Sidebar::ValidationResultComponent.new(result: validation_result_placeholder))
-          end
+          render(Storages::Admin::Sidebar::ValidationResultComponent.new(result: validation_result_placeholder))
         end

         health_status_container.with_row(mt: 2) do
```

finally **_deleting_** modules/storages/app/views/storages/admin/connection_validation/validate_connection.turbo_stream.erb as it will no longer be used

modules/storages/app/components/storages/admin/sidebar_component.html.erb
```diff
@@ -2,7 +2,10 @@
   component_wrapper do
     render(Primer::OpenProject::BorderGrid.new) do |border_grid|
       border_grid.with_row { render(Storages::Admin::Sidebar::HealthStatusComponent.new(storage: @storage)) }
-      border_grid.with_row { render(Storages::Admin::Sidebar::HealthNotificationsComponent.new(storage: @storage)) }
+      
+      if @storage.automatic_management_enabled?
+        border_grid.with_row { render(Storages::Admin::Sidebar::HealthNotificationsComponent.new(storage: @storage)) }
+      end
```
akabiru, line 8: 🟢 _fyi:_ the HealthNotificationsComponent was already using [conditional rendering](https://viewcomponent.org/guide/conditional_rendering.html) based on `@storage.automatic_management_enabled?` but perhaps this is more clear- my impression is that `border_grid.with_row {}` does not modify the DOM if the render result is nil _*unverified_

https://github.com/opf/openproject/blob/dev/modules/storages/app/components/storages/admin/sidebar/health_notifications_component.rb#L41-L43

modules/storages/app/common/storages/peripherals/one_drive_connection_validator.rb
```diff
@@ -0,0 +1,152 @@
...
+
+      def validate
+        maybe_is_not_configured
+          .or { tenant_id_wrong }
+          .or { client_id_wrong }
+          .or { client_secret_wrong }
+          .or { drive_id_wrong }
+          .or { request_failed_with_unknown_error }
+          .or { drive_with_unexpected_content }
+          .value_or(ConnectionValidation.new(type: :healthy, timestamp: Time.current, description: nil))
```
akabiru: 🟢 The monads are a really nice touch, but I wonder if we can follow a positive naming scheme- the negation based naming can sometimes be confusing (at least to me)
```suggestion
        maybe_storage_configured
          .or { validate_tenant_id }
          .or { validate_client_id }
          .or { validate_client_secret }
          .or { validate_drive_id }
          .or { request_failed_with_unknown_error }
          .or { drive_with_unexpected_content }
          .value_or(ConnectionValidation.new(type: :healthy, timestamp: Time.current, description: nil))
```

modules/storages/app/models/storages/connection_validation.rb
```diff
@@ -0,0 +1,33 @@
...
+#
+# You should have received a copy of the GNU General Public License
+# along with this program; if not, write to the Free Software
+# Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+#
+# See COPYRIGHT and LICENSE files for more details.
+#++
+
+module Storages
+  ConnectionValidation = Data.define(:type, :description, :timestamp)
```
akabiru: ✨

modules/storages/app/views/storages/admin/connection_validation/validate_connection.turbo_stream.erb
```diff
@@ -0,0 +1,32 @@
...
+along with this program; if not, write to the Free Software
+Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+
+See COPYRIGHT and LICENSE files for more details.
+
+++#%>
+
+<%= turbo_stream.update :connection_validation_result do %>
+  <%= render(Storages::Admin::Sidebar::ValidationResultComponent.new(result: @result)) %>
+<% end %>
```
akabiru: 🟠 https://github.com/opf/openproject/pull/15894#discussion_r1648856156 we can get rid of this file

modules/storages/spec/common/storages/peripherals/one_drive_connection_validator_spec.rb
```diff
@@ -0,0 +1,164 @@
...
+             :as_automatically_managed,
+             project_folder_id:,
+             storage:,
+             project: create(:project))
+    end
+    let(:files_result) do
+      Storages::StorageFiles.new(
+        [
+          Storages::StorageFile.new(id: project_folder_id, name: "I am your father"),
+          Storages::StorageFile.new(id: "noooooooooo", name: "testimony_of_luke_skywalker.md")
```
akabiru, line 136: 🤣 🙌🏾

modules/storages/spec/common/storages/peripherals/one_drive_connection_validator_spec.rb
```diff
@@ -0,0 +1,164 @@
...
+    end
+    let(:response) { ServiceResult.success(result: files_result) }
+
+    before do
+      project_storage
+    end
+
+    it "returns a validation failure" do
+      expect(subject.type).to eq(:warning)
+      expect(subject.description).to eq("Unexpected content found in the drive.")
```
akabiru, line 151: 🟠  I encountered this message while testing and I honestly had no clue what it means or what actions I should take next 🤷🏾 could we redefine it? Albeit I don't have a suggestion for a better one

---

Screenshot 2024-06-21 at 3 12 38 PM

modules/storages/spec/requests/storages/admin/connection_validation_spec.rb
```diff
@@ -0,0 +1,136 @@
...
+# You should have received a copy of the GNU General Public License
+# along with this program; if not, write to the Free Software
+# Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+#
+# See COPYRIGHT and LICENSE files for more details.
+#++
+
+require "spec_helper"
+
+RSpec.describe "connection validation", :skip_csrf do
```
akabiru, line 31: 🟢  dum Q: why do we need to "skip csrf" if the form is coming from within OP via turbo stream?

modules/storages/spec/requests/storages/admin/connection_validation_spec.rb
```diff
@@ -0,0 +1,136 @@
...
+  def xpath_for_label
+    "//turbo-stream[@target='connection_validation_result']/template/div/div/span[contains(@class, 'Label')]"
+  end
+
+  def xpath_for_description
+    "//turbo-stream[@target='connection_validation_result']/template/div/div/span[@data-test-selector='validation-result--description']"
+  end
+
+  # rubocop:enable Layout/LineLength
+end
```
akabiru, line 142: 🟢 I'm curious why a request spec rather than a feature spec? The latter would be more fitting IMO

akabiru (commented):
Nice one @Kharonus 🏆 I have a couple of 🟢 & 🟠 but nothing blocking. Shout if easier to chat/pair 🙂

modules/storages/app/controllers/storages/admin/connection_validation_controller.rb
Kharonus: That's interesting, will check your changes. ;)

modules/storages/spec/common/storages/peripherals/one_drive_connection_validator_spec.rb
Kharonus: fair point, will talk to @dominic-braeunlein about it

modules/storages/app/common/storages/peripherals/one_drive_connection_validator.rb
Kharonus: Ah, well, that is interesting. Especially due to the `or` concatenation I naturally read negative namings. It seems more natural then the positive way.

See this imaginary monologue.

"So, I have this storage.... I suspect it to be invalid."
"Maybe the storage is not fully configured?"
"Well, apparently it is... so, is then tenant id wrong?"
"Hmm, not... what about the client id?"
"Or the client secret?"
"Anything of that? No? Then apparently you are valid."

Get what I am pointing at? :D validations are "positive" when they are "negative", hence finding a problem means, your validation was justified.

akabiru: > Get what I am pointing at? :D validations are "positive" when they are "negative", hence finding a problem means, your validation was justified.

I see wym for sure, and perhaps [ActiveModel::Validations](https://api.rubyonrails.org/classes/ActiveModel/Validations.html) are a great conventional example; they only push to `errors` in the even "negative" event that there's a validation failure, otherwise they just return ?true.

Hence in your imaginany monologue 😄 mine would read like so:

"So I have this storage... I need to validate that everything is configured properly"
"Firstly, validated it is fully configured"
"Then, validate the tenant id is correct"
"Then, validate the client id is correct"
"Then, validate the client secret is correct"
"When there are invalid cases, collect them and display a unified response (similar to active_model_errors) to the user"
"When there are no errors, return- mark the storage connection as valid"

---

I reckon what's different with my monologue is that I presume we're collecting all errors- and displaying them once to the user, but AFAICT- we fail fast on the first error result? In which case that might be inconvenient if the user has to click the button multiple times because they are served one error at a time.

In which case, I wonder if just mixin in [ActiveModel::Validations](https://api.rubyonrails.org/classes/ActiveModel/Validations.html) and writing these as validate statements would be more fitting albeit less "Monad-ish" 😄

mereghost: > but AFAICT- we fail fast on the first error result? 

Here's the trick, except for the missing/empty values, it is impossible to validate if the previous validation has failed. If the tenant id is wrong, we can't validate the Client ID/Secret, if the Client ID/Secret is wrong we can't validate the Drive ID.

> In which case, I wonder if just mixin in [ActiveModel::Validations](https://api.rubyonrails.org/classes/ActiveModel/Validations.html) and writing these as validate statements would be more fitting albeit less "Monad-ish" 😄

There are always more than one way to do things. A `dry/validation` contract, `ActiveModel::Validations` or a chain of assertions all could do the same thing.

1. ActiveModel would be tied directly to an object (with certain assumptions). 
2. `Dry::Validation::Contract` would create an object just for validation of a hash/json that could be passed down after validation.
3. A chain of assertions (in this case using the Maybe/Option monad) makes explicit the validation process.

modules/storages/app/controllers/storages/admin/connection_validation_controller.rb
Kharonus: updated

modules/storages/app/components/storages/admin/sidebar/validation_result_component.html.erb
Kharonus: done

modules/storages/app/common/storages/peripherals/one_drive_connection_validator.rb
Kharonus: > Here's the trick, except for the missing/empty values, it is impossible to validate if the previous validation has failed. If the tenant id is wrong, we can't validate the Client ID/Secret, if the Client ID/Secret is wrong we can't validate the Drive ID.

Yes, and here we should design it, as the domain user story is telling us. And here, it was set, we do not show all errors. We do show the first error, that can happen. Hence, the implementation follows that story, and fails with the first error, not checking for the other validation rules.

modules/storages/app/views/storages/admin/connection_validation/validate_connection.turbo_stream.erb
Kharonus: removed

modules/storages/spec/common/storages/peripherals/one_drive_connection_validator_spec.rb
Kharonus: I talked to dominic, here is the outcome:
- we will add error codes to the description
- we won't add more information to the description text
- we will update the documentation, what the error/warning code means and what the possible explanation is and how to solve it.

modules/storages/spec/requests/storages/admin/connection_validation_spec.rb
Kharonus: I basically added two unit tests, one for the controller, one for the validator class I built. Unless there is a very good reason for a feature spec, I would not go for it.

modules/storages/spec/requests/storages/admin/connection_validation_spec.rb
Kharonus: good question, Christophe told me to use it. lemme check.

Edit: Well, without `:skip_csrf` every request returns 422.

modules/storages/app/components/storages/admin/sidebar_component.html.erb
Kharonus: Sadly, the separator of the border grid is shown, even when `render` returns `nil`. If I encapsulate the render method like this, the separator is not shown.

Do you have an idea, how to solve this with other means?

akabiru: Ah, yeah- I'd leave it out tbh- my proposal wasn't fully verified either way. Cheers for looking into it! 👍🏾

akabiru (approved):
:shipit: Cheers Eric!

PR #15899, opened 2024-06-19
Generalize sharing modal

klaustopher:
This extracts the sharing modal from the WorkPackage scope to a more generally usable thing.

- [x] Move the route config into a routing concern
- [x] Move `WorkPackages::Shares` and Bulk controller out of the WP scope
- [x] Remove the hard coded loading of the `WorkPackage` from the `SharesController` and generalize it 
- [x] Move the non-WorkPackage specific components into `components/share`
  - [x] Define `available_roles` in the controller and pass them down based on entity
  - [x] Define user types based on if the `Entity` is project scoped or not
  - [x] Permission checks need to be passed in from the controller as well (@aaron-contreras )
- [x] Move the Service classes into a generic one, leave contracts in the shared resources scope
  - [x] Proper contracts that are scoped to the entity and do the correct checks 
- [x] Get rid of all the `work_package` names in CSS classes, etc
- [x] Move i18n strings outside of the `work_package` scope
- [x] Move i18n strings outside of the `work_package` scope for JS files
- ~Generalize the Javascript component for the share modal: ?`WorkPackageShareModalComponent`~ We probably don't need that as we can add a new action for the modal to show up and then get rid of the entire component

---

Part of https://community.openproject.org/work_packages/55162/activity


aaron-contreras:
@klaustopher my bad 😅

app/controllers/concerns/shares/work_packages/authorization.rb
```diff
@@ -28,15 +28,22 @@
...
+  module WorkPackages
+    module Authorization
+      extend ActiveSupport::Concern
 
-        included do
-          def sharing_manageable?
-            User.current.allowed_in_project?(:share_work_packages, @work_package.project)
+      included do
+        def sharing_manageable?
+          # TODO: Fix this to check based on the entity
```
ulferts, line 38: ❓ I guess the TODO can be removed.

If that is in fact the case and this is now no longer work package specific, It should probably be moved out of the WorkPackages namespace.

app/controllers/shares_controller.rb
```diff
@@ -0,0 +1,376 @@
...
+                           ]
+                         else
+                           []
+                         end
+  end
+
+  def sharing_contract_scope
+    if @entity.is_a?(WorkPackage)
+      Shares::WorkPackages
+    end
```
ulferts, line 374: 💡 This file still includes a number of places that are either work packages specific or set out to end up having a very long list of case/if statements (e.g. `enterprise_check`, `available_roles`, `sharing_contract_scope`, `load_entity`). And then there is `Shares::WorkPackages::Authorization` which looks to follow along.

It might be a possibility to introduce a strategy objects per shareable entity where each centralizes the adaptation of this controller here but potentially also other places. That way, the controller wouldn't have to be touched every time a shareable is added.

Depending on how far ones to take that idea, the strategy could also be called from inside the components reducing the number of parameters that need to be passed around. The contracts might be another place that would profit from a strategy (or reusing the single one per sharable model) instead of inheriting.

app/components/shares/modal_body_component.rb
```diff
@@ -0,0 +1,204 @@
...
+#
+# You should have received a copy of the GNU General Public License
+# along with this program; if not, write to the Free Software
+# Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+#
+# See COPYRIGHT and LICENSE files for more details.
+#++
+
+module Shares
+  class ModalBodyComponent < ApplicationComponent # rubocop:disable OpenProject/AddPreviewForViewComponent
```
ulferts, line 30: With the intend of this being reused in other places, having preview/documentation for the Modal component seems like a good idea now or at least when the modal is reused for ProjectList

app/models/queries/members/entity_member_query.rb
```diff
@@ -28,7 +28,7 @@
 # See COPYRIGHT and LICENSE files for more details.
 # ++
 
-class Queries::Members::WorkPackageMemberQuery < Queries::Members::MemberQuery
+class Queries::Members::EntityMemberQuery < Queries::Members::MemberQuery
   def default_scope
     Member.joins(:member_roles).merge(MemberRole.only_non_inherited)
   end
```
ulferts, line 34: Not originating from this PR but the query has too broad a scope. It will return all Members (with non inherited member roles) and not just those that are entity specific. It probably hasn't caused any issues since it is supposedly always used in combination with a filter but it might be surprising if it is ever used in a different context and potentially dangerous if ever it is made accessible via the API.

ulferts (approved):
The change looks good functionality and code wise. I left a couple of remarks but they don't need to be resolved before merging.

app/controllers/concerns/shares/work_packages/authorization.rb
klaustopher: This is already done in the follow up PR #15971, see

https://github.com/opf/openproject/pull/15971/files#diff-8f52e03ddddb5d34ba543b34fe41505f557c5d25b2040f5d2b9c1b151184b9bcR404-R417

app/controllers/shares_controller.rb
klaustopher: I am 100% with you, this was also something I was worrying about, and I like the concept of a `WorkPackageSharingStrategy` hat has all of the logic about allowed checks, roles, etc in it. We will introduce this in #15971

app/components/shares/modal_body_component.rb
klaustopher: In #15971 we will be introducing a `ShareDialogComponent` that will be the entire dialog, currently this "main" component is only the body because it is rendered in a SPOT Modal in WorkPackages. For this I completely agree to add it to the lookbook

app/models/queries/members/entity_member_query.rb
klaustopher: Should we maybe rename it to `OnlyNonInherited`?

app/controllers/shares_controller.rb
klaustopher: Added as a TODO in the description of #15971

app/models/queries/members/entity_member_query.rb
klaustopher: I did the renaming in https://github.com/opf/openproject/pull/15971/commits/30139b99c6438b160591458479e268ca3444cf29

PR #15908, opened 2024-06-20
Fix small typo in lookbook

judithroth:
When the braces are in the wrong position the button will appear empty.

--

I wasn't sure who would be a good reviewer. If I should select someone else, please let me know :)


HDinger (approved):
Good catch, thanks 👍

PR #15916, opened 2024-06-21
Stop interrupting the sync process on error.

mereghost:
## What?

When an error during the sync process happened, we were raising an exception and that was stopping the other storages from completing their own sync processes

## How?

I moved the exception raising to outside of the processing block and the unhealthy event emission to the the `#perform` method.

Also added a test to verify the behaviour.

## And now?

This is a hacky fix. We need to make the sync job work over a specific storage / project as a more long term solution, but this will take sometime and will be addressed in a following PR.

### DIFF
--- a/modules/storages/app/common/storages/errors.rb
+++ b/modules/storages/app/common/storages/errors.rb
@@ -48,25 +48,17 @@ module Storages
       end
     end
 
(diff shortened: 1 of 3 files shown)


ba1ash (approved):
LGTM!

PR #15919, opened 2024-06-21
[Bug #55901] 403 when entering the overview page of a project in which a work package is shared with the current user

ulferts:
Without the permission, the project overview page of a work package shared with the user displays a 403.

### WP

https://community.openproject.org/wp/55901


aaron-contreras (approved):
👍

PR #15920, opened 2024-06-21
bump grape

ulferts:
Mostly an everyday bump of grape but:
*  the [`error_response` method has been removed](https://github.com/ruby-grape/grape/blob/master/UPGRADING.md#changes-in-rescue_from) - this does not lead to any visible changes
* the [path detection with `route_params` has been changed](https://github.com/ruby-grape/grape/blob/master/UPGRADING.md#recognizing-path). Because of that, on a parameter not conforming to the type (e.g. sending `api/v3/statuses/blubs` where an Integer is required) will lead to the path not being recognized resulting in a 404. Before, a 400 BadRequest was returned. It does not seem that the beviour of returning a 400 in such cases had been documented and returning a 404 is a valid response. There does not seem to be a way to continue returning a 400. 
* All unknown paths within the API realm now return an error object conforming to the other API errors.

### DIFF
--- a/lib/api/utilities/grape_helper.rb
+++ b/lib/api/utilities/grape_helper.rb
@@ -71,7 +71,7 @@ module API
           original_exception = $!
           representer = error_representer.new e
           resp_headers = instance_exec &headers
-          env["api.format"] = error_content_type
+          resp_headers["Content-Type"] = error_content_type
 
           if log == true
             OpenProject.logger.error original_exception, reference: :APIv3
--- a/lib/api/utilities/grape_helper.rb
+++ b/lib/api/utilities/grape_helper.rb
@@ -38,6 +38,10 @@ module API
           @env = env
           @options = {}
         end
+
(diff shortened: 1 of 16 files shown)


lib/api/utilities/grape_helper.rb
```diff
@@ -38,6 +38,10 @@ def initialize(env)
           @env = env
           @options = {}
         end
+
+        def error!(message, status = nil, headers = nil, backtrace = nil, original_exception = nil)
+          super
+        end
```
klaustopher, line 44: Not really understanding what this method does here? If it is just calling super with all the arguments passed as well, why does it exist?

klaustopher (approved):

lib/api/utilities/grape_helper.rb
ulferts: It is just there to make it public. Could also have used `send` but wasn't sure if that would catch all the cases.

PR #15923, opened 2024-06-21
TimeEntriesActivityRepresenter HAL fix (tooo long loading)

foxweb:
Hi @ulferts!

Referenced bug — https://community.openproject.org/projects/openproject/work_packages/54142/activity
PR https://github.com/opf/openproject/pull/15275 — it really doesn't fixed for our instance.

We are tested this completed fix — it really works. The load speed increase was up to 1000 times!


ulferts (changes requested):
As the failing spec highlights, the speedup comes at the cost of only rendering the projects links in the representer when links are embedded. 

But the desired behaviour is different. According to HAL, the links should always be rendered. The only part the server can decide to omit is embedded the linked resources. 

So in this case, we always need to render the links to the projects and can decide to not embed the projects. We do exactly this [already in the representer](https://github.com/opf/openproject/actions/runs/9640371655/job/26584048368?pr=15931).

So while improving the speed, the PR currently breaks the API. 

I am not sure how to proceed from here as I lack information on why the request is slow on your installation.

top4ek:
@foxweb @ulferts map fetches whole project entity from database, but we use only identifier and name while generating links. I believe to preserve current behavior we can try to pluck only this attributes, but we lose rails cache in this case. 
And, besides, I'm unsure if this approach will help, when OP instance has thousands of projects at once.

ulferts:
> map fetches whole project entity from database, but we use only identifier and name while generating links. I believe to preserve current behavior we can try to pluck only this attributes, ...

That is not the case. The same scope is currently used for both [rendering the links as well as the embedded objects](https://github.com/opf/openproject/pull/15923/files#diff-1c0b926e2b6a9209b4963a000552c591934d6ffe0b3cf7edcb63509ce241235dL48-L60). One could of course be more specific in that when not embedding the links, pluck is used. I still don't think that this will amount to a lot of speed difference.

foxweb:
> I still don't think that this will amount to a lot of speed difference.

We have 2096 projects, 1869 active. HTTP request to `/api/v3/time_entries/form` takes up the system over 60 seconds and returns 500 or 502 error.

If it completes successfully, JSON-response have 232000+ lines and over 8.3 Mb size, response time over 60 seconds.

Explanation: Rails iterate EACH type of activity and requests for EACH type ALL projects where the "Time and Costs" module is enabled.

photo_2024-06-25_12-33-30

ulferts:
@foxweb , how many different activity types do you have on your installation?

foxweb:
> @foxweb , how many different activity types do you have on your installation?

38

ulferts:
I tried to reproduce it with the values (2000 project, ~40 activity types) you provided, @foxweb but unfortunately did not manage to get similar results. The response took longer than I would have liked but with 2.5s was still far from what you experience.

A couple of questions:

* The sql requests are taking rather long on your machine (3s) whereas they are way quicker on mine (20ms). Those cannot by themselves result in the response time you report but might be a factor. Would it be possible for you to send an ANALYZE EXPLAIN for those?
* There are two sql requests issued (one from cache from line 56 of the time_entries_activity_representer.rb). After PR #15275 that line does not issue any SQL request any more and given that `embed_links` should be false in this case, the scope called in line 58 should not be called. Is the patch included in the instance that produced the output you shared? If it is:
* Are the full project resources included in the response? Only links should be but the size you reported might hint at more being in there than it is with my tests.

PR #15942, opened 2024-06-24
Fix UA logo conversion when styles are custom (OpenProject Light)

daggett-doofus:
https://community.openproject.org/work_packages/53987

Here is a reported issue from [Henriette Darge](https://community.openproject.org/users/39106). When choosing the OpenProject Light Theme in the administration/design section, the normal logo is still shown.

The reason is that when the theme 'OpenProject Light' is chosen, the dedicated logo is not set up properly. See the code reference: [OpenProject Light theme logo setup](https://github.com/opf/openproject/blob/09b3a68eb8a827fb246a15bb7546a601e8efb385/lib/open_project/custom_styles/color_themes.rb#L81).


ba1ash (approved):
LGTM

ba1ash (approved):

PR #15943, opened 2024-06-24
Display placeholder in case of no time entry activity

ulferts:
Display the placeholder text in case a time entry has no activity when listing it on the project overview page

image

https://community.openproject.org/wp/53200


aaron-contreras (approved):
Interesting how this tiny thing caused a browser crash

Nice find!

PR #15948, opened 2024-06-24
Document how to render validation errors inside a dialog with a form

judithroth:


HDinger (commented):
Hi @judithroth 

Looks good, short and on point 👍 I am wondering whether we want to show the code of `ComponentThatRendersAForm` as well to highlight that the Dialog::Body and Dialog::Footer components shall be used. Wdyt?

judithroth:
> Hi @judithroth
> 
> Looks good, short and on point 👍 I am wondering whether we want to show the code of `ComponentThatRendersAForm` as well to highlight that the Dialog::Body and Dialog::Footer components shall be used. Wdyt?

Usually I would like that idea very much :+1: 
It's hard to see in the PR diff, but actually the `ComponentThatRendersAForm` is from the paragraph above the new one I added. There it's used in a Dialog. So when reading the documentation top to button it would repeat - but it could be worth it for people skipping directly to this part :thinking: I will add to the text that this refers to the example above. Then it's clear and there is no duplication.

HDinger (approved):
Lgtm 👍 Thanks a lot

PR #15950, opened 2024-06-24
[#55899] refactor rename files command

Kharonus:
[#55899](https://community.openproject.org/wp/55899)

### WAT?

- use authentication strategies in rename files commands
- harmonize command interface
  - input data is now file id and new name
  - output is always a storage file in success case
- use shared tests to enforce common interface

### Trivia

- before nextcloud command used pathes and one drive used id and name
- new format is id + name for both, this required nextcloud query to make a preflight and postflight request
- changes to nextcloud command made some changes in sync service necessary ⚠️


modules/storages/spec/services/storages/nextcloud_group_folder_properties_sync_service_spec.rb
Kharonus, line 1: This spec needed some serious changes, only due to the fact, that one of the queries used, makes additional outbound requests. This indicates, that the unit of test might be too big.

If we'd stub the queries/commands response, instead of the outbound requests, this spec would be much more decoupled from this kind of changes.

modules/storages/spec/support/shared_examples_for_adapters/rename_file_command_examples.rb
```diff
@@ -0,0 +1,101 @@
...
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program; if not, write to the Free Software
+# Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+#
+# See COPYRIGHT and LICENSE files for more details.
+#++
+
+RSpec.shared_examples_for "rename_file_command: basic command setup" do
```
mereghost, line 31: This file is pure <3

modules/storages/spec/services/storages/nextcloud_group_folder_properties_sync_service_spec.rb
mereghost: Yes. This suite is messy as hell and probably should be redone.

mereghost (approved):
LGTM! Great work there.

PR #15951, opened 2024-06-24
Conform with RSpecRails/HaveHttpStatus cop

cbliard:
`have_http_status` matcher fails if the response object is a `Rack::MockResponse`, yet `RSpecRails/HaveHttpStatus` cop keeps complaining about it.

With a little hackery, disguise `Rack::MockResponse` into a `ActionDispatch::TestResponse` object to make it work and use `have_http_status` everywhere possible.

### DIFF
--- a/spec/support/have_http_status_with_rack_response.rb
+++ b/spec/support/have_http_status_with_rack_response.rb
@@ -32,10 +32,9 @@ module HaveHttpStatusWithRackResponse
       # `have_http_status` matcher would fail if the response object is a
       # `Rack::MockResponse`. Hack to disguise `Rack::MockResponse` into a
       # `ActionDispatch::TestResponse` object.
-      obj = ActionDispatch::Response.new(obj.status, obj.headers, obj.body).tap do |resp|
-        resp.request = ActionDispatch::Request.new({})
-      end
-      ::ActionDispatch::TestResponse.from_response(obj)
+      response = ActionDispatch::Response.new(obj.status, obj.headers, obj.body)
+      response.request = ActionDispatch::Request.new({})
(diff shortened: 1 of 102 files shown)


spec/support/have_http_status_with_rack_response.rb
```diff
@@ -0,0 +1,45 @@
...
+module HaveHttpStatusWithRackResponse
+  def as_test_response(obj)
+    if obj.is_a?(Rack::MockResponse)
+      # `have_http_status` matcher would fail if the response object is a
+      # `Rack::MockResponse`. Hack to disguise `Rack::MockResponse` into a
+      # `ActionDispatch::TestResponse` object.
+      obj = ActionDispatch::Response.new(obj.status, obj.headers, obj.body).tap do |resp|
+        resp.request = ActionDispatch::Request.new({})
+      end
+      ::ActionDispatch::TestResponse.from_response(obj)
```
toy: redefining `obj` is a bit confusing and  `tap` is superfluous here.
```suggestion
      response = ActionDispatch::Response.new(obj.status, obj.headers, obj.body)
      response.request = ActionDispatch::Request.new({})
      ::ActionDispatch::TestResponse.from_response(response)
```

toy (approved):
I only suggest to simplify the patch code

spec/support/have_http_status_with_rack_response.rb
cbliard: You're right. I kept too much stuff from the [super class code](https://github.com/rspec/rspec-rails/blob/591ef0b4ea153eaf8a4f1a285decba2ee601e55b/lib/rspec/rails/matchers/have_http_status.rb#L35-L55).
Thanks for the suggested change.

PR #15953, opened 2024-06-24
[#55362] Shared work packages in template projects cause several flaws

dombesz:
https://community.openproject.org/work_packages/55362

### DIFF
--- a/app/services/work_packages/copy_service.rb
+++ b/app/services/work_packages/copy_service.rb
@@ -41,15 +41,15 @@ class WorkPackages::CopyService
     self.contract_class = contract_class
   end
 
-  def call(send_notifications: nil, copy_attachments: true, **attributes)
+  def call(send_notifications: nil, copy_attachments: true, copy_share_members: true, **attributes)
     in_context(work_package, send_notifications:) do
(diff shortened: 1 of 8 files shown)


app/services/work_packages/copy_service.rb
```diff
@@ -111,4 +112,29 @@ def copy_work_package_attachments(copy)
...
+    return if role_ids.empty?
+
+    attributes = member
+                   .attributes.dup.except("id", "project_id", "entity_id", "created_at", "updated_at")
+                   .merge(role_ids:,
+                          project: target.project,
+                          entity: target,
+                          send_notifications:)
+
+    Members::CreateService
```
ulferts: Based on my quick comparison between the two, there shouldn't be a difference currently between using the `Members::CreateService` and the `Shares::CreateService` but semantically, the later would be 'more correct'. It might also be that the two diverge more later so I would use the `Shares::CreateService` to be on the safe side.

ulferts (approved):
Just nitpicking, @dombesz. Works like a charm.

app/services/work_packages/copy_service.rb
dombesz: Good point, thanks!

PR #15964, opened 2024-06-25
[#55898] Add uniqueness index to project_custom_field_project_mappings

judithroth:
to avoid duplicate rows if ActiveModel validations are bypassed.

https://community.openproject.org/work_packages/55898


akabiru (approved):
Amazing 👏🏾 Works brilliantly Judith! :shipit:

PR #15965, opened 2024-06-25
Inline create support dark theme.

Eric-Guo:
### Before:

iShot_2024-06-25_17 24 29

##### On dark:

iShot_2024-06-25_17 24 47

iShot_2024-06-25_17 25 05

##### On Light:

iShot_2024-06-25_17 33 42

iShot_2024-06-25_17 33 52

### DIFF
--- a/frontend/src/global_styles/content/work_packages/_table_inline_create.sass
+++ b/frontend/src/global_styles/content/work_packages/_table_inline_create.sass
@@ -14,10 +14,10 @@
     line-height: 1.6
 
 .wp-inline-create-row
-  background: #BEF3CA
+  background: var(--display-green-bgColor-muted)
 
   &:hover
-    background: darken(#BEF3CA, 5%) !important
+    background: var(--display-pine-bgColor-muted)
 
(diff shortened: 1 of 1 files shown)


HDinger (changes requested):
Hi @Eric-Guo 
thanks for the contribution! I think the used colors variables are however not ideal. The inline create row should be some kind of green whereas the `selection-bgColor` is blue. Further this is already used for the actual selection of rows. I think the inline create row should have a different color to differentiate them.

I'd suggest to use `--display-green-bgColor-muted` for the normal background color. And maybe `--display-pine-bgColor-muted` for the hover.

Eric-Guo:
@HDinger Thanks the suggestion, here is after change screen shot.

### After:

##### light

iShot_2024-06-25_21 47 57

iShot_2024-06-25_21 48 10

###### dark

iShot_2024-06-25_21 48 54

iShot_2024-06-25_21 49 09

HDinger (approved):
Thanks a lot! Lgtm 👍

PR #15971, opened 2024-06-25
[#55163] Build project-list sharing modal  

aaron-contreras:
## ToDo List

- [x] Add `ProjectQuery` as a shareable resource
- [x] Build the contracts to share `ProjectQuery`
- [x] Add a `ShareDialog` that will be added as an async dialog via Turbo
- [x] Introduce a `SharingStrategy` pattern that wraps all of the content of the methods `sharing_contract_scope`, `available_roles` and potentially more to not end up with big case statements in the `SharesController` as suggested by @ulferts in  https://github.com/opf/openproject/pull/15899#discussion_r1656863757
- [x] Change the Sharing modal to allow injecting components
  - [x] Inject a body component that adds the switch for making the query public or removing it (when allowed)
  - [x] Inject a custom empty state that shows a different message when there are no shares, but the query is marked as public
    - [ ] Nice to have: When in an empty state and toggling the query between public/private, the empty state text should be re-rendered without having to close and re-open the modal.
- [x] Currently the sharing controller is only accessible when we have the `share_work_packages` permission, we probably need to remove all permission checks from the shares controller and then handle permission checking with the `manageable?` method in the strategy and contracts.
- [x] `can_save?`, `can_rename?` checks for the project query index header component need to be changed to use the `visbile?` and `editable?` methods on the Query (@klaustopher)
- [x] Role names and descriptions should come from i18n files instead of directly from the DB (copy what we did for work package roles)
  - [x] Add role description translations
- [x] If you cannot view shares, hide the share button
- [x] Check all i18n strings
- [x] Show warning about project access

## Currently identified problems
- [x] Fix CSS top-layer issue with invite user form
- [x] Fix CSS issues with the modal  
  - [x] The `ActionMenu` Primer component uses an overlay to render its menu item list which ends up causing the container to span the entire width of the dialog for some reason
- [x] The modal currently closes after each action that is taken in the modal when sharing ProjectQueries 

## Documentation TODOs
- [x] Add `ShareDialogComponent` to Lookbook
- [x] Add documentation for the `keep-open-on-submit` (dccbdcaefb6183ce7bbf6f60af9b8f8a706dcbc0) functionality to https://qa.openproject-edge.com/lookbook/pages/patterns/dialogs

---

See: https://community.openproject.org/work_packages/55163


klaustopher:
When we add #15983 here, the modal already works in a super rough capacity

aaron-contreras:
Rebased onto https://github.com/opf/openproject/pull/15983 in order to get the fixes on there over on this branch

aaron-contreras:
Current modal rendering state

Known issues:
- [ ] Action Menu rendering inside the modal spans way too wide. I suspect this has something to do with the CSS top layer.

https://github.com/opf/openproject/assets/61627014/e53ba906-9152-416c-844e-203ea9c85e80

klaustopher:
https://github.com/opf/openproject/pull/15983 has been merged and the `generalize-sharing-modal` branch is based on current dev, so I am rebasing this branch bakc onto generalize sharing modal, please pull with. `git fetch --all && git reset --hard origin/implementation/55163-build-sharing-modal`

HDinger:
> Current modal rendering state
> 
> Known issues:
> 
>     * [ ]  Action Menu rendering inside the modal spans way too wide. I suspect this has something to do with the CSS top layer.
> 
> 
> Screen.Recording.2024-06-26.at.21.54.28.mov

This is actually an issue of Primer itself where the automatic width calulation of ActionMenus inside Dialogs do not work, see https://github.com/primer/view_components/issues/2926. As a wrokaround for now, we have to enforce the width of the ActionMenu manually. Unfortnately, we thus have to accept that the width is not always ideal.

app/contracts/shares/project_queries/base_extension.rb
```diff
@@ -1,8 +1,6 @@
-# frozen_string_literal: true
-
 # -- copyright
 # OpenProject is an open source project management software.
-# Copyright (C) 2010-2024 the OpenProject GmbH
+# Copyright (C) 2010-2023 the OpenProject GmbH
```
klaustopher: fix

app/components/projects/index_page_header_component.html.erb
```diff
@@ -21,7 +21,7 @@
         )
       end
 
-      if query.persisted? && OpenProject::FeatureDecisions.project_list_sharing_active?
+      if query.persisted? && OpenProject::FeatureDecisions.project_list_sharing_active? && can_view_shares?
```
klaustopher: I was thinking about this as well. But if the user cannot view the query, they will never reach the index page and thus this check does not really do anything.

aaron-contreras: I just realized after committing too. Probably superfluous so I'll remove it

app/components/projects/index_page_header_component.html.erb
```diff
@@ -21,14 +21,25 @@
         )
       end
 
+      if query.persisted? && OpenProject::FeatureDecisions.project_list_sharing_active?
```
toy: Are these all conditions needed and maybe better have a helper for this in ruby class?

app/components/shares/empty_state_component.rb
```diff
@@ -0,0 +1,58 @@
...
+
+      @strategy = strategy
+      @entity = strategy.entity
+    end
+
+    private
+
+    attr_reader :strategy, :entity
+
+    def blankslate_config # rubocop:disable Metrics/AbcSize
```
toy: Maybe better to have separate helpers for icon/heading_text/description_text? Otherwise at least better extract two methods with config for each case, so no need to build them.

app/components/shares/empty_state_component.rb
```diff
@@ -0,0 +1,58 @@
...
+#
+# You should have received a copy of the GNU General Public License
+# along with this program; if not, write to the Free Software
+# Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+#
+# See COPYRIGHT and LICENSE files for more details.
+#++
+
+module Shares
+  class EmptyStateComponent < ApplicationComponent # rubocop:disable OpenProject/AddPreviewForViewComponent
```
dombesz, line 30: When changing public state of the project query, the empty state does not update. 
<details><summary>(Click for demo)</summary>
<p>

empty_state_not_updated
</p>
</details> 

Feel free to address it separately as this is just a bug.

app/components/shares/modal_body_component.html.erb
```diff
@@ -1,11 +1,14 @@
...
-                                                   available_roles: @available_roles,
-                                                   sharing_manageable: @sharing_manageable,
-                                                   errors: @errors))
+      if strategy.custom_body_components?
+        strategy.additional_body_components.each do |component|
+          render(component.new(strategy: strategy, modal_body_container: modal_content))
+        end
+      end
+
+      modal_content.with_row(mt: (strategy.custom_body_components? ? 3 : nil)) do
```
toy: Any way to set `gap` instead of manually setting `margin-top` for all but first?
There is also a problem with `margin-top` for custom body components if it is added to all of them:
image

lookbook/docs/patterns/05-dialogs.md.erb
```diff
@@ -152,6 +148,23 @@ class TestController < ApplicationControler
 end
 ```
 
+### Dialogs that should stay open on form submission
+
+Because we normally want to close the dialog when a form is submitted successfully, we have implmented this as the default behaviour.  As described in the last section,
```
dombesz:
```suggestion
Because we normally want to close the dialog when a form is submitted successfully, we have impelmented this as the default behaviour.  As described in the last section,
```

dombesz (approved):
Very nice job @aaron-contreras and @klaustopher , just 2 minor things. Feel free to merge.

lookbook/docs/patterns/05-dialogs.md.erb
aaron-contreras: implemented***

hard to spell 😆

aaron-contreras:
```suggestion
Because we normally want to close the dialog when a form is submitted successfully, we have implemented this as the default behaviour.  As described in the last section,
```

app/components/shares/modal_body_component.rb
```diff
@@ -106,15 +86,15 @@ def type_filter_options
       end
     end
 
-    def type_filter_option_active?(option)
+    def type_filter_option_active?(option) # rubocop:disable Metrics/AbcSize
       principal_type_filter_value = current_filter_value(params[:filters], "principal_type")
       project_member_filter_value = current_filter_value(params[:filters], "also_project_member")
 
-      return false if principal_type_filter_value.nil? || project_member_filter_value.nil?
+      return false if principal_type_filter_value.nil? || (project_scoped_entity? && project_member_filter_value.nil?)
```
toy:
```suggestion
      return false if principal_type_filter_value.nil?
      return false if project_scoped_entity? && project_member_filter_value.nil?
```

app/components/shares/project_queries/empty_state_component.rb
```diff
@@ -0,0 +1,64 @@
...
+        @strategy = strategy
+        @entity = strategy.entity
+      end
+
+      private
+
+      attr_reader :strategy, :entity
+
+      def blankslate_config # rubocop:disable Metrics/AbcSize
+        @blankslate_config ||= {}.tap do |config|
```
toy: Same suggestion about separate methods or separate configs to not build them explicitly.

app/components/shares/project_queries/project_access_warning_component.html.erb
```diff
@@ -0,0 +1,3 @@
+<%- if query_is_public? || query_is_shared? %>
+  <%= container.with_row(mt: 3) { render(Primer::Alpha::Banner.new(icon: :info)) { I18n.t('sharing.project_queries.access_warning') } }%>
```
toy: margin-top will be there even if it is the first box in flex container

app/components/shares/project_queries/public_flag_component.html.erb
```diff
@@ -0,0 +1,14 @@
...
+container.with_row(mt: 3) do
+  render(Primer::Forms::ToggleSwitchForm.new(
+        name: "publish_project_query",
+        label: t('sharing.project_queries.public_flag.label', instance_name: Setting.app_title),
+        caption: t('sharing.project_queries.public_flag.caption'),
+        src: toggle_public_flag_link,
+        csrf_token: form_authenticity_token,
+        status_label_position: :start,
+        checked:,
+        enabled: can_publish?
```
toy: It should be either renaming `checked` to `published?`:
```suggestion
        checked: published?,
        enabled: can_publish?
```
or renaming `can_publish?` to `enabled`
```suggestion
        checked:,
        enabled:
```

app/components/shares/share_dialog_component.rb
```diff
@@ -0,0 +1,49 @@
...
+  class ShareDialogComponent < ApplicationComponent
+    include ApplicationHelper
+    include OpTurbo::Streamable
+    include OpPrimer::ComponentHelpers
+
+    def initialize(shares:, strategy:, errors:)
+      super
+
+      @strategy = strategy
+      @shares = shares
```
toy: Just for same order
```suggestion
      @shares = shares
      @strategy = strategy
```

app/components/shares/share_row_component.rb
```diff
@@ -34,17 +34,14 @@ class ShareRowComponent < ApplicationComponent # rubocop:disable OpenProject/Add
...
 
-    def initialize(share:,
-                   available_roles:,
-                   sharing_manageable:,
-                   container: nil)
+    def initialize(share:, strategy:, container: nil)
       super
 
+      @strategy = strategy
       @share = share
```
toy:
```suggestion
      @share = share
      @strategy = strategy
```

app/contracts/shares/project_queries/delete_contract.rb
```diff
@@ -0,0 +1,37 @@
...
+#
+# See COPYRIGHT and LICENSE files for more details.
+# ++
+
+module Shares
+  module ProjectQueries
+    class DeleteContract < Shares::DeleteContract
+      # DeleteContract has its own permission check and does not care about the role class,
+      # so we do not need to include the BaseExtension here.
+      delete_permission -> { model.entity.editable? }
```
toy, line 34: `user_allowed_to_manage?` can be used here

app/controllers/projects/queries_controller.rb
```diff
@@ -71,20 +71,15 @@ def update
...
 
-  def publish
-    call = Queries::Projects::ProjectQueries::PublishService
-             .new(user: current_user, model: @query)
-             .call(public: true)
-
-    render_result(call, success_i18n_key: "lists.publish.success", error_i18n_key: "lists.publish.failure")
-  end
+  def toggle_public
+    to_be_public = !@query.public?
```
toy, line 75: I think it is better to explicitly specify either as a separate action or in params whether published or unpublished state is desired, so that user seeing the state changes it to desired, not change it to opposite of what is in db and can be different from what user sees

app/controllers/shares_controller.rb
```diff
@@ -129,6 +130,20 @@ def bulk_destroy
 
   private
 
+  attr_reader :sharing_strategy
+
+  def check_if_viewable
+    return if sharing_strategy.viewable? || sharing_strategy.manageable?
```
toy, line 136: Should or-ing with `manageable?` be part of strategy instead?

app/menus/projects/menu.rb
```diff
@@ -104,7 +104,7 @@ def public_filters
 
     def my_filters
       persisted_filters
-        .select { |query| query.user == current_user }
+        .select { |query| query.user == current_user && !query.public? }
```
toy: Maybe better:
```suggestion
        .reject(&:public?)
        .select { |query| query.user == current_user }
```
And same for `shared_filters`.

It is probably not yet worth somehow combining filtering for three methods to not repeat calls.

app/models/queries/members/filters/entity_type_filter.rb
```diff
@@ -32,7 +32,10 @@ def type
   end
 
   def allowed_values
-    [[WorkPackage.name, WorkPackage.name]]
+    [
+      [WorkPackage.name, WorkPackage.name],
+      [ProjectQuery.name, ProjectQuery.name]
```
toy, line 37: Looks like double vision, I suggest either generating repetition or trying without repetition, as usages seem to call either first or last:
```suggestion
      [WorkPackage.name],
      [ProjectQuery.name]
```

app/models/sharing_strategies/base_strategy.rb
```diff
@@ -0,0 +1,83 @@
...
+      @entity = entity
+      @user = user
+    end
+
+    def available_roles
+      # format: [{ label: "Role name", value: 42, description: "Role description", default: true }]
+      raise NotImplementedError, "Override in a subclass and return an array of roles that should be displayed"
+    end
+
+    def vieable?
```
toy: Is it:
```suggestion
    def viewable?
```

app/services/authorization/user_permissible_service.rb
```diff
@@ -112,7 +112,9 @@ def allowed_in_single_standalone_entity?(permissions, entity)
       return false if entity.nil?
       return true if admin_and_all_granted_to_admin?(permissions)
 
-      cached_permissions(entity).intersect?(permissions)
+      permission_names = permissions.map { |perm| perm.name.to_sym }
+
+      cached_permissions(entity).intersect?(permission_names)
```
toy, line 117: Why was it working before?

app/views/shares/dialog.turbo_stream.erb
```diff
@@ -0,0 +1,4 @@
+<%= turbo_stream.dialog do
+  render(Shares::ShareDialogComponent.new(strategy: @sharing_strategy, shares: @shares, errors: @errors))
+end
+%>
```
toy:
```suggestion
end %>
```

toy (commented):
Mostly nitpicks, though no specs added

app/components/projects/index_page_header_component.html.erb
aaron-contreras: I would say all are needed and yes, it's better to have them in the ruby component

app/components/shares/modal_body_component.html.erb
aaron-contreras: I like the idea of using gap for spacing between rows but specifically for the project query sharing modal, this spacing is desired based on the Figma design:

image

Based on it, I think we can invert this and specifically in the project query sharing dialog add some bottom margin and space the rows with gap.

aaron-contreras: See e964e00c847a

app/components/projects/index_page_header_component.html.erb
toy: I meant if any more are needed, like `query.editable?`

app/components/shares/empty_state_component.rb
klaustopher: Yeah, we mentioned this above in the ticket. We would need to copy a bit of the share logic into the ProjectQueries controller (loading shares, rendering a turbo response, etc to render the modal.

Or we need to define the empty state as a seperate turbo frame and exchange that one from the ProjectQueries controller. That might actually work, will take a look

app/components/projects/index_page_header_component.html.erb
klaustopher: Yes, when you can see the query and it is persisted you can access the modal to see whom it is shared with

app/components/shares/project_queries/project_access_warning_component.html.erb
klaustopher: As this component will not be used anywhere outside of its intended scope that's not really an issue, but @aaron-contreras fixed this with the `gap` 👍

app/contracts/shares/project_queries/delete_contract.rb
klaustopher: Yeah, could be as well. This basically predates the introduction of that

app/controllers/projects/queries_controller.rb
klaustopher: The intention of a toggle endpoint is to switch something back and forth. The `ToggleSwitchForm` that we are using does not allow for seperate endpoints or a param

app/controllers/shares_controller.rb
klaustopher: You mean that `viewable?` here should also return true when it is "just" manageable? Let me think about it

app/menus/projects/menu.rb
klaustopher: I do not agree with this one. With `select` 3 times we are building and throwing away 3 arrays of AR objects. The checks are small enough and easy to understand what is filtered here, so I don't see a benefit of splitting those up.

app/models/queries/members/filters/entity_type_filter.rb
klaustopher: Orienting on other places where we use this as well

app/services/authorization/user_permissible_service.rb
klaustopher: For resoruces like project queries it wasn't. I will add tests for this in #16032

app/contracts/shares/project_queries/delete_contract.rb
klaustopher: Ah no, we're not using the mixin here. So this is correct as it is

app/controllers/shares_controller.rb
klaustopher: Thought about it again. We are trying to convey here that those two actions are accessible when the shares can be viewed or managed. So the code is correctly showing the intentions here.

PR #15972, opened 2024-06-25
Use jemalloc to prevent memory bloating

top4ek:
Use dynamically linked [jemalloc](https://jemalloc.net/) to prevent memory bloating in long living instances.


oliverguenther (changes requested):
Hi @top4ek, thanks for your contribution and interest in improvements. We build multiple architectures for our dockerfiles, so this PR will fail them: https://github.com/opf/openproject/blob/dev/.github/workflows/docker.yml#L105-L127

/cc @opf/ops

top4ek:
Oh, thanks, will look into this. 
Quick googling gives some problems with arm and ppc platforms so can we use it just for x86? 
I can try to poke it with RPi4 later and build image accordingly for arm, but not ppc. I'm not sure we need to use unofficial jemalloc enabled ruby images.

top4ek:
Added support for other architectures.  Using **USE_JEMALLOC=false** enviromnent by default. 
I was unable to test it on my RPi4(low memory, it just hangs on image building). Also there's opened issues on arm and ppc architectures in jemalloc repo. 
x86 is battle tested already for couple of weeks on OP and it's reduced memory usage at least twice. (My previous experience at different project with jemalloc allowed us to use 4 times more sidekiq instances on the same hardware limitations). 
Other architectures MUST be tested, but I don't have access to them.

docker/prod/setup/preinstall-common.sh
```diff
@@ -52,6 +52,18 @@ apt-get install -yq --no-install-recommends \
...
+if [ "$USE_JEMALLOC" = "true" ]; then
+	apt-get install -yq libjemalloc2
+	if [ "$ARCHITECTURE" = "x64" ]; then
+		LIB_ARCH="x86_64"
+	elif [ "$ARCHITECTURE" = "arm64" ]; then
+		LIB_ARCH="aarch64"
+	elif [ "$ARCHITECTURE" = "ppc64le" ]; then
+		LIB_ARCH="powerpc64le"
+	fi
+	export LD_PRELOAD=/usr/lib/${LIB_ARCH}-linux-gnu/libjemalloc.so.2
```
crohr: @top4ek aren't we supposed to also set that env variable when launching a container from this image?

top4ek: Hmm, looks like you're right. Just looked into fresh instance with this patch and it doesn't work. `export` is not set on build stage.

crohr: I think for now there would need to be an env variable basically saying "launch with jemalloc" (can reuse `USE_JEMALLOC` for that) that sets the `LD_PRELOAD` path in the entrypoint.

top4ek: I hoped to bake LD_PRELOAD straight into the image on build stage to make it more "systemwide". Quick googling said that i can't do that inside dockerfile's RUN.
Got some ideas, will try them.

machisuji:
I've tested this on ppc64 and double-checked via 

```
MALLOC_CONF=stats_print:true ruby -e "exit"
# prints
#   --- End jemalloc statistics ---
# at the end
```

to see if ruby is indeed using JEMALLOC which it is. So that's good.

So this PR is fine for x86 and ppc64 at least. And I reckon arm64 will be fine too. This can go ahead if we add the `export LD_PRELOAD=...` to the web (`docker/prod/web`) and worker (`docker/prod/worker`) scripts. Could you please amend that @top4ek?

machisuji (changes requested):
@top4ek Thanks for the PR. I would actually set the default for `USE_JEMALLOC` to true.

The only thing that's missing is the `LD_PRELOAD` export in the web and worker scripts as described in my other comment. I would check for `USE_JEMALLOC` there too so that we can still choose not to use it.

top4ek:
> `LD_PRELOAD` export in the web and worker scripts

The problem is that `LD_PRELOAD` is architecture dependent and it's can't be exported while building in RUN command. `/etc/environment` and `/etc/profile(.d)` don't cause any effect as per my tests. 
I beleive that web and worker scripts are fine, but I'm playing with docker ENTRYPOINT scripts ATM to set variable on start via additionally generated script on preinstall stage.  

> USE_JEMALLOC to true

Not sure if it's safe to do without tests as because of https://github.com/jemalloc/jemalloc/issues/2358

docker/prod/setup/preinstall-common.sh
crohr: You can use the existing entrypoint.sh file, and simply export the relevant variables if `USE_JEMALLOC` env variable is set to `true` by the user?

top4ek: Sure, but `entrypoint.sh` is for only all-in-one target, so I'm going to create additional entrypoint script for slim target and don't want to copy-paste same code in both scripts.

top4ek:
Added entrypoint for slim container. Show jemalloc status at admin/info page.

docker/prod/Dockerfile
```diff
@@ -5,6 +5,8 @@ LABEL maintainer="operations@openproject.com"
 ARG BUNDLER_VERSION="2.5.13"
 ARG NODE_VERSION="20.9.0"
 ARG BIM_SUPPORT=true
+ARG USE_JEMALLOC=false
```
machisuji: I would suggest renaming this to something like `JEMALLOC_SUPPORT` which defaults to true, whilst keeping `USE_JEMALLOC` with `false` as the default.

This way we build the images with JEMALLOC support by default without forcing anyone to use it.
But if they do want to, they can simply enable it without having to build their own image.

One could also add the build arg in the docker workflow which builds the images and set it true there.
But I would find it less of a hassle and potentially less confusing if it's just supported (but not used) by default.

machisuji (changes requested):
Hey @top4ek, I think this looks good and works fine. I just have one suggestion. Could you please have a look?

docker/prod/Dockerfile
top4ek: So in that case we even don't need JEMALLOCK_SUPPORT and build images with that lib for all platforms.
The main problem with SUPPORT var is when it's set to false, but runtime set to true, we will get lot's of warnings because of missing .so file in LD_PRELOAD.

machisuji: You're right. Just drop the JEMALLOC_SUPPORT var then and simply always install it, but only use it when USE_JEMALLOC is true.

machisuji (approved):
👍

machisuji:
Thanks for your contribution @top4ek! Merging it (into dev) now. That means it should become available officially with OpenProject 14.6 which is currently scheduled to be released in about two weeks.

PR #15979, opened 2024-06-26
Introduces a job that runs the AMPF sync on a per storage basis

mereghost:
## Why?

We run a long job with that syncs all the `Storages::Storage` that have automatically managed folders enabled with its provider.

This job is triggered basically all the time and at a schedule. Project Created, run it. Member added somewhere, run it. Sneezed in a funny manner, run it.

This leads to a lot of wasted resources as we are doing work that probably isn't required.

## What?

This PR aims to reduce the processing time and wasted resources by splitting the work on a per storage basis.

This would still avoid some race conditions while giving us slightly more flexibility when dealing with the synchronization process.

## How?

First all the logic from `Storages::ManageIntegrationJob` was moved to `Storages::AutomaticallyManagedStorageSyncJob`.

This new job works as mentioned above: on a single `Storages::Storage` and has some concurrency controls over it allowing at most 2 jobs for each storage (1 running/1 on queue).

This new job was then hooked to the `Member`s, `Project`s and `ProjectStorage`s events, so that it queues a job only for the relevant `Storages::Storage`, if any.

The original job then will only need to queue this new one for each storage to keep its original functionality and benefit from the granularity.

### DIFF
--- a/modules/storages/app/workers/storages/automatically_managed_storage_sync_job.rb
+++ b/modules/storages/app/workers/storages/automatically_managed_storage_sync_job.rb
@@ -29,41 +29,45 @@
 #++
 
 module Storages
-  class ManageStorageIntegrationsJob < ApplicationJob
+  class AutomaticallyManagedStorageSyncJob < ApplicationJob
     include GoodJob::ActiveJobExtensions::Concurrency
     extend ::DebounceableJob
+
+    queue_with_priority :above_normal
 
     good_job_control_concurrency_with(
       total_limit: 2,
       enqueue_limit: 1,
-      perform_limit: 1
+      perform_limit: 1,
+      key: -> { "StorageSyncJob-#{arguments.last.short_provider_type}-#{arguments.last.id}" }
     )
 
-    retry_on GoodJob::ActiveJobExtensions::Concurrency::ConcurrencyExceededError,
-             wait: 5,
-             attempts: 20
+    retry_on GoodJob::ActiveJobExtensions::Concurrency::ConcurrencyExceededError, wait: 5, attempts: 10
 
-    KEY = :manage_nextcloud_integration_job_debounce_happened_at
-    CRON_JOB_KEY = :"Storages::ManageStorageIntegrationsJob"
-
-    queue_with_priority :above_normal
-
-    class << self
-      def disable_cron_job_if_needed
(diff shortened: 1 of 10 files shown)


modules/storages/lib/open_project/storages/engine.rb
```diff
@@ -55,14 +55,25 @@ def self.permissions
...
-          OpenProject::Events::PROJECT_RENAMED,
-          OpenProject::Events::PROJECT_ARCHIVED,
-          OpenProject::Events::PROJECT_UNARCHIVED
+          OpenProject::Events::MEMBER_DESTROYED
         ].each do |event|
-          OpenProject::Notifications.subscribe(event) do |_payload|
-            ::Storages::ManageStorageIntegrationsJob.debounce
+          OpenProject::Notifications.subscribe(event) do |payload|
+            ::Storages::Storage.joins(project_storages: :project)
+                             .where(project_storages: { project_id: payload[:member].project_id }).find_each do |storage|
```
mereghost: I probably should extract this into a scope. 🤔

ba1ash: Yes, it makes sense :+1:

modules/storages/app/workers/storages/automatically_managed_storage_sync_job.rb
```diff
@@ -0,0 +1,55 @@
+#-- copyright
+#++
+
+module Storages
+  class AutomaticallyManagedStorageSyncJob < ApplicationJob
+    include GoodJob::ActiveJobExtensions::Concurrency
+
+    SINGLE_THREAD_DEBOUNCE_TIME = 4.seconds
+    class << self
+      def debounce(storage)
```
ba1ash: Let's extract debounce logic into a module and include it in both jobs? One thing that seems to be different is key construction and it can be defined within job itself. Wdyt?

ba1ash (commented):
I think you forgot to specify the queue for newly created job.
 `queue_with_priority :above_normal`

modules/storages/app/workers/storages/automatically_managed_storage_sync_job.rb
mereghost: Hmmm that might be useful.

ba1ash (approved):
Looks good to me. Well done!

PR #15983, opened 2024-06-26
Move ProjectQuery model to the main namespace

klaustopher:
With the current place where the model is, we break all sorts of automations that Rails bring us (i.e. automatically building proper routes with `url_for(@project_query)`. This will become really problematic when implementing sharing #15971.

This PR moves the class to the main namespace without any namespacing so it can be accessed like any other model


aaron-contreras (approved):
This looks good to me 👍 Did a short test with the project list sharing branch and path resolution is working as expected

I think Jens's should weight in too as you requested his review specifically

klaustopher:
@ulferts has pointed out that we should probably specify how we want to handle persisted queries in the future: https://community.openproject.org/projects/openproject/work_packages/55991

PR #15990, opened 2024-06-28
[#55960] Add the enabled in projects tab in storage edit view

judithroth:
https://community.openproject.org/work_packages/55960

### DIFF
--- a/app/components/settings/project_custom_fields/edit_form_header_component.html.erb
+++ b/app/components/settings/project_custom_fields/edit_form_header_component.html.erb
@@ -43,7 +43,7 @@ See COPYRIGHT and LICENSE files for more details.
         selected: project_custom_field_project_mappings_selected?,
         href: project_mappings_admin_settings_project_custom_field_path(@custom_field)
       ) do |tab|
-        tab.with_text { I18n.t("projects.settings.project_custom_fields.project_mappings.header") }
+        tab.with_text { t(:label_project_mappings) }
       end
     end
   end
--- a/modules/storages/app/components/storages/admin/edit_form_header_component.html.erb
+++ b/modules/storages/app/components/storages/admin/edit_form_header_component.html.erb
@@ -43,5 +43,20 @@ See COPYRIGHT and LICENSE files for more details.
       button.with_leading_visual_icon(icon: :trash)
       I18n.t("button_delete")
     end
+    header.with_tab_nav(label: nil, test_selector: :project_attribute_detail_header) do |tab_nav|
+      tab_nav.with_tab(
+        selected: tab_selected?(:edit),
+        href: edit_admin_settings_storage_path(@storage)
+      ) do |tab|
+        tab.with_text { t(:label_details) }
+      end
+
+      tab_nav.with_tab(
+        selected: tab_selected?(:project_mappings),
+        href: project_mappings_admin_settings_storage_path(@storage)
+      ) do |tab|
+        tab.with_text { t(:label_project_mappings) }
+      end
+    end
   end
 %>
--- a/modules/storages/app/components/storages/admin/edit_form_header_component.html.erb
+++ b/modules/storages/app/components/storages/admin/edit_form_header_component.html.erb
@@ -26,31 +26,22 @@ Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
 See COPYRIGHT and LICENSE files for more details.
 
 ++#%>
-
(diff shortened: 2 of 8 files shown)


app/components/settings/project_custom_fields/edit_form_header_component.html.erb
```diff
@@ -43,7 +43,7 @@ See COPYRIGHT and LICENSE files for more details.
         selected: project_custom_field_project_mappings_selected?,
         href: project_mappings_admin_settings_project_custom_field_path(@custom_field)
       ) do |tab|
-        tab.with_text { I18n.t("projects.settings.project_custom_fields.project_mappings.header") }
+        tab.with_text { t(:label_project_mappings) }
```
judithroth, line 46: I wasn't sure if it was too early to extract that to a common label, but it seems like this is very generic and could be used at multiple locations in the future as well. (Probably all of https://community.openproject.org/projects/openproject/work_packages/48317/activity)

modules/storages/app/components/storages/admin/edit_form_header_component.html.erb
```diff
@@ -43,5 +43,20 @@ See COPYRIGHT and LICENSE files for more details.
       button.with_leading_visual_icon(icon: :trash)
       I18n.t("button_delete")
     end
+    header.with_tab_nav(label: nil, test_selector: :project_attribute_detail_header) do |tab_nav|
```
judithroth: In the design, there's some nice spacing between the Page title and the tab navs. As far as I can tell, we don't do that at other places (yet?), so I was unsure if I should add special spacing here.

app/components/settings/project_custom_fields/edit_form_header_component.html.erb
akabiru: Good point, since it would be applied in various places makes sense to extract. I should mention that we recently had dev discussions on reducing the generic "label_" translations in favor of more feature specific nesting- such as we had before- but in this case, I think it makes sense as it's also very short 👍🏾

modules/storages/app/components/storages/admin/edit_form_header_component.html.erb
akabiru: Good eye, it's likely that it was just missed in previous designs- I would suggest adding the spacing and raising a bug/code maintenance to track the other pages so that it can be updated to the same spec. Perhaps one for the Frontend channel?

akabiru (commented):
Looks great to me @judithroth - such a nice start. 🙌🏾 

---

I will reach out to chat about feature flags, I'm having second thoughts about "master branch" approach as this is likely to be a long running feature. 🤔

modules/storages/app/components/storages/admin/edit_form_header_component.html.erb
judithroth: I added an open point for that: https://community.openproject.org/projects/openproject/work_packages/56057/activity 
I remember people telling me they preferred that to discussions in elements, but to be honest it feels a bit exaggerated afterwards now :sweat_smile: 

However, I assume we can probably keep it as is or the styles will be adapted globally and I can add that afterwards. So I think this topic this should not block this PR.

judithroth: It turned out that people want the more spacious spacing and I created a separate ticket for that: https://community.openproject.org/projects/openproject/work_packages/56060

app/components/settings/project_custom_fields/edit_form_header_component.html.erb
judithroth: I can totally relate to not wanting too many "label_" translations. Things like this can go wild. But I checked again and found not other place that seemed to fit. So I also think we should keep it there for now.

modules/storages/app/components/storages/admin/edit_form_header_component.html.erb
```diff
@@ -0,0 +1,64 @@
...
+    header.with_action_button(scheme: :danger,
+                              mobile_icon: :trash,
+                              mobile_label: I18n.t("button_delete"),
+                              type: :submit,
+                              aria: { label: I18n.t("storages.label_delete_storage") },
+                              test_selector: "storage-delete-button") do |button|
+      button.with_leading_visual_icon(icon: :trash)
+      I18n.t("button_delete")
+    end
+    if OpenProject::FeatureDecisions.enable_storage_for_multiple_projects_active?
```
akabiru, line 46: 👍🏾

akabiru (approved):
Works great! 👍🏾

PR #15992, opened 2024-06-28
Feature/51672 starring favorite project lists

toy:
[OP#51672](https://community.openproject.org/projects/openproject/work_packages/51672)

### DIFF
--- a/app/menus/projects/menu.rb
+++ b/app/menus/projects/menu.rb
@@ -120,14 +120,14 @@ module Projects
     end
 
     def persisted_filters
-      @persisted_filters = ::ProjectQuery
+      @persisted_filters ||= ::ProjectQuery
         .visible(current_user)
         .with_favored_by_user(current_user)
         .order(favored: :desc, name: :asc)
     end
 
     def favored_ids
-      @favored_ids = persisted_filters.select(&:favored).to_set(&:id)
+      @favored_ids ||= persisted_filters.select(&:favored).to_set(&:id)
     end
 
     def modification_params?
--- a/app/menus/projects/menu.rb
+++ b/app/menus/projects/menu.rb
@@ -121,7 +121,7 @@ module Projects
 
     def persisted_filters
       @persisted_filters = ::ProjectQuery.visible(current_user).order(:name)
-        .sort_by.with_index { |query, i| [query.id.in?(favored_ids) ? 0 : 1, i] }
+        .partition { |query| query.id.in?(favored_ids) }.flatten
     end
 
     def favored_ids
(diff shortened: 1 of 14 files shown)


app/menus/projects/menu.rb
```diff
@@ -117,6 +121,11 @@ def shared_filters
 
     def persisted_filters
       @persisted_filters = ::ProjectQuery.visible(current_user).order(:name)
+        .partition { |query| query.id.in?(favored_ids) }.flatten
```
klaustopher: We are partitioning the list and then directly flattening it again? Is this to move all favored queries to the front? If that's the case, than that's a really clever solution, but in a few weeks we will have forgotten what it does and why it works.

What do you think about adding a scope in the `Acts::Favoritable` module: 

```ruby
scope :with_favored_by_user, ->(user) {
  favorite = ::Favorite.arel_table

  join = arel_table
          .join(favorite, Arel::Nodes::OuterJoin)
          .on(
            favorite[:favored_type].eq(base_class.name),
            favorite[:favored_id].eq(arel_table[:id]),
            favorite[:user_id].eq(user.id)
          ).join_sources

  select(arel_table[Arel.star], favorite[:id].eq(nil).as("favored")).joins(join)
}
```

This will add a virtual attribute to the `ProjectQuery` that is loaded with that scope and we can use that for sorting:

```ruby
ProjectQuery.with_favored_by_user(current_user).order("favored DESC", "name ASC")
```

This returns favored queries first (PG is a bit dumb that it sorts `TRUE` as last fields).

The scope has an obvious name and additionally we don't have to do an additional query for the favorited IDs.

klaustopher (commented):
Implementation looks fine, except for the one mentioned solution that is pretty clever, but a bit unobvious in how and why it works. I have left a suggestion inline by moving the sorting to SQL and eliminating some extra queries.

app/menus/projects/menu.rb
toy: I tried the SQL way at first, and it felt more complex (also I didn't try using arel). Then I used sorting (see 9679ff8fa4c4cf56f0be12f0fcd0f6d63e627a74) and then switched to partition.

I agree that better and faster to rely on database, just the construction of outer join is not easily readable. Let me check how it looks in the end.

klaustopher: The good thing is that it is "hidden" away in the concerns code and we can test the implementation/query generation once and then when we use it can just rely on it

toy: The code is covered, but acts as favorable is tested only implicitly, so I'll note to create tests for it

klaustopher (approved):

PR #15993, opened 2024-06-28
migration to remove ProjectQuery namespace for entity_type of Member

toy:
Migration is needed after changes in #15983


klaustopher:
Up until now there was no way to create a membership on a project query for real users out there, so that migration will not do anything for anybody outside of our little circle of developers `rails runner "Member.where(entity_type: 'Queries::Projects::ProjectQuery').update_all(entity_type: 'ProjectQuery')` for everybody who has created some should just do the job

PR #16000, opened 2024-07-01
Display shared filters

klaustopher:
This PR displays shared filters in the menu. Also, I optimized loading the persisted filters only once from the database and then working on a cached list instead of loading the list multiple times.

---

Part of https://community.openproject.org/projects/openproject/work_packages/55162


ulferts (approved):
Change itself is good. Tests are failing legitimately. Speaking of tests, it would be good to have one on the feature level. Don't know if you plan to do it now, @klaustopher or in a later PR.

PR #16001, opened 2024-07-01
[#56027] disable token caching for one drive validation

Kharonus:
[#56027](https://community.openproject.org/work_packages/56027)

### WAT?

- strategies now have an additional data attribute `use_cache`
  - this is per default true
  - it can be switched off with calling `with_cache(:boolean)` on the strategy
- OneDrive validator now uses an uncached strategy


mereghost (approved):
Looks good.

PR #16007, opened 2024-07-01
[#55939] OneDrive/SharePoint storage with AMPF can be added as manual folder

judithroth:
https://community.openproject.org/work_packages/55939

Because managing of permissions is complex on OneDrive, a storage with automatically managed access and folders can not host projects with manually managed permissions.

---

Please notice that an existing file had to be renamed due to rubocop complaining.

### DIFF
--- a/modules/storages/app/models/storages/project_storage.rb
+++ b/modules/storages/app/models/storages/project_storage.rb
@@ -57,12 +57,8 @@ module Storages
         .where(storage: Storage.automatic_management_enabled)
     end
 
-    def automatic_management_possible?
-      storage.present? && storage.automatic_management_enabled?
-    end
-
-    def manual_management_possible?
-      storage&.manual_management_possible?
+    def project_folder_mode_possible?(project_folder_mode)
+      storage&.available_project_folder_modes&.include?(project_folder_mode)
     end
 
     def managed_project_folder_path
--- a/modules/storages/app/models/storages/project_storage.rb
+++ b/modules/storages/app/models/storages/project_storage.rb
@@ -58,7 +58,7 @@ module Storages
     end
 
     def project_folder_mode_possible?(project_folder_mode)
-      storage&.available_project_folder_modes&.include?(project_folder_mode)
+      storage.present? && storage.available_project_folder_modes&.include?(project_folder_mode)
     end
 
     def managed_project_folder_path
--- a/modules/storages/app/models/storages/project_storage.rb
+++ b/modules/storages/app/models/storages/project_storage.rb
@@ -61,6 +61,10 @@ module Storages
       storage.present? && storage.automatic_management_enabled?
     end
 
+    def manual_management_possible?
+      storage&.manual_management_possible?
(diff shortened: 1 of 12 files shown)


spec/support/oauth_connections_helpers.rb
```diff
@@ -0,0 +1,30 @@
+module OAuthConnectionsHelpers
+  def mock_one_drive_authorization_validation(with: {})
```
judithroth, line 2: If someone has a better suggestion on how to name this method please let me know :slightly_smiling_face:

modules/storages/app/models/storages/storage.rb
```diff
@@ -149,6 +149,14 @@ def automatic_management_enabled=(value)
 
     alias automatic_management_enabled automatically_managed
 
+    def available_project_folder_modes
```
Kharonus, line 152: 🟡 small improvement suggestion:

Do not code a "default" implementation here, instead use `raise Errors::SubclassResponsibility`  and implement Nextcloud logic on the nextcloud storage. it fits more the understanding of the domain, IMHO.

modules/storages/app/models/storages/project_storage.rb
```diff
@@ -57,8 +57,8 @@ class ProjectStorage < ApplicationRecord
         .where(storage: Storage.automatic_management_enabled)
     end
 
-    def automatic_management_possible?
-      storage.present? && storage.automatic_management_enabled?
+    def project_folder_mode_possible?(project_folder_mode)
+      storage&.available_project_folder_modes&.include?(project_folder_mode)
```
Kharonus: 🔴 you need to return this check to `storage.present?` because if the storage is nil, this line now  returns a non boolean value:

```ruby
b = nil
b&.include?(1)
=> nil
```

but it should return `false`.

modules/storages/spec/contracts/storages/project_storages/base_contract_spec.rb
Kharonus, line 1: 🟢 nice update of the contract specs

spec/support/oauth_connections_helpers.rb
Kharonus: eventually this will vanish for a VCR cassette, so I wouldn't bother too much. Name is fine.

Kharonus (changes requested):

spec/support/oauth_connections_helpers.rb
judithroth: Actuall I tried using a VCR cassette first (`it "shows no option for manually managed permissions", vcr: "auth/one_drive/user_token" do`) but had some issues with it, because it's a feature spec:
```
            An HTTP request has been made that VCR does not know how to handle:
              GET http://172.18.0.9:3000/__identify__
```
(The IP is from the docker container and probably needs to be ignored by VCR - I didn't dive into this).

Kharonus (approved):
LGTM

PR #16017, opened 2024-07-02
[55997] Add duration_format setting with hours_only by default

cbliard:
https://community.openproject.org/wp/55997

The `duration_format` setting can take two values: "hours_only" or "days_and_hours". The default value is "hours_only".

The `duration_format` setting is used to determine how durations are displayed in the frontend and in some exports.

It can be changed in Administration > Calendar and Dates > Working days and hours.

image

### DIFF
--- a/lib/chronic_duration.rb
+++ b/lib/chronic_duration.rb
@@ -102,7 +102,10 @@ module ChronicDuration
     month = days_per_month * day
     year = SECONDS_PER_YEAR
 
-    if seconds >= SECONDS_PER_YEAR && seconds % year < seconds % month
+    if opts[:format] == :hours_only
+      hours = seconds / 3600.0
+      seconds = 0
+    elsif seconds >= SECONDS_PER_YEAR && seconds % year < seconds % month
       years = seconds / year
       months = seconds % year / month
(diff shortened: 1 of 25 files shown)


lib/chronic_duration.rb
```diff
@@ -173,6 +176,14 @@ def output(seconds, opts = {})
...
       hours = hours_int if hours - hours_int == 0 # if hours end with .0
       minutes = seconds = 0
+    when :hours_only
+      dividers = {
+        hours: "h", keep_zero: true
+      }
+
+      hours = hours.round(2)
+      hours_int = hours.to_i
+      hours = hours_int if hours - hours_int == 0 # if hours end with .0
```
aaron-contreras, line 186: Clever

aaron-contreras (approved):
Excellent work! Gave it a thorough test

lib/chronic_duration.rb
cbliard: Thanks! But actually, it's not mine. I copied it from another line: https://github.com/opf/openproject/blob/5ac1787fd6c441e4973032b953b141affd7d4b26/lib/chronic_duration.rb#L85-L86

PR #16020, opened 2024-07-02
Feature/40437 emoji reactions to work package comments

jjabari-op:
# Ticket

https://community.openproject.org/projects/openproject/work_packages/40437

Support emoji reactions to work package comments aka "journals"

## Screenshots

Screenshot 2024-10-25 at 6 41 08 PM

A default set of reactions is defined and rendered out as unicode points; storing the reaction strings and evaluating the corresponding unicode points allows for more flexibility for emoji modifiers. E.g. adding skin-tone support should also be evaluated at run time based on the users preference.

# Merge checklist

- [x] Added/updated tests
- [x] Tested major browsers (Chrome, Firefox, Safari, ...)


app/components/work_packages/activities_tab/journals/item_component/add_reactions.html.erb
```diff
@@ -0,0 +1,39 @@
+<%=
+  component_wrapper do
+    if journal.available_emoji_reactions.any?
+      render(Primer::Alpha::Overlay.new(
+              title: I18n.t("reactions.action_title"),
+              padding: :condensed,
+              anchor_side: :outside_top,
+              visually_hide_title: true
+            )) do |d|
```
Kharonus: 🟡 please do use meaningful names as `overlay`

app/models/emoji_reaction.rb
```diff
@@ -0,0 +1,59 @@
...
+    eyes: "\u{1F440}"
+  }.freeze
+
+  AVAILABLE_EMOJIS = EMOJI_MAP.values.freeze
+
+  belongs_to :user
+  belongs_to :reactable, polymorphic: true
+
+  validates :reaction, presence: true
+  validates :user_id, uniqueness: { scope: %i[reactable_type reactable_id reaction] }
```
Kharonus, line 47: 🟡 Please check, whether those active model validations can be moved to the contracts.

config/locales/en.yml
```diff
@@ -546,6 +546,15 @@ en:
         Click to assign or change the color of this priority.
         It can be used for highlighting work packages in the table.
 
+  reactions:
+    action_title: "React"
+    add_reaction: "Add reaction"
+    react_with: "React with %{reaction}"
+    and_user: "and %{user}"
+    and_n_others_singular: "and %{n} other"
+    and_n_others_plural: "and %{n} others"
```
Kharonus: 🟡 we should use pluralisation from I18n here like:

```yaml
and_others:
  one: and 1 other
  others: and %{count} others
```

example: see [here](https://github.com/opf/openproject/pull/17017/files#diff-44438ce218f5287c58d0017f965d888715635d94280669896f75841fbd7b4cd7R258-R261)

config/locales/en.yml
```diff
@@ -934,6 +943,7 @@ en:
         could_not_be_copied: "%{dependency} could not be (fully) copied."
         does_not_exist: "does not exist."
         error_enterprise_only: "%{action} is only available in the OpenProject Enterprise edition"
+        error_not_found: "not found."
```
Kharonus: 🟡 prefixing the locale key with `error` again seems ... duplicate? Could we use `not_found` instead?

spec/contracts/emoji_reactions/base_contract_spec.rb
```diff
@@ -0,0 +1,124 @@
...
+
+      it_behaves_like "contract is valid"
+    end
+
+    context "when reactable is neither a journal nor a work package" do
+      let(:unknown_object) { build_stubbed(:project) }
+
+      before { emoji_reaction.reactable = unknown_object }
+
+      it_behaves_like "contract is invalid", base: :error_unauthorized
```
Kharonus: 🟡 error code should be `:invalid`

Kharonus (changes requested):
only a few found in pair review

app/models/emoji_reaction.rb
akabiru: Thanks, we can remove the `reaction` validation as it's redundant at contract and association level (belongs_to will ensure presence too).

However, we do not have a Uniqueness validator for contracts; and AFAICT from [a text search](https://github.com/search?q=repo%3Aopf%2Fopenproject+uniqueness%3A&type=code) similar cases, we fallback to the default AR validation for uniqueness

```ruby
An error occurred while loading ./spec/contracts/emoji_reactions/base_contract_spec.rb.
Failure/Error: validates :user_id, uniqueness: { scope: %i[reactable_type reactable_id reaction] }

ArgumentError:
  Unknown validator: 'UniquenessValidator'
```

PR #16032, opened 2024-07-03
Implement tests for sharing project queries

klaustopher:
Follow up to #15971 adding all the tests

---

## Tests

> NOTE: Not being able to edit means you can make changes to the query but then it only offers the "Save as" function

- [x] User Permissible service for non-project scoped resources are not tested yet
- [x] Tests for the `SharesController`
- [x] Tests for the sharing strategies 
- [x] Feature Tests for the entire ProjectQuery sharing workflow
  - [x] Adding a share and accessing the project query with the shared user (@aaron-contreras)
  - [x] Adding a share with edit and changing with the shared user ( @aaron-contreras )
  - [x] Making a query public and accessing it with another user (not allowed to edit) ( @aaron-contreras )
  - [x] Making a query public and adding a share with edit permission and then editing it as shared user ( @aaron-contreras )
  - [x] Making a query public and the owner not having the `manage_public_project_queries` and not being able to edit ( @aaron-contreras )

Found while writing specs
- [x] Re-work `UserDetailsComponent` to be entity specific (current work package specific details don't fit with a non-project scoped resource like a project query) ▶️ #16090 
- [x] As a user that has been shared a project list with edit permissions, I'm not able to edit the project query by clicking "Save"

---

See: https://community.openproject.org/work_packages/55163


app/contracts/queries/projects/project_queries/base_contract.rb
```diff
@@ -57,10 +57,10 @@ def user_is_logged_in
 
     def allowed_to_modify_private_query
```
aaron-contreras: I second guessed whether using `#model.editable?` here as it's quite broad but would be consistent with the rest of checks in other places.

klaustopher:
@HDinger notified me that the preview in LookBook was broken. This is fixed in [8f9e0af](https://github.com/opf/openproject/pull/16032/commits/8f9e0afa2d74d33645f4e392929d327feb39dd27) and this already uncovered a bug of a hardcoded title :D

klaustopher:
Approved!

PR #16036, opened 2024-07-04
Switch empty state when public flag is toggled

klaustopher:
One more left over ticket from #15971: When toggling the public state of the project list, re-render the modal

### DIFF
--- a/app/controllers/projects/queries_controller.rb
+++ b/app/controllers/projects/queries_controller.rb
@@ -28,6 +28,7 @@
 
 class Projects::QueriesController < ApplicationController
   include Projects::QueryLoading
+  include OpTurbo::ComponentStream
 
   # No need for a more specific authorization check. That is carried out in the contracts.
   no_authorization_required! :show, :new, :create, :rename, :update, :toggle_public, :destroy
@@ -79,7 +80,16 @@ class Projects::QueriesController < ApplicationController
              .new(user: current_user, model: @query)
              .call(public: to_be_public)
 
-    render_result(call, success_i18n_key: "#{i18n_key}.success", error_i18n_key: "#{i18n_key}.failure")
+    # Load shares and replace the modal
+    strategy = SharingStrategies::ProjectQueryStrategy.new(@query, user: current_user)
+    shares = strategy.shares_query({}).results
+    replace_via_turbo_stream(component: Shares::ModalBodyComponent.new(strategy:, shares:, errors: []))
+
+    respond_with_turbo_streams do |format|
+      format.html do
+        render_result(call, success_i18n_key: "#{i18n_key}.success", error_i18n_key: "#{i18n_key}.failure")
+      end
+    end
   end
 
   def destroy
--- a/app/controllers/projects/queries_controller.rb
+++ b/app/controllers/projects/queries_controller.rb
@@ -73,7 +73,7 @@ class Projects::QueriesController < ApplicationController
   end
 
   def toggle_public # rubocop:disable Metrics/AbcSize
-    to_be_public = !@query.public?
+    to_be_public = ActiveRecord::Type::Boolean.new.cast(params["value"])
     i18n_key = to_be_public ? "lists.publish" : "lists.unpublish"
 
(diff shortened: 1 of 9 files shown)


klaustopher:
I am very very confused why this does not work. It renders the exact same response body as other `replace` calls, but it is not replacing

Response that works:

```html
<turbo-stream action="replace" target="share_modal_body">
    <template>
      <turbo-frame id="share_modal_body">...</turbo-frame>
    </template>
</turbo-stream>
```

Response that does not work:

```html
<turbo-stream action="replace" target="share_modal_body">
    <template>
      <turbo-frame id="share_modal_body">...</turbo-frame>
    </template>
</turbo-stream>
```

If anybody has an idea, I'd be very happy to understand why

klaustopher:
It might boil down to how primer components mke the request. They generate a `fetch` request from their JS code.

https://github.com/primer/view_components/blob/ae8dbe23040ded74fab853ac3dc2068711c625f1/app/components/primer/alpha/toggle_switch.ts#L161-L172

@oliverguenther pointed out that the `Accept` header needs to contain `text/vnd.turbo-stream.html` but it doesn't since it just sets `Accept: */*`

klaustopher:
We have to wait for a decision on https://github.com/primer/view_components/issues/2940 if we can use the component or if we need to rebuild a bit from scratch so we can use turbo with it. We can already merge this and create anotehr follow up in the future

app/models/sharing_strategies/base_strategy.rb
```diff
@@ -79,5 +79,24 @@ def custom_empty_state_component?
...
+                 .call(params)
+
+      # Set default filter on the entity
+      @query.where("entity_id", "=", entity.id)
+      @query.where("entity_type", "=", entity.class.name)
+      if entity.respond_to?(:project)
+        @query.where("project_id", "=", entity.project.id)
+      end
+
+      @query.order(name: :asc) unless params[:sortBy]
```
toy, line 97: Maybe check for params is not needed as calls to order are appended

app/controllers/projects/queries_controller.rb
```diff
@@ -71,15 +72,27 @@ def update
...
              .new(user: current_user, model: @query)
              .call(public: to_be_public)
 
-    render_result(call, success_i18n_key: "#{i18n_key}.success", error_i18n_key: "#{i18n_key}.failure")
+    respond_to do |format|
+      format.turbo_stream do
+        # Load shares and replace the modal
+        strategy = SharingStrategies::ProjectQueryStrategy.new(@query, user: current_user)
+        shares = strategy.shares_query({}).results
+        replace_via_turbo_stream(component: Shares::ModalBodyComponent.new(strategy:, shares:, errors: []))
```
toy, line 88: Maybe better to pass only `strategy` to `ModalBodyComponent`, to not pass `shares` separately after getting it through `strategy method`

spec/controllers/projects/queries_controller_spec.rb
```diff
@@ -328,7 +328,7 @@
         it "passes variables to template" do
           allow(controller).to receive(:render).and_call_original
 
-          post "toggle_public", params: { id: 42 }
+          post "toggle_public", params: { id: 42, value: 1 }
```
toy: No spec with `value: 0`, only one without any to check authentication requirement

app/controllers/projects/queries_controller.rb
klaustopher: Yes, I will do exactly that in the [next PR](https://community.openproject.org/work_packages/56243/activity) because I will need to dynamically add something to the shares

app/models/sharing_strategies/base_strategy.rb
klaustopher: The default order for a member is created_at, so it makes sense to reorder here depending on params being present

spec/controllers/projects/queries_controller_spec.rb
klaustopher: done

toy (approved):

PR #16018, opened 2024-07-02
[#55966] Define "Enabled in Projects" as a fully qualified nested resource

akabiru:
https://community.openproject.org/work_packages/55966

Follows https://github.com/opf/openproject/pull/15990

The storages controller is already overloaded, defined a nested controller instead that can much more easily follow rails CRUD conventions.

----

```ruby
                                                  Prefix Verb   URI Pattern                                                                           Controller#Action
                 admin_settings_storage_project_storages GET    /admin/settings/storages/:storage_id/project_storages(.:format)                       storages/admin/projects/project_storages#index
                                                         POST   /admin/settings/storages/:storage_id/project_storages(.:format)                       storages/admin/projects/project_storages#create
              new_admin_settings_storage_project_storage GET    /admin/settings/storages/:storage_id/project_storages/new(.:format)                   storages/admin/projects/project_storages#new
                  admin_settings_storage_project_storage DELETE /admin/settings/storages/:storage_id/project_storages/:id(.:format)                   storages/admin/projects/project_storages#destroy
```

PR #16038, opened 2024-07-04
[#55961] Define the project lists query for activated storages

judithroth:
https://community.openproject.org/work_packages/55961

This also defines a basic table in the view, so it can be tested. I've choosen not to go with a new query filter for projects belonging to a storage because it seemed to be more work than necessary and not provide any further benefits, since for the API there already is such a filter (which also seems to be tailored to the needs of the API. It would not have worked for the table because the table should include archived projects as well).


modules/storages/app/views/storages/admin/storages/project_mappings.html.erb
```diff
@@ -34,3 +34,10 @@
     )
   )
 %>
+<%=
+  render(Storages::ProjectStorages::ProjectMapping::TableComponent.new(
+    query: @project_query,
+    params: params
```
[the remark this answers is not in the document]
judithroth: @akabiru I assume this is a false positive and could be added to brakeman.ignore, but I am not sure about it. Do you have a suggestion on how to handle this?

akabiru (changes requested):
Looking good Judith! ⭐ 

🟠 Could we add https://github.com/opf/openproject/pull/16018 first into `dev` first and rebase this PR so it fits into the new structure?

modules/storages/app/components/storages/project_storages/project_mapping/row_component.rb
```diff
@@ -0,0 +1,34 @@
...
+# You should have received a copy of the GNU General Public License
+# along with this program; if not, write to the Free Software
+# Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+#
+# See COPYRIGHT and LICENSE files for more details.
+#++
+
+# Purpose: Defines how to format the components within a table row of Projects
+# associated with a Storage
+module Storages::ProjectStorages::ProjectMapping
```
akabiru: 🟢 Given the new namespacing structure in https://github.com/opf/openproject/pull/16018 I think the following would be more fitting- as we essentially can drop the "project_mapping" naming across the board
```suggestion
module Storages::ProjectStorages
```

modules/storages/app/components/storages/project_storages/project_mapping/table_component.rb
```diff
@@ -0,0 +1,48 @@
...
+# Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+#
+# See COPYRIGHT and LICENSE files for more details.
+#++
+
+# Purpose: Defines a table based on TableComponent for listing the
+# Projects for a given Storage.
+# See also: row_component.rb, which contains a method
+# for every "column" defined below.
+module Storages::ProjectStorages::ProjectMapping
```
akabiru: 🟢 ditto:
```suggestion
module Storages::ProjectStorages
```

modules/storages/app/components/storages/project_storages/project_mapping/row_component.rb
akabiru: i.e. we'd have

1. `Storages::ProjectStorages::ProjectRowComponent` - for  projects
2. `Storages::ProjectStorages::StorageRowComponent` - for  storages _(in the future when the [current list is primerised](https://qa.openproject-edge.com/projects/demo-project/settings/project_storages/external_file_storages))_

judithroth:
> Looking good Judith! ⭐
> 
> 🟠 Could we add #16018 first into `dev` first and rebase this PR so it fits into the new structure?

Sure! - I thought the same actually so I can adjust wording and some file names :slightly_smiling_face:

modules/storages/app/components/storages/project_storages/project_mapping/row_component.rb
judithroth: I'd rather go with another namespace here. It's also nice if the files are grouped (not only by name but by a folder). What do you think about
`Storages::ProjectStorages::Projects::RowComponent` (and `Storages::ProjectStorages::Projects::TableComponent`) and keeping the other files as is?

akabiru: That also works- +1 on folder naming, can be further reused 👍🏾

modules/storages/app/components/storages/project_storages/projects/table_component.rb
```diff
@@ -0,0 +1,48 @@
...
+# Projects for a given Storage.
+# See also: row_component.rb, which contains a method
+# for every "column" defined below.
+module Storages::ProjectStorages::Projects
+  class TableComponent < Projects::TableComponent # rubocop:disable OpenProject/AddPreviewForViewComponent
+    include OpTurbo::Streamable
+
+    def sortable?
+      false
+    end
```
akabiru, line 39: Oddly enough this doesn't seem to kick in:

- [ ] I still see the "sorting icon" - which should be disabled in this feature.
- [ ] The action list is filled in with the default properties, in this feature spec- the menu actions are (1) Edit Project Folder (2) Remove project - I suggest either leaving them empty or defining placeholders.

Screenshot 2024-07-05 at 11 15 57 AM

akabiru: https://github.com/opf/openproject/assets/17295175/b2a0d89b-4261-4aaf-b8b5-2925c13161f4

akabiru (commented):
🟠 Almost there - two issues mentioned in the comments. 🛠️

akabiru (approved):
:shipit:

akabiru:
We agreed to address the issues separately 👍🏾

PR #16039, opened 2024-07-04
[OP#55966] Define "Add Projects" dialog that allows selecting one or many projects for activation with include subprojects option

akabiru:
https://community.openproject.org/work_packages/55966

Follows https://github.com/opf/openproject/pull/16018

🚧 _**Manual Folder Selection is not included**_, it will be added separately with the File Picker implementation

https://github.com/user-attachments/assets/13d96967-95c3-4e11-aa64-50cc5b1abe13

### DIFF
--- a/modules/storages/app/components/storages/admin/storages/add_projects_form_modal_component.rb
+++ b/modules/storages/app/components/storages/admin/storages/add_projects_form_modal_component.rb
@@ -29,7 +29,7 @@
 module Storages
   module Admin
     module Storages
-      class AddProjectsFormModalComponent < ApplicationComponent # rubocop:disable OpenProject/AddPreviewForViewComponent
+      class AddProjectsFormModalComponent < ApplicationComponent
         include OpPrimer::ComponentHelpers
         include OpTurbo::Streamable
 
--- a/modules/storages/app/components/storages/admin/storages/add_projects_form_modal_component.rb
+++ b/modules/storages/app/components/storages/admin/storages/add_projects_form_modal_component.rb
@@ -30,6 +30,7 @@ module Storages
   module Admin
     module Storages
       class AddProjectsFormModalComponent < ApplicationComponent # rubocop:disable OpenProject/AddPreviewForViewComponent
+        include OpPrimer::ComponentHelpers
         include OpTurbo::Streamable
 
         def initialize(project_storage:, **)
--- a/modules/storages/app/components/storages/admin/storages/add_projects_form_modal_component.rb
+++ b/modules/storages/app/components/storages/admin/storages/add_projects_form_modal_component.rb
@@ -1,5 +1,3 @@
-# frozen_string_literal: true
-
 #-- copyright
 # OpenProject is an open source project management software.
 # Copyright (C) 2012-2024 the OpenProject GmbH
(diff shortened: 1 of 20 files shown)


modules/storages/app/controllers/storages/admin/storages/project_storages_controller.rb
```diff
@@ -38,26 +40,39 @@ class Storages::Admin::Storages::ProjectStoragesController < ApplicationControll
...
+                              include_sub_projects: include_sub_projects?)
+                         .call
+
+    create_service.on_success { update_project_list_via_turbo_stream(url_for_action: :index) }
+
+    create_service.on_failure do
+      update_flash_message_via_turbo_stream(
+        message: join_flash_messages(create_service.errors),
+        full: true, dismiss_scheme: :hide, scheme: :danger
+      )
```
mereghost, line 70: 🟡 One trick here is to use a `return update_project....` on the success block, so that the method returns immediately, and then just handle a failure scenario below (or the other way around 😉).

I tend to find it clearer that way.

modules/storages/app/services/storages/last_project_folders/bulk_create_service.rb
```diff
@@ -0,0 +1,110 @@
...
+# along with this program; if not, write to the Free Software
+# Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+#
+# See COPYRIGHT and LICENSE files for more details.
+#++
+
+module Storages::LastProjectFolders
+  class BulkCreateService < ::BaseServices::BaseCallable
+    def initialize(user:, project_storages:)
+      super()
```
mereghost, line 34: 🟡 This call is unneeded as `BaseCallable` has no initializer.

modules/storages/app/services/storages/project_storages/bulk_create_service.rb
```diff
@@ -0,0 +1,152 @@
...
+# along with this program; if not, write to the Free Software
+# Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+#
+# See COPYRIGHT and LICENSE files for more details.
+#++
+
+module Storages::ProjectStorages
+  class BulkCreateService < ::BaseServices::BaseCallable
+    def initialize(user:, projects:, storage:, project_folder_mode:, include_sub_projects: false)
+      super()
```
mereghost, line 34: 🟡 Same as above :)

modules/storages/app/components/storages/admin/storages/add_projects_form_modal_component.rb
```diff
@@ -0,0 +1,63 @@
...
+module Storages
+  module Admin
+    module Storages
+      class AddProjectsFormModalComponent < ApplicationComponent
+        include OpPrimer::ComponentHelpers
+        include OpTurbo::Streamable
+
+        def initialize(project_storage:, **)
+          @project_storage = project_storage
+          @storage = project_storage.storage
```
mereghost: 🟡 Are we making heavy usage of `storage`? If not I'd go with just accessing it via `ProjectStorage`.

mereghost (approved):
Seems good! :D As usual I left some comments/suggestions that shouldn't block the merge as those are just icing in the cake.

modules/storages/app/components/storages/admin/storages/add_projects_form_modal_component.rb
akabiru: Actually no, just once [here](https://github.com/opf/openproject/pull/16039/files#diff-5d4738f1082585a072200ec3a4df128039e639e8acd0bdcee4c1738146db65f2R35)

```ruby
url: admin_settings_storage_project_storages_path(@storage),
```

Will get rid of the extra var 👍🏾

modules/storages/app/controllers/storages/admin/storages/project_storages_controller.rb
akabiru: Hmm, I somehow find that less clear- it needs an extra brain cell to "know" that return will exit on the block. IMO, I prefer the current approach as it's dead simple and clear that we're updating turbo streams based on the failure/success states- and then in the end responding with those turbo streams.

---

I may be missing your full picture- I'll ping you for a brief chat 😄

modules/storages/app/services/storages/last_project_folders/bulk_create_service.rb
akabiru: yeah, I think rubocop was shouting at me- anytime it sees an inheritance structure it expects the parent to be initialized for good measure

modules/storages/app/services/storages/project_storages/bulk_create_service.rb
akabiru: ditto :)

modules/storages/app/services/storages/last_project_folders/bulk_create_service.rb
mereghost: Weird. Mine doesn't complain and the issue never happens on CI =/

akabiru: As it's just a no-op I'd just leave it as is- it also seems like good practice to me

Screenshot 2024-07-16 at 1 01 31 PM

modules/storages/app/controllers/storages/admin/storages/project_storages_controller.rb
akabiru: _Update:_ We had a chat and settled to retain the current structure given the differentiating factor is what turbo streams are updated based on failure/success and consistent factor that we respond with all streams in the end. Cheers for the chat @mereghost 🍻

PR #16045, opened 2024-07-05
[#56012] all Viepoints are saved to the last BCF

Kharonus:
[#56012](https://community.openproject.org/work_packages/56012)

### WAT?

- refreshing topic state after creating a wp (topic)
  - A work package is the representation of a BCF topic in our domain
  - when a work package is created, a topic is created via the BCF API
  - the viewpoint is linked then to this topic and created alongside with the work package
  - *BEFORE*: the topic id was kept in state of the viewpoint service, so creating a second work package led to all viewpoints being referenced to the first topic 
  - *NOW*: the topic id is cleansed from the service, once the user hits the "Create" button in the frontend, so that a new topic is created, when the second work package is created
- added small script to docker dev container, that can setup the BIM dependencies

### DIFF
--- a/docker/dev/backend/scripts/setup-bim
+++ b/docker/dev/backend/scripts/setup-bim
@@ -1,0 +1,36 @@
+#!/bin/sh
+
+apt-get install -y wget unzip
+
+# https://learn.microsoft.com/en-gb/dotnet/core/install/linux-debian#debian-12
+wget --quiet https://packages.microsoft.com/config/debian/12/packages-microsoft-prod.deb -O /tmp/packages-microsoft-prod.deb
+dpkg -i /tmp/packages-microsoft-prod.deb
+rm /tmp/packages-microsoft-prod.deb
+
+apt-get update -qq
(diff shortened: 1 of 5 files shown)


docker/dev/backend/scripts/setup-bim
wielinde, line 1: Hey @Kharonus, great to see how you bake Docker support for BIM development into this PR. I share your view here. 

One note: the content of the file is somewhat similar to a couple of other places (docker prod, packaging and manual dev setup) where we solve similar tasks: the provisioning of the required OS packages and command line tools. Having those places. That was already complex to manage back in the days. Now, you introduce a fourth place here. Maybe we could have a talk about how to reduce it?

Kharonus: Yep, sure. The place I introduced has at least no effect, if you are not running the script in an existing container with 

```
docker compose exec --user root worker setup-bim
```

But, it might be helpful to actually harmonize all those places. But I wouldn't do so in this bug ticket.

apfohl (approved):

PR #16046, opened 2024-07-05
[#55962] Add the project list view, including the "project folder type" column

judithroth:
https://community.openproject.org/work_packages/55962


judithroth:
The CI does not like commas in branch names. I'll open a new one without commas :smile:

PR #16047, opened 2024-07-05
Shared spec for acts as favorable

toy:
Follow up for #15992

Besides the shared spec:
* Use it_behaves_like for "creates an audit trail on destroy" shared examples, as otherwise the subject overrides the one in outer scope + remove duplicate inclusion in Project spec
* Disambiguate module naming in acts as modules, as besides `ClassMethods` on top level, some had `ClassMethods` inside `InstanceMethods` with confusing `base.extend ClassMethods` both in `included` on top and `included`/`prepended` of `InstanceMethods`

### DIFF
--- a/lib_static/plugins/acts_as_customizable/lib/acts_as_customizable.rb
+++ b/lib_static/plugins/acts_as_customizable/lib/acts_as_customizable.rb
@@ -60,6 +60,7 @@ module Redmine
 
       module InstanceMethods
         def self.included(base)
+          base.extend AddClassMethods
           base.extend HumanAttributeName
         end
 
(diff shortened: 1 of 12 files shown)


lib_static/plugins/acts_as_customizable/lib/acts_as_customizable.rb
```diff
@@ -438,7 +438,7 @@ def custom_field_values_cache
           @custom_field_values_cache ||= {}
         end
 
-        module ClassMethods
```
klaustopher, line 441: It's sort of a convention to call the module that adds class methods `ClassMethods` see https://github.com/search?q=%22module+ClassMethods%22+language%3Aruby&type=code. Why are we changing it?

toy: I wrote in the PR description, because there two separate `ClassMethods` modules in one file, so one keeps the name and other gets a different one to disambiguate

klaustopher (approved):

PR #16059, opened 2024-07-08
Allow users with edit permission to save the query

klaustopher:
Extracted from #16032. @aaron-contreras has uncovered while writing tests that users cannot modify a query when they have the edit permission. This extracts the fix for this already so it can be tested already


aaron-contreras (approved):
Excellent

PR #16064, opened 2024-07-08
[55656] Remove derivedRemainingTime from form payload

cbliard:
`derivedRemainingTime` is not writeable. That's specified in the work package contract. But as the payload representer was using a lambda for `:writable` for `derivedRemainingTime`, the writability given by the contract was ignored, and it was always added if the lambda returned `true` (and it always returned true in work-based mode).

It produced issues like when copying a work package having a `derivedRemainingTime` being set, as reported in
https://community.openproject.org/wp/55656.

Removing the lambda makes the representer solely rely on the contract to determine the writability of the `derivedRemainingTime` property, and fixes the issue.

### DIFF
--- a/lib/api/utilities/payload_representer.rb
+++ b/lib/api/utilities/payload_representer.rb
@@ -51,12 +51,15 @@ module API
 
           # Note: `:writeable` is not a typo, it's used by declarative gem
           writable = property[:writeable]
+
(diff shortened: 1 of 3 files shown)


aaron-contreras (approved):
Nice catch, straightforward! 🍹

cbliard:
It looks like the tests have failed only locally. Flickering.

PR #16066, opened 2024-07-08
Remove no permission alert

klaustopher:
In the old version we rendered two alert boxes, primer states that this is not desirable. So, instead we removed the alert that the user is not allowed to edit the permissions and instead show a tooltip on the disabled input elements.

Also, we now show the role even for those that do not have permission to change the role. That information was already "leaked" anyway via the filters and the dialog makes a lot more sense this way.

---

Closes https://community.openproject.org/work_packages/56242/activity

### DIFF
--- a/app/components/shares/permission_button_component.rb
+++ b/app/components/shares/permission_button_component.rb
@@ -92,5 +92,9 @@ module Shares
 
       ["tooltip--left"]
     end
+
+    def editable?
+      strategy.manageable? && share.principal != User.current
+    end
   end
 end
--- a/app/components/shares/permission_button_component.rb
+++ b/app/components/shares/permission_button_component.rb
@@ -32,10 +32,10 @@ module Shares
     include OpPrimer::ComponentHelpers
     include OpTurbo::Streamable
 
-    def initialize(share:, available_roles:, **system_arguments)
+    def initialize(share:, strategy:, **system_arguments)
       super
 
-      @available_roles = available_roles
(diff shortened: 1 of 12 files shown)


app/components/shares/project_queries/public_flag_component.rb
```diff
@@ -56,6 +56,18 @@ def published?
...
+      def tooltip_message
+        return if can_publish?
+
+        I18n.t("sharing.project_queries.publishing_denied")
+      end
+
+      def tooltip_wrapper_classes
+        return ["d-flex", "flex-column"] if can_publish?
+
+        ["tooltip--bottom", "d-flex", "flex-column"]
```
toy: Better to not repeat `["d-flex", "flex-column"]`

app/components/shares/permission_button_component.rb
```diff
@@ -80,5 +80,21 @@ def form_inputs(role_id)
...
+    end
+
+    def tooltip_wrapper_classes
+      return [] if strategy.manageable?
+
+      ["tooltip--left"]
+    end
+
+    def editable?
+      strategy.manageable? && share.principal != User.current
```
toy, line 97: User can't edit own share?

klaustopher: No, you cannot edit your own share. This was already the case in the Workpackage implementation and got lost when generalizing

toy (approved):

PR #16077, opened 2024-07-09
Fix flickering cleanup_inherited_roles_service_integration_spec test

cbliard:
Test failing on https://github.com/opf/openproject/actions/runs/9845156852/job/27180064775?pr=16073.

The subject test was `Member.find_by(principal: users.last).roles`. As there are 3 different `Member` records having the same principal, it could return any of them. Most of the time it would always be the first one, which is a project role membership and satisfies the test conditions, but occasionnally, it would be the global role membership or the work package role membership and the test would fail.

Using `Member.where(principal: users.last).flat_map(&:roles)` and testing for all roles fixes the test flakiness.

Also changed some wording here and there to make the test easier to understand.


ulferts (approved):
👍 🙇

PR #16078, opened 2024-07-09
[#55962] Add the project list view, including the "project folder" custom column

judithroth:
https://community.openproject.org/work_packages/55962

### DIFF
--- a/modules/storages/app/models/queries/storages/projects/filter/storages_filter.rb
+++ b/modules/storages/app/models/queries/storages/projects/filter/storages_filter.rb
@@ -42,10 +42,8 @@ module Queries::Storages::Projects::Filter
     end
 
     def allowed_values
-      @allowed_values ||= Storages::ProjectStorage
-                            .includes(:storage)
-                            .distinct
-                            .pluck(:name, :storage_id)
+      @allowed_values ||= Storages::Storage
+                            .pluck(:name, :id)
     end
 
     def available?
--- a/modules/storages/app/models/queries/storages/projects/filter/storages_filter.rb
+++ b/modules/storages/app/models/queries/storages/projects/filter/storages_filter.rb
@@ -1,8 +1,8 @@
 # frozen_string_literal: true
 
-#-- copyright
+# -- copyright
 # OpenProject is an open source project management software.
-# Copyright (C) 2012-2024 the OpenProject GmbH
+# Copyright (C) 2024 the OpenProject GmbH
 #
 # This program is free software; you can redistribute it and/or
 # modify it under the terms of the GNU General Public License version 3.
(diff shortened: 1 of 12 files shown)


modules/storages/app/models/queries/storages/projects/filter/storages_filter.rb
```diff
@@ -0,0 +1,70 @@
...
+      else
+        raise "unsupported operator"
+      end
+    end
+
+    def where
+      nil
+    end
+  end
+end
```
akabiru, line 68: 🟠 Seems we're missing a unit test for the storage filter- would be worth adding one I think. Ex. https://github.com/opf/openproject/blob/dev/spec/models/queries/projects/filters/available_project_attributes_filter_spec.rb

akabiru (dismissed):
Looking great Judith! 👍🏾 One proposal to add a missing unit spec, and should be good to go! 👍🏾

modules/storages/spec/features/storages/admin/project_storages_spec.rb
```diff
@@ -37,11 +37,30 @@
...
+           project:,
+           storage:,
+           project_folder_mode: "automatic")
+  end
+  shared_let(:archived_project_project_storage) do
+    create(:project_storage,
+           project: archived_project,
+           storage:,
+           project_folder_mode: "inactive")
+  end
```
akabiru, line 63: 🟢  _nit_: I personally prefer the compact one-liner approach where possible, reduces "setup fatigue" IMO. The faster we can get to the tests without scrolling, the better 😄

modules/storages/spec/features/storages/admin/project_storages_spec.rb
```diff
@@ -37,11 +37,30 @@
...
+           :as_automatically_managed,
+           name: "My Nextcloud Storage")
+  end
+  shared_let(:project_storage) do
+    create(:project_storage,
+           project:,
+           storage:,
+           project_folder_mode: "automatic")
+  end
+  shared_let(:archived_project_project_storage) do
```
akabiru, line 58: Ah, seems there was already a typo 🧹 

```suggestion
  shared_let(:archived_project_storage) do
```

modules/storages/app/models/queries/storages/projects/filter/storages_filter.rb
```diff
@@ -0,0 +1,68 @@
...
+    end
+
+    def type
+      :list
+    end
+
+    def allowed_values
+      @allowed_values ||= Storages::Storage
+                            .pluck(:name, :id)
+    end
```
akabiru, line 47: Nice catch 👍🏾

akabiru (approved):
🚀

modules/storages/spec/features/storages/admin/project_storages_spec.rb
akabiru: Ah, it's not a typo! 🤦🏾

modules/storages/spec/features/storages/admin/project_storages_spec.rb
judithroth: I somewhat agree - it depends on how long the line gets. Sometimes it's also easier to see what the differences are between certain test objects. But if you only need 2 or 3 short arguments, I'd also go for oneliners :+1:

judithroth:
@akabiru ah, I missed that you already approved :smile: No need to do another review, I'll merge now :slightly_smiling_face:

PR #16079, opened 2024-07-09
[56194] Follow-up remaining issues for the Dark mode

bsatarnejad:
- [x] Light blue tooltip in BCF module 
- [x] Share modal input field is too dark
- [x] Filter value input field is too dark
- [x] Disabled Buttons in dark mode look weird
- [x] CKEditor Toolbar: WYSIWYG/Markdown toggle button is incomplete on dark
- [x] Spot-tooltips in dark mode have a bad contrast
- [x] Codemirror Markdown Source editor: Caret (blinking position) is black on black
- [x] Changes of attributes displayed in Baseline comparison unreadable
- [x] CKEditor edit mode: dark red # links to work packages are hard to read
- [x] Project specific notification settings in the my account settings have a white table header

https://community.openproject.org/projects/openproject/work_packages/56194/activity
https://community.openproject.org/projects/openproject/work_packages/56247/activity


app/views/custom_styles/_primer_color_mapping.erb
```diff
@@ -86,5 +86,7 @@
       --main-menu-bg-color: var(--overlay-bgColor);
       --type-form-conf-attribute--background: var(--overlay-bgColor);
       --select-arrow-bg-color-url: url("data:image/svg+xml;utf8,<svg fill='white' height='24' viewBox='0 0 24 24' width='24' xmlns='http://www.w3.org/2000/svg'><path d='M7 10l5 5 5-5z'/><path d='M0 0h24v24H0z' fill='none'/></svg>");
+      --button--primary-background-disabled-color: var(--primary-button-color--major1);
```
HDinger: You probably also have to override the `--button-primary-borderColor-disabled`. Otherwise it looks a bit weird.
Bildschirmfoto 2024-07-12 um 11 10 49

This is how it should like [in Primer](https://primer-lookbook.github.com/view-components/lookbook/inspect/primer/beta/button/playground?_display=%257B%2522theme%2522%253A%2522dark%2522%257D&scheme=primary&inactive=false&disabled=true):

Bildschirmfoto 2024-07-12 um 11 11 46

frontend/src/global_styles/layout/work_packages/_table_baseline.sass
```diff
@@ -46,7 +46,7 @@
     padding: 2px 2px 0 2px
 
   &--cell
-    background-color: #f3f3f3
+    background-color: var(--bgColor-muted)
```
HDinger: The contrast is too low, imho. Even in the light mode, you barely see the fields, and in dark mode it is even worse

Bildschirmfoto 2024-07-12 um 11 22 06

Bildschirmfoto 2024-07-12 um 11 22 52

frontend/src/global_styles/openproject/_forms.sass
```diff
@@ -76,7 +76,7 @@ $select-arrow: true !default
   border: $input-border
   border-radius: 2px
   background-color: transparent
-  color: var(--body-font-color)
+  color: var(--body-font-color) !important
```
HDinger: Why is that `important` needed here? Should we rather change the place where it gets overriden?

HDinger (changes requested):
Hey @bsatarnejad 

I have left some remarks. Further, please double check with the product team whether it is okay, that tooltips now changed from blue to black in the light mode. It feels very heavy to me.

frontend/src/global_styles/openproject/_forms.sass
bsatarnejad: In (webpack:///node_modules/@ng-select/ng-select/themes/default.theme.css), default color of input is #000 so the value that we set for inputs is ignored, so I added !important for the color that we set for inputs and force input value to use body-font-color.

HDinger: But you should be able to override that with a more specific selector in the `autocomplete.sass`. We already have a selector for the input there.

HDinger (approved):
The changes look good. Please rebase onto dev as we do not want that to be merged shortly before the relase

PR #16083, opened 2024-07-09
[56168] Fixed flash messages for api tokens

apfohl:
https://community.openproject.org/projects/openproject/work_packages/56168

This PR replaces classic flash messages for **Access Tokens** for the API tokens section with Primers banners to circumvent the problem of a not working close action on the old style flash messges.

image

image

### DIFF
--- a/app/controllers/my_controller.rb
+++ b/app/controllers/my_controller.rb
@@ -163,11 +163,9 @@ class MyController < ApplicationController
     result = APITokens::CreateService.new(user: current_user).call(token_name: params[:token_api][:token_name])
 
     result.on_success do |r|
-      flash[:info] = [
-        t("my.access_token.notice_reset_token", type: "API").html_safe,
-        content_tag(:strong, r.result.plain_value),
-        t("my.access_token.token_value_warning")
-      ]
+      token = Primer::Beta::Text.new(font_weight: :bold).render_in(view_context) { r.result.plain_value }
+      message = "#{t('my.access_token.notice_reset_token', type: 'API')} #{token} #{t('my.access_token.token_value_warning')}"
+      flash[:primer_banner] = { message:, scheme: :success }
 
       redirect_to action: "access_token"
     end
@@ -181,6 +179,7 @@ class MyController < ApplicationController
       respond_with_turbo_streams
     end
   end
+
   # rubocop:enable Metrics/AbcSize
 
   # rubocop:disable Metrics/AbcSize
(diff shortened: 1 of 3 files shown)


app/controllers/my_controller.rb
```diff
@@ -163,11 +163,9 @@ def generate_api_key
     result = APITokens::CreateService.new(user: current_user).call(token_name: params[:token_api][:token_name])
 
     result.on_success do |r|
-      flash[:info] = [
-        t("my.access_token.notice_reset_token", type: "API").html_safe,
-        content_tag(:strong, r.result.plain_value),
-        t("my.access_token.token_value_warning")
-      ]
+      token = Primer::Beta::Text.new(font_weight: :bold).render_in(view_context) { r.result.plain_value }
+      message = "#{t('my.access_token.notice_reset_token', type: 'API')} #{token} #{t('my.access_token.token_value_warning')}"
```
akabiru: 🟢 Some line breaks could improve the formatting

Updated https://github.com/opf/openproject/pull/16083/files#r1670613503

```suggestion
      message = join_flash_messages [
        t("my.access_token.notice_reset_token", type: "API"),
        token,
        t("my.access_token.token_value_warning")
      ]
```

akabiru (commented):
Looking good, one suggestion.

🟠 I'd also recommend adding one feature assertion for the primer flash. Ex. https://github.com/opf/openproject/blob/dev/modules/storages/spec/features/storages/admin/create_storage_spec.rb#L320-L324

app/controllers/my_controller.rb
akabiru: As it turns out, we have a [`join_flash_messages` helper](https://github.com/opf/openproject/blob/dev/app/helpers/flash_messages_output_safety_helper.rb#L38) from [`FlashMessagesOutputSafetyHelper`](https://github.com/opf/openproject/blob/dev/app/helpers/flash_messages_output_safety_helper.rb) which  uses [ActionView's `safe_join`](https://api.rubyonrails.org/v7.1.3/classes/ActionView/Helpers/OutputSafetyHelper.html#method-i-safe_join) under the hood- I'd use that instead.

apfohl: Nice find. That's awesome.

apfohl:
> Looking good, one suggestion.
> 
> 🟠 I'd also recommend adding one feature assertion for the primer flash. Ex. https://github.com/opf/openproject/blob/dev/modules/storages/spec/features/storages/admin/create_storage_spec.rb#L320-L324

I amended the controller specs for `MyController` to check the existence and correctness of the banners. Hope that's okay as well?

akabiru (approved):
Nice one :shipit:

PR #16084, opened 2024-07-09
[chore] change of logic for nextcloud storage hosts

Kharonus:
- instead of removing trailing slashes, we enforce them on storage hosts

### Hints

- this fixes [#56276](https://community.openproject.org/wp/56276)

### DIFF
--- a/modules/storages/app/models/storages/storage.rb
+++ b/modules/storages/app/models/storages/storage.rb
@@ -160,7 +160,11 @@ module Storages
     def uri
       return unless host
 
-      @uri ||= URI(host).normalize
+      @uri ||= if host.end_with?("/")
+                 URI(host).normalize
+               else
+                 URI("#{host}/").normalize
+               end
     end
 
     def connect_src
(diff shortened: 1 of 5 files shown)


wielinde:
Hey @Kharonus, without knowing all the details... Wouldn't it make sense to make the concatenation of host and path more robust? I can easily imagine where this leads to new issues. 

And if you still prefer adding a trailing slash then could we call it that method more specific? Instead of 'url' it actually is 'url_with_trailing_slash'. Then it looks at least a bit absurd 🤡

Kharonus:
Our problem is a bit, that we are misusing File.join to concatenate URI parts.

So, we did now one step: the storage uri will always be fetched with the method ".uri" this will now always return a uri with trailing slash.

In the end your point is right. We need a common way to construct those request URLs. Right now we have too many different ways to do so.

mereghost:
To compound the issue, `URI` adds a trailing slash to a `scheme://host` url but not when we have a path (`scheme://host/path`), and if there's one already it kinda throws a wrench on all the path joining.

We could, as a more robust fix, start using `URI.join`. It has its own peculiarities tho, as shown below.

```ruby
URI.join(storage.uri.origin, "#{storage.uri.path}/", "endpoint")
```

Kharonus:
Is that exclusive? For me, those are two different action within the same approach (making the usage of host urls more fail safe)

1) unify interface of `storage.uri`
2) add a new method `build_request_uri` to nextcloud util, that uses a more failsafe concatenation, maybe with `URI.join`

One could argue, that doing 2 can make 1 obsolete.

mereghost (approved):

PR #16086, opened 2024-07-09
Add owner to share

klaustopher:
Adds a virtual share for the owner so that they are included in the list of shares

---

Implements https://community.openproject.org/projects/openproject/work_packages/56243/activity

### DIFF
--- a/app/controllers/shares_controller.rb
+++ b/app/controllers/shares_controller.rb
@@ -32,7 +32,6 @@ class SharesController < ApplicationController
   include MemberHelper
 
   before_action :load_entity
-  before_action :load_shares, only: %i[index dialog]
   before_action :load_selected_shares, only: %i[bulk_update bulk_destroy]
   before_action :load_share, only: %i[destroy update resend_invite]
   before_action :enterprise_check, only: %i[index]
@@ -44,11 +43,11 @@ class SharesController < ApplicationController
   def dialog; end
 
   def index
-    unless @query.valid?
-      flash.now[:error] = query.errors.full_messages
+    unless sharing_strategy.query.valid?
+      flash.now[:error] = sharing_strategy.query.errors.full_messages
     end
 
-    render Shares::ModalBodyComponent.new(strategy: sharing_strategy, shares: @shares, errors: @errors), layout: nil
+    render Shares::ModalBodyComponent.new(strategy: sharing_strategy, errors: @errors), layout: nil
   end
 
   def create # rubocop:disable Metrics/AbcSize,Metrics/PerceivedComplexity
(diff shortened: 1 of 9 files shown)


app/controllers/shares_controller.rb
```diff
@@ -44,17 +43,19 @@ class SharesController < ApplicationController
...
 
-    render Shares::ModalBodyComponent.new(strategy: sharing_strategy, shares: @shares, errors: @errors), layout: nil
+    render Shares::ModalBodyComponent.new(strategy: sharing_strategy, errors: @errors), layout: nil
   end
 
   def create # rubocop:disable Metrics/AbcSize,Metrics/PerceivedComplexity
     overall_result = []
     @errors = ActiveModel::Errors.new(self)
 
+    had_shares_before_adding = sharing_strategy.shares.present?
```
aaron-contreras: Thank you.

app/controllers/shares_controller.rb
```diff
@@ -311,31 +314,21 @@ def load_entity # rubocop:disable Metrics/AbcSize
...
   def load_selected_shares
     @selected_shares = Member.includes(:principal)
                              .of_entity(@entity)
                              .where(id: params[:share_ids])
   end
+
+  def query_params
+    params
+      .slice(:filters, :sortBy, :groupBy)
+      .permit! # ParamsToQueryService will parse the data, so we can permit everything here
```
aaron-contreras, line 332: TIL this exists

app/models/sharing_strategies/base_strategy.rb
```diff
@@ -94,9 +95,14 @@ def shares_query(params) # rubocop:disable Metrics/AbcSize
...
 
-      @query.order(name: :asc) unless params[:sortBy]
+      @query.order(name: :asc) unless @query_params[:sortBy]
 
       @query
     end
+
+    def shares(reload: false)
+      @shares = nil if reload
+      @shares ||= query.results
```
aaron-contreras, line 105: Nice trick

aaron-contreras (approved):
This might be just a detail and not very important. I understand now that the owner is a virtual share amongst the other list of shares.

It is however either being included with every role filter unless there are now "real" shares matching the filter in which case it's removed. I leave a demo below:

https://github.com/opf/openproject/assets/61627014/667a5590-1dcb-4256-a9f5-214c17e2a774

Other than this, everything else is working as it used to 🍹

klaustopher:
Oh yes, this makes sense ... I think we can add a bit more smartness and look at the filter and the roles in there. Will add this tomorrow morning. Thanks for the review

klaustopher:
I added some code so that the owner is just shown when filtering for users and when the role matches

aaron-contreras (approved):
A+ work!

🍹

PR #16090, opened 2024-07-10
Move sub description of each share into the strategy

klaustopher:
Closes https://community.openproject.org/projects/openproject/work_packages/56244/activity

### DIFF
--- a/app/components/shares/user_details_component.html.erb
+++ b/app/components/shares/user_details_component.html.erb
@@ -5,7 +5,7 @@
         Primer::Beta::Link.new(
           font_weight: :semibold,
           href: principal_show_path,
-        ),
+        )
       ) { user.name }
     end
 
--- a/app/components/shares/user_details_component.html.erb
+++ b/app/components/shares/user_details_component.html.erb
@@ -1,67 +1,38 @@
-<%=
-  component_wrapper do
-    flex_layout do |flex|
-      flex.with_row do
-        render(Primer::Beta::Link.new(font_weight: :semibold, href: principal_show_path)) { user.name }
-      end
+<%= component_wrapper do
+  flex_layout do |flex|
+    flex.with_row do
+      render(
+        Primer::Beta::Link.new(
+          font_weight: :semibold,
+          href: principal_show_path,
+        ),
+      ) { user.name }
+    end
 
-      flex.with_row(classes: 'ellipsis') do
-        if manager_mode?
-          if user_is_a_group?
-            if project_group?
-              render(Primer::Beta::Text.new(color: :subtle)) { I18n.t("sharing.user_details.project_group")}
+    flex.with_row(classes: "ellipsis") do
+      if strategy.manageable?
+        if user_in_non_active_status?
+          if user.locked?
+            concat(render(Primer::Beta::Octicon.new(icon: :lock, color: :muted, mr: 1)))
+            concat(render(Primer::Beta::Text.new(color: :subtle)) { I18n.t("sharing.user_details.locked") })
+          elsif user.invited?
+            if invite_resent?
+              concat(render(Primer::Beta::Text.new(color: :subtle)) { I18n.t("sharing.user_details.invite_resent") })
             else
-              render(Primer::Beta::Text.new(color: :subtle)) { I18n.t("sharing.user_details.not_project_group")}
-            end
-          else
-            if user_in_non_active_status?
-              if user.locked?
-                concat(render(Primer::Beta::Octicon.new(icon: :lock, color: :muted, mr: 1)))
-                concat(render(Primer::Beta::Text.new(color: :subtle)) { I18n.t("sharing.user_details.locked") })
-              elsif user.invited?
-                if invite_resent?
-                  concat(render(Primer::Beta::Text.new(color: :subtle)) { I18n.t("sharing.user_details.invite_resent") })
-                else
-                  concat(render(Primer::Beta::Text.new(color: :subtle)) { I18n.t('sharing.user_details.invited') })
-                  concat(
-                    form_with(url: resend_invite_path, method: :post) do
-                      render(Primer::Beta::Button.new(type: :submit, px: 0, scheme: :link)) { I18n.t('sharing.user_details.resend_invite') }
-                    end
-                  )
-                end
-              end
-            else
-              if part_of_a_group?
-                if part_of_a_shared_group?
-                  if project_member?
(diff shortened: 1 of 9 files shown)


spec/components/shares/user_details_component_spec.rb
```diff
@@ -40,9 +40,16 @@
...
   shared_let(:user_lastname)  { "Hendricks" }
   shared_let(:group_name)     { "Cool group" }
+  shared_let(:strategy) do
+    SharingStrategies::WorkPackageStrategy.new(work_package, query_params: {})
+  end
 
   let(:invite_resent) { false }
 
+  before do
+    allow(strategy).to receive(:manageable?).and_return(manager_mode)
```
toy: I suggest renaming `manager_mode` to `manageable?`

app/models/sharing_strategies/work_package_strategy.rb
```diff
@@ -53,6 +53,38 @@ def viewable?
...
+          I18n.t("sharing.user_details.additional_privileges_project")
+        else
+          I18n.t("sharing.user_details.not_project_member")
+        end
+      elsif project_member?(share)
+        I18n.t("sharing.user_details.additional_privileges_project")
+      else
+        I18n.t("sharing.user_details.not_project_member")
+      end
+    end
```
toy: I suggest refactoring, for example all but one case share `I18n.t("sharing.user_details.")`

app/models/sharing_strategies/work_package_strategy.rb
```diff
@@ -64,5 +96,27 @@ def update_contract_class
     def delete_contract_class
       Shares::WorkPackages::DeleteContract
     end
+
+    private
+
+    def project_member?(share)
+      Member.exists?(project: share.entity.project, principal: share.principal, entity: nil)
```
toy, line 105: Does entity always has a project `share.entity.project`?

app/models/sharing_strategies/project_query_strategy.rb
```diff
@@ -88,6 +88,21 @@ def shares(reload: false)
...
+
+      if share.principal == entity.user
+        I18n.t("sharing.project_queries.user_details.owner")
+      elsif entity.public?
+        if share.principal.is_a?(User) && share.principal.allowed_globally?(:manage_public_project_queries)
+          I18n.t("sharing.project_queries.user_details.can_manage_public_lists")
+        elsif share.roles.any? { |role| role.builtin == Role::BUILTIN_PROJECT_QUERY_VIEW }
+          I18n.t("sharing.project_queries.user_details.can_view_because_public")
+        end
+      end
```
toy: `I18n.t("sharing.project_queries.user_details.…")` can be extracted

app/components/shares/user_details_component.html.erb
```diff
@@ -1,67 +1,38 @@
...
-        render(Primer::Beta::Link.new(font_weight: :semibold, href: principal_show_path)) { user.name }
-      end
+<%= component_wrapper do
+  flex_layout do |flex|
+    flex.with_row do
+      render(
+        Primer::Beta::Link.new(
+          font_weight: :semibold,
+          href: principal_show_path,
+        ),
```
toy:
```suggestion
        )
```

toy (changes requested):

app/models/sharing_strategies/work_package_strategy.rb
klaustopher: I have to concur here. This is already a super complicated tree of conditionals. But the way how it is designed now it is clear to follow a path and see what output it yields to. All optimizations like setting the i18n key to a variable and calling I18n.t with that lead to less readable code here.

I have tried that solution and it's actually worse to understand what is happening

app/models/sharing_strategies/work_package_strategy.rb
klaustopher: In the case of WorkPackages, yes.

app/models/sharing_strategies/project_query_strategy.rb
klaustopher: Same argument as the other case. It leads to less readable code if we somehow set a variable. It's already complicated enough as it is.

spec/components/shares/user_details_component_spec.rb
klaustopher: done

app/models/sharing_strategies/work_package_strategy.rb
toy: Hm, something like following seems easier to read (can't say for sure if guards are better than complete conditions):
```ruby
def share_description(share)
  key = share_description_key(share)
  return "" unless key

  I18n.t(key, scope: "sharing.user_details")
end

private

def share_description_key(share)
  if !manageable?
    if share.principal.invited?
      :invited
    else
      nil
    end
  elsif share.principal.is_a?(Group)
    if project_member?(share)
      :project_group
    else
      :not_project_group
    end
  elsif group_member?(share)
    if has_roles_via_group_membership?(share)
      if project_member?(share)
        :additional_privileges_project_or_group
      else
        :additional_privileges_group
      end
    elsif inherited_project_member?(share)
      :additional_privileges_project_or_group
    elsif project_member?(share)
      :additional_privileges_project
    else
      :not_project_member
    end
  elsif project_member?(share)
    :additional_privileges_project
  else
    :not_project_member
  end
end
```

klaustopher: I don't find that any more readable.

- We are seperating the generation of the translation key into multiple steps so it becomes harder to figure out what the final key will be.
- We have the overhead of adding another method
- We have to deal with the `key.nil?` edge case where we don't want a translation

If you 100% insist, that your version is better, I will change it, but functionally we are doing exactly the same thing, but each of us just thinks their personal style is better than the other.

app/models/sharing_strategies/project_query_strategy.rb
toy: In this case it is much less code, but still to me extracting variable to remove repetition sounds good, the only downside I imagine is that it is not possible to find occurrence by complete key.

Also either this should explicitly return `""` in else cases or the guard on the second line should return nil (explicitly or implicitly).

app/models/sharing_strategies/work_package_strategy.rb
toy: My point is removal of duplication, so that I don't need to read through prefixes to know that they are the same and there is no typo. I understand not wanting another method especially if it is private and should be separately in private section, then it can be in the same method.
Overall from my perspective having just the meaningful part different in the condition tree makes it easier to read.

toy (approved):

PR #16100, opened 2024-07-11
Feature/55233 Select/Exclude projects in project list explicitly via filter

toy:
[OP#55233](https://community.openproject.org/projects/openproject/work_packages/55233)

Prefer to see the diff with whitespace difference hidden.

image

### DIFF
--- a/app/models/queries/projects/filters/ancestor_filter.rb
+++ b/app/models/queries/projects/filters/ancestor_filter.rb
@@ -26,69 +26,63 @@
 # See COPYRIGHT and LICENSE files for more details.
 #++
 
-module Queries
-  module Projects
-    module Filters
-      class AncestorFilter < ::Queries::Projects::Filters::Base
-        def apply_to(_query_scope)
-          case operator
-          when "="
-            super
-              .where(exists_condition.exists)
-          when "!"
-            super
-              .where.not(exists_condition.exists)
-          else
-            raise "unsupported operator"
-          end
-        end
-
-        def where
-          nil
-        end
-
-        def type
-          :list
-        end
-
-        def self.key
-          :ancestor
-        end
-
-        private
-
-        def type_strategy
-          # Instead of getting the IDs of all the projects a user is allowed
-          # to see we only check that the value is an integer.  Non valid ids
-          # will then simply create an empty result but will not cause any
-          # harm.
-          @type_strategy ||= ::Queries::Filters::Strategies::IntegerList.new(self)
-        end
-
-        def exists_condition
-          Project.from("#{Project.table_name} ancestors")
-                 .where(ancestor_condition.and(ancestor_in_values_condition))
-                 .arel
-        end
-
-        def ancestor_condition
-          projects_table[:lft]
-            .gt(projects_ancestor_table[:lft])
-            .and(projects_table[:rgt].lt(projects_ancestor_table[:rgt]))
-        end
-
-        def ancestor_in_values_condition
-          projects_ancestor_table[:id].in(values)
-        end
-
-        def projects_table
-          Project.arel_table
-        end
-
-        def projects_ancestor_table
-          projects_table.alias(:ancestors)
-        end
-      end
+class Queries::Projects::Filters::AncestorFilter < Queries::Projects::Filters::Base
+  def apply_to(_query_scope)
+    case operator
+    when "="
+      super
+        .where(exists_condition.exists)
+    when "!"
+      super
+        .where.not(exists_condition.exists)
+    else
+      raise "unsupported operator"
     end
+  end
+
+  def where
+    nil
+  end
+
+  def type
+    :list
+  end
+
+  def self.key
+    :ancestor
+  end
+
+  private
+
+  def type_strategy
+    # Instead of getting the IDs of all the projects a user is allowed
+    # to see we only check that the value is an integer.  Non valid ids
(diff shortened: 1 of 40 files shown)


app/models/queries/projects/filters/id_filter.rb
```diff
@@ -26,8 +26,12 @@
...
-class Queries::Projects::Filters::IdFilter < Queries::Projects::Filters::ProjectFilter
-  def type
-    :integer
+class Queries::Projects::Filters::IdFilter < Queries::Projects::Filters::Base
+  include Queries::Filters::Shared::ProjectFilter::Required
+
+  def self.key = :id
+
+  def human_name
+    I18n.t(:label_project, default: :"attributes.project")
```
toy, line 35: Not sure whether it is better to use a default that exists in translations

app/models/queries/projects/filters/ancestor_filter.rb
```diff
@@ -26,69 +26,63 @@
...
+    when "="
+      super
+        .where(exists_condition.exists)
+    when "!"
+      super
+        .where.not(exists_condition.exists)
+    else
+      raise "unsupported operator"
+    end
+  end
```
aaron-contreras, line 41: Out of curiosity, why/what changes when you use this syntax as compared to the former nested modules style?

app/models/queries/projects/filters/id_filter.rb
aaron-contreras: I didn't get what you mean by this

config/locales/en.yml
```diff
@@ -2311,6 +2311,7 @@ en:
   label_product_version: "Product version"
   label_profile: "Profile"
   label_percent_complete: "% Complete"
+  label_project: "Project"
```
aaron-contreras, line 2314: I see the translation key was added, so why would the default be required/what does it bring with it?

spec/support/queries/filters/shared_filter_examples.rb
```diff
@@ -35,7 +35,13 @@
     described_class.create!(name: instance_key, context:, operator:, values:)
   end
   let(:name) { model.human_attribute_name((instance_key || expected_class_key).to_s.gsub("_id", "")) }
-  let(:model) { WorkPackage }
+  let(:model) do
+    if /^Queries::(?<model_plural>(?:[A-Z][a-z]+)+)::Filters?::(?:[A-Z][a-z]+)+Filter$/ =~ described_class.name
+      model_plural.singularize.constantize
+    else
+      raise "needs to be defined"
+    end
```
aaron-contreras, line 43: Neat!

aaron-contreras (approved):
Works great! The comments I left are to understand the why behind it.

Good job! 👍

app/models/queries/projects/filters/ancestor_filter.rb
klaustopher: With the chained syntax all the modules need to already exist:

```ruby
>> class Foo::Bar::Baz; end
(irb):1:in `<main>': uninitialized constant Foo (NameError)

>> module Foo; module Bar; end; end;
>> class Foo::Bar::Baz; end
=> nil
```

spec/support/queries/filters/shared_filter_examples.rb
klaustopher: Personally, I think we should get rid of ✨ magic 🪄 like this and have things like this be explicitly defined so that we don't have to guess what model is used but it is explicitly stated in the test. But that's not an issue of this PR and should be followed up in a Code Maintenance ticket.

klaustopher (approved):
good job!

app/models/queries/projects/filters/ancestor_filter.rb
toy: @aaron-contreras Main difference is constant resolution - in compact mode after current scope constant resolution will proceed right away to global scope, and in nested it will use every nested scope on the way

@klaustopher True in Ruby, but not when rails is involved - [zeitwerk handles implicit namespaces](https://github.com/fxn/zeitwerk?tab=readme-ov-file#implicit-namespaces)

app/models/queries/projects/filters/id_filter.rb
toy: This is exactly about your other question. I know that `attributes.project` is an existing meaningful translation, but it is not the right key, so I created the right key and not sure if I should default to an existing translation or not.

And now I think that default should be dropped, as crowdin AFAIK helps when keys are copied/renamed and leaving default is a bit of maintenance to remove it that will be forgotten and is not worth a ticket.

config/locales/en.yml
toy: Answered for app/models/queries/projects/filters/id_filter.rb

spec/support/queries/filters/shared_filter_examples.rb
toy: I first removed the `let` and started fixing the missing `model` definitions, than realised that it is wrong in at least some places, like for example modules/storages/spec/models/queries/storages/projects/filter/storages_filter_spec.rb should have `Project`, but spec was green with `WorkPackage`.
So a good idea to have a look at usages of all the shared filter examples

app/models/queries/projects/filters/id_filter.rb
toy: #16222

PR #16104, opened 2024-07-11
[#56334] introduced request url build and add test coverage

Kharonus:
[#56334](https://community.openproject.org/work_packages/56334)

### WAT?

- different ways of constructing URLs and paths has led to many bugs in the past
- this PR introduces a builder for urls and paths for file storages
  - it is meant to be the only used tool for construction of those urls
  - it has a clearly documented interface (due to its test)
    - it consumes unescaped(!) fragments and it produces escaped urls and paths
    - it does not preserve any slash delimiters, but normalizes the whole path (meaning trailing slashes get removed, intermediate multiple slashes are reduced to a single one)
- current PR only replaces nextcloud usage (one drive for a single PR due to reviewability)


modules/storages/app/common/storages/peripherals/storage_interaction/request_url_builder.rb
Kharonus, line 1: 🏳️ 

This file is the newly introduced url and path factory. It has a related unit tests to show its usage and expected behavior.

modules/storages/app/common/storages/peripherals/storage_interaction/nextcloud/util.rb
Kharonus, line 1: 🏳️

Please note, that explicitly the factory was introduced to replace those methods.

mereghost (approved):
Nice. That was a ton of work! I think the tests are extensive enough that it won't raise too many issues.

modules/storages/app/common/storages/peripherals/storage_interaction/request_url_builder.rb
```diff
@@ -0,0 +1,77 @@
...
+          def build(storage, *path_fragments)
+            host_uri = storage.uri
+
+            URI.join(host_uri.origin,
+                     ensure_sub_path(host_uri.path),
+                     *split_and_escape(path_fragments))
+               .to_s
+          end
+
+          def path(*path_fragments)
```
ba1ash: I think the name is not great. It is a noun, so seems like a property of object, but it is not called on an object. I suggest to have the name to be an action describing thing like: `escape_and_join`, `join`, `join_paths`. Something of that kind. What do you think?

modules/storages/app/common/storages/peripherals/storage_interaction/request_url_builder.rb
```diff
@@ -0,0 +1,77 @@
...
+#
+# See COPYRIGHT and LICENSE files for more details.
+#++
+
+module Storages
+  module Peripherals
+    module StorageInteraction
+      class RequestUrlBuilder
+        class << self
+          def build(storage, *path_fragments)
```
ba1ash: Are you sure about storage as a method parameter/dependency here? Why not to have uri passed here? Then function could be more general and reused in other parts of OP potentially.

modules/storages/app/components/storages/admin/forms/automatically_managed_project_folders_form_component.rb
```diff
@@ -94,8 +94,8 @@ def storage_provider_credentials_copy_instructions
     def provider_credentials_instructions_link
       render(
         Primer::Beta::Link.new(
-          href: Storages::Peripherals::StorageInteraction::Nextcloud::Util.join_uri_path(storage.host,
-                                                                                         "settings/admin/openproject"),
+          href: Storages::Peripherals::StorageInteraction::RequestUrlBuilder.build(storage,
```
ba1ash: I am not suggesting to do it here.
In my opinion the following namespace is absolutely sufficient from code user perspective. Therefore let's remove this two levels of nesting `::Peripherals::StorageInteraction`. Wdyt?
```suggestion
          href: Storages::RequestUrlBuilder.build(storage,
```

modules/storages/app/common/storages/peripherals/storage_interaction/request_url_builder.rb
Kharonus: I agree, we can pass the URI directly in here. My first intention was to tie it to a storage for usage.

but, after using it, I had to add a builder for relative paths. Those are not tight to the storage. So, better to loosen the tie and pass in URIs.

modules/storages/app/common/storages/peripherals/storage_interaction/request_url_builder.rb
Kharonus: I disagree with a noun being a bad method name. We have a ton of methods using nouns, because they do return a specific entity, object or constructed representation of a domain thing. And the noun is the description of that. Common examples:

- location
- uri
- header
- success
- failure

I'd like to keep the name, better yet, I'd like to rename `build` to `url` to make it consistent and more distinguishable.

`RequestUrlBuilder.url()`
`RequestUrlBuilder.path()`

modules/storages/app/components/storages/admin/forms/automatically_managed_project_folders_form_component.rb
Kharonus: Hmm, I don't understand. Removing those namespace identifiers will lead to the `RequestUrlBuild` not being found?

modules/storages/app/common/storages/peripherals/storage_interaction/request_url_builder.rb
ba1ash: It is still feels somehow uneasy to me, but there is no problem. Let's have it like that :+1:

modules/storages/app/components/storages/admin/forms/automatically_managed_project_folders_form_component.rb
ba1ash: Sorry for not being clear enough. I meant removing this 2 levels entirely throughout our application, because they do not much meaning, but using them makes names enormous. And wanted to know your opinion about that. Do they bring value to you?

Kharonus: The namespace means value for me due to its context.

e.g. a `Utils` class within `StorageInteraction::Nextcloud` means it is meant to be used there. Hence, when long namespace definitions like here are seen, then this is a good indicator, that the context is wrong. Which is, what you raised here.

So, When I was planning this url build, I opted for the commands/queries. When I replaced the previous utils method, I noticed, that we need this functionality outside, too. (here in the form components). With this information I agree with you, we should move it up.

PR #16106, opened 2024-07-12
No need right padding as spot-icon_dropdown takes 24px already.

Eric-Guo:
The purpose of PR is remove extra right padding for toolbar button `Create`, `Include projects`, `Baseline` and `Filter`, so given more space for query name.

#### Before:

Old-EN_2024-07-12_11 13 36

Old-CN_2024-07-12_11 13 08

#### After:

New-EN_2024-07-12_11 11 31

New-CN_2024-07-12_11 10 14

### DIFF
--- a/frontend/src/global_styles/layout/_toolbar.sass
+++ b/frontend/src/global_styles/layout/_toolbar.sass
@@ -159,6 +159,9 @@ $nm-color-success-background: #d8fdd1
     @include toolbar-element-sizing
     border: 1px solid var(--button-default-borderColor-rest)
 
+    &.-no-padding
+      padding-right: 0
+
   .button,
   .toolbar-input--affix
     background: var(--button-default-bgColor-rest)
--- a/frontend/src/global_styles/layout/_toolbar.sass
+++ b/frontend/src/global_styles/layout/_toolbar.sass
@@ -159,7 +159,7 @@ $nm-color-success-background: #d8fdd1
     @include toolbar-element-sizing
     border: 1px solid var(--button-default-borderColor-rest)
 
-    &.-no-padding
+    &.-has-dropdown
       padding-right: 0
 
   .button,
--- a/frontend/src/app/features/work-packages/components/wp-baseline/baseline-modal/baseline-modal.component.html
+++ b/frontend/src/app/features/work-packages/components/wp-baseline/baseline-modal/baseline-modal.component.html
@@ -6,7 +6,7 @@
   <button
     slot="trigger"
     type="button"
-    class="button"
+    class="button -no-padding"
     [class.-active]="opened"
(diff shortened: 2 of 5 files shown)


frontend/src/global_styles/layout/_toolbar.sass
```diff
@@ -167,6 +167,9 @@ $nm-color-success-background: #d8fdd1
     @include toolbar-element-sizing
     border: 1px solid var(--toolbar-item--border-color)
 
+    &.-no-padding
+      padding-right: 0
```
HDinger, line 163: I see multiple issues with such a rule:

* A class named `-no-padding` should remove all padding not only the right side.
* Such generic class are usually helpful for other places as well and should not be limited to the toolbar only.

HDinger (changes requested):
Hi @Eric-Guo 

thanks for this PR. I have to admit, that buttons look rather squeezed to me.. 

We already have a mechanism to wrap the buttons in a separate line on smaller screens. So I'd vote to rather have a look whether we can improve and let it wrap earlier than squeezing the buttons to a minimum.
Bildschirmfoto 2024-07-17 um 12 51 37

Eric-Guo:
> That buttons look rather squeezed to me..

Actually it's same as previous design of wp-fold-toggle-button (before this PR), which icon-pulldown have no padding:

iShot_2024-07-19_10 01 01

iShot_2024-07-19_10 03 00

So this PR is just make sure there is no extra padding for other icon-pulldown:

iShot_2024-07-19_10 02 05

frontend/src/global_styles/layout/_toolbar.sass
Eric-Guo: I checked other place, seems it's due to using different `icon_dropdown` and `spot-icon spot-icon_dropdown` lead such problem, I rename to more BEM style like, I have to admin it's not perfect, but at least better than No fix the spacing problem.

Feel free to propose a better solution for me, I'm more than glad to close this PR.

HDinger:
Hi @Eric-Guo 

I still think that instead of pushing the last bits out of the buttons we should rather let the whole toolbar wrap much earlier. With the amount of buttons that we have there now, we have to accept that the buttons have to move to a new line if there is not enough space. Maybe a `min-width` for the title would already do the trick.

piradata:
i think both can be bone, that padding on the dropdown icons really is unnecessary, and we could use two logic for the toolbar wrap, one just for the view size, and the other if the text needs to be shortened to fit

piradata:
i think both can be bone, that padding on the dropdown icons really is unnecessary, and we could use two logic for the toolbar wrap, one just for the view size, and the other if the text needs to be shortened to fit

PR #16115, opened 2024-07-12
[#55968] Implement project removal with warning dialog before removing the project

judithroth:
https://community.openproject.org/work_packages/55968


modules/storages/app/controllers/storages/admin/storages/project_storages_controller.rb
```diff
@@ -73,10 +72,24 @@ def create
     respond_with_turbo_streams(status: create_service.success? ? :ok : :unprocessable_entity)
   end
 
-  def destroy; end
+  def destroy_confirmation_dialog
+    respond_with_dialog Storages::ProjectStorages::DestroyConfirmationDialogComponent.new(
```
akabiru, line 77: 🟢 Q: Do we still need an async dialog for this use case? Given the use case does not require streamable forms. A "vanilla" dialog would perhaps be simpler and more fitting

modules/storages/app/controllers/storages/admin/storages/project_storages_controller.rb
```diff
@@ -73,10 +72,24 @@ def create
...
+      storage: @storage,
+      project_storage: @project_storage
+    )
+  end
+
+  def destroy
+    @project_storage.destroy
```
akabiru, line 84: 🟠 There's an edge case where a user can attempt to unlink a project from different tabs- in which case they would be met by an unhandled 404- in project attributes, we handled this via a before action [here](https://github.com/opf/openproject/blob/dev/app/controllers/admin/settings/project_custom_fields_controller.rb#L186-L189) shall we extend the same?

----


https://github.com/user-attachments/assets/37fbee7f-5817-409e-a5cd-e391ad6184c4

modules/storages/app/components/storages/project_storages/destroy_confirmation_dialog_component.html.erb
```diff
@@ -0,0 +1,86 @@
...
+
+++#%>
+
+<%=
+  primer_form_with(
+    model: @project_storage,
+    url: admin_settings_storage_project_storage_path(id: @project_storage),
+    data: {
+      turbo: true,
+      controller: 'disable-when-checked',
```
akabiru, line 36: Nice one, I didn't know we had this already 👍🏾

modules/storages/spec/components/storages/project_storages/destroy_confirmation_dialog_component_spec.rb
```diff
@@ -0,0 +1,75 @@
...
+    context "for a project with unmanaged folder" do
+      it "includes only the basic hint" do
+        project_storage = build_stubbed(:project_storage, storage:, project_folder_mode: "inactive")
+        component = described_class.new(storage:, project_storage:)
+        expect(component.text).to include("irreversible")
+        expect(component.text).not_to include("files will be deleted forever")
+      end
+    end
+  end
+end
```
akabiru, line 75: ✨

modules/storages/spec/components/storages/project_storages/destroy_confirmation_dialog_component_spec.rb
```diff
@@ -0,0 +1,75 @@
...
+      project_storage = build_stubbed(:project_storage, storage:)
+
+      component = described_class.new(storage:, project_storage:)
+      expect(component.heading).to include("OneDrive/SharePoint")
+      expect(component.heading).not_to include("Nextcloud")
+    end
+  end
+
+  describe "#text" do
+    let(:storage) { create(:storage, :as_generic) }
```
akabiru: ?
```suggestion
    let(:storage) { build_stubbed(:storage, :as_generic) }
```

akabiru (changes requested):
Looking good! Just a few Qs and requests for changes

modules/storages/app/controllers/storages/admin/storages/project_storages_controller.rb
akabiru: See https://github.com/opf/openproject/pull/15653

judithroth: Thanks for pointing that out!

modules/storages/app/controllers/storages/admin/storages/project_storages_controller.rb
judithroth: In general I agree with you and think from the contents of the dialog it does not need to be async. However, I didn't want to change the way the `action_menu` of a table row works and this currently only works with links. So as long as there is no pressing reason why the dialog should be embedded in the page, I'd avoid that refactoring to be honest.

akabiru (approved):

PR #16125, opened 2024-07-15
[#56160] 14.3 Documentation: Public, shared and favorite project list…

MayaBerd:
…s  https://community.openproject.org/work_packages/56160


docs/user-guide/projects/project-lists/README.md
```diff
@@ -13,12 +13,15 @@ In OpenProject you can create projects to collaborate with your team members, tr
...
-| [Export project list](#export-project-lists)                | You can export the project list to XLS or CSV.               |
-| [Project overarching reports](#project-overarching-reports) | How to create project overarching reports across multiple projects. |
+| Topic                                                        | Content                                                      |
+| ------------------------------------------------------------ | ------------------------------------------------------------ |
+| [Select project lists](#select-project-lists)                | Get an overview of all your projects in the project lists.   |
+| [Filter project lists](#project-lists-filters)               | Adjust filters in the project lists.                         |
+| [Favorite project lists](#favorite-project-lists)            | Mark project lists as favorite.                              |
+| [Share project lists](#share-project-lists-with-users-and-groups-enterprise-add-on) | Share project lists with users and groups.                   |
+| [Public project lists](#share-with-everyone-at-openproject)  | Share project lists with everyone at OpenProject.            |
```
marcalcobe: Hi @MayaBerd, we have been using the term share with everyone in your instance for that instead of public lists to avoid the confusion between sharing inside of an instance and making it public even outside of your instance (users that are not even logged in).

marcalcobe: As we discussed also in chat, I think we should get rid of the term public and start using "Share with everyone" or something similar, and I would also change the explanation to "Share project lists with everyone at your instance."

docs/user-guide/projects/project-lists/README.md
```diff
@@ -94,14 +95,20 @@ Projects can be filtered in OpenProject. The default view will list all currentl
 
 **Archived projects** - returns all projects that are not active, of which you were a member or have the right to see.
 
+**Public project lists** - shows all project lists that are public, i.e. that were shared with everyone within an OpenProject instance.
```
marcalcobe: @MayaBerd this entry will not exist as soon as the bug **56444** gets fixed because all the shared lists independent if they are shared individually or with everyone will go bellow "Shared project lists".

docs/user-guide/projects/project-lists/README.md
```diff
@@ -94,14 +95,20 @@ Projects can be filtered in OpenProject. The default view will list all currentl
...
+
 **My private project lists** - shows all the project lists that you have customized and saved.
 
+**Shared project lists** - shows all the project lists that were shared with you.
```
marcalcobe: @MayaBerd here we will need to add "that were shared with you or you shared with others."

docs/user-guide/projects/project-lists/README.md
```diff
@@ -148,6 +155,46 @@ By selecting *Overall activity* you can open a view in which all the latest glob
 
 ![overall activity meeting filter](actvity-global-filter.png)
 
+## Favorite project lists
+
+You can mark your private project lists as favorites. This includes shared, public and private project lists, but not the static ones, e.g. *Active projects* or *Archived projects*. To mark a project list as favorite click the star icon in the top right corner.
```
marcalcobe: @MayaBerd I would reformulate this as something like "You can mark project lists as favourites, both shared and private, but not the static ones..."

docs/user-guide/projects/project-lists/README.md
```diff
@@ -148,6 +155,46 @@ By selecting *Overall activity* you can open a view in which all the latest glob
...
+
+The star will turn yellow and the favorite project list will move to the top of the list within the respective sidebar section. If multiple project lists are favored, they will be listed alphabetically.
+
+![Favorite project list in OpenProject](star-yellow-project-list.png)
+
+You can remove the star by clicking on the star icon again. 
+
+## Share project lists 
+
+You can share project lists in OpenProject, except the static project lists, e.g. *Active projects* or project lists under *Project status* section in the lefthand menu. Project lists can either be made public, i.e. shared with everyone at OpenProject or with specific users or groups.
```
marcalcobe: @MayaBerd lets remove the public concept also from here and keep the "shared with everyone in your instance or with specific users or groups."

docs/user-guide/projects/project-lists/README.md
```diff
@@ -148,6 +155,46 @@ By selecting *Overall activity* you can open a view in which all the latest glob
...
+
+### Share with everyone at OpenProject
+
+You can share a project list with everyone at OpenProject, thus making it public. This means that the project list will become visible to all the users registered on your OpenProject instance.
```
marcalcobe: @MayaBerd same here, you are not really making it public. And the way users will see this option is always "Share project list with everyone at {instance name}" so maybe we should say "...with everyone in your instance" instead.

docs/user-guide/projects/project-lists/README.md
```diff
@@ -148,6 +155,46 @@ By selecting *Overall activity* you can open a view in which all the latest glob
...
+
+## Share project lists 
+
+You can share project lists in OpenProject, except the static project lists, e.g. *Active projects* or project lists under *Project status* section in the lefthand menu. Project lists can either be made public, i.e. shared with everyone at OpenProject or with specific users or groups.
+
+### Share with everyone at OpenProject
+
+You can share a project list with everyone at OpenProject, thus making it public. This means that the project list will become visible to all the users registered on your OpenProject instance. 
+
+To do that click the **Share** icon and turn on the **Share with everyone at OpenProject** switch.
```
marcalcobe: @MayaBerd The switch name will not always be "OpenProject" this last bit will be the real name of each instance.

docs/user-guide/projects/project-lists/README.md
```diff
@@ -148,6 +155,46 @@ By selecting *Overall activity* you can open a view in which all the latest glob
...
+
+![Share a project list with everyone in OpenProject](project-lists-share-with-everyone-button.png)
+It will be added to the menu on the left under *Public project lists*.
```
marcalcobe: @MayaBerd This section will not exist with the fix of **56444**. It will be added under "Shared project lists"

docs/user-guide/projects/project-lists/README.md
```diff
@@ -148,6 +155,46 @@ By selecting *Overall activity* you can open a view in which all the latest glob
...
+> Sharing project lists with users and groups is an Enterprise add-on and will only be displayed here for Enterprise on-premises and Enterprise cloud.
+
+You can share private project lists with specific users or groups in OpenProject. To do that navigate to a project list and click the **Share** icon. Then specify a user or a group using the search field, define whether they can only view or edit a project list and finally click the **Share** button. That user or group will see that shared project list under **Shared project lists**.  
+
+If a project list has already been shared, you will see the list of users that have access to the project list, including the list owner. You will can modify or revoke project list sharing privileges at any time.
+
+![Share project lists with users or groups in OpenProject](projects-list-share-enterprise.png)
+
+> [!NOTE]
+> Individual project permissions will override project lists. That means that within the project lists users will only see the projects that they already have access to.
```
marcalcobe: @MayaBerd the fact of which projects a user can see and which no its related with access to that project not to the permissions of that user on that project. Therefore, I would change the sentence to what we used in the info banner inside the modal "Users will only see the projects they have access to. Sharing project lists does not impact individual project permissions." or something similar.

marcalcobe (changes requested):
Hi @MayaBerd I've added comments to the review. Most of them are related to the fact that the term "Public project lists" should not really exist and should not have a separated entry menu in the sidebar. A [BUG](https://community.openproject.org/projects/openproject/work_packages/56444/activity) has been reported already in regards to this topic and will be fix soon, maybe it make sense to wait until the BUG is resolved to rework on the docs?

docs/user-guide/projects/project-lists/README.md
```diff
@@ -176,19 +176,22 @@ You can share a project list with everyone within your OpenProject instance. Thi
...
 
 ![Share a project list with everyone in OpenProject](project-lists-share-with-everyone-button.png)
-It will be added to the menu on the left under *Shared project lists*.
+The project list will move from *My project lists* section to *Shared project lists* section in the lefthand menu.
 
-![Public project lists in OpenProject]()
+![Public project lists in OpenProject](project-lists-share-with-everyone-list.png)
 
-UPDATE SCREENSHOT
+You can reverse the action by unselecting the star icon. The project list will return to *My project lists*.
```
marcalcobe: @MayaBerd I think this is not correct. Unselecting the star icon will not bring the list back to "My project lists". You need to switch off the share with everyone toggle for that.

docs/user-guide/projects/project-lists/README.md
```diff
@@ -176,19 +176,22 @@ You can share a project list with everyone within your OpenProject instance. Thi
...
 
-UPDATE SCREENSHOT
+You can reverse the action by unselecting the star icon. The project list will return to *My project lists*. 
+
+> ![TIP]
+> Using this function requires a *Manage public project lists* permission. This permission is automatically activated for administrators. If you want to grant this permission to other users, we recommend creating a [global role](../../../system-user-guide/users-permissions/roles-permissions/#global-role) to assign this permission.
 
 ### Share project lists with users and groups (Enterprise add-on)
 > [!NOTE] 
 > Sharing project lists with users and groups is an Enterprise add-on and will only be displayed here for Enterprise on-premises and Enterprise cloud.
```
marcalcobe, line 190: I would probably add the word "individual" in front of users here.

docs/user-guide/projects/project-lists/README.md
```diff
@@ -19,7 +19,7 @@ Your projects can be available publicly or internally. OpenProject does not limi
 | [Filter project lists](#project-lists-filters)               | Adjust filters in the project lists.                         |
 | [Favorite project lists](#favorite-project-lists)            | Mark project lists as favorite.                              |
 | [Share project lists with users and groups (Enterprise add-on)](#share-project-lists-with-users-and-groups-enterprise-add-on) | Share project lists with users and groups.                   |
```
marcalcobe: I would probably add the word "individual" in front of users here.

marcalcobe (changes requested):
Hi @MayaBerd,

I've added a couple of details to double check.

docs/user-guide/projects/project-lists/README.md
MayaBerd: @marcalcobe you are right! i mixed those up, will correct.

PR #16128, opened 2024-07-15
[#51847] added server-url to openapi-explorer

Kharonus:
[#51847](https://community.openproject.org/wp/51847)

### WAT?

- overwrite server url with real root path (including any path prefixes)


machisuji (approved):
👍

PR #16136, opened 2024-07-15
Allow specifiying a name for date picker names

klaustopher:
Closes https://community.openproject.org/projects/openproject/work_packages/56033

### DIFF
--- a/spec/lib/tabular_form_builder_spec.rb
+++ b/spec/lib/tabular_form_builder_spec.rb
@@ -279,7 +279,6 @@ JJ Abrams</textarea>
     it_behaves_like "wrapped in field-container by default"
 
     it "outputs element" do
-      puts output
       expect(output).to have_css "op-basic-single-date-picker"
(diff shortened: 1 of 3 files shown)


spec/lib/tabular_form_builder_spec.rb
```diff
@@ -268,6 +268,24 @@
...
+
+    subject(:output) do
+      builder.date_picker :last_login_on, options
+    end
+
+    it_behaves_like "labelled by default"
+    it_behaves_like "wrapped in field-container by default"
+
+    it "outputs element" do
+      puts output
```
aaron-contreras:
```suggestion
```

aaron-contreras (approved):
Nice find - just a leftover `puts` that needs cleanup

spec/lib/tabular_form_builder_spec.rb
klaustopher: woops, forgot that one. Good 👀

PR #1050, opened 2014-03-25
Meets Bug #3148: Comparison of history versions in wiki pages ignores newlines

netfighter:
```
* `#3148` Fix: Comparison of history versions in wiki pages ignores newlines
```

PR #16152, opened 2024-07-16
[55263] Fix layout issues when diffing wiki pages

nsalguero:
https://community.openproject.org/projects/openproject/work_packages/55263/activity

The nl2br() function is not needed and there are some incorrect generated HTML tags.

### DIFF
--- a/app/controllers/wiki_controller.rb
+++ b/app/controllers/wiki_controller.rb
@@ -265,6 +265,9 @@ class WikiController < ApplicationController
         helpers.format_text(@diff.content_from.data.text, disable_macro_expansion: true),
         helpers.format_text(@diff.content_to.data.text, disable_macro_expansion: true)
       ).build
+       .gsub(/<ins class="diff(ins|mod)">\n\r?<\/ins>/, '')
+       .gsub(/<del class="diff(del|mod)">\n\r?<\/del>/, '')
+       .gsub(/^<figure class="op-uc-figure">(<figure class="table op-uc-figure_align-center op-uc-figure">)/, '\1')
     else
       render_404
     end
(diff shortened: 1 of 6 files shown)


app/controllers/wiki_controller.rb
```diff
@@ -265,6 +265,9 @@ def diff
         helpers.format_text(@diff.content_from.data.text, disable_macro_expansion: true),
         helpers.format_text(@diff.content_to.data.text, disable_macro_expansion: true)
       ).build
+       .gsub(/<ins class="diff(ins|mod)">\n\r?<\/ins>/, '')
+       .gsub(/<del class="diff(del|mod)">\n\r?<\/del>/, '')
```
cbliard: Could you extract this in a dedicated method describing what's being done?

Something like 
```ruby
def remove_empty_newline_highlights(diff)
  diff.gsub(/<ins class="diff(ins|mod)">\n\r?<\/ins>/, "")
      .gsub(/<del class="diff(del|mod)">\n\r?<\/del>/, "")
end

File: app/controllers/wiki_controller.rb [hunk 2]
```diff
@@ -265,6 +265,9 @@ def diff
...
+       .gsub(/^<figure class="op-uc-figure">(<figure class="table op-uc-figure_align-center op-uc-figure">)/, '\1')
```
cbliard: I'm not sure to understand what this `.gsub(/^<figure ...` is doing. Could you explain please? Or explain how to reproduce the issue this code is fixing?

cbliard (changes requested):
Thank you for the proposed improvement. The wiki diff definitely reads better with your adjustements.

I saw that the `nl2br` method was introduced 10 years ago in PR #1050 but I can't find what it was supposed to fix anymore. I tried playing with it but could not find what could go wrong with the changes you introduced.

I just have a request to extract some code in methods to better reveal its intent. Then it will be good to merge.

cbliard:
@nsalguero I extracted the diffing logic in a class so that it can get tests. I added a test for the newlines removal. You can now add one for the `<figure>` replacement thing.

PR #16155, opened 2024-07-16
[55803] Allow free input of status % complete in administration

cbliard:
https://community.openproject.org/wp/55803

In Administration > Work packages > Statuses, when creating or editing a status, the % complete field has been replaced by an input field. Before it was a select with 10% steps which made it impossible to set values like "45%".

### DIFF
--- a/app/views/statuses/_form.html.erb
+++ b/app/views/statuses/_form.html.erb
@@ -31,12 +31,7 @@ See COPYRIGHT and LICENSE files for more details.
 
 <section class="form--section">
   <div class="form--field -required"><%= f.text_field "name", required: true, container_class: "-middle" %></div>
-  <div class="form--field">
-    <%= f.select :default_done_ratio,
-                 (0..100).step(10).map { |r| ["#{r} %", r] },
-                 container_class: "-xslim",
-                 label: WorkPackage.human_attribute_name(:done_ratio) %>
-  </div>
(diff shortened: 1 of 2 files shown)


app/views/statuses/_form.html.erb
```diff
@@ -31,12 +31,7 @@ See COPYRIGHT and LICENSE files for more details.
 
 <section class="form--section">
   <div class="form--field -required"><%= f.text_field "name", required: true, container_class: "-middle" %></div>
-  <div class="form--field">
-    <%= f.select :default_done_ratio,
-                 (0..100).step(10).map { |r| ["#{r} %", r] },
-                 container_class: "-xslim",
-                 label: WorkPackage.human_attribute_name(:done_ratio) %>
-  </div>
+  <div class="form--field"><%= f.number_field "default_done_ratio", min: 0, max: 100, container_class: "-tiny" %></div>
```
dombesz: The input field is too small in Firefox MacOS. Due to the stepper arrows, it doesn't fit 2 digit numbers.

image

dombesz (changes requested):

app/views/statuses/_form.html.erb
cbliard: good catch, I reduced it to a minimum, but did not check with all browsers. I'll fix that.

cbliard: @dombesz It should be fixed now. Can you test it again please?

dombesz (approved):
Works as expected, thanks!

PR #16161, opened 2024-07-16
[#56334] introduced url builder for all one drive commands/queries

Kharonus:
[#56334](https://community.openproject.org/work_packages/56334)

### WAT?
- construction of request URLs and paths now based on common url builder class
- removed util function to join and escape paths

### DIFF
--- a/modules/storages/app/common/storages/peripherals/storage_interaction/one_drive/download_link_query.rb
+++ b/modules/storages/app/common/storages/peripherals/storage_interaction/one_drive/download_link_query.rb
@@ -49,7 +49,7 @@ module Storages
             end
 
             Authentication[auth_strategy].call(storage: @storage) do |http|
-              handle_errors http.get(Util.join_uri_path(@storage.uri, uri_path_for(file_link.origin_id)))
+              handle_errors http.get(url_for(file_link.origin_id))
             end
           end
 
@@ -78,8 +78,8 @@ module Storages
             end
           end
 
-          def uri_path_for(file_id)
-            "/v1.0/drives/#{@storage.drive_id}/items/#{file_id}/content"
+          def url_for(file_id)
+            UrlBuilder.url(Util.drive_base_uri(@storage), "items", file_id, "content")
           end
         end
       end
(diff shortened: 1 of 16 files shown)


modules/storages/app/common/storages/peripherals/storage_interaction/one_drive/download_link_query.rb
```diff
@@ -78,8 +78,8 @@ def handle_errors(response)
             end
           end
 
-          def uri_path_for(file_id)
-            "/v1.0/drives/#{@storage.drive_id}/items/#{file_id}/content"
+          def url_for(file_id)
+            UrlBuilder.url(Util.drive_base_uri(@storage), "items", file_id, "content")
```
ba1ash, line 82: A path was returned by the method before. Is it fine to return a uri now?

ba1ash (approved):
New `UrlBuilder.url` and `UrlBuilder.path` look really nice after all! :+1: 
In a couple of places uri paths are replaced with full uri. I guess it is not a problem, right?

modules/storages/app/common/storages/peripherals/storage_interaction/one_drive/download_link_query.rb
Kharonus: ~probably overlooked. Will check, thanks~

Ah, answered question below, misunderstood it before.

Kharonus:
> New `UrlBuilder.url` and `UrlBuilder.path` look really nice after all! 👍 In a couple of places uri paths are replaced with full uri. I guess it is not a problem, right?

Before, when using `Util.using_admin_token` the url authority was set as the `origin` on the httpx object by this method. The generic, authenticated httpx object returned by the authentication strategies does not do this, hence all queries using the new strategies already have to use full urls.

In this fix, I did not replace the auth strategy, but even with the old method, a full url including authority will work, as this simply overrides the set `origin` on the http session.

mereghost (approved):
Awesomesauce! :shipit:

PR #16163, opened 2024-07-16
Don't blink custom fields form when hiding parts depending on field format

toy:
[OP#56842](https://community.openproject.org/work_packages/56842) follow up for #16142 ([OP#53574](https://community.openproject.org/work_packages/53574)).
Currently correct parts of custom field form are displayed for field type string (Text), so only for new custom field and for field of that type, when editing other fields there will be a blink when stimulus controller toggles the needed parts. Move the config to backend to load correct fields at start and send that config to stimulus.

### DIFF
--- a/app/views/admin/settings/project_custom_fields/new.html.erb
+++ b/app/views/admin/settings/project_custom_fields/new.html.erb
@@ -26,7 +26,7 @@ Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
 See COPYRIGHT and LICENSE files for more details.
 
 ++#%>
-<% content_controller 'admin--custom-fields', dynamic: true %>
+<% content_controller 'admin--custom-fields', dynamic: true, 'admin--custom-fields-format-config-value': custom_fields_format_config %>
 
 <% html_title t(:label_administration), t("settings.project_attributes.heading"), t('settings.project_attributes.new.heading') %>
 <% local_assigns[:additional_breadcrumb] = t('settings.project_attributes.new.heading') %>
--- a/app/views/admin/settings/project_custom_fields/new.html.erb
+++ b/app/views/admin/settings/project_custom_fields/new.html.erb
@@ -26,7 +26,6 @@ Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
 See COPYRIGHT and LICENSE files for more details.
 
 ++#%>
-<% content_controller 'admin--custom-fields', dynamic: true, 'admin--custom-fields-format-config-value': custom_fields_format_config %>
 
 <% html_title t(:label_administration), t("settings.project_attributes.heading"), t('settings.project_attributes.new.heading') %>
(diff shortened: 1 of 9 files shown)


app/views/admin/settings/project_custom_fields/new.html.erb
```diff
@@ -26,7 +26,7 @@ Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
 See COPYRIGHT and LICENSE files for more details.
 
 ++#%>
-<% content_controller 'admin--custom-fields', dynamic: true %>
+<% content_controller 'admin--custom-fields', dynamic: true, 'admin--custom-fields-format-config-value': custom_fields_format_config %>
```
ulferts: I think this could be moved into the _form partial so that it only gets called once. Or is there any field I am not aware  of outside of the form that is also controlled by the stimulus controller

frontend/src/stimulus/controllers/dynamic/admin/custom-fields.controller.ts
```diff
@@ -236,15 +242,12 @@ export default class CustomFieldsController extends Controller {
...
-    this.setActive(this.regexpTargets, !['list', 'bool', 'date', 'user', 'version'].includes(format));
-    this.setActive(this.searchableTargets, !['bool', 'date', 'float', 'int', 'user', 'version'].includes(format));
-    this.setActive(this.textOrientationTargets, format === 'text');
+    this.formatConfigValue.forEach(([targetsName, operator, formats]) => {
+      const active = operator == 'only' ? formats.includes(format) : !formats.includes(format);
+      const targets = this[`${targetsName}Targets` as keyof typeof this] as HTMLElement[];
+      if (targets) {
+        this.setActive(targets, active);
+      }
+    });
```
ulferts, line 251: ❓ admittedly I don't get how this get's rid of the blinking, but it does

ulferts (commented):
So far looks promising @toy .

app/views/admin/settings/project_custom_fields/new.html.erb
toy: Indeed, also I didn't see why is it needed to set stimulus controller for the content wrapper and not just for the form

frontend/src/stimulus/controllers/dynamic/admin/custom-fields.controller.ts
toy: Frontend part doesn't really change, just config is moved to backend, so it is backend part that gets rid of blinking, as fields are at start hidden or visible depending on the field format, so before stimulus toggles them

app/views/admin/settings/project_custom_fields/new.html.erb
toy: 821c21e978a634e34b739e2642bc4d78e24f3700 and also e19f46960f1d2d92ddf2b961a2404705d9091862

oliverguenther (approved):

PR #16167, opened 2024-07-17
[56194] Follow-up remaining issues for the Dark mode (second round)

bsatarnejad:
- [x] Hover over a selected notification: It has the same color as the background, making it unclear that something is selected.
- [x] Bg color of skeletons in notification center is very dark in dark mode.
- [x] White bg color in show mode in tables when activating header columns and header row

https://community.openproject.org/work_packages/56194/activity
https://community.openproject.org/work_packages/56502/activity

### DIFF
--- a/frontend/src/app/features/in-app-notifications/entry/in-app-notification-entry.component.sass
+++ b/frontend/src/app/features/in-app-notifications/entry/in-app-notification-entry.component.sass
@@ -28,7 +28,7 @@ $subject-font-size: 14px
     border-bottom: 1px solid var(--borderColor-default)
 
   &_selected
-    background: $ian-bg-selected-color
+    background: $ian-bg-selected-color !important
     &:hover
       background: $ian-bg-selected-hover-color
 
--- a/frontend/src/app/features/in-app-notifications/entry/in-app-notification-entry.component.sass
+++ b/frontend/src/app/features/in-app-notifications/entry/in-app-notification-entry.component.sass
@@ -5,7 +5,7 @@ $ian-bg-hover-color: var(--bgColor-neutral-muted)
 $ian-bg-read-color: var(--bgColor-muted)
 $ian-bg-read-hover-color: hsl(from $ian-bg-read-color h s calc(l - 0.07))
 $ian-bg-selected-color: var(--selection-bgColor)
-$ian-bg-selected-hover-color: hsl(from $ian-bg-selected-color calc(h - 7) s l)
+$ian-bg-selected-hover-color: hsl(from $ian-bg-selected-color calc(h - 5) s l)
 // This needs to be set in the itemSize of
 // the virtual scroller
 $ian-height: 100px
@@ -28,7 +28,7 @@ $subject-font-size: 14px
     border-bottom: 1px solid var(--borderColor-default)
 
   &_selected
-    background: $ian-bg-selected-color !important
+    background: $ian-bg-selected-color
     &:hover
       background: $ian-bg-selected-hover-color
 
--- a/frontend/src/app/features/in-app-notifications/entry/in-app-notification-entry.component.sass
+++ b/frontend/src/app/features/in-app-notifications/entry/in-app-notification-entry.component.sass
@@ -5,7 +5,7 @@ $ian-bg-hover-color: var(--bgColor-neutral-muted)
 $ian-bg-read-color: var(--bgColor-muted)
 $ian-bg-read-hover-color: hsl(from $ian-bg-read-color h s calc(l - 0.07))
 $ian-bg-selected-color: var(--selection-bgColor)
-$ian-bg-selected-hover-color: hsl(from $ian-bg-selected-color calc(h - 5) s l)
+$ian-bg-selected-hover-color: hsl(from var(--selection-bgColor) h calc(s - 30) l)
 // This needs to be set in the itemSize of
 // the virtual scroller
 $ian-height: 100px
--- a/frontend/src/app/shared/components/op-content-loader/op-content-loader.component.html
+++ b/frontend/src/app/shared/components/op-content-loader/op-content-loader.component.html
@@ -1,6 +1,6 @@
 <content-loader
   [backgroundColor]="'var(--skeletonLoader-bgColor)'"
-  [foregroundColor]="'var(--data-gray-color-muted)'"
+  [foregroundColor]="'var(--bgColor-neutral-muted)'"
   [baseUrl]="baseUrl"
   [speed]="3"
   [viewBox]="viewBox"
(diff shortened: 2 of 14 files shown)


frontend/src/app/features/in-app-notifications/entry/in-app-notification-entry.component.sass
```diff
@@ -28,7 +28,7 @@ $subject-font-size: 14px
     border-bottom: 1px solid var(--borderColor-default)
 
   &_selected
-    background: $ian-bg-selected-color
+    background: $ian-bg-selected-color !important
```
HDinger: Because of the `!important`, the hover effect styles below are completely ignored which is not what we want. I guess, you have to change ` $ian-bg-selected-hover-color` instead

frontend/src/app/shared/components/table/table.sass
```diff
@@ -8,7 +8,7 @@
     text-align: center
 
     &_heading
-      background-color: var(--fc-neutral-bg-color)
+      background-color: var(--display-gray-bgColor-muted)
```
HDinger: Why was this change necessary? the heading is now pretty dark

Bildschirmfoto 2024-07-17 um 12 18 41

**Before**
Bildschirmfoto 2024-07-17 um 12 18 27

frontend/src/global_styles/content/user-content/_table.sass
```diff
@@ -18,17 +18,16 @@
...
 
     @at-root .op-uc-container_editing &:hover
       background: rgba(26, 103, 163, 0.05)
 
     &_head
-      background: rgb(242, 242, 242)
-      background-clip: padding-box
-      border-bottom: 1px solid rgba(0, 0, 0, 0.1)
+      background: var(--display-gray-bgColor-muted)
+      border-bottom: 1px solid var(--borderColor-default)
```
HDinger: Both of these styles are overriden and not applied at all. Because of that, the headers are not shown differently at all in the dark mode (when editing e.g. the description)
Bildschirmfoto 2024-07-17 um 12 12 57

frontend/src/app/shared/components/op-content-loader/op-content-loader.component.html
```diff
@@ -1,5 +1,5 @@
 <content-loader
-  [backgroundColor]="'var(--bgColor-muted)'"
+  [backgroundColor]="'var(--skeletonLoader-bgColor)'"
```
HDinger, line 2: That change does not really make difference. It even looks like it is darker than before, so even worse visible

**Before**
Bildschirmfoto 2024-07-17 um 12 29 11


**After**
Bildschirmfoto 2024-07-17 um 12 30 38

HDinger (changes requested):
I left some remarks

frontend/src/app/shared/components/table/table.sass
bsatarnejad: Hi @HDinger 
--fc-neutral-text-color is a fullcalendar variable, can we use it for a table?
I think that it was set by mistake.

frontend/src/app/features/in-app-notifications/entry/in-app-notification-entry.component.sass
```diff
@@ -5,7 +5,7 @@ $ian-bg-hover-color: var(--bgColor-neutral-muted)
 $ian-bg-read-color: var(--bgColor-muted)
 $ian-bg-read-hover-color: hsl(from $ian-bg-read-color h s calc(l - 0.07))
 $ian-bg-selected-color: var(--selection-bgColor)
-$ian-bg-selected-hover-color: hsl(from $ian-bg-selected-color calc(h - 7) s l)
+$ian-bg-selected-hover-color: hsl(from $ian-bg-selected-color calc(h - 5) s l)
```
HDinger: The hover effect is still not visible in Firefox

frontend/src/global_styles/content/editor/_ckeditor.sass
```diff
@@ -41,6 +41,10 @@ ckeditor-augmented-textarea .op-ckeditor--wrapper
     th[style^="width:"]
       word-break: break-all
 
+  .table table th
+    border-color: var(--borderColor-default) !important
+    background-color: var(--bgColor-muted) !important
```
HDinger: Please get rid of the `!important`. There are ways to be more specifc and thus enforce our styles without that. Further, this style is duplicating the user-content table header styles. So please extract that into a mixin. Otherwise, one will somewhen change and the other will be forgotten

HDinger (changes requested):
Hi @bsatarnejad 

I have some remarks left.

frontend/src/global_styles/openproject/_mixins.sass
```diff
@@ -361,3 +361,7 @@ $scrollbar-size: 10px
     display: table
   &:after
     clear: both
+
+@mixin table-header-styles
```
HDinger: The name is a bit misleading as these are not our default table header styles, but rather those of tables inside an editor content. We already have a prefix for these kind of content called `op-uc`. Please use that here as well

frontend/src/app/features/in-app-notifications/entry/in-app-notification-entry.component.sass
```diff
@@ -5,7 +5,7 @@ $ian-bg-hover-color: var(--bgColor-neutral-muted)
 $ian-bg-read-color: var(--bgColor-muted)
 $ian-bg-read-hover-color: hsl(from $ian-bg-read-color h s calc(l - 0.07))
 $ian-bg-selected-color: var(--selection-bgColor)
-$ian-bg-selected-hover-color: hsl(from $ian-bg-selected-color calc(h - 5) s l)
+$ian-bg-selected-hover-color: hsl(from var(--selection-bgColor) h calc(s - 30) l)
```
HDinger, line 8: The issue is not that the color is not dark enough. Firefox calculates this to `rgba(0, 0, 0, 0)` in the dark mode, and thus making the effect invisible. Are you able to reproduce this issue? I am currently running on version 127.0.1

HDinger (changes requested):

frontend/src/app/features/in-app-notifications/entry/in-app-notification-entry.component.sass
HDinger: Ok, I updated to 128.0.2 and it seems to be solved in there 🤷

bsatarnejad: Hi, yes mine is 128.0.2 too, and it is ok there.

HDinger (approved):

PR #16180, opened 2024-07-18
[#55643] Extend API authentication to accept JWT issued by OpenID pro…

ba1ash:
…vider to other client.

https://community.openproject.org/work_packages/55643

- Add new warden authentication stategy to handle jwt issued by configured OIDC.
- Modify exisiting doorkeeper_oauth stategy to ignore jwts.
- Fill in WWW-Autheticate header with auth failure information.
- Make keycloak docker dev setup use postgres as a database.

### DIFF
--- a/lib_static/open_project/authentication.rb
+++ b/lib_static/open_project/authentication.rb
@@ -261,18 +261,25 @@ module OpenProject
       def response_header(
         default_auth_scheme: self.default_auth_scheme,
         scope: nil,
-        request_headers: {}
+        request_headers: {},
+        error: nil,
+        error_description: nil
       )
-        scheme = pick_auth_scheme auth_schemes(scope), default_auth_scheme, request_headers
+        scheme = pick_auth_scheme(auth_schemes(scope),
+                                  default_auth_scheme,
+                                  request_headers)
 
-        "#{scheme} realm=\"#{scope_realm(scope)}\""
+        header = %{#{scheme} realm="#{scope_realm(scope)}"}
+        header << %{ error="#{error}"}                         if error
+        header << %{ error_description="#{error_description}"} if error && error_description
+        header
       end
 
       def auth_schemes(scope)
         strategies = Array(Manager.scope_config(scope).strategies)
 
         Manager.auth_schemes
-          .select { |_, info| scope.nil? or not (info.strategies & strategies).empty? }
+          .select { |_, info| scope.nil? or not !info.strategies.intersect?(strategies) }
           .keys
       end
     end
--- a/lib_static/open_project/authentication.rb
+++ b/lib_static/open_project/authentication.rb
@@ -279,7 +279,7 @@ module OpenProject
         strategies = Array(Manager.scope_config(scope).strategies)
 
         Manager.auth_schemes
-          .select { |_, info| scope.nil? or not !info.strategies.intersect?(strategies) }
+          .select { |_, info| scope.nil? or info.strategies.intersect?(strategies) }
           .keys
       end
     end
--- a/lib_static/open_project/authentication/strategies/warden/doorkeeper_oauth.rb
+++ b/lib_static/open_project/authentication/strategies/warden/doorkeeper_oauth.rb
@@ -7,6 +7,8 @@ module OpenProject
         class DoorkeeperOAuth < ::Warden::Strategies::Base
           include FailWithHeader
 
+          # The strategy is supposed to handle bearer tokens that are not JWT.
+          # These tokens are issued by OpenProject
           def valid?
             access_token = ::Doorkeeper::OAuth::Token
                              .from_request(decorated_request, *Doorkeeper.configuration.access_token_methods)
--- a/lib_static/open_project/authentication/strategies/warden/doorkeeper_oauth.rb
+++ b/lib_static/open_project/authentication/strategies/warden/doorkeeper_oauth.rb
@@ -10,13 +10,15 @@ module OpenProject
           def valid?
             access_token = ::Doorkeeper::OAuth::Token
                              .from_request(decorated_request, *Doorkeeper.configuration.access_token_methods)
-            access_token.present? &&
-              begin
-                JWT.decode(access_token, nil, false)
-                false
-              rescue JWT::DecodeError
-                true
-              end
+            
+            # No access token found, so invalid strategy.
+            return false if access_token.blank?
+            
+            # We don't want JWT as our OAuth Bearer token
+            JWT.decode(access_token, nil, false)
+            false
+          rescue JWT::DecodeError
+            true
           end
 
           def authenticate!
--- a/lib_static/open_project/authentication/strategies/warden/doorkeeper_oauth.rb
+++ b/lib_static/open_project/authentication/strategies/warden/doorkeeper_oauth.rb
@@ -4,65 +4,52 @@ module OpenProject
   module Authentication
     module Strategies
       module Warden
-        ##
-        # Allows testing authentication via doorkeeper OAuth2 token
-        #
         class DoorkeeperOAuth < ::Warden::Strategies::Base
+          include FailWithHeader
+
           def valid?
-            ::Doorkeeper::OAuth::Token
-              .from_request(decorated_request, *Doorkeeper.configuration.access_token_methods)
-              .present?
+            access_token = ::Doorkeeper::OAuth::Token
+                             .from_request(decorated_request, *Doorkeeper.configuration.access_token_methods)
+            access_token.present? &&
+              begin
+                JWT.decode(access_token, nil, false)
+                false
+              rescue JWT::DecodeError
+                true
+              end
           end
 
           def authenticate!
-            token = ::Doorkeeper::OAuth::Token.authenticate(decorated_request,
-                                                            *Doorkeeper.configuration.access_token_methods)
-            return fail!("invalid token") unless token&.acceptable?(scope)
+            access_token = ::Doorkeeper::OAuth::Token.authenticate(decorated_request,
+                                                                   *Doorkeeper.configuration.access_token_methods)
+            return fail_with_header!(error: "invalid_token") if access_token.blank?
+            return fail_with_header!(error: "invalid_token") if access_token.expired? || access_token.revoked?
+            return fail_with_header!(error: "insufficient_scope") if !access_token.includes_scope?(scope)
 
-            if token.resource_owner_id.nil?
-              authenticate_client_credentials(token)
+            if access_token.resource_owner_id.nil?
+              user_id = ::Doorkeeper::Application
+                          .where(id: access_token.application_id)
+                          .where.not(client_credentials_user_id: nil)
+                          .pick(:client_credentials_user_id)
+              authenticate_user(user_id) if user_id
             else
-              authenticate_user(token.resource_owner_id)
+              authenticate_user(access_token.resource_owner_id)
             end
           end
 
           private
-
-          ##
-          # We allow applications to designate a user to be used for client credentials.
-          # When using client credentials flow, find this user and try to authenticate
-          def authenticate_client_credentials(token)
-            if client_credential_user = find_credential_app_user(token.application_id)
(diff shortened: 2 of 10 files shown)


lib/api/root.rb
```diff
@@ -39,13 +39,18 @@ class Root < ::API::RootAPI
...
-    OpenProject::Authentication.handle_failure(scope: API_V3) do |warden, _opts|
-      e = grape_error_for warden.env, self
+    OpenProject::Authentication.handle_failure(scope: API_V3) do |warden, opts|
+      e = grape_error_for(warden.env, self)
       error_message = I18n.t("api_v3.errors.code_401_wrong_credentials")
-      api_error = ::API::Errors::Unauthenticated.new error_message
-      representer = ::API::V3::Errors::ErrorRepresenter.new api_error
+      api_error = ::API::Errors::Unauthenticated.new(error_message)
+      representer = ::API::V3::Errors::ErrorRepresenter.new(api_error)
+      status = if opts[:message] == "insufficient_scope"
```
Kharonus: 🟡 minor: this look so small, a ternary operator would be cool?

```ruby
status = opts[:message] == "insufficient_scope" ? 403 : 401
```

lib_static/open_project/authentication/strategies/warden/doorkeeper_oauth.rb
```diff
@@ -4,65 +4,60 @@ module OpenProject
...
+
           def valid?
-            ::Doorkeeper::OAuth::Token
-              .from_request(decorated_request, *Doorkeeper.configuration.access_token_methods)
-              .present?
+            access_token = ::Doorkeeper::OAuth::Token
+                             .from_request(decorated_request, *Doorkeeper.configuration.access_token_methods)
+            access_token.present? &&
+              begin
+                JWT.decode(access_token, nil, false)
```
Kharonus: ❓ How do the strategy lookup work? Will all the strategies go through `valid?` Which will be the `winning_strategy`?

My question is: do we need to put some JWT validation into the doorkeeper OAuth, just to return `true` on a failed decoding? Or can we manage to change is so, that if a JWT token is sent, DoorkeeperOAuth would still send back "valid", but JWT OIDC will be the `winning_strategy`?

Intention is to get JWT dependency out of a validation for OAuth bearer tokens. (which have no restrictions of pattern)

lib_static/open_project/authentication/strategies/warden/doorkeeper_oauth.rb
```diff
@@ -4,65 +4,60 @@ module OpenProject
...
+                  end
+                else
+                  fail_with_header!(error: "insufficient_scope")
+                end
+              else
+                fail_with_header!(error: "invalid_token")
+              end
             else
-              authenticate_user(token.resource_owner_id)
+              fail_with_header!(error: "invalid_token")
```
Kharonus: 🟡 To reduce nesting, I normally prefer a coding style like:

```ruby
return fail_with_header!(error: "invalid_token") unless access_token.present?

return fail_with_header!(error: "invalid_token") if access_token.expired? || access_token.revoked?

return fail_with_header!(error: "insufficient_scope") if access_token.excludes_scope?(scope)

if access_token.resource_owner_id.nil?
  user_id = ::Doorkeeper::Application
              .where(id: access_token.application_id)
              .where.not(client_credentials_user_id: nil)
              .pick(:client_credentials_user_id)
  authenticate_user(user_id) if user_id
else
  authenticate_user(access_token.resource_owner_id)
end
```

WDYT? I find this much more readable.

Kharonus (changes requested):
As discussed before, no hard change requests, but some cosmetic ones, I'd like to mention. Of course, rubocop still lists some issues. ;)

lib_static/open_project/authentication/strategies/warden/doorkeeper_oauth.rb
ba1ash: The order defined in https://github.com/opf/openproject/blob/9428a5df4daeeb7ca1138fb0dd0576cf5a0041cc/config/initializers/warden.rb#L18.
The strategies are tried in this order. If a strategy is `#valid?` then its `#authenticate!` is called.  `#authenticate!` has three options: 
1. call `success!`. It means user has been found and next strategies in the order will not be tried.
2. call `fail!`. It means authentication failed for whatever reason and next strategies in the order will not be tried.
3. call `pass` explicitly or do not call `success!` or `fail!`. User has not been authenticated, but next strategies will be tried.

Speaking about  the JWT dependency. `doorkeeper-oauth` has possibility to use `jwt`s, but we do not use it in our application for now, because of that I decided to rely on bearer token format to distinguish strategies. I would keep it this way. It is a stable way to decided which strategy to use. At least for now.

lib_static/open_project/authentication/strategies/warden/doorkeeper_oauth.rb
ba1ash: I think it looks nice. I will rewrite it. Thanks for suggestion!

lib_static/open_project/authentication.rb
```diff
@@ -261,18 +261,25 @@ def scope_realm(scope = nil)
...
+        header << %{ error_description="#{error_description}"} if error && error_description
+        header
       end
 
       def auth_schemes(scope)
         strategies = Array(Manager.scope_config(scope).strategies)
 
         Manager.auth_schemes
-          .select { |_, info| scope.nil? or not (info.strategies & strategies).empty? }
+          .select { |_, info| scope.nil? or not !info.strategies.intersect?(strategies) }
```
mereghost: 🔴 The change from `empty?` to `intersect?` made the `not` superfluous and we ended up with a `not not`

The original was "if scope is nil and the intersection of strategies was not empty" which means "if the scope is nil or there is an intersection of strategies".  

```suggestion
          .select { |_, info| scope.nil? or info.strategies.intersect?(strategies) }
```

lib_static/open_project/authentication/strategies/warden/doorkeeper_oauth.rb
```diff
@@ -4,65 +4,52 @@ module OpenProject
   module Authentication
     module Strategies
       module Warden
-        ##
-        # Allows testing authentication via doorkeeper OAuth2 token
-        #
         class DoorkeeperOAuth < ::Warden::Strategies::Base
+          include FailWithHeader
+
           def valid?
```
mereghost, line 12: 🟡 If I get this right this method is checking:

1. Is there a token on the request?
2. No: return false
3. Yes: Is it a valid JWT?
4. No: Return true (valid)
5. Yes: return false (invalid)

It it to avoid doing OAuth using a JWT token? I think we could get a lot of love from future us if we leave a comment explaining the reasoning. =)

lib_static/open_project/authentication/strategies/warden/doorkeeper_oauth.rb
```diff
@@ -4,65 +4,52 @@ module OpenProject
...
-              .present?
+            access_token = ::Doorkeeper::OAuth::Token
+                             .from_request(decorated_request, *Doorkeeper.configuration.access_token_methods)
+            access_token.present? &&
+              begin
+                JWT.decode(access_token, nil, false)
+                false
+              rescue JWT::DecodeError
+                true
+              end
```
mereghost: 🟡 I find this method a bit hard to read with the block following the conditional. WDYT of this?

```suggestion
            access_token = ::Doorkeeper::OAuth::Token
                             .from_request(decorated_request, *Doorkeeper.configuration.access_token_methods)
            
            # No access token found, so invalid strategy.
            return false if access_token.blank?
            
            # We don't want JWT as our OAuth Bearer token
            JWT.decode(access_token, nil, false)
            false
          rescue JWT::DecodeError
            true
```

lib_static/open_project/authentication/strategies/warden/jwt_oidc.rb
```diff
@@ -0,0 +1,82 @@
...
+              ::Doorkeeper::Grape::AuthorizationDecorator.new(request)
+            )
+            @access_token.present? &&
+              begin
+                @unverified_payload, @unverified_header = JWT.decode(@access_token, nil, false)
+                @unverified_header.present? && @unverified_payload.present?
+              rescue JWT::DecodeError
+                false
+              end
+          end
```
mereghost: 🔴 Same as above, I think we could improve legibility and the same question about making it a stateful object rather and a stateless one.

```suggestion
          def valid?
            @access_token = ::Doorkeeper::OAuth::Token.from_bearer_authorization(
              ::Doorkeeper::Grape::AuthorizationDecorator.new(request)
            )
            return false if @access_token.blank?
                
            @unverified_payload, @unverified_header = JWT.decode(@access_token, nil, false)
            @unverified_header.present? && @unverified_payload.present?
          rescue JWT::DecodeError
            false
          end
```

lib_static/open_project/authentication/strategies/warden/doorkeeper_oauth.rb
```diff
@@ -4,65 +4,52 @@ module OpenProject
...
             user = User.find_by(id:)
             if user
               success!(user)
             else
-              fail!("No such user")
+              fail_with_header!(error: "invalid_token")
             end
           end
 
           def decorated_request
```
mereghost, line 55: 🔴 Does the strategy gets recreated at every request? If not, memoizing here can be really problematic as it will always check the memoized request.

lib_static/open_project/authentication/strategies/warden/jwt_oidc.rb
```diff
@@ -0,0 +1,82 @@
...
+
+          def authenticate!
```
mereghost, line 28: 🟡 This method is... long. 🤣 And there's a ton of nesting going on, maybe some guard clauses can probably help.

Quick question: what should be returned here? We have a lot of `return`s, which basically return `nil`, is it intended? Should `nil` be the return value?

mereghost (changes requested):
That's a lot of stuff! I have made some points that I'd like to understand better before approving it, but overall looks great.

lib_static/open_project/authentication.rb
ba1ash: Right, good catch. It was rubocop autofixer.

lib_static/open_project/authentication/strategies/warden/doorkeeper_oauth.rb
ba1ash: The logic is right like you described it. Sure, I will add it.

lib_static/open_project/authentication/strategies/warden/doorkeeper_oauth.rb
ba1ash: It looks fine to me. Let's have it this way. The point I like in my version is really small `begin..end` and then it is obvious which statement is expected to raise an error.

mereghost: Fine by me. =)

lib_static/open_project/authentication/strategies/warden/jwt_oidc.rb
ba1ash: I believe the method is called here:  https://github.com/wardencommunity/warden/blob/master/lib/warden/strategies/base.rb#L55C9-L55C22
So, no return value is expected. `returns` are just to stop following method execution.

mereghost: Cool then. :D

lib_static/open_project/authentication/strategies/warden/doorkeeper_oauth.rb
ba1ash: I believe it is recreated every request. I will check to make sure.

ba1ash: Yes, the object is new for every request.

mereghost (approved):

PR #16184, opened 2024-07-18
Fix writable_attributes when redefining attributes in child classes

cbliard:
There were cases were the writable parameter of a parent contract would take precedence of the child contract writable parameter when the parent condition is a lambda: a child could say `writable: true` and it would still be the parent `writable: -> { false }` that would "win" because when reducing and evaluating `writable_conditions`, only lambda are considered.

This commit fixes the issue by adding extensive tests and putting both lambdas AND true/false values in the `writable_conditions` hash.

### DIFF
--- a/app/contracts/base_contract.rb
+++ b/app/contracts/base_contract.rb
@@ -50,7 +50,7 @@ class BaseContract < Disposable::Twin
     end
 
     def writable_conditions
-      @writable_conditions ||= []
+      @writable_conditions ||= {}
     end
 
     def attribute_permissions
@@ -102,16 +102,13 @@ class BaseContract < Disposable::Twin
 
     def add_writable(attribute, writable)
       attribute_name = attribute.to_s.delete_suffix("_id")
+      writable_conditions[attribute_name] = writable
 
-      unless writable == false
-        writable_attributes << attribute_name
-        # allow the _id variant as well
-        writable_attributes << "#{attribute_name}_id"
-      end
+      return if writable == false
 
-      if writable.respond_to?(:call)
-        writable_conditions << [attribute_name, writable]
-      end
+      writable_attributes << attribute_name
+      # allow the _id variant as well
+      writable_attributes << "#{attribute_name}_id"
     end
   end
 
@@ -250,7 +247,8 @@ class BaseContract < Disposable::Twin
 
   def reduce_by_writable_conditions(attributes)
     collect_ancestor_attributes(:writable_conditions).each do |attribute, condition|
-      attributes -= [attribute, "#{attribute}_id"] unless instance_exec(&condition)
+      condition = !!instance_exec(&condition) if condition.respond_to?(:call)
+      attributes -= [attribute, "#{attribute}_id"] if condition == false
     end
 
     attributes
--- a/spec/contracts/base_contract_spec.rb
+++ b/spec/contracts/base_contract_spec.rb
@@ -150,7 +150,7 @@ RSpec.describe BaseContract do
         expect(contract.writable_attributes).to contain_exactly("name", "name_id")
       end
 
-      it "does not returns it when redefined as `writable: false` in the child class" do
+      it "does not return it when redefined as `writable: false` in the child class" do
         child_contract_class = Class.new(parent_contract_class) do
           attribute :name, writable: false
         end
--- a/spec/contracts/base_contract_spec.rb
+++ b/spec/contracts/base_contract_spec.rb
@@ -255,7 +255,4 @@ RSpec.describe BaseContract do
       it_behaves_like "the parent writable parameter is overridden by the child writable parameter"
     end
   end
-
-  describe "#reduce_by_writable_conditions"
-  describe "#reduce_writable_attributes"
 end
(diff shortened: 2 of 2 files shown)


app/contracts/base_contract.rb
```diff
@@ -102,16 +102,13 @@ def attribute_permission(attribute, permission)
...
-      unless writable == false
-        writable_attributes << attribute_name
-        # allow the _id variant as well
-        writable_attributes << "#{attribute_name}_id"
-      end
+      return if writable == false
 
-      if writable.respond_to?(:call)
-        writable_conditions << [attribute_name, writable]
-      end
```
aaron-contreras, line 114: _Sherlock Bliard 🕵️_

app/contracts/base_contract.rb
```diff
@@ -250,7 +247,8 @@ def reduce_writable_attributes(attributes)
 
   def reduce_by_writable_conditions(attributes)
     collect_ancestor_attributes(:writable_conditions).each do |attribute, condition|
-      attributes -= [attribute, "#{attribute}_id"] unless instance_exec(&condition)
+      condition = !!instance_exec(&condition) if condition.respond_to?(:call)
+      attributes -= [attribute, "#{attribute}_id"] if condition == false
```
aaron-contreras, line 251: To make sure I understand correctly, the behavior should now be:

- if the condition is a lambda, execute it to test if it resolves to `true`/`false`
- if it isn't, keep the condition that might be coming in as a primitive `true`/`false` boolean value
- use this condition to determine whether to remove the attribute from the list of writable ones

?

spec/contracts/base_contract_spec.rb
```diff
@@ -0,0 +1,261 @@
...
+      end
+      expect(contract_class.writable_attributes).to be_empty
+    end
+
+    it "returns attributes with `writable: nil`" do
+      contract_class = Class.new(BaseContract) do
+        attribute :name, writable: nil
+      end
+      expect(contract_class.writable_attributes).to contain_exactly("name", "name_id")
+    end
```
aaron-contreras, line 55: If this wasn't added as a spec and now being documented, I would spend quite some time trying to debug why an attribute is not writable if nil is provided 😆

spec/contracts/base_contract_spec.rb
```diff
@@ -0,0 +1,261 @@
...
+
+    it "returns attributes with `writable: -> { false }`" do
+      contract_class = Class.new(BaseContract) do
+        attribute :name, writable: -> { false }
+      end
+      expect(contract_class.writable_attributes).to contain_exactly("name", "name_id")
+    end
```
aaron-contreras, line 76: Why?

aaron-contreras: Because it's not being instantiated and therefore only prevents adding to `writable_attributes` if it's directly a primitive boolean? 

From the code, I see it only execs the lambdas when a contract is instantiated and now needs to look in the hierarchy and eval the lambda

spec/contracts/base_contract_spec.rb
```diff
@@ -0,0 +1,261 @@
...
+
+      it "returns it when redefined as `writable: true` in the child class" do
+        child_contract_class = Class.new(parent_contract_class) do
+          attribute :name, writable: true
+        end
+        contract = child_contract_class.new(model, user)
+        expect(contract.writable_attributes).to contain_exactly("name", "name_id")
+      end
+
+      it "does not returns it when redefined as `writable: false` in the child class" do
```
aaron-contreras:
```suggestion
      it "does not return it when redefined as `writable: false` in the child class" do
```

spec/contracts/base_contract_spec.rb
```diff
@@ -0,0 +1,261 @@
...
+        contract = child_contract_class.new(model, user)
+        expect(contract.writable_attributes).to be_empty
+      end
+
+      it_behaves_like "the parent writable parameter is overridden by the child writable parameter"
+    end
+  end
+
+  describe "#reduce_by_writable_conditions"
+  describe "#reduce_writable_attributes"
```
aaron-contreras: Remove these? or add a pending?

aaron-contreras (approved):
All behaving to spec now - just left some questions for my own understanding

Thorough examples, nicely done! 🤠

app/contracts/base_contract.rb
cbliard: Yes, exactly

app/contracts/base_contract.rb
cbliard: yep, not needed anymore as the writable value is always written to writable_conditions now (and it's a hash now too).

spec/contracts/base_contract_spec.rb
cbliard: That's why I added a test for it: it surprised me that an attribute is writable by default just because the value is `nil`.

spec/contracts/base_contract_spec.rb
cbliard: > Why?

I don't know, I only tested the existing behavior. I think evaluating the lambda only makes sense when you have a contract instance, and not a class. The class method `.writable_attributes` does not seem very useful IMHO.

spec/contracts/base_contract_spec.rb
cbliard: yes, I should remove them. Good point. I thought I would test them, but it's not needed.

cbliard: Done

PR #16195, opened 2024-07-19
Fix calendar test displaying Monday as non-working week day

cbliard:
The test ./modules/calendar/spec/features/calendar_dates_spec.rb:50 fails locally for me as I have my local timezone being strictly positive (GMT+02). It probably passes on CI because the machine is configured to use UTC as timezone.

When trying to get the isoWeekday, it gets the utc time, which is 22:00 the previous day, and this shifts all non-working week days: Tuesday-Saturday is working, and Sunday-Monday is non-working.

The fix is a bit ugly, but works.


cbliard:
The fix does NOT seem to work actually. It breaks another test around the same subject.

PR #16209, opened 2024-07-22
fix internal server error on missing host_name setting during host mi…

machisuji:
…stmatch check


oliverguenther (approved):

machisuji:
1 unrelated feature spec was failing.

PR #16221, opened 2024-07-22
[#56609] Fix further possible TypeErrors for gitlab hooks

judithroth:
See https://community.openproject.org/projects/openproject/work_packages/56609
And https://appsignal.com/openproject-gmbh/sites/62b06dacd2a5e41321946fcf/exceptions/incidents/699

### DIFF
--- a/modules/gitlab_integration/spec/lib/open_project/gitlab_integration/notification_handler/push_hook_spec.rb
+++ b/modules/gitlab_integration/spec/lib/open_project/gitlab_integration/notification_handler/push_hook_spec.rb
@@ -139,7 +139,7 @@ RSpec.describe OpenProject::GitlabIntegration::NotificationHandler::PushHook do
       end
 
       it "does not raise (Bugfix)" do
-        process
+        expect { process }.not_to raise_error
         expect(handler_instance).to have_received(:comment_on_referenced_work_packages).with(
           [work_package],
           gitlab_system_user,
--- a/modules/gitlab_integration/spec/lib/open_project/gitlab_integration/notification_handler/push_hook_spec.rb
+++ b/modules/gitlab_integration/spec/lib/open_project/gitlab_integration/notification_handler/push_hook_spec.rb
@@ -1,0 +1,151 @@
+#-- copyright
+# OpenProject is an open source project management software.
+# Copyright (C) 2012-2024 the OpenProject GmbH
+#
+# This program is free software; you can redistribute it and/or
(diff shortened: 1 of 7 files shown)


modules/gitlab_integration/spec/lib/open_project/gitlab_integration/notification_handler/push_hook_spec.rb
```diff
@@ -0,0 +1,151 @@
...
+      end
+
+      let(:comment) do
+        "**Pushed in refs/heads/main:** [Administrator]" \
+          "(https://www.gravatar.com/avatar/65a222b844ced567fe0ed2594c0b4abdf62efa1322a385c919c41e7bbc16d4fc?s=80&d=identicon) " \
+          "pushed [a265d6b7](http://c7e7cd2d54c3/openprojecttest/test/-/commit/a265d6b7bcf836b77ed9e32f824b231585c6a355) " \
+          "to [Test](http://c7e7cd2d54c3/openprojecttest/test) at 2024-07-22T11:18:29+02:00:\nMentioning OP##{work_package.id}\n"
+      end
+
+      it "does not raise (Bugfix)" do
```
mereghost, line 141: 🟡 While the test mentions not raising, there's no assertion to it. I get why but I'd prefer for it either have one assertion like `expect { process }.to_not raise_error(ExceptionOfInterest)` and the message expectation or the description changed to something along the lines of `handles commits without messages`

mereghost (approved):
Great work here.

I've used an traffic light system to make notes on what I think is:
🟢 Just me being weird or minor suggestions
🟡 Points that might warrant clarification discussion
🔴 (which I use sparingly) Major issues that need to be addressed before a merge

modules/gitlab_integration/spec/lib/open_project/gitlab_integration/notification_handler/push_hook_spec.rb
judithroth: Good point. When writing my thinking was that the test would raise and therefore "fail" if there still was an error, but I also agree that making things explicit is often the better choice.
So I added your suggested expectation, but since you already approved I assume you don't want to review again, right? :slightly_smiling_face:

mereghost: That was my only point and wasn't something that I feel should even block the merge, so feel free to :shipit: =)

PR #16222, opened 2024-07-22
remove default from label of project id filter

toy:
https://github.com/opf/openproject/pull/16100#discussion_r1677601544


aaron-contreras (approved):
👍

PR #16252, opened 2024-07-25
[#56253] Correct missing filter value in notification translation string

aaron-contreras:
Bug work package: https://community.openproject.org/work_packages/56253

Changes "responsible" to "accountable"


cbliard (approved):
Looks all good. I rebased it onto `release/14.4` as it does not happen anymore on `dev` since the primerization (it now uses only a generic message).

PR #16277, opened 2024-07-29
[56654]Instance custom logo is very small

bsatarnejad:
https://community.openproject.org/wp/56654

### DIFF
--- a/frontend/src/global_styles/common/header/logo.sass
+++ b/frontend/src/global_styles/common/header/logo.sass
@@ -5,6 +5,7 @@
 
   &--link
     width: 100%
+    height: 30px
     display: block
     background: no-repeat 20px 0
     text-indent: -9999em
--- a/frontend/src/global_styles/common/header/logo.sass
+++ b/frontend/src/global_styles/common/header/logo.sass
@@ -1,5 +1,6 @@
 .op-logo
   display: flex
+  align-items: center
   margin: 10px 0
   min-width: 230px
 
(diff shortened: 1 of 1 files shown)


frontend/src/global_styles/common/header/logo.sass
```diff
@@ -1,12 +1,11 @@
...
-  height: var(--header-height)
+  display: flex
+  margin: 10px 0
+  min-width: 230px
 
   &--link
-    margin-top: 12px
     width: 100%
     display: block
-    height: 30px
```
HDinger, line 9: I think we should keep the height restriction as it enforces the logo to be horizontally aligned with the other elements of the tooolbar which looks cleaner imho.

HDinger (commented):
I have one minor remark. Other than that it looks fine 👍

frontend/src/global_styles/common/header/logo.sass
bsatarnejad: Hi @HDinger, yes agree, when we set the height for the logo, it is aligned with other header elements. 👍 
I thought if we don't set a height for it, it will stretch to the whole height of header and will be bigger. 
with height:
Screenshot 2024-07-30 at 10 56 11
without height:
Screenshot 2024-07-30 at 10 55 45

HDinger (changes requested):
The logo is now not correctly aligned. It seems to be closer to the top, than to the bottom:

Bildschirmfoto 2024-08-01 um 14 05 33

HDinger:
The changes basically have no effect as the image is the same size as before. I talked about it with the ticket author and we agreed to reject the whole ticket for now. Sorry for the incovinience. 😞

PR #16283, opened 2024-07-30
[56761] Dark mode is not applied to drop downs in cost reports

bsatarnejad:
https://community.openproject.org/work_packages/56761/activity


HDinger (changes requested):
As just discussed, there are some visual differences in the group headers and the already selected items.

HDinger (approved):
Good to merge once the final Windows test is done 👍

PR #16289, opened 2024-07-30
[#54122] GitLab integration pipeline status not updated

judithroth:
https://community.openproject.org/work_packages/54122

Previously, the sorting was done on the "name" attribute, which actually just is the gitlab_id saved as a string. So after 10 pipelines, the sorting was wrong.


mereghost (approved):
:shipit:

PR #16293, opened 2024-07-31
Feature/56045 add link to storage provider in storage edit view

apfohl:
https://community.openproject.org/projects/openproject/work_packages/56045/activity

## what?

This PR adds a link to the storage's main directory to the edit page in admin settings.

It calculates the href by using the `:open_storage` query implementation of either type.


akabiru (changes requested):
Looking good! 🟠 It would be good to include some tests- component level unit tests come to mind as a cheap way to assert against the functionality (some examples exist for storages [here](https://github.com/opf/openproject/tree/dev/modules/storages/spec/components/storages) )

akabiru (approved):
📓 @apfohl will followup with an improvement the the unit spec

PR #16310, opened 2024-08-02
Block browser requests made outside of feature tests

cbliard:

When test is finished, Capybara calls `Capybara.reset!` which in turn calls `driver.reset!`. This one is responsible for stopping the browser by navigating to about:blank page and waiting for pending requests to complete.

With Cuprite, we observed that some requests could still be triggered from the browser after `Capybara.reset!` is called. It can interfere with test execution is some unexpected ways: when a request to the API is made, the settings are read from the database. If this happens right when the database is being rolled back to a previous savepoint (which happens when using `before_all` helper), then in postgres dapater code there is a `nil` reference instead of a result, and then it errs with "NoMethodError: undefined method 'clear' for nil".

You can run tests from `spec/features/work_packages/progress_modal_spec.rb` a couple of times to experiment the error. It happens mostly with tests having the most nested `before_all` calls.

We tried navigating to about:blank with Cuprite, but some requests were still made. So we looked for another fix.

Using a middleware to actively block requests outside of test execution fixed the issue. It is simple and avoids monkey-patching Cuprite.

### DIFF
--- a/spec/support/capybara/block_requests_made_outside_of_test.rb
+++ b/spec/support/capybara/block_requests_made_outside_of_test.rb
@@ -1,0 +1,57 @@
+# When test is finished, Capybara calls `Capybara.reset!` which in turn calls
+# `driver.reset!`. This one is responsible for stopping the browser by
+# navigating to about:blank page and waiting for pending requests to complete.
+#
+# With Cuprite, we observed that some requests could still be triggered from the
+# browser after `Capybara.reset!` is called. It can interfere with test
+# execution is some unexpected ways: when a request to the API is made, the
+# settings are read from the database. If this happens right when the database
+# is being rolled back to a previous savepoint (which happens when using
+# `before_all` helper), then in postgres dapater code there is a `nil` reference
+# instead of a result, and then it errs with "NoMethodError: undefined method
+# 'clear' for nil".
+#
+# You can run tests from `spec/features/work_packages/progress_modal_spec.rb` a
+# couple of times to experiment the error. It happens mostly with tests having
+# the most nested `before_all` calls.
+#
+# We tried navigating to about:blank with Cuprite, but some requests were still
+# made. So we looked for another fix.
+#
+# Using a middleware to actively block requests outside of test execution fixed
+# the issue.
+
+class RequestsBlocker
+  def initialize(app)
+    @app = app
+    @blocked = false
+  end
+
+  def block_requests!
+    @blocked = true
+  end
+
+  def unblock_requests!
+    @blocked = false
+  end
+
+  def call(env)
+    if @blocked
+      [500, {}, "RequestsBlocker is blocking further requests because test is finished."]
+    else
+      @app.call(env)
+    end
+  end
+end
+
+RSpec.configure do |config|
+  Capybara.app = RequestsBlocker.new(Capybara.app)
+
+  config.before(:each, type: :feature) do
+    Capybara.app.unblock_requests!
+  end
+
+  config.after(:each, type: :feature) do
+    Capybara.app.block_requests!
+  end
+end


spec/support/capybara/block_requests_made_outside_of_test.rb
```diff
@@ -0,0 +1,60 @@
...
+# We tried navigating to about:blank with Cuprite, but some requests were still
+# made. So we looked for another fix.
+#
+# Using a middleware to actively block requests outside of test execution fixed
+# the issue.
+
+class RequestsBlocker
+  class << self
+    attr_accessor :blocked
+  end
```
toy: If there is one instance used, why is it on class level?

toy (commented):
Is it only about requests made in-between tests or there is a possibility of requests from one test to happen during another test?

It feels like there should be a better solution, like some config is missing or broken or we should be using some cuprite method to handle it

spec/support/capybara/block_requests_made_outside_of_test.rb
cbliard: Because I initially called `RequestsBlocker.block_requests!` and changed to `Capybara.app.block_requests!` but did not change the attribute.

Good point, I'll move it to instance level.

cbliard:
> Is it only about requests made in-between tests or there is a possibility of requests from one test to happen during another test?

It is about API requests made while the database is being rolled back to a save point after a test is finished. The API call calls the authentication logic, which read the settings to check if the rest api is enabled, which touch the database to refresh the setting cache. When this API request happens while the database is being rolled back, things go wrong. I'm not sure of what is going on. Maybe the puma server thread and the rspec thread share the same db connection. The result is a failure with methods being called on `nil` and sometimes the test execution is stuck, maybe a deadlock somewhere.

The solution I took is adapted from https://github.com/teamcapybara/capybara/issues/1716

> It feels like there should be a better solution, like some config is missing or broken or we should be using some cuprite method to handle it

Maybe, but I could not make it work. I tried calling every thing I could think of (`visit("about:blank")`, `driver.browser.page.close`, `driver.browser.reset`, `driver.clear_network_traffic`), and I was still seeing the notification poll requests coming (I patched `Capybara::Server::Middleware::Counter` from `lib/capybara/server/middleware.rb` to log uris being accessed). I do not understand what was left that still triggered those requests.

FWIW, the solution I tried was adapted from https://github.com/rubycdp/cuprite/issues/128#issuecomment-703854564

The middleware app with explicit blocking was more reliable and simple.

toy (approved):
Probably better to create a maintenance ticket to check some time later on progress in https://github.com/rubycdp/cuprite/issues/128

aaron-contreras (approved):
Absolutely brilliant!

cbliard:
> Probably better to create a maintenance ticket to check some time later on progress in [rubycdp/cuprite#128](https://github.com/rubycdp/cuprite/issues/128)

I wanted to have a reproduction script to submit to capybara people to have the blocking being implemented directly there, and maybe also to cuprite people to have the browser stop sending requests after `#reset!` is called, but I could not reproduce the issue.

Here is the script I wrote:

```ruby
# frozen_string_literal: true

require "bundler/inline"

gemfile(true) do
  source "https://rubygems.org"

  git_source(:github) { |repo| "https://github.com/#{repo}.git" }

  gem "capybara"
  gem "cuprite"
  gem "puma"
  gem "rails", "7.1"
  gem "rspec-rails"
end

require "rspec-rails"
require "action_controller/railtie"

class TestApp < Rails::Application
  config.root = __dir__
  config.hosts << "example.org"
  config.session_store :cookie_store, key: "cookie_store_key"
  credentials.secret_key_base = "secret_key_base"

  config.logger = Logger.new($stdout)
  Rails.logger  = config.logger

  routes.draw do
    root "time#index"
    get "current_time" => "time#current_time"
  end
end

class TimeController < ActionController::Base
  include Rails.application.routes.url_helpers

  def index
    render html: <<~HTML.html_safe
      <!DOCTYPE html>
      <html>
      <head>
        <title>Current Time</title>
        <script>
          async function fetchTime() {
            const response = await fetch('/current_time');
            const data = await response.json();
            document.getElementById('time').innerText = data.time;
          }

          setInterval(fetchTime, 1000);
        </script>
      </head>
      <body>
        <h1>Current Time</h1>
        <div id="time"></div>
      </body>
      </html>
    HTML
  end

  def current_time
    render json: { time: Time.current.to_s }
  end
end

require "rspec/core"
require "capybara/cuprite"
require "capybara/rails"
require "capybara/rspec"

Capybara.javascript_driver = :cuprite
Capybara.register_driver(:cuprite) do |app|
  Capybara::Cuprite::Driver.new(app, window_size: [1200, 800])
end

RSpec.describe "TestApp", type: :feature do
  it "displays current time", js: true do
    visit "/"

    expect(page).to have_content "Current Time"
    expect(page).to have_content(Time.current + 2.seconds).to_s
  end

  it "waits if any additional request is being made" do
    puts "Checking"
    sleep 5
  end

  private

  def app
    Rails.application
  end
end

def wait_for_any_additional_request
  puts "Waiting to see if there are additional requests being made"
  sleep 5
end

# at_exit { wait_for_any_additional_request }
RSpec::Core::Runner.autorun
```

I tried adding a `binding.irb` on line 138 of `lib/capybara/session.rb` and some `puts "#{Time.current.iso8601(3)} Accessing #{uri}"` on line 13 of `lib/capybara/server/middleware.rb` of capybara-3.40.0. I did see additional requests being made with our code I was debugging this issue when it was occurring, but I don't with the simple script above.

I'm not sure how our setup differs from the script above and why I can't reproduce it. If you have any ideas?

PR #16334, opened 2024-08-05
[#55299] Project attributes: Fix phrasing of "required" and "visible" fields

dombesz:
Update custom field attribute names and descriptions for `required` and `visible`. Change the `visible` boolean field to `admin_only` and flip the logic around the flag.

  - Rename the text in the translation files.
  - Create a migration that renames `visible` to `admin_only` and flip the boolean values also flipping the default value from true to false.
  - Replace the `visible` occurences in the code with `admin_only`.

# Ticket
https://community.openproject.org/wp/55299
Also addresses https://community.openproject.org/wp/56897
# Merge checklist

- [X] Added/updated tests
- [X] ~~Tested major browsers (Chrome, Firefox, Edge, ...)~~

### DIFF
--- a/spec/migrations/rename_visible_to_admin_only_in_custom_fields_spec.rb
+++ b/spec/migrations/rename_visible_to_admin_only_in_custom_fields_spec.rb
@@ -51,6 +51,13 @@ RSpec.describe RenameVisibleToAdminOnlyInCustomFields, type: :model do
         .to change { CustomField.first.attributes.slice("visible", "admin_only") }
         .from("visible" => false)
         .to("admin_only" => true)
+
+      # it changes the default value to false
+      ProjectCustomField.reset_column_information
+      custom_field = ProjectCustomField.new
+      custom_field.save(validate: false)
+
+      expect(custom_field.admin_only).to be false
     end
   end
 
@@ -64,6 +71,13 @@ RSpec.describe RenameVisibleToAdminOnlyInCustomFields, type: :model do
         .to change { CustomField.first.attributes.slice("visible", "admin_only") }
         .from("admin_only" => false)
         .to("visible" => true)
+
+      # it changes the default value to tru
+      ProjectCustomField.reset_column_information
+      custom_field = ProjectCustomField.new
+      custom_field.save(validate: false)
+
+      expect(custom_field.visible).to be true
     end
   end
 end
--- a/spec/migrations/rename_visible_to_admin_only_in_custom_fields_spec.rb
+++ b/spec/migrations/rename_visible_to_admin_only_in_custom_fields_spec.rb
@@ -1,0 +1,69 @@
+#-- copyright
+# OpenProject is an open source project management software.
+# Copyright (C) 2012-2024 the OpenProject GmbH
+#
+# This program is free software; you can redistribute it and/or
+# modify it under the terms of the GNU General Public License version 3.
+#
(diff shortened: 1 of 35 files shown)


cbliard (commented):
Is it ok if renaming also happens for other custom fields?

"Required for all projects" sounds a bit off: for instance for user custom fields and groups custom fields, the custom field is not related to any projects. Same for Time Spent custom fields.

image

For work package custom fields, it's also strange to have "Required for all projects" and "For all projects" at the same time, as it's possible to have it required and being disabled for some projects.

And the description for "Searchable" feels wrong too as it is specific to project attributes / project custom fields.

Maybe the renaming should be limited to project attributes?

cbliard:
It also has an impact on the table view of work package custom fields

image

cbliard (changes requested):
The admin-only attribute can not be saved.

[admin-only attribute not kept.webm](https://github.com/user-attachments/assets/c03d0612-2f98-4a0e-90f2-dce6c473d707)

`:admin_only` is missing in permitted params in `app/models/permitted_params.rb:462`.

spec/migrations/rename_visible_to_admin_only_in_custom_fields_spec.rb
```diff
@@ -0,0 +1,69 @@
...
+    subject { ActiveRecord::Migration.suppress_messages { described_class.new.migrate(:up) } }
+
+    it "changes the visible field to admin_only and flips the value" do
+      ProjectCustomField.new(visible: false).save(validate: false)
+
+      expect { subject }
+        .to change { CustomField.first.attributes.slice("visible", "admin_only") }
+        .from("visible" => false)
+        .to("admin_only" => true)
+    end
```
cbliard: I really like how this test is written. Good job 👍

cbliard (commented):
Thanks.

It seems to work well, except the field admin_only which can not be saved. Other than that it matches the spec.

The renaming seems confusing for custom fields that are not project custom fields. I'd like to have @psatyal point of view about it.

dombesz:
> Thanks.
> 
> It seems to work well, except the field admin_only which can not be saved. Other than that it matches the spec.
> 
> The renaming seems confusing for custom fields that are not project custom fields. I'd like to have @psatyal point of view about it.

Thanks @cbliard , I agree the renaming seems to be tailored to the project attributes. One option could be to apply the renaming only on the project attributes, and not all the custom fields. cc @psatyal .

cbliard (approved):

cbliard:
> I agree the renaming seems to be tailored to the project attributes. One option could be to apply the renaming only on the project attributes, and not all the custom fields.

Yep, it looks like a good option. Parimal is back tomorrow.

cbliard:
> I agree the renaming seems to be tailored to the project attributes. One option could be to apply the renaming only on the project attributes, and not all the custom fields.

@dombesz I talked about it during today's frontend meeting. We agreed upon this:
* Apply renaming of "required" and "visible" attribute label and description only for project attributes (so it should be reverted back for all other custom fields)
* Apply renaming of "visible" attribute for all custom fields (which is user and project, so it's 👌 like it is right now). This will also address and fix https://community.openproject.org/projects/openproject/work_packages/56897/activity
* Target 14.5 instead of 14.4, because it's not a regression introduced in 14.4, and it would be too late for QA-ing for 14.4 before the release.

I'll update the work package description to reflect this.

PR #16354, opened 2024-08-06
[#56997] file names with + not working

Kharonus:
[#56997](https://community.openproject.org/work_packages/56997)

### WAT?
- added custom unescaping rule for + character


mereghost (approved):
++++

PR #16358, opened 2024-08-07
[chore] replaced ruby constants with config yaml

Kharonus:
### WAT?

- nextcloud integration dependencies are stored in yaml config file
- this way, the version numbers can be parsed automatically in the docs


as-op (approved):
👍  Super! I will teach our sitebuilder scripts to read and replace references in the docs with these values.
e.g. `[this markdown link gets replaces with the value of the yml field](modules/storages/config/nextcloud_dependencies.yml#dependencies.integration_app.min_version)`

PR #16361, opened 2024-08-07
[#56780] Avoid crashing the page on access token create

judithroth:
# Merge checklist

- [x] Added/updated tests
- [x] Added/updated documentation in Lookbook (patterns, previews, etc) => no new patterns
- [x] Tested major browsers (Chrome, Firefox, Edge, ...) => Tested in Chrome and Firefox

https://community.openproject.org/work_packages/56780

The previous behaviour was like this:
If the API Token Creation was not successful, the form in the dialog was
reloaded via turbo with error messages / form validations.
If the API Token Creation was successful, a full page reload was triggered
(so that the flash was shown and the table was updated).
Unfortunately, Turbo  does not recognize the full page reload and therefore
does not load and initialize javascript linked in the page - so our Angular
App did not get initialized.

Since people were unhappy about the flash in general (because it's easy
to be missed by the user who has to copy the new token), it was decided
to switch to using a modal to avoid the full page reload.

---

This is now the MR for only the fix of the actual bug, without the new "copy to clipboard" button, because there is still need of clarification for that.


akabiru (approved):
:shipit:

PR #16363, opened 2024-08-07
[#55299] Update Project attributes searchable hint text

dombesz:
Follow up modification for the WP:
https://community.openproject.org/work_packages/55299

Added the missing project attributes searchable field hint text.


cbliard (changes requested):
Nice 👍 

I noticed a regression with regards to the previous PR: the hint text for "Admin-only" should be the same for project and user custom fields. It should be "Check to make this attribute only visible to administrators. Users without admin rights will not be able to view or edit it.", but currently it's not the correct one on project attributes page:

image

Once this is fixed, it will be good to merge.

PR #16364, opened 2024-08-07
[#48274] Work packages export modal with settings & Job status modal

as-op:

* Turbofying and primerizing the export modal dialog
* New: Export options are no longer individual entries, but settings via form elements
* New: Column selection of the Query can be overwritten in the export modal
* New: Long text fields included in the PDF Report can be selected
* New: Atom is removed from the export modal
* New: PDFs are opened in a new tab (if supported by browser)
* Turbofying and primerizing the job status dialog 

## Screenshots

| Before      | After      |
| ------------- | ------------- |
| before |after |
|before-status|after-status|

* Form fields are included for all formats, hidden/shown by stimulus
* Existing wp export endpoint is used (GET)
* BCF download used the previous Angular export modal to display the job progress dialog (not the selection), it does now directly
* Replaced all usage of the Angular Job Modal with the turbo version
* removed from Angular: Export Modal, Job Modal and JobPage

# Ticket

https://community.openproject.org/work_packages/48274

https://www.figma.com/file/n0uj7NMxC4Q2nAzMyhdBXF/Export-work-packages-table?type=design&node-id=0-1&mode=design

# Merge checklist

- [x] Added/updated tests


app/components/work_packages/exports/modal_dialog_component.html.erb
```diff
@@ -0,0 +1,59 @@
...
+          export_formats_settings.each do |format|
+            component.with_item(
+              label: format[:label], icon: format[:icon], selected: format[:selected],
+              data: {
+                "work-packages--export--dialog-format-param": format[:id],
+                action: "click:segmented-control#select click->work-packages--export--dialog#formatChanged"
+              })
+          end
+        end %>
+        <%= render Primer::BaseComponent.new(tag: :hr, mt: 4, mb: 4) %>
```
HDinger: This line is duplicated multiple times, either with the exact same spacings or slighlty different. I think it makes sense to extract that a bit. Otherwise, we have to look for all places again, when the spacing should change.

 I see two possible ways:

* Either you extract that into a component `OpenProject::Common::Divider` with `4` as default spacing but make it configurable.
* Or you extract at least the `4` into a variable that you use everywhere

app/components/work_packages/exports/pdf/export_settings_component.html.erb
```diff
@@ -0,0 +1,36 @@
+<%=
+  flex_layout(data: {
+    "application-target": "dynamic",
+    controller: "work-packages--export--pdf--settings"
+  }) do |container| %>
+  <%= container.with_row do %>
+    <%= render(Primer::Alpha::RadioButtonGroup.new(
+      mb: 3,
```
HDinger: I guess that can be removed, as the divider takes care of spacing?

app/components/work_packages/exports/pdf/export_settings_component.rb
```diff
@@ -0,0 +1,71 @@
...
+            { value: "table",
+              label: I18n.t("export.dialog.pdf.export_type.options.table.label"),
+              caption: I18n.t("export.dialog.pdf.export_type.options.table.caption"),
+              component: WorkPackages::Exports::PDF::Table::ExportSettingsComponent },
+            { value: "report",
+              label: I18n.t("export.dialog.pdf.export_type.options.report.label"),
+              caption: I18n.t("export.dialog.pdf.export_type.options.report.caption"),
+              component: WorkPackages::Exports::PDF::Report::ExportSettingsComponent },
+            { value: "gantt",
+              label: (I18n.t("export.dialog.pdf.export_type.options.gantt.label") + enterprise_icon).html_safe,
```
HDinger: The check for the EE token is missing here. Currently, the icon is always displayed:
Bildschirmfoto 2024-08-22 um 10 55 41

app/components/work_packages/exports/pdf/report/export_settings_component.html.erb
```diff
@@ -0,0 +1,38 @@
+<%= flex_layout do |container| %>
+  <% container.with_row do |_columns| %>
+    <%= render WorkPackages::Exports::ColumnSelectionComponent.new(
+      query,
+      "columns-select-export-pdf-report",
+      I18n.t("export.dialog.columns.input_caption_report")
```
HDinger, line 6: In the visuals, there is a different text shown:
**Visuals**
Bildschirmfoto 2024-08-22 um 11 30 58

**Code**
Bildschirmfoto 2024-08-22 um 10 57 06

frontend/src/stimulus/controllers/dynamic/work-packages/export/form.controller.ts
```diff
@@ -0,0 +1,76 @@
...
+  submitForm(evt:CustomEvent) {
+    evt.preventDefault(); // Don't submit
+    const formatURL = this.element.getAttribute('action');
+    const searchParams = this.getExportParams();
+    const exportURL = `${formatURL}?${searchParams.toString()}`;
+    this.requestExport(exportURL)
+      .then((job_id) => this.showJobModal(job_id))
+      .catch((error) => {
+        // TODO: error handling
+        console.error(error);
```
HDinger: Is that something we should take care of?

modules/job_status/app/components/job_status/dialog/body_component.rb
```diff
@@ -0,0 +1,122 @@
...
+
+      def mime_type_pdf?
+        mime_type == "application/pdf"
+      end
+
+      def icon
+        return { icon: :alert, classes: "color-fg-danger" } if job.nil?
+        return { icon: :"x-circle", classes: "color-fg-danger" } if has_error?
+
+        { icon: :"issue-closed", classes: "color-fg-success" } if success_statuses.include?(job.status)
```
HDinger: Please do not set the classes directly, but rather use the system_arguments for that, e.g. `icon: :alert, color: :danger` and in the HTML you should be able to simply pass the hash without having to deconstruct it again

HDinger (changes requested):
Great job @as-op ! 👍 I really like how you structured the components. That was really easy to read and to understand.
I have some minor remarks, but nothing too critical:

- [ ] The loading modal has a different text then specified in the visuals. 

Bildschirmfoto 2024-08-22 um 11 50 32


- [ ] The loading animation looks somehow interrupted. Instead of one smooth animation, it breaks and seems to start from the beginning:

**Export loading**
Aug-22-2024 11-53-53

**For comparision: the WP table loading**

Aug-22-2024 11-54-42


- [ ]  On the success and the error modal, the close button is missing
Bildschirmfoto 2024-08-22 um 11 57 04

- [ ] I somehow managed to produce an error, by adding all available columns. However, there was no error message shown.
Bildschirmfoto 2024-08-22 um 11 04 56

Bildschirmfoto 2024-08-22 um 11 05 29

- [ ] The error message looks different from what is specified in Figma:The icon is different, and in Figma, there is a generic large message "Something went wrong" and the details in the description below

**Expected**
Bildschirmfoto 2024-08-22 um 12 00 39


**Actual**

Bildschirmfoto 2024-08-22 um 11 53 21

modules/job_status/app/components/job_status/dialog/body_component.rb
as-op: Thank you! Fixed.

app/components/work_packages/exports/pdf/report/export_settings_component.html.erb
as-op: Thank you! Fixed.

app/components/work_packages/exports/pdf/export_settings_component.rb
as-op: Good chat! Thank you! Fixed.

app/components/work_packages/exports/pdf/export_settings_component.html.erb
as-op: Indeed. Plus the margin was not applied by the RadioButtonGroup. Removed.

app/components/work_packages/exports/modal_dialog_component.html.erb
as-op: You are right, I've added [OpenProject::Common::Divider](https://github.com/opf/openproject/blob/8e7a9ceea5bec6e848d83a452a5017e88d4e9d95/app/components/open_project/common/divider_component.rb)
and a [lookbook preview](
https://github.com/opf/openproject/blob/b7c6d4b4c01e2ffa4d2dbf9ccb913d8c33157c7a/lookbook/previews/open_project/common/divider_preview.rb). Looks way better.

frontend/src/stimulus/controllers/dynamic/work-packages/export/form.controller.ts
as-op: Yes, thank you! Fixed now. There is a toast message now if the call for job modal fails.

as-op:
Thanks a lot for the very good feedback and detailed report, @HDinger!

I addresses & fixed nearly all of them, except the very first.  

> * [ ] The loading modal has a different text then specified in the visuals.

A background job status only has one message while processing, and that is already used by „Work packages are being exported“.
As the job status modal is used by various background jobs, e.g. copy project, create backup, and more, we need to keep it generic (and IMO the effort to change that is too much for now)

> * [x]  The loading animation looks somehow interrupted. Instead of one smooth animation, it breaks and seems to start from the beginning
> * [x]   On the success and the error modal, the close button is missing
> * [x]  I somehow managed to produce an error, by adding all available columns. However, there was no error message shown.
> * [x]  The error message looks different from what is specified in Figma:The icon is different, and in Figma, there is a generic large message "Something went wrong" and the details in the description below

I keep trying to reproduce the column error you had, because that looks suspicious and should® not happen at all.

Thanks again! 💙

Update: Found the bug. Column selection shows all available columns, including custom fields that are not valid in a project scoped query. I will filter them out.
UpdateUpdate: Fixed

app/components/work_packages/exports/pdf/gantt/export_settings_component.rb
```diff
@@ -0,0 +1,93 @@
...
+              { label: "A4", value: "A4", default: true },
+              { label: "A3", value: "A3" },
+              { label: "A2", value: "A2" },
+              { label: "A1", value: "A1" },
+              { label: "A0", value: "A0" },
+              { label: "Executive", value: "EXECUTIVE" },
+              { label: "Folio", value: "FOLIO" },
+              { label: "Letter", value: "LETTER" },
+              { label: "Tabloid", value: "TABLOID" }
+            ]
```
ulferts, line 87: ❓  I never thought about this but do those written out formats need to be I18ned? From my brief browsing, I would think that it is not necessary.

app/controllers/work_packages_controller.rb
```diff
@@ -84,13 +88,25 @@ def show
...
+
   protected
 
+  def load_and_authorize_in_optional_project_for_export
+    load_and_authorize_in_optional_project
+  end
+
+  def load_and_validate_query_for_export
+    load_and_validate_query
+  end
```
ulferts, line 103: ❓ Why this masking of methods?

config/locales/js-en.yml
```diff
@@ -378,19 +378,6 @@ en:
...
-        button_export: "Export"
-        column_widths:
-          narrow: "Narrow"
-          medium: "Medium"
-          wide: "Wide"
-          very_wide: "Very wide"
-        options:
-          date_zoom: "Date zoom"
-          paper_size: "Paper size"
-          column_widths: "Column widths"
```
ulferts, line 392: 👍 for cleaning up.

frontend/src/stimulus/controllers/dynamic/job-status-polling.controller.ts
```diff
@@ -0,0 +1,46 @@
...
+    this.stopPolling();
+    this.hideProgressIndicator();
+  }
+
+  stopPolling() {
+    clearInterval(this.interval);
+  }
+
+  hideProgressIndicator() {
+    const node = document.querySelector('#job-status-modal-dialog .op-loading-indicator');
```
ulferts: 💡 You could turn the loading indicator into a target as well. That way, it could more easily be accessed.

modules/job_status/app/components/job_status/dialog/dialog_component.html.erb
```diff
@@ -0,0 +1,22 @@
+<%= render(Primer::Alpha::Dialog.new(
+  title: I18n.t('job_status_dialog.title'),
+  visually_hide_title: true,
+  size: :large,
+  id: MODAL_ID
+)) do |dialog| %>
+  <% dialog.with_body do %>
+    <div class="op-loading-indicator" style="margin-bottom: -20px">
```
ulferts: Why not move this inline style into sass?

modules/job_status/config/routes.rb
```diff
@@ -1,4 +1,6 @@
 Rails.application.routes.draw do
+  get "/job_statuses/dialog/:job_uuid", to: "job_statuses#dialog", as: "job_status_dialog"
+  get "/job_statuses/dialog/:job_uuid/body", to: "job_statuses#dialog_body", as: "job_status_dialog_body"
   get "/job_statuses/:job_uuid",
```
ulferts, line 4: ❓ why not draw the routes as 

`job-statuses/:job_uuid/dialog` and `job-statuses/:job_uuid/dialog_body` or `job-statuses/:job_uuid/dialog/body`. 

It looks a bit off that the three routes don't follow the same pattern.

ulferts (approved):
I pulled again shortly before writing this, @as-op . The latest commits seem to have fixed the issue about the d&d on the column selection. As such, I only have comments now and the PR can be merged in my opinion.

I still think though, that additional work would help improving the export modal as when using it, I found a couple of UI/UX issues (in my opinion):

* The text below the "Add column text" states that I cannot add long text fields in table view. But when switching to Report, that limitation is no longer mentioned. But I still cannot add the long text columns there but need to scroll down (which is easy to overlook) and then add the long text fields as a separate element. A bit of explanatory text, also as to that I cannot have the long text fields mixed with the other columns is missing.
* The subtext below the "Add columns" makes it harder to understand (for me) that the next element in which the columns can be sorted are affected by it as well and that the two belong together. Would it be possible to have a section heading? Ideally we would have a selection that also allows to sort the elements but I understand that this is out of scope.
* The ng-select of the "Add columns" are oftentimes cut off and require the user to scroll to see all options. (e.g. CSV format)
image

* The options to export "Table", "Report" and "Gantt chart" are the same for both the "work packages" as well as for the "Gantt charts" module. That is confusing to me as I would have expected to only have the "Gantt chart" option when exporting in "Gantt" (or as it is the only option then, no option at all). And when exporting in "Work packages", I would have expected to only have "Table" and "Report". From my personal perspective, I would much rather we readd the option to activate the Gantt chart to the work package list as well but the current decision is towards the other direction.

My review focuses on the backend as I understand that @HDinger already covered the frontend.

modules/job_status/config/routes.rb
as-op: Thanks! You're right, I unified them as you suggested.

modules/job_status/app/components/job_status/dialog/dialog_component.html.erb
as-op: Done. I was not sure about it and then forgot. It's now extracted.

frontend/src/stimulus/controllers/dynamic/job-status-polling.controller.ts
as-op: The indicator was outside the controller element, I moved that to the dialog, so now it's a target. Much better. Thanks!

app/controllers/work_packages_controller.rb
as-op: It’s because I was not able to merge these two controller callbacks into one 
```
before_action :load_and_validate_query, only: :index, unless: -> { request.format.html? }
before_action :load_and_validate_query, only: :export_dialog, if: -> { request.format.html? }
```

but that did not work as calling the same filter with different options is [not supported](https://guides.rubyonrails.org/v7.1.3/action_controller_overview.html#filters).  
So, out of ideas I masked it. 
```
before_action :load_and_validate_query, only: :index, unless: -> { request.format.html? }
before_action :load_and_validate_query_for_export, only: :export_dialog, if: -> { request.format.html? }
```

Happy to do it differently, though.

app/components/work_packages/exports/pdf/gantt/export_settings_component.rb
as-op: Yes, I was asking this myself, too. But I opted for not translating, as it may be confusing to have the American special sizes translated. e.g. "Letter" vs. German "Brief"

as-op:
Hi @ulferts, I can agree with the UX remarks. I will copy the points to the work package to include the product team into the discussion.

I leave out the following 
> The ng-select of the "Add columns" are oftentimes cut off and require the user to scroll to see all options.

I talked with Oliver about it, but this is due to Primer dialogs using a different overlay technique (not z-index). Therefore the dropdown has to be attached to the dialog, resulting in extending the inner scroll area. If I remember right - in other dialogs with this situation, we scroll down on open, so it's visible. But I found that jumping to be equally unsatisfying. So I postponed the topic until ng-select can be changed/replaced.

Thanks a lot for the review!

app/controllers/work_packages_controller.rb
ulferts: I think you don't really need the 

`if: -> { request.format.html? }`

at the 

`before_action :load_and_authorize_in_optional_project_for_export, only: :export_dialog, if: -> { request.format.html? }` 

before_action declaration, because I wouldn't expect any other format to be requested of the action. But I might be missing something. To me it looks as if it would even be better to have the authorization checked regardless of the format.

In case I am not, then at least this statement and the one for index can simply be merged from:

```
before_action :load_and_authorize_in_optional_project,
              :check_allowed_export,
              :protect_from_unauthorized_export, only: :index
                
before_action :load_and_authorize_in_optional_project_for_export, only: :export_dialog, if: -> { request.format.html? }
```

to

```
before_action :load_and_authorize_in_optional_project,
                only: %i[index export_dialog]
before_action :check_allowed_export,
                :protect_from_unauthorized_export, only: :index
```

But the other one, `load_and_validate_query_for_export`, is admittedly tricky. On that action I again don't think that you need the `if: -> { request.format.html? }` but that does not help a lot.

One could merge the statements again but only at the expense of a very ugly if/unless statement:

```
before_action :load_and_validate_query, only: %i[index export_dialog], if: -> { params[:action] == "index" && !request.format.html? }
```

So there it is probably better to just stick to the masking.

PR #16366, opened 2024-08-07
Introduce RemoteIndentity to decouple remote user ids from oauth tokens

mereghost:
## Why

Right now, every time a OAuth Token is removed the user might lose access to any AMPF enabled project storages.

Of course, that's an unintended side effect from the way we deal with our data.

## What

`OAuthClientToken`s have a property called `origin_user_id` that we use to set permissions on remote folders for any given storage.

The same tokens can be revoked and removed for a bunch of different reasons and this can end up with users losing access to the project folders on the storage provider.

This PR introduces the missing abstraction that will deal with this issue.

## How

This PR introduces a new model called `RemoteIdentity` that links a `User` to an `OAuthClient` and contains an `origin_user_id`.

The idea is that this model will always hold the latest known remote identity for a given user on a oauth client, so that even if the token is gone, we can still know who this `User` is on the `storage provider` and therefore add/remove them from AMPF folders.

## Needed steps
- [x] Create RemoteIndentity
- [x] Update the ConnectionManager to create the RemoteIdentity
- [x] Update the Sync Jobs to use it
- [x] Update the Commands and Queries that rely on `origin_user_id`
- [x] Update the `Connected Users` list to rely on RemoteIdentity rather than tokens
- [x] Introduce an event on RemoteIdentity creation
- [ ] Introduce an event on RemoteIdentity removal (it is redundant but why not?)
- [x] Create a migration to move the identifier from `OAuthClientToken` to `RemoteIdentity`
- [ ] Create a migration to drop the `origin_user_id` column from `OAuthClientToken`
- [x] Quick sanity check through the UI

### DIFF
--- a/app/models/remote_identity.rb
+++ b/app/models/remote_identity.rb
@@ -2,7 +2,7 @@
 
 #-- copyright
 # OpenProject is an open source project management software.
-# Copyright (C) 2012-2024 the OpenProject GmbH
+# Copyright (C) the OpenProject GmbH
 #
 # This program is free software; you can redistribute it and/or
 # modify it under the terms of the GNU General Public License version 3.
@@ -28,11 +28,10 @@
 # See COPYRIGHT and LICENSE files for more details.
 #++
 
-FactoryBot.define do
-  factory :oauth_client_token, class: "::OAuthClientToken" do
-    sequence(:access_token) { |n| "1234567890-#{n}" }
-    sequence(:refresh_token) { |n| "2345678901-#{n}" }
-    oauth_client factory: :oauth_client
-    user factory: :user
-  end
+class RemoteIdentity < ApplicationRecord
+  belongs_to :user
+  belongs_to :oauth_client, class_name: "OAuthClient"
+
+  validates :user, uniqueness: { scope: :oauth_client }
+  validates :origin_user_id, :user, :oauth_client, presence: true
 end
--- a/app/services/oauth_clients/connection_manager.rb
+++ b/app/services/oauth_clients/connection_manager.rb
@@ -59,7 +59,6 @@ module OAuthClients
     end
 
     # rubocop:disable Metrics/AbcSize
-
     # The bearer/access token has expired or is due for renew for other reasons.
     # Talk to OAuth2 Authorization Server to exchange the renew_token for a new bearer token.
     def refresh_token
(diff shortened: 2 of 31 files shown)


app/models/remote_identity.rb
```diff
@@ -0,0 +1,37 @@
...
+#
+# See COPYRIGHT and LICENSE files for more details.
+#++
+
+class RemoteIdentity < ApplicationRecord
+  belongs_to :user
+  belongs_to :oauth_client, class_name: "OAuthClient"
+
+  validates :user, uniqueness: { scope: :oauth_client }
+  validates :origin_user_id, :user, :oauth_client, presence: true
```
Kharonus, line 36: 🟡 should we create a very simple contract for the validations? So that we keep the validations out of the ApplicationRecord.

app/services/oauth_clients/connection_manager.rb
```diff
@@ -104,28 +103,18 @@ def code_to_token(code)
...
-        return ServiceResult.failure(errors: oauth_client_token.errors) unless oauth_client_token.valid?
+        OAuthClientToken.transaction do
+          oauth_client_token = create_client_token(rack_access_token)
 
-        OpenProject::Notifications.send(
-          OpenProject::Events::OAUTH_CLIENT_TOKEN_CREATED,
-          integration_type: @oauth_client.integration_type
-        )
-      end
+          oauth_client_token.save!
```
Kharonus, line 108: 🔴 we need to rollback on a failure here, too. Or is this, what the `!` is for?

Kharonus (changes requested):
Very good work in a short time. Looks pretty solid, some smaller requests.

db/migrate/20240808125617_populate_remote_identities.rb
```diff
@@ -0,0 +1,55 @@
...
+
+    OAuthClientToken.where.not(origin_user_id: nil)
+                    .select(:id, *fields)
+                    .find_in_batches do |batch|
+      identities = batch.map { |record| record.slice(*fields) }
+
+      RemoteIdentity.insert_all(identities, unique_by: %i[user_id oauth_client_id])
+    end
+
+    OAuthClientToken.update_all(origin_user_id: nil)
```
Kharonus, line 41: ❓ why not dropping the column? And adding it in `down`?

app/services/oauth_clients/connection_manager.rb
mereghost: `save!` raises an error when saving, thus rolling back the transaction.

db/migrate/20240808125617_populate_remote_identities.rb
mereghost: No special reason, but I prefer to make field removal into their own migration.

app/models/remote_identity.rb
mereghost: I could. I just avoided it as it was simpler and faster this way and I was running against time here.

Kharonus: As you created services already, dropping a create contract should be easy enough.

app/services/oauth_clients/connection_manager.rb
Kharonus: alright, cool. then this turn to 🟢

db/migrate/20240808125617_populate_remote_identities.rb
Kharonus: alright, especially with wieland latest message make sense to keep it separated.

app/models/remote_identity.rb
mereghost: Thing is that I dodged `BaseService` because I don't need half of it and the service works like an `Upsert`, so it means a lot more changes to it. =)

PR #16376, opened 2024-08-08
[#57052] Add optimistic locking to oauth_client_tokens

ba1ash:

I try to prevent race condition on `oauth_client_token` updates. This race condition can lead to an inconsistent state when OP has already dead refresh_token which NC does not accept. Then user has to go through OAuth grant flow again. 

I enable optimistic locking for `oauth_client_tokens`. It prevents a race condition for update and destroy calls. `ActiveRecord::StaleObjectError` is raised then. The exception is caught and retry happens once. 
https://api.rubyonrails.org/v7.1.3.4/classes/ActiveRecord/Locking/Optimistic.html
image

# Ticket

https://community.openproject.org/wp/57052
# Merge checklist

- [ ] Added/updated tests

### DIFF
--- a/app/services/oauth_clients/connection_manager.rb
+++ b/app/services/oauth_clients/connection_manager.rb
@@ -33,7 +33,6 @@ require "uri/http"
 
 module OAuthClients
   class ConnectionManager
-    # Nextcloud API endpoint to check if Bearer token is valid
     TOKEN_IS_FRESH_DURATION = 10.seconds.freeze
 
     def initialize(user:, configuration:)
--- a/db/migrate/20240808133947_add_lock_version_to_oauth_client_tokens.rb
+++ b/db/migrate/20240808133947_add_lock_version_to_oauth_client_tokens.rb
@@ -1,0 +1,35 @@
+# frozen_string_literal:true
+
+#-- copyright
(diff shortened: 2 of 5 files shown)


mereghost (approved):
LGTM!

Raising the error for now is good enough. This will allow us to double check if that's the main source of the problem and then address it for next release.

wielinde:
Meeeeh, what shall a user do then? That problem is not created by the user. Can't be solved by the user. And it is fixable by us. We serve the users. Let's fix it for the users. And yes, it takes longer to build.

ba1ash:
> Meeeeh, what shall a user do then? That problem is not created by the user. Can't be solved by the user. And it is fixable by us. We serve the users. Let's fix it for the users. And yes, it takes longer to build.

Sure, let's do it.

mereghost:
> Meeeeh, what shall a user do then? That problem is not created by the user. Can't be solved by the user. And it is fixable by us. We serve the users. Let's fix it for the users. And yes, it takes longer to build.

Sure, my point was: we aren't even sure that this is the root cause right now. We might be fixing a straw man.

Ideally we could deploy this fix on a live, heavily used system like community as a step to diagnose the issue, while not being part of the release.

PR #16381, opened 2024-08-09
[#56780] Add copy to clipboard button for newly created API token

judithroth:
https://community.openproject.org/work_packages/56780/activity

Actually, the whole layout of the dialog was changed, because of UX
implications of having a copy to clipboard button and to make the
user more aware that he has something to do.

From a technical side, I switched from using a blankslate to using
a flex layout, because the blankslate would enforce the text to be
centered and the input label and warning are left aligned and do not
look good when being centered. I also tried having them below the
blankslate (so the blankslate would only have an icon and an heading)
but then there's a large margin between the heading and the rest of
the content.

# Merge checklist

- [x] Added/updated tests
- [x] Added/updated documentation in Lookbook (patterns, previews, etc) => no new patterns
- [x] Tested major browsers (Chrome, Firefox, Edge, ...) => Tested in Chrome

### DIFF
--- a/app/components/my/access_token/access_token_created_dialog_component.html.erb
+++ b/app/components/my/access_token/access_token_created_dialog_component.html.erb
@@ -34,10 +34,35 @@ See COPYRIGHT and LICENSE files for more details.
     size: :large
   )) do |dialog|
     dialog.with_body do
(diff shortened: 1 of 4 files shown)


akabiru (approved):
🥇

PR #16383, opened 2024-08-09
[#55480] Fix Include subprojects return wrong total sum

Reza98Sh:
Fix bug OP#55480 Project and Include subprojects return wrong total sum values (button 'Include projects') 

# Ticket
https://community.openproject.org/work_packages/55480

I have changed the numberOfProjectsInFilter$. When this.includeSubprojects is true, it collects all subproject hrefs and updates the number.

[Screencast from 2024-08-09 15-48-28.webm](https://github.com/user-attachments/assets/4f60b47d-8515-4f5a-8acf-799801979e44)

If this method is not considered best practice, please let me know. 
I am doing my best to improve it.


Reza98Sh:
@oliverguenther
Excuse me, could you please check this and run the workflow? It seems to need approval from a maintainer. 
Thank you!

PR #16389, opened 2024-08-09
[#57038] Trigger all necessary hooks when a project is removed from a storage

judithroth:
https://community.openproject.org/work_packages/57038

When removing a project from a storage, all necessary hooks, like e.g. deleting files for automatically managed project folders, should be run. The DeleteService takes care of that.

# Ticket
https://community.openproject.org/work_packages/57038

# Merge checklist

- [x] Added/updated tests
- [x] Added/updated documentation in Lookbook (patterns, previews, etc) => nothing to do
- [x] Tested major browsers (Chrome, Firefox, Edge, ...) => Tested in Chrome

### DIFF
--- a/modules/storages/app/controllers/storages/admin/storages/project_storages_controller.rb
+++ b/modules/storages/app/controllers/storages/admin/storages/project_storages_controller.rb
@@ -109,9 +109,20 @@ class Storages::Admin::Storages::ProjectStoragesController < ApplicationControll
   end
 
   def destroy
-    Storages::ProjectStorages::DeleteService
+    result = Storages::ProjectStorages::DeleteService
       .new(user: current_user, model: @project_storage)
       .call
+
+    # rubocop:disable Rails/ActionControllerFlashBeforeRender
+    result.on_success do
+      flash[:primer_banner] = { message: I18n.t(:notice_successful_delete) }
+    end
+
+    result.on_failure do |failure|
+      error = failure.errors.map(&:message).join("; ")
+      flash[:primer_banner] = { message: t("project_storages.remove_project.deletion_failure_flash", error:), scheme: :danger }
+    end
+    # rubocop:enable Rails/ActionControllerFlashBeforeRender
 
     redirect_to admin_settings_storage_project_storages_path(@storage)
   end
--- a/modules/storages/app/controllers/storages/admin/storages/project_storages_controller.rb
+++ b/modules/storages/app/controllers/storages/admin/storages/project_storages_controller.rb
@@ -119,7 +119,7 @@ class Storages::Admin::Storages::ProjectStoragesController < ApplicationControll
     end
 
     result.on_failure do |failure|
-      error = failure.errors.map(&:message).join("; ")
+      error = failure.errors.map(&:message).to_sentence
       flash[:primer_banner] = { message: t("project_storages.remove_project.deletion_failure_flash", error:), scheme: :danger }
     end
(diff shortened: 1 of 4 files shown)


modules/storages/app/controllers/storages/admin/storages/project_storages_controller.rb
```diff
@@ -81,7 +81,21 @@ def destroy_confirmation_dialog
...
+      .call
+
+    # rubocop:disable Rails/ActionControllerFlashBeforeRender
+    result.on_success do
+      flash[:primer_banner] = { message: I18n.t(:notice_successful_delete) }
+    end
+
+    result.on_failure do |failure|
+      error = failure.errors.map(&:message).join("; ")
+      flash[:primer_banner] = { message: t("project_storages.remove_project.deletion_failure_flash", error:), scheme: :danger }
```
akabiru: 🟠 Q: Does i18n/Active model errors have special handling for `error:`? AFAICT, the `deletion_failure_flash` does not interpolate the errors and the `;` join seems odd to me.

modules/storages/spec/features/storages/admin/project_storages_spec.rb
```diff
@@ -315,13 +329,23 @@
         expect(page).to have_text(project.name)
         project_storages_index_page.click_menu_item_of("Remove project", project)
 
+        # The original DeleteService would try to remove actual files from actual storages,
+        # which is of course not possible in this test since no real storage is used.
```
akabiru, line 351: 🟢 Makes sense- good that the service is already well unit tested so we can confidently stub it out 👍🏾

modules/storages/spec/features/storages/admin/project_storages_spec.rb
```diff
@@ -315,13 +329,23 @@
...
+
         page.within("dialog") do
           expect(page).to have_button("Remove", disabled: true)
           check "Please, confirm you understand and want to remove this file storage from this project"
-          expect(page).to have_button("Remove", disabled: false, wait: 3) # ensure button is clickable
+          wait_for(page).to have_button("Remove", disabled: false) # ensure button is clickable
           click_on "Remove"
         end
 
+        expect(page).to have_text(I18n.t(:notice_successful_delete))
```
akabiru: 🟢 in a dev discussion, we settled to use strings when asserting against i18n as a better way of tracking errors in the main en translations
```suggestion
        expect(page).to have_text("Successful deletion.")
```

akabiru (approved):
:shipit: 

_A couple of non-blocking remarks_

modules/storages/spec/features/storages/admin/project_storages_spec.rb
judithroth: Oh, that one is really good to know! Because I usually also prefer using Strings but thought the "default" here was not using them.
Thanks for mentioning it!

modules/storages/app/controllers/storages/admin/storages/project_storages_controller.rb
judithroth: Good catch! Actually I tried testing this one but wasn't able to create an error via the service artificially.

modules/storages/config/locales/en.yml
```diff
@@ -65,13 +65,13 @@ en:
       inactive: No specific folder
       manual: Existing folder manually managed
     remove_project:
+      deletion_failure_flash: 'Failed to remove the project from the storage: '
```
judithroth: Thanks @mereghost :)

PR #16395, opened 2024-08-12
Update migration to use raw SQL rather than AR models

mereghost:
Updating the migration as requested by @machisuji.


machisuji (approved):
Brilliant, thank you! 👍

PR #16419, opened 2024-08-13
[52233] Fix multiple issues related to % complete being editable

cbliard:

Multiple fixes:

- [x]  **Update administration texts**

  Update some texts in administration related to % Complete now being editable.
  See this comment for more context: 
  https://community.openproject.org/projects/openproject/work_packages/52233/activity#activity-61

  The texts must be visible only if the feature flag for % complete edition is active.

- [x] **Make error messages less noisy**

  Also update some errors messages to be more relevant: if work or remaining work are negative, then it's noisy to also tell that % complete is not consistent with work and remaining work or that remaining work exceed work.

- [x] **Handle setting remaining work when % complete is 100%**

  If the input value for % Complete is 100%, Work cannot be derived (it could be any value), so Remaining work must be unset if Work is unset, or set to 0h if Work is set. An error message is displayed otherwise.

- [x] **Make errors messages consistent**

  Work, Remaining work, and % Complete are capitalized,  first letter of the error message is capitalized, the field name is not present in the error message, the field name is still present for other usages (like API).

- [x] **Detect invalid percentage value**
  
  When given percentage value can not be parsed, an error is now displayed and the other values are not derived.

- [x] **Keep work if set**

  Currently, work is derived if remaining work and % complete are provided, regardless of work being set or empty. This is not correct.

  If work is empty, it's ok to set it from derived values, but if work is set, its value does not change. Instead if % Complete value changes, Remaining work should be derived, and if Remaining work value changes, % Complete should be derived

  Some of these scenarios are written and formalized in [Testing: Work estimates and progress: Pop-over scenarios (% Complete) (#57370)](https://community.openproject.org/projects/openproject/work_packages/57370/activity).

- [x] **Derive work when switching from work-based to status-based if only remaining work is set**

  If a work package has only remaining work being set and progress calculation mode is changed from work-based to status-based, then work package gets a % complete value.

  Remaining work and % complete being both present, work has to be derived.

- [x] **Clear % Complete when switching from status-based to work-based if work is 0h**

  It existed before making % complete editable. Better fix it now.

## Screenshots

**Update administration texts**

Before:

image

After:
image

**Make error messages less noisy**

Before:

image

After:

image

**Handle setting remaining work when % complete is 100%**

Before:

image

After:

image

**Detect invalid percentage value**

Before:

image
% Complete is changed from "50%" to "invalid", but there is no error and remaining work gets derived to "10h" (because internally % complete is coerced to 0%).

After:

image
Much better!

**Update administration texts**

Only two texts have been updated: the field hint text and the warning text when switching from status-based to work-based.

The warning text when switching from work-based to status-based has not been changed.

For the Note text in the pop-over, another PR will be opened with an alternative approach.

**Make error messages less noisy**

In work package contract, skip adding more errors if already in an error state for progress values.

**Handle setting remaining work when % complete is 100%**

More `if`s and more tests to detect the special case `done_ratio == 100`. A bit similar to work being 0h and % complete cannot be derived.

**Make errors messages consistent**

Use `validation_message` parameter of the primer input field to have finer control over the message instead of relying on primer default implementation which gets the full message and changing the `format` i18n key as it can alter the experience for error messages in other context (in the API for instance).

**Detect invalid percentage value**

Introduce a `PercentageConverter` class very much like `DurationConverter`, and coerce the percentage string at the work package level by overriding `#done_ratio=`. Checking if the string looks like a percentage is done with a regexp. `PercentageConverter` converts to float so that it can be reused elsewhere. Coercing to integer is done automatically by ActiveRecord.

**Derive work when switching from work-based to status-based if only remaining work is set**

Reuse the methods already existing from `WorkPackages::Progress::MigrateValuesJob` (used for migration to 14.0.0), move them to parent class `WorkPackages::Progress::Job` so that they are available in `WorkPackages::Progress::ApplyStatusesChangeJob`. Then call them.

# Ticket

[Feature: Allow % Complete edition in work-based progress calculation mode (#52233)](https://community.openproject.org/projects/openproject/work_packages/52233/activity)

# Merge checklist

- [x] Added/updated tests

### DIFF
--- a/spec/services/work_packages/set_attributes_service/derive_progress_values_work_based_spec.rb
+++ b/spec/services/work_packages/set_attributes_service/derive_progress_values_work_based_spec.rb
@@ -110,6 +110,13 @@ RSpec.describe WorkPackages::SetAttributesService::DeriveProgressValuesWorkBased
       include_examples "update progress values", description: "keeps % complete"
     end
 
+    context "when work is changed and percent complete is set to the same value as before" do
+      let(:set_attributes) { { estimated_hours: 20.0, done_ratio: 70 } }
+      let(:expected_derived_attributes) { { remaining_hours: 6.0 } }
+
+      include_examples "update progress values", description: "derives remaining work from work and % complete"
+    end
+
     context "when both work and percent complete are unset" do
       let(:set_attributes) { { estimated_hours: nil, done_ratio: nil } }
       let(:expected_kept_attributes) { %w[remaining_hours] }
--- a/spec/services/work_packages/set_attributes_service/derive_progress_values_work_based_spec.rb
+++ b/spec/services/work_packages/set_attributes_service/derive_progress_values_work_based_spec.rb
@@ -79,31 +79,31 @@ RSpec.describe WorkPackages::SetAttributesService::DeriveProgressValuesWorkBased
       work_package.clear_changes_information
     end
 
-    context "when work is unset" do
+    context "when work is cleared" do
       let(:set_attributes) { { estimated_hours: nil } }
       let(:expected_derived_attributes) { { remaining_hours: nil } }
       let(:expected_kept_attributes) { %w[done_ratio] }
 
-      include_examples "update progress values", description: "keeps % complete, and unsets remaining work"
+      include_examples "update progress values", description: "keeps % complete, and clears remaining work"
     end
 
-    context "when remaining work is unset" do
+    context "when remaining work is cleared" do
       let(:set_attributes) { { remaining_hours: nil } }
       let(:expected_derived_attributes) { { estimated_hours: nil } }
       let(:expected_kept_attributes) { %w[done_ratio] }
 
-      include_examples "update progress values", description: "keeps % complete, and unsets work"
+      include_examples "update progress values", description: "keeps % complete, and clears work"
     end
 
-    context "when % complete is unset" do
+    context "when % complete is cleared" do
       let(:set_attributes) { { done_ratio: nil } }
       let(:expected_derived_attributes) { { remaining_hours: nil } }
       let(:expected_kept_attributes) { %w[estimated_hours] }
 
-      include_examples "update progress values", description: "keeps work, and unsets remaining work"
+      include_examples "update progress values", description: "keeps work, and clears remaining work"
     end
 
-    context "when both work and remaining work are unset" do
+    context "when both work and remaining work are cleared" do
       let(:set_attributes) { { estimated_hours: nil, remaining_hours: nil } }
       let(:expected_kept_attributes) { %w[done_ratio] }
 
@@ -117,14 +117,14 @@ RSpec.describe WorkPackages::SetAttributesService::DeriveProgressValuesWorkBased
       include_examples "update progress values", description: "derives remaining work from work and % complete"
     end
 
-    context "when both work and percent complete are unset" do
+    context "when both work and percent complete are cleared" do
       let(:set_attributes) { { estimated_hours: nil, done_ratio: nil } }
       let(:expected_kept_attributes) { %w[remaining_hours] }
 
       include_examples "update progress values", description: "keeps remaining work"
     end
 
-    context "when both remaining work and percent complete are unset" do
+    context "when both remaining work and percent complete are cleared" do
       let(:set_attributes) { { remaining_hours: nil, done_ratio: nil } }
       let(:expected_kept_attributes) { %w[estimated_hours] }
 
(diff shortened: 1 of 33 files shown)


spec/services/work_packages/set_attributes_service/derive_progress_values_work_based_spec.rb
```diff
@@ -79,45 +79,52 @@
...
-    context "when remaining work is unset" do
+    context "when remaining work is cleared" do
       let(:set_attributes) { { remaining_hours: nil } }
-      let(:expected_derived_attributes) { { done_ratio: nil } }
-      let(:expected_kept_attributes) { %w[estimated_hours] }
+      let(:expected_derived_attributes) { { estimated_hours: nil } }
+      let(:expected_kept_attributes) { %w[done_ratio] }
 
-      include_examples "update progress values", description: "keeps work, and unsets % complete"
+      include_examples "update progress values", description: "keeps % complete, and clears work"
```
dombesz, line 95: Not sure if this is a bug, but the behaviour is not entirely consistent. When both work and remaining work fields are changed, it clears the % complete instead of work.

clearing work

dombesz (approved):
Nice job Christophe! 👏

I have one observation, clearing values in a certain order feels like playing whack-a-mole. Perhaps we could keep track of which values were cleared and not saved, then keep them cleared on subsequent changes. Albeit, this goes against some of the specs we have in the progress modal spec, such as `"Case 33-14: when work and % complete are cleared, and then work"`.
Nothing to act on at the moment, just a food for thought.

<details>
<summary> (Here's a demo of the issue) </summary>

clearing values


</details>

I also left another comment, feel free to merge after checking it.

spec/services/work_packages/set_attributes_service/derive_progress_values_work_based_spec.rb
cbliard: Actually this is consistent. Well... kind-of.

In the initial implementation, some errors were displayed when all 3 fields were modified and the values were not consistent. Parimal objected that if we enter two values, then we should compute the third one, and discard the value entered by the user. Some values are more important than other: we try to keep the value of Work if it's entered, or % Complete.

Internally it tries to keep track of which values have been modified by the user.

So about you scenario, you start with all values set
Then you clear Remaining work. As a result, either % Complete or Work should be cleared. We debated this and decided that it made more sense to clear Work because they act together. That's why Work is cleared too.
Then you enter a value for Work. As a result, Remaining work is calculated and replaces the value entered.
Then you clear Remaining work again. And again, either % Complete or Work should be cleared. And this time, Work has been entered by the user which gives more importance to this value. So it's the value of % Complete which is cleared.

We can also see it like this: you clear Remaining work, and Work gets cleared as well. You think "No that's not what I want" and you put the value back in Work. Then you delete Remaining work and % Complete gets cleared. And that's what you wanted initially: keeping only Work for some reason.

Does it make more sense?

I agree that these scenarios are complicated and the logic is brittle. Other choices could have been made. That's why we documented them in the work package with all test cases.

cbliard: And about this scenario in the test case, it is correct: this is the first part of your test: 
- All values are set, when remaining work is cleared => keeps % complete, and clears work

Then in the second part, it's different as the Work is being set as well Remaining work is cleared. Both information are sent, and it always start from the values from the database (the server is stateless. The state is held in the frontend):
- All values are set, when work is changed to the same value and remaining work is cleared => % complete is cleared.

This is the spec example from line 310.

cbliard: It should also be easier to understand with the hints (from the other PR):

[demo with hints.webm](https://github.com/user-attachments/assets/2b4f5e69-1dc2-4da0-86c5-d93528cef73e)

dombesz: Thank you for the thorough explanation, it makes sense.

cbliard:
> I have one observation, clearing values in a certain order feels like playing whack-a-mole. Perhaps we could keep track of which values were cleared and not saved, then keep them cleared on subsequent changes. Albeit, this goes against some of the specs we have in the progress modal spec, such as `"Case 33-14: when work and % complete are cleared, and then work"`. Nothing to act on at the moment, just a food for thought.
> 
> (Here's a demo of the issue)
> clearing values

Actually, it does keep track of which values are entered or not, and then to prevent having an error like "% complete does not match work and remaining work" like it is currently on edge, it stops tracking some fields under some conditions.

Here when % complete and remaining work have been manually entered, and work is being entered, it stops tracking % complete so that it can be derived from work and remaining work. It makes sense when values are entered, but not when values are empty.

So I'd qualify it as a bug and will try to fix it this morning and then merge.

Thanks for finding it.

cbliard:
As discussed in another thread, I'll merge this one and fix the bug separately to be sure it makes it before the feature freeze.

PR #16423, opened 2024-08-13
[chore] introduce new interface for set permission command

Kharonus:
[#57284](https://community.openproject.org/wp/57284)

- consolidate set permission command interface and behaviour

- introduce input data definition
- add shared test examples (incomplete)
- refactored only(!) one drive command due to better existent test coverage
- nextcloud can be refactored as a next step


modules/storages/app/common/storages/peripherals/storage_interaction/inputs/set_permissions.rb
```diff
@@ -0,0 +1,69 @@
...
+          def valid?
+            file_id.is_a?(String) && file_id.present? && valid_permissions?
+          end
+
+          private
+
+          def valid_permissions?
+            permissions.is_a?(Hash) &&
+              permissions.all? do |user_id, user_permissions|
+                user_id.is_a?(Symbol) &&
```
Kharonus: This is up for discussion: Is their any opinion about pinning usage to symbols or strings?

modules/storages/app/services/storages/one_drive_managed_folder_sync_service.rb
```diff
@@ -65,21 +65,30 @@ def call
 
     private
 
+    # rubocop:disable Metrics/AbcSize
     def apply_permission_to_folders
       info "Setting permissions to project folders"
       active_project_storages_scope.includes(:project).where.not(project_folder_id: nil).find_each do |project_storage|
```
Kharonus: This is the main change, of how the interface is used.

modules/storages/app/common/storages/peripherals/storage_interaction/inputs/set_permissions.rb
mereghost: 🟡 I'd leave as strings as it will require no casting. Instead of type checking, I'd either call `stringify_keys` or `transform_keys(&:to_s)` on the permissions hash.

modules/storages/app/common/storages/peripherals/storage_interaction/inputs/set_permissions.rb
```diff
@@ -0,0 +1,69 @@
...
+
+module Storages
+  module Peripherals
+    module StorageInteraction
+      module Inputs
+        # permissions - A map of user ids to a map of all file permissions.
+        # IMPORTANT: the user ids are considered to be the ids of the remote identities.
+        #   Example:
+        #   {
+        #     "d6e00f6d-1ae7-43e6-b0af-15d99a56d4ce": {
```
mereghost: 🟢 Something about using the `origin_user_id` as a key bothers me, but since I can't put a finger on it, I'm okay with it.

modules/storages/app/common/storages/peripherals/storage_interaction/inputs/set_permissions.rb
```diff
@@ -0,0 +1,69 @@
...
+        #   }
+        SetPermissions = Data.define(:file_id, :permissions) do
+          def valid?
+            file_id.is_a?(String) && file_id.present? && valid_permissions?
+          end
+
+          private
+
+          def valid_permissions?
+            permissions.is_a?(Hash) &&
```
mereghost: 🟢 Having the `origin_user_id` as the key makes it more difficult to validate the data, otherwise this could be a good place to introduce `dry-validation`.

modules/storages/app/common/storages/peripherals/storage_interaction/one_drive/set_permissions_command.rb
```diff
@@ -45,25 +45,33 @@ def delete? = permission_ids.any? && user_ids.empty?
             def update? = permission_ids.any? && user_ids.any?
           end
 
-          def self.call(storage:, auth_strategy:, path:, permissions:)
-            new(storage).call(auth_strategy:, path:, permissions:)
+          def self.call(storage:, auth_strategy:, input_data:)
```
mereghost, line 54: We probably should add some YARD docs here on what is the `input_data`, so it makes easier for the IDEs/Editors/Our future selves

modules/storages/app/common/storages/peripherals/storage_interaction/one_drive/set_permissions_command.rb
```diff
@@ -45,25 +45,33 @@ def delete? = permission_ids.any? && user_ids.empty?
...
           def initialize(storage)
             @storage = storage
           end
 
-          def call(auth_strategy:, path:, permissions:)
+          # rubocop:disable Metrics/AbcSize
+          def call(auth_strategy:, input_data:)
             with_tagged_logger do
+              info "Validating input data"
+              validate_input_data(input_data).on_failure { return _1 }
```
mereghost: 🔴 We probably should move this to the `.call` method, so we don't even instantiate the service if `InputData` is invalid.

modules/storages/app/services/storages/one_drive_managed_folder_sync_service.rb
mereghost: 🟡 I'd keep the block variable as `origin_user_id` otherwise it can lead future devs to think it is a `User#id`.

modules/storages/app/common/storages/peripherals/storage_interaction/inputs/set_permissions.rb
mereghost: 🟢 Also, type checking gives me the creeps, but that's just me. XD

mereghost (changes requested):
I think you're on a good path, but some bits and pieces might need some polish. =)

modules/storages/app/common/storages/peripherals/storage_interaction/inputs/set_permissions.rb
Kharonus: This evolved mostly around the data model, that we have this mapping in our data. any user (through its roles) has those five permissions.

modules/storages/app/common/storages/peripherals/storage_interaction/one_drive/set_permissions_command.rb
Kharonus: Well, this would work, if we make `#call` private. And only execute the command with `.call`. Should we change that behaviour?

modules/storages/app/common/storages/peripherals/storage_interaction/one_drive/set_permissions_command.rb
Kharonus: Good idea.

modules/storages/app/common/storages/peripherals/storage_interaction/inputs/set_permissions.rb
Kharonus: I will give it a shot.

modules/storages/app/common/storages/peripherals/storage_interaction/one_drive/set_permissions_command.rb
Kharonus: validation is now handled outside

modules/storages/app/services/storages/one_drive_managed_folder_sync_service.rb
Kharonus: done

mereghost (approved):

PR #16427, opened 2024-08-13
Cleanup after Remote Identity introduction

mereghost:
Renaming methods and variables, removing columns and wishing for the best 🤞🏼

### DIFF
--- a/modules/storages/spec/factories/storage_factory.rb
+++ b/modules/storages/spec/factories/storage_factory.rb
@@ -93,7 +93,7 @@ FactoryBot.define do
     sequence(:host) { |n| "https://host#{n}.example.com/" }
 
     trait :as_automatically_managed do
-      automatically_managed { true }
+      automatic_management_enabled { true }
       username { "OpenProject" }
       password { "Password123" }
     end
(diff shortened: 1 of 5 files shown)


modules/storages/spec/factories/storage_factory.rb
```diff
@@ -144,8 +144,10 @@
                                       "MISSING_NEXTCLOUD_LOCAL_OAUTH_CLIENT_REFRESH_TOKEN"),
              token_type: "bearer")
 
-      create(:remote_identity, oauth_client: storage.oauth_client, user: evaluator.oauth_client_token_user,
```
Kharonus, line 147: 🟡 If you already did a small cleanup here, maybe you can quickly shoot at line number 96? Replace it with `automatic_management_enabled` to remove the deprecated method there?

Kharonus (approved):
LGTM, just a small plea, but it is no merge blocker.

modules/storages/spec/factories/storage_factory.rb
mereghost: got it.

PR #16429, opened 2024-08-13
Implementation/56922 trigger a login dialog when the user has no oauth access grant

akabiru:

Prompt user to login to the storage provider so that manual folder selection is possible and background jobs that set up permissions and such can find a ready state once a storage is activated for n projects.

## Screenshots

https://github.com/user-attachments/assets/12a80c12-3d20-412e-ab96-cee830ec829d

https://github.com/user-attachments/assets/9673ec59-5a81-4d7c-ae70-270c2dec69d6

Reuse the existing OAuth Access Grant Nudge flow for single projects (https://github.com/opf/openproject/pull/14792) splitting the grant success component and updating the design so that these are reusable in both scenarios.

# Ticket

https://community.openproject.org/work_packages/56922

# Merge checklist

- [x] Added/updated tests
  - Existing Flash Modal Pattern will be documented in https://community.openproject.org/wp/56496


modules/storages/app/components/storages/admin/storages/oauth_access_granted_modal_component.html.erb
```diff
@@ -0,0 +1,40 @@
+<%=
+  component_wrapper do
+    render(
+      Primer::Alpha::Dialog.new(
+        id: dialog_id,
+        title:,
+        data: {
+          "application-target": "dynamic",
+          controller: "auto-show-dialog",
```
judithroth, line 9: Did you try `open: true` like suggested [here](https://qa.openproject-edge.com/lookbook/inspect/primer/alpha/dialog/initally_open)? :slightly_smiling_face:  If you did and it did not work for you for any reason, it would probably be nice to document that use case in `auto-show-dialog.controller.ts`.

akabiru: I wasn't aware of this option, thanks! I'll update it 🙌🏾

akabiru: Ok, I tried the `open: true` option and it's a bit wonky, it pushes down the whole content, not sure why. 🤔 Seems buggy. I may have come across this issue in the past but didn't dig any further, good idea to document this in if we can't fix it now.

Screenshot 2024-08-20 at 10 07 53 AM

akabiru: Ahm nvm- there's a `position` sys arg 🤦🏾 😅

akabiru: I've logged a bug for this https://community.openproject.org/wp/57342

judithroth: Thanks for trying it. That looks quite broken. Last time I used it, it worked as it should, but then I went for `respond_with_dialog` for other reason. So I think using your Stimulus Controller is the better way to do it this time :+1:

judithroth (approved):

PR #16388, opened 2024-08-09
Feature/56573 release custom field columnsfilters in project lists to ce

toy:
# What are you trying to accomplish?
Release custom field columns/filters in project lists to Community Edition

I also made `wait_for_network_idle` and `wait_for_reload` warn including the location if they are used in specs that are not using cuprite, which currently shows 784 places.

# Ticket
[#56573](https://community.openproject.org/projects/openproject/work_packages/56573)

# Merge checklist

- [x] Added/updated tests
- [ ] Added/updated documentation in Lookbook (patterns, previews, etc)
- [ ] Tested major browsers (Chrome, Firefox, Edge, ...)

PR #16439, opened 2024-08-14
Feature/55234 always visible text filter in project list

dombesz:
Extract the name and identifier filter from the advanced filters to a permanent filter.

## Screenshots
image

The approach is to re-use the existing filtering logic, and apply it to the extra search input field. This way the new filter element integrates well into the existing filtering system, automatically saving the filter's value on the query.

# Ticket
https://community.openproject.org/wp/55234

# Merge checklist

- [x] Added/updated tests

### DIFF
--- a/spec/features/projects/projects_index_spec.rb
+++ b/spec/features/projects/projects_index_spec.rb
@@ -532,7 +532,7 @@ RSpec.describe "Projects index page", :js, :with_cuprite, with_settings: { login
         load_and_open_filters admin
 
         # value selection defaults to "active"'
-        expect(page).to have_css('li[filter-name="active"]')
+        expect(page).to have_css('li[data-filter-name="active"]')
 
         projects_page.expect_projects_listed(parent_project,
                                              child_project,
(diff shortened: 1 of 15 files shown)


spec/features/projects/projects_index_spec.rb
```diff
@@ -280,51 +280,51 @@ def load_and_open_filters(user)
...
 
-  describe "project attributes visibility restrictions" do
-    let(:user) do
-      create(:user,
-             member_with_roles: {
-               development_project => create(:existing_project_role, permissions:),
-               project => create(:existing_project_role)
-             })
-    end
+    context "for project attributes", with_ee: %i[custom_fields_in_projects_list] do
```
cbliard: I thought the enterprise action `custom_fields_in_projects_list` was removed?
https://github.com/opf/openproject/commit/792def7b90b

cbliard (approved):
Nice work 👏

spec/features/projects/projects_index_spec.rb
dombesz: We still seem to have the enterprise check with `EnterpriseToken.allows_to?(:custom_fields_in_projects_list)` in the filter component:

https://github.com/opf/openproject/blob/caf494516c086786bd3b2737eba89a9a915ed0eb/app/components/projects/projects_filters_component.html.erb#L108-L117

~~Should I remove that ?~~

Yes, removing it. ( #16388 )

PR #16441, opened 2024-08-14
Filter the project storage members to show only users

mereghost:

The Project Storage Members list was raising an exception since the inclusion of `RemoteIdentity`.

This was caused by the lack of filtering out groups and placeholder users as those cannot login on storage anyway.

Simplest fix was to change the `includes` call to a `preload` and filter the members to only `User`s.

# Ticket

Related WP: [OP#57260](https://community.openproject.org/projects/openproject/work_packages/57260/)

# Merge checklist

- [x] Added/updated tests

### DIFF
--- a/modules/storages/app/controllers/storages/project_settings/project_storage_members_controller.rb
+++ b/modules/storages/app/controllers/storages/project_settings/project_storage_members_controller.rb
@@ -41,9 +41,11 @@ class Storages::ProjectSettings::ProjectStorageMembersController < Projects::Set
   model_object Storages::ProjectStorage
 
   def index
-    @memberships = Member
+    @project_users = Member
                    .of_project(@project)
-                   .includes(principal: :remote_identities, roles: :role_permissions)
+                   .joins(:principal)
+                   .preload(roles: :role_permissions, principal: :remote_identities)
+                   .where(principal: { type: "User" })
                    .paginate(page: page_param, per_page: per_page_param)
 
     render "/storages/project_settings/project_storage_members/index"
--- a/modules/storages/spec/features/view_project_storage_members_spec.rb
+++ b/modules/storages/spec/features/view_project_storage_members_spec.rb
@@ -33,10 +33,13 @@ require_module_spec_helper
 
 RSpec.describe "Project storage members connection status view" do
   let(:user) { create(:user) }
+  let(:group) { create(:group, members: [group_user]) }
+  let(:placeholder_user) { create(:placeholder_user) }
   let(:admin_user) { create(:admin) }
   let(:connected_user) { create(:user) }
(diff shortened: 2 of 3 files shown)


modules/storages/app/controllers/storages/project_settings/project_storage_members_controller.rb
```diff
@@ -41,9 +41,11 @@ class Storages::ProjectSettings::ProjectStorageMembersController < Projects::Set
   model_object Storages::ProjectStorage
 
   def index
-    @memberships = Member
+    @project_users = Member
                    .of_project(@project)
-                   .includes(principal: :remote_identities, roles: :role_permissions)
+                   .joins(:principal)
+                   .preload(roles: :role_permissions, principal: :remote_identities)
```
akabiru, line 47: 🟢 I didn't know of [preload](https://apidock.com/rails/v5.2.3/ActiveRecord/QueryMethods/preload) the docs suggests it's similar to [includes](https://apidock.com/rails/ActiveRecord/QueryMethods/includes)? Is this just a semantic change? Or is there a perf benefit? AFAICT- the resulting query seem identical. 🤔

modules/storages/spec/features/view_project_storage_members_spec.rb
```diff
@@ -76,11 +79,16 @@
...
-        [disconnected_user, "Not connected. The user should login to the storage via the following link."]
+        [disconnected_user, "Not connected. The user should login to the storage via the following link."],
+        [group_user, "Not connected. The user should login to the storage via the following link."]
       ].each do |(principal, status)|
         expect(page).to have_css("#member-#{principal.id} .name", text: principal.name)
         expect(page).to have_css("#member-#{principal.id} .status", text: status)
       end
+
+      [placeholder_user, group].each do |principal|
+        expect(page).to have_no_css("#member-#{principal.id} .name", text: principal.name)
```
akabiru, line 90: 💯

akabiru (approved):
🥇 Nice one

modules/storages/app/controllers/storages/project_settings/project_storage_members_controller.rb
akabiru: > _Simplest fix was to change the `includes` call to a `preload` and filter the members to only `User`s._

Ah, I see you allude to this- but I'm still unclear on how that solves it, if you don't mind expounding for my benefit 😄

mereghost: `preload` always issues separate queries for the associations.

`includes` will decide how to approach depending on the final query:
1. One single query with joins
2. Multiple queries

So `includes` need to validate all the associations, which would make this break as not all principals have `remote_identities`.

PR #16449, opened 2024-08-15
[#56440] Add 🔍 to projects autocompleter

judithroth:

- Add the magnifier icon to the projects select for enabling a storage for multiple projects at once
- Fix the cursor position in relation to the magnifier icon on the projects select for project attributes
- Have a reusable implementation for both that does not copy the sass code
- Have a simple / pure CSS solution to not mess with the angular component which is used in multiple places and not only for projects

## Screenshots

Before:
Screenshot from 2024-08-15 10-46-33

After:
Screenshot from 2024-08-15 10-47-12

Have an additional CSS class that controls the icon being added, so the angular code of the component does not have to be reworked.

# Ticket

https://community.openproject.org/work_packages/56440

# Merge checklist

- [ ] Added/updated tests => not necessary, since it's only styling
- [x] Tested major browsers 
   - [x] Chrome
   - [x] Firefox
   - [x] Edge

### DIFF
--- a/modules/storages/app/forms/storages/admin/storages/add_projects_autocompleter_form.rb
+++ b/modules/storages/app/forms/storages/admin/storages/add_projects_autocompleter_form.rb
@@ -38,6 +38,7 @@ module Storages
               visually_hide_label: true,
               validation_message: project_ids_error_message,
               autocomplete_options: {
+                with_search_icon: true,
                 openDirectly: false,
                 focusDirectly: false,
                 multiple: true,
--- a/app/components/_index.sass
+++ b/app/components/_index.sass
@@ -7,5 +7,4 @@
 @import "open_project/common/submenu_component"
 @import "filter/filters_component"
 @import "projects/row_component"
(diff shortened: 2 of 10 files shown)


modules/storages/app/forms/storages/admin/storages/add_projects_autocompleter_form.rb
```diff
@@ -38,6 +38,7 @@ class AddProjectsAutocompleterForm < ApplicationForm
               visually_hide_label: true,
               validation_message: project_ids_error_message,
               autocomplete_options: {
+                with_search_icon: true,
```
judithroth, line 41: @akabiru I tried to find another way to get the option in there but miserably failed at that :see_no_evil: 
After I have been looking around, I am also not so sure we want to introduce a new way for adding options. I am not too happy about this approach but still see it as the least intrusive way.

akabiru (approved):
🚀

PR #16470, opened 2024-08-19
Bug/57316 robots follow action links unnecessarily

ulferts:
# Ticket
https://community.openproject.org/wp/57316

Avoid robots triggering exports of project lists. Also preventing the access to activities. That should already be covered by the robots.txt so this is an extra safety net.

To prepare for loading the whole page repeatedly once the projects list are powered by turbo, the three modals have been changed to be rendered asynchronously.


HDinger (approved):
LGTM 👍

PR #16484, opened 2024-08-20
[#57285] added shared examples for testing set permissions command

Kharonus:
# Ticket
[#57285](https://community.openproject.org/work_packages/57285)

- having a consistent interface for both sharepoint and nextcloud command for setting permissions
- using a consistent test interface for the testable behaviour

- added shared examples for set permission commands
- removed deprecated authentication from onedrive util
- migrated one drive specs to new shared examples


modules/storages/spec/common/storages/peripherals/storage_interaction/one_drive/set_permissions_command_spec.rb
Kharonus, line 1: this test now uses the shared examples. The previous VCR cassettes are still in use, as the overall requests didn't change.

modules/storages/spec/services/storages/project_storages/delete_service_spec.rb
```diff
@@ -145,4 +139,14 @@
...
+
+  private
+
+  def command_class_reference
+    if storage.provider_type_nextcloud?
+      Storages::Peripherals::StorageInteraction::Nextcloud::DeleteFolderCommand
+    else
+      Storages::Peripherals::StorageInteraction::OneDrive::DeleteFolderCommand
+    end
+  end
```
mereghost, line 151: Can't we use the `Registry` to solve this here?

mereghost (approved):
Just a minor point. Looks great! =)

modules/storages/spec/services/storages/project_storages/delete_service_spec.rb
Kharonus: well, honestly I'm not sure. I just fixed a rubocop here.

this is the code:

```ruby
    let(:command_double) { class_double(command_class_reference, call: ServiceResult.success) }

    before do
      Storages::Peripherals::Registry
        .stub("#{storage.short_provider_type}.commands.delete_folder", command_double)
    end
```

before it was an unverified double on `:delete_folder_command`. I understand we need the double to expect calls on it? Is this solvable with the registry alone?

Kharonus: Well, it works :D. It's just a bit weird, If I would inline it, then we would have this:

`Registry.stub("key", class_double(Registry.resolve("key"), call: Success))`

Is this weird? or fine? Asking for a friend.

mereghost: We only do `.call` over the command, right? So we could do:

```ruby
Registry.stub("key", ->(*) { Success })
```

lambdas respond to call and allow us full control. You could even set some expectations inside the lambda to validate arguments and what not.

PR #16493, opened 2024-08-21
[#57028] Fix Removal of a project leaves some js components  unusable

judithroth:
# Ticket
https://community.openproject.org/work_packages/57028

Since this is a redirect and turbo would only replace the DOM contents, some JS events would not work after the contents were replaced. They rely on event handlers being registered and currently this only happens on full DOM load.
So we don't use turbo for this request any more to avoid this situation. There's already another work package to fix the behaviour of these elements, but that would be more work, so to be sure for now we're just using the "fast" solution here.

# Merge checklist

- [ ] Added/updated tests => not necessary
- [ ] Added/updated documentation in Lookbook (patterns, previews, etc) => not necessary
- [ ] Tested major browsers (Chrome, Firefox, Edge, ...) => not necessary


akabiru (approved):
:shipit:

PR #16495, opened 2024-08-21
Implementation/57028 removal of a project leaves some angular components as not clickable

akabiru:
https://community.openproject.org/work_packages/57028

This reverts commit https://github.com/opf/openproject/commit/f3a0db2f149dfef639e88aa4500421f49e2549b2.

Update the controller to respond with partial turbo streams instead of a full page redirect with turbo enabled. This caused DOM load issues with angular components, as Turbo would not break out of it's navigation frame- to remount the angular web components. The solution here is also used in other similar pages such as project attributes.

https://github.com/user-attachments/assets/e9585878-c1f6-4a04-a29b-207946f0c005

### DIFF
--- a/modules/storages/app/components/storages/project_storages/destroy_confirmation_dialog_component.html.erb
+++ b/modules/storages/app/components/storages/project_storages/destroy_confirmation_dialog_component.html.erb
@@ -28,54 +28,68 @@ See COPYRIGHT and LICENSE files for more details.
 ++#%>
 
 <%=
-  primer_form_with(
-    model: @project_storage,
-    url: admin_settings_storage_project_storage_path(id: @project_storage),
-    data: {
-      controller: 'disable-when-checked',
-      'disable-when-checked-reversed-value': true
-    },
-    method: :delete
-  ) do |form|
-    render(Primer::Alpha::Dialog.new(
-      id: id,
-      title: nil,
-      test_selector: id,
-      size: :large
-    )) do |dialog|
-      dialog.with_body do
-        flex_layout do |layout|
-          layout.with_row(mb: 3) do
-            render(Primer::Beta::Heading.new(tag: :h1, text_align: :center)) do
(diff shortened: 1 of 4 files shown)


modules/storages/app/components/storages/project_storages/destroy_confirmation_dialog_component.html.erb
```diff
@@ -28,54 +28,68 @@ See COPYRIGHT and LICENSE files for more details.
...
+              I18n.t("button_close")
+            end
+
+            footer.with_component(
+              Primer::ButtonComponent.new(scheme: :danger, type: :submit,
+                                          test_selector: "remove-project-storage-button",
+                                          disabled: true,
+                                          data: {
+                                            "disable-when-checked-target": "effect",
+                                            "close-dialog-id": id
```
akabiru, line 90: @judithroth One odd thing is I had to specify the `close-dialog-id` - in other cases, the dialog closes automatically on submit 🤷🏾

modules/storages/app/components/storages/project_storages/destroy_confirmation_dialog_component.html.erb
```diff
@@ -28,54 +28,68 @@ See COPYRIGHT and LICENSE files for more details.
...
-      title: nil,
-      test_selector: id,
-      size: :large
-    )) do |dialog|
-      dialog.with_body do
-        flex_layout do |layout|
-          layout.with_row(mb: 3) do
-            render(Primer::Beta::Heading.new(tag: :h1, text_align: :center)) do
-              render(Primer::Beta::Octicon.new(icon: :alert, color: :danger, size: :medium))
+  component_wrapper do
```
akabiru, line 31: This is just for semantics and convention, but since async dialogs replace stream content via JS, it's technically not necessary- which is why it worked before without it.

judithroth (approved):
Just replacing parts of the page is of course another solution to the problem :+1: 
:shipit:

PR #16497, opened 2024-08-21
[#57367] Disable edit for OneDrive/Sharepoint with AMPF enabled

akabiru:
https://community.openproject.org/work_packages/57367

OneDrive/SharePoint storages that have AMPF activated do not have this option as the user can not change any settings. and the button should not be shown in that case.


akabiru:
_more flaky tests_

Kharonus (approved):
LGTM

PR #16498, opened 2024-08-21
[#57368] Remove "No Specific folder" as on option in OneDrive/Sharepoint AMPF storage

akabiru:
https://community.openproject.org/work_packages/57368

There is no use case with a _OneDrive/SharePoint_ storage that has the **permission inheritance chain** broken to allow a storage with no specific folder. Because no one would have access to anything.

---

Screenshot 2024-08-21 at 3 48 01 PM

---

Screenshot 2024-08-21 at 3 50 18 PM

### DIFF
--- a/modules/storages/spec/models/storages/project_storage_spec.rb
+++ b/modules/storages/spec/models/storages/project_storage_spec.rb
@@ -88,7 +88,7 @@ RSpec.describe Storages::ProjectStorage do
       context "when the storage is a one drive storage" do
         let(:storage) { build_stubbed(:one_drive_storage, :as_automatically_managed) }
 
-        it "returns true for project_folder_mode inactive" do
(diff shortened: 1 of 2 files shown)


modules/storages/spec/models/storages/project_storage_spec.rb
```diff
@@ -89,7 +89,7 @@
         let(:storage) { build_stubbed(:one_drive_storage, :as_automatically_managed) }
 
         it "returns true for project_folder_mode inactive" do
```
judithroth:
```suggestion
        it "returns false for project_folder_mode inactive" do
```

judithroth (approved):
LGTM (but Dominic asked to wait for people to be back from vacation before merge, so this is just for the record at the moment)

PR #16514, opened 2024-08-23
Bug/57401 sorting by link custom field causes exception

toy:
# Ticket
[OP#57401](https://community.openproject.org/projects/openproject/work_packages/57401)

Allow ordering by custom field of link format

Adding sorting as for string format + add tests for ordering by custom fields of all formats

# Merge checklist

- [ ] Added/updated tests

### DIFF
--- a/app/models/custom_field/order_statements.rb
+++ b/app/models/custom_field/order_statements.rb
@@ -38,7 +38,7 @@ module CustomField::OrderStatements
       else
         [select_custom_option_position]
       end
-    when "string", "text", "date", "bool"
+    when "string", "text", "date", "bool", "link"
       if multi_value?
         [select_custom_values_as_group]
       else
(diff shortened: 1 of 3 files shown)


ulferts (approved):
🙇 for adding all the specs.

The functionality is also fine and works as expected.

What I have been wondering about though, is that the specs are currently work package `Query` specific. The same code is however also used in the `ProjectQuery` and there the order statement are included in a different manner. From my point of view it would help to copy over the specs to the `ProjectQuery` realm as well. That would in my expectation also make the subsequent changes to be carried out for the optimization of the performance a safer change. But this needn't be done in this PR necessarily.

toy:
True, also it looks like only for work packages there is a logic to change the placement of null values (`null_handling` method). I'll create a maintenance ticket for that.

PR #16524, opened 2024-08-26
[#56992] Wording of error message for Nextcloud host in storage form is misleading

ba1ash:
# Ticket
https://community.openproject.org/work_packages/56992

image


akabiru (approved):
:shipit:

PR #16528, opened 2024-08-26
[#57068] Nextcloud "recheck connection" checks for GroupFolderApp though AMPF is deactivated

ba1ash:
# Ticket

https://community.openproject.org/work_packages/57068


Kharonus (approved):
Yep, I think, that covers all cases. I was a bit confused about the boolean juggling in the validator line 86, but it makes sense and does the job. So -> LGTM.

PR #16530, opened 2024-08-27
Simplify work packages controller filters

as-op:
Apply suggestion of https://github.com/opf/openproject/pull/16364#discussion_r1731396582


ulferts (approved):
👍

as-op:
Thanks for the lightning fast review :D

PR #16531, opened 2024-08-27
Copy project page not support turbo navigate yet.

Eric-Guo:
### Before (page content not fully loaded)

image

image

### After (page content loaded)

image

### After

Because fix the copy project page is hard for me, so just let copy project page refresh.

# Merge checklist

- [ ] Added/updated tests
- [x] Tested major browsers (Chrome, Safari)


oliverguenther (approved):
Good catch, thanks @Eric-Guo

PR #16537, opened 2024-08-27
Expands logging to CopyFileLinksService

mereghost:
## Related WP: [OP#56357](https://community.openproject.org/projects/openproject/work_packages/56357)

Just more of the same, just adding some extra logging here and there and some cleanups

### DIFF
--- a/modules/storages/app/common/storages/peripherals/storage_interaction/one_drive/files_info_query.rb
+++ b/modules/storages/app/common/storages/peripherals/storage_interaction/one_drive/files_info_query.rb
@@ -33,6 +33,7 @@ module Storages
     module StorageInteraction
       module OneDrive
         class FilesInfoQuery
+          include TaggedLogging
           using ServiceResultRefinements
 
           def self.call(storage:, auth_strategy:, file_ids: [])
--- a/modules/storages/app/common/storages/peripherals/storage_interaction/one_drive/files_info_query.rb
+++ b/modules/storages/app/common/storages/peripherals/storage_interaction/one_drive/files_info_query.rb
@@ -52,17 +52,20 @@ module Storages
               )
             end
 
-            result = Array(file_ids).map do |file_id|
-              file_info_result = FileInfoQuery.call(storage: @storage, auth_strategy:, file_id:)
+            with_tagged_logger do
+              info "Retrieving file information for #{file_ids.join(', ')}"
+              result = Array(file_ids).map do |file_id|
+                file_info_result = FileInfoQuery.call(storage: @storage, auth_strategy:, file_id:)
 
-              file_info_result.on_failure do |failed_result|
-                return failed_result if failed_result.error_source.module_parent == AuthenticationStrategies
+                file_info_result.on_failure do |failed_result|
+                  return failed_result if failed_result.error_source.module_parent == AuthenticationStrategies
+                end
+
+                wrap_storage_file_error(file_id, file_info_result)
               end
 
-              wrap_storage_file_error(file_id, file_info_result)
+              ServiceResult.success(result:)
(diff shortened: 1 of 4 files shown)


modules/storages/app/common/storages/peripherals/storage_interaction/one_drive/files_info_query.rb
```diff
@@ -33,6 +33,7 @@ module Peripherals
     module StorageInteraction
       module OneDrive
         class FilesInfoQuery
+          include TaggedLogging
```
Kharonus, line 36: ❓ isn't there a usage of `with_tagged_logger` needed and missing in the one drive query?

modules/storages/app/services/storages/file_links/copy_file_links_service.rb
```diff
@@ -122,13 +127,14 @@ def source_files_info(source_file_links)
 
       def target_files_map
         Peripherals::Registry
-          .resolve("#{@source.storage.short_provider_type}.queries.file_path_to_id_map")
-          .call(storage: @source.storage, auth_strategy:, folder: Peripherals::ParentFolder.new(@target.project_folder_location))
+          .resolve("#{@target.storage.short_provider_type}.queries.file_path_to_id_map")
+          .call(storage: @target.storage, auth_strategy:, folder: Peripherals::ParentFolder.new(@target.project_folder_location))
```
Kharonus, line 131: 🟢 uuh, was that a big bug? :D what were possible consequences? 😂

modules/storages/spec/services/storages/file_links/copy_file_links_service_spec.rb
```diff
@@ -0,0 +1,88 @@
...
+        expect(link.origin_id).to eq(source_links[index].origin_id)
+      end
+    end
+  end
+
+  context "when AMPF is enabled" do
+    let(:files_info) { class_double(Storages::Peripherals::StorageInteraction::Nextcloud::FilesInfoQuery) }
+    let(:file_path_to_id) { class_double(Storages::Peripherals::StorageInteraction::Nextcloud::FilePathToIdMapQuery) }
+    let(:auth_strategy) do
+      Storages::Peripherals::StorageInteraction::AuthenticationStrategies::Strategy.new(key: :basic_auth)
```
Kharonus, line 44: 🟡 why not the "usual" usage as `Storages::Peripherals::StorageInteraction::AuthenticationStrategies::BasicAuth.strategy`?

Kharonus (changes requested):
I only have two small issues. If there is a good reason for it, speak up! :P

modules/storages/app/services/storages/file_links/copy_file_links_service.rb
mereghost: Not many, as we didn't expect 2 different instances, but... now it is handled.

In other words: the target project_storage is a copy of the source, so it should not  be a different storage at all but it is clearer this way.

modules/storages/spec/services/storages/file_links/copy_file_links_service_spec.rb
mereghost: I want the actual strategy for matching the expectations. Using the `.strategy` will create a new object every time and won't match later on, as it isn't a value object.

modules/storages/app/common/storages/peripherals/storage_interaction/one_drive/files_info_query.rb
mereghost: Good catch

Kharonus (approved):
LGTM

PR #16540, opened 2024-08-27
Allow "Accountable" to be a settable field via inbound mails

aaron-contreras:
# Ticket
https://community.openproject.org/work_packages/55206

I'd like to be able to set the "Accountable" field on work packages via inbound mails just like we can with setting an Assignee.

It was a matter of extending the already existing mail handler to accept an extra field and updating the test coverage to ensure it's behaving as expected.

# Merge checklist

- [x] Added/updated tests
- [ ] ~Tested major browsers (Chrome, Firefox, Edge, ...)~


oliverguenther (approved):
Nice 👍

PR #16558, opened 2024-08-29
[#57503] Time based filters like "Created on" don't trigger auto loading

dombesz:
# Ticket
https://community.openproject.org/work_packages/57503

Trigger the reload event for input fields on every input event instead of the change event.

Filter through all the reload event target elements and bind them to an input event if the field is an input field.
Otherwise bind the reload for the change event (for select, multi select, ng-select, checkboxes).

# Merge checklist

- [x] Added/updated tests


ulferts (approved):
Kudos for this change @dombesz. It is way simpler to read than before and fixes the problems.

PR #16565, opened 2024-08-30
Implementation/57515 show a list of all projects in which a given custom field is activated

akabiru:
# Ticket

Follows #16543 

https://community.openproject.org/wp/57515

List projects where a given custom field is enabled. In cases where the CF is enabled for all projects, show a notice of the same.

## Screenshots

https://github.com/user-attachments/assets/99655ce6-5393-4d74-bb52-a1ecb5f22c80

This PR largely builds on the Project Attributes functionality - as a _"Project Attribute"_ is a special type of Custom Field known in the domain as "_ProjectCustomField_". The project queries are kept separate since the join table is different in both cases

# Merge checklist

- [x] Added/updated tests


app/components/admin/custom_fields/custom_field_projects/row_component.rb
Kharonus, line 1: ❓ Why extracting row and table component logic to the `admin` namespace, if it is not used elsewhere? Why couldn't it stay at `settings` namespace?

app/models/queries/projects/filters/available_project_custom_fields_filter.rb
```diff
@@ -0,0 +1,70 @@
...
+  end
+
+  def available?
+    User.current.admin?
+  end
+
+  def apply_to(_query_scope)
+    case operator
+    when "="
+      super.with_available_custom_fields(values)
```
Kharonus: 🟡 maybe I am missing part of the logic you have written/extracted here, but why exactly is this line not using `super.with_available_project_custom_fields(values)` in a filter name `AvailableProjectCustomFieldsFilter`?

app/components/admin/custom_fields/custom_field_projects/row_component.rb
Kharonus: I think I found it, you are using it in the view template from `admin/custom_fields/custom_fields_projects/index.html.erb`.

Kharonus (commented):
Not yet approving it, but also unsure, if the comment is hinting at something that needs change. :D

app/components/admin/custom_fields/custom_field_projects/row_component.rb
akabiru: Precisely. `{Admin::CustomFields::CustomFieldProjects,Settings::ProjectCustomFields}::{RowCompoent,TableComponent}` are reserved - only the inheritance is elevated to the primary model in this context which is a `CustomField` - a "ProjectAttribute" is only a special type of custom field identified as `ProjectCustomField` via an STI model.

app/models/queries/projects/filters/available_project_custom_fields_filter.rb
akabiru: The naming is confusing - I should rename this filter to `AvailableCustomFieldProjectsFilter` to clarify that we're querying against `CustomFieldsProject` which is the join table for all custom field types, except "ProjectCustomField" aka "Project Attributes" which uses a special join table `ProjectCustomFieldsProjectMapping`

Hence, the scopes in use are different as both filters query different tables. Good eye! 👍🏾

app/components/admin/custom_fields/custom_field_projects/row_component.rb
akabiru: https://github.com/opf/openproject/pull/16565/commits/f240ef707725f375f6cd2678e54c17e0b0d8d861

Kharonus (approved):
LGTM, fine work

PR #16567, opened 2024-08-30
Bug/56771 meeting timestamp in edit form not the same as in details

ulferts:
# Ticket

https://community.openproject.org/wp/56771

The PR ensures that times and time zones displayed in the meetings module are always in the time zone of the current user. The bug originally reported in the work package could not be reproduced but before, different code paths have been used for rendering the time zone and the calculation on times was in part achieved by setting the request's time to be in the current user's time.  

* Querying for `User.current.time_zone` will now always return a value. It will never be `nil`. Either the time zone selected by the user or the time zone set as the default (`Setting.user_default_timezone`) is returned. This already used to be the case for logged in users. For those, in case no default is set, UTC is now returned. Internally, it was already handled as such as both server and database are running on UTC. The internal users (e.g. `AnonymousUser`) follow the same pattern. 
* The i18n helpers for `format_time` and `format_date_as_time` have been simplified. There used to be code in there that was working with the localtime. Since a few rails versions back, the server is always running in UTC so there is no need for this code. The only times now necessary are UTC and the user's local time. The server's local time should have no effect. Those methods get their signature updated to use named parameters.
* `formatted_time_zone_offset` is added to `I18n` which will give the currently active time offset to the user's time zone compared to UTC. This is now employed throughout the meeting module.
* Because time/date calculations are now handled at the places that matter, the before action that changed the request's time zone is no longer necessary and is thus removed.
* The user settings time zone select field has it's blank value removed. This is in line with the user always having UTC as their fallback timezone. Along the same line of thinking it would make sense to remove the blank value from the `Setting.user_default_timezone` select and set the value to UTC as a default but this is not done in the scope of this PR.

# Merge checklist

- [x] Added/updated tests

### DIFF
--- a/app/models/anonymous_user.rb
+++ b/app/models/anonymous_user.rb
@@ -49,7 +49,7 @@ class AnonymousUser < User
 
   def mail; nil end
 
-  def time_zone; nil end
+  def time_zone; ActiveSupport::TimeZone["Etc/UTC"] end
 
   def rss_key; nil end
 
--- a/app/models/anonymous_user.rb
+++ b/app/models/anonymous_user.rb
@@ -49,7 +49,7 @@ class AnonymousUser < User
 
   def mail; nil end
 
-  def time_zone; ActiveSupport::TimeZone["Etc/UTC"] end
+  def time_zone; ActiveSupport::TimeZone[Setting.user_default_timezone.presence || "Etc/UTC"] end
 
   def rss_key; nil end
 
--- a/app/models/anonymous_user.rb
+++ b/app/models/anonymous_user.rb
@@ -27,31 +27,25 @@
 #++
 
 class AnonymousUser < User
-  validate :validate_unique_anonymous_user, on: :create
-
-  # There should be only one AnonymousUser in the database
-  def validate_unique_anonymous_user
-    errors.add :base, "An anonymous user already exists." if AnonymousUser.any?
-  end
-
-  def available_custom_fields
-    []
-  end
-
-  # Overrides a few properties
-  def logged?; false end
-
-  def builtin?; true end
-
-  def admin; false end
+  include Users::FunctionUser
 
   def name(*_args); I18n.t(:label_user_anonymous) end
 
-  def mail; nil end
+  def self.first
+    anonymous_user = super
 
-  def time_zone; ActiveSupport::TimeZone[Setting.user_default_timezone.presence || "Etc/UTC"] end
+    if anonymous_user.nil?
+      (anonymous_user = new.tap do |u|
+        u.lastname = "Anonymous"
+        u.login = ""
+        u.firstname = ""
+        u.mail = ""
+        u.status = User.statuses[:active]
+      end).save
 
-  def rss_key; nil end
+      raise "Unable to create the anonymous user." if anonymous_user.new_record?
+    end
 
-  def destroy; false end
+    anonymous_user
+  end
 end
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -409,6 +409,12 @@ class User < Principal
     @time_zone ||= (pref.time_zone.blank? ? nil : ActiveSupport::TimeZone[pref.time_zone])
   end
 
+  def reload(*)
+    @time_zone = nil
+
+    super
+  end
+
   def wants_comments_in_reverse_order?
     pref.comments_in_reverse_order?
   end
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -398,7 +398,7 @@ class User < Principal
   end
 
   def log_successful_login
-    update_attribute(:last_login_on, Time.now)
+    update_attribute(:last_login_on, Time.current)
   end
 
   def pref
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -543,46 +543,12 @@ class User < Principal
 
   # Returns the anonymous user.  If the anonymous user does not exist, it is created.  There can be only
   # one anonymous user per database.
-  def self.anonymous # rubocop:disable Metrics/AbcSize
-    RequestStore[:anonymous_user] ||=
-      begin
-        anonymous_user = AnonymousUser.first
-
-        if anonymous_user.nil?
-          (anonymous_user = AnonymousUser.new.tap do |u|
-            u.lastname = "Anonymous"
-            u.login = ""
-            u.firstname = ""
-            u.mail = ""
-            u.status = User.statuses[:active]
-          end).save
-
-          raise "Unable to create the anonymous user." if anonymous_user.new_record?
-        end
-        anonymous_user
-      end
+  def self.anonymous
+    RequestStore[:anonymous_user] ||= AnonymousUser.first
   end
 
   def self.system
-    system_user = SystemUser.first
-
-    if system_user.nil?
-      system_user = SystemUser.new(
-        firstname: "",
-        lastname: "System",
-        login: "",
-        mail: "",
-        admin: true,
-        status: User.statuses[:active],
-        first_login: false
-      )
-
-      system_user.save(validate: false)
-
-      raise "Unable to create the automatic migration user." unless system_user.persisted?
-    end
-
-    system_user
+    SystemUser.first
   end
 
   protected
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -406,7 +406,7 @@ class User < Principal
   end
 
   def time_zone
-    @time_zone ||= (pref.time_zone.blank? ? nil : ActiveSupport::TimeZone[pref.time_zone])
+    @time_zone ||= ActiveSupport::TimeZone[pref.time_zone] || ActiveSupport::TimeZone["Etc/UTC"]
   end
 
   def reload(*)
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -688,6 +688,6 @@ class User < Principal
   end
 
   def self.default_admin_account_changed?
-    !User.active.find_by_login("admin").try(:current_password).try(:matches_plaintext?, "admin") # rubocop:disable Rails/DynamicFindBy
+    !User.active.find_by_login("admin").try(:current_password).try(:matches_plaintext?, "admin")
   end
 end
--- a/app/views/users/_preferences.html.erb
+++ b/app/views/users/_preferences.html.erb
@@ -30,8 +30,10 @@ See COPYRIGHT and LICENSE files for more details.
   <%= render Settings::TimeZoneSettingComponent.new(
     "time_zone",
     form: pref_fields,
+    include_blank: false,
     container_class: (defined? input_size) ? "-#{input_size}" : "-wide"
   )
+
   %>
   <div class="form--field">
     <%= pref_fields.select :theme, theme_options_for_select, container_class: '-middle' %>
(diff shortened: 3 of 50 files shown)


ulferts:
Rubocop is complaining about the complexity of a method touched within the PR. The PR only changes the call signature of a method called from within the complained about method. Therefore, I deem it ok to not improve that method.

app/models/user.rb
```diff
@@ -398,15 +398,21 @@ def log_failed_login
...
+    update_attribute(:last_login_on, Time.current)
   end
 
   def pref
     preference || build_preference
   end
 
   def time_zone
-    @time_zone ||= (pref.time_zone.blank? ? nil : ActiveSupport::TimeZone[pref.time_zone])
+    @time_zone ||= ActiveSupport::TimeZone[pref.time_zone] || ActiveSupport::TimeZone["Etc/UTC"]
```
cbliard, line 409: This could be contracted to `ActiveSupport::TimeZone[pref.time_zone || "Etc/UTC"]` like other methods

And as `pref.time_zone` always return something now and falls back to "Etc/UTC", I think the second part `|| ActiveSupport::TimeZone["Etc/UTC"]` is dead code.

So it could simply be `@time_zone ||= ActiveSupport::TimeZone[pref.time_zone]`.

```suggestion
    @time_zone ||= ActiveSupport::TimeZone[pref.time_zone]
```

app/models/anonymous_user.rb
```diff
@@ -49,7 +49,7 @@ def name(*_args); I18n.t(:label_user_anonymous) end
 
   def mail; nil end
 
-  def time_zone; nil end
+  def time_zone; ActiveSupport::TimeZone[Setting.user_default_timezone.presence || "Etc/UTC"] end
```
cbliard: Why not rely on the default implementation of `User#time_zone`? Implementation is 
```ruby
class User
  def time_zone
    @time_zone ||= ActiveSupport::TimeZone[pref.time_zone] || ActiveSupport::TimeZone["Etc/UTC"]
  end
end
```

As a` pref` record does not exist for `AnonymousUser`, it would be automatically built without touching the db and its `time_zone` will return the right value, so it should work as expected.

Same applies for `DeletedUser` and `SystemUser`.

app/views/users/_preferences.html.erb
```diff
@@ -30,8 +30,10 @@ See COPYRIGHT and LICENSE files for more details.
   <%= render Settings::TimeZoneSettingComponent.new(
     "time_zone",
     form: pref_fields,
+    include_blank: false,
     container_class: (defined? input_size) ? "-#{input_size}" : "-wide"
   )
+
```
cbliard: Unneeded new line

lib_static/redmine/i18n.rb
```diff
@@ -138,17 +138,14 @@ def format_time(time, include_date = true)
 
       time = time.to_time if time.is_a?(String)
       zone = User.current.time_zone
-      local = if zone
-                time.in_time_zone(zone)
-              else
-                (time.utc? ? time.to_time.localtime : time)
-              end
+      local = time.in_time_zone(zone)
```
cbliard, line 150: Nice 👏

modules/meeting/app/controllers/meetings_controller.rb
```diff
@@ -97,8 +96,8 @@ def create # rubocop:disable Metrics/AbcSize
 
     if call.success?
       text = I18n.t(:notice_successful_create)
-      if User.current.time_zone.nil?
-        link = I18n.t(:notice_timezone_missing, zone: Time.zone)
+      unless User.current.pref.time_zone?
+        link = I18n.t(:notice_timezone_missing, zone: formatted_time_zone_offset)
         text += " #{view_context.link_to(link, { controller: '/my', action: :settings, anchor: 'pref_time_zone' },
                                          class: 'link_to_profile')}"
```
cbliard, line 102: I tried it. When there is no time zone set for the user, the message is displayed inviting the user to set a time zone, but the notification disappears after 5 seconds (or so). **This is extremely short to read and take action**. I suggest the success toast should not autoclose in this case. Probably worth a new bug work package for it.

modules/meeting/app/controllers/meetings_controller.rb
```diff
@@ -289,17 +288,6 @@ def load_meetings(query)
...
-  def set_time_zone(&)
-    zone = User.current.time_zone
-    if zone.nil?
-      localzone = Time.current.utc_offset
-      localzone -= 3600 if Time.current.dst?
-      zone = ::ActiveSupport::TimeZone[localzone]
-    end
-
-    Time.use_zone(zone, &)
-  end
```
cbliard, line 301: Goodbye buggy code (a dst is not always 1h).

lib_static/redmine/i18n.rb
```diff
@@ -133,18 +129,25 @@ def format_time_as_date(time, format = nil)
       end
     end
 
-    def format_time(time, include_date = true)
+    def format_time(time, include_date: true, format: Setting.time_format)
```
cbliard, line 146: It's hard to know that this method uses user time zone without reading its code. We could rename it, but I propose adding some docs for it.

```suggestion
    # Formats the given time as a time string according to the user's time zone
    # and optional specified format.
    #
    # @param time [Time, String] The time to format. Can be a Time object or a
    #   String.
    # @param include_date [Boolean] Whether to include the date in the formatted
    #   output. Defaults to true.
    # @param format [String] The strftime format to use for the time. Defaults
    #   to the format in `Setting.time_format`.
    # @return [String, nil] The formatted time string, or nil if the time is not
    #   provided.
    def format_time(time, include_date: true, format: Setting.time_format)
```

lib_static/redmine/i18n.rb
```diff
@@ -116,15 +116,11 @@ def link_regex
     # Format the time to a date in the user time zone if one is set.
     # If none is set and the time is in utc time zone (meaning it came from active record), format the date in the system timezone
     # otherwise just use the date in the time zone attached to the time.
-    def format_time_as_date(time, format = nil)
+    def format_time_as_date(time, format: nil)
```
cbliard, line 123: Documentation is not up-to-date anymore

Here is a proposal:

```ruby
    # Formats the given time as a date string according to the user's time zone and
    # optional specified format.
    #
    # @param time [Time, String] The time to format. Can be a Time object or a String.
    # @param format [String, nil] The strftime format to use for the date. If nil, the default
    #   date format from `Setting.date_format` is used.
    # @return [String, nil] The formatted date string, or nil if the time is not provided.
    def format_time_as_date(time, format: nil)
```

spec/lib/redmine/i18n_spec.rb
```diff
@@ -35,43 +35,35 @@ module OpenProject
     let(:format) { "%d/%m/%Y" }
     let(:user) { build_stubbed(:user) }
 
-    after do
-      Time.zone = nil
-    end
+    describe "#format_time_as_date" do
```
cbliard, line 38: There are no tests taking time parameter as string. Not sure if it's used that way in our code and how it deals with time strings without timezone information.
Not sure it's worth it though.

cbliard: same for `#format_time`.

app/models/user.rb
cbliard: Ah, I got it now, it's in case the user time zone is invalid. It would return `nil`, so it needs to adapt to it. That was the last commit purpose.

Discard my previous comment.

cbliard (approved):
Really nice refactoring 👏  It simplifies the code dealing with time zones and removes a lot of useless code which is always a good sign.

I only have some minor improvement suggestions.
Feel free to merge whenever you see fit.

For that long method rubocop is complaining about, if it's ok to keep it, then it could be `# rubocop:disable Metrics/AbcSize`.

app/models/anonymous_user.rb
ulferts: After reading your comment I thought, "right he is". Then I tried it out:

```
pry(main)> User.system.build_preference
  SystemUser Load (0.5ms)  SELECT "users".* FROM "users" WHERE "users"."type" = $1 ORDER BY "users"."id" ASC LIMIT $2  [["type", "SystemUser"], ["LIMIT", 1]]
  UserPreference Load (0.2ms)  SELECT "user_preferences".* FROM "user_preferences" WHERE "user_preferences"."user_id" = $1 LIMIT $2  [["user_id", 1], ["LIMIT", 1]]
=> #<UserPreference:0x000000030eaf0148 id: nil, user_id: 1, settings: {}, created_at: Mon, 08 Jul 2024 07:17:52.591795000 UTC +00:00, updated_at: Mon, 08 Jul 2024 07:17:52.591795000 UTC +00:00>
```

That happens because the user objects are persisted. It wouldn't happen on unpersisted models. 

I think this would not amount to much additional time per request when done but on the other hand the code necessary is quite easy to maintain so I would keep it as it is. 

An improvement I see, though is to extract code shared between the three internal user objects into a module.

What do you think @cbliard?

app/views/users/_preferences.html.erb
ulferts: Removed in [b2e6d67](https://github.com/opf/openproject/pull/16567/commits/b2e6d67d19a5355cc2caed4fef5621330b3bbc38)

lib_static/redmine/i18n.rb
ulferts: Applied in [db1e9f3](https://github.com/opf/openproject/pull/16567/commits/db1e9f3e4753c041a1b239cecc0c2f8ca828da3b), thanks.

spec/lib/redmine/i18n_spec.rb
ulferts: I removed the option to pass in strings in [ff45d0d](https://github.com/opf/openproject/pull/16567/commits/ff45d0d0a229e27425abc5eec98278839d85b95f). The tests at least are green and from the calls I looked at randomly, no calls are made using strings. But there are a lot of those. I still think it is ok to make that change with the automatic tests not having spotted anything and with the QA still there to catch things.

What is your opinion, @cbliard

app/models/anonymous_user.rb
cbliard: Nice find.

> I think this would not amount to much additional time per request when done but on the other hand the code necessary is quite easy to maintain so I would keep it as it is.

Yeah, I was not really into the time it would take, but rather the logic duplication in the 3 classes, and in the fallback to Etc/UTC with a regular user.

> An improvement I see, though is to extract code shared between the three internal user objects into a module.

Yep 👍

spec/lib/redmine/i18n_spec.rb
cbliard: If passing strings is never used, then yes definitely worth removing it.

And I don't think adding tests was worth it, but I was wondering what the time zone would be when parsing a string to a time.

modules/meeting/app/controllers/meetings_controller.rb
ulferts: I created a [ticket for this](https://community.openproject.org/wp/58002). There are quite a number of places where we present times and not always do we tell the user that it might not be in their local time (e.g. work package activities). So I think there should be some nudging whenever the time zone is not set.

app/models/anonymous_user.rb
ulferts: Did that in [cd615f5](https://github.com/opf/openproject/pull/16567/commits/cd615f5108034852b1473eb775e55574951163bb).

ulferts:
I'll merge this one in after the release branch for 14.6 has been created.

EinLama:
The release branch for 14.6 has been created 👍🏻

PR #16568, opened 2024-08-30
[57258] Fix issues found during review

cbliard:
# Ticket
https://community.openproject.org/wp/57258

Two issues have been found during review:

- [x] Derivation hints should be displayed in status-based mode
- [x] Hints when remaining work is increased/decreased like work should be reworked

This PR addresses them.

## Screenshots

**Derivation hints should be displayed in status-based mode**

Before:

image

After:

image

**Hints when remaining work is increased/decreased like work should be reworked**

The amount by which it is changed is indicated in the hint text.

Before:

Aug-29-2024_11-33-42

After: 

new behavior with amount

**Derivation hints should be displayed in status-based mode**

Call the right method to set the caption from the calculated hint.
Then some hints were displayed inappropriately so they had to be fixed.

**Hints when remaining work is increased/decreased like work should be reworked**

The delta is added to the derived hint in the work package. A `WorkPackage::ProgressHint` class is introduced to manage the logic of converting the keys and the delta to a translated message with the duration converted into hours or hours and days depending on the configuration.

# Merge checklist

- [x] Added/updated tests

### DIFF
--- a/app/forms/work_packages/progress_form.rb
+++ b/app/forms/work_packages/progress_form.rb
@@ -151,6 +151,7 @@ class WorkPackages::ProgressForm < ApplicationForm
       name:,
       value: field_value(name),
       label:,
+      caption: field_hint_message(name),
       readonly: true,
       classes: "input--readonly",
       placeholder: ("-" if placeholder)
@@ -199,9 +200,10 @@ class WorkPackages::ProgressForm < ApplicationForm
   end
 
   def default_field_options(name)
+    action = name == :status_id ? "change" : "input"
     data = { "work-packages--progress--preview-progress-target": "progressInput",
              "work-packages--progress--touched-field-marker-target": "progressInput",
-             action: "input->work-packages--progress--touched-field-marker#markFieldAsTouched" }
+             action: "#{action}->work-packages--progress--touched-field-marker#markFieldAsTouched" }
 
     if @focused_field == name
       data[:"work-packages--progress--focus-field-target"] = "fieldToFocus"
--- a/app/forms/work_packages/progress_form.rb
+++ b/app/forms/work_packages/progress_form.rb
@@ -197,10 +197,9 @@ class WorkPackages::ProgressForm < ApplicationForm
   end
 
   def default_field_options(name)
-    action = name == :status_id ? "change" : "input"
     data = { "work-packages--progress--preview-progress-target": "progressInput",
              "work-packages--progress--touched-field-marker-target": "progressInput",
-             action: "#{action}->work-packages--progress--touched-field-marker#markFieldAsTouched" }
+             action: "work-packages--progress--touched-field-marker#markFieldAsTouched" }
 
     if @focused_field == name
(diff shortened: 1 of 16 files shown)


app/forms/work_packages/progress_form.rb
```diff
@@ -199,9 +197,10 @@ def as_percent(value)
   end
 
   def default_field_options(name)
+    action = name == :status_id ? "change" : "input"
     data = { "work-packages--progress--preview-progress-target": "progressInput",
              "work-packages--progress--touched-field-marker-target": "progressInput",
-             action: "input->work-packages--progress--touched-field-marker#markFieldAsTouched" }
+             action: "#{action}->work-packages--progress--touched-field-marker#markFieldAsTouched" }
```
aaron-contreras: Now that I think of it, we could even omit the action and it would use the default event for each input type. Meaning for select fields it would use "change", for input fields would use "input" and we don't need to specify this. But we could also argue that being explicit about the event is probably best.
```suggestion
             action: "work-packages--progress--touched-field-marker#markFieldAsTouched" }
```

frontend/src/stimulus/controllers/dynamic/work-packages/progress/preview-progress.controller.ts
```diff
@@ -61,7 +61,11 @@ export default class PreviewProgressController extends Controller {
     };
 
     this.progressInputTargets.forEach((target) => {
-      target.addEventListener('input', this.debouncedPreview);
+      if (target.tagName.toLowerCase() === 'select') {
+        target.addEventListener('change', this.debouncedPreview);
+      } else {
+        target.addEventListener('input', this.debouncedPreview);
+      }
```
aaron-contreras, line 68: Same comment as above - whatever you see more fit

spec/support/edit_fields/progress_edit_field.rb
```diff
@@ -87,18 +87,31 @@ def clear
...
+
+  def status_field?
+    field_name == "work_package_status_id"
+  end
+
   def focus
     return if focused?
 
     input_element.click
+    input_element.click if status_field? # to close the dropdown
```
aaron-contreras, line 114: Clever

aaron-contreras (approved):
All looks good! Left a question open to your decision. Merge if disregarded :)

app/forms/work_packages/progress_form.rb
cbliard: Good catch! I tried it and it works. I'll commit it.

frontend/src/stimulus/controllers/dynamic/work-packages/progress/preview-progress.controller.ts
cbliard: I don't get it. The first argument is mandatory. How can I make both identical?

app/forms/work_packages/progress_form.rb
cbliard: commited

PR #16570, opened 2024-08-30
Bug/57508 custom fields with format string text bool link and date dont forbid multi select internally and have handling in ordering

toy:
# Ticket
[OP#57508](https://community.openproject.org/work_packages/57508)

What are you trying to accomplish?
Cleanup custom field ordering from unsupported cases, and check that multi_value (and allow_non_open_versions) are not assignable on model level and not only on UI level.

# Merge checklist

- [ ] Added/updated tests

### DIFF
--- a/app/models/custom_field.rb
+++ b/app/models/custom_field.rb
@@ -268,6 +268,10 @@ class CustomField < ApplicationRecord
     field_format == "list"
   end
 
+  def user?
+    field_format == "user"
+  end
+
   def version?
     field_format == "version"
   end
@@ -281,7 +285,7 @@ class CustomField < ApplicationRecord
   end
 
   def multi_value_possible?
-    %w[version user list].include?(field_format)
+    version? || user? || list?
   end
 
   def allow_non_open_versions_possible?
--- a/app/models/custom_field.rb
+++ b/app/models/custom_field.rb
@@ -281,13 +281,11 @@ class CustomField < ApplicationRecord
   end
 
   def multi_value_possible?
-    %w[version user list].include?(field_format) &&
-      [ProjectCustomField, WorkPackageCustomField, TimeEntryCustomField, VersionCustomField].include?(self.class)
+    %w[version user list].include?(field_format)
   end
 
   def allow_non_open_versions_possible?
-    version? &&
-      [ProjectCustomField, WorkPackageCustomField, TimeEntryCustomField, VersionCustomField].include?(self.class)
+    version?
   end
 
   ##
--- a/app/models/custom_field.rb
+++ b/app/models/custom_field.rb
@@ -277,17 +277,9 @@ class CustomField < ApplicationRecord
     field_format == "bool"
   end
 
-  def multi_value?
-    multi_value
-  end
-
   def multi_value_possible?
     %w[version user list].include?(field_format) &&
       [ProjectCustomField, WorkPackageCustomField, TimeEntryCustomField, VersionCustomField].include?(self.class)
-  end
-
-  def allow_non_open_versions?
-    allow_non_open_versions
   end
 
   def allow_non_open_versions_possible?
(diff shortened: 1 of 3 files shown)


toy:
@ulferts I have removed checking for custom field class from `multi_value_possible?` and `allow_non_open_version_possible?` checks, do you remember why where they needed (added in [OP#34073](https://community.openproject.org/projects/stadt-koeln/work_packages/34073))?

app/views/custom_fields/_form.html.erb
```diff
@@ -118,7 +118,7 @@ See COPYRIGHT and LICENSE files for more details.
     </fieldset>
   <% end %>
 
-  <% if @custom_field.new_record? || @custom_field.version? || @custom_field.allow_non_open_versions_possible? %>
```
ulferts, line 121: Thanks for the cleanup in here. It was confusing before.

app/models/custom_field.rb
```diff
@@ -277,22 +280,12 @@ def boolean?
...
-  end
-
   def multi_value_possible?
-    %w[version user list].include?(field_format) &&
-      [ProjectCustomField, WorkPackageCustomField, TimeEntryCustomField, VersionCustomField].include?(self.class)
-  end
-
-  def allow_non_open_versions?
-    allow_non_open_versions
+    %w[version user list].include?(field_format)
```
ulferts: You could also write it like this:

```
version? || user? || list?
```

which would be more in line with the implemenation of `allow_non_open_versions_possible?`. But that is just an idea and not a necessity.

ulferts:
> @ulferts I have removed checking for custom field class from `multi_value_possible?` and `allow_non_open_version_possible?` checks, do you remember why where they needed (added in [OP#34073](https://community.openproject.org/projects/stadt-koeln/work_packages/34073))?

From the looks of the [commit that introduced it](https://github.com/opf/openproject/commit/9edc7244a94#diff-694e91add82ff0c4a4d4af8ab641148e51cf7db9c077c6156e58c2253d091a44R265-R268) it looks like it was introduced to prevent the select being displayed for custom fields on TimeEntries. That was later on changed to also allow multi select for TimeEntry- and VersionCustomField. So I guess it is ok to remove the limitation as `OpenProject::CustomFieldFormat` covers which customized model should get the fields and currently all of those in there should have that option.

ulferts (approved):
Nice change here @toy.

I have one idea but feel free to ignore it. Good to merge in any case.

app/models/custom_field.rb
toy: I also noticed it, but too quickly ignored probably thinking about efficiency, but `version? || user? || list?` is also more explicit

toy: Made me wonder, also not important in this case, but it is not even slower:
```ruby
require 'benchmark/ips'

class CustomField
  attr_reader :field_format

  def initialize(field_format)
    @field_format = field_format
  end

  def list?
    field_format == "list"
  end

  def version?
    field_format == "version"
  end

  def user?
    field_format == "user"
  end

  def multi_value_possible1?
    %w[version user list].include?(field_format)
  end

  def multi_value_possible2?
    version? || user? || list?
  end
end

cfs = [
  CustomField.new('version'),
  CustomField.new('user'),
  CustomField.new('list'),
  CustomField.new('foobar'),
]

Benchmark.ips do |x|
  x.report('include?') { cfs.each(&:multi_value_possible1?) }

  x.report('x? || y? || z?') { cfs.each(&:multi_value_possible2?) }

  x.compare!
end
```
gives
```
Warming up --------------------------------------
            include?   173.006k i/100ms
      x? || y? || z?   185.909k i/100ms
Calculating -------------------------------------
            include?      1.731M (± 0.6%) i/s  (577.81 ns/i) -      8.823M in   5.098375s
      x? || y? || z?      1.851M (± 0.4%) i/s  (540.34 ns/i) -      9.295M in   5.022755s

Comparison:
      x? || y? || z?:  1850702.9 i/s
            include?:  1730668.9 i/s - 1.07x  slower
```

PR #16587, opened 2024-09-02
[#57016] updated status API specification

Kharonus:
# Ticket
[#57016](https://community.openproject.org/work_packages/57016)

- more complete status API specification

### DIFF
--- a/docs/api/apiv3/components/schemas/status_model.yml
+++ b/docs/api/apiv3/components/schemas/status_model.yml
@@ -9,52 +9,45 @@ properties:
   id:
     type: integer
     description: Status id
-    readOnly: true
     minimum: 1
   name:
     type: string
     description: Status name
-    readOnly: true
-  position:
-    type: integer
-    description: Sort index of the status
-    readOnly: true
-  isDefault:
-    type: boolean
-    readOnly: true
   isClosed:
     type: boolean
-    description: are tickets of this status considered closed?
-    readOnly: true
+    description: Indicates, whether work package of this status are considered closed
+  color:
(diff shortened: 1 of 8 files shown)


docs/api/apiv3/components/schemas/status_model.yml
```diff
@@ -9,52 +9,45 @@ properties:
...
   isReadonly:
     type: boolean
-    description: are tickets of this status read only?
-    readOnly: true
+    description: Indicates, whether work package of this status are readonly
+  excludedFromTotals:
+    type: boolean
+    description: |-
+      Indicates, whether work package of this status are excluded from totals of `Work`, `Remaining work`, and
+      `% Complete` in a hierarchy.
```
cbliard, line 33: Kudos for also adding that one 👍

cbliard (changes requested):
Looks good to me 👍 

The review was a bit tedious to do. I would have appreciated to have separate commits for purely cosmetic changes like changing attributes order to match the order returned by the API (good thnig BTW), changing single quotes to double quotes, moving stuff around without modifying, etc, and for the intended change like adding `color` and `excludeFromTotals`. That would have been easier to review.

Also in some places, you used "status" for the plural form whereas "statuses" is used everywhere else in the app. It should be updated to statuses for consistency.

It would be great if you could also update `docs/api/apiv3/tags/statuses.yml` to match your changes (visible on http://localhost:3000/api/docs/#?route=tag--Statuses--local-properties).

cbliard (approved):
Great 👏

PR #16590, opened 2024-09-02
Add spacing between title and icons for auth providers

aaron-contreras:
Missing styling for the registration form due to ID selector specificity.

# Ticket
https://community.openproject.org/projects/openproject/work_packages/57366

Adding some vertical spacing between the heading for the auth provider section and the actual auth provider icons.

## Screenshots

**Before**
image

**After**
image

Based on the fact that the heading element is only present under a given condition, this is when the spacing is needed. Based on this, using a sibling CSS selector `~` to determine when to conditionally add this top margin to the `provider-list`

# Merge checklist

- [ ] Added/updated tests


frontend/src/global_styles/content/_accounts.sass
```diff
@@ -119,3 +119,6 @@
 
   &--agreement
     font-size: inherit
+
+.login-auth-providers-title ~ .login-auth-provider-list
```
bsatarnejad: Hi @aaron-contreras It correctly fix the bug, but I have an alternative way that help us with reducing custom css.

You can add primer class 'mb-3' to the element with 'login-auth-providers-title' class:

```
      <h3 class="login-auth-providers-title mb-3"><span>
          <%= local_assigns[:omniauth_title] || I18n.t('account.login_with_auth_provider')%>
        </span></h3>
```
It adds 1rem margin bottom to this element.
https://primer.style/foundations/css-utilities/margin

bsatarnejad (changes requested):

frontend/src/global_styles/content/_accounts.sass
aaron-contreras: Ah, great! I forget this CSS is available to use outside Primer view components. Thank you :)

bsatarnejad (approved):
LGTM 👍

PR #16591, opened 2024-09-02
Rework the Nextcloud AMPF Sync test suite

mereghost:
~~Soon...~~

The current tests for the Nextcloud AMPF Sync stub the requests and their XML responses. This makes making changes to the spec, or adapting results a pain.

This PR aims to fix this by moving the stubs to the commands and queries that are called.


apfohl (approved):
Look okay. Even though I find the big sync spec really hard to follow. The tons of setup code are hard to match to the actual use cases. But the use case tests are fine.

mereghost:
> Look okay. Even though I find the big sync spec really hard to follow. The tons of setup code are hard to match to the actual use cases. But the use case tests are fine.

Yeah. I'd love to find a way to shrink the setup but so far I've failed in this endeavor.

Kharonus (approved):
❤️ it.

The improvement on comprehension and readability of this spec is awesome.

Kharonus:
Regarding big setups:

More important then test setup is always the test cases and expectations. I often times skip reading, how `rename_file` is set up, when I read

```ruby
          expect(rename_file).to have_received(:call)
                                   .with(storage:, auth_strategy:, file_id: renamed_storage.project_folder_id,
                                         name: renamed_storage.managed_project_folder_name).once
```

it just don't matter for understanding the test case itself. Which is why the current improvement already is a big step.

However, the complex setup is a perfect indicator, that this service is getting close to a god class. ;)

PR #16593, opened 2024-09-02
Bug/57612 work package report export fails if work package has unavailable attachments

as-op:
# Ticket

If there are faulty attachments (attachment exists in db and work_package.attachments, but work_package.attachments[0].file.local_file throws an error. So the export is stopped.

https://community.openproject.org/work_packages/57612

=> Treat it like "attachment not found", catch and log the error but don't stop the export.

# Merge checklist

- [x] Added/updated tests


as-op:
> [!NOTE] 
> Failing spec is unrelated.

dombesz (approved):

PR #16605, opened 2024-09-03
[#57615] Report type work package export doesn't display the correct query title

as-op:
# Ticket

Bug: title of PDF report/PDF table is not correct

https://community.openproject.org/work_packages/57615

* add query title parameter (which may be just a text from the frontend and not saved into the query) to the export dialog

* add the missing parameter

# Merge checklist

- [X] Added/updated tests


aaron-contreras (approved):
Nice job on the specs!

as-op:
Thanks @aaron-contreras ❤️

PR #16619, opened 2024-09-04
[#57536] Remove menu entry in Administration submenu

EinLama:
The user can access the "Projects" Submenu" entry instead. This entry here has been redundant for a while and should be removed to avoid confusion for the user.

# Ticket
https://community.openproject.org/wp/57536


ulferts (approved):
👍 

Hooray on your first merged PR, @EinLama 🎉

PR #16627, opened 2024-09-04
Implementation/57579 define custom fields add projects dialog for multi project activation

akabiru:
# Ticket

https://community.openproject.org/work_packages/57579

Follows: #16565 

Enable admins to activate custom fields on multiple projects at once

## Screenshots

https://github.com/user-attachments/assets/3259e3a0-9d31-4791-bfdc-4e453a95c32e

Continue building upon project attributes foundation- elevate the abstraction to the primary model in this context which is a `CustomField` - a "ProjectAttribute" is only a special type of custom field identified as `ProjectCustomField` via an STI model.

# Merge checklist

- [x] Added/updated tests


app/contracts/custom_fields/custom_field_projects/base_contract.rb
```diff
@@ -0,0 +1,52 @@
...
+  module CustomFieldProjects
+    class BaseContract < ::ModelContract
+      attribute :project_id
+      attribute :custom_field_id
+
+      validate :select_custom_fields_permission
+      validate :not_for_all
+
+      def select_custom_fields_permission
+        return if user.allowed_in_project?(:select_custom_fields, model.project)
```
akabiru, line 39: @ulferts the PR adopts the `select_custom_fields` permission that should be reusable by "non-admin" users- but at present the multiple-project activation is only visible to admin users.

*_Update: We had a chat and agreed to retain as is_

akabiru:
_🚧 The feature test failure is unrelated, but I can look into it_

app/services/custom_fields/custom_field_projects/bulk_create_service.rb
```diff
@@ -0,0 +1,125 @@
...
+        @custom_field = custom_field
+        @include_sub_projects = include_sub_projects
+      end
+
+      def perform
+        service_call = validate_permissions
+        service_call = validate_contract(service_call, incoming_mapping_ids) if service_call.success?
+        service_call = perform_bulk_create(service_call) if service_call.success?
+
+        service_call
```
apfohl, line 47: 🟡 You could use the bind chain or even better the `>>` operator from the `ServiceResultRefinements`.

app/services/custom_fields/custom_field_projects/bulk_create_service.rb
```diff
@@ -0,0 +1,125 @@
...
+        set_attributes_results = project_ids.map do |id|
+          set_attributes(project_id: id, custom_field_id: @custom_field.id)
+        end
+
+        if (failures = set_attributes_results.select(&:failure?)).any?
+          service_call.success = false
+          service_call.errors = failures.map(&:errors)
+        else
+          service_call.result = set_attributes_results.map(&:result)
+        end
```
apfohl, line 72: 🟡 Sorry if I'm a bit direct here, but I would treat a `ServiceResult` as a finite object and would rather create a new one and see every `ServiceResult` as immutable. That would also play nicely when using `bind` or `>>`. You would not need to have the `service_call` parameter at all for this function, because this function would only be called after the one before succeeds.

app/services/custom_fields/custom_field_projects/bulk_create_service.rb
```diff
@@ -0,0 +1,125 @@
...
+
+        service_call
+      end
+
+      def perform_bulk_create(service_call)
```
apfohl, line 77: 🟡 Maybe just pass in the actual result and call the method with `Bind`, as described above.

apfohl (approved):
🟡 You could leave it like that, but I would encourage you to give `Bind`/`>>` a try in the `BulkService`. I can help you if you wish.

akabiru:
> 🟡 You could leave it like that, but I would encourage you to give `Bind`/`>>` a try in the `BulkService`. I can help you if you wish.

_Update:_ We synced and will effect updates in a separate PR 👍🏾

PR #16630, opened 2024-09-04
Ignore FileLink renames when rendering activities

mereghost:
#### Related WP: [OP#53995](https://community.openproject.org/projects/openproject/work_packages/53995)

# What?

Sometimes, any change to a work package would add entries about the addition of file links to the current user.

After trying to pinpoint it for some time, we could finally nail it to file renames on the remote.

# Why?

Some basics first: our `StorableJournal` stores the **storage name** and **file link name**, the reason being that we want to still render the information even if the file link or storage has been removed.

So, the journal rendering was pretty naive: if the current entry is file name nil, this should be a removal, otherwise an file link was added.

On a rename we would land on the *file link added* case. Also, previous mentions would use the new name of the file, due to the way we render the links, so it would basically look like the same file was re-added multiple times.

# And the fix?

We still need to keep track of the most current file name to render the **file link removed** message, so instead of avoid the creation of the Journal entry, we skip the render if both old and new file link names are present.

So we went from this:
image

to this:

Screenshot from 2024-09-04 17-32-35

### DIFF
--- a/lib/open_project/journal_formatter/file_link.rb
+++ b/lib/open_project/journal_formatter/file_link.rb
@@ -35,6 +35,8 @@ class OpenProject::JournalFormatter::FileLink < JournalFormatter::Base
     id = key.to_s.sub("file_links_", "").to_i
     label, old_value, value, storage = format_details(id, values)
 
+    return if old_value && value
+
     if options[:html]
       label, old_value, value = format_html_details(label, old_value, value)
       value = format_html_file_link_detail(id, value)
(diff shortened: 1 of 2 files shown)


mereghost:
Note that once a file link is removed after being renamed, the activity entry of when it was added will refer to the original file link name.

Kharonus (approved):
I see the fix and I agree with it.

Still it leaves some taste behind. So, if within the life of a file link it gets renamed, the journal would look like:

```
File link to my_first_name.txt added
# and later
File link to my_second_name.txt removed
```
this leaves a few questions open. But I can't figure a good way of handling this right now, especially, as with nextcloud every user can have its own file name.

PR #16640, opened 2024-09-05
[#56802] Document the new component and use it to replace streamable_flash_banner

mrmir:
# Ticket
https://community.openproject.org/work_packages/56802

## Screenshots

# Merge checklist

- [ ] Added/updated tests


modules/meeting/app/components/meetings/update_flash_component.html.erb
```diff
@@ -1,21 +1,40 @@
...
-          }
-        )) do |component|
-          component.with_action_button(
-              tag: :a,
-              href: meeting_path(@meeting),
-              data: { turbo: false },
-              size: :medium) { I18n.t("label_meeting_reload") }
-          I18n.t("notice_meeting_updated")
+    if @button
+      flex_layout do |flex|
```
mrmir: @oliverguenther do you know if this is the correct way to style this? I think we may have missed something before because when I trigger a full width flash, it stays the same width but in the full style and doesn't expand like I was assuming it would

Not sure if using flex with the `op-toast--wrapper` is working as intended tbh

mrmir: @HDinger same question :)

lookbook/docs/patterns/08-flash-banner.md.erb
mrmir, line 1: @oliverguenther Does this work? Do I need to add more? Less? And what about the usage via the stimulus controller? Does that need to be documented here as well?

app/views/layouts/base.html.erb
```diff
@@ -117,10 +117,13 @@ See COPYRIGHT and LICENSE files for more details.
...
     <div class="content-overlay"></div>
     <main id="content-wrapper" class="<%= initial_classes %>">
-      <%= render_primer_banner_message %>
-      <%= render_streameable_primer_banner_message %>
+      <%# Banner messages are being rendered using primer %>
+      <div id="banner-messages" class="op-primer-banner">
+        <%= render_primer_banner_message %>
+      </div>
       <%# Banner messages are being rendered using primer %>
       <div id="banner-messages" class="op-primer-banner"></div>
```
mrmir: Is this not an extra #banner-messages div?

mrmir (commented):
Hi @oliverguenther, 

I pushed some small docs changes that I noticed were broken. Also noticed an extra div in the base layout, potentially. 

Behaviour wise it looks okay to me, but there is this issue - When you edit a meeting in a new tab that is already closed, you get the 'This meeting is not editable anymore.' flash. But this then gets replaced by the 'This page has been updated...' flash. Should that be happening? I thought we were going to do stackable flashes or whatever to allow more than one at a time. 

It looks jumpy and broken sometimes imo because of the timed poll for the update flash, where in some instances the first flash is only there for a fraction of a second before being replaced. 

I'm not sure what the solution is here, but maybe we could remove that first flash altogether and replace it with an instance of the second? As they are kind of doing similar things now.

mrmir:
Looks good to me now :)

PR #16645, opened 2024-09-06
Feature/57347 add validation connection check unexpected content for nextcloud storages

apfohl:
# Ticket
https://community.openproject.org/projects/openproject/work_packages/57347/github

Adding a validator for the `NextcloudConnectionValidation` that checks for unforseen folder contents inside the group folder on Nextcloud side.

I just programmed it!

# Merge checklist

- [ ] Added/updated tests

### DIFF
--- a/modules/storages/app/common/storages/peripherals/nextcloud_connection_validator.rb
+++ b/modules/storages/app/common/storages/peripherals/nextcloud_connection_validator.rb
@@ -39,6 +39,7 @@ module Storages
         @storage = storage
       end
 
+      # rubocop:disable Metrics/AbcSize
       def validate
         maybe_is_not_configured
           .or { host_url_not_found }
@@ -46,12 +47,15 @@ module Storages
           .or { version_mismatch }
           .or { with_unexpected_content }
           .or { group_folder_not_found }
-          .or { request_failed_with_unknown_error }
+          .or { capabilities_request_failed_with_unknown_error }
+          .or { files_request_failed_with_unknown_error }
           .value_or(ConnectionValidation.new(type: :healthy,
                                              error_code: :none,
                                              timestamp: Time.current,
                                              description: nil))
       end
+
+      # rubocop:enable Metrics/AbcSize
 
       private
 
@@ -164,15 +168,7 @@ module Storages
       # rubocop:disable Metrics/AbcSize
       def with_unexpected_content
         return None() unless @storage.automatic_management_enabled?
-
-        if files.failure?
-          return Some(
-            ConnectionValidation.new(type: :error,
-                                     error_code: :err_unknown,
-                                     timestamp: Time.current,
-                                     description: I18n.t("storages.health.connection_validation.unknown_error"))
-          )
-        end
+        return None() if files.failure?
 
         expected_folder_ids = @storage.project_storages
                                       .where(project_folder_mode: "automatic")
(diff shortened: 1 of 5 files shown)


modules/storages/app/common/storages/peripherals/nextcloud_connection_validator.rb
```diff
@@ -59,6 +60,17 @@ def query
                      .call(storage: @storage, auth_strategy:)
       end
 
+      def files
```
Kharonus, line 68: 🟡 minor, but if we have now a meaningful name to distinguish outbound requests, can we rename the other query to `capabilities`?

modules/storages/app/common/storages/peripherals/nextcloud_connection_validator.rb
```diff
@@ -59,6 +60,17 @@ def query
...
+        @files ||= Peripherals::Registry
+                     .resolve("#{@storage.short_provider_type}.queries.files")
+                     .call(
+                       storage: @storage,
+                       auth_strategy: Peripherals::Registry
```
Kharonus: 🟡 also minor, can we extract the userless auth strategy to a private method, too.

naming the current `noop`, and this one `userless`. IMHO a small improvement of readability

modules/storages/config/locales/en.yml
```diff
@@ -217,7 +217,9 @@ en:
         placeholder: Check your connection against the server.
         subtitle: Connection validation
         tenant_id_wrong: The configured directory (tenant) id is invalid. Please check the configuration.
-        unexpected_content: Unexpected content found in the drive.
+        unexpected_content:
+          nextcloud: Unexpected content found in the managed folder.
```
Kharonus: 🟡 not sure about the AC of this feature, but should we use the wording `in the managed group folder`? we do use it elsewhere, too.

modules/storages/spec/common/storages/peripherals/nextcloud_connection_validator_spec.rb
```diff
@@ -141,11 +150,42 @@
         end
       end
     end
+
+    context "if the request returns unexpected files" do
```
Kharonus: 🔴 missing testcase, what happens if the `files` query returns a failure (not_found can be a good example). This opens then the question, is this maybe even another error? a la: "group folder not set up". Maybe we need an opinion from dominic here.

Kharonus (changes requested):
I do have one bigger and a couple of very small requests.

modules/storages/spec/common/storages/peripherals/nextcloud_connection_validator_spec.rb
apfohl: I will integrate a test for the failure case for now, and we will discuss this case later.

modules/storages/app/common/storages/peripherals/nextcloud_connection_validator.rb
```diff
@@ -45,6 +45,7 @@ def validate
           .or { missing_dependencies }
           .or { version_mismatch }
           .or { request_failed_with_unknown_error }
+          .or { with_unexpected_content }
```
Kharonus, line 48: 🟡 I think, we need to change the order here. first check for unexpected content, then unknown errors. As the unknown errors should catch everything, that is not handled by the previous rules.

PR #16649, opened 2024-09-06
[#57529] Add options for sum totals calculation mode in hierarchies

aaron-contreras:
# Ticket
https://community.openproject.org/work_packages/57529

Adding a new setting in order to allow users to determine the way in which they'd like sum totals of percent complete in work package hierarchies to be calculated.

1. **Simple average (default)**: Work is ignored and the total % Complete will be a simple average of % Complete values of work packages in the hierarchy.
2. **Weighted by work**: The total % Complete will be weighted against the Work of each work package in the hierarchy. Work packages without Work will be ignored.

## Screenshots
Not yet

1. Extend the `UpdateAncestorsService` to calculate the total % complete based on the current Setting value.
2. Add a radio group entry to the progress tracking settings page to be able to toggle between the modes.
3. Add a job that recalculates all total % complete values that gets triggered when the mode is changed in the settings page.

# Merge checklist

- [x] Added/updated tests


spec/features/admin/progress_tracking_spec.rb
```diff
@@ -28,7 +28,9 @@
 
 require "spec_helper"
 
-RSpec.describe "Progress tracking admin page", :cuprite, :js,
+RSpec.describe "Progress tracking admin page",
+               :js,
+               :with_cuprite,
```
cbliard: I was not lucky, I copied it over from the only spec that uses `:cuprite`: `spec/features/admin/languages_settings_spec.rb`. I thought that it had been renamed at some point.

app/services/settings/update_service.rb
```diff
@@ -47,6 +47,10 @@ def set_setting_value(name, value)
     Setting[name] = new_value
     if name == :work_package_done_ratio && old_value != "status" && new_value == "status"
       WorkPackages::Progress::ApplyStatusesChangeJob.perform_later(cause_type: "progress_mode_changed_to_status_based")
+    elsif name == :total_percent_complete_mode && old_value != "work_weighted_average" && new_value == "work_weighted_average"
+      WorkPackages::Progress::ApplyTotalPercentCompleteModeChangeJob
+        .perform_later(mode: new_value,
+                       cause_type: "total_percent_complete_mode_changed_to_work_weighted_average")
```
cbliard: Does it mean that both "Progress calculation mode" and "total % complete calculation mode" can't be changed at the same time?

cbliard (commented):
Nice work so far 👏 

I only looked at code so far. I need to play a bit with it now.

spec/services/work_packages/update_ancestors_service_spec.rb
```diff
@@ -795,6 +796,213 @@ def call_update_ancestors_service(work_package)
     end
   end
 
+  describe "simple average mode for total % complete calculation",
+           with_settings: { total_percent_complete_mode: "simple_average" } do
+    subject(:call_result) do
+      described_class.new(user:, work_package: parent)
+                      .call(%i(remaining_hours))
+    end
```
cbliard: That's not exactly how this class is supposed to be called: it is called on a child which has been modified, and we check that the ancestors got updated as well.

Calling it like you did actually hides a bug: in the `WorkPackages::UpdateAncestors::Loader`, there can be structured data (the `WorkPackageLikeStruct`), but also plain `WorkPackage` instances, and they not respond to `status_excluded_from_totals`.

I'll fix that.

cbliard: Fixed in 1874fb3003e.

app/services/work_packages/update_ancestors_service.rb
```diff
@@ -120,15 +120,43 @@ def derive_done_ratio(ancestor, loader)
...
 
+  def calculate_simple_average_percent_complete(work_package, loader)
+    all_done_ratios = children_done_ratio_values(work_package, loader)
+
+    if work_package.done_ratio.present? && !work_package.status.excluded_from_totals
+      all_done_ratios << work_package.done_ratio
+    end
+
+    progress = all_done_ratios.sum.to_f / all_done_ratios.count
+    progress.round
```
cbliard, line 152: Found another issue: this fails if all work packages are excluded: the sum returns 0.0, count returns 0, so progress will be `NaN`, and `NaN.round` raises.

In this case, ∑ % complete should be cleared.

I'll fix that.

cbliard: Fixed in f0408482463

app/services/settings/update_service.rb
cbliard: Ah, I did not think enough for this one. It's called once for each modified setting, so it will be called for both and 2 jobs will be triggered.
It's probably not ideal to have the journal being updated twice, but that will do for now.

app/workers/work_packages/progress/sql_commands.rb
```diff
@@ -188,7 +188,6 @@ def update_totals
     if Setting.total_percent_complete_mode == "work_weighted_average"
       update_total_percent_complete_in_work_weighted_average_mode
     elsif Setting.total_percent_complete_mode == "simple_average"
-      update_total_percent_complete_in_work_weighted_average_mode
```
cbliard: Nice!

cbliard (approved):
Well done. I think it's good to go now.

PR #16652, opened 2024-09-06
Save 2 seconds by clicking redirect link directly in a test

cbliard:
It can be done because this has been fixed in 8fe4a222148.

And fix the redirection not working properly if the link is clicked. Fixed by explicitly setting `window.href`.

### DIFF
--- a/frontend/src/stimulus/controllers/dynamic/job-status-polling.controller.ts
+++ b/frontend/src/stimulus/controllers/dynamic/job-status-polling.controller.ts
@@ -40,14 +40,18 @@ export default class JobStatusPollingController extends Controller<HTMLElement> 
     setTimeout(() => element.click(), 50);
   }
 
-  redirectClick(_:Event) {
-    this.userInteraction = true;
+  redirectClick(event:PointerEvent) {
+    this.followLink(event.target as HTMLLinkElement);
   }
 
   redirectTargetConnected(element:HTMLLinkElement) {
     setTimeout(() => {
-      this.userInteraction = true;
-      window.location.href = element.href;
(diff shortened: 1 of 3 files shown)


frontend/src/stimulus/controllers/dynamic/job-status-polling.controller.ts
```diff
@@ -40,14 +40,18 @@ export default class JobStatusPollingController extends Controller<HTMLElement>
     setTimeout(() => element.click(), 50);
   }
 
-  redirectClick(_:Event) {
-    this.userInteraction = true;
+  redirectClick(event:PointerEvent) {
+    this.followLink(event.target as HTMLLinkElement);
```
as-op, line 45: Before the change the actual navigating was done by the browser. 
As we now set windows.href ourselves, i think it's best to ensure stopping the browser one via the event.

```suggestion
    event.preventDefault();
    this.followLink(event.target as HTMLLinkElement);
```

as-op (changes requested):
👍  Thank you for following up on this!

frontend/src/stimulus/controllers/dynamic/job-status-polling.controller.ts
cbliard: Thanks, I committed your suggestion.

I think the behavior when clicking is intercepted by turbo-something, and it tries to replace the turbo-frames from the current page with the turbo-frames from the response, which in turn totally skips the redirection which normally happens. Not entirely sure though.

as-op (approved):
👍

PR #16653, opened 2024-09-06
[#55234] Fix text filter width on project list

dombesz:
# Ticket
https://community.openproject.org/work_packages/55234

Add a larger width to the project list search input.

## Screenshots
image

Added an override for the project list input filter, because the primer provided css was not working.

# Merge checklist

- [ ] Added/updated tests


HDinger (commented):
The current solution is weird, as the override is basically setting the width and not the `input_width` which should be responsible for it. We should fix it in the primer repo as soon as possible. Due to the time pressure, I am fine with merging this for now.

PR #16676, opened 2024-09-09
[#57037] Migrate UserPreference settings with the "default" theme to "light"

aaron-contreras:
# Ticket
https://community.openproject.org/work_packages/57037

A lot of areas were broken display-wise when upgrading from 9.x versions. This happened if users still had in their `UserPreference` theme settings the `default` option. I'm bringing these "default" values to "light" which is the current default setting for the theme setting.

## Screenshots
**Before**
Screenshot showing the transparency issue

**After**
Screenshot showcasing the fixed transparency

This was missed when upgrading OpenProject at some point so this should bring everything back in order for users that still have the "default" theme setting which is no longer supported, and keep the preferences for said users who have since moved to one of the valid options like "light" or "light_high_contrast".

A migration should suffice.

# Merge checklist

- [x] Added/updated tests

### DIFF
--- a/db/migrate/20240909151818_migrate_legacy_default_theme_to_light_theme.rb
+++ b/db/migrate/20240909151818_migrate_legacy_default_theme_to_light_theme.rb
@@ -29,29 +29,29 @@
 #++
 
 class MigrateLegacyDefaultThemeToLightTheme < ActiveRecord::Migration[7.1]
-  class MigrationUserPreference < ActiveRecord::Base
-    self.table_name = "user_preferences"
-  end
-
   # Migrate legacy `default` theme to `light` theme
   #
   # Moving from:
   # {
   #   "settings": {
-  #     "theme": "default"
+  #     "theme": "default",
+  #     "foo": "bar"
   #   }
   # }
   #
   # To:
   # {
   #   "settings": {
-  #     "theme": "light" # current default theme
+  #     "theme": "light", # current default theme
+  #     "foo": "bar"
   #   }
   # }
   def up
-    MigrationUserPreference
-      .where(settings: { theme: "default" })
-      .update_all(settings: { theme: "light" })
+    execute <<-SQL.squish
+      UPDATE user_preferences
+      SET settings = jsonb_set(settings, '{theme}', '"light"')
+      WHERE settings->>'theme' = 'default';
+    SQL
   end
 
   # no-op
--- a/db/migrate/20240909151818_migrate_legacy_default_theme_to_light_theme.rb
+++ b/db/migrate/20240909151818_migrate_legacy_default_theme_to_light_theme.rb
@@ -1,0 +1,59 @@
+# frozen_string_literal: true
+
+#-- copyright
+# OpenProject is an open source project management software.
+# Copyright (C) the OpenProject GmbH
+#
+# This program is free software; you can redistribute it and/or
+# modify it under the terms of the GNU General Public License version 3.
+#
+# OpenProject is a fork of ChiliProject, which is a fork of Redmine. The copyright follows:
+# Copyright (C) 2006-2013 Jean-Philippe Lang
+# Copyright (C) 2010-2013 the ChiliProject Team
+#
+# This program is free software; you can redistribute it and/or
+# modify it under the terms of the GNU General Public License
+# as published by the Free Software Foundation; either version 2
+# of the License, or (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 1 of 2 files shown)


db/migrate/20240909151818_migrate_legacy_default_theme_to_light_theme.rb
```diff
@@ -0,0 +1,59 @@
...
+# along with this program; if not, write to the Free Software
+# Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+#
+# See COPYRIGHT and LICENSE files for more details.
+#++
+
+class MigrateLegacyDefaultThemeToLightTheme < ActiveRecord::Migration[7.1]
+  class MigrationUserPreference < ActiveRecord::Base
+    self.table_name = "user_preferences"
+  end
```
toy: As the class is nested, I think it is cleaner to name it `UserPreference` and not specify `table_name`

db/migrate/20240909151818_migrate_legacy_default_theme_to_light_theme.rb
```diff
@@ -0,0 +1,59 @@
...
+  # {
+  #   "settings": {
+  #     "theme": "light" # current default theme
+  #   }
+  # }
+  def up
+    MigrationUserPreference
+      .where(settings: { theme: "default" })
+      .update_all(settings: { theme: "light" })
+  end
```
toy: This will not find settings if anything but theme is set, which is probably the main case.
I assume either walking using `find_each` and filtering/updating in ruby, or updating in SQL, finding using `jsonb_extract_path` and updating using `jsonb_set` should be done. SQL way should be faster, so less chances of overwriting settings if user is changing them in parallel.

spec/migrations/migrate_legacy_default_theme_to_light_theme_spec.rb
```diff
@@ -0,0 +1,56 @@
...
+RSpec.describe MigrateLegacyDefaultThemeToLightTheme, type: :model do
+  RSpec::Matchers.define_negated_matcher :not_change, :change
+
+  # Silencing migration logs, since we are not interested in that during testing
+  subject(:run_migration) { ActiveRecord::Migration.suppress_messages { described_class.new.up } }
+
+  shared_let(:user_with_default_theme) { create(:user, preferences: { settings: { theme: "default" } }) }
+  shared_let(:user_with_light_theme) { create(:user, preferences: { settings: { theme: "light" } }) }
+  shared_let(:user_with_light_high_contrast_theme) { create(:user, preferences: { settings: { theme: "light_high_contrast" } }) }
+  shared_let(:user_with_dark_theme) { create(:user, preferences: { settings: { theme: "dark" } }) }
```
toy: Something besides theme should be set, `settings: { foo: "bar", theme: "default" }`

toy (changes requested):

db/migrate/20240909151818_migrate_legacy_default_theme_to_light_theme.rb
aaron-contreras: I just followed the pattern in other migrations for naming

db/migrate/20240909151818_migrate_legacy_default_theme_to_light_theme.rb
aaron-contreras: Ah interesting, didn't think of this. Now looking at the SQL that's spit out, you're right. Thought it was a partial match and inferred that it's a jsonb column. Nice catch

aaron-contreras: I actually can't replicate the above. I've ran this migration against a couple of different scenarios with different users and it does find and update the right columns. But I understand the plain SQL way is probably a bit faster.

db/migrate/20240909151818_migrate_legacy_default_theme_to_light_theme.rb
toy: Indeed a lot have prefix or suffix, but some don't. I can't come up with a big benefit of the prefix, maybe it is just a bit of cargo cult

db/migrate/20240909151818_migrate_legacy_default_theme_to_light_theme.rb
toy: I'm not sure I get your second comment.
`MigrationUserPreference.where(settings: { theme: "default" }).update_all(settings: { theme: "light" })` will run following SQL: `UPDATE "user_preferences" SET "settings" = $1 WHERE "user_preferences"."settings" = $2  [["settings", "{\"theme\":\"light\"}"], ["settings", "{\"theme\":\"default\"}"]]`, so it will only update rows with one key value pair.

toy (approved):

PR #16688, opened 2024-09-10
[#57664] ActiveRecord::RecordNotFound in ActivitiesController#index

ba1ash:
# Ticket

https://community.openproject.org/work_packages/57664

Fix error when viewing inactive user activity.

`#determine_author` can raise `ActiveRecord::RecordNotFound` if user does not exist or is inactive. Use `rescue_from` to catch it and send appropriate response.
## Screenshots

image

# Merge checklist

- [x] Added/updated tests

### DIFF
--- a/spec/controllers/activities_controller_spec.rb
+++ b/spec/controllers/activities_controller_spec.rb
@@ -88,6 +88,16 @@ RSpec.describe ActivitiesController do
           expect(response).to have_http_status(:not_found)
         end
       end
+
+      describe "inactive user activity" do
+        let!(:inactive_user) { create(:user, status: User.statuses[:locked]) }
+
+        it "renders 404" do
+          get "index", params: { user_id: inactive_user.id }
+          expect(response).to have_http_status(:not_found)
+          expect(response).to render_template "common/error"
+        end
+      end
     end
 
     describe "with activated activity module" do
--- a/app/controllers/activities_controller.rb
+++ b/app/controllers/activities_controller.rb
@@ -45,20 +45,18 @@ class ActivitiesController < ApplicationController
 
   accept_key_auth :index
 
+  rescue_from ActiveRecord::RecordNotFound do |exception|
+    op_handle_warning "Failed to find all resources in activities: #{exception.message}"
+    render_404(message: I18n.t(:error_can_not_find_all_resources))
+  end
+
   def index
     @events = @activity.events(from: @date_from.to_datetime, to: @date_to.to_datetime)
 
     respond_to do |format|
-      format.html do
-        respond_html
-      end
-      format.atom do
-        respond_atom
-      end
+      format.html { respond_html }
+      format.atom { respond_atom }
     end
-  rescue ActiveRecord::RecordNotFound => e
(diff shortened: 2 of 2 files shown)


mereghost (approved):

spec/controllers/activities_controller_spec.rb
```diff
@@ -101,6 +101,14 @@
         expect(response).to be_successful
         expect(response).to render_template "index"
       end
+
+      describe "with not existent user_id" do
+        it "renders 404" do
+          get "index", params: { project_id: project.id, user_id: 123123123 }
+          expect(response).to have_http_status(:not_found)
+          expect(response).to render_template "common/error"
+        end
```
cbliard: Can you add a test with an inactive user instead? It would better reflect what happened in production.

cbliard (changes requested):
Thanks for the fix. I have a few remarks.

The PR description is lacking explanations, making it harder to review. It could make it even harder for any future developer stumbling upon it. Same for the commit message, it's not helpful. It should describe what you fixed and why. The screenshot is not helpful as well: I thought it was the error happening before the fix. Just a label saying "After fix" would have helped.

If I understand correctly, the user is fetched in `#determine_author`. It looks up only active users. If the user is inactive (locked for instance), then the method will raise an `ActiveRecord::RecordNotFound` error. The `index` action is intended to catch it, but it can't work outside of this method. `determine_author` being called in a before action, that's why it's not caught.

You fixed it by using `rescue_from` helper which works in action AND in action hooks, fixing the issue.

Here is a proposal for the commit message:
```
Fix error when viewing inactive user activity

`#determine_author` can raise `ActiveRecord::RecordNotFound` if user does not
exist or is inactive. Use `rescue_from` to catch it and send appropriate
response.

Fixes https://community.openproject.org/wp/57664
Fixes https://appsignal.com/openproject-gmbh/sites/66b224a4d30d867bed8a1772/exceptions/incidents/461
```

Can you also modify your test to test against an inactive user instead of a non-existing user id?

Also this is not a critical bug and it's not a regression either, so it should target current dev branch.

ba1ash:
@cbliard Thanks for your review!
Sure, I can do it.

spec/controllers/activities_controller_spec.rb
ba1ash: Yes, of course.

spec/controllers/activities_controller_spec.rb
```diff
@@ -88,6 +88,16 @@
           expect(response).to have_http_status(:not_found)
         end
       end
+
+      describe "inactive user activity" do
+        let!(:inactive_user) { create(:user, status: User.statuses[:locked]) }
```
cbliard, line 93: That's valid.

You can also use `create(:locked_user)` next time. It takes care of setting the status to `User.statuses[:locked]` among other things.

cbliard (changes requested):
Thanks for the changes.

It still needs to target `dev` and not `release/14.5` because it's not a critical bug nor a regression. Even if there should be no issue with it that's how we agreed to work for bugs.

cbliard (approved):
Nice! Thanks a lot for bringing it in.

PR #16692, opened 2024-09-10
Ambiguous button_save_as translation fix

top4ek:
Remove ambiguous translation in Reporting module. This leads us to different translations in production and development modes at least.

## Screenshots
There're two translations of button_save_as menu item. [Here](https://github.com/opf/openproject/blob/dev/config/locales/en.yml#L1456) and [here](https://github.com/opf/openproject/blob/dev/modules/reporting/config/locales/en.yml#L35). 

Use new translation key button_save_report_as inside reporting module leaving global button_save_as untouched.


oliverguenther (approved):
Good catch, thanks.

Please note that the translations within the `crowdin/` folders will be lost on the next translation update. They are being generated from crowdin (https://www.openproject.org/docs/development/translate-openproject/)

PR #16694, opened 2024-09-10
Implementation/57516 projects should be removable from list with the help of a menu item

akabiru:
# Ticket

https://community.openproject.org/work_packages/57516

A project should be removable from a custom field on the project list via a menu action item

## Screenshots

Kapture 2024-09-13 at 13 38 48

> [!IMPORTANT]  
> Adding a primary key to an existing table has a seemingly higher opportunity cost compared to the
alternative; composite primary keys which will have the carrying cost of introducing different data structure-
which have a higher chance of introducing conflicts or requiring additional handling (ergo- _carrying cost_
>
> In the migration- defining a PRIMARY KEY via Rails migration ensures an autoincremting column is defined as a BIGSERIAL type and existing records are backfilled automatically
>

### DIFF
--- a/app/components/admin/custom_fields/custom_field_projects/row_component.rb
+++ b/app/components/admin/custom_fields/custom_field_projects/row_component.rb
@@ -43,7 +43,7 @@ module Admin
         private
 
         def more_menu_detach_project
-          if User.current.admin && project.active?
+          if User.current.allowed_in_project?(:select_custom_fields, project)
             {
               scheme: :default,
               icon: nil,
--- a/app/controllers/admin/custom_fields/custom_field_projects_controller.rb
+++ b/app/controllers/admin/custom_fields/custom_field_projects_controller.rb
@@ -31,6 +31,7 @@
 class Admin::CustomFields::CustomFieldProjectsController < ApplicationController
   include OpTurbo::ComponentStream
   include OpTurbo::DialogStreamHelper
+  include ApplicationComponentStreams
   include FlashMessagesOutputSafetyHelper
 
   layout "admin"
--- a/app/controllers/admin/custom_fields/custom_field_projects_controller.rb
+++ b/app/controllers/admin/custom_fields/custom_field_projects_controller.rb
@@ -133,8 +133,8 @@ class Admin::CustomFields::CustomFieldProjectsController < ApplicationController
   end
 
   def find_custom_field_project_to_destroy
-    @project = Project.find(params.to_unsafe_h[:custom_fields_project][:project_id])
-    @custom_field_project = CustomFieldsProject.find_by!(custom_field: @custom_field, project: @project)
+    @custom_field_project = CustomFieldsProject.find_by!(custom_field: @custom_field,
+                                                         project: params[:custom_fields_project][:project_id])
   rescue ActiveRecord::RecordNotFound
     respond_with_project_not_found_turbo_streams
   end
--- a/app/controllers/admin/custom_fields/custom_field_projects_controller.rb
+++ b/app/controllers/admin/custom_fields/custom_field_projects_controller.rb
@@ -31,6 +31,7 @@
 class Admin::CustomFields::CustomFieldProjectsController < ApplicationController
   include OpTurbo::ComponentStream
   include OpTurbo::DialogStreamHelper
+  include FlashMessagesOutputSafetyHelper
 
   layout "admin"
 
@@ -39,9 +40,10 @@ class Admin::CustomFields::CustomFieldProjectsController < ApplicationController
   before_action :require_admin
   before_action :find_model_object
 
-  before_action :available_custom_fields_projects_query, only: :index
+  before_action :available_custom_fields_projects_query, only: %i[index destroy]
   before_action :initialize_custom_field_project, only: :new
   before_action :find_projects_to_activate_for_custom_field, only: :create
+  before_action :find_custom_field_project_to_destroy, only: :destroy
 
   menu_item :custom_fields
 
@@ -70,6 +72,23 @@ class Admin::CustomFields::CustomFieldProjectsController < ApplicationController
     end
 
     respond_to_with_turbo_streams(status: create_service.success? ? :ok : :unprocessable_entity)
+  end
+
+  def destroy
+    delete_service = ::CustomFields::CustomFieldProjects::DeleteService
+                         .new(user: current_user, model: @custom_field_project)
+                         .call
+
+    delete_service.on_success { render_project_list(url_for_action: :index) }
+
+    delete_service.on_failure do
+      update_flash_message_via_turbo_stream(
+        message: join_flash_messages(delete_service.errors.full_messages),
+        full: true, dismiss_scheme: :hide, scheme: :danger
+      )
+    end
+
+    respond_to_with_turbo_streams(status: delete_service.success? ? :ok : :unprocessable_entity)
   end
 
   def default_breadcrumb; end
@@ -112,6 +131,18 @@ class Admin::CustomFields::CustomFieldProjectsController < ApplicationController
   rescue ActiveRecord::RecordNotFound
     update_flash_message_via_turbo_stream message: t(:notice_project_not_found), full: true, dismiss_scheme: :hide,
                                           scheme: :danger
+    update_project_list_via_turbo_stream
+
+    respond_with_turbo_streams
+  end
+
+  def find_custom_field_project_to_destroy
+    @project = Project.find(params.to_unsafe_h[:custom_fields_project][:project_id])
+    @custom_field_project = CustomFieldsProject.find_by!(custom_field: @custom_field, project: @project)
+  rescue ActiveRecord::RecordNotFound
+    update_flash_message_via_turbo_stream(
+      message: t(:notice_project_not_found), full: true, dismiss_scheme: :hide, scheme: :danger
+    )
     update_project_list_via_turbo_stream
 
     respond_with_turbo_streams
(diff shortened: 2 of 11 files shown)


app/services/custom_fields/custom_field_projects/delete_service.rb
```diff
@@ -0,0 +1,53 @@
...
+
+      # `custom_fields_projects` table has no `id` column, hence no primary key. #destroy method would not work
+      # Note: `delete_all` goes straight to the database and does not trigger callbacks
+      #
+      # @return [Integer] number of rows deleted
+      def delete(custom_field_id:, project_id:)
+        CustomFieldsProject.transaction do
+          CustomFieldsProject.where(custom_field_id:, project_id:).delete_all
+        end
+      end
```
akabiru: \cc @ulferts any thoughts on this approach?

app/controllers/admin/custom_fields/custom_field_projects_controller.rb
```diff
@@ -110,11 +129,14 @@ def find_projects_to_activate_for_custom_field
...
   rescue ActiveRecord::RecordNotFound
-    update_flash_message_via_turbo_stream message: t(:notice_project_not_found), full: true, dismiss_scheme: :hide,
-                                          scheme: :danger
-    update_project_list_via_turbo_stream
+    respond_with_project_not_found_turbo_streams
+  end
 
-    respond_with_turbo_streams
+  def find_custom_field_project_to_destroy
+    @project = Project.find(params.to_unsafe_h[:custom_fields_project][:project_id])
```
ulferts: I am rather sure that you don't need the `to_unsafe_h` here. You could just write

```
@project = Project.find(params[:custom_fields_project][:project_id])
```

ulferts: Come to think of it, I don't understand why the project is fetched at all. Couldn't it also be written like this:

```
@custom_field_project = CustomFieldsProject.find_by!(custom_field: @custom_field, 
                                                     project: params[:custom_fields_project][:project_id])
```

That of course is only true if `@project` is really not used at all. But you'd safe a database request if it needn't be loaded.

app/services/custom_fields/custom_field_projects/delete_service.rb
ulferts: I think it is ok to do this given what the data structure currently offers. 

But there might be a better way of doing it. I don't have first hand experience with it yet, though, so I don't particularly know the limitations. Since [Rails 7.1.2 composite primary keys are supported](https://blog.saeloun.com/2024/08/06/active-record-composite-primary-key/). I still don't quite understand how much of [this](https://guides.rubyonrails.org/active_record_composite_primary_keys.html) is only available in edge and how much is already in 7.1.2. so you'd have to look that up for yourself. 

But my quick test where I changed the `CustomFieldsProject` to include the specification for the composite primary key

```
class CustomFieldsProject < ApplicationRecord
  self.primary_key = %i[project_id custom_field_id]

  belongs_to :project
  belongs_to :custom_field
end
```

Let me to already do this in the console:

```
[3] pry(main)> CustomFieldsProject.find([992, 152]).destroy
  CustomFieldsProject Load (0.4ms)  SELECT "custom_fields_projects".* FROM "custom_fields_projects" WHERE "custom_fields_projects"."project_id" = $1 AND "custom_fields_projects"."custom_field_id" = $2 LIMIT $3  [["project_id", 992], ["custom_field_id", 152], ["LIMIT", 1]]
  TRANSACTION (0.1ms)  BEGIN
  CustomFieldsProject Destroy (0.7ms)  DELETE FROM "custom_fields_projects" WHERE "custom_fields_projects"."project_id" = $1 AND "custom_fields_projects"."custom_field_id" = $2  [["project_id", 992], ["custom_field_id", 152]]
  TRANSACTION (2.2ms)  COMMIT
=> #<CustomFieldsProject:0x0000000332b9e620 custom_field_id: 152, project_id: 992>
```

I can now also query for e.g. `CustomFieldsProject.last`. It still isn't a proper composite key in the database but it might well be good enough.

If that doesn't work, it might still be worthwhile to add an id column to the table and work through the necessary changes. It will probably make a whole lot of code simpler (e.g. in the controller). But I'd understand if you don't want to do that now.

app/components/admin/custom_fields/custom_field_projects/row_component.rb
```diff
@@ -37,8 +37,31 @@ def wrapper_uniq_by
...
 
         def more_menu_items
-          []
+          @more_menu_items ||= [more_menu_detach_project].compact
         end
+
+        private
+
+        def more_menu_detach_project
+          if User.current.admin && project.active?
```
ulferts: You could also query for `User.current.allowed_to?(:select_custom_fields, project)`. This is more specific than the admin check (only admins will visit that page anyway), but allows to later on open up the page to non admins more easily. At the same time, it will check for the project being active as no permissions are granted on archived projects.

ulferts (commented):
Hey @akabiru, 

I did not do a proper review as I only browsed through the code and didn't check out the functionality itself. I have a proposal on how you could add a composite primary key and by that work around the missing id column. Other than that, I added smaller remarks.

app/services/custom_fields/custom_field_projects/delete_service.rb
akabiru: Thank you! In the end adding an ID column seems like the better solution, albeit with higher opportunity cost compared to the composite primary keys which will have the carrying cost of a different data structure- which I have already hit issues with in one part of the application (ng auto-completer)

app/controllers/admin/custom_fields/custom_field_projects_controller.rb
akabiru: Great catch, missed that! 👍🏾

app/components/admin/custom_fields/custom_field_projects/row_component.rb
akabiru: Good call! 👍🏾

db/migrate/20240917105829_add_primary_key_to_custom_fields_projects.rb
```diff
@@ -0,0 +1,50 @@
...
+          UPDATE custom_fields_projects
+          SET id = cte.row_num
+          FROM cte
+          WHERE custom_fields_projects.project_id = cte.project_id
+            AND custom_fields_projects.custom_field_id = cte.custom_field_id;
+        SQL
+      end
+    end
+  end
+end
```
akabiru: > [!IMPORTANT]  
> Adding a primary key to an existing table has a seemingly higher opportunity cost compared to the
alternative; [composite primary keys](https://guides.rubyonrails.org/active_record_composite_primary_keys.html) which will have the carrying cost of introducing different data structure- which have a higher chance of introducing conflicts or requiring additional handling (ergo- _carrying cost_
>
> In the migration- defining a PRIMARY KEY via Rails migration ensures an auto-incrementing column is defined as a _BIGSERIAL_ type with the default value set to `nextval('custom_fields_projects_id_seq'::regclass)` and existing records are backfilled via `row_number()` pg window function
> 

---

```ruby
Migrating to AddPrimaryKeyToCustomFieldsProjects (20240917105829)
== 20240917105829 AddPrimaryKeyToCustomFieldsProjects: migrating ==============
-- add_column(:custom_fields_projects, :id, :primary_key)
  TRANSACTION (0.8ms)  BEGIN
   (12.7ms)  ALTER TABLE "custom_fields_projects" ADD "id" bigserial primary key
   -> 0.0132s
-- execute("WITH cte AS ( SELECT row_number() OVER () AS row_num, project_id, custom_field_id FROM custom_fields_projects ) UPDATE custom_fields_projects SET id = cte.row_num FROM cte WHERE custom_fields_projects.project_id = cte.project_id AND custom_fields_projects.custom_field_id = cte.custom_field_id;")
   (1.1ms)  WITH cte AS ( SELECT row_number() OVER () AS row_num, project_id, custom_field_id FROM custom_fields_projects ) UPDATE custom_fields_projects SET id = cte.row_num FROM cte WHERE custom_fields_projects.project_id = cte.project_id AND custom_fields_projects.custom_field_id = cte.custom_field_id;
   -> 0.0015s
== 20240917105829 AddPrimaryKeyToCustomFieldsProjects: migrated (0.0150s) =====
```

```sql
                                Table "public.custom_fields_projects"
     Column      |  Type  | Collation | Nullable |                      Default
-----------------+--------+-----------+----------+----------------------------------------------------
 custom_field_id | bigint |           | not null |
 project_id      | bigint |           | not null |
 id              | bigint |           | not null | nextval('custom_fields_projects_id_seq'::regclass)
Indexes:
    "custom_fields_projects_pkey" PRIMARY KEY, btree (id)
    "index_custom_fields_projects_on_custom_field_id_and_project_id" UNIQUE, btree (custom_field_id, project_id)
Foreign-key constraints:
    "fk_rails_12fb30588e" FOREIGN KEY (project_id) REFERENCES projects(id) ON UPDATE CASCADE ON DELETE CASCADE
    "fk_rails_e51cefe60d" FOREIGN KEY (custom_field_id) REFERENCES custom_fields(id) ON UPDATE CASCADE ON DELETE CASCADE
```

cc @ulferts @machisuji

akabiru: https://github.com/opf/openproject/pull/16694/commits/922bd158ef38b08acffa45737b0f4cab4b219e50

> _An ActiveRecord primary key migration translates to a BIGSERIAL PRIMARY KEY
constraint in Postgres. BigSerial is an auto-incrementing numeric type so there
is no need to manually backfill_

ba1ash (approved):

PR #16695, opened 2024-09-10
Attempt to fix lost oauth token issue.

ba1ash:
# Ticket
https://community.openproject.org/wp/57835

It seems like enabling optimistic locking in https://github.com/opf/openproject/pull/16376 was not enough. After `ActiveRecord::StaleObjectError` is raised the sql request for token is cached. Then there is an attempt to refresh stale token, NC answers with `invalid_request` and we delete the token. Actually, the deletion should not happen, because the object should be stale... Anyways, it is just an attempt to fix this.

## Screenshots

This PR makes sure request for OAuthClientToken is not cached by ActiveRecord query cache..

# Merge checklist

- [ ] Added/updated tests

### DIFF
--- a/modules/storages/app/common/storages/peripherals/storage_interaction/authentication_strategies/oauth_user_token.rb
+++ b/modules/storages/app/common/storages/peripherals/storage_interaction/authentication_strategies/oauth_user_token.rb
@@ -77,8 +77,11 @@ module Storages
               return Failures::Builder.call(code: :error, log_message:, data:)
             end
 
-            current_token = OAuthClientToken.find_by(user: @user,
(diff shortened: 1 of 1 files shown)


wielinde:
@ba1ash if that is really causing the error then optimistic locking in rails is broken, right? We should not care about caching. that is a lower implementation level at which "ideally" we should not be worrying about, right?

ba1ash:
> @ba1ash if that is really causing the error then optimistic locking in rails is broken, right? We should not care about caching. that is a lower implementation level at which "ideally" we should not be worrying about, right?

@wielinde cached sql request is only a part of the issue. I updated the work package with a diagram that show that optimistic locking is not enough.

PR #16713, opened 2024-09-12
[#57856] Export modal should not show a scrollbar for default settings

as-op:
# Ticket

https://community.openproject.org/work_packages/57856

Avoid showing a scrollbar for no apparent reason.

## Screenshots

| before | after |
|---|---|
| scrollbar | noscollbar|

1) Reduce the standard padding of the spacer. Solution suggested by Marc.
2) Remove the padding below the Angular drop & drop component. To fix the scrollbar for even more other settings combinations. To avoid breaking other places where this component is used, the padding is only removed for the export modal.

> [!NOTE]
> There will always be a valid reason to show the scrollbar for certain combinations,
> as it depends on: screen size, number of columns and column rows, column name lengths and their translations

# Merge checklist

### DIFF
--- a/app/components/_index.sass
+++ b/app/components/_index.sass
@@ -8,3 +8,4 @@
 @import "filter/filters_component"
 @import "projects/row_component"
 @import "op_primer/border_box_table_component"
+@import "work_packages/exports/modal_dialog_component"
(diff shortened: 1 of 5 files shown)


aaron-contreras (approved):
Nice 👍

PR #16719, opened 2024-09-12
Make expandable list button text more specific

cbliard:
# Ticket

No ticket initially, then scope increased with the feedbacks so this ticket was created to track the changes: https://community.openproject.org/projects/meetings-stream/work_packages/57911/activity

I don't like this label:

image

"Show/hide 30 items". That does not sound right. And I suppose screen readers do not read that properly either.
I'd like to display "Show 30 items" when it's hidden, and "Hide 30 items" when it's shown.

I was annoyed by that for a long time, and today I stumbled upon it while reading the lookbook, so why not fix it.

Now there are more to it:
* Change the phrasing of "Show 4 items"/"Hide 4 items" at the end of the participant section with:
  * "Show 4 more" (when collapsed)
  * "Show less" (when expanded)
* Increase margin above the Show link from 4px to 8px
* Remove the text ("Attached files are available...") from below the "Participants" heading
  This text seems to have been introduced as a bug in a prior change (regression)

## Screenshots

Before:

image

Now:

image

image

There are two buttons. One "Show", one "Hide". Only one is displayed at once. Clicking one changes the one being displayed.

Also fixed the action which did not trigger when pressing "enter". By not passing the action name explicitly, stimulus uses the default action, which should always work better, and is less verbose code-wise.

Also support @cuttoff_limit being nil, in which case all items are displayed (because when changing the value in lookbook, it crashed when I cleared the field before setting a new value).

Also increased to margin of the button from 4px to 8px as agreed in [frontend meeting](https://community.openproject.org/projects/gmbh/meetings/3590).

# Merge checklist

- [ ] Added/updated tests
  - nope, that's just a minor refactoring. I'll update existing tests if some fail.

### DIFF
--- a/app/components/op_primer/expandable_list_component.html.erb
+++ b/app/components/op_primer/expandable_list_component.html.erb
@@ -2,7 +2,7 @@
   component_wrapper(data: wrapper_data_attributes) do
     if elements.count == 0
       render(Primer::Beta::Text.new(color: :subtle)) { t("label_meeting_no_participants") }
-    elsif elements.count <= @cutoff_limit
+    elsif @cutoff_limit.nil? || elements.count <= @cutoff_limit
       flex_layout do |list|
         elements.each do |item|
           list.with_row(mt: 1) { item.to_s }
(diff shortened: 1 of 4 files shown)


oliverguenther (approved):
Nice 👏

psatyal:
Looks much better, @cbliard. Thank you!

Screenshot 2024-09-13 at 11 06 24
Screenshot 2024-09-13 at 11 06 32

Some feedback:

- The spacing between the users and "Show 4 items" is okay now.
- The term "item" was always an odd choice. Let's change the text to "Show 6 more" and "Show less". (Or simply "Show more" and "Show less")
- The text below "Participants" is wrong; the participant section is not supposed to have text at all. This seems like a regression. Could you please remove it? The spacing also seems too much between the end of the text and the first participant (32px vs. 16px). 

This increases a scope a little bit. I'll create a WP with these requirements if it's okay with you.

Cheers
Parimal

psatyal:
Created work package: https://community.openproject.org/projects/meetings-stream/work_packages/57911/activity

bsatarnejad (approved):
LGTM 👍

PR #16736, opened 2024-09-16
Configurable dev ports

toy:
Make it possible to configure both backend and frontend host/port using environment variables (or .env file).

# Merge checklist

- [ ] Added/updated tests


frontend/cli_to_rails_proxy.js
```diff
@@ -1,9 +1,10 @@
 const PROXY_HOSTNAME = process.env.PROXY_HOSTNAME || 'localhost';
+const PORT = process.env.PORT || '3000';
 
 const PROXY_CONFIG = [
   {
     "context": ['/**'],
-    "target": `http://${PROXY_HOSTNAME}:3000`,
+    "target": `http://${PROXY_HOSTNAME}:${PORT}`,
```
cbliard, line 7: I'm wondering what this `PROXY_HOSTNAME` is supposed to be.

In `frontend/cli_to_rails_proxy.js`, it's used with port 3000 so it seems to be the backend host, and in `frontend/package.json` it's used with port 4200 so it looks like it's the frontend host.

Shouldn't this `PROXY_HOSTNAME` be removed entirely and use `HOST` or `FE_HOST` instead?
If so, the documentation at `docs/development/running-tests/README.md:704` should be updated too.

`PROXY_HOSTNAME` is also used in `docker-compose.yml`, but I do not fully understand how it is supposed to work in combination with `HOST` and `PORT` as I'm not using this docker setup regularly.

cbliard (commented):
I love it 👏 

I will happily replace my crappy aliases to use these variables instead when starting my dev servers

frontend/cli_to_rails_proxy.js
toy: I was really not sure about touching `PROXY_HOSTNAME`, as it is in compose file, but you may be right and it should be replaced with `HOST`.
I'm also not sure if there is a need for separate `HOST` and `FE_HOST` variables, maybe there should be just `HOST`, `PORT` and `FE_PORT`.

cbliard: Probably someone from the Titan team can help, as they are all using docker.

machisuji:
@toy 

On this branch when accessing `localhost:4200` - which usually works just fine just the same as `localhost:3000` - I get a proxy error:

```
11:17:20 angular.1 | <e> [webpack-dev-server] [HPM] Error occurred while proxying request localhost:4200/ to http://localhost:3100/ [ECONNREFUSED] (https://nodejs.org/api/errors.html#errors_common_system_errors)
```

I'm not sure how the `localhost:3100` comes about, but this seems to be specific to this branch as it still works perfectly fine on the dev branch.

machisuji:
> @toy
> 
> On this branch when accessing `localhost:4200` - which usually works just fine just the same as `localhost:3000` - I get a proxy error:
> 
> ```
> 11:17:20 angular.1 | <e> [webpack-dev-server] [HPM] Error occurred while proxying request localhost:4200/ to http://localhost:3100/ [ECONNREFUSED] (https://nodejs.org/api/errors.html#errors_common_system_errors)
> ```
> 
> I'm not sure how the `localhost:3100` comes about, but this seems to be specific to this branch as it still works perfectly fine on the dev branch.

I have fixed the proxy issue by amending the .env.example file. It must use the `FE_PORT` in the `OPENPROJECT_DEV_URL`.

I have also amended the `docker-compose.yml` to expose the backend and frontend ports. This is technically unrelated to the PR and a separate bug that we ever removed the exposed ports which made it impossible to use the docker setup locally with http directly.

Either way this fix needed to align with your PR, however, which is why I made it on this branch.

I've tested both the direct and docker setups and they both work as expected for me now. So from my perspective this is good to be merged.

But it would indeed be a good idea to ask someone from the @opf/titan team to double check their docker setup with TLS is still working with these changes.

machisuji (approved):
👍

frontend/cli_to_rails_proxy.js
machisuji: @cbliard @toy In the context of the docker setup `PROXY_HOSTNAME` refers to the container host name, for instance `backend`, for the backend container. So `localhost` wouldn't work here. Hence this should remain and not be replaced with `HOST`.

cbliard: Thanks for the explanation. That makes more sense. 
So when `PROXY_HOSTNAME` is not set, it should default to either `HOST` or `FE_HOST` then?

machisuji: The fallback should be `HOST`.

docker-compose.yml
```diff
@@ -66,6 +67,8 @@ services:
     depends_on:
       - db
       - cache
+    ports:
+      - "${PORT}:3000"
```
toy: @machisuji Shouldn't this fallback:
```suggestion
      - "${PORT:-3000}:3000"
```

docker-compose.yml
```diff
@@ -87,6 +90,8 @@ services:
       - network
     depends_on:
       - backend
+    ports:
+      - "${FE_PORT}:4200"
```
toy: @machisuji This too:
```suggestion
      - "${FE_PORT:-4200}:4200"
```

frontend/cli_to_rails_proxy.js
```diff
@@ -1,9 +1,10 @@
 const PROXY_HOSTNAME = process.env.PROXY_HOSTNAME || 'localhost';
```
toy: @machisuji So it should be:
```suggestion
const PROXY_HOSTNAME = process.env.PROXY_HOSTNAME || process.env.HOST || 'localhost';
```

akabiru:
Looking great - Please hold merging until @Kharonus / @apfohl have a look as they're more familiar with the Titan docker setup.

I couldn't get this PR to work with the Docker TLS setup out of the box with the following env: I'm likely missing sth obvious.

```yml
# Local backend development host and port
HOST=openproject.local
PORT=3000
# Local frontend development host and port
FE_HOST=openproject.local
FE_PORT=4200
```

https://github.com/user-attachments/assets/3b5ae89e-012f-429b-9474-ec5a3c8c0f9f

docker-compose.yml
```diff
@@ -87,6 +90,8 @@ services:
       - network
     depends_on:
       - backend
+    ports:
+      - "${FE_PORT:-4200}:4200"
```
Kharonus: 🔴 I want to raise one issue here.

in the past we had port bindings in the `docker-compose.yml`. We removed them because exposed ports cannot be closed by writing anything in the `docker-compose.override.yml`. Therefore, we considered the practice of not exposing ANY port in the `docker-compose.yml`.

This led to having the `docker-compose.override.example.yml` to contain all port bindings. Hence, I'd remove that port bindings from here.

It is very easy to understand, that a port is not acessible from the host, and then a dev can use the `docker-compose.override.example.yml` to get the "default" port bindings.

Prematurely bound ports is always a hassle, when you want to up stacks and a port is not available.

Kharonus (changes requested):

docker-compose.yml
toy: Can surely be removed, though there is a solution for overriding using yaml tags: [!reset](https://docs.docker.com/reference/compose-file/merge/#reset-value) or [!override](https://docs.docker.com/reference/compose-file/merge/#replace-value)

machisuji: @toy In that case it can be removed, yes. But please update the override example accordingly to use the `FE_PORT` in the frontend section.

Also we should amend the docker development readme to point out that this override file is necessary to be able to access the setup under HTTP://localhost:3000 as documented.

toy: @machisuji Override already has both ports added

docker-compose.yml
```diff
@@ -38,6 +38,7 @@ x-op-backend: &backend
   environment:
     LOCAL_DEV_CHECK: "${LOCAL_DEV_CHECK:?The docker-compose file for OpenProject has moved to https://github.com/opf/openproject-deploy}"
     RAILS_ENV: development
+    OPENPROJECT_HOST__NAME: "${HOST:-localhost}:${PORT:-3000}"
```
toy: @machisuji Should this be moved to override example?

docker-compose.yml
machisuji: Ah right! All good then!

machisuji: That is this line can be removed.

docker-compose.yml
machisuji: Quite possibly. I will give it a go in a bit.

docker-compose.yml
toy: Already [documented](https://www.openproject.org/docs/development/development-environment/docker/#:~:text=You%20also%20will%20want%20to%20create%20a%20docker%2Dcompose.override.yml%20file%2C%20which%20can%20contain%20the%20port%20exposure%20for%20your%20containers)

docker-compose.yml
toy: I moved it, as it feels odd to set environment variable for setting if it supposed to be changed in override file

machisuji (approved):
👍

PR #16737, opened 2024-09-17
[57092] Right side of the instance upper banner looks misaligned

bsatarnejad:
# Ticket
https://community.openproject.org/projects/openproject/work_packages/57092/activity

Align the header buttons.

## Screenshots
Before:
Screenshot 2024-09-17 at 06 45 21

After:

Screenshot 2024-09-17 at 06 44 33

Replace op-icon with octicon for bell icon. And override the width of op-app-header--primer-button to auto, because it has ''Button--iconOnly classwith 2rem width.
# Merge checklist

- [ ] Added/updated tests

### DIFF
--- a/frontend/src/app/features/in-app-notifications/bell/in-app-notification-bell.component.html
+++ b/frontend/src/app/features/in-app-notifications/bell/in-app-notification-bell.component.html
@@ -3,7 +3,7 @@
   data-test-selector="op-ian-bell"
   [href]="notificationsPath()"
 >
-  <op-icon icon-classes="icon-bell"></op-icon>
+  <svg bell-fill-icon  size="small"></svg>
   <ng-container *ngIf="(unreadCount$ | async) as unreadCount">
     <span
       *ngIf="unreadCount !== 0"
--- a/frontend/src/app/features/in-app-notifications/bell/in-app-notification-bell.component.html
+++ b/frontend/src/app/features/in-app-notifications/bell/in-app-notification-bell.component.html
@@ -3,7 +3,7 @@
   data-test-selector="op-ian-bell"
   [href]="notificationsPath()"
 >
-  <svg bell-fill-icon  size="small"></svg>
+  <svg bell-fill-icon size="small"></svg>
   <ng-container *ngIf="(unreadCount$ | async) as unreadCount">
     <span
       *ngIf="unreadCount !== 0"
--- a/frontend/src/app/shared/components/icon/icon.module.ts
+++ b/frontend/src/app/shared/components/icon/icon.module.ts
@@ -29,6 +29,7 @@ import {
   StarFillIconComponent,
   StarIconComponent,
   XIconComponent,
+  BellFillIconComponent,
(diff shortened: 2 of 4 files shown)


mrmir:
Hi @bsatarnejad, thanks for this change. Is it also possible to center the icons vertically? To me is still looks like the avatar is off-centered compared to the other 3 icons. What do you think?

frontend/src/app/features/in-app-notifications/bell/in-app-notification-bell.component.html
```diff
@@ -3,7 +3,7 @@
   data-test-selector="op-ian-bell"
   [href]="notificationsPath()"
 >
-  <op-icon icon-classes="icon-bell"></op-icon>
+  <svg bell-fill-icon  size="small"></svg>
```
toy:
```suggestion
  <svg bell-fill-icon size="small"></svg>
```

toy (commented):
@mrmir is right, top is 11px, bottom is 8px

Could you please also make the avatar defined size, so that the width of avatar doesn't jump from 0 when the avatar gets loaded?

bsatarnejad:
> Could you please also make the avatar defined size, so that the width of avatar doesn't jump from 0 when the avatar gets loaded?

Hi @mrmir yes I agree, in my eyes, avatar and other icons are not aligned, I think it is because they have different sizes, Avatar is larger than other icons. I remove the line-height for op-app-menu--item-action, it fixes the issue you mentioned. Could you please have another look?

bsatarnejad:
> @mrmir is right, top is 11px, bottom is 8px
> 
> Could you please also make the avatar defined size, so that the width of avatar doesn't jump from 0 when the avatar gets loaded?

Hi @toy 
thanks for the feedback. The avatar has width and height of 36px. To prevent the avatar from being squeezed by the flexbox layout, the min-width property was added(min-width=36px). This ensures the avatar retains a minimum size, even when the layout is compressed or constrained.

toy:
It is now vertically centred 👍
Jumping is still there, did you push everything? See the video, reload in the middle of it:

https://github.com/user-attachments/assets/07a4f429-b087-43f3-bde0-6a6b91642d69

mrmir (approved):
The alignment is good now, thanks! Feel free to merge once the jumping issue is sorted out.

toy:
Jump is now just by few pixels due to avatar, maybe better to fix sizes of all icons. I assume the search bar is appearing not at start due to it being created by angular.

https://github.com/user-attachments/assets/425664db-35a5-40fb-901c-d2f0064a5005

bsatarnejad:
> Jump is now just by few pixels due to avatar, maybe better to fix sizes of all icons. I assume the search bar is appearing not at start due to it being created by angular.
> 
>  jump.mov

Hi @toy 
thanks for the feedback, I shared this jumping issue in our design daily, we decide to create a separate WP for it and fix the issue there: https://community.openproject.org/projects/openproject/work_packages/57956

PR #16741, opened 2024-09-17
[#55392] page number should reset when changing the sort order via the configure view modal

EinLama:
# Ticket

https://community.openproject.org/wp/55392

As stated in the ticket, this PR accomplishes that the **project list** pagination is always reset to the first page whenever the sorting is changed. Before this PR, this would only work when the user changes the sorting by clicking on the table header. Now, the pagination is also reset when the sorting is changed by means of the configure view modal.

This PR also includes some minor spelling fixes and doc comment updates in nearby files.

I saw two major ways of tackling this bug:
1. by resetting the pagination in the view layer, i.e. when the sorting is changed, the hidden `page` input field is emptied.
2. by resetting the pagination in the controller (on server side).

Option 2 has two major flaws: For one, it leads to buggy behavior when you click on a link with a different sorting than yours (and a page != 1). Even worse, it will not allow you to open multiple tabs for the project list and apply different sort  settings in them without resetting your current page (thanks for pointing that out @oliverguenther ).

Option one seemed like the better approach. To do so, I added a dedicated Stimulus controller for this purpose. But the resulting markup was rather messy. Moving the logic into the _existing_ sort-by controller had two benefits: it cleaned up the markup while simultaneously allowing us to adopt this bugfix for other paginated, sortable views in the future.

# Merge checklist

- [x] Added/updated tests


oliverguenther (changes requested):
Works great for the use-case of changing the attributes you sort by.

When you change _only_ the sort direction through the dialog, the page is not reset.
The bug description doesn't really specify what to do here, but I reckon it should be reset too

[Kapture 2024-09-20 at 09.47.43.webm](https://github.com/user-attachments/assets/fd7229f2-cb63-4de2-93fc-b8ba90d3d9dc)


/edit Not sure what I did here, but I cannot reproduce this problem

EinLama:
Great catch! Yes, changing asc/desc should also reset the pagination to stay consistent with sorting by table header.

Will fix!

oliverguenther (approved):

PR #16749, opened 2024-09-18
[57812] Changed CustomOptions to become a hierarchy

apfohl:
# Ticket
https://community.openproject.org/projects/openproject/work_packages/57812/activity

## Screenshots

# Merge checklist

- [ ] Added/updated tests

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -174,7 +174,7 @@ jobs:
         with:
           context: git
           labels: |
-            io.artifacthub.package.readme-url=https://raw.githubusercontent.com/opf/openproject/refs/heads/dev/docker/prod/REAMDE.md
+            io.artifacthub.package.readme-url=https://raw.githubusercontent.com/opf/openproject/refs/heads/dev/docker/prod/README.md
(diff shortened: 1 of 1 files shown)


db/migrate/20240917092519_create_custom_option_hierarchies.rb
```diff
@@ -0,0 +1,16 @@
+class CreateCustomOptionHierarchies < ActiveRecord::Migration[7.1]
+  def change
+    create_table :custom_option_hierarchies, id: false do |t|
```
akabiru: 🟠 I would leave the id present- ActiveRecord won't play well with nil primary key. For example, `#destroy` won't work which means you can't leverage things like AR callbacks or the in-house DeleteService which [relies on #destroy out of the box](https://github.com/opf/openproject/pull/16694#discussion_r1758641198).

Of course you have the option to [define composite primary keys](https://guides.rubyonrails.org/active_record_composite_primary_keys.html) but I've recently learnt [_the hard way_](https://github.com/opf/openproject/pull/16694) the change in data structure can be messy

PR #16752, opened 2024-09-18
Chore/upgrade turbo

cbliard:
# Ticket

None

I tried using turbo morph instead of a custom rendering with iodiomorph to update the progress popover. It did not work.

The turbo has been upgraded and a stimulus controller has been merged in the process, which is still valuable, hence this pull request.

The turbo lib have been upgraded to 8.0.10 because there was a bug which prevented morph from working when triggered from javascript. It was fixed in this version.

Also using morph needed some `data-turbo-permanent` on the inputs getting focus so that the value and the focus are kept, which is why the 2 controllers have been merged.

# Merge checklist

- [ ] Added/updated tests


aaron-contreras (approved):
Nicely done inlining the focus controller too

cbliard:
The one failing test comes from `dev` branch and has been fixed in #16757
Merging

PR #16757, opened 2024-09-19
fix[Op#57516]: Reset column info and primary key after migration for fresh state

akabiru:
_Follows #16694_ 

ActiveRecord & FactoryBot definitions would have cached column and primary key info. When recreating the previous state, let's reset.

```
 1) AddPrimaryKeyToCustomFieldsProjects adds an `id` primary key column with backfilled values
     Failure/Error: create_list(:custom_fields_project, 5)

     ActiveRecord::StatementInvalid:
       PG::UndefinedColumn: ERROR:  column "id" does not exist
       LINE 1: ...stom_field_id", "project_id") VALUES ($1, $2) RETURNING "id"
```

https://community.openproject.org/wp/57516


cbliard (approved):
👏

PR #16762, opened 2024-09-19
Bug/57577 broken ordering by multi value custom fields

toy:
# Ticket
[OP#57577](https://community.openproject.org/wp/57577)

Fix inconsistencies of ordering by multi value custom fields (list, user, version). Preparing for [OP#57305](https://community.openproject.org/wp/57305)

# Merge checklist

- [x] Added/updated tests

### DIFF
--- a/spec/models/queries/projects/project_query_custom_field_order_spec.rb
+++ b/spec/models/queries/projects/project_query_custom_field_order_spec.rb
@@ -58,12 +58,18 @@ RSpec.describe ProjectQuery, "order using CustomFieldOrder" do
 
     before { projects }
 
+    project_attributes = ->(project) do
+      {
+        id: project.id,
+        values: project.custom_values.map(&:value).sort
+      }
+    end
+
     context "in ascending order" do
       let(:direction) { :asc }
 
       it "returns the correctly sorted result" do
-        expect(query_results.map(&:id))
-          .to eq projects.map(&:id)
+        expect(query_results).to eq_array(projects, &project_attributes)
       end
     end
 
@@ -71,8 +77,7 @@ RSpec.describe ProjectQuery, "order using CustomFieldOrder" do
       let(:direction) { :desc }
 
       it "returns the correctly sorted result" do
-        expect(query_results.map(&:id))
-          .to eq projects_desc.map(&:id)
+        expect(query_results).to eq_array(projects_desc, &project_attributes)
       end
     end
   end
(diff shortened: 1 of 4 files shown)


spec/models/queries/projects/project_query_custom_field_order_spec.rb
```diff
@@ -58,21 +58,26 @@ def project_without_cf_value
 
     before { projects }
 
+    work_package_attributes = ->(work_package) do
+      {
+        id: work_package.id,
+        values: work_package.custom_values.map(&:value).sort
+      }
+    end
```
ulferts, line 66: I guess this is a C&P mistake. It should be named `project_attributes` with a project passed in.

spec/models/queries/projects/project_query_custom_field_order_spec.rb
```diff
@@ -188,28 +193,39 @@ def project_without_cf_value
...
+            project_with_cf_value(*id_by_value.fetch_values("20", "3", "100")), # 100, 3, 20
+            project_with_cf_value(*id_by_value.fetch_values("20", "100", "3")), # 100, 3, 20
+            project_with_cf_value(*id_by_value.fetch_values("100", "3", "20")), # 100, 3, 20
+            project_with_cf_value(*id_by_value.fetch_values("100", "20", "3")), # 100, 3, 20
+            project_with_cf_value(*id_by_value.fetch_values("20", "100")),      # 100, 20
+            project_with_cf_value(*id_by_value.fetch_values("3")),              # 3
+            project_with_cf_value(*id_by_value.fetch_values("3", "20")),        # 3, 20
+            project_with_cf_value(*id_by_value.fetch_values("20")),             # 20
+            project_without_cf_value # TODO: decide on order of absent values
           ]
```
ulferts, line 209: 👍 for adding the additional cases where the same value is permutated.

ulferts (approved):
Apart from the one tiny naming issue, the changes are good. So merge at will once that is resolved.

The structure of the commits really helped me work through the PR. Thanks for that.

It appears to me that at least part of the performance problems related to sorting by multi value user custom fields are already resolved by this PR. At least on the data I have available, the performance was no longer abhorrent. 

You were right about the need to have the selected values for multi fields ordered in the same way as is now used when ordering work packages and projects. That can be a different topic so please open a ticket for this.

toy:
> The structure of the commits really helped me work through the PR. Thanks for that.

:heart:

> It appears to me that at least part of the performance problems related to sorting by multi value user custom fields are already resolved by this PR. At least on the data I have available, the performance was no longer abhorrent.

Yes it should be ~3 times faster after this PR

> You were right about the need to have the selected values for multi fields ordered in the same way as is now used when ordering work packages and projects. That can be a different topic so please open a ticket for this.

👍

PR #16775, opened 2024-09-23
Feature/57990 preserve percent complete over remaining work when migrating

cbliard:
# Ticket

https://community.openproject.org/wp/57990

In 14.0, when the progress values were made consistent, a data migration was done to transform all progress values of all work packages to make them consistent. To do so there were some general rules: as the % Complete value is always a derived value in 14.0, it makes sense to preserve Work and Remaining work in 14.0, and make % Complete a derived value.

In 14.6, we'll make [% Complete editable again](https://community.openproject.org/projects/openproject/work_packages/52233/activity), and we want to preserve that value more than before. For people not having made the jump to 14.x because they feared losing their % Complete value, we'll update the data migration so that Work and % Complete are preserved, and Remaining work is derived. It better reflect how the application is working now with regards to the progress values.

Update the job to derive % Complete only when it's unset, and alway derive Remaining work from the 2 others as last step.

# Merge checklist

- [x] Added/updated tests

### DIFF
--- a/spec/migrations/update_progress_calculation_spec.rb
+++ b/spec/migrations/update_progress_calculation_spec.rb
@@ -863,14 +863,17 @@ RSpec.describe UpdateProgressCalculation, type: :model do
           .to eql [nil, 50]
       end
 
-      it "creates a journal for the work package transitioning ∑ % complete from 90% to 20%" do
+      it "creates a journal for the work package transitioning ∑ remaining work from 50h to 80h " \
+         "and ∑ % complete from 90% to 20%" do
         expect(wp_wrong.journals.count).to eq(2)
 
         expect(wp_wrong.journals.first.get_changes.keys)
           .not_to include("derived_done_ratio")
 
-        expect(wp_wrong.journals.last.get_changes["derived_done_ratio"])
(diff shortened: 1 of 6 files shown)


spec/migrations/update_progress_calculation_spec.rb
```diff
@@ -858,14 +863,14 @@ def expect_migrates(from:, to:)
           .to eql [nil, 50]
       end
 
-      it "creates a journal for the work package transitioning ∑ % complete from 90 to 50" do
+      it "creates a journal for the work package transitioning ∑ % complete from 90% to 20%" do
```
dombesz: Should this spec also contain an expectation to a journal entry of setting the remaining work to the correct value?

dombesz (approved):
Nice job @cbliard, I just have one question.

spec/migrations/update_progress_calculation_spec.rb
cbliard: This context was testing specifically what happened to total % complete, as it is reset in older journals, and updated in the last one, but it can indeed test for more. 

I added a commit to also test for ∑ remaining work in the same test.

PR #16776, opened 2024-09-23
Bug/57550 custom field with format version are ordered as strings

toy:
# Ticket
[OP#57550](https://community.openproject.org/wp/57550)

Order versions as versions instead of as strings

Using collation that treats runs of digits as one number rather than as separate digits.

Questionable: 

1. Creating collation with `IF NOT EXISTS` to reduce small chance of conflict. Also not dropping it in down direction
2. I'm not sure if keeping `order_by_semver_name` is meaningful
3. It is also now using `order` instead of `reorder`, no complaints from tests about that

# Merge checklist

- [x] Added/updated tests

### DIFF
--- a/app/models/queries/work_packages/selects/property_select.rb
+++ b/app/models/queries/work_packages/selects/property_select.rb
@@ -95,7 +95,7 @@ class Queries::WorkPackages::Selects::PropertySelect < Queries::WorkPackages::Se
     },
     version: {
       association: "version",
-      sortable: [->(table_name = Version.table_name) { Version.semver_sql(table_name) }, "name"],
+      sortable: "name",
       default_order: "ASC",
       null_handling: "NULLS LAST",
       groupable: "#{WorkPackage.table_name}.version_id"
--- a/app/models/versions/scopes/order_by_semver_name.rb
+++ b/app/models/versions/scopes/order_by_semver_name.rb
@@ -32,21 +32,7 @@ module Versions::Scopes
 
     class_methods do
       def order_by_semver_name
-        reorder semver_sql, :name
-      end
-
-      # Returns an sql for ordering which:
-      # * Returns a substring from the beginning of the name up until the first alphabetical character e.g. "1.2.3 "
-      #   from "1.2.3 ABC"
-      # * Replaces all non numerical character groups in that substring by a blank, e.g "1.2.3 " to "1 2 3 "
-      # * Splits the result into an array of individual number parts, e.g. "{1, 2, 3, ''}" from "1 2 3 "
-      # * removes all empty array items, e.g. "{1, 2, 3}" from "{1, 2, 3, ''}"
-      def semver_sql(table_name = Version.table_name)
-        sql = <<~SQL
-          array_remove(regexp_split_to_array(regexp_replace(substring(#{table_name}.name from '^[^a-zA-Z]+'), '\\D+', ' ', 'g'), ' '), '')::int[]
-        SQL
-
-        Arel.sql(sql)
+        order :name
       end
     end
   end
--- a/db/migrate/20240920152544_set_versions_name_collation.rb
+++ b/db/migrate/20240920152544_set_versions_name_collation.rb
@@ -1,0 +1,13 @@
+class SetVersionsNameCollation < ActiveRecord::Migration[7.1]
+  def up
+    execute <<-SQL.squish
+      CREATE COLLATION IF NOT EXISTS versions_name (provider = icu, locale = 'en-u-kn-true');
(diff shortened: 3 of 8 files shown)


ulferts (approved):
Nice change here @toy. And good find in the first place of using the collation.

The reason why I did not merge this now is that I have a recommendation I would like to talk through with you.

Currently there is a discrepancy between ordering a work package list by a version custom field as opposed to ordering it by the hard wired version property. For the later, the null values are always placed last. I propose to remove this in the same PR, or at least to do it right after. It would only require to remove the `null_handling` statement defined for version in `property_select.rb`.

The other thing I would include in this PR is the cleanup of the `Version.sort_by_name` scope which I believe has no more merit. 

Lastly, it is a bit weird that there is now a `name` and a `semver_name` order available in the API which basically do the same but I would still keep the two to not break potential integrations.

toy:
> Currently there is a discrepancy between ordering a work package list by a version custom field as opposed to ordering it by the hard wired version property. For the later, the null values are always placed last. I propose to remove this in the same PR, or at least to do it right after. It would only require to remove the `null_handling` statement defined for version in `property_select.rb`.

[Related discussion](https://community.openproject.org/projects/openproject/work_packages/57554/activity) about ordering of absent values for custom fields. What about start date and due date columns? The PR #8000 and [OP#32156](https://community.openproject.org/projects/openproject/work_packages/32156) don't describe the reason for adding null_handling that puts empty values always at the end for 3 properties.

> The other thing I would include in this PR is the cleanup of the `Version.sort_by_name` scope which I believe has no more merit.

You mean `Version.order_by_name` (or `Version.order_by_semver_name`)?

Just saw that `Version.order_by_name` orders by lowered `name` which according to few tests preserves collation, so still orders versions not as plain strings. We can change the collation to also be not case sensitive, but it will make it [non deterministic](https://www.postgresql.org/docs/current/collation.html#COLLATION-NONDETERMINISTIC)

> Lastly, it is a bit weird that there is now a `name` and a `semver_name` order available in the API which basically do the same but I would still keep the two to not break potential integrations.

I'm wondering how does Queries::Versions::Orders::NameOrder work, if it [overrides private `order` to receive no arguments](https://github.com/opf/openproject/blob/dev/app/models/queries/versions/orders/name_order.rb#L38) while [in base it receives one](https://github.com/opf/openproject/blob/dev/app/models/queries/orders/base.rb#L71), so `apply_to` will raise

ulferts:
> [Related discussion](https://community.openproject.org/projects/openproject/work_packages/57554/activity) about ordering of absent values for custom fields. What about start date and due date columns? The PR https://github.com/opf/openproject/pull/8000 and [OP#32156](https://community.openproject.org/projects/openproject/work_packages/32156) don't describe the reason for adding null_handling that puts empty values always at the end for 3 properties.

I'd just remove it for versions so that it the same whenever one orders by version name, regardless of whether that is done on a custom field or on the hard wired property. But it is also not super important so if you want it done in the scope of [the null value order feature](https://community.openproject.org/wp/57554) that would also be fine with me. And apparently there is also the same discrepancy between date custom fields and due and start date so I definitely see the benefit in having all of them harmonized.

> You mean Version.order_by_name (or Version.order_by_semver_name)?

Yes, I mean the `order_by_name` method. Sorry for the confusion.

>  Just saw that Version.order_by_name orders by lowered name which according to few tests preserves collation, so still orders versions not as plain strings. We can change the collation to also be not case sensitive, but it will make it [non deterministic](https://www.postgresql.org/docs/current/collation.html#COLLATION-NONDETERMINISTIC)

At least in my test, the change in the collation already seems to do that even without the transformation to LOWER. I did not check why:

image

> I'm wondering how does Queries::Versions::Orders::NameOrder work, if it [overrides private order to receive no arguments](https://github.com/opf/openproject/blob/dev/app/models/queries/versions/orders/name_order.rb#L38) while [in base it receives one](https://github.com/opf/openproject/blob/dev/app/models/queries/orders/base.rb#L71), so apply_to will raise

It fails of course :). Apparently it is just broken. Could be fixed in a separate PR but since the change of simply removing the method override is simple, I'd include it.

toy:
> At least in my test, the change in the collation already seems to do that even without the transformation to LOWER. I did not check why:

Just to share here the learning that it is caused by using ICU collation: it is not ignoring the case, but ICU collation is more sophisticated than just comparing character codes while applying some language rules - it compares characters on multiple levels and case difference is less important than case-ignoring difference. So following is the correct collation: `a 0`, `A 0`, `à 0`, `a 1`, `A 1`, `à 1`

PR #16777, opened 2024-09-23
[#57928] user cant be removed from global role

EinLama:
# Ticket
https://community.openproject.org/wp/57928

Global permissions can now be removed from users by an admin.

## Screenshots

The `DeleteService` makes a distinction between a project member and a "global" user. Based on the result, the correct contract for the use case is used to check the permissions for the operation.

# Merge checklist

- [x] Added/updated tests

### DIFF
--- a/app/contracts/members/global_delete_contract.rb
+++ b/app/contracts/members/global_delete_contract.rb
@@ -27,7 +27,7 @@
 #++
 
 module Members
-  class DeleteContract < DeleteBaseContract
-    delete_permission :manage_members
+  class GlobalDeleteContract < DeleteBaseContract
+    delete_permission :admin
   end
 end
--- a/app/services/members/delete_service.rb
+++ b/app/services/members/delete_service.rb
@@ -63,7 +63,7 @@ class Members::DeleteService < BaseServices::Delete
 
   def default_contract_class
     # We have different contracts for project roles and global roles
-    if model.project.present?
+    if model.project_role?
       "#{namespace}::DeleteContract".constantize
     else
       "#{namespace}::GlobalDeleteContract".constantize
--- a/app/services/members/delete_service.rb
+++ b/app/services/members/delete_service.rb
@@ -60,4 +60,13 @@ class Members::DeleteService < BaseServices::Delete
       .new(member.principal, current_user: user, contract_class: EmptyContract)
       .call
   end
+
+  def default_contract_class
+    # We have different contracts for project roles and global roles
+    if model.project.present?
+      "#{namespace}::DeleteContract".constantize
+    else
+      "#{namespace}::GlobalDeleteContract".constantize
+    end
+  end
 end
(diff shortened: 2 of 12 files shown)


app/contracts/members/delete_contract.rb
```diff
@@ -27,15 +27,7 @@
 #++
 
 module Members
-  class DeleteContract < ::DeleteContract
+  class DeleteContract < DeleteBaseContract
```
ulferts: ❓ what is the reason for not having this names something project specific. That way it would be more easily discernible from the `GlobalDeleteContract`

app/contracts/members/global_delete_contract.rb
```diff
@@ -0,0 +1,33 @@
...
+# along with this program; if not, write to the Free Software
+# Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+#
+# See COPYRIGHT and LICENSE files for more details.
+#++
+
+module Members
+  class GlobalDeleteContract < DeleteBaseContract
+    delete_permission :admin
+  end
```
ulferts: This contract I would like to see unit speced, similar to how it is [done for the `Members::DeleteContract`](https://github.com/opf/openproject/blob/dev/spec/contracts/members/delete_contract_spec.rb). I feel that specing the contracts is very important since they are the safeguard against faulty behaviour.

app/services/members/delete_service.rb
```diff
@@ -60,4 +60,13 @@ def cleanup_for_group(member)
       .new(member.principal, current_user: user, contract_class: EmptyContract)
       .call
   end
+
+  def default_contract_class
+    # We have different contracts for project roles and global roles
+    if model.project.present?
```
ulferts: There would also be the `#project_role?` helper method. Don't know if that it really better.

app/services/members/delete_service.rb
```diff
@@ -60,4 +60,13 @@ def cleanup_for_group(member)
...
+      "#{namespace}::DeleteContract".constantize
+    else
+      "#{namespace}::GlobalDeleteContract".constantize
+    end
```
ulferts, line 70: A [spec for this](https://github.com/opf/openproject/blob/dev/spec/contracts/members/delete_contract_spec.rb) would ensure that the behaviour is safe against regression.

ulferts (approved):
The functionality and the code change is good. 

For me, a couple of specs are added to guard the code against regression. That can however also happen in a subsequent PR, so that the change here can be QAed and shipped in the upcomming bug fix release.

app/services/members/delete_service.rb
EinLama: Seems like a good fit here, added!

app/services/members/delete_service.rb
EinLama: added with 2e4b0eb70f30

app/contracts/members/delete_contract.rb
EinLama: I figured it would be easier to merge bugfixes back into `dev` later if I don't refactor class names that are already in use throughout the application.

However, I have now pushed a commit where the class names are normalized and I think it's much more readable, just like you suggested.

ulferts:
@EinLama, nice all good and clean now. Merge at will once the CI agrees.

PR #16778, opened 2024-09-23
use format_date to format timestamp instead of parsing date from stringified timestamp

toy:
Replace odd code

Use `format_date` helper on timestamp, which handles converting to date and applying same format as everywhere else

# Merge checklist

- [ ] Added/updated tests


ulferts (approved):
👍 for cleaning up.

PR #16788, opened 2024-09-24
[57993] Browser tab is missing breadcrumb

bsatarnejad:
# Ticket
https://community.openproject.org/projects/openproject/work_packages/57993/activity

Show selected tab in the html title.

## Screenshots
Before:
Screenshot 2024-09-24 at 11 30 50

After:
Screenshot 2024-09-24 at 11 30 27

### DIFF
--- a/app/components/admin/design_header_component.html.erb
+++ b/app/components/admin/design_header_component.html.erb
@@ -27,7 +27,7 @@ See COPYRIGHT and LICENSE files for more details.
 
 ++#%>
 
-<% helpers.html_title t(:label_administration), t(:label_custom_style), @title %>
+<% helpers.html_title t(:label_administration), t(:label_custom_style), selected_tab(@tabs)[:label] %>
 <%=
   render(Primer::OpenProject::PageHeader.new) do |header|
     header.with_title { t(:label_custom_style) }


app/components/admin/design_header_component.html.erb
```diff
@@ -27,7 +27,7 @@ See COPYRIGHT and LICENSE files for more details.
 
 ++#%>
 
-<% helpers.html_title t(:label_administration), t(:label_custom_style), @title %>
+<% helpers.html_title t(:label_administration), t(:label_custom_style), selected_tab(@tabs)[:label] %>
```
EinLama, line 30: No issue, but an observation: when `selected_tab` cannot find the currently selected tab, it will [return the first entry of the given Array as fallback](https://github.com/opf/openproject/blob/488e7ede14d4967fe74184a289113e1087832604/app/helpers/tabs_helper.rb#L41). In other words: `@tabs.first` is the fallback if no currently selected tab can be found.

Furthermore, if `@tabs` _was_ an empty Array, `@tabs.first` would evaluate to `nil`. And `nil[:label]` would raise an Error.

Luckily `@tabs` is hard-coded to be a list with three entries (see [CustomStylesHelper#design_tabs](https://github.com/opf/openproject/blob/488e7ede14d4967fe74184a289113e1087832604/app/helpers/custom_styles_helper.rb#L30). It is very unlikely that we remove the tabs in the future and keep the selected_tabs helper here at the same time. So we are fine without adding hard-to-read guards here 😀

EinLama (approved):
LGTM 👍🏻

app/components/admin/design_header_component.html.erb
bsatarnejad: Hi @EinLama thanks for explaining 👍

PR #16792, opened 2024-09-24
[#58010] fixed input data for upload query

Kharonus:
# Ticket
[OP#58010](https://community.openproject.org/work_packages/58010)

- changed input data for file upload to be blank id, and not parent folder object
- the id as a value is very specific, while the parent folder object was merely used as a wrapping object for safe handling and validation. validation alas is now done in the contract, so we can use the same approach as in `set_permission` input data


akabiru (approved):
Code LGTM 👍🏾

PR #16793, opened 2024-09-24
Bug[Op#57989]: unlinking a project within a multi-activation page does not retain the current page

akabiru:
#### Ticket

https://community.openproject.org/wp/57989

#### What are you trying to accomplish?

Unlinking a project from a custom field, project attribute or storage (aka Multi-project activation) should retain the current page. Right now, after removal the project list re-renders on the first page which is not desired behaviour.

Further, clicking on the pagination links did not update the browser URL with `?page` param

##### Screenshots

_Before_

https://github.com/user-attachments/assets/9245cac3-dc35-45dc-937b-eb1c0c4940a1

_After_

### Custom Fields

https://github.com/user-attachments/assets/eae3f051-a5d6-40e5-ad4b-11b81ea0420d

### Project Attributes

https://github.com/user-attachments/assets/148a533a-72b8-44b2-82e3-c4c79f701cf4

### File Storages

https://github.com/user-attachments/assets/ca067ad2-6332-47a7-8f37-a911c50459d0

#### What approach did you choose and why?

- [x] Specify [turbo application visit to "advance" ](https://turbo.hotwired.dev/handbook/frames#promoting-a-frame-navigation-to-a-page-visit)within the project list turbo frame such that clicking pagination links (ex. <url>?page=1) also updates the history state, and the URL updates accordingly. This allows for any further actions to retain the current page.

    _*Navigating within a frame can be configured as either [an application visit ("advance" or "replace") or as a _restoration_ visit](https://turbo.hotwired.dev/handbook/drive#page-navigation-basics)_

- [x] Set the `page` param in the destroy links so that responses render in the current page

Set the turbo action to "advance" within the project list turbo frame such that clicking pagination links (ex. <url>?page=1) also updates the history state, and the URL updates accordingly. This allows for any further actions to retain the current page.

#### Merge checklist

- [x] Added/updated tests

### DIFF
--- a/app/helpers/pagination_helper.rb
+++ b/app/helpers/pagination_helper.rb
@@ -215,6 +215,12 @@ module PaginationHelper
     #  "advance" - push a new entry onto the history stack.
     #  "replace" - replace the current history entry.
     # See: https://turbo.hotwired.dev/reference/attributes
+    #
+    # Example: Promoting a Frame Navigation to a Page Visit
+    #   By default navigation within a turbo frame does not change the rest of the browser's state,
+    #   but you can promote a frame navigation a "Visit" by setting the turbo-action attribute to "advance".
+    #   See: https://turbo.hotwired.dev/handbook/frames#promoting-a-frame-navigation-to-a-page-visit
+    #
     def turbo_action
       @options[:turbo_action] # rubocop:disable Rails/HelperInstanceVariable
     end
--- a/app/helpers/pagination_helper.rb
+++ b/app/helpers/pagination_helper.rb
@@ -202,12 +202,21 @@ module PaginationHelper
     def link(text, target, attributes)
       new_attributes = attributes.dup
       new_attributes["data-turbo-stream"] = true if turbo?
+      new_attributes["data-turbo-action"] = turbo_action if turbo_action.present?
 
       super(text, target, new_attributes)
     end
 
     def allowed_params
       @options[:allowed_params] # rubocop:disable Rails/HelperInstanceVariable
+    end
+
+    # Customize the Turbo visit action for pagination links. Can be set to "advance" or "replace".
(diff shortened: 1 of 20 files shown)


app/helpers/pagination_helper.rb
```diff
@@ -210,6 +211,14 @@ def allowed_params
       @options[:allowed_params] # rubocop:disable Rails/HelperInstanceVariable
     end
 
+    # Customize the Turbo visit action for pagination links. Can be set to "advance" or "replace".
+    #  "advance" - push a new entry onto the history stack.
```
Kharonus, line 215: ❓ is one of those the default action? Or is there a third option, how turbo behaves (when the `data-turbo-action` is unset)?

If so, I'd miss that in the comment to get the understanding.

Kharonus (approved):
Looks pretty solid to me and the outcome is working.

I only have one question for the "documentation" of the turbo related behavior in code.

app/helpers/pagination_helper.rb
akabiru: Good eye- within a Turbo Frame navigation is "captured"- only the frame is changed without changing other parts of the application i.e. the browser history is also not affected by default. Hence Turbo, allows [_promoting a Frame Navigation to a Visit_](https://turbo.hotwired.dev/handbook/frames#promoting-a-frame-navigation-to-a-page-visit)

Outside of Turbo Frames- [by default Turbo drive sets the visit type to "advance"](https://turbo.hotwired.dev/handbook/drive#application-visits)

I agree it's rather less clear, and it did in fact take me a while to find the docs for this behaviour. I'll attempt to explain better in the comment

akabiru: https://github.com/opf/openproject/pull/16793/commits/998eb22dd3ee3b5db016c00c0c6a317272558925

PR #16803, opened 2024-09-25
[#58018] no group folder content validation if ampf is disabled

Kharonus:
# Ticket
[OP#58018](https://community.openproject.org/work_packages/58018)

- group folder existence and content should not get validated, if ampf is not enabled
- if userless authentication fails, a specific error should return this as a validation result


modules/storages/spec/common/storages/peripherals/nextcloud_connection_validator_spec.rb
```diff
@@ -47,38 +47,18 @@
...
-    let(:storage) { create(:nextcloud_storage_configured, :as_automatically_managed) }
+  context "if storage is not configured for automatic folder management" do
+    let(:storage) { create(:nextcloud_storage_configured, :as_not_automatically_managed) }
     let(:app_enabled) { true }
     let(:app_version) { Storages::SemanticVersion.parse("2.6.3") }
-    let(:group_folder_enabled) { true }
-    let(:group_folder_version) { Storages::SemanticVersion.parse("17.0.1") }
-    let(:project_folder_id) { "1337" }
+    # let(:group_folder_enabled) { true }
+    # let(:group_folder_version) { Storages::SemanticVersion.parse("17.0.1") }
```
apfohl: Maybe remove?

apfohl (changes requested):
Just a small cleanup

modules/storages/spec/common/storages/peripherals/nextcloud_connection_validator_spec.rb
Kharonus: true :D missed it.

apfohl (approved):

PR #16805, opened 2024-09-25
[57812] Added model and migration for custom field of type hierarchy

apfohl:
# Ticket
https://community.openproject.org/projects/openproject/work_packages/57812/activity

This PR implements a new data structure for storing hierarchical item for use in the new custom field of type hierarchy. Therefore every custom field of type `hierarchy` now holds a reference to a ´root´-item. From there the hierarchy starts.

I used the ´closure_tree´ gem as discussed in the architecture decision.

# Merge checklist

- [ ] Added/updated tests


app/models/custom_value/hierarchy_strategy.rb
apfohl, line 1: 🟡 I'm not sure about this change, if it already goes to deep into frontend.

config/initializers/custom_field_format.rb
apfohl, line 1: 🟡 I'm not sure about this change, if it already goes to deep into frontend.

lib/api/v3/utilities/custom_field_injector.rb
apfohl, line 1: 🟡 I'm not sure about this change, if it already goes to deep into frontend.

app/models/custom_field/Hierarchy/item.rb
Kharonus, line 1: 🟡 you might want to rename the directory to `snake_case`, so in this time lowering the case of the first letter ;)

app/models/custom_value/hierarchy_strategy.rb
Kharonus: I'm not sure, what I read here at all. 😁 Where an when is this strategy called?

config/initializers/custom_field_format.rb
Kharonus: 🔴 I'd either not commit that or already put that behind a feature flag, as we do not want to show hierarchy in the creation dropdown, until we are done with it.

I'd tend to adding the feature flag directly. If you want, I can commit that to your branch

lib/api/v3/utilities/custom_field_injector.rb
Kharonus: 🟢  no harm done here, so fine for me.

Kharonus (changes requested):

config/initializers/custom_field_format.rb
Kharonus: solved

Kharonus: solved

Kharonus (approved):
LGTM

PR #16809, opened 2024-09-25
[#57904] meeting reload button will take you to your previous spot on the page

EinLama:
# Ticket
https://community.openproject.org/wp/57904

Whenever another user makes changes to a meeting, you will get a flash notification that encourages you to reload in order to see the updated contents of the page.

Until now, your viewport was moved back all the way up to the beginning of the page whenever you clicked on that reload button.

With this feature, clicking the reload button will remember your current position on the page and move you back down again.

Caveat: if you do not use the reload button, but trigger a reload manually, your position will be discarded.

## Screenshots
https://github.com/user-attachments/assets/14a83810-0e59-4329-8296-6ec55ad17db5

Introduced a new boolean value that can be set via Stimulus to enable this feature. If done so, the pollForChangesController will track the last known scroll position before reloading the page.

Restoring the last known position only happens when the reload button is used. This is done so that the last known position on the page is NOT restored when you re-enter a meeting page that you've visited previously.

The scroll position is remembered in the `sessionStorage`. This means it is exclusive to the current tab. Additionally, the current path with the meeting id is used as a key. This avoids positional changes on the wrong page when visiting multiple meeting pages after another.

# Merge checklist

- [ ] Added/updated tests

### DIFF
--- a/app/views/layouts/base.html.erb
+++ b/app/views/layouts/base.html.erb
@@ -116,7 +116,7 @@ See COPYRIGHT and LICENSE files for more details.
       </div>
     <% end %>
     <div class="content-overlay"></div>
-    <main id="content-wrapper" class="<%= initial_classes %>">
+    <%= content_tag :main, id: "content-wrapper", class: initial_classes, data: stimulus_content_data do %>
       <%# Primerized flash messages are being rendered separately  %>
       <div id="primerized-flash-messages" class="op-primer-flash">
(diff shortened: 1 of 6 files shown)


EinLama:
Opinions please: should this feature have a feature spec? Since those are rather costly, and this functionality is mostly "nice to have", I deemed it not worth it in this case.

oliverguenther (approved):

PR #16818, opened 2024-09-26
add instructions for adding self-signed certificate to docker install…

tiroessler:
Adding instructions for adding self-signed certificate to a docker installation


oliverguenther (approved):
Nice, thanks 👍

PR #16829, opened 2024-09-27
[56343] Move Colours page into a tab of Admin/Design page

bsatarnejad:
# Ticket
https://community.openproject.org/projects/openproject/work_packages/56343/activity

We want to remove admin/colors page and move its content to a separate tab in admin/design page.

## Screenshots
Screenshot 2026-07-10 at 09 00 13


app/components/colors/edit_page_header_component.rb
```diff
@@ -41,7 +41,7 @@ def initialize(color:)
     def breadcrumb_items
       [
         { href: admin_index_path, text: t(:label_administration) },
-        { href: colors_path, text: t(:label_color_plural) },
+        { href: custom_style_path(tab: :default_colors), text: t(:label_custom_style) },
```
HDinger, line 44: I think, the edit page should show the tab you are on. Currently it says "Administration / Desgin / **Orange**". My proposal would be "Administration / Desgin / Default colors: **Orange**"

app/views/colors/new.html.erb
```diff
@@ -34,7 +34,7 @@ See COPYRIGHT and LICENSE files for more details.
     header.with_title { t("colors.new.label_new_color") }
     header.with_breadcrumbs(
       [{ href: admin_index_path, text: t(:label_administration) },
-       { href: colors_path, text: t(:label_color_plural) },
+       { href: custom_style_path(tab: :default_colors), text: t(:label_custom_style) },
```
HDinger, line 37: Same as before.. I think it should match the actual structur: "Administration / Design / Default colors: **New color**"

app/views/custom_styles/_default_colors.html.erb
```diff
@@ -47,23 +35,35 @@ See COPYRIGHT and LICENSE files for more details.
...
+      <%= render(Primer::Beta::Text.new(display: :flex, align_items: :center, font_size: :small)) do %>
+        <%= render(
+              Primer::Box.new(
+                tag: :span,
+                classes: "color--default-color-swatch",
+                aria: { hidden: true },
+                style: "background-color: #{color.hexcode}"
+              )
+            ) %>
+        <%= color.hexcode %>
```
HDinger: I think this mighte be worth extracting it into it's own component

app/views/custom_styles/show.html.erb
```diff
@@ -28,33 +28,34 @@ See COPYRIGHT and LICENSE files for more details.
...
-  <%= form_tag update_design_themes_path, method: :post, class: "form" do %>
-    <section class="form--section">
-      <fieldset class="form--fieldset">
-        <legend class="form--fieldset-legend"><%= I18n.t("admin.custom_styles.color_theme") %></legend>
-        <div class="form--field">
-          <div class="form--field-container">
-            <%= styled_select_tag "theme",
-                                  options_for_select(@theme_options, @current_theme),
-                                  container_class: "-slim" %>
+  <% if theme_tab? %>
```
HDinger: The name is not really well chosen.. I'd prefer a more self-explaining name like `show_theme_selector?`

frontend/src/global_styles/layout/_colors.sass
```diff
@@ -86,6 +86,14 @@
   a.-dark
     color: white
 
+.color--default-color-swatch
+  border: 1px solid var(--borderColor-default)
+  border-radius: 4px
+  display: inline-block
+  height: 12px
+  margin-right: 8px
+  width: 12px
```
HDinger: Some of those can be primer variables

app/controllers/colors_controller.rb
```diff
@@ -34,19 +34,16 @@ class ColorsController < ApplicationController
...
-  menu_item :colors
+  menu_item :custom_style
 
   def index
-    @colors = Color.all.sort_by(&:name)
     respond_to do |format|
-      format.html
+      format.html { redirect_to colors_settings_path }
     end
   end
```
HDinger: I think we don't need this method anymore, do we?

app/views/colors/confirm_destroy.html.erb
HDinger, line 1: I am nearly 100% sure that this file (and it's route and tests) are not used/needed any more

spec/controllers/colors_controller_spec.rb
```diff
@@ -41,7 +41,16 @@
     def fetch
       get "index"
     end
+
+    def expect_redirect_to
+      custom_style_path(tab: :default_colors)
+    end
+
```
HDinger: Can be removed once the index method is removed

spec/features/custom_styles/default_colors_administration_spec.rb
HDinger, line 1: It probably makes sense to move that to the custom_styles folder

HDinger (changes requested):
Looks good overall. I think there is some more cleanup that can be done..

One other thing: I know that it was not explicitly named in the ticket, but it would be nice to also primerize the edit and create form for colors given that they are rather small.

app/controllers/colors_controller.rb
bsatarnejad: Removed

app/forms/admin/settings/color_form.rb
```diff
@@ -0,0 +1,92 @@
...
+          input_width: :medium
+        )
+
+        color_form.group(layout: :horizontal, classes: "color--hexcode-form-row") do |hexcode_group|
+          hexcode_group.text_field(
+            name: :hexcode,
+            label: attribute_name(:hexcode),
+            required: true,
+            autocomplete: "off",
+            validation_message: validation_message_for(:hexcode),
```
HDinger: Why do you add the error message by hand here? If the error is attached to the model, it is shown automatically.

app/components/colors/default_color_swatch_component.rb
```diff
@@ -0,0 +1,61 @@
...
+#
+# You should have received a copy of the GNU General Public License
+# along with this program; if not, write to the Free Software
+# Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+#
+# See COPYRIGHT and LICENSE files for more details.
+# ++
+
+module Colors
+  class DefaultColorSwatchComponent < ApplicationComponent
```
HDinger: I would like to see a preview for this

frontend/src/global_styles/layout/_colors.sass
```diff
@@ -88,12 +88,21 @@
...
 .color--default-color-swatch
   border: 1px solid var(--borderColor-default)
-  border-radius: 4px
+  border-radius: var(--borderRadius-small)
   display: inline-block
-  height: 12px
-  margin-right: 8px
-  width: 12px
+  min-height: var(--base-size-12)
+  min-width: var(--base-size-12)
```
HDinger: This belongs to the component and thus should be in a default_color_swatch.component.sass

app/components/colors/default_color_swatch_component.rb
HDinger: Imho, the name is a bit off.. The component is generic and can be used outside the "default colors" tab. So including the name `default` in here feels confusing.. I'd suggest to just name it `Colors::SwatchComponent`

app/components/colors/default_color_swatch_component.rb
```diff
@@ -0,0 +1,61 @@
...
+    def initialize(color:, **system_arguments)
+      super()
+
+      @color = color
```
HDinger: The only thing you need from this color is the hexcode, so maybe it is easier to pass that directly.

app/components/colors/swatch_component.rb
```diff
@@ -0,0 +1,61 @@
...
+# See COPYRIGHT and LICENSE files for more details.
+# ++
+
+module Colors
+  class DefaultColorSwatchComponent < ApplicationComponent
+    def initialize(color:, **system_arguments)
+      super()
+
+      @color = color
+      @system_arguments = system_arguments
```
HDinger, line 38: What do you think about adding an option like `show_value` which shows the hexcode as text next to the swatch when being set to `true`. I guess, this is a use case that will appear quite often for this componentn.

frontend/src/global_styles/layout/_colors.sass
```diff
@@ -88,12 +88,21 @@
...
+.color--default-color-list-swatch
+  margin-right: var(--base-size-8)
+
+.color--hexcode-form-row
+  align-items: flex-end
+  width: fit-content
+
 
+.color--hexcode-form-swatch
+  margin-bottom: 10px
```
HDinger: It took me a while to realize that those styles actually belong together. Part of the reason is that the names are again not self explaining. Further, it would have helped if the rules were nested. Since this is specific for that form, I'd also vote to move it to a color_form.sass

```suggestion
.color-form--hexcode-row
  align-items: flex-end
  width: fit-content
  
  // Needed for proper alignment of the swatch next to the input. Due to the label being part of the input, we need to do that manually. 
  .color-form--hexcode-row-swatch
    margin-bottom: 10px
```

HDinger (changes requested):

app/forms/admin/settings/color_form.rb
bsatarnejad: Removed 👍🏼

lookbook/previews/colors/swatch_component_preview.rb
```diff
@@ -0,0 +1,44 @@
...
+#
+# You should have received a copy of the GNU General Public License
+# along with this program; if not, write to the Free Software
+# Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+#
+# See COPYRIGHT and LICENSE files for more details.
+#++
+
+module Colors
+  # @logical_path Colors
```
HDinger:
```suggestion
  # @logical_path OpenProject
```

For it to appear under the OpenProject folder. Otherwise it would look left alone

Image

app/components/colors/swatch_component.rb
```diff
@@ -29,32 +29,49 @@
...
       @system_arguments = system_arguments
     end
 
     def call
+      return swatch unless show_value?
+
+      render(Primer::Box.new(tag: :span, classes: "color--swatch-with-value")) do
+        safe_join([swatch, value])
+      end
+    end
```
HDinger: This looks way too complicated

```rb
def initialize
   @system_arguments[:classes] = class_names(
        @system_arguments[:classes],
        "op-color-swatch"
      )
end

def call
  flex_layout(**system_arguements) do |flex|
    flex.with_colum do
      render(Primer::Box.new(**swatch_arguments))
      end
    if show_value?
      flex.with_colum do
        render(Primer::Beta::Text.new(tag: :span)) { hexcode }
      end
    end
  end
end

def swatch_arguements
  {  
     tag: span,
     aria: @system_arguments.fetch(:aria, {}).merge(hidden: true),
    style: "background-color: #{hexcode}"
  }
end
```

```sass
.op-color-swatch
  border: 1px solid var(--borderColor-default)
  border-radius: var(--borderRadius-small)
  display: inline-flex
  align-items: center
  gap: var(--base-size-8)
  min-height: var(--base-size-12)
  min-width: var(--base-size-12)
```

Then you can get rid of all the helper rendering methods which is somehow bad style anyway. Further, this clearly separates the system arguments from the swatch arguments, and shortens the CSS

HDinger (changes requested):

lookbook/previews/colors/swatch_component_preview.rb
bsatarnejad: My bad :(

app/components/colors/swatch_component.rb
```diff
@@ -39,37 +39,31 @@ def initialize(hexcode:, show_value: false, **system_arguments)
...
-      render(Primer::Box.new(tag: :span, classes: "color--swatch-with-value")) do
-        safe_join([swatch, value])
+      render(Primer::OpenProject::FlexLayout.new(classes: "op-color-swatch--with-value", align_items: :center)) do |flex|
+        flex.with_column do
+          render(Primer::Box.new(**swatch_arguments))
+        end
+
+        flex.with_column do
+          render(Primer::Beta::Text.new(tag: :span)) { hexcode }
+        end
```
HDinger: Although, this is already better, it is not what I had in mind.. My idea was to always render the flex around and thus extract the swatch_arguments from the system_arguments. If needed, the swatch_arguments can be another named param but currently you are mixing that heavily. 

Let me try to explain in an example: 
* If I write `Color::SwatchComponent.new(hexcode: "#fff", show_value: false, mr: 2)`, the margin will be added at the end of the component (e.g. when I want to show something after it)
* f I write `Color::SwatchComponent.new(hexcode: "#fff", show_value: true, mr: 2)`, the margin will be added between the swatch and the text. I have no choice to add a margin right at the end of the component itself.

This is why, my [original proposal](https://github.com/opf/openproject/pull/16829#discussion_r3577521330) was to add a flex wrapper which contains the system_arguments (and is always rendered). The text column is the one that is rendered conditionally, which also makes more sense from a component design point of view. The swatch_arguments are independent of the system_arguments are always applied on the swatch as the name implies.

HDinger (changes requested):

app/components/colors/swatch_component.rb
```diff
@@ -61,12 +67,12 @@ def show_value?
...
 
     def swatch_arguments
-      @system_arguments.merge(
+      {
         tag: :span,
-        classes: helpers.class_names("op-color-swatch", @system_arguments[:classes]),
-        aria: @system_arguments.fetch(:aria, {}).merge(hidden: true),
-        style: [@system_arguments[:style], "background-color: #{hexcode}"].compact.join("; ")
-      )
+        classes: "op-color-swatch--indicator",
```
bsatarnejad, line 72: I am not really sure about the class name, WDYT?

HDinger: I don't have a better name either, so let's stick with that 👍

HDinger (approved):
Awesome, thanks a lot 👍 🙇 Please squash the commits when merging

PR #16841, opened 2024-09-27
Bug/57305 sorting by custom field has strong impact on performance for the project list

toy:
# Ticket
[OP#57305](https://community.openproject.org/wp/57305) + [OP#57554](https://community.openproject.org/wp/57554)

## Screenshots

# Merge checklist

- [ ] Added/updated tests

### DIFF
--- a/app/models/custom_field/order_statements.rb
+++ b/app/models/custom_field/order_statements.rb
@@ -75,6 +75,8 @@ module CustomField::OrderStatements
   # Returns the join statement that is required to group objects by their value
   # of the custom field.
   def group_by_join_statement
+    return join_for_group_by_list_sql if field_format == "list"
+
     order_join_statement
   end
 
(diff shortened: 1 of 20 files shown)


app/models/queries/work_packages/selects/custom_field_select.rb
```diff
@@ -105,9 +77,22 @@ def summable_scope(query)
...
+
+      ->(query, grouped) {
+        Queries::WorkPackages::Selects::WorkPackageSelect
+          .scoped_column_sum(summable_scope(query), select, grouped:, query:)
+      }
     else
-      "COALESCE(ROUND(SUM(value::NUMERIC), 2)::FLOAT, 0.0) #{name}"
+      false
     end
   end
```
toy, line 98: Diff may make it confusing, this is extracted from `set_summable!`

app/models/custom_field/order_statements.rb
```diff
@@ -27,152 +27,144 @@
...
-  def order_by_list_sql
-    columns = multi_value? ? "array_agg(co_sort.position ORDER BY co_sort.position)" : "co_sort.position"
-    limit = multi_value? ? "" : "LIMIT 1"
+  private
 
+  def can_be_used_for_grouping? = field_format.in?(%w[list date bool int float string link])
+
+  # Template for all the join statements.
+  #
+  # For single value custom fields thejoin ensures single value for every
```
ulferts: Typo 'thejoin' -> 'the join'

ulferts (changes requested):
The change looks good functionality and also performance wise, in the tests I have conducted. It is now possible to group by way more fields which when looking at community is nice. 

There are some issues with the added grouping options, though. Of those, the first is but a display issue. But the second one leads to entries being sorted into the wrong group.

#### Grouping by integer cf
image

Here, the grouping happens with an integer cf. The groups however are presented as floats.

#### Grouping by link cf

image

The backend correctly returns just one item in the group. The frontend still places more entries into the group.

toy:
@ulferts I wasn't able to reproduce the problem with grouping, we should look at it together

toy:
Grouping by link is fixed by 804986d91121d149c2da2e97997d2ed3fe0597ad, the issue was already there, but it was not possible to group by link custom field. Similar problem exists for grouping by bool custom fields, as absence of value will lead to `nil` used for grouping/ordering, but default value (which can only be `true` or `false`) used for work package custom field value, so currently there will be odd groups or adding to wrong group, the issue exists on dev.
Very confusing result is on the following screenshot, where only work packages with ids 40 and 5 have value set for boolean custom field, so only first 3 groups are correct (`false`, `true` and `-`), following groups are created because custom field value is available for type TASK, so defaults to `false`, but not available for other types, so has null value, and as frontend expects groups to be in order, it creates a group every time next row group value is different from one for current row. But the false group must be in the api  response, as otherwise algorithm will not be be able to match it with row value and create it, so error doesn't show up, or at least that bad.

toy:
Last commit dc28e44b5e932542d9e97f5f4e0a899bb46dd2e2 is also extracted as #17201

ulferts (approved):
👍

PR #16842, opened 2024-09-27
[#57814] add detail page for hierarchy custom fields

Kharonus:
# Ticket
[OP#57814](https://community.openproject.org/work_packages/57814)

- editing of details of custom field of type hierarchy is possible
- form is written with primer components
  - spacing is off, but this is known for primer forms 
- existing custom fields of type hierarchy are not usable yet

# Screenshots

image

### DIFF
--- a/app/components/custom_fields/details_component.html.erb
+++ b/app/components/custom_fields/details_component.html.erb
@@ -1,0 +1,21 @@
+<%=
+  component_wrapper do
+    flex_layout do |content|
+      content.with_row(mb: 3) do
+        render Primer::Alpha::Banner.new(scheme: :default, icon: :info, dismiss_scheme: :hide) do
+          I18n.t("custom_fields.admin.notice.remember_items_and_projects")
+        end
+      end
+
+      content.with_row do
+        primer_form_with(
+          model:,
+          scope: :custom_field,
+          id: "custom_field_form",
+          url: custom_field_path(model),
+          method: :put
+        ) { |form| render CustomFields::DetailsForm.new(form) }
+      end
+    end
+  end
+%>
--- a/app/forms/custom_fields/details_form.rb
+++ b/app/forms/custom_fields/details_form.rb
@@ -1,0 +1,72 @@
+# frozen_string_literal: true
+
+#-- copyright
+# OpenProject is an open source project management software.
+# Copyright (C) the OpenProject GmbH
+#
+# This program is free software; you can redistribute it and/or
+# modify it under the terms of the GNU General Public License version 3.
+#
+# OpenProject is a fork of ChiliProject, which is a fork of Redmine. The copyright follows:
+# Copyright (C) 2006-2013 Jean-Philippe Lang
+# Copyright (C) 2010-2013 the ChiliProject Team
+#
+# This program is free software; you can redistribute it and/or
+# modify it under the terms of the GNU General Public License
+# as published by the Free Software Foundation; either version 2
+# of the License, or (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program; if not, write to the Free Software
+# Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+#
+# See COPYRIGHT and LICENSE files for more details.
+#++
+
+module CustomFields
+  class DetailsForm < ApplicationForm
+    form do |details_form|
+      details_form.text_field(
+        name: :name,
+        label: I18n.t(:label_name),
+        required: true
+      )
+
+      details_form.check_box(
+        name: :multi_value,
+        label: I18n.t("activerecord.attributes.custom_field.multi_value")
+      )
+
+      details_form.check_box(
+        name: :required,
+        label: I18n.t("activerecord.attributes.custom_field.is_required"),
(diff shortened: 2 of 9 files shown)


apfohl (approved):

app/views/custom_fields/edit.html.erb
```diff
@@ -33,9 +33,13 @@ See COPYRIGHT and LICENSE files for more details.
 
 <%= error_messages_for 'custom_field' %>
 
-<%= labelled_tabular_form_for @custom_field, as: :custom_field,
-                              url: custom_field_path(@custom_field),
-                              html: {method: :put, id: 'custom_field_form'} do |f| %>
-  <%= render partial: 'form', locals: { f: f } %>
-  <%= styled_button_tag t(:button_save), class: '-primary -with-icon icon-checkmark' %>
+<% if @custom_field.field_format == "hierarchy" %>
```
akabiru: 🟢 It would be nice to have encapsulated predicates: string vs symbol comparisons are easy pitfals.

```ruby
<% if @custom_field.field_format_hierarchy? %>
```

app/components/custom_fields/details_component.html.erb
```diff
@@ -0,0 +1,21 @@
+<%=
+  component_wrapper do
+    flex_layout do |content|
+      content.with_row(mb: 3) do
+        render Primer::Alpha::Banner.new(scheme: :default, icon: :info, dismiss_scheme: :hide) do
```
akabiru, line 5:
```suggestion
        render Primer::Alpha::Banner.new(scheme: :default, icon: :info, dismiss_scheme: :none) do
```

🟢 Seems a sticky banner is more fitting in this case?

A dismissible one seems odd for an info banner that's not a "flash" notice.

Screenshot 2024-10-01 at 2 20 38 PM

app/forms/custom_fields/details_form.rb
```diff
@@ -0,0 +1,72 @@
...
+    form do |details_form|
+      details_form.text_field(
+        name: :name,
+        label: I18n.t(:label_name),
+        required: true
+      )
+
+      details_form.check_box(
+        name: :multi_value,
+        label: I18n.t("activerecord.attributes.custom_field.multi_value")
```
akabiru, line 42: 🟢  Missing `caption` here: without caption, I'm not sure what "Allow Multi-select" actually does

app/forms/custom_fields/details_form.rb
```diff
@@ -0,0 +1,72 @@
...
+      details_form.check_box(
+        name: :searchable,
+        label: I18n.t("activerecord.attributes.custom_field.searchable"),
+        caption: I18n.t("custom_fields.instructions.searchable")
+      )
+
+      details_form.submit(name: :submit, label: I18n.t(:button_save), scheme: :default)
+    end
+  end
+end
```
akabiru, line 72: 🟠 🧑‍🎨 I couldn't find the designs but I think[ some grouping](https://primer-lookbook.github.com/view-components/lookbook/pages/forms/inputs/check_box_group) could touch up the UI and assign more meaning to the checkboxes- right now it feels like a _long_ list of checkboxes which doesn't look too good

akabiru (approved):
👍🏾 Just a few non-blocking remarks

app/components/custom_fields/details_component.html.erb
Kharonus: This is explicitly sketched in the wire frames, should I ask Marc again?

app/forms/custom_fields/details_form.rb
Kharonus: Also part of the wire frames, multi_value is an already existing thing. And it exists without caption. I wouldn't introduce it without a request from the design to do so.

app/forms/custom_fields/details_form.rb
Kharonus: The designs are in the parent [EPIC](https://community.openproject.org/wp/36033). here is the direct link:

https://www.figma.com/design/oTMDTvcwrx35qExTkkSKBn/Custom-fields?node-id=386-8676&t=F9b5GLm3QqZWEdB1-0

There are currently no groups intended, but of course, there is the matter with the vertical spacing, which we talked about last week. And conclusion was, I'd ignore spacings, as Henriette will introduce them.

app/views/custom_fields/edit.html.erb
Kharonus: good idea, I will introduce that.

app/components/custom_fields/details_component.html.erb
akabiru: It might be a case of simple copy paste- it's a tiny thing but would be good to clarify 👍🏾

app/forms/custom_fields/details_form.rb
akabiru: Agree it's worth checking with Design- I couldn't find the figma designs on the work package. Still, I think it's a good chance to improve the UX on that page- it hasn't been in the best state IMO.

app/forms/custom_fields/details_form.rb
akabiru: Thanks for the link! The grouping was also a UX question/proposal: we could group some checkboxes that change similar categories. I'm not too familiar with the model, so it's a proposal that can be ignored if already considered.

--

I was imagining sth like

```
This is setting group one
  - [ ] settings_check_box_1
  - [ ] settomgs_check_box_2

This is settings group two
  - [ ] settings_check_box_3
  - [ ] settings_check_box_4
```

Screenshot 2024-10-01 at 3 10 54 PM

Kharonus:
I took those three topics to the UX team:
- dismissable banner
- grouping of inputs
- caption for multi_value

Merging this PR, but the answers of it will eventually influence the upcoming changes.

PR #16845, opened 2024-09-28
No need breadcrumbs in HomeScreen.

Eric-Guo:

Original home screen have no breadcrumbs, but after 96129661235d380f32e8ce091714d671d2f0b1d2 it has.

### Before

image

### After

image

### Original (OpenProject Community)

image

Just remove no need breadcrumbs

# Merge checklist

- [ ] Added/updated tests

### DIFF
--- a/app/views/homescreen/index.html.erb
+++ b/app/views/homescreen/index.html.erb
@@ -26,12 +26,6 @@ Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
 See COPYRIGHT and LICENSE files for more details.
 
 ++#%>
-<%=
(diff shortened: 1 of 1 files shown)


HDinger (changes requested):
Hi @Eric-Guo 

as you probably know we are pushing towards a harmonized UI with the help of a design system and reusable components. Therefore, we applied our PageHeader component here, but forget to remove the old header which caused the duplication. 

However, instead of removing the PageHeader, the old custom headline should be removed. This was also discovered during the QA (see [here](https://community.openproject.org/projects/design-system/work_packages/53809/activity))

Eric-Guo:
Thanks the info, closing this PR now.

PR #16868, opened 2024-10-01
[#57815] add item page for hierarchies

Kharonus:
# Ticket
[OP#57815](https://community.openproject.org/work_packages/57815)

- add a new tab named "Items" to custom fields of type hierarchy

- add pure blankslate component to be shown 
- add new route `/custom_fields/:id/items`
- add tab Items under condition, that custom field is of type hierarchy
- add item button not functional, creation of items will  be in a later PR

# Screenshots

image

### DIFF
--- a/app/components/admin/custom_fields/hierarchy/item_component.html.erb
+++ b/app/components/admin/custom_fields/hierarchy/item_component.html.erb
@@ -28,33 +28,21 @@ See COPYRIGHT and LICENSE files for more details.
 ++#%>
 
 <%=
-  render(Primer::OpenProject::PageHeader.new(test_selector: "custom-fields--page-header")) do |header|
-    header.with_title { @custom_field.name }
-
-    header.with_breadcrumbs(breadcrumbs_items)
-
-    header.with_tab_nav(label: nil, test_selector: :custom_field_detail_header) do |tab_nav|
-      tab_nav.with_tab(
-        selected: tab_selected?(:edit),
-        href: edit_custom_field_path(@custom_field)
-      ) do |tab|
-        tab.with_text { t(:label_details) }
-      end
-
-      if @custom_field.field_format_hierarchy?
-        tab_nav.with_tab(
-          selected: tab_selected?(:items),
-          href: custom_field_items_path(@custom_field)
-        ) do |tab|
-          tab.with_text { t(:label_item_plural) }
+  component_wrapper do
+    flex_layout do |container|
+      container.with_row(mb: 3) do
+        render Primer::Beta::Blankslate.new(border: true) do |component|
+          component.with_visual_icon(icon: "list-ordered")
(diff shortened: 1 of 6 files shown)


app/components/admin/custom_fields/edit_form_header_component.html.erb
```diff
@@ -41,6 +41,15 @@ See COPYRIGHT and LICENSE files for more details.
         tab.with_text { t(:label_details) }
       end
 
+      if @custom_field.field_format_hierarchy?
```
akabiru, line 44: ✨

app/components/admin/custom_fields/hierarchy/item_component.html.erb
```diff
@@ -0,0 +1,49 @@
...
+<%=
+  component_wrapper do
+    flex_layout do |container|
+      container.with_row(mb: 3) do
+        render Primer::Beta::Blankslate.new(border: true) do |component|
+          component.with_visual_icon(icon: "list-ordered")
+          component.with_heading(tag: :h3).with_content(I18n.t("custom_fields.admin.items.blankslate.title"))
+          component.with_description { I18n.t("custom_fields.admin.items.blankslate.description") }
+        end
+      end
```
akabiru, line 39: 🟢 🧑‍🎨  design proposal: a  [`+ Item` primary action ](https://qa.openproject-edge.com/lookbook/inspect/primer/beta/blankslate/primary_action) within the blank slate might be more fitting than placing the + item below it. In some cases, I see it even done twice- as a primary action in the blankslate and a button below which is a nice pattern IMO - more convenient for the user and encourages them to "action" 😄

akabiru (commented):
Looking good, although missing (feature) tests 🟠 any plans to add those?

app/components/admin/custom_fields/hierarchy/item_component.html.erb
Kharonus: interesting proposal, will take it to design, too.

Kharonus: I took the question to the frontend discussion, with the result:
- no primary action of the blankslate, as a duplicated primary action is not a good design
- and the primary action below is needed, because once we fill the list with items, the blankslate vanishes but we still need the primary action

Kharonus:
Feature tests are planned, once there are more use cases implemented. IMHO we need at least:
- custom field creation ✅ 
- custom field editing (details) ✅ 
- custom field deletion ✅ 
- custom field editing (inserting items) ❌

additional use cases are (which then can be used to amend the feature spec):
- insert items under items
- editing items
- moving items
- deleting items
- soft deleting items

So, from this list, I'm missing one use case to write a sensible feature spec. ;)

apfohl (approved):

PR #16871, opened 2024-10-01
[57932] Make `Mails::MailerJob` retry failed emails

cbliard:
# Ticket

https://community.openproject.org/wp/57932

Make OpenProject retry sending emails when sending fails for a temporary reason (short outage of provider for instance).

Mailer jobs were discarded instead of being retried. Here is why:

Previously the `retry_on StandardError` defined in the `Mails::MailerJob` class was ignored because there is also a `rescue_from StandardError` declared after it. As `retry_on` is implemented using `rescue_from`, and the handlers are evaluated in reverse order, the last declared `rescue_from` would be picked up and the retry logic would not be triggered. That's why mail jobs were discarded instead of being retried.

Here is how it was fixed:

The issue was fixed by inheriting directly from `ActionMailer::MailDeliveryJob` instead of redefining its methods. This
way the `rescue_from` is declared first in the parent class, and then `retry_on` is declared second in the child class, meaning it will take precedence (because declared last) and be picked up.

The needed shared logic from `ApplicationJob` is extracted to a new
`SharedJobSetup` module and included in both `ApplicationJob` and
`Mails::MailerJob`.

# Merge checklist

- [ ] Added/updated tests

### DIFF
--- a/app/workers/application_job.rb
+++ b/app/workers/application_job.rb
@@ -30,13 +30,7 @@ require "active_job"
 
 class ApplicationJob < ActiveJob::Base
   include ::JobStatus::ApplicationJobWithStatus
-
-  ##
-  # By default, do not log the arguments of a background job
-  # to avoid leaking sensitive information to logs
-  self.log_arguments = false
-
-  around_perform :prepare_job_context
+  include SharedJobSetup
 
   ##
   # Return a priority number on the given payload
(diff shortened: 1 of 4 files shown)


machisuji (changes requested):
Thanks a lot for fixing that, @cbliard ! While it doesn't make sense to test rails ActiveJob logic, we have clearly been using it wrong in the past and didn't notice that it didn't work as intended. This is why I think this justifies a spec to test that we are using it correctly.

Could you please add one?

cbliard:
Sure, I can try adding one. I'm curious to see what I will end up with.

cbliard:
@machisuji I added a test for it. Can you review it again?

Also I'm concerned that it will be a bit harder to find errors with unit tests: if a mailer produces a `StandardError` for some reason, then the exception will be catched and the mailing job reenqueued for a retry. Previously it would have raised the error so it was easy to spot.

You can easily see it if you comment the line `retry_on ...` in the `Mails::MailerJob` and run the spec: it displays the "ArgumentError" error and the test fails. For that test it's ok because that's what we test, but if we introduce a bug in a mailer, will our test suite still catch it?

machisuji:
> @machisuji I added a test for it. Can you review it again?
> 
> Also I'm concerned that it will be a bit harder to find errors with unit tests: if a mailer produces a `StandardError` for some reason, then the exception will be catched and the mailing job reenqueued for a retry. Previously it would have raised the error so it was easy to spot.
> 
> You can easily see it if you comment the line `retry_on ...` in the `Mails::MailerJob` and run the spec: it displays the "ArgumentError" error and the test fails. For that test it's ok because that's what we test, but if we introduce a bug in a mailer, will our test suite still catch it?

Great! Thanks for adding the test!

As for your concern, I suppose it does obscure the failure for a test in case we do break a mailer somewhere. But usually it should still be found because we would probably test that a mail did go out and contained certain contents?

I suppose you could amend the spec helper with a `retry_stopped` (`ActiveSupport::Notifications.subscribe "retry_stopped.active_job" { }` (see [active support instrumentation](https://guides.rubyonrails.org/active_support_instrumentation.html#retry-stopped-active-job)) subscription and raise the error there in all specs except for this retry spec?

machisuji (approved):
👍

cbliard:
> I suppose you could amend the spec helper with a `retry_stopped` (`ActiveSupport::Notifications.subscribe "retry_stopped.active_job" { }` (see [active support instrumentation](https://guides.rubyonrails.org/active_support_instrumentation.html#retry-stopped-active-job)) subscription and raise the error there in all specs except for this retry spec?

I get the idea but that would not work: `retry_stopped` notification happens when the number of attempts has been reached, right before the job is discarded and the error is raised.

Well, the jobs are not performed that often. It should not be too much of a problem.

PR #16881, opened 2024-10-02
[#58143] Add the new role and apply defaults

dombesz:
# Ticket
Both tickets are addressed in this PR:
- [Add the new role and apply defaults](https://community.openproject.org/wp/58143)
- [Extend api with the email address for the autocompleters](https://community.openproject.org/wp/58145)

Add a new global permission "View users' email addresses" and add a new global builtin role "Standard global role" in order to have a place for assigning roles for every user in the system.
Enable the "View users' email addresses" on this new role by default.

Enforce displaying the user's email and allow searching users by email addresses via the API based on the "View users' email addresses" permission.

# Merge checklist

- [x] Added/updated tests

### DIFF
--- a/app/models/global_role.rb
+++ b/app/models/global_role.rb
@@ -31,4 +31,15 @@ class GlobalRole < Role
     super
       .where(type: "GlobalRole")
   end
+
+  def self.standard
+    standard_global_role = where(builtin: BUILTIN_STANDARD_GLOBAL).first
+    if standard_global_role.nil?
+      standard_global_role = create(name: "Standard global role", position: 0) do |role|
+        role.builtin = BUILTIN_STANDARD_GLOBAL
+      end
+      raise "Unable to create the standard global role." if standard_global_role.new_record?
+    end
+    standard_global_role
+  end
 end
--- a/app/controllers/members_controller.rb
+++ b/app/controllers/members_controller.rb
@@ -118,6 +118,10 @@ class MembersController < ApplicationController
     current_user.allowed_in_project?({ controller:, action: }, @project)
   end
 
+  def user_allowed_to_view_emails?
+    current_user.allowed_globally?(:view_user_email)
+  end
+
   def build_members
     paths = API::V3::Utilities::PathHelper::ApiV3Path
     principals = @principals.map do |principal|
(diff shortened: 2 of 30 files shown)


app/models/global_role.rb
```diff
@@ -31,4 +31,15 @@ def self.givable
...
+  def self.standard
+    standard_global_role = where(builtin: BUILTIN_STANDARD_GLOBAL).first
+    if standard_global_role.nil?
+      standard_global_role = create(name: "Standard global role", position: 0) do |role|
+        role.builtin = BUILTIN_STANDARD_GLOBAL
+      end
+      raise "Unable to create the standard global role." if standard_global_role.new_record?
+    end
+    standard_global_role
+  end
```
aaron-contreras, line 44: Would this not work too?
```suggestion

  def self.standard
    standard_global_role = find_or_create_by(name: "Standard global role", builtin: BUILTIN_STANDARD_GLOBAL, position: 0)
    raise "Unable to create the standard global role." unless standard_global_role.persisted?
    standard_global_role
  end
```

aaron-contreras (commented):
Just one suggestion! Perhaps there's a reason not to do it this way but I'm curious about it.

app/models/global_role.rb
dombesz: The `find_or_create_by(name: "Standard global role", builtin: BUILTIN_STANDARD_GLOBAL, position: 0)` could be problematic if let's say the name or position differs. I would rather use the `builtin: BUILTIN_STANDARD_GLOBAL` as a finding condition only.

aaron-contreras: Okay sure, just wondering the reason behind it. All looks good 🌴

aaron-contreras (approved):

app/models/global_role.rb
dombesz: Hi @aaron-contreras ,

Thanks for the review, I will try to summarise the scenario. Let's say the role already exists in the database, with a different position. Position will be reassigned at some point, because it is managed by the acts as list gem. If we have the `position:0` in the finder conditions, the existing role wouldn't be found and it would try to create a new Role, which is incorrect. Same goes for the name too, if the role exists with a different name, but with the same `builtin` flag, then we end up creating a duplicate role for the `BUILTIN_STANDARD_GLOBAL` flag. We also cannot solely rely on the name when finding the existing role, because that would lead to a modification of an existing role that wasn't done by the system.

Perhaps it would be better to have a uniqueness validation on the `builtin: Role::BUILTIN_STANDARD_GLOBAL`, but since we already have this pattern in other places, I considered it out of scope for this PR.

We have this pattern implemented for the ProjectRoles:

https://github.com/opf/openproject/blob/fb607a24d32585171d55b1fedffc059af17706f4/app/models/project_role.rb#L37-L46

PR #16884, opened 2024-10-02
Implementation/58105 item persistence service

apfohl:
# Ticket
[OP#58105](https://community.openproject.org/projects/openproject/work_packages/58105/github)

PART 1

Implement a service capable of manipulation a hierarchy of items for the custom field of type `hierarchy`.

- `initialize`
- `generate_root`
- `insert_item`

- Dry::Validation contracts for input checking
- Dry::Monads for error handling

# Merge checklist

- [ ] Added/updated tests


spec/contracts/custom_fields/hierarchy/insert_item_contract_spec.rb
```diff
@@ -0,0 +1,119 @@
...
+          { parent:, label: "A label" }
+        ].each { |params| expect(subject.call(params)).to be_success }
+      end
+    end
+
+    context "when inputs are invalid" do
+      it "creates a failure result" do
+        [
+          { parent:, label: "A label", short: "" },
+          { parent:, label: "A label", short: nil },
```
Kharonus, line 106: 🔴  we might have a problem with that case. If I interpret the usage in the service correct, it will call it, when `short` is not passed.

Solutions:
a) adapt contract, so that passing `nil` is allowed, while `""` still is forbidden.
b) adapt service to call contract only with non-nil shorts or no short key at all.

To make b) work we would need to create a hash from the method parameters before putting them into the call. something like: 

```ruby
input = {parent:, label:, short:}.compact

CustomFields::Hierarchy::InsertItemContract
  .new
  .call(**input)
```

app/contracts/custom_fields/hierarchy/insert_item_contract.rb
```diff
@@ -0,0 +1,57 @@
...
+    class InsertItemContract < Dry::Validation::Contract
+      params do
+        required(:parent).filled
+        required(:label).filled(:string)
+        optional(:short).filled(:string)
+      end
+
+      rule(:parent) do
+        if value.is_a?(CustomField::Hierarchy::Item)
+          unless value.persisted?
```
Kharonus, line 42: 🟡 can we combine conditions and thus shrink the indention a bit? 

```ruby
return key.failure("Parent must be of type 'Item'") unless value.is_a?(CustomField::Hierarchy::Item)

unless value.persisted?
  key.failure("Parent must exist")
end
```

app/contracts/custom_fields/hierarchy/generate_root_contract.rb
```diff
@@ -0,0 +1,43 @@
...
+
+module CustomFields
+  module Hierarchy
+    class GenerateRootContract < Dry::Validation::Contract
+      params do
+        required(:hierarchy_root)
+      end
+
+      rule(:hierarchy_root) do
+        key.failure("Hierarchical root already set") unless value.nil?
```
Kharonus, line 39: 🟡 change to `if value.present?` to battle this godless `unless` :D

app/contracts/custom_fields/hierarchy/service_initialization_contract.rb
```diff
@@ -0,0 +1,43 @@
...
+
+module CustomFields
+  module Hierarchy
+    class ServiceInitializationContract < Dry::Validation::Contract
+      params do
+        required(:field_format).filled(:string)
+      end
+
+      rule(:field_format) do
+        key.failure("Custom field must have field format 'hierarchy'") unless value == "hierarchy"
```
Kharonus: 🟡 same here: can we use `if value != "hierarchy"`?

spec/services/custom_fields/hierarchy/hierarchical_item_service_spec.rb
```diff
@@ -0,0 +1,98 @@
...
+  describe "#generate_root" do
+    let(:service) { described_class.new(custom_field) }
+
+    context "with valid hierarchy root" do
+      it "creates a root item successfully" do
+        expect(service.generate_root).to be_success
+      end
+    end
+
+    context "with invalid hierarchy root" do
```
Kharonus: ❓ I don't understand the test case. So, is this an item object which cannot get persisted? 

Maybe renaming it to "if persistence of hierarchy root fails"? Otherwise some transitional head work is needed when reading the test - as in "an invalid hierarchy root will lead to a persistence error". But this is no hill I gonna die upon.

Kharonus (changes requested):
I think, we should take care of some small bobs and bits, wdyt?

spec/contracts/custom_fields/hierarchy/insert_item_contract_spec.rb
apfohl: `short` is indeed passed, having `short: nil` in the hash actually creates the key in the hash. Therefor you have three cases: `short: ""`, `short: nil` and no `short`.

app/contracts/custom_fields/hierarchy/insert_item_contract.rb
apfohl: Early return not possible it's is already the shortest form.

spec/contracts/custom_fields/hierarchy/insert_item_contract_spec.rb
Kharonus: discussed personally, there is indeed a faulty call from the service. Another test case discovered that.

spec/services/custom_fields/hierarchy/hierarchical_item_service_spec.rb
```diff
@@ -0,0 +1,105 @@
...
+
+    context "with valid parameters" do
+      it "inserts an item successfully" do
+        result = service.insert_item(parent:, label:)
+        expect(result).to be_success
+      end
+    end
+
+    context "with invalid short" do
+      it "fails to insert an item" do
```
Kharonus: 🔴 description does not match. I'd suggest:

```ruby
    context "with valid parameters" do
      it "inserts an item successfully without short" do
        result = service.insert_item(parent:, label:)
        expect(result).to be_success
      end

      it "inserts an item successfully with short" do
        result = service.insert_item(parent:, label:, short:)
        expect(result).to be_success
      end
    end
```

Kharonus (changes requested):

Kharonus (approved):
LGTM! nice!

PR #16926, opened 2024-10-09
[58147] Add caption to clarify status % Complete field in administration

cbliard:
# Ticket

https://community.openproject.org/wp/58147

In Administration > Work packages > Status > Edit or Create page, add a caption to % Complete field to indicate that it is only used in status-based mode.

As the page has been touched, it has been primerized as well

## Screenshots

Before:

image

After:

image

For primerization:
* Create a dedicated form object
* Reuse the angular component for color selection. The color select field can be reused in other forms by using `#color_select_list`.
  * FYI I tried primerizing this color selector with a `SelectPanel` but it did not work well enough.
* Fixed minor bugs that occurred after the edit form is submitted and it has errors: when the form is displayed again, it was displaying the updated name in the header instead of displaying the original name. Likewise, the "make default" checkbox could disappear.

# Merge checklist

- [ ] Added/updated tests
  - current test suite should continue working


app/forms/statuses/form.rb
```diff
@@ -0,0 +1,128 @@
...
+      statuses_form.text_field(
+        label: attribute_name(:name),
+        name: :name,
+        required: true,
+        input_width: :medium
+      )
+
+      statuses_form.text_field(
+        label: attribute_name(:default_done_ratio),
+        name: :default_done_ratio,
```
ulferts, line 51: 💡 In this form, the labels could have been deduced from the name. Don't know if that is always the case but it looks like a good default in case we want to optimize this. 

Not necessary to address in this PR.

app/forms/statuses/form.rb
```diff
@@ -0,0 +1,128 @@
...
+      statuses_form.check_box(
+        label: attribute_name(:is_closed),
+        name: :is_closed
+      )
+
+      unless already_default_status?
+        statuses_form.check_box(
+          label: attribute_name(:is_default),
+          name: :is_default
+        )
```
ulferts, line 71: I personally would prefer to still display the checkbox in a disabled state if the current status is the default one. And there are probably even better ways but this would be an easy improvement IMO.

app/views/statuses/edit.html.erb
```diff
@@ -27,19 +27,20 @@ See COPYRIGHT and LICENSE files for more details.
 
 ++#%>
 
-<% html_title t(:label_administration), "#{t(:label_edit)} #{Status.model_name.human} #{h @status.name}" %>
+<% html_title t(:label_administration), "#{t(:label_edit)} #{Status.model_name.human} #{h @status.name_was}" %>
```
ulferts, line 30: It existed before but the `h` can by now be removed since the string will be escaped anyway.

app/views/statuses/edit.html.erb
```diff
@@ -27,19 +27,20 @@ See COPYRIGHT and LICENSE files for more details.
...
 
 <%=
   render Primer::OpenProject::PageHeader.new do |header|
-    header.with_title { @status.name }
+    header.with_title { @status.name_was }
     header.with_breadcrumbs([{ href: admin_index_path, text: t("label_administration") },
                              { href: admin_settings_work_packages_general_path, text: t(:label_work_package_plural) },
                              { href: statuses_path, text: t(:label_status) },
-                             @status.name])
+                             @status.name_was])
```
ulferts, line 38: 👍 for using the former name on this one.

app/views/statuses/edit.html.erb
```diff
@@ -27,19 +27,20 @@ See COPYRIGHT and LICENSE files for more details.
...
   end
 %>
 
-<%= labelled_tabular_form_for(@status) do |f| %>
-  <%= render partial: 'form', locals: { f: f } %>
-  <%= styled_button_tag t(:button_save), class: '-primary -with-icon icon-checkmark' %>
+<%= error_messages_for :status %>
```
ulferts, line 42: 💡 I don't know if there are error cases defined for status on `base` rather than on an attribute. The way the errors are currently displayed is fine but duplicates the message:

image

Out of scope for this PR but changing the error message helper so that e.g. via a flag it only displays errors on `base` could be a simple improvement.

ulferts (approved):
I only added ideas and one nit-picking. Good to merge if you decide that this is the scope you want it to merge with, @cbliard .

Bonus points for adding the ColorSelect to the form.

app/views/statuses/edit.html.erb
ulferts: Primer btw suggests to use [banners for a summarizing errors](https://primer.style/ui-patterns/forms/overview#interactive-summary-of-errors). Could also be applied to `base` errors.

app/forms/statuses/form.rb
cbliard: Yes, indeed, but I'm wary about deviating from the simplicity of using the existing helpers. There is a bit of duplication, but it's not a big deal as it reads well. I may change my mind later.

app/forms/statuses/form.rb
cbliard: I only ported the existing behavior without questioning it.
When looking at it with Parimal some months ago, he was not satisfied with this way of choosing which attribute is the default, but did not want to touch it for the moment. I'll leave it like this too.

app/views/statuses/edit.html.erb
cbliard: Yes, but showing `base` errors only could be problematic too: checkboxes do not have any inline validation message, unless they are part of a checkbox group, and this checkbox group must have a name matching the one in the model to get the inline validation message being displayed.

To try it, check both "is default" and "is readonly". There will be an error message on "is default" saying that it can't be active when "is readonly" is also checked. It is displayed in the error banner, but not inline. That's a bit of a regression compared to how the page behaved as before the label and the box border would be in red (without message though).

Anyway, this has to be solved globally, and I think Henriette is the person who should coordinate that. I'll show her the page to get her feedback about it.

app/views/statuses/edit.html.erb
cbliard: I tried but this does not work as expected:

with `h`:

image

without `h`:

image

Note the difference in the tab title in the left sidebar.

So it's still valid, there no XSS injection, but for consistency, if the tags are escaped everywhere (header, breadcrumbs, etc), they should be escaped in the title as well, not stripped.

About where it's done: the string is used to set the title using meta_tags, and this meta_tags calls `strip_tags` at some point (when calling [`#normalize_title`](https://github.com/kpumuk/meta-tags/blob/00358f4160eafe3e5c21cb9466a5d794e541b5e2/lib/meta_tags/text_normalizer.rb#L8-L34)).

app/forms/statuses/form.rb
cbliard: We [discussed it](https://community.openproject.org/meetings/3690#item-5949) in frontend daily, and it ended up as a bug: [Bug: Statuses edit page should show all checkboxes and disable those leading to impossible states (#58360)](https://community.openproject.org/projects/openproject/work_packages/58360/activity)

PR #16930, opened 2024-10-10
[#57816] create hierarchy items

Kharonus:
# Ticket
[OP#57816](https://community.openproject.org/work_packages/57816)

- added form for item creation
- added borderbox in items component

- enable creation of hierarchy items on top most level

- enable turbo stream updates to render inline form to create items
- add form validation
- render existing items in borderbox


apfohl (approved):

app/components/admin/custom_fields/hierarchy/items_component.rb
```diff
@@ -0,0 +1,53 @@
...
+        include OpPrimer::ComponentHelpers
+
+        def initialize(custom_field:, new_item_form_data: { show: false })
+          super
+          @custom_field = custom_field
+          @new_item_form_data = new_item_form_data
+        end
+
+        def items
+          # TODO: This must be context aware (breadcrumbs)
```
akabiru, line 43: 🟢 open TODO

akabiru (approved):
🥇

app/components/admin/custom_fields/hierarchy/items_component.rb
Kharonus: Ah, yes, that's a todo for the feature, but not for the implementation. There is a follow up implementation ticket, that inserts logic in there dependent on the the current hierarchy level.

spec/support/i18n_lazy_loading.rb
```diff
@@ -28,16 +28,19 @@
...
-    locale.to_s.delete_prefix("js-").to_sym if locale
-    # name = File.basename(path, ".*")
-    # locale = name.split("_").first.delete_prefix("js-")
-    # locale&.to_sym
+    name = File.basename(path, ".*")
+    locale = name.split("_").first
+
+    return if locale.nil?
+
+    locale.tr("js-", "").to_sym
```
cbliard: `tr` is cryptic and a lot less intention-revealing than other methods like `delete_prefix` or `gsub`.
And also it does not work as expected as it deletes every `j`, `s`, and `-` characters instead of deleting the word `js-`

```
>> "js-ja".tr("js-", "")
=> "a"
>> "js-ja".delete("js-")
=> "a"
>> "js-ja".delete_prefix("js-")
=> "ja"
```

Better use `delete_prefix`.

PR #16939, opened 2024-10-11
[58105] Added update_item and delete_branch methods to HierarchicalItemService

apfohl:
# Ticket
https://community.openproject.org/projects/openproject/work_packages/58105/github

See headline

same as last time

# Merge checklist

- [ ] Added/updated tests


app/contracts/custom_fields/hierarchy/update_item_contract.rb
```diff
@@ -0,0 +1,61 @@
...
+  module Hierarchy
+    class UpdateItemContract < Dry::Validation::Contract
+      params do
+        required(:item).filled
+        optional(:label).filled(:string)
+        optional(:short).filled(:string)
+      end
+
+      rule(:item) do
+        if value.is_a?(CustomField::Hierarchy::Item)
```
Kharonus: 🔴 @mereghost Did you update this here, too? In regard of checking for the type in the schema already?

app/contracts/custom_fields/hierarchy/update_item_contract.rb
```diff
@@ -0,0 +1,61 @@
...
+        else
+          key.failure("Item must be of type 'Item'")
+        end
+      end
+
+      rule(:label) do
+        next unless key?
+
+        if CustomField::Hierarchy::Item.exists?(parent_id: values[:item].parent, label: value)
+          key.failure("Label must be unique within the same hierarchy level")
```
Kharonus: 🟢 I will put this message on localization once I start with the update action in the UI. The localization will get merged today.

app/models/custom_field/hierarchy/item.rb
```diff
@@ -32,5 +32,5 @@ class CustomField::Hierarchy::Item < ApplicationRecord
   self.table_name = "hierarchical_items"
 
   belongs_to :custom_field
-  has_closure_tree order: "sort_order", numeric_order: true
+  has_closure_tree order: "sort_order", numeric_order: true, dependent: :delete_all
```
Kharonus: 👍🏽

app/services/custom_fields/hierarchy/hierarchical_item_service.rb
```diff
@@ -55,7 +55,22 @@ def insert_item(parent:, label:, short: nil)
...
+          .new
+          .call({ item:, label:, short: }.compact)
+          .to_monad
+          .bind { |attributes| update_item_attributes(item:, attributes:) }
+      end
+
+      def delete_branch(item:)
+        return Failure(:item_is_root) if item.root?
+
+        # CustomField::Hierarchy::Item sets "dependent: :destroy"
```
Kharonus: 🟡 can we remove the comment? It is already oudated :D

app/services/custom_fields/hierarchy/hierarchical_item_service.rb
```diff
@@ -67,13 +82,18 @@ def create_root_item
         Success(item)
       end
 
-      def create_child_item(validation)
-        item = CustomField::Hierarchy::Item
-                 .create(parent: validation[:parent], label: validation[:label], short: validation[:short])
+      def create_child_item(validation:)
+        item = validation[:parent].children.create(label: validation[:label], short: validation[:short])
```
Kharonus, line 122: 🟡 Hmm, not sure, but I do not like this change. What is the benefit? Before the code was very clear, that a new Item is created. Now I must understand, that the item to create is referenced here as `.children`.

🔴 Does this even work, if the created item is not the only child?

spec/services/custom_fields/hierarchy/hierarchical_item_service_spec.rb
```diff
@@ -63,39 +65,99 @@
...
         expect(result).to be_success
       end
     end
 
     context "with invalid item" do
       it "fails to insert an item" do
-        allow(CustomField::Hierarchy::Item)
-          .to receive(:create).and_return(instance_double(CustomField::Hierarchy::Item,
-                                                          persisted?: false, errors: "some errors"))
+        # rubocop:disable RSpec/VerifiedDoubles
```
Kharonus: ❓ This cop should not trigger, as `instance_double` is a verified double?

Kharonus (changes requested):

spec/services/custom_fields/hierarchy/hierarchical_item_service_spec.rb
mereghost: But the wrapping double isn't. I'll see if I can remove this by adding the correct double here.

app/services/custom_fields/hierarchy/hierarchical_item_service.rb
mereghost: No! :P Sure.

app/services/custom_fields/hierarchy/hierarchical_item_service.rb
mereghost: Because that's how `closure_tree` works. It keeps intermediate representations of the tree for lookup, so we need to go through its code.

From another perspective, it is clearer to the random RoR dev as `children.create` maps to "I'm creating some associated record to the caller". :P

spec/services/custom_fields/hierarchy/hierarchical_item_service_spec.rb
mereghost: Fixed.

app/services/custom_fields/hierarchy/hierarchical_item_service.rb
Kharonus: Can you confirm, that this does not break, if the parent already has persisted children, when adding a new one?

app/contracts/custom_fields/hierarchy/generate_root_contract.rb
```diff
@@ -32,11 +32,7 @@ module CustomFields
   module Hierarchy
     class GenerateRootContract < Dry::Validation::Contract
       params do
-        required(:hierarchy_root)
-      end
-
-      rule(:hierarchy_root) do
-        key.failure("Hierarchical root already set") unless value.nil?
+        required(:hierarchy_root).value(:nil?)
```
Kharonus: 🟢 nice!

app/contracts/custom_fields/hierarchy/update_item_contract.rb
```diff
@@ -0,0 +1,54 @@
...
+    class UpdateItemContract < Dry::Validation::Contract
+      params do
+        required(:item).filled(type?: CustomField::Hierarchy::Item)
+        optional(:label).filled(:string)
+        optional(:short).filled(:string)
+      end
+
+      rule(:item) do
+        key.failure("must exist") if value.new_record?
+        key.failure("must not be a root item") if value.root?
```
Kharonus, line 42: ❓ is `root?` an extension method from `closure_tree`? Or can't I simply find it in our codebase?

app/services/custom_fields/hierarchy/hierarchical_item_service.rb
```diff
@@ -55,25 +54,64 @@ def insert_item(parent:, label:, short: nil)
...
+
+        item.destroy ? Success() : Failure(item.errors)
+      end
+
+      def get_branch(item:)
+        Success(item.ancestors.reverse)
+      end
+
+      def move_item(item:, new_parent:)
+        Success(new_parent.append_child(item))
```
Kharonus, line 95: 🟢 seems like closure tree indeed comes with some nice helpers.

Kharonus (approved):
LGTM generally, can you answer the last two questions though?

app/contracts/custom_fields/hierarchy/update_item_contract.rb
mereghost: [Yep](https://github.com/ClosureTree/closure_tree/tree/master?tab=readme-ov-file#instance-methods), from closure_tree.

app/services/custom_fields/hierarchy/hierarchical_item_service.rb
mereghost: Nice, huh?

app/services/custom_fields/hierarchy/hierarchical_item_service.rb
mereghost: Yes. It works. Why? Because `.children` is an association proxy that don't override anything but ensure that the fields for the association are correctly set - and in case of closure tree, sets up the rest of the fields and optimizations.

PR #16940, opened 2024-10-11
Save OIDC tokens to OpenProject database

ba1ash:
# Ticket

https://community.openproject.org/wp/58365

Storing access and refresh tokens for users that initiate their session through OIDC, so that we can later use them to obtain tokens for accessing third party resources, such as Nextcloud.

Storing the access and refresh token on a new table that is intended to store all access tokens for the current session's user, including the ones later obtained through Token Exchange. This provides a clear place to look for credentials and additionally keeps the back-channel logout logic separated from token storage (as opposed to storing the tokens on the `user_session_link`.

# Merge checklist

- [x] Added/updated tests

### DIFF
--- a/app/models/sessions/user_session.rb
+++ b/app/models/sessions/user_session.rb
@@ -1,3 +1,5 @@
+# frozen_string_literal: true
+
 #-- copyright
 # OpenProject is an open source project management software.
 # Copyright (C) the OpenProject GmbH
--- a/modules/openid_connect/app/services/openid_connect/associate_user_token.rb
+++ b/modules/openid_connect/app/services/openid_connect/associate_user_token.rb
@@ -1,3 +1,5 @@
+# frozen_string_literal: true
+
 #-- copyright
 # OpenProject is a project management system.
 # Copyright (C) the OpenProject GmbH
--- a/modules/openid_connect/app/services/openid_connect/associate_user_token.rb
+++ b/modules/openid_connect/app/services/openid_connect/associate_user_token.rb
@@ -28,34 +28,26 @@
 
 module OpenIDConnect
   class AssociateUserToken
-    def initialize(session)
-      @session = session
+    def initialize(user)
+      @user = user
     end
 
-    def call(access_token:, refresh_token: nil, assume_idp: false)
+    def call(access_token:, refresh_token: nil, known_audiences: [], clear_previous: false)
       if access_token.blank?
-        Rails.logger.error("Could not associate token to session: No access token")
+        Rails.logger.error("Could not associate token to user: No access token")
         return
       end
 
-      user_session = find_user_session
-      if user_session.nil?
-        Rails.logger.error("Could not associate token to session: User session not found")
+      if @user.nil?
+        Rails.logger.error("Could not associate token to user: Can't find user")
         return
       end
 
-      token = user_session.oidc_user_tokens.build(access_token:, refresh_token:)
+      @user.oidc_user_tokens.destroy_all if clear_previous
+
+      token = @user.oidc_user_tokens.build(access_token:, refresh_token:, audiences: Array(known_audiences))
       # We should discover further audiences from the token in the future
-      token.audiences << UserToken::IDP_AUDIENCE if assume_idp
-
       token.save! if token.audiences.any?
-    end
-
-    private
-
-    def find_user_session
-      private_session_id = @session.id.private_id
-      ::Sessions::UserSession.find_by(session_id: private_session_id)
     end
   end
 end
--- a/modules/openid_connect/app/services/openid_connect/associate_user_token.rb
+++ b/modules/openid_connect/app/services/openid_connect/associate_user_token.rb
@@ -1,12 +1,12 @@
 #-- copyright
-# OpenProject is an open source project management software.
+# OpenProject is a project management system.
 # Copyright (C) the OpenProject GmbH
 #
 # This program is free software; you can redistribute it and/or
 # modify it under the terms of the GNU General Public License version 3.
 #
 # OpenProject is a fork of ChiliProject, which is a fork of Redmine. The copyright follows:
-# Copyright (C) 2006-2013 Jean-Philippe Lang
+# Copyright (C) 2006-2017 Jean-Philippe Lang
 # Copyright (C) 2010-2013 the ChiliProject Team
 #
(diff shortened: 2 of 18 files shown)


NobodysNightmare:
An alternative approach to storing the tokens on the `UserSessionLinks` table might be to associate another table to the session that specifically stores tokens only.

In general, we are not done yet: One next step will be to write code that can take a valid access token stored here and exchange it through OAuth 2.0 Token Exchange for a token that can be used to access a third party service. Eventually there will be multiple third party services, that we want to exchange tokens for and then we want to reuse those tokens until they expire.

This means we could store _all_ of those tokens in the same table (e.g. with an additional `audience` column) along with our own tokens. I'd prefer this approach for storing all tokens the same way. On the other hand, some special logic will be required, since for token exchange we will always want to use our own tokens to perform the exchange.

NobodysNightmare:
I did what I suggested above. This PR now contains both, the original version (check the diff of commit 1) and the one suggested by me (check the diff of the total PR). My personal prefererence is the current version of this PR.

modules/openid_connect/app/services/openid_connect/create_open_project_token.rb
```diff
@@ -0,0 +1,65 @@
...
+        return
+      end
+
+      user_session = find_user_session
+      if user_session.nil?
+        Rails.logger.error("Could not associate token to session: User session not found")
+        return
+      end
+
+      user_session.oidc_user_tokens.open_project.create!(access_token:, refresh_token:)
```
NobodysNightmare: TBD: Probably we should look at the actual audiences of the token instead of just assuming "open project" only.

Maybe the token already gives us more audiences.

NobodysNightmare:
One other thought that crossed my mind today and was then repeated by @wielinde:

Should we store the tokens on the session or on the user?

Arguments _against_ session:
* We'd have to push the correct session into all storage-services, instead of pushing the user there
* A session is harder to obtain than a user (imagine calling a service from a background worker)

The argument _for_ session:
* Right now we obtain the tokens in the context of an SSO session, so they may be affected by IDP logout
    * thus it's not safe to use tokens from arbitrary sessions
    * but maybe we can obtain a token for long-term offline use instead? (that could be bound to user safely)

modules/openid_connect/app/services/openid_connect/create_open_project_token.rb
NobodysNightmare: Solved that now. First error was that we'd be the audience, which is not correct per spec. Only the IDP needs to be in the audience, because the access token is exchanged there.

We will now always assume the IDP to be in the audience for a token received via Authorization Code flow. Since we are dealing with access tokens here, there's no guaranteed way to determine additional details about the token. ID Tokens are guaranteed to by JWTs according to OpenID Core, but access tokens may remain opaque. Figuring out further audiences was extracted into a follow up chunk: https://community.openproject.org/projects/cross-application-user-integration-stream/work_packages/60162

NobodysNightmare:
> but maybe we can obtain a token for long-term offline use instead? (that could be bound to user safely)

For Keycloak in default config, this is possible by adding the `offline_access` scope to the list of requested scopes. This will make sure that the lifetime of refresh tokens is not limited and they are not revoked after logout anymore.

So we could attach tokens to the user instead of the session, if we can ensure that they are not expiring. I see elevated potential for misconfiguration here, but on the other hand it still sounds like the cleanest available solution to me.

edit: However, seeing that `offline_access` is a standardized scope name, we could make sure that there is a properly configured OIDC provider available...

app/models/sessions/user_session.rb
```diff
@@ -33,6 +33,8 @@ module Sessions
   class UserSession < ::ApplicationRecord
     self.table_name = "sessions"
 
+    belongs_to :user
```
mereghost, line 38: Note: there's no FK or index over `user_id` so queries over it can be extra slow.

modules/openid_connect/app/services/openid_connect/associate_user_token.rb
```diff
@@ -0,0 +1,53 @@
...
+#
+# You should have received a copy of the GNU General Public License
+# along with this program; if not, write to the Free Software
+# Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+#
+# See COPYRIGHT and LICENSE files for more details.
+# +
+
+module OpenIDConnect
+  class AssociateUserToken
```
mereghost, line 32: 🟢 Should we add `Service` to the name of this object?

modules/openid_connect/app/services/openid_connect/associate_user_token.rb
```diff
@@ -0,0 +1,53 @@
...
+        Rails.logger.error("Could not associate token to user: Can't find user")
+        return
+      end
+
+      @user.oidc_user_tokens.destroy_all if clear_previous
+
+      token = @user.oidc_user_tokens.build(access_token:, refresh_token:, audiences: Array(known_audiences))
+      # We should discover further audiences from the token in the future
+      token.save! if token.audiences.any?
+    end
```
mereghost, line 53: 🟡 The code implies that this creates a OIDC user token not really `associate` an existing one. The name of the object should probably match this.

And then probably also follow the established pattern of `OpenIDConnect::UserTokens::CreateService`.

mereghost (approved):
The work seems solid, but slightly off the already existing patterns in some places. Nothing that can't be addressed later if this unblocks others work.

NobodysNightmare:
> The work seems solid, but slightly off the already existing patterns in some places. Nothing that can't be addressed later if this unblocks others work.

Your remarks about the naming are valid. My currently preferred approach would be:

1. Merge the two pending PRs that depend on each other (this and #17469)
2. Rename `AssociateUserToken` to `UserTokens::CreateService`

I pretty clearly wrote this code before realizing that there was a naming patterns across the larger codebase.

app/models/sessions/user_session.rb
NobodysNightmare: That's a good hint. I would have probably missed that, because I merely added this accessor for convenience.

Though, double thinking about this, the performance penalty is when accessing the inverse relation, correct? `user_session.user` will have already loaded a `user_sessions` row and find a user by its (indexed) `id`.

Anyhow: We should keep that in mind, once we think about using this relation further.

PR #16944, opened 2024-10-14
[#52147] quick action table headers

EinLama:
# Ticket
https://community.openproject.org/wp/52147 

Clicking on the table header in the project list offers a variety of actions to perform on this column.

Sort by a column, move columns around, use a filter by a column value, add or remove columns. All this now at the convenience of a click.

## Screenshots

https://github.com/user-attachments/assets/b2afbe5d-377a-4378-8577-7be15ca04250

I went with Primers `ActionMenu` as it seemed like a good fit. Where possible, existing functionality of the project list was kept. For example, remembering the query parameters for previous changes when making further changes (first sort, then change available columns, then sort again).

To interact with the filter component and configure view components, I used Stimulus. All other operations are triggered by visiting a URL.

In general, the quick action table headers should be easily portable to other tables. This is why all of the action menu methods went into the general sort_helper. Model-specific infos such as custom filter mappings (see next paragraph) are provided from the view when calling the component.

About column filter mappings: Some actions are not suitable for certain columns. This depends on the column itself and its position within the table. You cannot move the leftmost column further to the left. At the same time, you cannot filter by columns where we have no fitting filter for. For mapping existing filters to certain columns, I have added constant to `projects_helper.rb`. When applying the action menu to other tables, the models must provide their column-filter-mapping in a similar fashion.

# Merge checklist

- [x] Added/updated tests


ulferts:
I didn't do a proper review but rather only tried out the PR for a bit. I really enjoyed playing around with it cause the UX felt nicely direct. But while playing around, I stumbled over two things:

image

* The table headers are no longer ensured to be fully visible. E.g. "Version" is cut off. That used to be different before.
* There is no quick action on text columns. I get that they are not sortable but one could move or remove such a column via the quick actions.

EinLama:
Thanks for the feedback, Jens. I agree and have added a style fix as well as the option to use the action menu on non-sortable columns, too.

What's left now is to also allow the "mark as favorite" column to work with the action menu with the same reasoning. Currently looking into that.

app/helpers/sort_helper.rb
aaron-contreras, line 1: Excellent!

frontend/src/global_styles/primer/_overrides.sass
```diff
@@ -117,3 +117,9 @@ sub-header,
 // Todo: Remove once https://github.com/primer/view_components/pull/3087 is merged
 .FormControl-spacingWrapper
   row-gap: var(--stack-gap-normal)
+
+// When our generic table uses the quick action menu, column headers are cut off and no longer fully readable.
+// This fix ensures that column headers are always displayed. An example for such a table is the project list.
+.generic-table
+  .generic-table--action-menu-button
+    min-width: max-content !important
```
aaron-contreras, line 125: 👍

frontend/src/stimulus/controllers/dynamic/filter/filters-form.controller.ts
```diff
@@ -142,6 +142,11 @@ export default class FiltersFormController extends Controller {
     this.displayFiltersValue = !this.displayFiltersValue;
   }
 
+  showDisplayFilters() {
+    this.displayFiltersValue = true;
+    this.displayFiltersValueChanged();
```
aaron-contreras, line 147: Is this actually necessary? IIRC when these are declared Stimulus controller values, they are automatically observed and call the `valueNameValueChanged` without having to invoke it.

aaron-contreras (approved):
🌴 Great work. Left a question as I recall some of the behavior in a Stimulus controller might be duplicated. If not, feel free to merge with this approval

frontend/src/stimulus/controllers/dynamic/filter/filters-form.controller.ts
EinLama: I know what you mean. I had the same assumption initially and was surprised by this behavior as well.

To be sure, I tried commenting out the call to `showDisplayFilters` again -> the filter component is not toggled automatically when this line is missing.

EinLama:
Thank you 🎉

PR #16951, opened 2024-10-14
[57251] Fix mention notification email sometimes not being sent correctly on update

cbliard:
# Ticket

https://community.openproject.org/wp/57251

When a notification exists for a user for a work package journal for a reason other than :mentioned, then its `mail_alert_sent` value is `nil`.

On a subsequent update of the work package which mentions the user, the existing notification is updated. In this update, the `mail_alert_sent` value must be changed from `nil` to `false` if the reason is :mentioned. Without it, the immediate notification email for the mention would not be sent (which is the buggy behavior).

In `Notifications::CreateFromModelService`, there is some logic in place to create a new notification or update the existing one. In the `update_notification` logic, there is already some code to update `read_ian` value so that the in-app notification is visible again.

I added some logic next to it to also update `mail_alert_sent`. The value true is kept so that the email is sent only once.
# Merge checklist

- [x] Added/updated tests

### DIFF
--- a/app/services/notifications/create_from_model_service.rb
+++ b/app/services/notifications/create_from_model_service.rb
@@ -129,6 +129,7 @@ class Notifications::CreateFromModelService
     Notifications::UpdateService
       .new(model: existing_notification, user:, contract_class: EmptyContract)
       .call(read_ian: strategy.supports_ian?(reason) ? false : nil,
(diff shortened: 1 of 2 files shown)


oliverguenther (approved):
Great catch!

PR #16957, opened 2024-10-14
[58360] Disable "is read-only" checkbox in admin statuses when "is default" is checked

cbliard:
# Ticket

https://community.openproject.org/wp/58360
=> disable "is read-only" checkbox if "is default" checkbox is checked

https://community.openproject.org/wp/58411 
=> restore the enterprise banner on "new board" page

- The "Default value" checkbox must be visible for all types.
  - With caption: "New work packages are by default set to this type. They cannot be read-only."
- When the Default checkbox is checked, the Read-only checkbox must be unchecked and disabled.
- When the Read-only checkbox is enabled, the Default checkbox is still available (it's not disabled).

While adding a helper to check for enterprise banner presence, it revealed that [the enterprise banner was not shown on the new board page](https://community.openproject.org/wp/58411). Previous test helper was testing the presence, but not the visibility. It has been fixed in this same PR.

For the checkboxes: use a stimulus controller to dynamically enable/disable/uncheck the is-readonly checkbox when the is-default checkbox is clicked.

For the banner: use the correct invocation of the enterprise-banner and add some css to adapt the banner width to match the boards grid width.

# Merge checklist

- [x] Added/updated tests

### DIFF
--- a/frontend/src/stimulus/controllers/dynamic/admin/statuses.controller.ts
+++ b/frontend/src/stimulus/controllers/dynamic/admin/statuses.controller.ts
@@ -1,0 +1,56 @@
+/*
+ * -- copyright
+ * OpenProject is an open source project management software.
+ * Copyright (C) the OpenProject GmbH
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU General Public License version 3.
+ *
+ * OpenProject is a fork of ChiliProject, which is a fork of Redmine. The copyright follows:
+ * Copyright (C) 2006-2013 Jean-Philippe Lang
+ * Copyright (C) 2010-2013 the ChiliProject Team
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU General Public License
+ * as published by the Free Software Foundation; either version 2
+ * of the License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ *
+ * See COPYRIGHT and LICENSE files for more details.
+ * ++
+ */
+
+import { Controller } from '@hotwired/stimulus';
+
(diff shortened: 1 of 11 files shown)


frontend/src/stimulus/controllers/dynamic/admin/statuses.controller.ts
```diff
@@ -0,0 +1,56 @@
...
+    'isReadonlyCheckbox',
+  ];
+
+  declare readonly isDefaultCheckboxTarget:HTMLInputElement;
+  declare readonly isReadonlyCheckboxTarget:HTMLInputElement;
+
+  updateReadonlyCheckboxDisabledState() {
+    // do nothing if the readonly checkbox is restricted due to enterprise
+    // edition not being available
+    if (this.isReadonlyCheckboxTarget.dataset.restricted === 'true') {
```
ulferts, line 45: 💡 It would also be possible to use a [stimulus value](https://stimulus.hotwired.dev/reference/values) to pass in the restricted value explicitly and not via a somewhat nondescript data attribute. I don't know though, if it is possible to have stimulus value attributes defined in places other than where the controller is referenced. Because if it would be necessary to do it directly where the controller is referenced, then it would have to be done in two places (`new/edit.html.erb`). And if it would be necessary to do it in two places, it might not actually be an improvement.

spec/support/matchers/has_enterprise_banner.rb
```diff
@@ -0,0 +1,49 @@
...
+#
+# You should have received a copy of the GNU General Public License
+# along with this program; if not, write to the Free Software
+# Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+#
+# See COPYRIGHT and LICENSE files for more details.
+#++
+
+RSpec::Matchers.define :have_enterprise_banner do |**args|
+  include TestSelectorFinders
```
ulferts, line 32: Nice, thanks.

ulferts (approved):
Looks good and is free to merge.

I have one possible improvement on the code structure of which I don't know whether it actually turns out to be an improvement.

frontend/src/stimulus/controllers/dynamic/admin/statuses.controller.ts
cbliard: That's the thing: the value has to be in the element referencing the controller, and I did not want to duplicate it (as you guessed). That's what I learned when using it in the [progress tracking admin page](https://github.com/opf/openproject/blob/13277cf4bae73205426c94ae658d74838a12779c/app/views/admin/settings/progress_tracking/show.html.erb#L45) where the value of `Setting.work_package_done_ratio` is passed as a value.

PR #16961, opened 2024-10-15
Restore the form that was removed in 14.4

mereghost:
This restores the save button for [OP#58400](https://community.openproject.org/work_packages/58400)


HDinger (changes requested):
The original commit moved the PageHeader and the form inside of the partial. So simply restoring the form here, will result in doubled content.

Bildschirmfoto 2024-10-16 um 08 26 45
 
 I will push a change in a bit.

mereghost:
Thanks @HDinger ! It works with the DB plugin. =)

HDinger (approved):

Kharonus (approved):

PR #16972, opened 2024-10-16
[#57817] add deletion action to item action menu

Kharonus:
- https://community.openproject.org/work_packages/57817

# Ticket
[OP#57817](https://community.openproject.org/work_packages/57817)

- enable deletion of hierarchy items

- added dialog component to safe guard item deletion
- added controller actions and routes to open dialog and delete item


app/controllers/admin/custom_fields/hierarchy/items_controller.rb
```diff
@@ -63,6 +65,18 @@ def create
           respond_with_turbo_streams
         end
 
+        def destroy
+          # TODO: user persistence service
+          @hierarchy_item.destroy
```
mereghost: Since I merged the persistence service, I think you could already implement this change here.

mereghost (changes requested):
Just one small change, since we have the service merged already.

app/components/admin/custom_fields/hierarchy/item_component.html.erb
```diff
@@ -46,10 +46,14 @@ See COPYRIGHT and LICENSE files for more details.
...
       # Actions
       item_container.with_column do
-        render(Primer::Beta::IconButton.new(scheme: :default,
-                                            icon: "kebab-horizontal",
-                                            "aria-label": I18n.t("custom_fields.admin.items.more_actions")))
+        render(Primer::Alpha::ActionMenu.new(data: { test_selector: "op-hierarchy-item--action-menu" })) do |menu|
+          menu.with_show_button(icon: "kebab-horizontal",
+                                scheme: :invisible,
+                                "aria-label": I18n.t("custom_fields.admin.items.actions"))
+          delete_item(menu)
```
brunopagno: At first read I though the intention was to delete something, not to add something to the menu :shrug: 

Do we have a standard here or could we name the method something else?

brunopagno (approved):
Hey, from my point of view this looks good. I was able to understand things and test it locally so :heavy_check_mark: :tada:

Just left a comment if you want to overthink things a bit.

app/components/admin/custom_fields/hierarchy/item_component.html.erb
Kharonus: no standard, but we can definitely improve method name

app/controllers/admin/custom_fields/hierarchy/items_controller.rb
Kharonus: definitely, will merge dev in and update this.

brunopagno (approved):
I was happy before, I'm happier now :+1:

mereghost (approved):

PR #16974, opened 2024-10-16
Fix CostType redirect to index after deletion.

top4ek:
Fix report deletion redirect.

1. Go to /cost_reports
2. Select report or save new one. Browser is located to /cost_reports/:id
3. Delete report.
4. Browser is still located /cost_reports/:id
5. Click on Update button leads to 500 error as `@query` is not populated
  
Nullify ID after destroy.  Actually, i'm not sure if it correct way to fix it as I still don't get the way controller works completely :)

# Work Package

https://community.openproject.org/wp/58441


ulferts (approved):
Much obliged @top4ek

PR #16982, opened 2024-10-17
docs/user-guide/time-and-costs/progress-tracking/README.md: fix % Com…

levindu:
When no field is set and one of Work and Remaining Work is filled,  the % Complete will be 0%, not 100%.

When Work match Remaining Work value, (Work - Remaining Work) / Work should be 0.


cbliard (approved):
You're totally right! 
Thanks for taking the time to fix it.

PR #16983, opened 2024-10-17
Ensure that deleting the Custom Field removes the entire tree

mereghost:
Related WP: [OP#58343](https://community.openproject.org/projects/document-workflows-stream/work_packages/58343/activity)

Ensure the expected behaviour.


brunopagno (approved):
LGTM :tm:

Kharonus (approved):
nice catch

PR #16987, opened 2024-10-18
Bug/57526 migration 20230608151123 add validity period to journals is very slow

ulferts:
# Ticket

https://community.openproject.org/wp/57526
https://community.openproject.org/wp/58090

In cases with very large data sets in the journals table, especially with a lot of versions per journable, the former implementation of the migration would perform poorly. This was the case because a version of the journal was touched multiple times within the same transaction when fixing inconsistent timestamps on journals.

The migration now will now touch each journal only once at the most because it fixes the inconsistencies from the highest version backwards down to the first one. Additionally, a utility table is created that can be used within each step to create the connection between a journal and its succeeding one. That utitilty table is then also used when actually writing the `validity_period`.

A test is added to prevent regressions.

# Merge checklist

- [x] Added/updated tests

### DIFF
--- a/db/migrate/20230608151123_add_validity_period_to_journals.rb
+++ b/db/migrate/20230608151123_add_validity_period_to_journals.rb
@@ -4,8 +4,12 @@ class AddValidityPeriodToJournals < ActiveRecord::Migration[7.0]
 
     reversible do |direction|
       direction.up do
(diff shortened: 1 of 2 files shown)


machisuji (approved):
Great effort, thank you! 👍

PR #16988, opened 2024-10-18
[#58426] add feature spec for custom fields of type hierarchy

Kharonus:
# Ticket
[OP#58426](https://community.openproject.org/work_packages/58426)
also fixes: [OP#58466](https://community.openproject.org/work_packages/58466)

- UI and main use cases are covered by a feature spec

- add a couple of test selectors
- add page objects

### DIFF
--- a/app/components/admin/custom_fields/hierarchy/delete_item_dialog_component.html.erb
+++ b/app/components/admin/custom_fields/hierarchy/delete_item_dialog_component.html.erb
@@ -28,7 +28,9 @@ See COPYRIGHT and LICENSE files for more details.
 ++#%>
 
 <%=
-  render(Primer::Alpha::Dialog.new(id: DIALOG_ID, title: "Delete item")) do |dialog|
+  render(Primer::Alpha::Dialog.new(id: DIALOG_ID,
+                                   title: "Delete item",
+                                   data: { test_selector: TEST_SELECTOR })) do |dialog|
     dialog.with_header(variant: :large)
     dialog.with_body do
       "Are you sure you want to delete this item from the current hierarchy level?"
--- a/app/components/admin/custom_fields/hierarchy/items_component.html.erb
+++ b/app/components/admin/custom_fields/hierarchy/items_component.html.erb
@@ -32,7 +32,10 @@ See COPYRIGHT and LICENSE files for more details.
     flex_layout do |container|
       if items.empty? && !show_new_item_form?
         container.with_row(mb: 3) do
-          render Primer::Beta::Blankslate.new(border: true) do |component|
+          render Primer::Beta::Blankslate.new(
+            border: true,
+            test_selector: "op-custom-fields--hierarchy-items-blankslate"
+          ) do |component|
             component.with_visual_icon(icon: "list-ordered")
             component.with_heading(tag: :h3).with_content(I18n.t("custom_fields.admin.items.blankslate.title"))
             component.with_description { I18n.t("custom_fields.admin.items.blankslate.description") }
--- a/app/components/admin/custom_fields/hierarchy/new_item_form_component.html.erb
+++ b/app/components/admin/custom_fields/hierarchy/new_item_form_component.html.erb
@@ -31,6 +31,7 @@ See COPYRIGHT and LICENSE files for more details.
   primer_form_with(
     url: custom_field_items_path(@custom_field),
     method: :post,
+    data: { test_selector: "op-custom-fields--new-item-form" }
   ) do |f|
     render(CustomFields::Hierarchy::NewItemForm.new(f, custom_field: @custom_field, label: @label, short: @short))
   end
--- a/app/components/custom_fields/details_component.html.erb
+++ b/app/components/custom_fields/details_component.html.erb
@@ -1,9 +1,14 @@
 <%=
   component_wrapper do
     flex_layout do |content|
-      content.with_row(mb: 3) do
-        render Primer::Alpha::Banner.new(scheme: :default, icon: :info, dismiss_scheme: :hide) do
-          I18n.t("custom_fields.admin.notice.remember_items_and_projects")
+      if has_no_items_or_projects?
+        content.with_row(mb: 3) do
+          render Primer::Alpha::Banner.new(scheme: :default,
+                                           icon: :info,
+                                           dismiss_scheme: :hide,
+                                           test_selector: "op-custom-fields--new-hierarchy-banner") do
+            I18n.t("custom_fields.admin.notice.remember_items_and_projects")
+          end
         end
       end
 
--- a/app/views/custom_fields/_form.html.erb
+++ b/app/views/custom_fields/_form.html.erb
@@ -37,7 +37,10 @@ See COPYRIGHT and LICENSE files for more details.
   data-admin--custom-fields-format-config-value="<%= OpenProject::CustomFieldFormatDependent.stimulus_config %>"
 >
   <div class="form--field -required" id="custom_field_name_attributes">
-    <%= f.text_field :name, required: true, container_class: '-middle' %>
+    <%= f.text_field :name,
+                     required: true,
+                     container_class: "-middle",
+                     "data-test-selector": "op-custom-fields--new-custom-field-name" %>
   </div>
   <% if @custom_field.type == 'ProjectCustomField' %>
     <div class="form--field -required">
--- a/spec/features/custom_fields/hierarchy_custom_field_spec.rb
+++ b/spec/features/custom_fields/hierarchy_custom_field_spec.rb
@@ -1,0 +1,131 @@
+# frozen_string_literal: true
+
+#-- copyright
+# OpenProject is an open source project management software.
+# Copyright (C) the OpenProject GmbH
+#
+# This program is free software; you can redistribute it and/or
+# modify it under the terms of the GNU General Public License version 3.
+#
+# OpenProject is a fork of ChiliProject, which is a fork of Redmine. The copyright follows:
+# Copyright (C) 2006-2013 Jean-Philippe Lang
+# Copyright (C) 2010-2013 the ChiliProject Team
+#
+# This program is free software; you can redistribute it and/or
+# modify it under the terms of the GNU General Public License
+# as published by the Free Software Foundation; either version 2
+# of the License, or (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program; if not, write to the Free Software
+# Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+#
+# See COPYRIGHT and LICENSE files for more details.
+#++
+
+require "spec_helper"
+require "support/pages/custom_fields/hierarchy_page"
+
+RSpec.describe "custom fields of type hierarchy", :js, :with_cuprite do
+  let(:user) { create(:admin) }
+  let(:custom_field_index_page) { Pages::CustomFields::IndexPage.new }
+  let(:new_custom_field_page) { Pages::CustomFields::NewPage.new }
+  let(:hierarchy_page) { Pages::CustomFields::HierarchyPage.new }
+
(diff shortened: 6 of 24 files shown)


spec/features/custom_fields/hierarchy_custom_field_spec.rb
```diff
@@ -0,0 +1,131 @@
...
+#
+# You should have received a copy of the GNU General Public License
+# along with this program; if not, write to the Free Software
+# Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+#
+# See COPYRIGHT and LICENSE files for more details.
+#++
+
+require "spec_helper"
+require "support/pages/custom_fields/hierarchy_page"
```
mereghost: 🟢 The support folder is already required by `spec_helper`, this shouldn't be needed.

app/components/admin/custom_fields/hierarchy/delete_item_dialog_component.html.erb
```diff
@@ -28,7 +28,9 @@ See COPYRIGHT and LICENSE files for more details.
 ++#%>
 
 <%=
-  render(Primer::Alpha::Dialog.new(id: DIALOG_ID, title: "Delete item")) do |dialog|
+  render(Primer::Alpha::Dialog.new(id: DIALOG_ID,
+                                   title: "Delete item",
+                                   data: { test_selector: TEST_SELECTOR })) do |dialog|
```
mereghost, line 33: ❓ Does this end up on the production rendered HTML?

app/components/admin/custom_fields/hierarchy/items_component.html.erb
```diff
@@ -32,7 +32,10 @@ See COPYRIGHT and LICENSE files for more details.
     flex_layout do |container|
       if items.empty? && !show_new_item_form?
         container.with_row(mb: 3) do
-          render Primer::Beta::Blankslate.new(border: true) do |component|
+          render Primer::Beta::Blankslate.new(
+            border: true,
+            test_selector: "op-custom-fields--hierarchy-items-blankslate"
```
mereghost, line 37: 🟡 Any reason this isn't a constant like the other one?

app/components/admin/custom_fields/hierarchy/new_item_form_component.html.erb
```diff
@@ -31,6 +31,7 @@ See COPYRIGHT and LICENSE files for more details.
   primer_form_with(
     url: custom_field_items_path(@custom_field),
     method: :post,
+    data: { test_selector: "op-custom-fields--new-item-form" }
```
mereghost, line 34: 🟡 Same question as above

app/components/custom_fields/details_component.html.erb
```diff
@@ -1,9 +1,14 @@
...
     flex_layout do |content|
-      content.with_row(mb: 3) do
-        render Primer::Alpha::Banner.new(scheme: :default, icon: :info, dismiss_scheme: :hide) do
-          I18n.t("custom_fields.admin.notice.remember_items_and_projects")
+      if has_no_items_or_projects?
+        content.with_row(mb: 3) do
+          render Primer::Alpha::Banner.new(scheme: :default,
+                                           icon: :info,
+                                           dismiss_scheme: :hide,
+                                           test_selector: "op-custom-fields--new-hierarchy-banner") do
```
mereghost, line 9: 🟡 Same same.

app/components/custom_fields/details_component.rb
```diff
@@ -32,5 +32,13 @@ module CustomFields
   class DetailsComponent < ApplicationComponent
     include OpPrimer::ComponentHelpers
     include OpTurbo::Streamable
+
+    alias_method :custom_field, :model
+
+    def has_no_items_or_projects?
```
mereghost, line 38: 🟡 Can we leverage the `render?` method

app/views/custom_fields/_form.html.erb
```diff
@@ -37,7 +37,10 @@ See COPYRIGHT and LICENSE files for more details.
   data-admin--custom-fields-format-config-value="<%= OpenProject::CustomFieldFormatDependent.stimulus_config %>"
 >
   <div class="form--field -required" id="custom_field_name_attributes">
-    <%= f.text_field :name, required: true, container_class: '-middle' %>
+    <%= f.text_field :name,
+                     required: true,
+                     container_class: "-middle",
+                     "data-test-selector": "op-custom-fields--new-custom-field-name" %>
```
mereghost, line 43: 🟡 More non-constant selectors.

spec/features/custom_fields/hierarchy_custom_field_spec.rb
```diff
@@ -0,0 +1,131 @@
...
+
+RSpec.describe "custom fields of type hierarchy", :js, :with_cuprite do
+  let(:user) { create(:admin) }
+  let(:custom_field_index_page) { Pages::CustomFields::IndexPage.new }
+  let(:new_custom_field_page) { Pages::CustomFields::NewPage.new }
+  let(:hierarchy_page) { Pages::CustomFields::HierarchyPage.new }
+
+  it "lets you create, update and delete a custom field of type hierarchy",
```
mereghost, line 39: 🔴 Wouldn't be better to split those cases into separate tests?

spec/support/pages/custom_fields/hierarchy_page.rb
```diff
@@ -0,0 +1,117 @@
...
+
+      def switch_tab(tab)
+        @tab = tab.downcase
+
+        within_test_selector("custom_field_detail_header") do
+          click_on "Items"
+        end
+      end
+
+      def expect_empty_items_banner(visible:)
```
mereghost: 😕 I have mixed feeling about hidden assertions in helpers as makes the tests kinda opaque. That being said, it is just me so don't bother too much with it.

mereghost (changes requested):
Bunch of questions, rants and of course, no suggestions. 🤣

app/components/admin/custom_fields/hierarchy/delete_item_dialog_component.html.erb
Kharonus: I need to relay the question, as I remember I asked the same thing some time ago :D

spec/features/custom_fields/hierarchy_custom_field_spec.rb
Kharonus: No, IMHO that's exactly the opposite of what we should do with e2e tests.

Reasoning:
The ramp up of the test suite takes ages, compared to the actual test run. This is why I wrote the work with the custom field like a story. You start by creating it, do a lot of nasty stuff to it, and then you end by deleting it. Also, an expensive test like this is meant to test UI and integration, not all of the use cases. Which is why I kept some things out for now.

spec/support/pages/custom_fields/hierarchy_page.rb
Kharonus: My goal was to make the test itself easily readable without bothering to much about capybara syntax. Things like `click_on` and `fills_in` are already doing that quite nicely, but the stuff about test_selectors add nothing to the readability imho.

app/components/admin/custom_fields/hierarchy/delete_item_dialog_component.html.erb
Kharonus: When looking at community, I still see the html attributes `data-test-selector`. So I guess, the answer is yes, it is part of production rendered HTML.

Does this lead to another issue, you had in mind?

mereghost: Nope. I'm OK with it existing in production, I was just wondering if it got removed.

Some testing stuff is bound to escape, like the oil level dipstick on car engines. =)

spec/features/custom_fields/hierarchy_custom_field_spec.rb
mereghost: I'd separate them to avoid eventual state leaks to change the behaviour. 

I tend to see feature specs as user stories with clear intent in each test and acceptance criteria for it. We may need to involve others in this discussion.

@apfohl @brunopagno WDYT?

spec/support/pages/custom_fields/hierarchy_page.rb
mereghost: `click_on` and `fill_in` represent user actions.

This sets an expectation/assertion making it completely invisible to the test as you scan for `expect(thing).to eq(zeug)` making parsing what is being tested pretty hard. =/

Kharonus: Is it then a general opposition of using POs? They are meant to encapsulate these kind of things into more readable things.

spec/features/custom_fields/hierarchy_custom_field_spec.rb
brunopagno: In the past I was bitten by having a single/few tests because not all errors would be visible right away. Then you get to the situation where you fix something and it explodes again further into the test.

I understand the argument against, though, that tests might take much longer.

I guess it's a question if we can "pay the price" of longer running tests.

spec/support/pages/custom_fields/hierarchy_page.rb
mereghost: Nah, I'm cool with PO's... even with managers. :P

Seriously, POs are ok the issue is the hidden assertion. When parsing specs, I search for `expect(something).to something` and not `page.expect_something`.

I classify this as worse than shared specs used only in one place.

Kharonus: okay, understood. I'll recheck the code with having this kind of glasses on.

spec/features/custom_fields/hierarchy_custom_field_spec.rb
apfohl: I see also two sides. The one side don't wanting an even longer running test suite and the other side wanting to have easier maintainable specs. It seems like you can't have both at the moment. Honestly I would say we should go the way of doing it like @Kharonus. Define the domain story (you could even draw it) and play that thing through. Maybe we could use `# region REGION_NAME` `# endregion` to visually highlight the steps. The IDE allows to collapse them.

In general I would not delay on feature spec development to much. Invest more into automated domain model unit testing.

spec/support/pages/custom_fields/hierarchy_page.rb
apfohl: Maybe we can extend the `expect` language with new custom matchers. Not sure if that is easily doable. I'm thinking of some kind of fluent syntax extension:

```ruby
expect(page).when(:visible).to have(BANNER).when_not(:visible).not_to have(BANNER)
```

Or different.

spec/features/custom_fields/hierarchy_custom_field_spec.rb
Kharonus: Is separating the test into clearly distinct `regions` something you would approve @mereghost ?

app/components/custom_fields/details_component.rb
Kharonus: I'd say no, as I'd need to create a separate component just for the banner.

app/components/admin/custom_fields/hierarchy/items_component.html.erb
Kharonus: Honestly there is no good reason to extract those inline selectors to variables. I did it for the dialog, because it was already in the template I copied.

But those selectors are unique. They are not reused.

app/components/admin/custom_fields/hierarchy/new_item_form_component.html.erb
Kharonus: see above

app/components/custom_fields/details_component.html.erb
Kharonus: see above

app/views/custom_fields/_form.html.erb
Kharonus: see above

spec/features/custom_fields/hierarchy_custom_field_spec.rb
mereghost: I'd say it would alleviate the issue at least. 👍🏼

spec/features/custom_fields/hierarchy_custom_field_spec.rb
Kharonus: done

mereghost (approved):

PR #17001, opened 2024-10-21
Enable item editing of hierarchical custom fields

brunopagno:
# Ticket

https://community.openproject.org/wp/57818

A button to enable editting items.
image

And a form for editting items.
image

I needed to make the Item a component_wrapper so we can update it independently of the rest (and to show validation errors)

# Merge checklist

- [ ] Added/updated tests

### DIFF
--- a/app/components/admin/custom_fields/hierarchy/item_form_component.rb
+++ b/app/components/admin/custom_fields/hierarchy/item_form_component.rb
@@ -29,14 +29,17 @@
 module Admin
   module CustomFields
     module Hierarchy
-      class NewItemFormComponent < ApplicationComponent
+      class ItemFormComponent < ApplicationComponent
         include OpTurbo::Streamable
 
-        def initialize(custom_field:, label: nil, short: nil)
+        def initialize(custom_field:, hierarchy_item:, url:, method:, label: nil, short: nil)
           super
           @custom_field = custom_field
-          @label = label
-          @short = short
+          @hierarchy_item = hierarchy_item
+          @url = url
+          @method = method
+          @label = label || @hierarchy_item.label
+          @short = short || @hierarchy_item.short
         end
 
         def items_path
--- a/app/controllers/admin/custom_fields/hierarchy/items_controller.rb
+++ b/app/controllers/admin/custom_fields/hierarchy/items_controller.rb
@@ -41,7 +41,7 @@ module Admin
 
         before_action :require_admin
         before_action :find_model_object, except: %i[destroy deletion_dialog]
-        before_action :find_custom_field_and_item, only: %i[destroy deletion_dialog]
+        before_action :find_custom_field_and_item, only: %i[destroy deletion_dialog edit update]
 
         menu_item :custom_fields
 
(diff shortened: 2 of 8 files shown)


app/components/admin/custom_fields/hierarchy/edit_item_form_component.html.erb
Kharonus, line 1: 🔴 very early feedback:

I think you should not write a new form component. The only difference is the method, this one can be get by any other indicator.

app/controllers/admin/custom_fields/hierarchy/items_controller.rb
```diff
@@ -53,13 +53,42 @@ def new
...
+
+          respond_with_turbo_streams
+        end
+
+        def update
+          ::CustomFields::Hierarchy::HierarchicalItemService
+            .new
+            .update_item(item: @hierarchy_item, label: item_input[:label], short: item_input[:short])
+            .either(
+              ->(_) {
```
Kharonus: 🟡 if we have multi-line blocks, IMHO we should use `do end`, not curlies

Kharonus (changes requested):

app/components/admin/custom_fields/hierarchy/edit_item_form_component.html.erb
brunopagno: Not only the method, it does also render a different form as well.

I made the `edit` explicitly separate of the `new` because I saw some examples that kept them separate on the codebase.

But I don't mind explore using the same form for both, though. I don't really have strong opinions on this, yet

app/controllers/admin/custom_fields/hierarchy/items_controller.rb
brunopagno: Agreed. I usually rely on rubocop magic to complain about these. But this is an easy and happy fix :)

mereghost: @Kharonus you know that this can change semantics too? Not in this case but overall.

app/forms/custom_fields/hierarchy/item_form.rb
```diff
@@ -62,21 +62,19 @@ class NewItemForm < ApplicationForm
...
       private
 
       def validation_message_for(attribute)
-        @custom_field
-          .errors
-          .messages_for(attribute)
-          .to_sentence
-          .presence
+        @hierarchy_item.errors.messages_for(attribute).to_sentence.presence ||
+          @custom_field.errors.messages_for(attribute).to_sentence.presence
```
brunopagno, line 77: I think there's a bit of semantics here to decide if we can operate only on top of `hierarchy_item` instead of adding information to the `custom_field`. But for now I kept both.

app/components/admin/custom_fields/hierarchy/item_form_component.rb
```diff
@@ -29,14 +29,17 @@
 module Admin
   module CustomFields
     module Hierarchy
-      class NewItemFormComponent < ApplicationComponent
+      class ItemFormComponent < ApplicationComponent
         include OpTurbo::Streamable
 
-        def initialize(custom_field:, label: nil, short: nil)
+        def initialize(custom_field:, hierarchy_item:, url:, method:, label: nil, short: nil)
```
brunopagno, line 35: Parameters look dirty, but at the same time seem necessary. I'm open to ideas on how to do this.

app/forms/custom_fields/hierarchy/item_form.rb
Kharonus: I agree, especially after seeing, what Marcello does.

In the end, this form should work fully without the custom field, and only with the item.

Kharonus (approved):
LGTM

it does the job of adding the edit form. The improvements of e.g. removing `@custom_field` from the item form will then be part of marcellos PR with the rework.

PR #17017, opened 2024-10-22
Implement drill up/down the hierarchy field

mereghost:
## What?

We want to be able to navigate to (and back to) the different branches of the hierarchy custom field and, of course, do the normal CRUD operations over the branch child nodes.

## How?

This PR makes a lot of changes on the way we normally have been doing this, by going in a more `turbo_frames` first approach.

First it makes the turbo frame use the `morph` refresh mechanism and also the action `advance` so that the browser location and history are updated on every navigation. This required some tweaks as the `Primer::Breadcrumbs` don't allow us to pass extra attributes to its hyperlinks (besides target and href).

The result is a clean approach with a more standard RoR view approach and following the hotwire guidelines.

There's a couple of interesting pieces that I learned around this: for turbo frame responses a specific layout needs to be used. This ensures a minimalist response and focuses on just the frame being acted upon.

## Known issues

While you can have multiple items in edit mode and update them separately, the moment you click on new, the entire state resets. This is due to the way we render the form inside the border-box footer needing the re-render of the entire component.

There's no easy way around it due to the usage of the primer components. We would need to be able to somehow generate just a HTML snipped of a single row of a `BorderBox` so that we could append/prepend/update it.

This will probably become very relevant the moment we try to make the **Insert Above** or **Insert Below** actions.

### Extra Notes
Special thanks for @akabiru that put up with my silly and ranty questions and to @brunopagno that helped me clean this mess. <3

### DIFF
--- a/app/components/admin/custom_fields/hierarchy/item_component.html.erb
+++ b/app/components/admin/custom_fields/hierarchy/item_component.html.erb
@@ -39,7 +39,7 @@ See COPYRIGHT and LICENSE files for more details.
       flex_layout(align_items: :center, justify_content: :space_between, test_selector: "op-custom-fields--hierarchy-item") do |item_container|
         item_container.with_column(flex_layout: true) do |item_information|
           item_information.with_column(mr: 2) do
-            render(Primer::Beta::Link.new(href: custom_field_item_path(model.root.custom_field_id, model), underline: false)) do
+            render(Primer::Beta::Link.new(href: custom_field_item_path(@root.custom_field_id, model), underline: false)) do
               render(Primer::Beta::Text.new(font_weight: :bold)) { model.label }
             end
           end
--- a/app/components/admin/custom_fields/hierarchy/item_component.html.erb
+++ b/app/components/admin/custom_fields/hierarchy/item_component.html.erb
@@ -42,8 +42,10 @@ See COPYRIGHT and LICENSE files for more details.
       flex_layout(align_items: :center, justify_content: :space_between) do |item_container|
         item_container.with_column(flex_layout: true) do |item_information|
           item_information.with_column(mr: 2) do
-            render(Primer::Beta::Text.new(font_weight: :bold)) do
-              @hierarchy_item.label
+            render(Primer::Beta::Link.new(href: custom_field_item_path(@hierarchy_item.root, @hierarchy_item), underline: false)) do
+              render(Primer::Beta::Text.new(font_weight: :bold)) do
+                @hierarchy_item.label
+              end
             end
           end
 
@@ -52,16 +54,21 @@ See COPYRIGHT and LICENSE files for more details.
               render(Primer::Beta::Text.new(color: :subtle)) { short_text }
             end
           end
+        end
 
-          # Actions
-          item_container.with_column do
-            render(Primer::Alpha::ActionMenu.new(data: { test_selector: "op-hierarchy-item--action-menu" })) do |menu|
-              menu.with_show_button(icon: "kebab-horizontal",
-                                    scheme: :invisible,
-                                    "aria-label": I18n.t("custom_fields.admin.items.actions"))
-              edit_action_item(menu)
-              deletion_action_item(menu)
-            end
+        if @hierarchy_item.children.exists?
+          item_information.with_column(mr: 2) do
+            render(Primer::Beta::Text.new) { children_count }
+          end
+        end
+
+        item_container.with_column do
+          render(Primer::Alpha::ActionMenu.new(data: { test_selector: "op-hierarchy-item--action-menu" })) do |menu|
+            menu.with_show_button(icon: "kebab-horizontal",
+                                  scheme: :invisible,
+                                  "aria-label": I18n.t("custom_fields.admin.items.actions"))
+            edit_action_item(menu)
+            deletion_action_item(menu)
           end
         end
       end
--- a/app/components/admin/custom_fields/hierarchy/item_component.html.erb
+++ b/app/components/admin/custom_fields/hierarchy/item_component.html.erb
@@ -28,42 +28,37 @@ See COPYRIGHT and LICENSE files for more details.
 ++#%>
 
 <%=
-  component_wrapper(data: { test_selector: "op-custom-fields--hierarchy-item" }) do
+  component_wrapper(tag: "turbo-frame", refresh: :morph) do
     if show_edit_form?
       render Admin::CustomFields::Hierarchy::ItemFormComponent.new(
-        custom_field: @custom_field,
-        hierarchy_item: @hierarchy_item,
-        url: custom_field_item_path(@custom_field, @hierarchy_item),
-        method: :put,
-        label: @edit_item_form_data.fetch(:label, nil),
-        short: @edit_item_form_data.fetch(:short, nil)
+        target_item: model,
+        url: custom_field_item_path(@root.custom_field_id, model),
+        method: :put
       )
     else
-      flex_layout(align_items: :center, justify_content: :space_between) do |item_container|
+      flex_layout(align_items: :center, justify_content: :space_between, test_selector: "op-custom-fields--hierarchy-item") do |item_container|
         item_container.with_column(flex_layout: true) do |item_information|
           item_information.with_column(mr: 2) do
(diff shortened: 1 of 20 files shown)


app/components/admin/custom_fields/hierarchy/item_component.rb
```diff
@@ -45,6 +45,10 @@ def short_text
           "(#{@hierarchy_item.short})"
         end
 
+        def children_count
+          pluralize(@hierarchy_item.children.count, "sub-item")
```
Kharonus: 🟡 of course in the end we need i18n here to pluralize this stuff. You know the patterns?

app/views/admin/custom_fields/hierarchy/items/index.html.erb
```diff
@@ -31,4 +31,6 @@ See COPYRIGHT and LICENSE files for more details.
 
 <%= render(Admin::CustomFields::EditFormHeaderComponent.new(custom_field: @custom_field, selected: :items)) %>
 
-<%= render(Admin::CustomFields::Hierarchy::ItemsComponent.new(custom_field: @custom_field)) %>
+<%= turbo_frame_tag :hierarchy_list, data: { turbo_action: :advance } do %>
```
mereghost: `advance` needed to be added here due to the `Breadcrumbs` not allowing us to pass down tags to the `a` elements.

config/locales/en.yml
```diff
@@ -254,6 +254,11 @@ en:
           short: "Short name"
       notice:
         remember_items_and_projects: "Remember to set items and projects in the respective tabs for this custom field."
+      hierarchy:
+        subitems:
+          zero: no sub-items
```
Kharonus, line 259: 🟡 "zero" is currently a dead locale key, as those are not rendered. I'd rather remove it and add it, once we need this text.

app/components/admin/custom_fields/hierarchy/items_component.rb
```diff
@@ -35,20 +35,44 @@ class ItemsComponent < ApplicationComponent
...
+        def initialize(item:, new_item_form_data: { show: false })
+          super(item)
           @new_item_form_data = new_item_form_data
         end
 
-        def items
-          # TODO: This must be context aware (breadcrumbs)
-          @custom_field.hierarchy_root.children
+        def item_header
+          render(Primer::Beta::Breadcrumbs.new) do |loaf|
```
Kharonus, line 54: ❤️ perfect naming of variables

app/controllers/admin/custom_fields/hierarchy/items_controller.rb
```diff
@@ -73,45 +85,43 @@ def destroy
...
-          update_via_turbo_stream(component: ItemsComponent.new(custom_field: @custom_field, new_item_form_data:),
-                                  status: :unprocessable_entity)
+          update_via_turbo_stream(
+            component: ItemsComponent.new(item: @hierarchy_item, new_item_form_data:), status: :unprocessable_entity
+          )
         end
 
-        def find_model_object(object_id = :custom_field_id)
-          super
+        def find_model_object
```
Kharonus, line 133: 🟡 if we do not call super, lets maybe rename the method to `find_custom_field`.

app/services/custom_fields/hierarchy/hierarchical_item_service.rb
```diff
@@ -84,7 +84,7 @@ def delete_branch(item:)
       # @param item [CustomField::Hierarchy::Item] the parent of the node
       # @return [Success(Array<CustomField::Hierarchy::Item>)]
       def get_branch(item:)
-        Success(item.ancestors.reverse)
+        Success(item.self_and_ancestors.reverse)
```
Kharonus, line 87: 🟢 nice

Kharonus (commented):
some first feedback

app/controllers/admin/custom_fields/hierarchy/items_controller.rb
mereghost: This leads to other errors. This dude here does a bunch of stuff for the sidebars etc.

config/locales/en.yml
mereghost: This is just protection against weird cases. :P

app/components/admin/custom_fields/hierarchy/items_component.html.erb
```diff
@@ -28,42 +28,33 @@ See COPYRIGHT and LICENSE files for more details.
 ++#%>
 
 <%=
-  component_wrapper(tag: "turbo-frame") do
+  component_wrapper(tag: 'turbo-frame', refresh: :morph, data: { turbo_action: :advance }) do
```
brunopagno: This is totally out of context, but it bothers me a bit that our linter does not lint the code in `.html.erb` files. Notice here the single quotes~

I feel the urge to fix this :sweat_smile:

app/components/admin/custom_fields/hierarchy/item_component.rb
```diff
@@ -35,31 +35,33 @@ class ItemComponent < ApplicationComponent
         include OpTurbo::Streamable
         include OpPrimer::ComponentHelpers
 
-        def initialize(custom_field:, hierarchy_item:, edit_item_form_data: { show: false })
-          super
-          @custom_field = custom_field
-          @hierarchy_item = hierarchy_item
+        def initialize(item:, edit_item_form_data: { show: false })
```
brunopagno: I am thinking we could also change the `edit_item_form_data` parameter to be just a boolean, since we don't pass down other attributes along anymore.

e.g.
```ruby
def initialize(item:, show_edit_form: false)
```

Possibly we can do that at a separate moment, though

app/components/admin/custom_fields/hierarchy/items_component.html.erb
mereghost: There's [erb_lint](https://github.com/Shopify/erb_lint) that Shopify ressurected recently. Might be worth taking a look at.

app/components/admin/custom_fields/hierarchy/item_component.html.erb
```diff
@@ -28,42 +28,44 @@ See COPYRIGHT and LICENSE files for more details.
...
+        method: :put
       )
     else
-      flex_layout(align_items: :center, justify_content: :space_between) do |item_container|
+      flex_layout(align_items: :center, justify_content: :space_between, test_selector: "op-custom-fields--hierarchy-item") do |item_container|
         item_container.with_column(flex_layout: true) do |item_information|
           item_information.with_column(mr: 2) do
-            render(Primer::Beta::Text.new(font_weight: :bold)) do
-              @hierarchy_item.label
+            render(Primer::Beta::Link.new(href: custom_field_item_path(model.root.custom_field_id, model), underline: false)) do
```
Kharonus: 🟡 minor: can we use `@root` instead of `model.root` if we expose it?

app/components/admin/custom_fields/hierarchy/items_component.rb
```diff
@@ -35,19 +35,42 @@ class ItemsComponent < ApplicationComponent
...
+
+        def initialize(item:, new_item: nil)
+          super(item)
+          @new_item = new_item
         end
 
-        def items
-          # TODO: This must be context aware (breadcrumbs)
-          @custom_field.hierarchy_root.children
+        def branch_root = model
```
Kharonus: 🟡 this one is unused, isn't it? Can we remove it?

app/contracts/custom_fields/hierarchy/update_item_contract.rb
```diff
@@ -45,12 +45,7 @@ class UpdateItemContract < Dry::Validation::Contract
       rule(:label) do
         next unless key?
 
-        label_already_exists = CustomField::Hierarchy::Item
-          .where(parent_id: values[:item].parent_id, label: value)
-          .where.not(id: values[:item].id)
-          .exists?
-
-        if label_already_exists
+        if values[:item].siblings.where(label: value).any?
```
Kharonus, line 48: 🎉 nice

Kharonus (approved):
Nice one, changes are way fewer then I'd expected.

Only have two non-blocking issues.

Also I added [this bug](https://community.openproject.org/wp/58700) for the left margin. It seems a rather easy fix in CSS, but I won't inject it in your PR, but simply add a small bugfix after it.

PR #17021, opened 2024-10-22
Bug/58555 reminder mail not sent in case more than 15 notifications are to be reported about

ulferts:
# Ticket

https://community.openproject.org/wp/58555

The notification_path is not defined in the context of the template being rendered. It will thus result in an hard error and this means that the mail is not sent at all. I haven't been able to figure out why the path is not defined but it needs to be the `notifications_url` in any case since the full url needs to be available as the link is followed from the mail client.

A second thing that was changed is the concurrency key although no errors have been reported about that. But with the way it was defined before, only one of multiple users sharing the same name could have a `Mail::ReminderJob` scheduled at the same time. The reason for this is that sometimes the user object is passed in. On it, `to_s` will be called to determine the concurrency_key. If two people share the same name, then adding the job for the second user will be prevented. This is changed by using the id of the user object instead. 

# Merge checklist

- [x] Added/updated tests


machisuji (approved):
👍

PR #17027, opened 2024-10-23
[58467] Changed add hierarchical item behavior

apfohl:
# Ticket
https://community.openproject.org/projects/document-workflows-stream/work_packages/58467/github

When adding a new item in the hierarchy the item form will stay open for the next entry.

I added the additional rendering of the item form for the turbo stream response to `admin/custom_fields/hierarchy/items#create`.

### DIFF
--- a/app/controllers/admin/custom_fields/hierarchy/items_controller.rb
+++ b/app/controllers/admin/custom_fields/hierarchy/items_controller.rb
@@ -58,7 +58,10 @@ module Admin
             .new
             .insert_item(**item_input)
             .either(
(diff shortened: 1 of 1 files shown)


mereghost:
Won't this cause a double send of the turbo-frame?

apfohl:
> Won't this cause a double send of the turbo-frame?

Why? It only adds the form to the response.

apfohl:
@mereghost Ah, now I see what you mean lol.

mereghost (approved):

PR #17045, opened 2024-10-25
[58256] Header and sidebar menu should be hidden while pressing zen mode button 

bsatarnejad:
# Ticket
https://community.openproject.org/projects/openproject/work_packages/58256/activity

Update core with new version of primer view components. Listen to the toggledZenMode event and set/remove zen-mode class on body element.


frontend/src/app/core/setup/globals/global-listeners.ts
```diff
@@ -84,6 +84,17 @@ export function initializeGlobalListeners():void {
...
     });
 
+  // Listen for 'zenModeToggled' event to toggle Zen Mode styling on the body.
+  // Adds 'zen-mode' class if active; removes it if not.
+  window.addEventListener('zenModeToggled', (event:CustomEvent) => {
+    const isActive = event.detail.active;
+    if (isActive) {
+      document.body.classList.add('zen-mode');
+    } else {
+      document.body.classList.remove('zen-mode');
```
ulferts: 💡 You can simplify the whole method by using [toggle](https://developer.mozilla.org/en-US/docs/Web/API/DOMTokenList/toggle)

```
window.addEventListener('zenModeToggled', (event:CustomEvent) => {
  document.body.classList.toggle('zen-mode', event.detail.active)
}
```

bsatarnejad: Hi @ulferts thanks for good hintwise, much better now 👍

ulferts (approved):
👍

PR #17062, opened 2024-10-28
Bug/58656 undesired auto scroll to latest emoji reaction

jjabari-op:
[58656](https://community.openproject.org/work_packages/58656)

The describe issue occurs because the browser performs a browser-native auto-scroll to the reaction button, which was in focus after the update stream gets applied. The reaction button is in focus whenever a user uses the overlay to select an emoji. This seems to be the default behavior of Primer.

My solution was to defocus all reaction buttons before performing the update stream request.


akabiru (approved):
Nice catch! Tested and working well 👍🏾

PR #17067, opened 2024-10-28
Fix small issues in the anonymize-sql-dump script

cbliard:
- Reuse AdminUserSeeder to create the admin user. Previous code did not set a valid email, which prevented the user from being saved (That's why language could not be set).
- drop the db before creating it, to avoid "database already exists" errors.
- Remove owner from ALTER statements in dump file, avoiding "ERROR: must be able to SET ROLE " errors when loading the sql file.


cbliard:
Failing tests are flickering ones. They pass locally as well as on other branches.
Merging.

PR #17070, opened 2024-10-28
Bug/57107 mobile tap twice on comment input to start typing

jjabari-op:
[57107](https://community.openproject.org/projects/communicator-stream/work_packages/57107)


jjabari-op:
The issue still not seem to be resolved, but I'm closing this PR as it got stale

lindenthal:
Hi @jjabari-op, is there a specific reason you didn't finished this?

wielinde:
@lindenthal We don't have a solution for that problem. We got a solution that worked for Android but not for iOS Safari.

PR #17075, opened 2024-10-29
[#58520] add an endpoint to fetch the hierarchy

Kharonus:
# Ticket
[OP#58520](https://community.openproject.org/work_packages/58520)

- we need an API endpoint to fetch the hierarchy structure of a custom field of type hierarchy
- the tree is represented as a depth-first flat list

- adding an endpoint GET /custom_fields/:id/items
- adding an endpoint GET /custom_field_items/:id (needed for the self links)
- adding API specification for both
- adding parameter validation contract for the tree structure
- amended legacy tests

## Hint for review

This review is easily separable into 

a) the API specification files, located under `/docs`
b) the code, in `/app`, `/lib/api`, and `/spec`

those could be separate PRs, but sadly I have those changes on the same commits.

### DIFF
--- a/docs/api/apiv3/components/examples/hierarchy_item_collection_filtered_response.yml
+++ b/docs/api/apiv3/components/examples/hierarchy_item_collection_filtered_response.yml
@@ -10,43 +10,52 @@ value:
         id: 1338
         label: Stormtroopers
         short: ST
+        depth: 1
         _links:
           self:
-            href: /api/v3/custom_field/42/items/1338
+            href: /api/v3/custom_field_items/1338
+            name: Stormtroopers
           parent:
-            href: /api/v3/custom_field/42/items/1337
+            href: /api/v3/custom_field_items/1337
+            name: null
           children:
-            - href: /api/v3/custom_field/42/items/1340
+            - href: /api/v3/custom_field_items/1340
               name: ST-377
-            - href: /api/v3/custom_field/42/items/1341
+            - href: /api/v3/custom_field_items/1341
               name: ST-422
       - _type: HierarchyItem
         id: 1340
         label: ST-377
         short: null
+        depth: 2
         _links:
           self:
-            href: /api/v3/custom_field/42/items/1340
+            href: /api/v3/custom_field_items/1340
+            name: ST-377
           parent:
-            href: /api/v3/custom_field/42/items/1338
+            href: /api/v3/custom_field_items/1338
+            name: Stormtroopers
           children:
-            - href: /api/v3/custom_field/42/items/1480
+            - href: /api/v3/custom_field_items/1480
               name: ST-377-200
-            - href: /api/v3/custom_field/42/items/1481
+            - href: /api/v3/custom_field_items/1481
               name: ST-377-201
       - _type: HierarchyItem
         id: 1341
         label: ST-422
         short: null
+        depth: 2
         _links:
           self:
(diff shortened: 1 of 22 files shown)


docs/api/apiv3/components/examples/hierarchy_item_collection_filtered_response.yml
```diff
@@ -0,0 +1,53 @@
...
+        _links:
+          self:
+            href: /api/v3/custom_field/42/items/1338
+          parent:
+            href: /api/v3/custom_field/42/items/1337
+          children:
+            - href: /api/v3/custom_field/42/items/1340
+              name: ST-377
+            - href: /api/v3/custom_field/42/items/1341
+              name: ST-422
```
brunopagno: I think you're right. Looks simple and elegant.

image

docs/api/apiv3/components/examples/hierarchy_item_collection_response.yml
```diff
@@ -0,0 +1,45 @@
...
+value:
+  _type: Collection
+  total: 37
+  count: 37
+  _embedded:
+    elements:
+      - _type: HierarchyItem
+        id: 1337
+        label: null
+        short: null
```
brunopagno, line 12: This makes me think again if we should copy data to the root item, instead of keeping it only in the custom field.

I imagine that in most cases whoever is consuming the API will also have to fetch the custom field anyways, so this is probably a non issue.

Commenting just not to leave unsaid.

brunopagno (commented):
I say the API spec looks beautiful. A work of art. Deserving of praise.
:tada:

docs/api/apiv3/components/examples/hierarchy_item_collection_response.yml
Kharonus: I also think, the custom field is everytime loaded first. Without fetching it first, you won't have the id and also not the field format. Only with the format of `hierarchy` (and later `list`) the client is interested in retrieving any items for it.

brunopagno: Yes. The only argument against would if if we ever want to extract hierarchy away from custom fields, but that would be solving a problem that does not exist yet.

So, all good :+1:

lib/api/v3/custom_fields/hierarchy/items_api.rb
```diff
@@ -0,0 +1,118 @@
...
+              validation = GetItemsParameterContract.new(hierarchy_root:).call(params)
+              handle_validation_errors(validation) if validation.failure?
+
+              start_item = ::CustomField::Hierarchy::Item.find_by(id: validation[:parent]) || hierarchy_root
+              depth = validation[:depth] || -1
+
+              sub_tree = ::CustomFields::Hierarchy::HierarchicalItemService
+                           .new
+                           .hashed_subtree(item: start_item, depth:)
+                           .value!
```
apfohl: 🟡 Should we handle an error case here?

apfohl (changes requested):

lib/api/v3/custom_fields/hierarchy/items_api.rb
Kharonus: added an either handling

apfohl (approved):

mereghost (commented):
LGTM.

mereghost (approved):

PR #17063, opened 2024-10-28
make watcher spec less flaky

EinLama:
The watcher spec is sometimes red on dev.

```
❯ script/github_pr_errors --display-rerun-info https://github.com/opf/openproject/actions/runs/11553087641/job/32153513118
Looking for the workflow run with id 11553087641
  Branch: dev
  Commit SHA: 5412c5b16ba398451441c4f15c19f22094d9a116
  Commit message: Merge pull request #17044 from opf/feature/46309-custom-field-no-longer-added-to-all-projects-when-added-to-a-type
  Run started at: 2024-10-28 14:34:55 +0300
  ✗ Test suite: failure  https://github.com/opf/openproject/actions/runs/11553087641
    ✗ Units + Features: failure  https://github.com/opf/openproject/actions/runs/11553087641/job/32153513118
Failures explanation:

  1) Watcher tab within a full screen behaves like watchers tab when auto completing users with view_user_email permissions shows the email address of all users
     Failure/Error: expect(target_dropdown).to have_css(".ng-option", text: other_user.name)
       expected to find visible css ".ng-option" with text "Other Bobbit" within #<Capybara::Node::Element tag="ng-dropdown-panel" path="//HTML[1]/BODY[1]/NG-DROPDOWN-PANEL[1]"> but there were no matches. Also found "BB\n Bob Bobbit bobmail47.bobbit@bob.com", which matched the selector but not all filters.
     Shared Example Group: "watchers tab" called from ./spec/features/work_packages/tabs/watcher_tab_spec.rb:172
     # ./spec/features/work_packages/tabs/watcher_tab_spec.rb:148:in `block (5 levels) in <top (required)>'
     # ./spec/support/capybara.rb:16:in `block (2 levels) in <top (required)>'
     # ./spec/support/shared/with_env.rb:58:in `block (2 levels) in <top (required)>'
     # ./spec/support/shared/with_direct_uploads.rb:204:in `block (2 levels) in <top (required)>'
     # ./spec/support/rspec_retry.rb:24:in `block (2 levels) in <top (required)>'
     # ./spec/support/cuprite_setup.rb:126:in `block (2 levels) in <top (required)>'

Finished in 9 minutes 51 seconds (files took 57.33 seconds to load)
2358 examples, 1 failure, 2 pending, 1 error occurred outside of examples

To be with the exact same source files as CI, use this command:
  git checkout 5412c5b16ba398451441c4f15c19f22094d9a116

Tests group #28
[...]

PR #17080, opened 2024-10-29
[#58651] Add reactions button should not be hidden when user has exhausted all reactions

akabiru:
# Ticket

https://community.openproject.org/work_packages/58651

The current pattern hides the "add reactions" button based on whether there are existing reactions or not. This is an inconsistent pattern and can be confusing to the user.

E.g. See Gif below- when "😕" is clicked the "add reactions" banner doesn't display because the logic resolves that there is a button there for the user to click. It only renders back when no reaction exists.

image

## Screenshots

Kapture 2024-10-29 at 18 34 25

Available reactions are never hidden, rather highlighted when active and act as toggle buttons. This pattern is similar to GitHub's emoji reactions

image

# Merge checklist

- [ ] Added/updated tests

### DIFF
--- a/app/components/work_packages/activities_tab/journals/item_component/add_reactions.html.erb
+++ b/app/components/work_packages/activities_tab/journals/item_component/add_reactions.html.erb
@@ -1,34 +1,39 @@
 <%=
   component_wrapper do
-    if journal.available_emoji_reactions.any?
-      render(Primer::Alpha::Overlay.new(
-              title: I18n.t("reactions.action_title"),
-              padding: :condensed,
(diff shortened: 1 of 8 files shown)


jjabari-op (approved):
Great fix! Way better UX now! I only see an unrelated spec failing in the CI, so I approve this PR 👍

akabiru:
> Great fix! Way better UX now! I only see an unrelated spec failing in the CI, so I approve this PR 👍

Thanks! yes, the flaky spec is a known issue being resolved in #17063

PR #17086, opened 2024-10-30
[58689] Eager load relations on work_packages api

cbliard:
# Ticket

https://community.openproject.org/wp/58689

Make `/api/v3/work_packages` endpoint faster.

Eager loaded relations are users (`author`, `assigned_to`, `responsible`), `priority`, and `bcf_issue`. 
- Users have a new `API::V3::WorkPackages::EagerLoading::Users`.
- `priority` and `bcf_issue` have been added in the `includes` of `API::V3::WorkPackages::WorkPackagesEagerLoadingWrapper. 

On my tests, it reduces the overall request duration by 30%. Not as much as one could hope, but still better.

Most of the time is spent rendering the JSON. That's where the N+1 was.
Still there is time consumed to render all URL, and also to sanitize the description of each work package. Not sure how to reduce that time though.

# Merge checklist

- [X] Added/updated tests

### DIFF
--- a/lib/api/v3/work_packages/eager_loading/principals.rb
+++ b/lib/api/v3/work_packages/eager_loading/principals.rb
@@ -1,0 +1,64 @@
+#-- copyright
+# OpenProject is an open source project management software.
+# Copyright (C) the OpenProject GmbH
+#
+# This program is free software; you can redistribute it and/or
+# modify it under the terms of the GNU General Public License version 3.
+#
+# OpenProject is a fork of ChiliProject, which is a fork of Redmine. The copyright follows:
+# Copyright (C) 2006-2013 Jean-Philippe Lang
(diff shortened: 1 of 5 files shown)


cbliard:
It does not work correctly: only users are preloaded. groups and placeholder users are not. I'm working on a fix.

machisuji (approved):
@cbliard 👍 Looks good to me, thanks for tuning that a little! It looks like it does break some specs, though. Once those are updated this is good to merge!

Even if only users are preloaded that's already an improvement, isn't it?

Mennder (approved):
Done

cbliard:
It's fixed now.

@machisuji yes, that's still an improvement. I'm just disappointed that I did not find something more impactful, but we are rendering a *lot* of stuff for just one work package.

PR #17090, opened 2024-10-30
Hierarchy items reorder up and down actions

brunopagno:
# Ticket
https://community.openproject.org/wp/57821/

Allow context actions for reordering hierarchical items up and down

## Screenshots
image

There was a decision on what kind of parameter should we send from the frontend to reorder. We went with an "index" based parameter on the target position we wish to prepend the item because this is most likely to work well on drag & drop commands.

# Merge checklist
- [x] Added/updated tests

### DIFF
--- a/app/controllers/admin/custom_fields/hierarchy/items_controller.rb
+++ b/app/controllers/admin/custom_fields/hierarchy/items_controller.rb
@@ -94,7 +94,7 @@ module Admin
 
         def move
           item_service
-            .reorder_item(item: @active_item, new_sort_order: params[:new_sort_order].to_i)
+            .reorder_item(item: @active_item, new_sort_order: params.require(:new_sort_order))
 
           redirect_to(custom_field_items_path(@custom_field), status: :see_other)
         end
--- a/app/controllers/admin/custom_fields/hierarchy/items_controller.rb
+++ b/app/controllers/admin/custom_fields/hierarchy/items_controller.rb
@@ -92,6 +92,13 @@ module Admin
             )
         end
 
+        def move
+          item_service
+            .reorder_item(item: @active_item, new_sort_order: params[:new_sort_order].to_i)
+
+          redirect_to(custom_field_items_path(@custom_field), status: :see_other)
+        end
+
         def destroy
           item_service
             .delete_branch(item: @active_item)
--- a/app/services/custom_fields/hierarchy/hierarchical_item_service.rb
+++ b/app/services/custom_fields/hierarchy/hierarchical_item_service.rb
@@ -103,17 +103,11 @@ module CustomFields
       def reorder_item(item:, new_sort_order:)
         return Success(item) if item.siblings.empty?
 
-        new_sort_order = [0, new_sort_order].max
+        new_sort_order = [0, new_sort_order.to_i].max
 
         return Success(item) if item.sort_order == new_sort_order
 
-        target_item = item.siblings.find_by(sort_order: new_sort_order)
-        if target_item.present?
-          target_item.prepend_sibling(item)
-        else
-          target_item = item.siblings.last
-          target_item.append_sibling(item)
-        end
+        update_item_order(item:, new_sort_order:)
 
         Success(item.reload)
       end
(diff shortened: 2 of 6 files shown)


config/routes.rb
```diff
@@ -175,9 +175,12 @@
...
         resource :project, controller: "/admin/custom_fields/custom_field_projects", only: :destroy
         resources :items, controller: "/admin/custom_fields/hierarchy/items" do
-          get :deletion_dialog, on: :member
-          get :new_child, on: :member, action: :new
-          post :new_child, on: :member, action: :create
+          member do
+            get :deletion_dialog
+            get :new_child, action: :new
+            post :new_child, action: :create
+            put :move
```
mereghost: 🟡 WDYT about a PATCH or POST here? We are technically just updating the resource (hence PATCH), OTOH this triggers a lot of side-effects on the other items so POST would also be accurate.

app/controllers/admin/custom_fields/hierarchy/items_controller.rb
```diff
@@ -87,6 +87,13 @@ def update
             )
         end
 
+        def move
+          item_service
+            .reorder_item(item: @active_item, new_sort_order: params[:new_sort_order].to_i)
```
mereghost: 🟡 `nil.to_i` and `"".to_i` both become 0. Is that intended?

app/services/custom_fields/hierarchy/hierarchical_item_service.rb
```diff
@@ -97,11 +97,24 @@ def move_item(item:, new_parent:)
...
+
+        target_item = item.siblings.find_by(sort_order: new_sort_order)
+        if target_item.present?
+          target_item.prepend_sibling(item)
+        else
+          target_item = item.siblings.last
+          target_item.append_sibling(item)
+        end
+
+        Success(item.reload)
```
apfohl: 🟡 I would guess this database query is not necessary as the return values of this method is not read.

Also, what happens, if the ordering fails for some reason?

mereghost (approved):
Looks good. Just a couple of questions/ideas.

config/routes.rb
brunopagno: I think you have a good point. I was looking at other examples and we have both routes [with put](https://github.com/opf/openproject/blob/dev/config/routes.rb#L509) and [with post](https://github.com/opf/openproject/blob/dev/config/routes.rb#L348).

I'll adjust this.

apfohl (changes requested):

app/services/custom_fields/hierarchy/hierarchical_item_service.rb
brunopagno: > this database query is not necessary as the return values of this method is not read

I am not sure what to do with this one. It is a bit more defensive to keep the reload if in the future we use this action for something else that depends on the response. But on the other hand it's an extra unnecessary query.

We could replace for an empty `Success`, similar to the `delete_branch` method.

...

> what happens, if the ordering fails for some reason?

I did not think of a path where this operation fails :grimacing:. The worse that can happen is that no reordering happens, but the operation itself is still successful, I think.

app/controllers/admin/custom_fields/hierarchy/items_controller.rb
brunopagno: I think I handled it better now. Just moved the handling to the service instead of doing magic here. And now requesting the parameter to be present.

app/services/custom_fields/hierarchy/hierarchical_item_service.rb
brunopagno: Okay, just for documenting here, I went with removing the object from the success response.

app/services/custom_fields/hierarchy/hierarchical_item_service.rb
```diff
@@ -97,11 +97,18 @@ def move_item(item:, new_parent:)
 
       # Reorder the item along its siblings.
       # @param item [CustomField::Hierarchy::Item] the parent of the node
-      # @param new_sort_order [Integer] the new parent of the node
+      # @param new_sort_order [Integer] the new position of the node
       # @return [Success(CustomField::Hierarchy::Item)]
```
apfohl: To be honest I would remove the comments. At least the return type needs to be fixed.

apfohl (changes requested):
Missed a comment.

app/services/custom_fields/hierarchy/hierarchical_item_service.rb
brunopagno: I like rdoc. They express the intention of the developer. Downside is giving maintenance, but that's fine

mereghost (approved):

apfohl (approved):

PR #17099, opened 2024-10-31
Implements the ability to add items above or below another

mereghost:
## what?

We want the users to be able to add items above or below a certain point in the hierarchy.

## How?

We are now inserting on a specific sort order, that's passed by the form as a hidden field and also a URL parameter to control where the form should be shown.

This PR also makes sure that the appearance of the diverse screens match the figma designs.

## Known issues

1. We are missing some icons
2. We still can't edit and insert at the same time.

image
image
image

Small video demo:

https://github.com/user-attachments/assets/70019468-6251-4aa6-8e38-7d37e6f1ae85

### DIFF
--- a/app/services/custom_fields/hierarchy/hierarchical_item_service.rb
+++ b/app/services/custom_fields/hierarchy/hierarchical_item_service.rb
@@ -49,12 +49,12 @@ module CustomFields
       # @param label [String] the node label/name that must be unique at the same tree level
       # @param short [String] an alias for the node
       # @return [Success(CustomField::Hierarchy::Item), Failure(Dry::Validation::Result), Failure(ActiveModel::Errors)]
-      def insert_item(parent:, label:, short: nil)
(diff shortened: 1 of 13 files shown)


app/services/custom_fields/hierarchy/hierarchical_item_service.rb
```diff
@@ -49,12 +49,12 @@ def generate_root(custom_field)
       # @param label [String] the node label/name that must be unique at the same tree level
       # @param short [String] an alias for the node
       # @return [Success(CustomField::Hierarchy::Item), Failure(Dry::Validation::Result), Failure(ActiveModel::Errors)]
-      def insert_item(parent:, label:, short: nil)
+      def insert_item(parent:, label:, short: nil, sort_order: nil)
```
brunopagno, line 53: super small, but we're missing docs for the `sort_order` param

brunopagno (approved):
These are good changes. I like much more how the ItemComponent is looking now :clap:

PR #17110, opened 2024-11-04
Make Hierarchy::Item#short unique per level

mereghost:
#### Relate WP: [OP#58852](https://community.openproject.org/projects/document-workflows-stream/work_packages/58852)

Gives `#short` the same treatment we have for labels.

### DIFF
--- a/app/contracts/custom_fields/hierarchy/update_item_contract.rb
+++ b/app/contracts/custom_fields/hierarchy/update_item_contract.rb
@@ -31,6 +31,8 @@
 module CustomFields
   module Hierarchy
     class UpdateItemContract < Dry::Validation::Contract
+      config.messages.backend = :i18n
+
       params do
         required(:item).filled(type?: CustomField::Hierarchy::Item)
         optional(:label).filled(:string)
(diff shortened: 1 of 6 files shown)


apfohl (approved):

config/locales/en.yml
```diff
@@ -299,8 +299,16 @@ en:
...
+        short:
+          not_unique: "must be unique within the same hierarchy level"
+        item:
+          root_item: "cannot be a root item"
+          not_persisted: "must be an already existing item"
     rules:
       label: "Label"
+      short: "Short"
+      item: "Item"
+      parent: "Item"
```
Kharonus: 🔴 This one will clash with my current PR. Why not translating key `parent` to `Parent`? And if it is a very specific translation (for this contract), then it should be below a namespace.

app/contracts/custom_fields/hierarchy/update_item_contract.rb
```diff
@@ -31,23 +31,29 @@
...
+
       params do
         required(:item).filled(type?: CustomField::Hierarchy::Item)
         optional(:label).filled(:string)
         optional(:short).filled(:string)
       end
 
       rule(:item) do
-        key.failure("must exist") if value.new_record?
-        key.failure("must not be a root item") if value.root?
```
Kharonus, line 42: 🟢 nice

Kharonus (changes requested):

config/locales/en.yml
mereghost: I'm ok with that.

Kharonus (approved):
LGTM

PR #17114, opened 2024-11-05
Bug/58787 notification banner is showing up on all pages

jjabari-op:
[Ticket](https://community.openproject.org/projects/communicator-stream/work_packages/58787)

The reason for this bug was the global instantiation of the ian center/store/bell state, which was used to let the new activity tab stimulus controller trigger immediate ian center/store/bell state updates. 

Why this was required: The activity tab usually receives workpackage-specific notifications faster than the ian center/store/bell - thus the capability of the immediate update helps to keep the UI in sync.

The approach is now switched to a simple client-side event mechanism instead of calling the ian store method directly. This fixes the reported bug while still enable the immediate update from the new activity tab controller.


akabiru (approved):
Amazing, thanks for resolving this quickly @jjabari-op 🙌🏾 

---
_Primerized Activity tab OFF_
- [x] Notification reload banner is no longer present globally
- [x] Notification reload banner is only triggered within notification center

_Primerized Activity tab ON_

- [x] [Continuously update notification center](https://community.openproject.org/projects/communicator-stream/work_packages/58253)

PR #17119, opened 2024-11-05
Feature/59287 fixed set of project stages and gates editable on project overview page

dombesz:
# Ticket
https://community.openproject.org/wp/59287

Add the foundation structure for the Stages and Gates epic. It consists of:
- [x] Adding a feature flag
- [x] Adding the stages and gates set up as instance wide models
- [x] Setting up Project related stages and gates models
- [x] Seeding a predefined set of stages and gates as outlined in PM2 specification
- [ ] ~~Translating the predefined seeds~~ (Happens on Crowdin)
- [x] Handle data migration for existing instances
- [x] Add newly specced Gate seeds

The approach chosen is to have a base model called `LifeCycle` to share behaviour between the `Stage` and `Gate` models. `Stage` and `Gate` models are mostly the same, except the type.

Additionally a second set of project specific `Project::LifeCycle` models are defined alongside with `Project::Stage` and `Project::Gate`. These models hold project specific information about each `LifeCycle`, such as the start and end dates of life cycles defined inside each project. The `Project::Stage` required a `start_date` and `end_date` fields, while the `Project::Gate` requires a `date` field. These are all mapped in the `Project::LifeCycle` model with STI.

# Merge checklist

- [X] Added/updated tests


app/models/life_cycle.rb
```diff
@@ -0,0 +1,47 @@
...
+
+  def initialize(*args)
+    if instance_of? LifeCycle
+      # Do not allow directly instantiating this class
+      raise NotImplementedError, "Cannot instantiate the base LifeCycle class directly. " \
+                                 "Use Stage or Gate instead."
+    end
+
+    super
+  end
```
toy: Will `self.abstract_class = true` work instead? As a downside I see only the need to set `table_name` explicitly as default will be to use separate per concrete class.

ulferts:
Idea discussed between @dombesz , @toy and @ulferts 

```
Project::LifeCycleStepDefinition { has_many :life_cycle_steps }
Project::GateDefinition < LifeCycleStepDefinition { has_many :gates }
Project::StageDefinition < LifeCycleStepDefinition { has_many :stages }
Project::LifeCycleStep { belongs_to :definition, class: '' }
Project::Gate < LifeCycleStep 
Project::Stage < LifeCycleStep

Project { has_many :life_cycle_steps }
```

app/models/project.rb
```diff
@@ -86,6 +86,8 @@ class Project < ApplicationRecord
   has_many :notification_settings, dependent: :destroy
   has_many :project_storages, dependent: :destroy, class_name: "Storages::ProjectStorage"
   has_many :storages, through: :project_storages
+  has_many :project_life_cycles, class_name: "Project::LifeCycle", dependent: :destroy
+  has_many :life_cycles, through: :project_life_cycles
```
toy: We discussed that this relation may not really be needed

app/models/project/gate.rb
```diff
@@ -0,0 +1,38 @@
...
+#
+# You should have received a copy of the GNU General Public License
+# along with this program; if not, write to the Free Software
+# Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+#
+# See COPYRIGHT and LICENSE files for more details.
+#++
+
+class Project::Gate < Project::LifeCycle
+  validates :date, presence: true
```
toy, line 34: Do you think it is better to have a separate column for `date` and not reuse `start_date` (probably aliasing it)?

app/models/project/life_cycle.rb
```diff
@@ -0,0 +1,47 @@
...
+# along with this program; if not, write to the Free Software
+# Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+#
+# See COPYRIGHT and LICENSE files for more details.
+#++
+
+class Project::LifeCycle < ApplicationRecord
+  belongs_to :project, optional: false
+  belongs_to :life_cycle, optional: false, class_name: "::LifeCycle"
+  has_many :work_packages, inverse_of: :project_life_cycle, dependent: :nullify
```
toy: Was it clarified if work package can or can not belong to multiple stages/gates?

app/models/life_cycle.rb
```diff
@@ -0,0 +1,47 @@
...
+#++
+
+class LifeCycle < ApplicationRecord
+  has_many :project_life_cycles, class_name: "Project::LifeCycle", dependent: :destroy
+  has_many :projects, through: :project_life_cycles
+
+  belongs_to :color, optional: false
+
+  validates :name, presence: true
+  validates :type, inclusion: { in: %w[Stage Gate], message: :must_be_a_stage_or_gate }
```
toy: I think this is already validated on framework level by inheritance

app/models/project/life_cycle.rb
```diff
@@ -0,0 +1,47 @@
...
+
+  attr_readonly :life_cycle_id
+
+  validates :type, inclusion: { in: %w[Project::Stage Project::Gate], message: :must_be_a_stage_or_gate }
```
toy: Same point about validation on framework level

app/models/project/life_cycle.rb
```diff
@@ -0,0 +1,47 @@
...
+
+  def initialize(*args)
+    if instance_of? Project::LifeCycle
+      # Do not allow directly instantiating this class
+      raise NotImplementedError, "Cannot instantiate the base Project::LifeCycle class directly. " \
+                                 "Use Project::Stage or Project::Gate instead."
+    end
+
+    super
+  end
```
toy: Same question about using `self.abstract_class = true `

config/initializers/new_framework_defaults_7_1.rb
```diff
@@ -167,7 +167,7 @@
 # behavior would allow assignment but silently not persist changes to the
 # database.
 #++
-# Rails.application.config.active_record.raise_on_assign_to_attr_readonly = true
+Rails.application.config.active_record.raise_on_assign_to_attr_readonly = true
```
toy, line 170: I think it should be removed from here as you already added it to application.rb

spec/support/shared/project_life_cycle_helpers.rb
```diff
@@ -0,0 +1,48 @@
...
+# You should have received a copy of the GNU General Public License
+# along with this program; if not, write to the Free Software
+# Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+#
+# See COPYRIGHT and LICENSE files for more details.
+#++
+
+RSpec.shared_examples_for "a Project::LifeCycle event" do
+  it "inherits from Project::LifeCycle" do
+    expect(described_class < Project::LifeCycle).to be true
```
toy:
```suggestion
    expect(described_class).to be < Project::LifeCycle
```

spec/support/shared/project_life_cycle_helpers.rb
```diff
@@ -0,0 +1,48 @@
...
+  end
+
+  describe "associations" do
+    it { is_expected.to belong_to(:project).required(true) }
+    it { is_expected.to belong_to(:life_cycle).required(true) }
```
toy: required argument is true by default, so can be:
```suggestion
    it { is_expected.to belong_to(:project).required }
    it { is_expected.to belong_to(:life_cycle).required }
```

spec/support/shared/life_cycle_helpers.rb
```diff
@@ -0,0 +1,50 @@
...
+# You should have received a copy of the GNU General Public License
+# along with this program; if not, write to the Free Software
+# Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+#
+# See COPYRIGHT and LICENSE files for more details.
+#++
+
+RSpec.shared_examples_for "a LifeCycle event" do
+  it "inherits from LifeCycle" do
+    expect(described_class < LifeCycle).to be true
```
toy:
```suggestion
    expect(described_class).to be < LifeCycle
```

spec/support/shared/life_cycle_helpers.rb
```diff
@@ -0,0 +1,50 @@
...
+  end
+
+  describe "associations" do
+    it { is_expected.to have_many(:project_life_cycles).dependent(:destroy) }
+    it { is_expected.to have_many(:projects).through(:project_life_cycles) }
+    it { is_expected.to belong_to(:color).required(true) }
```
toy:
```suggestion
    it { is_expected.to belong_to(:color).required }
```

spec/seeders/basic_data/life_cycle_seeder_spec.rb
```diff
@@ -0,0 +1,191 @@
...
+          type: Stage
+          color_name: :default_color_green_yellow
+      SEEDING_DATA_YAML
+    end
+
+    shared_examples_for "creates the life_cycles seeds" do
+      it "creates the corresponding life cycles with the given attributes" do
+        expect(LifeCycle.count).to eq(5)
+        expect(Stage.find_by(name: "Initiating")).to have_attributes(
+          color_id: Color.find_by(name: "Orange (dark)").id
```
toy: I think it reads better without inversion:
```suggestion
          color: have_attributes(name: "Orange (dark)")
```

spec/models/project/life_cycle_spec.rb
```diff
@@ -0,0 +1,70 @@
...
+    context "when the Gate is already saved" do
+      subject { create :project_gate }
+
+      it "does not allow updating the life_cycle" do
+        expect do
+          subject.life_cycle = build :gate
+        end.to raise_error ActiveRecord::ReadonlyAttributeError
+      end
+    end
+  end
```
toy: Confusing nesting of context, but it can be probably all replaced with a `have_readonly_attribute` test

spec/models/project/life_cycle_spec.rb
```diff
@@ -0,0 +1,70 @@
...
+    expect { described_class.new }.to raise_error(NotImplementedError)
+  end
+
+  it "cannot be instantiated with an invalid type" do
+    expect { described_class.new(type: "InvalidType") }.to raise_error(ActiveRecord::SubclassNotFound)
+  end
+
+  it "can be instantiated with a valid type" do
+    expect { described_class.new(type: "Project::Gate") }.not_to raise_error
+  end
```
toy: These two tests feel like testing rails itself

spec/models/life_cycle_spec.rb
```diff
@@ -0,0 +1,48 @@
...
+    expect { described_class.new }.to raise_error(NotImplementedError)
+  end
+
+  it "cannot be instantiated with an invalid type" do
+    expect { described_class.new(type: "InvalidType") }.to raise_error(ActiveRecord::SubclassNotFound)
+  end
+
+  it "can be instantiated with a valid type" do
+    expect { described_class.new(type: "Gate") }.not_to raise_error
+  end
```
toy: Same point about feeling like this tests the framework

db/migrate/20241114130242_add_active_flag_to_project_life_cycles.rb
```diff
@@ -0,0 +1,5 @@
+class AddActiveFlagToProjectLifeCycles < ActiveRecord::Migration[7.1]
+  def change
+    add_column :project_life_cycles, :active, :boolean, default: false, null: false
```
toy: I think it is better to add the column during creation of the table

db/migrate/20241030154245_create_life_cycles.rb
```diff
@@ -0,0 +1,39 @@
...
+  def change
+    create_table :life_cycles do |t|
+      t.string :type
+      t.string :name
+      t.belongs_to :color, type: :int
+
+      t.timestamps
+    end
+  end
+end
```
toy: As separate creation of tables is needed only during development, I think it is better to combine all migrations into one. As far as I understand PostgreSQL has transaction DDL, so inconsistency should not be a problem

db/migrate/20241105175000_add_life_cycle_to_work_packages.rb
```diff
@@ -0,0 +1,8 @@
+class AddLifeCycleToWorkPackages < ActiveRecord::Migration[7.1]
+  def change
+    add_reference :work_packages,
+                  :project_life_cycle,
+                  foreign_key: { to_table: :project_life_cycles },
```
toy: I think here rails can guess the table, so can be:
```suggestion
                  foreign_key: true,
```

db/migrate/20241030154245_create_life_cycles.rb
```diff
@@ -0,0 +1,39 @@
...
+#
+# See COPYRIGHT and LICENSE files for more details.
+#++
+
+class CreateLifeCycles < ActiveRecord::Migration[7.1]
+  def change
+    create_table :life_cycles do |t|
+      t.string :type
+      t.string :name
+      t.belongs_to :color, type: :int
```
toy: Could you clarify why type: :int is specified when colors.id is `bigint`?
Also why no foreign key, color is not optional and we probably don't want deletion of referenced color.
And I suggest using either only `belongs_to` or only `references` in migrations of one PR (seems code has both methods in existing migrations).

config/locales/en.yml
```diff
@@ -1108,6 +1112,18 @@ en:
...
+            type:
+              must_be_a_stage_or_gate: "must be either Project::Stage or Project::Gate"
+        project/gate:
+          attributes:
+            base:
+              date_range_not_allowed: "Cannot assign `start_date` or `end_date` to a Gate"
+        project/stage:
+          attributes:
+            base:
+              date_not_allowed: "Cannot assign `date` to a Stage"
```
toy: These are very development centric which feels odd, but it should not be possible to trigger them from UI, so not sure if something should be done about it.

app/seeders/common.yml
```diff
@@ -69,6 +69,21 @@ colors:
...
+    hexcode: "#682D91"
+  - reference: :default_color_red_dark
+    t_name: Red (dark)
+    hexcode: "#F05823"
+  - reference: :default_color_magenta_light
+    t_name: Magenta (light)
+    hexcode: "#EC038A"
+  - reference: :default_color_green_yellow
+    t_name: Green Yellow
+    hexcode: "#B1D13A"
```
toy: I'm not sure about this, but maybe better to name according to stage or gate they are used for? Downside is that reusing them will make it confusing and renaming stage or gate will not change the color name

app/seeders/basic_data/color_seeder.rb
```diff
@@ -29,6 +29,17 @@ module BasicData
   class ColorSeeder < ModelSeeder
     self.model_class = Color
     self.seed_data_model_key = "colors"
+    self.attribute_names_for_lookups = %i[name]
+
+    def seed_models!(colors)
+      model_class.transaction do
```
toy: I don't think transaction here is needed, as it will be already called from another transaction

app/seeders/basic_data/life_cycle_seeder.rb
```diff
@@ -0,0 +1,75 @@
...
+      required_colors = models_data.pluck("color_name")
+      required_color_map = required_colors.each_with_object({}) do |reference, colors|
+        color_name = color_seeder.mapped_models_data.dig(reference, :name) or
+                     raise ArgumentError, "Could not find required color #{reference} in seed data definition."
+        colors[color_name] = reference
+      end
+      existing_names = Color.where(name: required_color_map.keys).pluck(:name)
+      missing_names = required_color_map.keys - existing_names
+      required_color_map.values_at(*missing_names)
+    end
```
toy: I think it should be simpler to create another seeder for life cycle colors, like `ColorSchemeSeeder` that also seeds colors

toy (changes requested):
Looks good, most my points are minor, also we agreed to change names of classes as in the comment from @ulferts

app/models/life_cycle.rb
dombesz: Unfortunately the STI mechanism is not compatible with abstract classes.
Defining an abstract parent class will tell rails that there is no table associated with it, and it disables the STI. Even if you define `self.abstract_class = false` for the child classes, the STI mechanism will not be able to infer the common table name. Defining the `table_name` in each child won't work either, because STI mechanism won't kick in and the `type` field won't be correctly utilised for inferring child classes.

toy: Yes, you are correct.
Then it should be enough to validate presence of type, currently validation of inclusion is doing it, but as rails will already check that type is a name of a subclass, I think we don't need to repeat it.

app/seeders/common.yml
dombesz: I didn't use the stage gate names for the downside you mention.

toy: What do you think about adding some prefix, like `PM2` or `Life Cycle` to all color names? `PM2 Orange`, `Life Cycle Orange` or `Life Cycle: Orange` look odd, but it separates them from other colors

app/models/project/life_cycle.rb
dombesz: Yes we have specified that a work package can be optionally assigned to one life cycle.
It is specified [in another epic that is out of scope](https://community.openproject.org/wp/59177) for the current development.

> A stage/gate can be an attribute of a work package (similar to versions).
> A work package can be optionally assigned to stages and gates (no multi-select).

I wouldn't remove it, because it is already done, but it won't get much use until we develop the other epic anyway.

toy: I asked as I remember that there was a question in one of the tickets about allowing work package to be connected to multiple stages/gates, but I failed to find it while reviewing, so was wondering if you know if there was a decision

app/models/life_cycle.rb
dombesz: I agree, the reason behind adding this validation is that the `type` will be a user supplied value and in case it is incorrect, the page will fail with an 500 error. Having this validation will gracefully handle the eventual type mismatch coming from the UI and will display the error message.
Perhaps an improvement would be to add the error to the `:base` and not on the `:type`, as we won't display the type as a separate input field on the LifeCycle creation form.

app/models/project/life_cycle.rb
dombesz: Same reason as [here](https://github.com/opf/openproject/pull/17119/files#r1850134591).
Perhaps I would add another validation that if the associated `life_cycle` is a `Gate` then allow `Project::Gate` as a type, so we won't end up with a type mismatch between a `LifeCycle` and a `Project::LifeCycle`.

config/initializers/new_framework_defaults_7_1.rb
dombesz: Normally I would like to have the configuration defined here, but unfortunately this file it is not being loaded correctly in production and the test environment, hence the definition is also done in the application.rb .

I would rather leave this here, and add a comment in the application.rb to remove that definition, once the issue is cleared.

db/migrate/20241030154245_create_life_cycles.rb
dombesz: Good point! Not sure, this is how it was defined for the other relations in the migrations and I just followed this pattern:
https://github.com/opf/openproject/blob/0ae0349d9bd3f305bcabe2892937c4b3f4417dd4/db/migrate/20180504144320_add_color_to_statuses_and_enumerations.rb#L31-L37
https://github.com/opf/openproject/blob/0ae0349d9bd3f305bcabe2892937c4b3f4417dd4/db/migrate/tables/types.rb#L40

They are old and I'm not sure if there was a specific reason to do this.
Looking through the structure sql, we have a few more examples of `*_id` columns defined as `integer` when their corresponding primary key columns are bigint.
This could cause a problem at some point I guess, I logged a [code maintenance work package](https://community.openproject.org/wp/59476) to solve the other examples.
Edit: As it turns out, those migrations were from pre rails 5.1 times, when the database ids had `integer` type. In rails 5.1 primary keys were changed to `bigint`.

db/migrate/20241030154245_create_life_cycles.rb
dombesz: I would rather keep one table creation per migration, in case we have to roll back a migration it is going to be simpler. Unless there is a benefit from merging all the migrations together.

app/models/project/life_cycle.rb
dombesz: Unfortunately I'm not aware of that discussion. Since this is out of scope for the epic, I would suggest leaving it as is. We can revisit it at a latter point if the specification changes.

dombesz:
@toy I believe I addressed your comments. I will do the discussed renaming in another PR.

app/models/project/gate.rb
```diff
@@ -0,0 +1,40 @@
...
+  alias_attribute :date, :start_date
+
+  validates :date, presence: true
+  validate :end_date_not_allowed
+
+  def end_date_not_allowed
+    if end_date.present?
+      errors.add(:base, :end_date_not_allowed)
+    end
+  end
```
toy, line 41: Just a note that this could have been replaced by `validates :end_date, absence: true`, but it would not be possible to add the error on base

app/models/life_cycle.rb
toy: I at first tried `LifeCycle.new(type: "Color")` and saw it raising an exception, found that this is checked in a call to [`find_sti_class`](https://github.com/rails/rails/blob/5f14ba8d647d241c1705513d718436853faf8464/activerecord/lib/active_record/inheritance.rb#L311C13-L311C27) from [overridden `new` method](https://github.com/rails/rails/blob/5f14ba8d647d241c1705513d718436853faf8464/activerecord/lib/active_record/inheritance.rb#L56), but then realised that the problem is that type will be assigned not during initialisation, but afterwards which is not checked by framework, so `Gate.new.update(type: "Color", name: "Name", color: Color.last)` without this validation inserts an invalid row after which simple `LifeCycle.all.to_a` raises because of the exception from `find_sti_class`.

app/models/life_cycle.rb
toy: Not a merge stopper, but I still think that given validation of type to be in know list, customising initialiser is not needed

app/models/project/life_cycle.rb
toy: See comment for initializer of `::LifeCycle`

spec/models/project/life_cycle_spec.rb
```diff
@@ -0,0 +1,60 @@
...
+    end
+
+    context "when the Gate is already saved" do
+      subject { create :project_gate }
+
+      it "does not allow updating the life_cycle" do
+        expect(subject).to have_readonly_attribute(:life_cycle_id)
+      end
+    end
+  end
```
toy: I meant that this whole describe can be replaced by a check to have_readonly_attribute, see [shoulda matcher docs](https://github.com/thoughtbot/shoulda-matchers/blob/6f4de5ad6d0d09486c2c713fb1a0bf38169d399e/lib/shoulda/matchers/active_record/have_readonly_attribute_matcher.rb)

db/migrate/20241030154245_create_life_cycles.rb
toy: Migrations are connected, so it is not possible to roll back them not from the end, and I assume rolling back only one of the migrations can be helpful for this PR, but probably not really needed afterwards. From my perspective benefit is just in grouping related changes to database, so there is no need to search for commit or even the PR that introduced the change, but it is all in one file.

toy (approved):
LGTM, only deciding on migrations being in one file or not, and merging renaming into this should be done before merging the PR

app/models/project/gate.rb
dombesz: Good point, I didn't knew about `absence: true`, unfortunately in this case wouldn't work as you said.

app/models/life_cycle.rb
dombesz: I believe it is required, because otherwise the validations of the child classes can be circumvented. Let's consider the following scenario:
 1. Instantiate a `Project::LifeCycleStep.new`
 2. Assign attributes outside of the initialisation also including the `type: "Project::Stage"`
 3. The class of the already initialised object won't change to `Project::Stage` after the assignment, and the validations declared in the `Project::Stage` class won't be applied.
 4. An invalid object will be saved to the database.

```ruby
[1] pry(main)> lc = Project::LifeCycleStep.new
=> #<Project::LifeCycleStep:0x00000003144794d0
 id: nil,
 type: nil,
 start_date: nil,
 end_date: nil,
 active: false,
 position: nil,
 project_id: nil,
 definition_id: nil,
 created_at: nil,
 updated_at: nil>
[2] pry(main)> lc.assign_attributes(type: "Project::Stage", definition: Project::LifeCycleStepDefinition.first, project: Project.first)
  Project::LifeCycleStepDefinition Load (1.8ms)  SELECT "project_life_cycle_step_definitions".* FROM "project_life_cycle_step_definitions" ORDER BY "project_life_cycle_step_definitions"."id" ASC LIMIT $1  [["LIMIT", 1]]
  Project Load (0.5ms)  SELECT "projects".* FROM "projects" ORDER BY "projects"."id" ASC LIMIT $1  [["LIMIT", 1]]
=> nil
[3] pry(main)> lc.valid?
=> true
[4] pry(main)> lc
=> #<Project::LifeCycleStep:0x00000003144794d0
 id: nil,
 type: "Project::Stage",
 start_date: nil,
 end_date: nil,
 active: false,
 position: nil,
 project_id: 1,
 definition_id: 1,
 created_at: nil,
 updated_at: nil>
[5] pry(main)> lc.save
  TRANSACTION (11.8ms)  BEGIN
  Project::LifeCycleStep Load (18.0ms)  SELECT "project_life_cycle_steps".* FROM "project_life_cycle_steps" WHERE (1 = 1) AND ("project_life_cycle_steps"."position" IS NOT NULL) ORDER BY "project_life_cycle_steps"."position" DESC LIMIT $1  [["LIMIT", 1]]
  Project::LifeCycleStep Create (3.9ms)  INSERT INTO "project_life_cycle_steps" ("type", "position", "project_id", "definition_id", "created_at", "updated_at") VALUES ($1, $2, $3, $4, $5, $6) RETURNING "id"  [["type", "Project::Stage"], ["position", 1], ["project_id", 1], ["definition_id", 1], ["created_at", "2024-11-22 11:21:33.569747"], ["updated_at", "2024-11-22 11:21:33.569747"]]
  TRANSACTION (2.7ms)  COMMIT
=> true
[6] pry(main)> Project::LifeCycleStep.find(lc.id).valid? 
  Project::LifeCycleStep Load (12.4ms)  SELECT "project_life_cycle_steps".* FROM "project_life_cycle_steps" WHERE "project_life_cycle_steps"."id" = $1 LIMIT $2  [["id", 1], ["LIMIT", 1]]
  Project Load (0.4ms)  SELECT "projects".* FROM "projects" WHERE "projects"."id" = $1 LIMIT $2  [["id", 1], ["LIMIT", 1]]
  Project::LifeCycleStepDefinition Load (0.2ms)  SELECT "project_life_cycle_step_definitions".* FROM "project_life_cycle_step_definitions" WHERE "project_life_cycle_step_definitions"."id" = $1 LIMIT $2  [["id", 1], ["LIMIT", 1]]
=> false

```

app/models/project/life_cycle.rb
dombesz: See above https://github.com/opf/openproject/pull/17119#discussion_r1853763577 .

spec/models/project/life_cycle_spec.rb
dombesz: I still have to keep the context, because the subject should be instantiable for the test to work.

db/migrate/20241030154245_create_life_cycles.rb
dombesz: I added it to the [renaming PR](https://github.com/opf/openproject/pull/17248) to avoid merge conflicts.

app/models/life_cycle.rb
toy: It is still possible to create an invalid Stage if Gate is initialised and then switched dynamically - create a gate with only `start_date`/`date` and change type to Stage. So it is a problem of STI, if you really want to ensure no invalid instances, all validation should move from sub classes to base class and apply depending on the value of `type` attribute
```rb
gate = Project::Gate.new
gate.attributes = {start_date: Date.today, type: 'Project::Stage', definition: Project::LifeCycleStepDefinition.first, project: Project.first}
gate.valid? # => true
gate.save
Project::LifeCycleStep.find(gate.id).valid? # => false
Project::LifeCycleStep.find(gate.id).tap(&:validate).errors # => #<ActiveModel::Errors [#<ActiveModel::Error attribute=end_date, type=blank, options={}>]>
```

dombesz: I don't like the idea of moving class related validations into the parent model, it defeats the purpose of the whole structure.
I would rather add the following validation to ensure no invalid objects can be created:

```ruby
class Project::Stage < Project::LifeCycleStep
  validates :type, inclusion: { in: %w[Project::Stage], message: :must_be_a_stage }
end

class Project::Gate < Project::LifeCycleStep
  validates :type, inclusion: { in: %w[Project::Gate], message: :must_be_a_gate }
end
```
WDYT?

toy (approved):
LGTM

PR #17120, opened 2024-11-05
Reordering by drag and drop

brunopagno:
# Ticket
https://community.openproject.org/wp/57822

Drag and drop reordering of hierarchy items

## Screenshots

https://github.com/user-attachments/assets/69867986-8e7d-4407-9a56-1686e44adb56

We have a few things to mention here:
1. Each row in the list will be draggable (on the drag handle) and will serve as a drop zone
2. Moving and item up will insert above of the highlighted item, and moving down will insert below
3. We are able to use turbo frames thanks to some clever request attributes. We make a `fetch` (POST) submiting a form like request to reorder the item, and when the request responds we trigger a Turbo.visit which reloads the frame beautifully

Building UI is not easy :weary: 

# Merge checklist

### DIFF
--- a/frontend/src/stimulus/controllers/dynamic/admin/hierarchy-item.controller.ts
+++ b/frontend/src/stimulus/controllers/dynamic/admin/hierarchy-item.controller.ts
@@ -32,40 +32,55 @@ import { Controller } from '@hotwired/stimulus';
 import * as Turbo from '@hotwired/turbo';
 
 export default class HierarchyItemController extends Controller {
+  private currentHover:HTMLElement | null;
+
   connect() {}
 
   dragstart(event:DragEvent) {
     const element = dataNode(event.target);
+    element.style.opacity = '0.4';
 
     if (event.dataTransfer) {
       event.dataTransfer.setDragImage(element, 25, 25);
       event.dataTransfer.effectAllowed = 'move';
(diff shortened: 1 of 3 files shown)


frontend/src/stimulus/controllers/dynamic/admin/hierarchy-item.controller.ts
```diff
@@ -0,0 +1,125 @@
...
+
+import { Controller } from '@hotwired/stimulus';
+import * as Turbo from '@hotwired/turbo';
+
+export default class HierarchyItemController extends Controller {
+  private currentHover:HTMLElement | null;
+
+  connect() {}
+
+  dragstart(event:DragEvent) {
```
apfohl, line 39: 🟢 Can these methods be camel cased?

apfohl (approved):

frontend/src/stimulus/controllers/dynamic/admin/hierarchy-item.controller.ts
brunopagno: These are browser events, and [MDN](https://developer.mozilla.org/en-US/docs/Web/API/DragEvent) is using them as lowercase. We could rename them for internal use, but I think it's better to keep them according to the convention.

PR #17125, opened 2024-11-06
Implementation/58521 api add endpoint for fetching a branch

apfohl:
# Ticket
https://community.openproject.org/projects/document-workflows-stream/work_packages/58521/

Added API endpoint for fetching the branch of a custom field item.

### DIFF
--- a/docs/api/apiv3/paths/custom_field_item_branch.yml
+++ b/docs/api/apiv3/paths/custom_field_item_branch.yml
@@ -1,0 +1,63 @@
+# /api/v3/custom_field_items/{id}/branch
+---
+get:
+  summary: Get a custom field hierarchy item's branch
+  operationId: get_custom_field_item_branch
+  description: |-
+    Retrieves the branch of a single custom field item specified by its unique identifier.
+
+    A branch is list of all ancestors, starting with the root item and finishing with the item itself.
+  parameters:
+    - name: id
+      description: The custom field item's unique identifier
+      in: path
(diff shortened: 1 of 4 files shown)


docs/api/apiv3/paths/custom_field_item_branch.yml
```diff
@@ -0,0 +1,61 @@
+# /api/v3/custom_field_items/{id}/branch
+---
+get:
+  summary: Get a custom field hierarchy item's branch
+  operationId: get_custom_field_item_branch
+  description: |-
+    Retrieves the branch of a single custom field item specified by its unique identifier.
```
Kharonus, line 7: 🟡 I'd like to explain here, what we understand under the term `branch`. Something like.

> A branch is list of all ancestors, starting with the root item and finishing with the item itself.

Kharonus (changes requested):
I think the API spec needs a small improvement, then it's green from my side.

Kharonus (approved):

docs/api/apiv3/paths/custom_field_item_branch.yml
wielinde: Interesting. Intuitively I would have assumed that a branch is everything below an item (descendants) and not above (ascendants). So, apparently it is not clear to me not to @Kharonus.

Ideas:
- ItemWithAscendants
- ItemWithPath (inspired by Graph API)

PR #17130, opened 2024-11-06
[#58522] add drop down selection to hierarchy cfs

Kharonus:
# Ticket
[OP#58522](https://community.openproject.org/work_packages/58522)

- the user should see a dropdown for custom fields of type hierarchy

- amend hierarchy namespaces, routes and strategies
- amend select-edit-field and multi-select-edit-field to be able to react on depth of options

#### Hint
- The indention of the options is still 1 level too big. The reason for this will get fixed in [OP#58868](https://community.openproject.org/work_packages/58868)


apfohl (approved):
Looks good so far.

PR #17139, opened 2024-11-07
[58868] Reworked depth calculation for ItemsAPI for hierarchical custom field

apfohl:
# Ticket
https://community.openproject.org/projects/document-workflows-stream/work_packages/58868/

Optimize depth calculation for ItemsAPI to not have a second database query.

We hooked into the flattening algorithm for the hash based tree representation and introduced a depth counter for every level of hierarchy.


mereghost (approved):
Looks good. 👍🏼 

It just occurred to me that for non depth limited queries the `Item.root_and_descendants_preordered` or the `Item#self_and_descendants_preordered` gives us the same result as the helper method here.

PR #17145, opened 2024-11-07
Overwrite version of fetched journal entries for the activity tab

ulferts:
# Ticket

https://community.openproject.org/wp/58513

Before changing the algorithm, journals were aggregated by creating a new one and removing the predecessor. That however lead to the version numbers having gaps. 

Overwriting the `version` by a custom select based on a window function ensures the version to increased with a fixed increment of 1. 

# Merge checklist

- [ ] Added/updated tests

### DIFF
--- a/app/controllers/work_packages/activities_tab_controller.rb
+++ b/app/controllers/work_packages/activities_tab_controller.rb
@@ -213,7 +213,9 @@ class WorkPackages::ActivitiesTabController < ApplicationController
   end
 
   def find_journal
-    @journal = Journal.find(params[:id])
+    @journal = Journal
+      .with_sequence_version
+      .find(params[:id])
   rescue ActiveRecord::RecordNotFound
     respond_with_error(I18n.t("label_not_found"))
   end
@@ -309,7 +311,9 @@ class WorkPackages::ActivitiesTabController < ApplicationController
   end
 
   def generate_time_based_update_streams(last_update_timestamp)
-    journals = @work_package.journals
+    journals = @work_package
+                 .journals
+                 .with_sequence_version
 
     if @filter == :only_comments
       journals = journals.where.not(notes: "")
--- a/app/models/journals/scopes/with_sequence_version.rb
+++ b/app/models/journals/scopes/with_sequence_version.rb
@@ -1,0 +1,55 @@
+# -- copyright
+# OpenProject is an open source project management software.
+# Copyright (C) 2010-2024 the OpenProject GmbH
+#
+# This program is free software; you can redistribute it and/or
+# modify it under the terms of the GNU General Public License version 3.
+#
+# OpenProject is a fork of ChiliProject, which is a fork of Redmine. The copyright follows:
+# Copyright (C) 2006-2013 Jean-Philippe Lang
+# Copyright (C) 2010-2013 the ChiliProject Team
+#
+# This program is free software; you can redistribute it and/or
+# modify it under the terms of the GNU General Public License
+# as published by the Free Software Foundation; either version 2
+# of the License, or (at your option) any later version.
+#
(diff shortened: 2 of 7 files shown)


wielinde:
@ulferts In the future we will attach much more objects to work package comments, such as threads. So we will need great referential integrity and simplicity when working with comments/journals. Will this solution be easy to work with? For example, a query: "Give me all threads that are attached to journal 34".

wielinde:
To answer my question myself: References should not use the version number but the journal ID.

akabiru:
> @ulferts In the future we will attach much more objects to work package comments, such as threads. So we will need great referential integrity and simplicity when working with comments/journals. Will this solution be easy to work with? For example, a query: "Give me all threads that are attached to journal 34".

@wielinde we discussed this as a bandaid solution for 15.0 - the plan is still to go ahead with the migration in the next release- and reverse this work.

oliverguenther (approved):
LGTM. Tested this on a community dump and saw all versions in examples that were linked as fixed.

app/controllers/work_packages/activities_tab_controller.rb
```diff
@@ -213,7 +213,10 @@ def find_project
   end
 
   def find_journal
-    @journal = Journal.find(params[:id])
+    @journal = Journal
+      .with_sequence_version
+      .where(id: params[:id])
+      .first
```
akabiru: 🟢 should raise `ActiveRecord::RecordNotFound` handled below
```suggestion
    @journal = Journal
      .with_sequence_version
      .find_by!(id: params[:id])
```

app/models/journals/scopes/with_sequence_version.rb
```diff
@@ -0,0 +1,55 @@
...
+    extend ActiveSupport::Concern
+
+    class_methods do
+      def with_sequence_version
+        joins(
+          <<~SQL.squish
+            JOIN LATERAL (
+              SELECT
+                journals_rank.id,
+                ROW_NUMBER() OVER (PARTITION BY journable_id ORDER BY version ASC) sequence_version
```
akabiru: ✨ Amazing.

---

One thing to note is `row_number` [_"generates a new row number for every row, regardless of duplicates within a partition."_](https://blog.jooq.org/sql-trick-row_number-is-to-select-what-dense_rank-is-to-select-distinct/) so we can only rely on the `sequence_version` when journables are already scoped- otherwise we'll get duplicates.

akabiru (approved):
🚀

app/controllers/work_packages/activities_tab_controller.rb
ulferts: I applied it sort of (used `find`) but merged it into the one commit. I'd like to keep it to a single commit so that it is easier to revert.

app/models/journals/scopes/with_sequence_version.rb
ulferts: Good pointer. In this case, I think we are safe given that the subselect is filtered down to the same journable (id and type) by the `where`s

ulferts:
The failing spec

`rspec ./spec/features/activities/work_package/activities_spec.rb:301` 

is green locally (ran it 5 times). I'll try to rerun it to see if that magically fixes it on the CI as well.

akabiru:
> The failing spec
> 
> `rspec ./spec/features/activities/work_package/activities_spec.rb:301`
> 
> is green locally (ran it 5 times). I'll try to rerun it to see if that magically fixes it on the CI as well.

I'm afraid we have several flaky tests on this feature - I'd be happy to log a maintenance on this one for now.

ulferts:
Different specs failed this time so I put it down to flakiness and merge the PR.

PR #17150, opened 2024-11-08
Bug/59082 the comment box in activity requires a second click for keyboard focus

jjabari-op:
[Ticket](https://community.openproject.org/projects/communicator-stream/work_packages/59082)


akabiru (approved):
Nice one 🥇 Thanks for the tests too!

PR #17164, opened 2024-11-11
load storage/activity data only for admins

ulferts:
# Ticket

https://community.openproject.org/wp/59202

Avoid the performance penalty caused by loading the storage and activity information at least for non admins. 

The problem needs to be fixed for admins independently but at least the bulk of users should be freed from it.


oliverguenther (approved):

ulferts:
The failing specs are flickering. The two test runs resulted in different specs failing.

PR #17173, opened 2024-11-11
[#58865] add enterprise banner to new custom field form

Kharonus:
# Ticket
[OP#58865](https://community.openproject.org/work_packages/58865)

- do not allow creation of hierarchy custom fields if no enterprise token is active

- show banner below form field selection, if
  - hierarchy is selected AND
  - enterprise token is not active
- disable submit button, if the enterprise banner is shown

### DIFF
--- a/app/helpers/custom_fields_helper.rb
+++ b/app/helpers/custom_fields_helper.rb
@@ -68,7 +68,7 @@ module CustomFieldsHelper
     field_name = "#{name}[custom_field_values][#{custom_field.id}]"
     field_id = "#{name}_custom_field_values_#{custom_field.id}"
 
-    field_format = OpenProject::CustomFieldFormat.find_by_name(custom_field.field_format)
+    field_format = OpenProject::CustomFieldFormat.find_by(name: custom_field.field_format)
 
     tag = case field_format.try(:edit_as)
           when "date"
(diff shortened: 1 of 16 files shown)


app/views/custom_fields/_form.html.erb
```diff
@@ -62,6 +59,17 @@ See COPYRIGHT and LICENSE files for more details.
                  }
     %>
   </div>
+  <div class="form--field-extra-actions" <%= format_dependent.attributes(:enterpriseBanner) %>>
+    <%= angular_component_tag "opce-enterprise-banner",
+                              inputs: {
+                                collapsible: true,
+                                opReferrer: "custom-field-hierarchy",
+                                textMessage: "PLACEHOLDER - upgrade to Starkiller Base",
```
dominic-braeunlein, line 67: Hi @Kharonus, here is still a placeholder UI copy in the PR.

dominic-braeunlein (changes requested):

app/helpers/custom_fields_helper.rb
```diff
@@ -198,21 +198,25 @@ def custom_field_formats_for_select(custom_field)
...
 
   def label_for_custom_field_format(format_string)
-    format = OpenProject::CustomFieldFormat.find_by_name(format_string)
+    show_enterprise_text = format_string == "hierarchy" && !EnterpriseToken.allows_to?(:custom_field_hierarchies)
 
-    if format
-      format.label.is_a?(Proc) ? format.label.call : I18n.t(format.label)
-    end
+    format = OpenProject::CustomFieldFormat.find_by(name: format_string)
+    label = if format
```
apfohl: 🟡 Could be rewritten to `format&.label.is_a?(Proc)`.

apfohl (approved):
One small thing you can irgnore.

app/views/custom_fields/_form.html.erb
Kharonus: Yes, there is a [TASK](https://community.openproject.org/wp/59219) for Maya, to provide the correct text.

My take was, as this is behind a feature flag, we can merge it even with the placeholder. But before removing the flag, we need to replace it.

app/helpers/custom_fields_helper.rb
Kharonus: rewritten with a bit of reordering...

PR #17181, opened 2024-11-12
Bump IfcConvert version

top4ek:
Bump IfcConvert tool to v0.7.11 

Got some models unable to convert with 0.6. After migration to Debian 12 we can easily switch official builds.


dombesz (approved):
Thanks @top4ek for your contribution!👍

PR #17187, opened 2024-11-12
Bug/59239 notification bubbles and mark as read buttons not responding correctly to changes

jjabari-op:
[Ticket](https://community.openproject.org/work_packages/59239)


akabiru (approved):
Elegant solution! ✨

PR #17189, opened 2024-11-12
[58857] Added specs for hierarchy items API and representer

apfohl:
https://community.openproject.org/projects/openproject/work_packages/58857/

- Added request specs for /api/v3/custom_field/{id}/items
- Added request specs for /api/v3/custom_field_items/{id}
- Added request specs for /api/v3/custom_fiels_items/{id}/branch
- Added representer specs for HierarchyItemRepresenter
- Make shift repair of Dry::Validation::Contract english localization

### DIFF
--- a/spec/lib/api/v3/custom_fields/hierarchy/hierarchy_item_representer_rendering_spec.rb
+++ b/spec/lib/api/v3/custom_fields/hierarchy/hierarchy_item_representer_rendering_spec.rb
@@ -1,0 +1,223 @@
+# frozen_string_literal: true
+
+#-- copyright
+# OpenProject is an open source project management software.
+# Copyright (C) the OpenProject GmbH
+#
+# This program is free software; you can redistribute it and/or
+# modify it under the terms of the GNU General Public License version 3.
+#
+# OpenProject is a fork of ChiliProject, which is a fork of Redmine. The copyright follows:
+# Copyright (C) 2006-2013 Jean-Philippe Lang
+# Copyright (C) 2010-2013 the ChiliProject Team
+#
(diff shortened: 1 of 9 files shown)


lib/api/v3/utilities/path_helper.rb
```diff
@@ -216,7 +216,7 @@ def self.custom_field_item(id)
           end
 
           def self.custom_field_item_branch(id)
-            "#{custom_field_item(id)}/branch"
+            "#{root}/custom_field_items/#{id}/branch"
```
Kharonus: ❓ why replacing this? What, if the resource is renamed, then we have here a possible error, that only one of the methods is updated?

spec/lib/api/v3/custom_fields/hierarchy/hierarchy_item_representer_rendering_spec.rb
```diff
@@ -0,0 +1,223 @@
...
+      it_behaves_like "property", :label do
+        let(:value) { item.label }
+      end
+
+      it_behaves_like "property", :short do
+        let(:value) { item.short }
+      end
+
+      it_behaves_like "property", :depth do
+        let(:value) { item.depth }
```
Kharonus, line 162: 🔴 moment, why does this work?? `item.depth` would execute the the depth query of an `HierarchicalItem`, it should not be the same as the rendered value in the JSON?

Kharonus (changes requested):

Kharonus (approved):
Perfect, if (non-flaky) specs are green, you can merge it,

PR #17193, opened 2024-11-12
Implementation/59254 ensure feature flag for primerised activity is always true via env var

jjabari-op:
[Ticket](https://community.openproject.org/projects/communicator-stream/work_packages/59254/activity)

Meant as a substitute for #17192


jjabari-op:
The CI shows random flaky specs (not related to the change in this PR as far as I can see it). I already tried to re-run the CI - both times different specs are failing.

PR #17196, opened 2024-11-13
Fix/bump gems

ulferts:

Regular gem bumps


aaron-contreras:
There's a known issue with rackup, puma and new Rails version compatibility. The fix is merged into Puma's main branch but there's not tag released for it yet.

https://github.com/puma/puma/pull/3532#issuecomment-2433486115

PR #17197, opened 2024-11-13
Enable activities pending specs, address flakiness

akabiru:
https://community.openproject.org/wp/59282

- [ ] `spec/features/activities/work_package/activities_spec.rb:369`(https://github.com/opf/openproject/actions/runs/11852444503/job/33030633302)


aaron-contreras:
Would something like this be helpful? You can set the polling interval to 1 second and this guarantees that RSpec waits are longer than the time it takes the poll to happen. Would make it simpler to deal with the flakiness

```ruby
RSpec.describe "boards onboarding tour",
               :js,
               with_settings: { notifications_polling_interval: 1_000 } do
```

akabiru:
> Would something like this be helpful? You can set the polling interval to 1 second and this guarantees that RSpec waits are longer than the time it takes the poll to happen. Would make it simpler to deal with the flakiness
> 
> ```ruby
> RSpec.describe "boards onboarding tour",
>                :js,
>                with_settings: { notifications_polling_interval: 1_000 } do
> ```

Cheers for checking in @aaron-contreras - it's slightly different as the _new_ "activities tab polling" is separate from the "old" notifications polling interval. At present, the activities tab polling interval [is hardcoded to 10s](https://github.com/opf/openproject/blob/dev/app/components/work_packages/activities_tab/index_component.rb#L70-L77) but can be changed in test env via `WORK_PACKAGES_ACTIVITIES_TAB_POLLING_INTERVAL_IN_MS` - albeit not the most fitting approach.

akabiru:
Duplicate of #17952

PR #17200, opened 2024-11-13
Feature/58163 project specific stage administration

ulferts:
# Ticket

https://community.openproject.org/wp/58163

* [x] A new menu is added to the project menu below the &quot;Project settings&quot; labeled &quot;Project lifecycle&quot;    
* [x] It looks and behaves similar to the project attributes administration.
  * [x] Individual Toggle
  * [x] Bulk enable/disable
  * [x] Filtering
* [x] To see and access the project administration a permission is required.    
    * [x] The permissions is called &quot;Select project stages and gates&quot;.        
    * [x] Roles having the &#39;edit project&#39; permission will receive the new permission both when seeding and when migrating.        
* [x] Within this administration, all currently existing stages and gates are listed.    
    * [x] The name and type (stage/gate) is displayed.        
        * [x] The type (rhombus for gate, arrow for stage) are displayed in the color defined for the stage/gate.            
    * [x] They are listed in the order defined in the instance&#39;s administration.        
        * [x] The order cannot be changed.            
* [x] The project administrator can activate and deactivate a gate/stage.
* [x] The new functionality is hidden behind the feature flag developed as part of ###58159

Out of scope from the original requirements as of now as it has not been implemented yet:

* Disabled stages/gates will not be displayed:    
    * In the overview page        
    * In the project list (see ##58160)       

## Screenshots

image

# Changes of note

* The settings page helpers for specs are split up into individual files. Before, it was possible to go to the various pages by `visit_tab!('backlogs')`. The menu however no longer uses tabs and adding all the helpers to that one page led to some cluttering in there.
* A component for displaying the type of a life cycle has been added and documented in the lookbook.

# Merge checklist

- [x] Added/updated tests


ulferts:
The failing feature specs are flickering and weren't broken by the changes within this PR.

app/components/projects/life_cycle_type_component.html.erb
```diff
@@ -0,0 +1,8 @@
+<%= flex_layout(align_items: :center) do |type_container|
```
toy, line 1: Is align_items needed?

app/components/projects/settings/life_cycles/show_page_header_component.html.erb
```diff
@@ -0,0 +1,9 @@
+<%= render Primer::OpenProject::PageHeader.new do |header| %>
+  <%= header.with_title(variant: :default) { t('projects.settings.life_cycle.header.title') } %>
+  <%= header.with_description {
+    t('projects.settings.life_cycle.header.description',
+      overview_url: project_path(@project),
+      admin_settings_url: admin_settings_project_life_cycles_path).html_safe
```
toy: I think it is better to use `_html` suffix for the translation key instead of using `html_safe`

app/components/projects/settings/life_cycles/show_page_header_component.html.erb
```diff
@@ -0,0 +1,9 @@
+<%= render Primer::OpenProject::PageHeader.new do |header| %>
+  <%= header.with_title(variant: :default) { t('projects.settings.life_cycle.header.title') } %>
```
toy: Default variant seem to be same as medium variant which is the default (is confusion intended?), but variant seem to be not set in most cases
```suggestion
  <%= header.with_title { t('projects.settings.life_cycle.header.title') } %>
```

app/components/projects/settings/life_cycles/show_page_header_component.rb
```diff
@@ -0,0 +1,46 @@
...
+# ++
+
+module Projects::Settings::LifeCycles
+  class ShowPageHeaderComponent < ApplicationComponent
+    include ApplicationHelper
+
+    def initialize(project: nil)
+      super
+      @project = project
+    end
```
toy: Why not use `options :project` instead? If not, it should be `project:`, as project is required

app/components/projects/settings/life_cycles/show_page_header_component.rb
```diff
@@ -0,0 +1,46 @@
...
+
+    def breadcrumb_items
+      [{ href: project_overview_path(@project.id), text: @project.name },
+       { href: project_settings_general_path(@project.id), text: I18n.t("label_project_settings") },
```
toy: I assume it was copied from project attributes, as it is better to use slug (`@project`) instead of project id (`@project.id`)

app/components/projects/settings/life_cycles/index_component.rb
```diff
@@ -0,0 +1,63 @@
...
+
+        attr_reader :project,
+                    :life_cycle_definitions
+
+        def initialize(project:, life_cycle_definitions:)
+          super
+
+          @project = project
+          @life_cycle_definitions = life_cycle_definitions
+        end
```
toy:
```suggestion
        options :project,
                    :life_cycle_definitions
```

app/components/projects/settings/life_cycles/step_component.rb
```diff
@@ -0,0 +1,66 @@
...
+
+        attr_reader :definition,
+                    :project
+
+        def initialize(project:, definition:)
+          super
+
+          @definition = definition
+          @project = project
+        end
```
toy:
```suggestion
        options :definition,
                    :project
```

app/components/projects/settings/life_cycles/step_component.html.erb
```diff
@@ -0,0 +1,37 @@
...
+        end
+        title_container.with_column(pt: 1) do
+          render(Projects::LifeCycleTypeComponent.new(definition))
+        end
+      end
+      # py: 1 quick fix: prevents the row from bouncing as the toggle switch currently changes height while toggling
+      definition_container.with_column(py: 1, mr: 2) do
+        # buggy currently:
+        # small variant +  status_label_position: :start leads to a small bounce while toggling
+        # behavior can be seen on primer's viewbook as well -> https://view-components-storybook.eastus.cloudapp.azure.com/view-components/lookbook/inspect/primer/alpha/toggle_switch/small
```
toy: Does the preview work for you? In local development I get exception due to absent route, and created a [bug ticket](https://community.openproject.org/wp/59918)

app/components/projects/settings/life_cycles/show_page_header_component.html.erb
```diff
@@ -0,0 +1,9 @@
...
+  <%= header.with_description {
```
toy: Should it be:
```suggestion
  <%= header.with_description do
```

app/components/projects/settings/life_cycles/step_component.html.erb
```diff
@@ -0,0 +1,37 @@
...
+        # quick fix: don't display loading indicator which is causing the bounce
+        render(Primer::Alpha::ToggleSwitch.new(
+          src: toggle_project_settings_life_cycle_path(
+            project_life_cycle: {
+              definition_id: definition.id,
+              type: definition.type
```
toy: I think it is better to decide on type based on definition in controller, so only definition id is needed

config/routes.rb
```diff
@@ -518,6 +527,9 @@
           delete :unlink
         end
       end
+      # TODO: This is for now only added to be able to create a link
+      # There is no controller behind this.
+      resources :project_life_cycles, only: %i[index]
```
toy: Should probably be `project_life_cycle_step_definitions`

config/routes.rb
```diff
@@ -244,6 +244,15 @@
             put :disable_all_of_section
           end
         end
+        resource :life_cycle, only: %i[show] do
```
toy: Shouldn't it be `resources` with only `index` instead?

app/controllers/projects/settings/life_cycles_controller.rb
```diff
@@ -0,0 +1,114 @@
...
+
+  def show; end
+
+  def toggle
+    step_params = params
+                    .require(:project_life_cycle)
+                    .permit(:definition_id, :type)
+                    .to_h
+                    .symbolize_keys
+                    .merge(active: params["value"] == "1")
```
toy: Documentation for `upsert_all` says that type casting is done, so there is no need to cast manually.

Unfortunate that name `value` is hardcoded in `ToggleSwitch`

app/components/projects/settings/life_cycles/step_component.html.erb
```diff
@@ -0,0 +1,37 @@
...
+      end
+      # py: 1 quick fix: prevents the row from bouncing as the toggle switch currently changes height while toggling
+      definition_container.with_column(py: 1, mr: 2) do
+        # buggy currently:
+        # small variant +  status_label_position: :start leads to a small bounce while toggling
+        # behavior can be seen on primer's viewbook as well -> https://view-components-storybook.eastus.cloudapp.azure.com/view-components/lookbook/inspect/primer/alpha/toggle_switch/small
+        # quick fix: don't display loading indicator which is causing the bounce
+        render(Primer::Alpha::ToggleSwitch.new(
+          src: toggle_project_settings_life_cycle_path(
+            project_life_cycle: {
```
toy: To me namespacing feels unneeded here, but feel free to ignore

app/controllers/projects/settings/life_cycles_controller.rb
```diff
@@ -0,0 +1,114 @@
...
+      [{
+        project_id: @project.id,
+        definition_id:,
+        active:,
+        type: project_type_for_definition_type(type)
+      }]
+    )
+  end
+
+  def upsert_all_steps(active: true)
```
toy: Why set default value for `active`?

app/controllers/projects/settings/life_cycles_controller.rb
```diff
@@ -0,0 +1,114 @@
...
+
+  def load_life_cycle_definitions
+    @life_cycle_definitions = Project::LifeCycleStepDefinition.order(position: :asc)
+  end
+
+  def deny_access_on_feature_flag
+    deny_access unless OpenProject::FeatureDecisions.stages_and_gates_active?
+  end
+
+  def upsert_one_step(definition_id:, type:, active: true)
```
toy: Why set default value for `active`?

app/controllers/projects/settings/life_cycles_controller.rb
```diff
@@ -0,0 +1,114 @@
...
+  end
+
+  def upsert_all_steps(active: true)
+    upsert_all(
+      @life_cycle_definitions.map do |definition|
+        {
+          project_id: @project.id,
+          definition_id: definition.id,
+          active:,
+          type: project_type_for_definition_type(definition.type)
```
toy: Wouldn't it be better to have a method like `step_class`, to have instead:
```suggestion
          type: definition.step_class
```

app/controllers/projects/settings/life_cycles_controller.rb
```diff
@@ -0,0 +1,114 @@
...
+  before_action :load_life_cycle_definitions, only: %i[show enable_all disable_all]
+
+  menu_item :settings_life_cycles
+
+  def show; end
+
+  def toggle
+    step_params = params
+                    .require(:project_life_cycle)
+                    .permit(:definition_id, :type)
```
toy: I think it is better to find definition by id, instead of working with id and type

app/controllers/projects/settings/life_cycles_controller.rb
```diff
@@ -0,0 +1,114 @@
...
+  end
+
+  private
+
+  def load_life_cycle_definitions
+    @life_cycle_definitions = Project::LifeCycleStepDefinition.order(position: :asc)
+  end
+
+  def deny_access_on_feature_flag
+    deny_access unless OpenProject::FeatureDecisions.stages_and_gates_active?
```
toy: Shouldn't it be `deny_access(not_found: true)`? It is not available to anyone, so not there

app/controllers/projects/settings/life_cycles_controller.rb
```diff
@@ -0,0 +1,114 @@
...
+  def project_type_for_definition_type(definition_type)
+    case definition_type
+    when Project::StageDefinition.name
+      Project::Stage
+    when Project::GateDefinition.name
+      Project::Gate
+    else
+      raise NotImplementedError, "Unknown life cycle definition type"
+    end
+  end
```
toy: Not needed with a method of definition returning the desired class

app/views/projects/settings/life_cycles/show.html.erb
```diff
@@ -0,0 +1,33 @@
...
+along with this program; if not, write to the Free Software
+Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+
+See COPYRIGHT and LICENSE files for more details.
+
+++#%>
+<div data-turbo="true">
+  <%= render Projects::Settings::LifeCycles::ShowPageHeaderComponent.new(project: @project) %>
+
+  <%= render(Projects::Settings::LifeCycles::IndexComponent.new( project: @project, life_cycle_definitions: @life_cycle_definitions)) %>
```
toy: I need to get to enabling erb_lint in the project
```suggestion
  <%= render Projects::Settings::LifeCycles::IndexComponent.new(project: @project, life_cycle_definitions: @life_cycle_definitions) %>
```

db/migrate/20241127161228_grant_select_project_life_cycle_permission.rb
```diff
@@ -0,0 +1,7 @@
+require_relative "migration_utils/permission_adder"
+
+class GrantSelectProjectLifeCyclePermission < ActiveRecord::Migration[7.1]
+  def up
+    ::Migration::MigrationUtils::PermissionAdder.add(:edit_project, :select_project_life_cycle)
```
toy, line 5: Just a remark, probably not for this PR - would be better to have at least `having:` named argument

db/migrate/20241125161226_unique_index_on_project_life_cycle_steps.rb
```diff
@@ -0,0 +1,8 @@
+class UniqueIndexOnProjectLifeCycleSteps < ActiveRecord::Migration[7.1]
+  def change
+    add_index :project_life_cycle_steps,
+              %i[project_id definition_id],
+              unique: true,
+              name: "index_project_life_cycle_steps_on_project_id_and_definition_id"
```
toy: I assume default index name is explicitly specified to be able to search for it, but maybe better to add "unique" to the name?

modules/backlogs/spec/support/pages/projects/settings/backlogs.rb
```diff
@@ -0,0 +1,49 @@
...
+# See COPYRIGHT and LICENSE files for more details.
+# ++
+
+require "support/pages/page"
+
+module Pages
+  module Projects
+    module Settings
+      class Backlogs < Pages::Page
+        attr_accessor :project
```
toy: I think `attr_reader` would be better, as assigning is not used

spec/support/pages/projects/settings/general.rb
```diff
@@ -0,0 +1,49 @@
...
+# See COPYRIGHT and LICENSE files for more details.
+# ++
+
+require "support/pages/page"
+
+module Pages
+  module Projects
+    module Settings
+      class General < Pages::Page
+        attr_accessor :project
```
toy: I think `attr_reader` would be better

spec/support/pages/projects/settings/life_cycle.rb
```diff
@@ -0,0 +1,100 @@
...
+# See COPYRIGHT and LICENSE files for more details.
+# ++
+
+require "support/pages/page"
+
+module Pages
+  module Projects
+    module Settings
+      class LifeCycle < Pages::Page
+        attr_accessor :project
```
toy: I think `attr_reader` would be better

spec/support/pages/projects/settings/life_cycle.rb
```diff
@@ -0,0 +1,100 @@
...
+        end
+
+        def path
+          "/projects/#{project.identifier}/settings/life_cycle"
+        end
+
+        # Checks if the life cycle steps are listed in the order given and with the correct toggle state.
+        # @param life_cycle_definitions [Hash{LifeCycleElement => Boolean}]
+        def expect_listed(**life_cycle_definitions)
+          if life_cycle_definitions.size > 1
```
toy: No need for the check before `each_cons`

spec/support/pages/projects/settings/life_cycle.rb
```diff
@@ -0,0 +1,100 @@
...
+          self.project = project
+        end
+
+        def path
+          "/projects/#{project.identifier}/settings/life_cycle"
+        end
+
+        # Checks if the life cycle steps are listed in the order given and with the correct toggle state.
+        # @param life_cycle_definitions [Hash{LifeCycleElement => Boolean}]
+        def expect_listed(**life_cycle_definitions)
```
toy: As this check is used to validate the result of filter, I think it is better to find all row elements and check that they match the order and active status of passed definitions

spec/features/projects/life_cycle/active_in_project_spec.rb
```diff
@@ -0,0 +1,135 @@
...
+                                            ready_to_execute_gate => true,
+                                            executing_stage => true,
+                                            ready_to_close_gate => true,
+                                            closing_stage => true)
+
+      # The user can filter the life cycle steps
+      project_life_cycle_page.filter_by("Clo")
+
+      project_life_cycle_page.expect_listed(ready_to_close_gate => true,
+                                            closing_stage => true)
```
toy: This doesn't check that others are not visible

spec/features/projects/life_cycle/active_in_project_spec.rb
```diff
@@ -0,0 +1,135 @@
...
+      # Activate the stages to be found within the project
+      project_life_cycle_page.toggle(initiating_stage)
+      project_life_cycle_page.toggle(ready_to_close_gate)
+      project_life_cycle_page.toggle(closing_stage)
+
+      wait_for_network_idle
+
+      # Expect the activation state to be kept after a reload
+      visit home_path
+      project_life_cycle_page.visit!
```
toy: Will `page.refresh` or `visit current_path` not work?

spec/features/projects/life_cycle/active_in_project_spec.rb
```diff
@@ -0,0 +1,135 @@
...
+                                            ready_to_close_gate => false,
+                                            closing_stage => false)
+
+      # Activate the stages to be found within the project
+      project_life_cycle_page.toggle(initiating_stage)
+      project_life_cycle_page.toggle(ready_to_close_gate)
+      project_life_cycle_page.toggle(closing_stage)
+
+      wait_for_network_idle
+
```
toy, line 81: Feels odd that state is checked only after reloading the page

toy (commented):
Looks good, mostly minor notes

app/controllers/projects/settings/life_cycles_controller.rb
```diff
@@ -0,0 +1,114 @@
...
+  def enable_all
+    upsert_all_steps(active: true)
+
+    redirect_to action: :show
+  end
+
+  private
+
+  def load_life_cycle_definitions
+    @life_cycle_definitions = Project::LifeCycleStepDefinition.order(position: :asc)
```
toy: I've added `default_scope { order(:position) }` in the PR for global stage admin

app/components/projects/life_cycle_type_component.html.erb
ulferts: Yes, that seems to be necessary as without that alignment specification the text isn't in the middle of the icon:

image

with it, icon and text are aligned:

image

spec/features/projects/life_cycle/active_in_project_spec.rb
ulferts: Thanks for that pointer. It uncovered a bug in the filtering I relied on.

spec/features/projects/life_cycle/active_in_project_spec.rb
ulferts: Refresh will of course work as such. But I am uncertain that the reload actually happened before the subsequent check is carried out. I haven't looked into the code so I am not 100% on this but I wouldn't expect either of the two methods to actually wait before then carrying out the `expect_listed`.  I moved the reloading into a method of the helper page that also checks explicitly that the home page is loaded in between.

db/migrate/20241127161228_grant_select_project_life_cycle_permission.rb
ulferts: I agree. I had to look up the meaning of the variables.

app/components/projects/settings/life_cycles/step_component.html.erb
ulferts: Yes, that works for me without problems in my local lookbook instance:

http://localhost:3000/lookbook/inspect/OpenProject/Projects/life_cycle_type/stage

app/controllers/projects/settings/life_cycles_controller.rb
ulferts: Good advice. It comes at the cost of having to load the definition in the controller when only a single step is toggled but that should still be good enough. 

It allowed to simplify a lot of the other parts you remarked on. Especially, it allowed getting rid of two almost identical methods for upserting.

config/routes.rb
ulferts: Not sure about that. Admittedly, it is semantics. To me using project_life_cycles is distinctive enough and I'd expect that the distinction between step_definitions and steps is only relevant from a technical perspective. To the user I would assume it is actually somewhat of the same. So with the url within the project being `projects/:id/settings/life_cycle` it makes sense to me that the url within the administration would be `admin/settings/project_life_cycles`. Ideally, I would go for `admin/projects/life_cycles` but that is a different story.

config/routes.rb
ulferts: To me the project only has a single life cycle. So from my point of view, we either keep it like that, or go for an index action with `life_cycle_steps`. 

But similar to my argument above, I think that users will refer to the content as the life cycle and not as the life cycle steps. So to me, it is not the individual steps that a user will refer to but rather the whole group will be seen as one entity.

app/controllers/projects/settings/life_cycles_controller.rb
ulferts: This is probably a question of believe as there are good arguments and bad for/against using default scopes. I am aware that e.g. in meetings, the default order is currently being removed. In my book a default order is not as bad as other default scopes which can seriously mess up a day. But here I am not as decided.

My argument for keeping an explicit order basically boils down to:
* It is explicit. So when you read the code, you are not surprised as to what happens.

But I would be open to removing it as well (in the other PR).

ulferts:
Thanks for the thorough review, @toy . It sparked quite some improvements I think. I fear this lead to additional code to review.

I simply resolved all the conversations which I think to be addressed. Those where I am of a different opinion I left open and commented on.

ulferts:
The flickering spec is addressed by #17412

app/components/projects/settings/project_custom_field_sections/index_page_header_component.html.erb
```diff
@@ -1,8 +1,7 @@
 <%= render Primer::OpenProject::PageHeader.new do |header| %>
   <%= header.with_title(variant: :default) { t('projects.settings.project_custom_fields.header.title') } %>
```
toy: As you changed the description key, I suggest to also remove the `(variant: :default)`

config/routes.rb
toy: I agree that having all words of life cycle step definitions is to much, though we can use `path: 'life_cycle'` option to have internally complete identification of path and sane url

config/routes.rb
toy: I suggest `resources :life_cycle_steps, only: %i[index], path: 'life_cycle'`. My reason is that rails intention for `resource` is when there is one object, so `show` displaying a list feels confusing

db/migrate/20241127161228_grant_select_project_life_cycle_permission.rb
toy: https://community.openproject.org/wp/60020

spec/features/projects/life_cycle/active_in_project_spec.rb
toy: I pointed this, as going to home page looks unneeded and extends the duration of the test. Quickly checking the code, told me that both `visit` and `refresh` are not checking if page load is finished

app/controllers/projects/settings/life_cycles_controller.rb
toy: Hm, my argument is expectation, for models that have very explicit order I expect it to be used by default and it looks unexpected when it is explicitly set

app/views/projects/settings/life_cycles/show.html.erb
```diff
@@ -0,0 +1,34 @@
...
+GNU General Public License for more details.
+
+You should have received a copy of the GNU General Public License
+along with this program; if not, write to the Free Software
+Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+
+See COPYRIGHT and LICENSE files for more details.
+
+++#%>
+<div>
```
toy: Is `div` it needed?

app/components/projects/settings/project_custom_field_sections/custom_field_row_component.html.erb
```diff
@@ -5,7 +5,7 @@
     }) do |custom_field_container|
       custom_field_container.with_column(flex_layout: true) do |title_container|
         title_container.with_column(pt: 1, mr: 2) do
-          render(Primer::Beta::Text.new) do
+          render(Primer::Beta::Text.new(classes: 'filter-target-visible-text')) do
```
toy, line 8: This removes a probably unintentional ability to search by type, by filtering for `string`, `bool`, …

app/components/projects/settings/life_cycles/step_component.html.erb
```diff
@@ -0,0 +1,30 @@
+<%=
+  flex_layout(align_items: :center,
+              justify_content: :space_between) do |step_container|
+    step_container.with_column(flex_layout: true) do |title_container|
+      title_container.with_column(pt: 1, mr: 3) do
+        render(Primer::Beta::Text.new(classes: 'filter-target-visible-text')) { definition.name }
```
toy: Removes ability to filter by step type, not sure such unintentional feature should be there or not

app/components/projects/settings/life_cycles/show_page_header_component.html.erb
```diff
@@ -0,0 +1,9 @@
+<%= render Primer::OpenProject::PageHeader.new do |header| %>
+  <%= header.with_title { t('projects.settings.life_cycle.header.title') } %>
+  <%= header.with_description do
+    t('projects.settings.life_cycle.header.description_html',
+      overview_url: project_path(project),
+      admin_settings_url: admin_settings_project_life_cycles_path)
+  end %>
+  <%= header.with_breadcrumbs(breadcrumb_items) %>
+<% end %>
```
toy: Can be simplified to:
```suggestion
<%= render Primer::OpenProject::PageHeader.new do |header|
  header.with_title { t('projects.settings.life_cycle.header.title') }
  header.with_description do
    t('projects.settings.life_cycle.header.description_html',
      overview_url: project_path(project),
      admin_settings_url: admin_settings_project_life_cycles_path)
  end
  header.with_breadcrumbs(breadcrumb_items)
end %>
```

app/components/projects/settings/life_cycles/step_component.rb
```diff
@@ -0,0 +1,53 @@
...
+
+        options :definition,
+                :project
+
+        def active_in_project?
+          project
+            .life_cycle_steps
+            .detect { |project_lc| project_lc.definition_id == definition.id }
+            &.active
+        end
```
toy: Looks like I forgot to write about this - feels not good that all steps will be iterated for every step/row, maybe better to find definitions and matching steps in index component, and pass both to the every step component instead? But not sure if it will be better

app/components/projects/settings/life_cycles/show_page_header_component.html.erb
ulferts: Well spotted, Thanks

app/components/projects/settings/life_cycles/step_component.html.erb
ulferts: Yes it does, but it also removes the weird behaviour that when searching for 'ing'  or 'on' all the rows were kept because that was part of row because each row has that listed (sometimes not visible) as well. For now, I would think this is acceptable. 

If more filtering is desired, we should change it to a backend side filtering that can properly define what to search for.

app/views/projects/settings/life_cycles/show.html.erb
ulferts: No, it isn't. I removed it. Thanks.

app/components/projects/settings/project_custom_field_sections/custom_field_row_component.html.erb
ulferts: This seems to be a [duplicate comment](https://github.com/opf/openproject/pull/17200#discussion_r1878488884)

app/controllers/projects/settings/life_cycles_controller.rb
ulferts: I would like to leave it in for now then. Once the other PR is moved, I would be ok with removing the order here (I still lean towards leaving it). But currently, it would need to be kept in here to ensure the order.

config/routes.rb
ulferts: I took up the idea of using a path option but settled for `project_life_cycles` given that I could imagine other life cycles as well. Again, ideally I would go for `admin/projects/life_cycles` but since the other project specific routes aren't in line with that, I think it prudent to keep the `project_` prefix.

app/components/projects/settings/life_cycles/step_component.rb
ulferts: The problem here is that it is not guaranteed that a corresponding `life_cycle_step` exists for every definition. We don't seed them on project & definition creation. Otherwise it would be a good approach to work through that list instead of through the definitions.   

An approach would be to build the non existing steps in the index action and then pass that step into the step component. Would definitely improve the interface as only the step would have to be passed into the `StepComponent`. The `definition` could then be accessed from the step within the component. This all comes at the cost of increased complexity in the index component.

Performancewise, I'd not expect it to be better though. It exchanges a non optimal loop for AR model instantiation. As there are probably not many steps, rather sure there are less than 20 and very sure there are less than 100, I would also expect the impact of the non optimal loop to be minimal.

ulferts:
I worked in or commented on your remarks @toy. Please take another look.

app/components/projects/settings/life_cycles/step_component.html.erb
toy: It can be an option to mark also type element with the class and handle multiple, but probably more correct to not support this unintentional feature

app/components/projects/settings/life_cycles/step_component.rb
toy: Absent steps can be also simply nils, or only boolean for active state can be passed.

```rb
def life_cycle_definitions_and_step_active
  active_ids = project.life_cycle_steps.where(active: true).pluck(:definition_id).to_set

  life_cycle_definitions.all.map do |definition|
    [definition.id, definition.id.in?(active_ids)]
  end
end

# …

life_cycle_definitions_and_step_active.each do |definition, active|
  # …
  Projects::Settings::LifeCycles::StepComponent.new(project:, definition:, active?: active))
  # …
end
```

ulferts: Moved it. I didn't need the project in the step component so the change is one parameter for one which is nice.

config/routes.rb
ulferts: It is now implemented like that. This comes at the cost of the life_cycle_steps being the only index action among `project/settings/...`. So we should eventually move the other controllers' actions as well.

ulferts:
Again, @toy, I would request a review from your side.

app/controllers/projects/settings/life_cycle_steps_controller.rb
```diff
@@ -26,33 +26,33 @@
...
 
-  menu_item :settings_life_cycles
+  menu_item :settings_life_cycle_steps
 
-  def show; end
+  def index; end
 
   def toggle
-    definition = Project::LifeCycleStepDefinition.where(id: params[:definition_id])
+    definition = Project::LifeCycleStepDefinition.where(id: params[:id])
```
toy, line 41: Very much nitpick - value is still an array, so there may be a bit of confusion why relation is assigned to variable with singular name

config/initializers/menus.rb
```diff
@@ -639,9 +643,9 @@
 
   project_menu_items.each do |key, options|
     menu.push :"settings_#{key}",
-              { controller: "/projects/settings/#{key}", action: "show" },
+              { controller: "/projects/settings/#{key}", action: "show" }.merge(options.slice(:action)),
```
toy, line 646:
```suggestion
              { controller: "/projects/settings/#{key}", action: "show", **options.slice(:action) },
```

config/locales/en.yml
```diff
@@ -2592,7 +2592,7 @@ en:
   label_none_parentheses: "(none)"
   label_not_contains: "doesn't contain"
   label_not_equals: "is not"
-  label_life_cycle_plural: "Project lifecycle"
+  label_life_cycle_step_plural: "Project lifecycle"
```
toy, line 2595:
```suggestion
  label_life_cycle_steps: "Project lifecycle"
```

config/routes.rb
toy: I didn't notice that it will stand out in this way, but I still think it is better

toy (approved):
Looks good to me

config/initializers/menus.rb
ulferts: I find that harder to understand the the merge command.

app/controllers/projects/settings/life_cycle_steps_controller.rb
ulferts: The variable name is correct in being singular since only a single element will ever be returned. So I find that appropriate. One could switch to using a `find` and wrap that into an array explicitly but that I find too cumbersome.

config/locales/en.yml
ulferts: I changed that explicitly after noticing that a translation was no longer applied. I currently don't remember which component but there is a hard wired expectation that there is a label with a plural suffix. Otherwise I wouldn't even have put it into the way to nondescript label section.

PR #17201, opened 2024-11-13
move creation of angular boostrap __ng2-bootstrap-has-run mark, so it is set after turbo:load too

toy:
The `__ng2-bootstrap-has-run` class gets lost when turbo replaces body, so it should be set after `turbo:load` too. Without it `expect_angular_frontend_initialized` will randomly fail depending on timing, so whether it manages the presence of the class check before turbo replaced body or not.

# Merge checklist

- [ ] Added/updated tests

### DIFF
--- a/frontend/src/app/app.module.ts
+++ b/frontend/src/app/app.module.ts
@@ -414,6 +414,8 @@ export class OpenProjectModule implements DoBootstrap {
     if (root) {
       appRef.bootstrap(ApplicationBaseComponent, root);
     }
+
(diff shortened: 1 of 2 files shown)


oliverguenther (approved):
Good catch. I wonder if we could rid of that check altogether though

toy:
@oliverguenther Do you mean that if we don't need the check at all (https://github.com/opf/openproject/pull/17206 waiting for resolution for failure of non feature test to see) or that it can be done by checking something instead of setting a class on body (I spent some time looking for answer while creating this PR, but was not successful)

PR #17211, opened 2024-11-15
Bug/58738 work package creation event not displayed if aggregation includes a comment

jjabari-op:
[Ticket](https://community.openproject.org/projects/communicator-stream/work_packages/58738)

### DIFF
--- a/app/components/work_packages/activities_tab/journals/item_component.html.erb
+++ b/app/components/work_packages/activities_tab/journals/item_component.html.erb
@@ -30,6 +30,16 @@
                                                     classes: "work-packages-activities-tab-journals-item-component--user-name ellipsis hidden-for-mobile") do
                     truncated_user_name(journal.user)
                   end
+                  if journal.initial?
+                      header_start_container.with_column(
+                        mr: 1,
+                        classes: "work-packages-activities-tab-journals-item-component-details--journal-type hidden-for-mobile"
+                      ) do
+                        render(Primer::Beta::Text.new(font_size: :small, color: :subtle, mt: 1)) do
+                          I18n.t("activities.work_packages.activity_tab.created_on")
+                        end
+                      end
+                    end
                   header_start_container.with_column(mr: 1, classes: "hidden-for-mobile") do
                     render(Primer::Beta::Text.new(font_size: :small, color: :subtle, mt: 1)) { format_time(journal.updated_at) }
                   end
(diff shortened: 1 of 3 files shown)


app/components/work_packages/activities_tab/journals/item_component.html.erb
```diff
@@ -30,6 +30,16 @@
                                                     classes: "work-packages-activities-tab-journals-item-component--user-name ellipsis hidden-for-mobile") do
                     truncated_user_name(journal.user)
                   end
+                  if journal.initial?
+                      header_start_container.with_column(
+                        mr: 1,
+                        classes: "work-packages-activities-tab-journals-item-component-details--journal-type hidden-for-mobile"
```
akabiru, line 36: 🟠 Is this meant to be hidden in mobile?

akabiru (changes requested):
Unfortunately, this doesn't seem to resolve the issue- at least from my tests, I saw the same behavior

| Mobile | Desktop |
|--------|--------|
| Screenshot 2024-11-18 at 10 55 59 AM | Screenshot 2024-11-18 at 10 55 44 AM |

jjabari-op:
Hi @akabiru! Thank you for your review. Could you double check with this statement in the ticket: "The comment and the created event share one single event if aggregated"? You're expecting a standalone creation event item in the timeline additionally to the comment, right? But I think the visual merge is correct, it was just missing the "created this on" IMO.

> Unfortunately, this doesn't seem to resolve the issue- at least from my tests, I saw the same behavior
> 
> Mobile	Desktop
> Screenshot 2024-11-18 at 10 55 59 AM	Screenshot 2024-11-18 at 10 55 44 AM

akabiru:
> Hi @akabiru! Thank you for your review. Could you double check with this statement in the ticket: "The comment and the created event share one single event if aggregated"? You're expecting a standalone creation event item in the timeline additionally to the comment, right? But I think the visual merge is correct, it was just missing the "created this on" IMO.
> 
> > Unfortunately, this doesn't seem to resolve the issue- at least from my tests, I saw the same behavior
> > Mobile	Desktop
> > Screenshot 2024-11-18 at 10 55 59 AM	Screenshot 2024-11-18 at 10 55 44 AM

Ah, you're right @jonasjabari - I had the wrong expectation. I see the "created this on" text. I somewhat expected the standalone "+ User created" journal item.

akabiru (approved):

app/components/work_packages/activities_tab/journals/item_component.html.erb
akabiru: @jonasjabari I'm not sure I understand the reason for the `.hidden-for-mobile` class here

jjabari-op: Yes, it we don't have enough width for it on mobile

PR #17212, opened 2024-11-15
[chore] make flickering disappear

Kharonus:
- unflicker InsertItemContract test

- use test locale for localisation specs to avoid errors due to lazy i18n loading

### DIFF
--- a/spec/contracts/custom_fields/hierarchy/insert_item_contract_spec.rb
+++ b/spec/contracts/custom_fields/hierarchy/insert_item_contract_spec.rb
@@ -67,15 +67,27 @@ RSpec.describe CustomFields::Hierarchy::InsertItemContract do
       it "is invalid" do
         result = subject.call(params)
         expect(result).to be_failure
-        expect(result.errors.to_h).to include(label: ["must be unique within the same hierarchy level."])
+        expect(result.errors.to_h).to include(label: [I18n.t("dry_validation.errors.rules.label.not_unique")])
       end
 
-      context "if locale is set to 'de'" do
+      context "if another locale is set" do
+        let(:mordor) { "agh burzum-ishi krimpatul" }
+
+        before do
+          I18n.config.enforce_available_locales = false
+          I18n.backend.store_translations(
+            :mo,
+            { dry_validation: {
+              errors: { rules: { label: { not_unique: mordor } } }
+            } }
+          )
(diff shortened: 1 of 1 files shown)


brunopagno (approved):
:tada:

apfohl (approved):

spec/contracts/custom_fields/hierarchy/insert_item_contract_spec.rb
```diff
@@ -67,15 +67,27 @@
...
-        expect(result.errors.to_h).to include(label: ["must be unique within the same hierarchy level."])
+        expect(result.errors.to_h).to include(label: [I18n.t("dry_validation.errors.rules.label.not_unique")])
       end
 
-      context "if locale is set to 'de'" do
+      context "if another locale is set" do
+        let(:mordor) { "agh burzum-ishi krimpatul" }
+
+        before do
+          I18n.config.enforce_available_locales = false
```
ulferts, line 77: This change will most likely (didn't check) spill over to the specs running after this one. So it should be reverted in an `after` block.

Kharonus: sure, thanks for being very vigilant

spec/contracts/custom_fields/hierarchy/insert_item_contract_spec.rb
```diff
@@ -67,15 +67,31 @@
...
+            :mo,
+            { dry_validation: {
+              errors: { rules: { label: { not_unique: mordor } } }
+            } }
+          )
+        end
+
+        after do
+          I18n.config.enforce_available_locales = true
+        end
```
brunopagno, line 88: Now that I see the comment above, this will also leak, no? Would it be better to store the original value and reassign it here, or is that overkill?

Kharonus: Well, this is the original one. It is also the default. Not sure, if we should bother with variable assignment.

brunopagno: It's all good, you get to make a call. My idea was to be a bit extra defensive. If we (or Rails) ever decides to make a change to this default :upside_down_face: which is very unlikely.

PR #17213, opened 2024-11-15
make sure limit_self_registration is included in saml provider hash

machisuji:
WP [#59375](https://community.openproject.org/projects/openproject/work_packages/59375/activity)


toy (approved):
Code looks correct from my perspective, I only suggest a test that gets false in the hash when it is set so in the model

PR #17214, opened 2024-11-15
[#59344] fix copy folder query to fetch correct id

Kharonus:
# Ticket
[OP#59344](https://community.openproject.org/wp/59344)

- copying projects should also correctly copy managed project folders

- remove trailing slashes before getting id from result (hacky)
- need OP#57850 to make this more robust


brunopagno (approved):
:+1:

PR #17223, opened 2024-11-18
Implementation/59288 add stages and gates to the project overview page

dombesz:
# Ticket
https://community.openproject.org/wp/59288

Display Stages and Gates on the Project Overview page

## Screenshots

Similar implementation to ProjectCustomFields, as there are a lot of similarities.

# Merge checklist

- [x] Added/updated tests


config/locales/en.yml
```diff
@@ -1186,16 +1189,38 @@ en:
...
+              end_date_not_allowed: "Cannot assign end date to a Project::Gate"
+            type:
+              type_and_class_name_mismatch: "must be a Project::Gate"
+            date:
+              non_continuous_dates: "can’t be earlier than the previous %{step}'s end date."
+        project/stage:
+          attributes:
+            date_range:
+              start_date_must_be_before_end_date: "start date must be before the end date."
+              non_continuous_dates: "can’t be earlier than the previous %{step}'s end date."
```
EinLama: same as above: nitpick: two different styles for `'` in this sentence. `can't` has a different apostrophe than `%{step}'s`. Is this by choice or chance? AFAIK it should be `'` everywhere.

config/locales/en.yml
```diff
@@ -1186,16 +1189,38 @@ en:
...
+              type_and_class_name_mismatch: "must match the instantiated class name"
         project/gate:
           attributes:
             base:
-              end_date_not_allowed: "Cannot assign `end_date` to a Project::Gate"
+              end_date_not_allowed: "Cannot assign end date to a Project::Gate"
+            type:
+              type_and_class_name_mismatch: "must be a Project::Gate"
+            date:
+              non_continuous_dates: "can’t be earlier than the previous %{step}'s end date."
```
EinLama: nitpick: two different styles for `'` in this sentence. `can't` has a different apostrophe than `%{step}'s`. Is this by choice or chance? AFAIK it should be `'` everywhere.

app/models/project/stage.rb
```diff
@@ -29,5 +29,36 @@
...
+  def not_set?
+    start_date.blank? || end_date.blank?
+  end
+
+  def range_set?
+    !not_set?
+  end
+
+  def range_incomplete?
+    start_date.blank? ^ end_date.blank?
```
EinLama, line 54: Clever use of XOR 👍🏻

app/models/project/stage.rb
```diff
@@ -29,5 +29,36 @@
 class Project::Stage < Project::LifeCycleStep
   # This ensures the type cannot be changed after initialising the class.
   validates :type, inclusion: { in: %w[Project::Stage], message: :must_be_a_stage }
-  validates :start_date, :end_date, presence: true
+  validate :validate_date_range
+
+  def working_days_count
+    return nil if not_set?
+
+    Day.working.from_range(from: start_date, to: end_date).count
```
EinLama, line 37: This would be much more complicated if we had the option of setting work-week-days per project (or system-wide) 😮

frontend/src/turbo/turbo-global-listeners.ts
```diff
@@ -64,4 +64,13 @@ export function addTurboGlobalListeners() {
     activateFlashNotice();
     activateFlashError();
   });
+
+  document.addEventListener('turbo:before-morph-element', (event) => {
+    const element = event.target as HTMLElement;
+
+    // In case the element is an OpenProject custom dom element, morphing is prevented.
+    if (element.tagName.startsWith('OPCE-')) {
```
EinLama: If the HTML spec ever changes, we encounter a wonky Browser, or if we ever migrate to XHTML, the tagName might not be all-uppercase, see [MDN](https://developer.mozilla.org/en-US/docs/Web/API/Element/tagName#examples). While unlikely, I suggest ruling out this hard-to-debug edge-case and safeguard this condition by always converting the tagName to uppercase, just in case.

```suggestion
    if (element.tagName.toUpperCase().startsWith('OPCE-')) {
```

lib/primer/open_project/forms/date_picker.html.erb
```diff
@@ -0,0 +1,24 @@
+<%= render(FormControl.new(input: @input, tag: :"primer-datepicker-field")) do %>
```
EinLama, line 30: Consider adding the Copyright comment above here.

lib/primer/open_project/forms/date_picker.rb
```diff
@@ -0,0 +1,19 @@
+# frozen_string_literal: true
```
EinLama, line 1: Consider adding the Copyright comment here :)

lib/primer/open_project/forms/dsl/range_date_picker_input.rb
```diff
@@ -0,0 +1,21 @@
+# frozen_string_literal: true
```
EinLama, line 1: Consider adding the Copyright comment here :)

lib/primer/open_project/forms/dsl/single_date_picker_input.rb
```diff
@@ -0,0 +1,33 @@
+# frozen_string_literal: true
```
EinLama, line 1: Consider adding the Copyright comment here :)

modules/overviews/app/components/project_life_cycles/sections/edit_component.rb
```diff
@@ -0,0 +1,41 @@
...
+
+module ProjectLifeCycles
+  module Sections
+    class EditComponent < ApplicationComponent
+      include ApplicationHelper
+      include OpTurbo::Streamable
+      include OpPrimer::ComponentHelpers
+
+      def life_cycle_steps
+        model.life_cycle_steps.active.eager_load(:definition).order(position: :asc)
```
EinLama: You could re-use previous logic here:

```suggestion
        model.available_life_cycle_steps
```

modules/overviews/app/components/project_life_cycles/sections/edit_dialog_component.html.erb
```diff
@@ -0,0 +1,32 @@
...
+                                   id: "edit-project-life-cycles-dialog")) do |d|
+    d.with_header(variant: :large)
+    d.with_body(classes: "Overlay-body_autocomplete_height") do
+      render(::ProjectLifeCycles::Sections::EditComponent.new(model))
+    end
+    d.with_footer do
+      component_collection do |footer_collection|
+        footer_collection.with_component(Primer::ButtonComponent.new(
+          data: {
+            'close-dialog-id': "edit-project-life-cycles-dialog"
```
EinLama: Since this ID refers to the Dialog id defined in line 4, it would make sense to move this into a constant/method within the component. That way, if the ID changes in the future, you only need to change it in one place.

e.g.

```ruby
module ProjectLifeCycles
  module Sections
    class EditComponent
       def dialog_id
         "edit-project-life-cycles-dialog"
       end
# ...

# here:
'close-dialog-id': dialog_id
```

modules/overviews/app/components/project_life_cycles/sections/project_life_cycles/show_component.rb
```diff
@@ -0,0 +1,79 @@
...
+            end
+          when Project::Stage
+            render(Primer::Beta::Text.new) do
+              concat [
+                helpers.format_date(model.start_date),
+                helpers.format_date(model.end_date)
+              ].join(" - ")
+            end
+          end
+        end
```
EinLama: This could be simplified so that you don't have to check which type the model is.

```suggestion
        def render_value
          render (Primer::Beta::Text.new) do
            concat [
              model.start_date,
              model.end_date
            ]
            .compact
            .map { |d| helpers.format_date(d) }
            .join(" - ")
          end
        end
```

modules/overviews/app/components/project_life_cycles/sections/show_component.rb
```diff
@@ -0,0 +1,51 @@
...
+      include ApplicationHelper
+      include OpPrimer::ComponentHelpers
+      include OpTurbo::Streamable
+
+      def initialize(project:)
+        super
+
+        @project = project
+        @life_cycle_steps =
+          @project.life_cycle_steps.active.eager_load(:definition).order(position: :asc)
```
EinLama:
```suggestion
        @life_cycle_steps = @project.available_life_cycle_steps
```

modules/overviews/app/components/project_life_cycles/sections/show_component.html.erb
```diff
@@ -0,0 +1,34 @@
...
+            controller: 'async-dialog'
+          },
+          test_selector: "project-life-cycles-edit-button",
+          aria: { label: I18n.t(:label_edit) }
+        )
+      end
+
+      flex_layout do |details_container|
+        @life_cycle_steps.each_with_index do |life_cycle_step, i|
+          margin =  i == @life_cycle_steps.size - 1 ? 0 : 3
```
EinLama: bonus space

```suggestion
          margin = i == @life_cycle_steps.size - 1 ? 0 : 3
```

modules/overviews/app/controllers/overviews/overviews_controller.rb
```diff
@@ -49,6 +76,50 @@ def update_project_custom_values
...
+                     model: @project,
+                     contract_class: ProjectLifeCycleSteps::UpdateContract)
+                .call(permitted_params.project_life_cycles)
+
+      update_via_turbo_stream(
+        component: ProjectLifeCycles::Sections::EditComponent.new(service_call.result),
+        method: "morph"
+      )
+      # TODO: :unprocessable_entity is not nice, change the dialog logic to accept :ok
+      # without dismissing the dialog, alternatively use turbo frames instead of streams.
```
EinLama, line 101: While true, I think it is safe to say that this is outside of the scope of this PR 👍🏻

modules/overviews/lib/overviews/engine.rb
```diff
@@ -47,7 +47,8 @@ class Engine < ::Rails::Engine
         OpenProject::AccessControl.permission(:view_project)
           .controller_actions
           .push(
-            "overviews/overviews/show"
+            "overviews/overviews/show",
+            "overviews/overviews/project_life_cycles_sidebar"
```
EinLama: It surprised me that the side bar is part of `view_project`. If I understand it correctly, you must have both `view_project` and `view_project_stages_and_gates` permissions to see the sidebar.

Isn't this superfluous as you wouldn't see the entire overview page to begin with if you lacked the `view_project` permission? 🤔 

In other words: is it necessary to add this line?

spec/contracts/project_life_cycle_steps/base_contract_spec.rb
```diff
@@ -0,0 +1,153 @@
...
+require "spec_helper"
+require "contracts/shared/model_contract_shared_context"
+
+RSpec.describe ProjectLifeCycleSteps::BaseContract do
+  include_context "ModelContract shared context"
+
+  let(:contract) { described_class.new(project, user) }
+  let(:project) { build_stubbed(:project) }
+
+  context "with authorised user" do
```
EinLama: Americanize:

```suggestion
  context "with authorized user" do
```

spec/contracts/project_life_cycle_steps/base_contract_spec.rb
```diff
@@ -0,0 +1,153 @@
...
+          allow(project).to receive(:valid?)
+
+          contract.validate
+          expect(project).to have_received(:valid?).with(:saving_life_cycle_steps)
+        end
+      end
+    end
+  end
+
+  context "with unauthorised user" do
```
EinLama:
```suggestion
  context "with unauthorized user" do
```

spec/models/project/stage_spec.rb
```diff
@@ -33,13 +33,112 @@
...
+      subject.start_date = Date.parse("2024-11-25")
+      subject.end_date = Date.parse("2024-11-27")
+
+      allow(Day).to receive(:working).and_return(Day)
+      allow(Day).to receive(:from_range)
+                .with(from: subject.start_date, to: subject.end_date)
+                .and_return([])
+
+      expect(subject.working_days_count).to eq(0)
+    end
```
EinLama: Can you explain why you added this spec?

This spec is a bit odd as it tests the internals of the method in question. I wonder if it wouldn't be better to just supply some more test cases with specific date ranges and expected amounts of working days?

Right now, this spec would break if the `working_days_count` method changes its implementation, even if the result returned was the same.

spec/support/pages/projects/show.rb
```diff
@@ -56,7 +54,12 @@ def expect_no_visible_sidebar
         expect(page).to have_no_css(".op-grid-page--grid-container")
       end
 
-      def within_async_loaded_sidebar(&)
+      def expect_visible_sidebar
+        expect_angular_frontend_initialized
+        expect(page).to have_css(".op-grid-page--grid-container")
```
EinLama: I think the CSS class here is wrong 😁 shouldn't it rather be:

```suggestion
        expect(page).to have_css(".op-grid-page--sidebar")
```

EinLama (commented):
Works nicely! I have added some remarks, most of which are small things. I could not find any major flaws, great job!

modules/overviews/lib/overviews/engine.rb
dombesz: Yep, it is superfluous, I think it is just a typo. Thanks!

spec/models/project/stage_spec.rb
dombesz: I agree, the tests became a bit more brittle having this expectation. However, I believe the costs weights the benefits. AFAIK it is ok to mock external objects that are independent from the class, because I'm not mocking the internal methods of this class.

This way I can make sure the right behaviour is called, without duplicating all the tests from the `day_spec.rb`.

EinLama: I see, that makes sense.

With that in mind, we should check if the allowed methods were actually called:
```ruby
      expect(Day).to have_received(:working).with(no_args)
      expect(Day).to have_received(:from_range).with(from: subject.start_date, to: subject.end_date)
```

dombesz: Very good point, for some reason that was left off 😅

dombesz:
Thanks @EinLama  for the quick review, I believe I addressed all your comments.

spec/features/projects/life_cycle/overview_page/dialog/permission_spec.rb
```diff
@@ -0,0 +1,84 @@
...
+
+RSpec.describe "Edit project stages and gates on project overview page", :js, :with_cuprite,
+               with_flag: { stages_and_gates: true } do
+  include_context "with seeded projects and stages and gates"
+  let(:user) { create(:user) }
+  let(:overview_page) { Pages::Projects::Show.new(project) }
+  let(:permissions) { [] }
+
+  before do
+    allow(User).to receive(:current).and_return user
```
EinLama: You could also use the `current_user` [helper](https://github.com/opf/openproject/blob/1d73cab4a8932ccf8ecd40ab4a81f52ddb54313c/spec/support/authentication_helpers.rb#L104) here:

```ruby
current_user { user }
```

spec/features/projects/life_cycle/overview_page/dialog/update_spec.rb
```diff
@@ -0,0 +1,155 @@
...
+RSpec.describe "Edit project stages and gates on project overview page", :js, :with_cuprite,
+               with_flag: { stages_and_gates: true } do
+  include_context "with seeded projects and stages and gates"
+  shared_let(:overview) { create :overview, project: }
+
+  let(:overview_page) { Pages::Projects::Show.new(project) }
+
+  before do
+    # TODO: Could this work for all feature specs?
+    allow(User).to receive(:current).and_return admin
```
EinLama: You could also use the `current_user` [helper](https://github.com/opf/openproject/blob/1d73cab4a8932ccf8ecd40ab4a81f52ddb54313c/spec/support/authentication_helpers.rb#L104) here:

```ruby
current_user { admin }
```

spec/features/projects/life_cycle/overview_page/sidebar_spec.rb
```diff
@@ -0,0 +1,153 @@
...
+        expect(fields.size).to eq(7)
+
+        expect(fields[0].text).to include("Initiating")
+        expect(fields[1].text).to include("Ready for Planning")
+        expect(fields[2].text).to include("Planning")
+        expect(fields[3].text).to include("Ready for Executing")
+        expect(fields[4].text).to include("Executing")
+        expect(fields[5].text).to include("Ready for Closing")
+        expect(fields[6].text).to include("Closing")
+      end
```
EinLama: This could be simplified:

```suggestion
    overview_page.within_life_cycles_sidebar do
      expected_stages = [
        "Initiating",
        "Ready for Planning",
        "Planning",
        "Ready for Executing",
        "Executing",
        "Ready for Closing",
        "Closing"
      ]

      fields = page.all(".op-project-life-cycle-container")
      expect(fields.map(&:text)).to eq(expected_stages)
    end
```

spec/features/projects/life_cycle/overview_page/sidebar_spec.rb
```diff
@@ -0,0 +1,153 @@
...
+      life_cycle_ready_for_executing_definition.move_to_bottom
+
+      overview_page.visit_page
+
+      overview_page.within_life_cycles_sidebar do
+        fields = page.all(".op-project-life-cycle-container")
+
+        expect(fields.size).to eq(7)
+
+        expect(fields[0].text).to include("Initiating")
```
EinLama: Same simplification possible here :)

EinLama (approved):
Nice one 👏🏻

PR #17230, opened 2024-11-19
[#59391] rework file links to paginated collection

Kharonus:
# Ticket
[OP#59391](https://community.openproject.org/wp/59391)

- show again a correct counter on the files tab in work package detail view

- files tab counter now only calls for file links of wp with page size 0
- file links endpoint returns paginated collections
- frontend requests file links with page size -1 (INFINITE)
- file sync is still executed for ALL file links of a work package, not only the requested page
  - this is practically no change for the product, as we do not fetch single pages
  - but fetching only the total numbers with page size 0 now does not trigger a sync

### DIFF
--- a/modules/storages/lib/api/v3/file_links/join_origin_status_to_file_links_relation.rb
+++ b/modules/storages/lib/api/v3/file_links/join_origin_status_to_file_links_relation.rb
@@ -1,5 +1,33 @@
 # frozen_string_literal: true
 
+#-- copyright
+# OpenProject is an open source project management software.
+# Copyright (C) the OpenProject GmbH
+#
+# This program is free software; you can redistribute it and/or
+# modify it under the terms of the GNU General Public License version 3.
+#
+# OpenProject is a fork of ChiliProject, which is a fork of Redmine. The copyright follows:
+# Copyright (C) 2006-2013 Jean-Philippe Lang
+# Copyright (C) 2010-2013 the ChiliProject Team
+#
+# This program is free software; you can redistribute it and/or
+# modify it under the terms of the GNU General Public License
+# as published by the Free Software Foundation; either version 2
+# of the License, or (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program; if not, write to the Free Software
+# Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+#
+# See COPYRIGHT and LICENSE files for more details.
+#++
+
 module API
   module V3
     module FileLinks
@@ -7,8 +35,7 @@ module API
         # @param [Hash] id_status_map A hash mapping file link IDs to their origin status
         # in the format { 137: "view_allowed", 142: "error" }
         def self.create(id_status_map)
-          sanitized_sql = ActiveRecord::Base.send(
-            :sanitize_sql_array,
+          sanitized_sql = ActiveRecord::Base.sanitize_sql_array(
             [origin_status_join(id_status_map.size), *id_status_map.flatten]
           )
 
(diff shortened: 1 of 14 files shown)


brunopagno (approved):
Code changes look good to me. I am assuming the pagination change is considered a non breaking API change and thus can be included without any issues to consumers.

:tada:

Kharonus:
@brunopagno This is an important point. My take on it would be, that usage of it is in the most cases the same. If you do not define a pagesize, you get a standard back. only if you have more then 30 file links in one collection, you suddenly won't get all back.

This could be considered breaking, but the solution for it is quite simple and needs only adding a `pageSize=-1` to the request's query params.

modules/storages/lib/api/v3/file_links/join_origin_status_to_file_links_relation.rb
```diff
@@ -0,0 +1,31 @@
+# frozen_string_literal: true
+
```
apfohl, line 2: You might add the copyright thingy

modules/storages/lib/api/v3/file_links/join_origin_status_to_file_links_relation.rb
```diff
@@ -0,0 +1,31 @@
...
+module API
+  module V3
+    module FileLinks
+      class JoinOriginStatusToFileLinksRelation
+        # @param [Hash] id_status_map A hash mapping file link IDs to their origin status
+        # in the format { 137: "view_allowed", 142: "error" }
+        def self.create(id_status_map)
+          sanitized_sql = ActiveRecord::Base.send(
```
apfohl: Do you need to use `send`? Can't you just call the method? `sanitize_sql_array([...])`

modules/storages/lib/api/v3/file_links/work_packages_file_links_api.rb
```diff
@@ -28,50 +28,78 @@
...
+            return ::Storages::FileLink.none if file_links.empty?
 
-      unless query.valid?
-        message = I18n.t("api_v3.errors.missing_or_malformed_parameter", parameter: "filters")
-        raise ::API::Errors::InvalidQuery.new(message)
-      end
+            sync_result = ::Storages::FileLinkSyncService
+                            .new(user: current_user)
+                            .call(file_links)
+                            .result
```
apfohl, line 42: Will it always succeed?

apfohl (changes requested):

modules/storages/lib/api/v3/file_links/join_origin_status_to_file_links_relation.rb
Kharonus: done

modules/storages/lib/api/v3/file_links/join_origin_status_to_file_links_relation.rb
Kharonus: Uh, interesting. I took this code sample from rubocop's improvements. Will try it.

brunopagno:
> This could be considered breaking, but the solution for it is quite simple and needs only adding a `pageSize=-1` to the request's query params.

@Kharonus the behaviour change _could_ be considered breaking. Do we have any previous events similar to this to take into account? Would be nice if the decision was made for us :sweat_smile: 

I personally do not feel the change is a big deal, but if any consumer depends on this behaviour they could be annoyed, even if the fix on the client side is simple.

modules/storages/lib/api/v3/file_links/join_origin_status_to_file_links_relation.rb
Kharonus: works, good catch!

modules/storages/lib/api/v3/file_links/work_packages_file_links_api.rb
Kharonus: The sync service indeed is always successful. While single storage syncs can fail, the sync service only accumulates successes into its own result.

Kharonus:
Well, TBH in the past we did not really bothered with breaking changes to the API. We renamed properties, changed JSON representations, and still released them without a deprecation round.

For this thing here a deprecation round would be adding this behavior behind some kind of feature flag.

It is possible, but it would raise of course again the question, if the change is breaking enough. It should not happen, that all custom clients suddenly break. And for those, where big collections are an issue, the fix is rather simple.

brunopagno:
@Kharonus sounds good enough to me. Thanks for the context there :bow:

apfohl (approved):

PR #17235, opened 2024-11-20
[59539] Add automatic scheduling mode

cbliard:
# Ticket

https://community.openproject.org/wp/59539

(Part of https://community.openproject.org/wp/42388)

# Pull preview

Available here: https://pr-17235-42388-new-autom-ip-3-70-132-70.my.preview.run/

What works:
* work packages are manually scheduled by default
* work packages being automatically scheduled are always moved to their soonest start date
  * unless they are parent, in which case they follow their children's dates. If children are automatically scheduled, it behaves as expected
* work package switches to automatic mode when becoming a successor
* work package switches to manual mode when no longer a successor

What's next:
* switch scheduling mode to automatic when becoming a parent.

What's missing:
* forbid dates modifications when in automatic scheduling mode
  * disable inputs in date picker
* automatic rescheduling of a successor manually scheduled when switching to automatic
* [x] automatic rescheduling of a successor with 2 predecessor when the closest relation is deleted
  * [ ] automatic switch child of successor from automatic to manual when latest predecedes relation is deleted. successor keeps automatic because it has children

Implement new automatic scheduling mode

Migration:
- [x] Work packages are manually scheduled by default
- [x] Data migration in a separate job
  - [x] All non-successor and non-parent work packages are switched to manual scheduling mode to preserve their dates
  - [x] Closest follows relations have lag adjusted to preserve their dates if a rescheduling happens

Logic changes:
- [X] Parent dates and ignore non-working days attribute are derived from all children regardless of their scheduling mode (before, only automatically scheduled children were considered)
Feature implementation:
- [x] When scheduling happens, a work package is moved to its soonest date.
- [x] When some non-working days are removed, successor need to be rescheduled if the relation from predecessor to successor covers a removed non-working day
- [x] When last predecessor is removed, the successor switches to manual if it has dates, and keeps automatic if it does not have dates (`spec/services/work_packages/set_schedule_service_working_days_spec.rb:374`)

## Screenshots

TBD

TBD

# Merge checklist

- [X] Added/updated tests


cbliard:
Both failing tests are passing locally.

Merging.

PR #17236, opened 2024-11-20
Implementation/59466 change data model to allow start and end times for time entries

klaustopher:
# Ticket
https://community.openproject.org/projects/stream-time-and-costs/work_packages/59466/github

## Screenshots

# Merge checklist

- [ ] Added/updated tests


app/models/journal/base_journal.rb
```diff
@@ -37,7 +37,7 @@ def journaled_attributes
   end
 
   def self.journaled_attributes
-    @journaled_attributes ||= column_names.map(&:to_sym) - excluded_attributes
+    @journaled_attributes ||= columns.reject(&:virtual?).map { |col| col.name.to_sym } - excluded_attributes
```
klaustopher, line 40: I actually planned to add virtual attributes here, but I will do this later. But I think it makes sense in general, when we have virtual columns in the journal stuff, we should not try to back this up

oliverguenther (approved):

PR #17237, opened 2024-11-20
[chore] render item shorts in value selector

Kharonus:
- the hierarchy item value selector does not render shorts yet, but we want them

## Screenshots
image

- the select component renders HalResource arrays
- the rendered string is, what a resource answers on `.name`
- the hierarchy items do not have a name, so they fall back to `self_link.title`
- this PR adds the short to the self link title


brunopagno (approved):
:ship: :ship: :ship:

PR #17247, opened 2024-11-21
[59069] Implemented filtering for hierarchies

brunopagno:
# Ticket
https://community.openproject.org/wp/59069 & https://community.openproject.org/wp/59070

Include filtering for hierarchy items

## Screenshots

https://github.com/user-attachments/assets/c5a51404-6313-465f-8624-f3696c0215fd

We introduced a new filter strategy that uses the already existing equals and notequals operators

# Merge checklist

- [X] Added/updated tests

### DIFF
--- a/app/models/queries/filters/shared/custom_fields/hierarchy.rb
+++ b/app/models/queries/filters/shared/custom_fields/hierarchy.rb
@@ -1,3 +1,5 @@
+# frozen_string_literal: true
+
 #-- copyright
 # OpenProject is an open source project management software.
 # Copyright (C) the OpenProject GmbH
@@ -26,27 +28,37 @@
 # See COPYRIGHT and LICENSE files for more details.
 #++
 
-module Queries::Filters
-  STRATEGIES = {
-    list: Queries::Filters::Strategies::List,
-    list_all: Queries::Filters::Strategies::ListAll,
-    list_optional: Queries::Filters::Strategies::ListOptional,
-    shared_with_user_list_optional: Queries::Filters::Strategies::WorkPackages::SharedWithUser::ListOptional,
-    integer: Queries::Filters::Strategies::Integer,
-    date: Queries::Filters::Strategies::Date,
-    datetime_past: Queries::Filters::Strategies::DateTimePast,
-    string: Queries::Filters::Strategies::String,
-    text: Queries::Filters::Strategies::Text,
-    search: Queries::Filters::Strategies::Search,
-    float: Queries::Filters::Strategies::Float,
-    inexistent: Queries::Filters::Strategies::Inexistent,
-    empty_value: Queries::Filters::Strategies::EmptyValue,
-    hierarchy: Queries::Filters::Strategies::Hierarchy
-  }.freeze
+module Queries
+  module Filters
+    module Shared
+      module CustomFields
+        class Hierarchy < Base
+          def ar_object_filter?
+            true
+          end
 
-  ##
-  # Wrapper class for invalid filters being created
-  class InvalidError < StandardError; end
+          def value_objects
+            CustomField::Hierarchy::Item
+              .where(id: @values)
+              .map { |item| HierarchyItemFilterAdapter.new(item:) }
+          end
+        end
 
-  class MissingError < StandardError; end
+        class HierarchyItemFilterAdapter
+          attr_reader :name
+
+          delegate :id, to: :item
+
+          def initialize(item:)
+            @item = item
+            @name = item.label
+          end
+
+          private
+
+          attr_accessor :item
+        end
+      end
+    end
+  end
 end
--- a/spec/features/work_packages/table/queries/hierarchy_cf_filter_spec.rb
+++ b/spec/features/work_packages/table/queries/hierarchy_cf_filter_spec.rb
@@ -1,0 +1,96 @@
+#-- copyright
+# OpenProject is an open source project management software.
+# Copyright (C) the OpenProject GmbH
+#
+# This program is free software; you can redistribute it and/or
+# modify it under the terms of the GNU General Public License version 3.
+#
+# OpenProject is a fork of ChiliProject, which is a fork of Redmine. The copyright follows:
+# Copyright (C) 2006-2013 Jean-Philippe Lang
+# Copyright (C) 2010-2013 the ChiliProject Team
+#
+# This program is free software; you can redistribute it and/or
+# modify it under the terms of the GNU General Public License
+# as published by the Free Software Foundation; either version 2
+# of the License, or (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program; if not, write to the Free Software
+# Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+#
+# See COPYRIGHT and LICENSE files for more details.
+#++
+
+require "spec_helper"
+
+RSpec.describe "Work package filtering by user custom field", :js do
+  let(:project) { create(:project) }
+  let(:type) { project.types.first }
+  let(:wp_table) { Pages::WorkPackagesTable.new(project) }
+  let(:filters) { Components::WorkPackages::Filters.new }
+  let(:hierarchy_root) { create(:hierarchy_item) }
+  let!(:hierarchy_cf) do
+    create(:hierarchy_wp_custom_field, hierarchy_root:).tap do |cf|
+      type.custom_fields << cf
+      project.work_package_custom_fields << cf
+    end
+  end
+  let(:service) { CustomFields::Hierarchy::HierarchicalItemService.new }
+  let!(:luke) { service.insert_item(parent: hierarchy_root, label: "luke").value! }
+  let!(:leia) { service.insert_item(parent: hierarchy_root, label: "leia").value! }
+
+  let!(:wp_luke) do
+    create(:work_package, project:, subject: "Luke's wp").tap do |wp|
+      wp.custom_field_values = { hierarchy_cf.id => luke.id }
+      wp.save!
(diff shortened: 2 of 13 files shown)


brunopagno:
@apfohl I opened a PR in the current branch. We can for sure create a separate branch and reapply the changes here if you prefer. But alternatively we could keep it simple and just do a `Squash and Merge` to keep the tree clean whenever we decide to merge this one.

apfohl:
It's good. Let's just squash it.

app/models/queries/filters/shared/custom_fields/hierarchy.rb
```diff
@@ -0,0 +1,64 @@
...
+
+          def initialize(item:)
+            @item = item
+            @name = item.label
+          end
+
+          private
+
+          attr_accessor :item
+        end
```
ulferts, line 60: ❓ What is the purpose of this class here?

ulferts (commented):
This was a quick fly-over glance and not a proper review, i.e. I haven't tried out the functionality and didn't spent a long time to think this through. Especially the actual SQL used for filtering would be important to look at in this case. 

To that end it would be helpful to know the expected filter behaviour. E.g. is filtering on a category (encompassing multiple items below) intended to be possible and if so, what is the intended outcome.

An integration spec on the filter would serve to guarantee the quality of the changes as well as explain the expected filter results. https://github.com/opf/openproject/blob/dev/spec/models/query/results_custom_field_filter_integration_spec.rb could be a good place for this. And of course a feature spec to safeguard the whole functionality.

But the structure of the change seems to be going into the right direction.

app/models/queries/filters/shared/custom_fields/hierarchy.rb
brunopagno: @ulferts it's because of the expected values on the [query_filter_instance_representer.rb](https://github.com/opf/openproject/blob/dev/lib/api/v3/queries/filters/query_filter_instance_representer.rb#L76-L100). It seems to be designed to operate on a type of object and we needed the adapter so that it would render correctly.

Same for the path we added to the `path_helper.rb`

Kharonus: Specifically we are talking about the filter representer being very specific about needing a `.name` method on entities. the hierarchy items do not have a name. To not pollute either the record or the representer, an adapter was created, wrapping the hierarchy item into a readable model for the filter representer.

spec/features/work_packages/table/queries/hierarchy_cf_filter_spec.rb
```diff
@@ -0,0 +1,96 @@
...
+# You should have received a copy of the GNU General Public License
+# along with this program; if not, write to the Free Software
+# Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+#
+# See COPYRIGHT and LICENSE files for more details.
+#++
+
+require "spec_helper"
+
+RSpec.describe "Work package filtering by user custom field", :js do
```
Kharonus: 🔴 spec description not correct, copy paste intrusion 😝

spec/features/work_packages/table/queries/hierarchy_cf_filter_spec.rb
```diff
@@ -0,0 +1,96 @@
...
+      wp_table.expect_work_package_listed(wp_luke, wp_leia)
+    end
+
+    context "when any" do
+      it "shows the work package matching the hierarchy cf filter" do
+        filters.open
+
+        # Filtering by hierarchy (=)
+
+        filters.add_filter_by(hierarchy_cf.name, "is", [luke.label], hierarchy_cf.attribute_name(:camel_case))
```
Kharonus, line 76: ❓ ooc: how much effort is it to have a clone of the `is` filter operator which is called `is any`, being more precise in what it is doing?

question opts for maybe suggesting within our current epic scope a renaming (in code) from `is` to `is any`, to easily distinguish between `is any` and `is all` in future.

Kharonus (changes requested):
Pretty cool, that in the end it is indeed relatively small code changes. great code discovery to find all the needed locations.

Kharonus:
@brunopagno you already added a UI test as Jens suggested. For the correct wire up of schemas, representers and all the stuff, this is a nice and valuable thing. But it is not proper (IMHO) for a full functionality test of the filter. For the current operators (is any and is not) this is no big deal, but my take would be, that we write requests specs (API integration) once we have the `is any (with descendants)` for the work package collection, to really cover which work packages are filtered and which not.

requests specs are much faster then UI tests, and hence can be used to cover all relevant functionality of filtering the wp collection.

spec/features/work_packages/table/queries/hierarchy_cf_filter_spec.rb
brunopagno: We initially had implemented a custom operator for hierarchies, but after some investigation it turned out the behaviour would be exactly the same as the [Equals operator](https://github.com/opf/openproject/blob/dev/app/models/queries/operators/equals.rb).

And after looking into other operators, I found that there is already a [EqualsAll operator](https://github.com/opf/openproject/blob/dev/app/models/queries/operators/equals_all.rb).

Which leads me to think that using the default operators here is better because it is more aligned with the rest of the application. Meaning I don't think it is better to explicitly duplicate it for the `is any` label

brunopagno:
> requests specs are much faster then UI tests

@Kharonus that's a very valid point. I think I caught a glimpse of previous discussions around the topic in the past. To be honest at this point I don't have strong opinions, so I would be fine if we want to ditch these feature specs for API tests.

apfohl (approved):

ulferts (commented):
The code change is still fine so nothing changed from my initial, brief glance-over. 

I personally wouldn't have written the detailed specs on the various filters including their operators in a feature spec. Rather, I would have written one feature spec to safeguard the end to end functionality and then have integrational, still somewhat unit specs on the level of checking that the expected results are returned for every operator.

What is amis on the user level is I think the filter value selection which currently is a flat list. 

image

Something that I stumbled over is that the filter results are not as expected (by me). In case I have two work packages and only one is of a type that has the cf enabled. The other doesn't have the cf configured at all. And that one work package with the matching type has a value set. If the filtering then happens with the "is not" operator, the work package that does not have the matching value is not returned. 

The above behaviour however is true for all custom field filters I checked. It still constitutes a bug I think as it goes against the user's expectations.

What is also surprising is the SQL that is generated:

```SQL
SELECT
	"work_packages"."id"
FROM
	"work_packages"
	LEFT OUTER JOIN "projects" ON "projects"."id" = "work_packages"."project_id"
WHERE
	(
		(
			WORK_PACKAGES.ID IN (
				SELECT
					WORK_PACKAGES.ID
				FROM
					WORK_PACKAGES
					LEFT OUTER JOIN CUSTOM_VALUES ON CUSTOM_VALUES.CUSTOMIZED_TYPE = 'WorkPackage'
					AND CUSTOM_VALUES.CUSTOMIZED_ID = WORK_PACKAGES.ID
					AND CUSTOM_VALUES.CUSTOM_FIELD_ID = 541
					JOIN CUSTOM_FIELDS_TYPES ON CUSTOM_FIELDS_TYPES.TYPE_ID = WORK_PACKAGES.TYPE_ID
					AND CUSTOM_FIELDS_TYPES.CUSTOM_FIELD_ID = 541
					JOIN CUSTOM_FIELDS_PROJECTS ON CUSTOM_FIELDS_PROJECTS.PROJECT_ID = WORK_PACKAGES.PROJECT_ID
					AND CUSTOM_FIELDS_PROJECTS.CUSTOM_FIELD_ID = 541
				WHERE
					(
						CUSTOM_VALUES.VALUE IS NULL
						OR CUSTOM_VALUES.VALUE NOT IN ('14')
					)
			)
		)
		AND (PROJECTS.ID IN (6))
	)
	AND (
		WORK_PACKAGES.ID IN (
			SELECT
				WORK_PACKAGES.ID
			FROM
				WORK_PACKAGES
				LEFT OUTER JOIN CUSTOM_VALUES ON CUSTOM_VALUES.CUSTOMIZED_TYPE = 'WorkPackage'
				AND CUSTOM_VALUES.CUSTOMIZED_ID = WORK_PACKAGES.ID
				AND CUSTOM_VALUES.CUSTOM_FIELD_ID = 541
				JOIN CUSTOM_FIELDS_TYPES ON CUSTOM_FIELDS_TYPES.TYPE_ID = WORK_PACKAGES.TYPE_ID
				AND CUSTOM_FIELDS_TYPES.CUSTOM_FIELD_ID = 541
				JOIN CUSTOM_FIELDS_PROJECTS ON CUSTOM_FIELDS_PROJECTS.PROJECT_ID = WORK_PACKAGES.PROJECT_ID
				AND CUSTOM_FIELDS_PROJECTS.CUSTOM_FIELD_ID = 541
			WHERE
				(
					CUSTOM_VALUES.VALUE IS NULL
					OR CUSTOM_VALUES.VALUE NOT IN ('14')
				)
		)
	)
	AND "work_packages"."id" IN (
		SELECT
			"work_packages"."id"
		FROM
			"work_packages"
			INNER JOIN "projects" ON "projects"."id" = "work_packages"."project_id"
		WHERE
			"projects"."active" = TRUE
	)
ORDER BY
	WORK_PACKAGES.ID
``` 

So with the exception of the WHERE statement on the project, the subselect for the custom field values is repeated. This looks like potential for optimizations.

brunopagno:
That's briliant feedback, thanks for making the time @ulferts 

> I would have written one feature spec to safeguard the end to end functionality and then have integrational, still somewhat unit specs on the level of checking that the expected results are returned for every operator.

Fair enough. I am not sure I follow what you're thinking but I assume this aligns more with what @Kharonus was pointing out. Improvements can still be done. I might take some time to look into it this week, then.

> What is amis on the user level is I think the filter value selection which currently is a flat list.

This is happening on https://github.com/opf/openproject/pull/17281

> So with the exception of the WHERE statement on the project, the subselect for the custom field values is repeated. This looks like potential for optimizations.

Interesting. I had not noticed this behaviour when initially implementing the feature, but probably because we were still only trying to handle a single item filter. This feels like something that we could dive really deep into. I am not sure how to tackle yet, but I'll take upon myself to follow up after synchronising with the team

PR #17248, opened 2024-11-21
Rename LifeCycle to Project::LifeCycleStepDefinition and co

dombesz:
# Ticket
Part of https://community.openproject.org/wp/59287

Apply the class renamings discussed here: https://github.com/opf/openproject/pull/17119#issuecomment-2476923696 .

# Merge checklist

- [X] Added/updated tests


db/migrate/20241105175000_add_life_cycle_to_work_packages.rb
```diff
@@ -1,5 +1,8 @@
 class AddLifeCycleToWorkPackages < ActiveRecord::Migration[7.1]
   def change
-    add_reference :work_packages, :project_life_cycle, foreign_key: true, null: true
+    add_reference :work_packages,
+                  :life_cycle_step,
```
toy: I think `project_ life_cycle_step` should be more clear, though lengthier
If you change it `foreign_key` can be just `true`

db/migrate/20241030154245_create_project_life_cycle_step_definitions.rb
```diff
@@ -26,11 +26,12 @@
...
 #++
 
-class CreateLifeCycles < ActiveRecord::Migration[7.1]
+class CreateProjectLifeCycleStepDefinitions < ActiveRecord::Migration[7.1]
   def change
-    create_table :life_cycles do |t|
+    create_table :project_life_cycle_step_definitions do |t|
       t.string :type
       t.string :name
+      t.integer :position, default: 1, null: true
```
toy: For ordering wherever they are listed?

db/migrate/20241105151352_create_project_life_cycle_steps.rb
```diff
@@ -0,0 +1,15 @@
+class CreateProjectLifeCycleSteps < ActiveRecord::Migration[7.1]
+  def change
+    create_table :project_life_cycle_steps do |t|
+      t.string :type
+      t.date :start_date
+      t.date :end_date
+      t.boolean :active, default: false, null: false
+      t.integer :position, default: 1, null: true
```
toy: If this is for ordering when listing, then I think it should either follow the order of definitions or use `start_date`

toy (commented):
Nice, just few notes. Git rename detection for this PR makes files changed very confusing

db/migrate/20241105151352_create_project_life_cycle_steps.rb
dombesz: Yes, this is for the acts_as_list plugin. The default ordering follows the order of definition, however the list can be drag-and-drop reordered from the admin interface.

After running the seeders, the order is set correctly:

```ruby
[4] pry(main)> Project::LifeCycleStepDefinition.all.pluck(:id, :name, :position)
  Project::LifeCycleStepDefinition Pluck (0.8ms)  SELECT "project_life_cycle_step_definitions"."id", "project_life_cycle_step_definitions"."name", "project_life_cycle_step_definitions"."position" FROM "project_life_cycle_step_definitions"
=> [[1, "Initiating", 1], [2, "Ready for Planning", 2], [3, "Planning", 3], [4, "Ready for Executing", 4], [5, "Executing", 5], [6, "Ready for Closing", 6], [7, "Closing", 7]]
```

db/migrate/20241030154245_create_project_life_cycle_step_definitions.rb
dombesz: Yes, this is being used by the acts_as_list plugin.

dombesz:
I addressed your comments, the failing spec is unrelated to the changes. Please take a look.

db/migrate/20241105151352_create_project_life_cycle_steps.rb
toy: But what would be the point of reordering? Names are coming from definitions, and it make sense to be able to reorder them, but why changing order of date assignments for specific project?

dombesz: Sorry I misunderstood your initial question. Yes, it does not makes sense to have the list in the `Project::LifeCycleStep`, it is sufficient for the `Project::LifeCycleStepDefinition`.

dombesz:
Additionally I removed the `acts_as_list` from the `Project::LifeCycleStep` and merged the migrations on this RP to avoid conflicts. Please take a look.

PR #17258, opened 2024-11-22
[#59572] fixed rendering of hierarchy values in wp table views

Kharonus:
- 

# Ticket
[OP#59572](https://community.openproject.org/wp/59572)

- render hierarchy item values correctly in work package table

- use text display field (default) to also handle resources and represent them


brunopagno (approved):
Tested this locally, does the job briliantly :clap:

PR #17270, opened 2024-11-25
Hierarchy item filter "is any with descendants"

brunopagno:
# Ticket
https://community.openproject.org/wp/59071

Add a new filter

## Screenshots

https://github.com/user-attachments/assets/ae64e523-5433-4f43-bb22-86b878048560

The idea here is to leverage the default Equals filter, but include hierarchy items from the whole branch in the query.

# Merge checklist
- [x] Added/updated tests

### DIFF
--- a/app/models/queries/operators/custom_fields/hierarchies/equals_with_descendants.rb
+++ b/app/models/queries/operators/custom_fields/hierarchies/equals_with_descendants.rb
@@ -1,5 +1,3 @@
-# frozen_string_literal: true
-
 #-- copyright
 # OpenProject is an open source project management software.
 # Copyright (C) the OpenProject GmbH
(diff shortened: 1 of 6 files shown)


config/locales/en.yml
```diff
@@ -2336,6 +2336,7 @@ en:
   label_environment: "Environment"
   label_estimates_and_progress: "Estimates and progress"
   label_equals: "is"
+  label_equals_with_descendants: "is any with descendants"
```
brunopagno, line 2401: I am not sure what is the usual process for adding new translation keys

app/models/queries/operators/custom_fields/hierarchies/equals_with_descendants.rb
```diff
@@ -0,0 +1,50 @@
...
+
+          def self.sql_for_field(values, db_table, db_field)
+            items = CustomField::Hierarchy::Item.find(values)
+
+            actual_values = items.map do |item|
+              item.self_and_descendants.collect(&:id)
+            end.flatten.uniq
+
+            Equals.sql_for_field(actual_values, db_table, db_field)
+          end
```
brunopagno: We can refactor this to use the HierarchyService, but this would be the thought process behind the filter.

Collect the ids, and perform a Equals filter.

Happy to hear other ideas

brunopagno: Already refactored~

config/locales/en.yml
brunopagno: From the info gathering I did, nothing needs to be done

apfohl (approved):

PR #17272, opened 2024-11-25
Store timezones for TimeEntries and finalize how we determine end time

klaustopher:
# Ticket
https://community.openproject.org/projects/stream-time-and-costs/work_packages/59466

During the discussion today we realized that it is important to also persist the timezone of the tracked time entries so that we can properly determine what the time is. We used to anticipate that we could use the user's current timezone from the preferences, but as this might change in the future and would destroy old time entries, we also need to persist the timezone.

We also deducted that we will not need an `end_time` field but it will be implied by `start_time + hours`.

## Screenshots

# Merge checklist

- [x] Added/updated tests


as-op (approved):
👍

PR #17279, opened 2024-11-26
[#59343] Spacing between table columns is off if name is long

mrmir:
# Ticket

https://community.openproject.org/work_packages/59343

# Other points
- I made the width for the main column span 2 columns instead of 3 as that works better for meetings, but this can be reverted if necessary ofc
- I think it makes sense to also introduce "narrow" columns at some point, or at least have the actions column behave differently, because that seems to be taking up too much space imo. I tried to make this happen as part of this PR, but after spending a bit too much time was not able to, so here's a WP to [track it](https://community.openproject.org/wp/59741)

# Merge checklist

- [ ] Added/updated tests

### DIFF
--- a/app/components/op_primer/border_box_table_component.sass
+++ b/app/components/op_primer/border_box_table_component.sass
@@ -15,12 +15,19 @@
     justify-content: flex-start
     align-items: center
 
+    &--row-item
+      &:not(:first-child)
+        margin-left: 6px
+
+      &:not(:last-child)
+        margin-right: 6px
+
     &--mobile-heading,
     &--row-label
       display: none
 
-    &--wide-column
-      grid-column: span 3
+    &--main-column
+      grid-column: span 2
 
 @media screen and (max-width: $breakpoint-md)
   .op-border-box-grid
--- a/app/components/op_primer/border_box_table_component.sass
+++ b/app/components/op_primer/border_box_table_component.sass
@@ -15,12 +15,13 @@
     justify-content: flex-start
     align-items: center
 
+    &--heading,
     &--row-item
       &:not(:first-child)
-        margin-left: 6px
+        padding-left: 6px
 
       &:not(:last-child)
-        margin-right: 6px
+        padding-right: 6px
 
     &--mobile-heading,
     &--row-label
--- a/lookbook/docs/patterns/12-tables.md.erb
+++ b/lookbook/docs/patterns/12-tables.md.erb
@@ -35,7 +35,7 @@ end
 
 
 
-### Border Box Table specific options
+### Border Box Table specifics
 
 #### Mobile behavior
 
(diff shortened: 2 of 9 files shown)


app/components/op_primer/border_box_table_component.sass
```diff
@@ -15,12 +15,19 @@
     justify-content: flex-start
     align-items: center
 
+    &--row-item
+      &:not(:first-child)
+        margin-left: 6px
+
+      &:not(:last-child)
+        margin-right: 6px
+
```
HDinger: I assume that this was done to avoid that the wrapped text is glued directly to the next column, right? However, this is not ideal. First, it only intends the rows, not the header, thus creatig a misalignment. Second, it relys on the main column being the first one. If that is not the case, it will not work any more. And last but not least, if I do not have a main column at all, it intends unneccessarily. 

Bildschirmfoto 2024-11-27 um 08 15 09

I'd rather suggest to give the main column (including its header) some padding.

lookbook/docs/patterns/12-tables.md.erb
```diff
@@ -60,18 +60,22 @@ mobile_labels :users
...
 # On desktop, show these columns
 columns :title, :users, :created_at
 
-# Make title column wider than the others (factor 3 using grid span)
-wide_columns :title
+# Make title column wider than the others (factor 2 using grid span)
+main_column :title
 ```
-
+Note: Ideally, one one main column will be present for each table.
```
HDinger, line 78: Question for discussiob: If we even write, that ideally there should only be one main column, why don't we limit that in code? The identifier is even named in singular..

HDinger (changes requested):
Looks good overall 👍 I have only a minor remark. I would also like to clarify with the product team whether the component can have more than one main column.

lookbook/docs/patterns/12-tables.md.erb
mrmir: Decision from the front-end meeting: Mir update Lookbook docs to mention that it's good practice to only have one "Main":, but it's not a hard restriction.

So this should be fine as is imo, with the name being singular and the documentation pushing you towards just one, but if more are needed that can be done.

HDinger: Fine for me 👍 Thanks for double checking 🙇

app/components/op_primer/border_box_table_component.sass
```diff
@@ -17,10 +17,17 @@
...
       &:not(:last-child)
-        margin-right: 6px
+        padding-right: 6px
+
+    &--heading
+      &:not(:first-child)
+        padding-left: 6px
+
+      &:not(:last-child)
+        padding-right: 6px
```
HDinger: Please combine that, so that we do not have to repeat the code. If we decide the change spacing somewhen later on, the chances to forget about one or the other is lower.

```
&--heading,
&--row-item
  &:not(:first-child)
    .....
```

HDinger (approved):
I just have a minor remark. The rest is looking good 👍

mrmir:
Merging as the failing specs are unrelated

PR #17287, opened 2024-11-27
Fix for bulk editting work packages with hierarchy items

brunopagno:
# Ticket
[The link to the ticket number 59734](https://community.openproject.org/projects/document-workflows-stream/work_packages/59734/activity)

Fix bulk edit for work packages that have custom fields with format hierarchy

### DIFF
--- a/app/helpers/custom_fields_helper.rb
+++ b/app/helpers/custom_fields_helper.rb
@@ -172,6 +172,19 @@ module CustomFieldsHelper
                         options_for_select(base_options + custom_field.possible_values_options(project)),
                         id: field_id,
                         multiple: custom_field.multi_value?)
+    when "hierarchy"
+      base_options = [[I18n.t(:label_no_change_option), ""]]
+      result = CustomFields::Hierarchy::HierarchicalItemService.new
+        .get_descendants(item: custom_field.hierarchy_root, include_self: false)
+        .either(
+          ->(items) { items },
+          ->(_) { [] }
+        )
+      options = base_options + result.map do |item|
+        label = item.short.present? ? "#{item.label} (#{item.short})" : item.label
+        [label, item.id]
+      end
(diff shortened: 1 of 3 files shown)


app/helpers/custom_fields_helper.rb
```diff
@@ -172,6 +172,14 @@ def custom_field_tag_for_bulk_edit(name, custom_field, project = nil) # rubocop:
                         options_for_select(base_options + custom_field.possible_values_options(project)),
                         id: field_id,
                         multiple: custom_field.multi_value?)
+    when "hierarchy"
+      options = [[I18n.t(:label_no_change_option), ""]]
+      service = CustomFields::Hierarchy::HierarchicalItemService.new
+      options += service.get_descendants(item: custom_field.hierarchy_root, include_self: false).value!.map do |item|
```
Kharonus: 🟡 This goes into a direction, we had already on some other places:

is `.value!` enough for the failure handling? it means: "Throw on failure". So, while the method currently cannot return a failure, this might change in future, the interface is a `Result`.

Should we have small `.either` handling here? Maybe raising (as it is now) or returning empty options together with an error log entry?

Kharonus (approved):
Nice, now using the persistence layer makes it more decoupled. 👍🏽 

I have one issue regarding code style and excellence. Not a blocker for a merge, but as we still have some time, lets wrap our head shortly around this case.

app/helpers/custom_fields_helper.rb
brunopagno: I meant to do that when I refactored, but forgot. Adjusted it now.

Kharonus: yep, makes sense now. Also this code will eventually get replaced by the specific hierarchy selector (hopefully)

Kharonus (approved):

brunopagno:
image

I ran the failed test on CI locally and it worked. I'll assume this is fine

PR #17289, opened 2024-11-27
[#59765] disable hierarchy cfs from custom actions

Kharonus:
# Ticket
[OP#59354](https://community.openproject.org/work_packages/59354)

- custom actions should not break

- exclude custom fields of type hierarchy from being listed as available custom actions
- this disables setting custom fields of type hierarchy as a custom action, as all the options have to be preloaded. This can make the form very slow for big hierarchies.


brunopagno (approved):
code change make sense to me

Kharonus:
Changed the code a bit, @brunopagno

brunopagno (approved):
Nice! I like the use of the scope here. I wonder if the name could be better, but did not think of anything more meaningful, so I think we can go ahead.

PR #17290, opened 2024-11-27
Fix failing feature specs for wp filters

brunopagno:
# Ticket
Related to https://community.openproject.org/projects/openproject/work_packages/59556

Fix failing feature specs

Work around javascript tricks between `null` and `undefined` by transforming the value into a boolean.


Kharonus (approved):
If this fixes the specs, all green.

the change makes sense in the angular context.

PR #17296, opened 2024-11-28
Enable morphing for turbo drive

ulferts:
# Ticket

https://community.openproject.org/wp/59779

Enable morphing with TurboDrive without breaking functionality.

* Add the necessary headers that
  * Enable morphing
  * Keep the scroll position
* Exempt the top menu and various angular components from being morphed.

# Merge checklist

- [ ] Added/updated tests

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -433,7 +433,7 @@ module ApplicationHelper
   # To avoid the menu flickering, disable it
   # by default unless we're in test mode
   def initial_menu_styles(side_displayed)
-    Rails.env.test? || !side_displayed ? "" : "display:none"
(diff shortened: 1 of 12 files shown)


frontend/src/turbo/turbo-global-listeners.ts
```diff
@@ -64,4 +64,13 @@ export function addTurboGlobalListeners() {
     activateFlashNotice();
     activateFlashError();
   });
+
+  document.addEventListener('turbo:before-morph-element', (event) => {
+    const element = event.target as HTMLElement;
+
+    // In case the element is an OpenProject custom dom element, morphing is prevented.
+    if (element.tagName.startsWith('OPCE-')) {
```
dombesz: According to this https://github.com/opf/openproject/pull/17223#discussion_r1888464672 , it would be advisable to use `toUpperCase` for comparing the tagname.
```suggestion
    if (element.tagName.toUpperCase().startsWith('OPCE-')) {
```

PR #17298, opened 2024-11-28
[59163] Auto focus label field of new hierarchy item

apfohl:
# Ticket
[59163](https://community.openproject.org/projects/openproject/work_packages/59163)

Automatically set focus to the `label` input field when the form is opened for adding a new hierarchy item for a custom field of type hierarchy.

I added `autofocus: true` to the input.


brunopagno (approved):
:clap: :tada:

PR #17302, opened 2024-11-28
[chore] remove user no longer has errornous behaviour

Kharonus:
- removing edge cases from code base

- updating NC 28 to NC 29 allows removal of user from its last group
- removed special handling from tests

### DIFF
--- a/modules/storages/spec/common/storages/peripherals/storage_interaction/nextcloud/remove_user_from_group_command_spec.rb
+++ b/modules/storages/spec/common/storages/peripherals/storage_interaction/nextcloud/remove_user_from_group_command_spec.rb
@@ -72,16 +72,10 @@ RSpec.describe Storages::Peripherals::StorageInteraction::Nextcloud::RemoveUserF
     before do
       create_group(auth, storage, group)
       add_user_to_group(user, group)
-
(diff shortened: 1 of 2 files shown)


brunopagno (commented):
What happens if a customer wants to keep using nextcloud 28? Would they become blocked of upgrading openproject if that's the case?

mereghost:
> What happens if a customer wants to keep using nextcloud 28? Would they become blocked of upgrading openproject if that's the case?

IIRC our integrity checks would report it as "non-compatible".

brunopagno (approved):

Kharonus:
uhm, maybe we should bump the min NC version then? :D good idea, will ask @dominic-braeunlein 

so, dominic:
- ~shall we bump the supported min NC version to 29.0.9? In 15.1? This was the version I used for testing the group user removal.~
- we only have checks for versions of our integration app and group folder app
  - we could add a check for NC version, too
  - we could add the supported version of NC to our integration app (e.g. version 2.7.2 is supporting only 29 and up)

Kharonus:
All in all, I'd take the version discussion out of this PR, which basically just amends the tests. no prod code changes. WDYT?

PR #17304, opened 2024-11-28
enable ordering work packages by hierarchy items

brunopagno:
# Ticket
https://community.openproject.org/wp/59766

Enable ordering by custom field hierarchies on work packages

## Screenshots

https://github.com/user-attachments/assets/3fc7c336-3c6f-4e93-8473-e617a2176e39

# Merge checklist

- [x] Added/updated tests

### DIFF
--- a/app/models/custom_field/order_statements.rb
+++ b/app/models/custom_field/order_statements.rb
@@ -31,7 +31,7 @@ module CustomField::OrderStatements
   # value of the custom field.
   def order_statement
     case field_format
-    when "string", "date", "bool", "link", "int", "float", "list", "user", "version"
+    when "string", "date", "bool", "link", "int", "float", "list", "user", "version", "hierarchy"
       "cf_order_#{id}.value"
     end
   end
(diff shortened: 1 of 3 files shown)


app/models/custom_field/order_statements.rb
```diff
@@ -167,4 +169,9 @@ def join_for_order_by_version_sql
       multi_value:
     )
   end
+
+  def join_for_order_by_hierarchy_sql
+    table_name = CustomField::Hierarchy::Item.quoted_table_name
+    join_for_order_sql(value: "item.label", join: "INNER JOIN #{table_name} item ON item.id = cv.value::bigint")
```
Kharonus, line 175: ⚠️ Do I read this correctly, that this is now sorted by label (as it is joined into the values table). So we should keep that in mind, whenever we decide to change the display representation in the wp table view. (e.g. no longer starting with the label, but some shorts of ancestors).

Kharonus (approved):

app/models/custom_field/order_statements.rb
brunopagno: I just implemented the simples possible way forward.

To be fair this could also be extended to handle multi values... I think it's fair to bring to discussion.

brunopagno:
image

Flaky test looks green locally on the branch

PR #17310, opened 2024-11-29
[#59807] added form field instructions

Kharonus:
# Ticket
[OP#59807](https://community.openproject.org/work_packages/59807)

- new custom field form should have an instructions on multi-select form field

- added the instruction


bsatarnejad (approved):
LGTM 👍

PR #17314, opened 2024-11-29
OP#59817 Remove activity menu for anonymous users with login required

top4ek:
[OP#59817](https://community.openproject.org/projects/openproject/work_packages/details/59817/overview
)

Remove top menu Activity item for anonymous users when login is required as it's leads to nowhere(same login page).

Fresh installation shows only Activity item in top menu and leads to the same login page when login is required., so I removed that item when user is anonymous and login is required. 
As we don't have any items in top menu in this case and menu is not rendered, I've removed spec for that.
Returning empty array when there's no items thus we don't need to group them(otherwise we have 500 with heading? access to nil)

### DIFF
--- a/config/initializers/menus.rb
+++ b/config/initializers/menus.rb
@@ -43,6 +43,7 @@ Redmine::MenuManager.map :top_menu do |menu|
   menu.push :activity,
             { controller: "/activities", action: "index" },
             context: :modules,
+            if: Proc.new { User.current.logged? || !Setting.login_required? },
             icon: "history"
 
   menu.push :work_packages,


spec/features/menu_items/top_menu_item_spec.rb
```diff
@@ -138,12 +138,6 @@ def click_link_in_open_menu(title)
     context "as an anonymous user" do
       let(:user) { create(:anonymous) }
 
-      context "when login_required", with_settings: { login_required: true } do
-        it "redirects to login" do
-          expect(page).to have_current_path /login/
-        end
-      end
-
```
HDinger: Please do not simply delete this test because the redirect should still work and be tested. I think you just have to set `  let(:open_menu) { false }` for this context to work again.

HDinger (changes requested):
Hi @top4ek 

thanks a lot for your contribution! Functionality looks good, I just have a minor remark with regards to the test.

spec/features/menu_items/top_menu_item_spec.rb
top4ek: Not sure why :top_menu is forced before examples, but all other specs fail even if we don't open top menu there.

lib/redmine/menu_manager/top_menu_helper.rb
```diff
@@ -215,6 +215,8 @@ def more_top_menu_items
 
   def module_top_menu_item_groups
     items = more_top_menu_items
+    return items if items.empty?
```
apfohl: Even though an empty array is an empty array in ruby, I would suggest not returning the empty `items` array here, but the `item_groups` array. You can do this in a way,  that add `&` in line 223: `unless items.first&.heading?`. If you do it that way the `reduce` is operating again on an empty list or containing a single item.

Let me know if I should explain more in depth.

apfohl (changes requested):
I have a suggestion to the code in `module_top_menu_items`.

lib/redmine/menu_manager/top_menu_helper.rb
top4ek: Doing that gives us empty menu dropdown as `unless items.first&.heading?` adds to item_groups `[{:title=>nil, :items=>[]}]`

Screenshot_20241204_195215

top4ek: As a variant?
```ruby
def module_top_menu_item_groups
    item_groups = []
    items = more_top_menu_items
    return item_groups if items.blank?
```

apfohl (commented):
LGTM

Maybe @HDinger can have a look. I can’t approve it anymore. It makes sense to me, as far as I remember.

HDinger:
Hi @top4ek 

Sorry, for the long delay. This somehow slipped through 🤦‍♀️ 
We currently changed the implementation of the menu which is why there are some merge conflicts. Could you please resolve them?

top4ek:
Hi @HDinger 
look like new menu system doesn't require logic modification, just checking user and settings is enough. 
Menu itself still created with only one "home" item.

Also could you check my [other PRs](https://github.com/opf/openproject/issues?q=is%3Apr%20is%3Aopen%20author%3Atop4ek). They are quite old also.

HDinger (approved):
Thanks for the patience @top4ek 🙇

PR #17316, opened 2024-12-02
PDF export for time sheets

as-op:
# Ticket

https://community.openproject.org/projects/stream-time-and-costs/work_packages/59824

Export logged time as PDF time sheets

## Screenshots

<table><tr><td>
button
Button (in Figma this is in a drop down menu, should be done in a separate PR, here menu is out of scope)
</td></tr><tr><td>
worker
File is created in the background worker
</td></tr><tr><td>
pdf
The PDF 
</td></tr></table>

* Adapting from the XLS export for a cost query
* Reusing of existing PDF code (e.g. for the cover page)

# Merge checklist

- [x] Export skeleton PDF via the background worker and a button in the cost report view
- [x] Don't navigate to .../job_statuses/xyz, use the "inline" progress dialog
- [x] Apply [styling](https://www.figma.com/design/FdU4mgz2EkJQTQSvzojPRi/Time-and-costs-reports?node-id=1-2) 
- [x] Ellipse text in wp subject cell (missing prawn table support)
- [x] Table splitting on page breaks with merged rows (missing prawn-table support for splitting)
- [x] Use cost report query filter & data
- [x] Use new start_time and time_zone for time on "time" column
- [x] If an entry has its end time on a different day than the start date, it is indicated by a (+days), e.g. "23:00 - 04:00 (+1d)"
- [x] Verify used values for header/footer/cover page vs wireframes
- [x] Verify column widths with translations
- [x] Added/updated tests


frontend/src/stimulus/controllers/dynamic/costs/export.controller.ts
```diff
@@ -0,0 +1,60 @@
+import { Controller } from '@hotwired/stimulus';
```
klaustopher, line 31:
```suggestion
/*
 * -- copyright
 * OpenProject is an open source project management software.
 * Copyright (C) the OpenProject GmbH
 *
 * This program is free software; you can redistribute it and/or
 * modify it under the terms of the GNU General Public License version 3.
 *
 * OpenProject is a fork of ChiliProject, which is a fork of Redmine. The copyright follows:
 * Copyright (C) 2006-2013 Jean-Philippe Lang
 * Copyright (C) 2010-2013 the ChiliProject Team
 *
 * This program is free software; you can redistribute it and/or
 * modify it under the terms of the GNU General Public License
 * as published by the Free Software Foundation; either version 2
 * of the License, or (at your option) any later version.
 *
 * This program is distributed in the hope that it will be useful,
 * but WITHOUT ANY WARRANTY; without even the implied warranty of
 * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
 * GNU General Public License for more details.
 *
 * You should have received a copy of the GNU General Public License
 * along with this program; if not, write to the Free Software
 * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
 *
 * See COPYRIGHT and LICENSE files for more details.
 * ++
 */


import { Controller } from '@hotwired/stimulus';
```

modules/reporting/app/controllers/cost_reports_controller.rb
```diff
@@ -76,23 +76,38 @@ def check_cache
...
-    unless performed?
-      respond_to do |format|
-        format.html do
-          session[report_engine.name.underscore.to_sym].try(:delete, :name)
-          render locals: { menu_name: project_or_global_menu }
-        end
+    respond_to do |format|
+      format.html { render_html }
+      format.xls { export(:xls) }
+      format.pdf { export(:pdf) }
```
klaustopher, line 84: Question is here, we're making the timesheet "the" representation of the cost export as a PDF. WDYT about adding a second parameter like `variant` or something where we specify that we want the cost export or the timesheet. In a later iteration we could even make this a parameter

modules/reporting/app/controllers/cost_reports_controller.rb
```diff
@@ -524,16 +541,25 @@ def build_query(filters, groups = {})
...
       h
     end
-    cookie[:name] = @query.name if @query.name
-    session[report_engine.name.underscore.to_sym] = cookie
+  end
+
+  def cookie_groups
+    @query.group_bys.inject({}) do |h, group|
+      ((h[:"#{group.type}s"] ||= []) << group.underscore_name.to_sym) && h
+    end
```
klaustopher, line 562: 🌟 for the cleanup

modules/reporting/app/controllers/cost_reports_controller.rb
```diff
@@ -566,4 +592,8 @@ def find_optional_report(query = "1=0")
     end
   rescue ActiveRecord::RecordNotFound
   end
+
+  def session_name
+    report_engine.name.underscore.to_sym
```
klaustopher:
```suggestion
    @session_name ||= report_engine.name.underscore.to_sym
```

just a micro optimization as we seem to call this a lot

modules/reporting/app/models/cost_query.rb
```diff
@@ -85,6 +85,25 @@ def self.exists_in?(project, user)
     public(project).or(private(project, user)).exists?
   end
 
+  # rubocop:disable Metrics/AbcSize
+  def self.build_query(project, filters, groups = {})
```
klaustopher, line 89: Just a hint, you can just append the directive after the method definition. Then the cop is disabled in that method, and you don't need to manually enable it

```ruby
def self.build_query(project, filters, groups = {}) # rubocop:disable Metrics/AbcSize
```

modules/reporting/app/workers/cost_query/pdf/export_job.rb
```diff
@@ -0,0 +1,40 @@
+require "active_storage/filename"
+
+class CostQuery::PDF::ExportJob < Exports::ExportJob
```
klaustopher: suggestion, something like `ExportTimesheetJob`?

modules/reporting/app/workers/cost_query/pdf/timesheet_generator.rb
```diff
@@ -0,0 +1,385 @@
...
+  end
+
+  def cover_page_heading
+    heading
+  end
+
+  def cover_page_dates
+    dates_range = all_entries.group_by(&:spent_on).sort
+    start_date = dates_range.first&.first
+    end_date = dates_range.last&.first
```
klaustopher:
```suggestion
    start_date, end_date = all_entries.map(&:spent_on).minmax
```

Just to show off some nice ruby functions :)

modules/reporting/app/workers/cost_query/pdf/timesheet_generator.rb
```diff
@@ -0,0 +1,385 @@
...
+      content: entry.comments,
+      text_color: COMMENT_FONT_COLOR,
+      font_style: :italic,
+      colspan: table_columns_widths.size
+    }]
+  end
+
+  def table_header_columns
+    [
+      { content: I18n.t(:"activerecord.attributes.time_entry.spent_on"), rowspan: 1 },
```
klaustopher: A shortcut to those would be `TimeEntry.human_attribute_name(:spent_on)`. Not needed to change, might just read a bit nicer

klaustopher (approved):
Great job 👏 

I left some comments inline. We seem to work under the assumption that the Timesheet is the only PDF export we will ever generate for reports. I think it will be easy enough to add another parameter that tells us what type of PDF we want to generate, so that there is no need to do anything right now.

All in all, awesome work. Especially the edge case handling for weird PDF behavior.

modules/reporting/app/workers/cost_query/pdf/timesheet_generator.rb
as-op: That is nicer indeed. ^^

modules/reporting/app/controllers/cost_reports_controller.rb
as-op: Yes, true. This should and will be a parameter when a cost report PDF is implemented.

PR #17317, opened 2024-12-02
[59174] Added group by for hierarchy items

apfohl:
# Ticket
[59174](https://community.openproject.org/projects/document-workflows-stream/work_packages/59174)

Enable `GROUP BY` for hierarchy items.

## Screenshots
image

Fixed generating the API response that creates the row keys for dividing the workpackage table into groups.

### DIFF
--- a/app/models/custom_field/hierarchy/hierarchy_item_adapter.rb
+++ b/app/models/custom_field/hierarchy/hierarchy_item_adapter.rb
@@ -28,22 +28,17 @@
 # See COPYRIGHT and LICENSE files for more details.
 #++
 
-module Queries
-  module Filters
-    module Shared
-      module CustomFields
-        class Hierarchy < Base
-          def ar_object_filter?
-            true
-          end
+class CustomField::Hierarchy::HierarchyItemAdapter
+  attr_reader :name
 
-          def value_objects
-            CustomField::Hierarchy::Item
-              .where(id: @values)
-              .map { |item| CustomField::Hierarchy::HierarchyItemAdapter.new(item:) }
-          end
-        end
-      end
-    end
+  delegate :id, :label, :short, :to_s, to: :item
+
+  def initialize(item:)
+    @item = item
+    @name = item.label
   end
+
+  private
+
+  attr_accessor :item
 end
--- a/app/models/custom_field/hierarchy/hierarchy_item_adapter.rb
+++ b/app/models/custom_field/hierarchy/hierarchy_item_adapter.rb
@@ -29,14 +29,11 @@
 #++
 
 class CustomField::Hierarchy::HierarchyItemAdapter
-  attr_reader :name
-
   delegate :id, :label, :short, :to_s, to: :item
 
-  def initialize(item:)
-    @item = item
-    @name = item.label
-  end
+  def initialize(item:) = @item = item
+
+  def name = @item.label
 
   private
 
--- a/app/models/custom_field/hierarchy/item.rb
+++ b/app/models/custom_field/hierarchy/item.rb
@@ -36,5 +36,5 @@ class CustomField::Hierarchy::Item < ApplicationRecord
 
   scope :including_children, -> { includes(children: :children) }
 
-  def to_s = label
+  def to_s = short.nil? ? label : "#{label} (#{short})"
 end
--- a/app/models/custom_field/hierarchy/item.rb
+++ b/app/models/custom_field/hierarchy/item.rb
@@ -35,4 +35,6 @@ class CustomField::Hierarchy::Item < ApplicationRecord
   has_closure_tree order: "sort_order", numeric_order: true, dont_order_roots: true, dependent: :destroy
 
   scope :including_children, -> { includes(children: :children) }
+
+  def to_s = label
 end
--- a/app/models/query/results/group_by.rb
+++ b/app/models/query/results/group_by.rb
@@ -117,8 +117,11 @@ module ::Query::Results::GroupBy
     CustomFields::Hierarchy::HierarchicalItemService
       .new
       .get_descendants(item: custom_field.hierarchy_root, include_self: false)
-      .fmap { |list| list.filter { |item| keys.include?(item.label) } }
-      .fmap { |list| list.index_by(&:label) }
+      .fmap do |list|
+        list.map { |item| CustomField::Hierarchy::HierarchyItemAdapter.new(item:) }
+            .filter { |item| keys.include?(item.label) }
+            .index_by(&:label)
+      end
       .either(
         ->(list) { list },
         ->(error) do
@@ -127,6 +130,7 @@ module ::Query::Results::GroupBy
         end
       )
   end
+
   # rubocop:enable Metrics/AbcSize
 
   def transform_list_custom_field_keys(custom_field, groups)
--- a/app/models/query/results/group_by.rb
+++ b/app/models/query/results/group_by.rb
@@ -101,9 +101,7 @@ module ::Query::Results::GroupBy
 
     groups.transform_keys do |key|
       if custom_field.multi_value?
-        (key ? key.split(".") : []).map do |subkey|
-          items[subkey]
-        end
+        Array(key&.split(".")).map { |subkey| items[subkey] }
       else
         items[key] || nil
       end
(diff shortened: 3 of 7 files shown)


app/models/query/results/group_by.rb
```diff
@@ -89,11 +89,46 @@ def transform_custom_field_keys(groups)
...
+          items[subkey]
+        end
+      else
+        items[key] || nil
+      end
+    end
+  end
+
+  # rubocop:disable Metrics/AbcSize
+  def hierarchy_items_for_keys(custom_field, groups)
```
brunopagno, line 112: I spent some time trying to think of a way to do this so that we don't need to disable rubocop rules and I don't know of an obvious one. But this is a sign of something not ideal, possible the rubocop check itself.

I would not block things for correctness, though, but good to pay attention if we start seeing more of these.

brunopagno (commented):
The code looks okay, but when trying to group by on my local I get the following error

Is this an additional edge case, maybe or am I missing something on my setup?

```
NoMethodError (undefined method `name' for an instance of CustomField::Hierarchy::Item):
  
lib/api/decorators/aggregation_group.rb:84:in `map'
lib/api/decorators/aggregation_group.rb:84:in `set_links!'
lib/api/decorators/aggregation_group.rb:37:in `initialize'
lib/api/v3/work_packages/work_package_aggregation_group.rb:38:in `initialize'
app/services/api/v3/work_package_collection_from_query_service.rb:100:in `new'
app/services/api/v3/work_package_collection_from_query_service.rb:100:in `block in generate_groups'
app/services/api/v3/work_package_collection_from_query_service.rb:99:in `each'
app/services/api/v3/work_package_collection_from_query_service.rb:99:in `map'
app/services/api/v3/work_package_collection_from_query_service.rb:99:in `generate_groups'
app/services/api/v3/work_package_collection_from_query_service.rb:67:in `results_to_representer'
app/services/api/v3/work_package_collection_from_query_service.rb:47:in `call'
lib/api/v3/queries/helpers/query_representer_response.rb:38:in `block in query_representer_response'
lib/api/root_api.rb:263:in `raise_invalid_query_on_service_failure'
lib/api/v3/queries/helpers/query_representer_response.rb:35:in `query_representer_response'
lib/api/v3/queries/queries_api.rb:134:in `block (3 levels) in <class:QueriesAPI>'
```

app/models/query/results/group_by.rb
apfohl: I tried three different versions. This one looked best to me.

apfohl:
> The code looks okay, but when trying to group by on my local I get the following error
> 
> Is this an additional edge case, maybe or am I missing something on my setup?
> 
> ```
> NoMethodError (undefined method `name' for an instance of CustomField::Hierarchy::Item):
>   
> lib/api/decorators/aggregation_group.rb:84:in `map'
> lib/api/decorators/aggregation_group.rb:84:in `set_links!'
> lib/api/decorators/aggregation_group.rb:37:in `initialize'
> lib/api/v3/work_packages/work_package_aggregation_group.rb:38:in `initialize'
> app/services/api/v3/work_package_collection_from_query_service.rb:100:in `new'
> app/services/api/v3/work_package_collection_from_query_service.rb:100:in `block in generate_groups'
> app/services/api/v3/work_package_collection_from_query_service.rb:99:in `each'
> app/services/api/v3/work_package_collection_from_query_service.rb:99:in `map'
> app/services/api/v3/work_package_collection_from_query_service.rb:99:in `generate_groups'
> app/services/api/v3/work_package_collection_from_query_service.rb:67:in `results_to_representer'
> app/services/api/v3/work_package_collection_from_query_service.rb:47:in `call'
> lib/api/v3/queries/helpers/query_representer_response.rb:38:in `block in query_representer_response'
> lib/api/root_api.rb:263:in `raise_invalid_query_on_service_failure'
> lib/api/v3/queries/helpers/query_representer_response.rb:35:in `query_representer_response'
> lib/api/v3/queries/queries_api.rb:134:in `block (3 levels) in <class:QueriesAPI>'
> ```

@brunopagno Can you tell me when this happens?

app/models/custom_field/hierarchy/item.rb
```diff
@@ -35,4 +35,6 @@ class CustomField::Hierarchy::Item < ApplicationRecord
   has_closure_tree order: "sort_order", numeric_order: true, dont_order_roots: true, dependent: :destroy
 
   scope :including_children, -> { includes(children: :children) }
+
+  def to_s = label
```
Kharonus: 🟡 Not sure, where this `to_s` is actually called , but from my gut feeling the string representation should be `"#{label} (#{short})" if a short is available. this is a reoccurring pattern, and we could even simplify then locations like `custom_fields_helper:184` and `hierarchy_strategy:52`.

Kharonus (changes requested):
There is this thing with the string representation. Not major, but if we can do this maybe a bit better, I think it is worth thinking about.

app/models/custom_field/hierarchy/item.rb
brunopagno: +1. I like this idea. Small difference, but much less cognitive effort.

brunopagno (approved):
I think this looks good. :tada: 

I would still try to get Eric's comment as well cause it seems small enough.

app/models/custom_field/hierarchy/item.rb
apfohl: It is used indirectly the the grouping representers. It is used for the rows that a created for the heading groups. I don't wanted to introduce Item specific logic in the representers. If I change this to `label` and `short` it will be printed in the rows.

brunopagno: But wouldn't that be a good thing?

EDIT: I'm happy either way. I already approved the PR, so feel free to make a decision with @Kharonus

Kharonus (approved):
The to_s fix looks good, if we can find the source for the failing test, then this is approved.

app/models/query/results/group_by.rb
```diff
@@ -89,11 +89,50 @@ def transform_custom_field_keys(groups)
...
+    keys = groups.keys.map { |k| k ? k.split(".") : [] }.flatten.uniq
+
+    CustomFields::Hierarchy::HierarchicalItemService
+      .new
+      .get_descendants(item: custom_field.hierarchy_root, include_self: false)
+      .fmap do |list|
+        list.map { |item| CustomField::Hierarchy::HierarchyItemAdapter.new(item:) }
+            .filter { |item| keys.include?(item.label) }
```
mereghost: Those 2 operations can be joined using `filter_map`.

```ruby
list.filter_map { |item| CustomField::Hierarchy::HierarchyItemAdapter.new(item:) if keys.include?(item.label) }
```

app/models/custom_field/hierarchy/hierarchy_item_adapter.rb
```diff
@@ -0,0 +1,44 @@
...
+#++
+
+class CustomField::Hierarchy::HierarchyItemAdapter
+  attr_reader :name
+
+  delegate :id, :label, :short, :to_s, to: :item
+
+  def initialize(item:)
+    @item = item
+    @name = item.label
```
mereghost: What do you think of, instead of using an instance variable, just using a "getter method" on the lines of 

```
def name = @item.label
```

app/models/query/results/group_by.rb
```diff
@@ -89,11 +89,50 @@ def transform_custom_field_keys(groups)
...
       transform_single_custom_field_keys(custom_field, groups)
     end
   end
 
+  def transform_hierarchy_custom_field_keys(custom_field, groups)
+    items = hierarchy_items_for_keys(custom_field, groups)
+
+    groups.transform_keys do |key|
+      if custom_field.multi_value?
+        (key ? key.split(".") : []).map do |subkey|
```
mereghost: `String#split` always returns an array. We are doing a `truthy` check if `falsey` we are going for an empty array.

We could probably do something on the lines of

```ruby
Array(key&.split(".")).map
```

mereghost (approved):
LGTM.

I just left some code suggestions that are by no means blocking (or maybe even good 🤣 )

Kharonus (approved):

PR #17322, opened 2024-12-02
[#59354] add missing dot to fix scope

Kharonus:
# Ticket
[OP#59354](https://community.openproject.org/work_packages/59354)

- fix the scope

- fix the damn scope

### DIFF
--- a/app/models/work_package_custom_field.rb
+++ b/app/models/work_package_custom_field.rb
@@ -51,7 +51,7 @@ class WorkPackageCustomField < CustomField
 
   scope :usable_as_custom_action, -> {
     where.not(field_format: %w[hierarchy])
-    order(:name)
+         .order(:name)
   }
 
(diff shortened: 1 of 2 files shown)


NobodysNightmare:
Do we have any unit tests in this area of the code that could have spotted this mistake? Should we add some?

akabiru (approved):
🔨

Kharonus:
That is a valid point. I admit, I struggled, because I do no agree with the design of the custom fields. Their models have way too much logic and responsibilities.

~But I found a unit test for this controller and will add test cases for the scopes.~

Wait, no, the test I found is for the work package model. I will create a new model test, but will only add scope tests for now.

NobodysNightmare (approved):
Great, thanks for adding the tests. As far as I can tell they look good.

PR #17327, opened 2024-12-02
Updates GoodJob to version 3.99

mereghost:
Preparation steps for the eventual 4.X series upgrade.

Detailed instructions are on the [gem's repo](https://github.com/bensheldon/good_job?tab=readme-ov-file#upgrading-v3-to-v4).

@machisuji Feel free to take over and run as many tests as necessary.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -126,7 +126,7 @@ gem "multi_json", "~> 1.15.0"
 gem "oj", "~> 3.16.0"
 
 gem "daemons"
-gem "good_job", "= 3.26.2" # update should be done manually in sync with saas-openproject version.
+gem "good_job", "= 3.99.1" # update should be done manually in sync with saas-openproject version.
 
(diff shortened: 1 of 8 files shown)


mereghost:
I noticed some DelayedJob leftovers (like tables etc). Can I remove all those references too?

mereghost:
@machisuji Is this enough for you to give it a try?

brunopagno (commented):
From code standpoint it looks correct. All migrations seem to be validating and creating/removing things appropriately.

Is there something specific that's worth testing locally?

brunopagno (approved):
Okay I'm marking this as approved as I tested a few things with success:

1. switched to this branch
2. ran the migrations
3. created a "reminder" to validate that the worker was doing it's thing
4. validated that the reminder was working as expected
5. switched back from this branch to my working one
6. validated that the application was still working

I think for local env this should be good enough

PR #17334, opened 2024-12-03
Beautify development highlight

NobodysNightmare:
This might be an issue dependent on the OS or
available fonts, but to me the header in development mode read like "developmerdevelopmer", because the text was wider than the image width we allocated.

By increasing the width to 120px, there is now some whitespace left between the images (hopefully for most OSes), which looks nicer IMO and makes the dev-hint more readable.

### Before

Desktop:
image

Mobile:
image

### After

Desktop:
image

Mobile:
image

I was briefly thinking about using 100px width, which would have looked like `developmentdevelopment` on my machine. But assuming that the text width is different depending on the exact device, I figured it would be safer to always count on _some_ amount of whitespace.

# Merge checklist

- [x] ~~Added/updated tests~~
- Tested major browsers
    - [x] Chrome (actually looked acceptable before already, with a small amount of whitespace. Now has more whitespace, but still looks nice)
    - [x] Firefox

### DIFF
--- a/frontend/src/global_styles/common/header/app-header.sass
+++ b/frontend/src/global_styles/common/header/app-header.sass
@@ -83,4 +83,4 @@
         padding-right: 1rem
 
   &.op-app-header_development
-    background-image: url("data:image/svg+xml,<svg xmlns='http://www.w3.org/2000/svg' width='90' height='20'><text x='0' y='1em' fill='rgba(0, 0, 0, 0.5)'>development</text></svg>")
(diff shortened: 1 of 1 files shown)


klaustopher (approved):

NobodysNightmare:
CI tests have been failing for 5 attempts now, though the affected tests are changing, so apparently these are flickering tests.

I can't for the life of me imagine a way that changing the width of an image would make CSS selectors flickeringly unavailable, so I am rather confident that my changes did not introduce additional test flicker.

That said, it still doesn't feel great to merge on a failing pipeline.

PR #17339, opened 2024-12-03
Remove inappropriate status tripping up turbo

ulferts:
# Ticket

https://community.openproject.org/wp/59369

Fix the issue described in the linked bug.

Returning `unprocessable_entity` on a GET request, with no parameters provided is not appropriate and trips up turbo leading to a reinitialization of the angular app and ckeditor.
It was added when enabling turbo drive (#16486) where when rendering a response to a POST form request a non ok status needs to be returned. I believe here, it was added erroneously.

# Merge checklist


HDinger (approved):
Works as expected. Nice catch 👍

PR #17341, opened 2024-12-03
Implementation/59721 add setedit reminder buttondialog to workpackage page

jjabari-op:
[Ticket](https://community.openproject.org/projects/communicator-stream/work_packages/59721/activity)


app/contracts/reminders/base_contract.rb
```diff
@@ -59,6 +60,10 @@ def validate_remindable_exists
       errors.add :remindable, :not_found if model.remindable.blank?
     end
 
+    def validate_remind_at_present
```
akabiru, line 63: 🟢  worth dropping in a unit test https://github.com/opf/openproject/blob/dev/spec/contracts/reminders/base_contract_spec.rb

app/controllers/work_packages/reminders_controller.rb
```diff
@@ -0,0 +1,206 @@
...
+
+    case @errors.find { |error| error.attribute == :remind_at }&.type
+    when :blank
+      handle_blank_error
+    when :datetime_must_be_in_future
+      handle_future_error
+    end
+
+    @errors.delete(:remind_at)
+  end
```
akabiru, line 146: 🟠 I think this method and related calls would do well as an extracted simple PORO that can be unit tested- also reduce the perceived complexity on the controller

app/models/reminder.rb
```diff
@@ -33,6 +33,21 @@ class Reminder < ApplicationRecord
...
+
+  def self.upcoming_and_visible_to(user)
+    visible_reminders = visible(user)
+    reminder_notifications_for_reminders = ReminderNotification.where(reminder: visible_reminders)
+
+    visible_reminders
+      .where(completed_at: nil)
+      .where.not(id: reminder_notifications_for_reminders.select(:reminder_id))
+  end
+
```
akabiru: Nice one covering these- 🟠  it would be worth add the corresponding model unit specs in https://github.com/opf/openproject/blob/dev/spec/models/reminder_spec.rb

frontend/src/app/features/work-packages/components/wp-reminder-modal/wp-reminder.modal.html
```diff
@@ -0,0 +1,46 @@
...
+      #frameElement
+      id="reminder_modal_body"
+      src="{{this.frameSrc}}">
+      <op-content-loader
+        viewBox="0 0 180 80"
+      >
+        <svg:rect height="10" rx="2" ry="2" width="110" x="0" y="0"/>
+        <svg:rect height="10" rx="2" ry="2" width="25" x="120" y="0"/>
+        <svg:rect height="10" rx="2" ry="2" width="25" x="155" y="0"/>
+        <svg:rect height="25%" rx="2" ry="2" width="180" x="0" y="16"/>
```
akabiru: 🟠 I noticed this content loader skeleton comes from the shares modal- could we ask design for one that matches the reminders modal?

frontend/src/app/features/work-packages/components/wp-buttons/wp-reminder-button/wp-reminder-button.component.ts
```diff
@@ -0,0 +1,103 @@
...
+      .apiV3Service
+      .work_packages
+      .id(this.workPackage.id as string)
+      .reminders
+      .get()
+      .pipe(
+        map((collection:CollectionResource) => { return collection.total; }),
+      );
+  }
+}
```
akabiru, line 103: 🧹 Several eslint issues that need fixing

frontend/src/app/features/work-packages/components/wp-reminder-modal/wp-reminder.modal.ts
```diff
@@ -0,0 +1,113 @@
...
+      .apiV3Service
+      .work_packages
+      .id(this.workPackage.id as string)
+      .reminders
+      .get()
+      .pipe(
+        map((collection:CollectionResource) => { return collection.total > 0; }),
+      );
+  }
+}
```
akabiru, line 113: 🧹 Several eslint issues that need fixing

lib/api/v3/reminders/reminder_representer.rb
```diff
@@ -0,0 +1,41 @@
...
+
+module API
+  module V3
+    module Reminders
+      class ReminderRepresenter < ::API::Decorators::Single
+        include API::Decorators::DateProperty
+
+        property :id
+
+        date_time_property :remind_at
```
akabiru, line 38: 🔴 Since we're adding the reminder API either way, shall we take the chance to represent it fully? by adding the note and creator association

akabiru: 🔴 It would also be good to see either a representer unit spec or an API spec for this

app/models/reminder.rb
akabiru: 🟠  Once the tests are added. I think the method to exclude reminders that have associated reminder notifications can be simplified via [where.missing](https://github.com/rails/rails/pull/34727)

Performance seems ~about the same~ faster - as the later results in no sub-queries and reads easier IMO

```suggestion
  # Currently, reminders are personal, meaning
  # they are only visible to the user who created them.
  def self.visible(user)
    where(creator: user)
  end

  def self.upcoming_and_visible_to(user)
    visible(user)
      .where(completed_at: nil)
      .where.missing(:reminder_notifications)
  end
```

frontend/src/app/features/work-packages/components/wp-buttons/wp-reminder-button/wp-reminder-button.component.ts
```diff
@@ -0,0 +1,103 @@
...
+          filter((id) => id === this.workPackage.id?.toString()),
+          startWith(null),
+          switchMap(() => this.countReminders()),
+        ),
+      this.storeService.unread$
+        .pipe(
+          startWith(0),
+          switchMap(() => this.countReminders()),
+        ),
+    );
```
akabiru, line 85: Q: Why do we need to reference unread notifications? 🤔

frontend/src/app/features/work-packages/components/wp-reminder-modal/wp-reminder.modal.html
```diff
@@ -0,0 +1,46 @@
...
+    </button>
+  </div>
+  <p
+    class="spot-modal--subheader fgColor-muted"
+    [textContent]="this.text.subtitle">
+  </p>
+
+  <div class="spot-modal--body spot-container">
+    <turbo-frame
+      #frameElement
```
akabiru, line 33: Nice one 👍🏾

lib/api/v3/reminders/reminders_api.rb
```diff
@@ -0,0 +1,49 @@
...
+                                     .upcoming_and_visible_to(User.current)
+            ReminderCollectionRepresenter.new(reminders,
+                                              self_link: api_v3_paths.work_package_reminders(@work_package.id),
+                                              current_user: User.current)
+          end
+        end
+      end
+    end
+  end
+end
```
akabiru, line 49: 🔴 ditto: would be good to see a basic API spec. I think we also need to to update the swagger open-api yml as well. cc tagging @Kharonus - who oversees the API at OP

spec/features/work_packages/reminders_spec.rb
```diff
@@ -0,0 +1,322 @@
...
+
+      work_package_page.expect_and_dismiss_flash(type: :success,
+                                                 message: I18n.t("work_package.reminders.success_deletion_message"))
+      work_package_page.expect_reminder_button_without_count
+      expect(Reminder.upcoming_and_visible_to(user_with_permissions).count).to eq(0)
+    end
+
+    describe "validations" do
+      it "renders errors on the date field or time field when the reminder is in the past" do
+        two_am = "02:00".to_time
```
akabiru: Screenshot 2024-12-06 at 2 09 25 PM

spec/features/work_packages/reminders_spec.rb
```diff
@@ -0,0 +1,322 @@
...
+
+  context "without permissions to manage own reminders" do
+    current_user { user_without_permissions }
+
+    it "does not render the reminder button when visiting the work package page" do
+      work_package_page.visit!
+      work_package_page.expect_no_reminder_button
+    end
+  end
+end
```
akabiru, line 385: Nice work on the feature specs! 👍🏾

akabiru (changes requested):
Quite the feat, nicely done! 👏🏾 I've noted a few general issues

- [x] Missing unit test coverage for model, contract and API changes - annotated in the code
- [x] Missing API specification & tests
- [x] Several eslint and rubocop issues

UI Wise
- [x] Missing modal cancel button as per [the design](https://www.figma.com/design/hC75Mf34GcQ2LFdm2pQu0l/Reminders?node-id=2-15928&node-type=frame&t=OWW2AAJ77CgnLgo4-0)
   - [x] Separately, it would be good to check with UX/Product if we can adopt "Remove reminder" instead of "Delete reminder" and "Schedule" instead of "Save" - from previous mention [people seemed in favour of the idea](https://matrix.to/#/!lesfhfcbTsKJRLPltu:openproject.org/$GL5v5x_-FTH12i5iqxho5myfRFKnu9TOZ0-jFRYQcbU?via=openproject.org)
 - [x] On hover of the "Set reminder icon" no label is displayed
 - [x] The reminders clock icon is off-centered to the left
 
Kapture 2024-12-06 at 14 24 41

 - [x] Content loader skeleton does not match the reminders UI
 - [x] Duplicate form error messages
Screenshot 2024-12-06 at 11 03 09 AM

- [x] Unhandled 404 when reminder edit modal is open, and the reminder notification is created whilst it's still open (you can test this by creating a reminder in the next minute, open the edit modal and wait for the reminder notification to be created then attempt to edit it)

Screenshot 2024-12-06 at 11 00 29 AM

- [x] icon is smaller than counter

Screenshot 2024-12-09 at 2 02 10 PM

lib/api/v3/reminders/reminders_api.rb
Kharonus: Yes, please, an inserting of this new endpoint into the OpenAPI specification would be great.

Also, looking at the representer, this is also not a big step. Seems to be like a very small endpoint.

Kharonus: You can though extract that into a separate PR, or even include me in the work (as in pairing on it), to avoid too many loops.

jjabari-op:
Thank you @akabiru for the comprehensive and detailed review! I'm on it! :)

akabiru:
@jjabari-op updating against dev as there were several recent CI stabalisation fixes that would be nice to have in

app/controllers/work_packages/reminders_controller.rb
akabiru: https://community.openproject.org/wp/60127

lib/api/v3/reminders/reminders_api.rb
akabiru: 👍🏾 
Screenshot 2024-12-12 at 12 53 17 PM

akabiru (approved):
🥇

PR #17346, opened 2024-12-04
fix broken link to storages documentation

brunopagno:
# Ticket
https://community.openproject.org/projects/openproject/work_packages/57896/activity

Fix a broken link

### DIFF
--- a/config/locales/js-en.yml
+++ b/config/locales/js-en.yml
@@ -1455,4 +1455,4 @@ en:
         network_on: "Network is back. We are trying."
         timeout: >
           OpenProject could not provide you access to the project folder within the expected period of time. Please, try once again.<br><br>
-          If that problem persists please contact your OpenProject administrator to check the <a href="https://www.openproject.org/docs/system-admin-guide/files/external-file-storages/file-storage-troubleshooting/">health status of the file storage setup</a>.
+          If that problem persists please contact your OpenProject administrator to check the <a href="https://www.openproject.org/docs/system-admin-guide/files/external-file-storages/health-status/">health status of the file storage setup</a>.
--- a/config/locales/js-en.yml
+++ b/config/locales/js-en.yml
@@ -1453,6 +1453,3 @@ en:
       waiting_subtitle:
         network_off: "There is a network problem."
         network_on: "Network is back. We are trying."
-        timeout: >
-          OpenProject could not provide you access to the project folder within the expected period of time. Please, try once again.<br><br>
-          If that problem persists please contact your OpenProject administrator to check the <a href="https://www.openproject.org/docs/system-admin-guide/files/external-file-storages/health-status/">health status of the file storage setup</a>.
--- a/modules/storages/app/components/storages/open_project_storage_modal_component.html.erb
+++ b/modules/storages/app/components/storages/open_project_storage_modal_component.html.erb
@@ -1,7 +1,7 @@
 <%= render(
     Primer::Alpha::Dialog.new(id: self.class.dialog_id,
                               title: 'Open project folder modal',
-                              data: @data)
+                              data:)
 ) do |dialog| %>
     <% dialog.with_header(display: :none) %>
     <% dialog.with_body(id: self.class.dialog_body_id, p: 2) do %>
--- a/modules/storages/app/components/storages/open_project_storage_modal_component.rb
+++ b/modules/storages/app/components/storages/open_project_storage_modal_component.rb
@@ -29,15 +29,13 @@
 #++
 
 class Storages::OpenProjectStorageModalComponent < ViewComponent::Base
+  attr_reader :controller, :project_storage_open_url, :redirect_url, :state
+
   def initialize(project_storage_open_url:, redirect_url:, state:, **options)
     super
-    controller = "storages--open-project-storage-modal"
-    @data = {
-      controller:,
-      "application-target": "dynamic",
-      "#{controller}-project-storage-open-url-value": project_storage_open_url,
-      "#{controller}-redirect-url-value": redirect_url
-    }
+    @controller = "storages--open-project-storage-modal"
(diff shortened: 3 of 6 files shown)


config/locales/js-en.yml
```diff
@@ -1455,4 +1455,4 @@ en:
         network_on: "Network is back. We are trying."
         timeout: >
           OpenProject could not provide you access to the project folder within the expected period of time. Please, try once again.<br><br>
-          If that problem persists please contact your OpenProject administrator to check the <a href="https://www.openproject.org/docs/system-admin-guide/files/external-file-storages/file-storage-troubleshooting/">health status of the file storage setup</a>.
+          If that problem persists please contact your OpenProject administrator to check the <a href="https://www.openproject.org/docs/system-admin-guide/files/external-file-storages/health-status/">health status of the file storage setup</a>.
```
brunopagno: I can fix this in other languages as well, but I am not sure if this affects crowdin process. What would be preferred to do?

Kharonus: 🟡 Can we do it as we do it with the static links we are using for example in the storage form? 

```ruby
    def caption
      href = ::OpenProject::Static::Links[:storage_docs][:one_drive_drive_id_guide][:href]
      I18n.t("storages.instructions.one_drive.drive_id",
             drive_id_link_text: render(Primer::Beta::Link.new(href:, target: "_blank")) do
               I18n.t("storages.instructions.one_drive.documentation_link_text")
             end)
    end
```

Kharonus: This way, the localisation does not have to be updated, if the link is changing

brunopagno: The thing is this translation is happening on a [javascript controller](https://github.com/opf/openproject/blob/dev/frontend/src/stimulus/controllers/dynamic/storages/open-project-storage-modal.controller.ts#L136). So probably more work needs to be done to inject the URL the way you propose.

I think it's still worth updating the link since it's broken, but a refactor for this can come as a future step.

Kharonus: Nah, I think this is not as difficult. This is a stimulus controller, hence you can pass values in. Like it is done with `projectStorageOpenUrlValue` and `redirectUrlValue`.

And those values are built in rails context.

Kharonus: This means, of course, we will have to move the locale key away from the `js` yml. But, as you nevertheless are invalidating all other languages, I see no big difference. :D

brunopagno: Fair enough. Lemme try, then :sweat_smile:

modules/storages/app/components/storages/open_project_storage_modal_component.rb
```diff
@@ -29,15 +29,13 @@
...
+
   def initialize(project_storage_open_url:, redirect_url:, state:, **options)
     super
-    controller = "storages--open-project-storage-modal"
-    @data = {
-      controller:,
-      "application-target": "dynamic",
-      "#{controller}-project-storage-open-url-value": project_storage_open_url,
-      "#{controller}-redirect-url-value": redirect_url
-    }
```
brunopagno, line 40: I found out that trying to call a `render` during initialization explodes everything. So I needed to refactor this guy a bit.

Functionality *should* be the same as before.

modules/storages/app/components/storages/open_project_storage_modal_component.html.erb
```diff
@@ -1,7 +1,7 @@
 <%= render(
     Primer::Alpha::Dialog.new(id: self.class.dialog_id,
                               title: 'Open project folder modal',
-                              data: @data)
+                              data: data)
```
Kharonus: 🟡 can be shortened to `data:)`, no?

modules/storages/app/components/storages/open_project_storage_modal_component.rb
```diff
@@ -48,4 +46,24 @@ def self.dialog_id
   def self.dialog_body_id
     "open-project-storage-modal-body-component"
   end
+
+  def data
+    @data ||= {
```
Kharonus, line 51: 🟢 yeah, nice, lazy building of the data. Did you test it, does the controller comes up correctly? But I think this is quite the improvement to how it was before :D

modules/storages/config/locales/en.yml
```diff
@@ -344,6 +344,11 @@ en:
       success:
         subtitle: You are being redirected
         title: Integration setup completed
+      timeout:
+        link_text: health status of the file storage setup
+        subtitle: 'OpenProject could not provide you access to the project folder within the expected period of time. Please, try once again.<br><br> If that problem persists please contact your OpenProject administrator to check the %{storages_health_link}.
+
```
Kharonus: 🟡 did an empty line sneak in? and a final line break? And probably you should use " instead of ' for the string, by the convention of our ruby side of code.

Kharonus (changes requested):
just a minor thing left.

modules/storages/app/components/storages/open_project_storage_modal_component.rb
Kharonus: nice find

modules/storages/config/locales/en.yml
brunopagno: It was a rake task magic thing :eyes: 
Nice catch, though. :bow:

modules/storages/app/components/storages/open_project_storage_modal_component.rb
brunopagno: image

It does the thing it should do, so, yep.

modules/storages/app/components/storages/open_project_storage_modal_component.html.erb
brunopagno: it seems to be the case

Kharonus (approved):
LGTM

PR #17354, opened 2024-12-04
[#59866] fixed update contract

Kharonus:
# Ticket
[OP#59866](https://community.openproject.org/work_packages/59866)

- removing short of item must work

- add early check to short rule for update contract
- similar to create contract

### DIFF
--- a/app/contracts/custom_fields/hierarchy/update_item_contract.rb
+++ b/app/contracts/custom_fields/hierarchy/update_item_contract.rb
@@ -52,6 +52,7 @@ module CustomFields
 
       rule(:short) do
         next if schema_error?(:item)
+        next unless key?
 
         key.failure(:not_unique) if values[:item].siblings.exists?(short: value)
       end
--- a/spec/contracts/custom_fields/hierarchy/update_item_contract_spec.rb
+++ b/spec/contracts/custom_fields/hierarchy/update_item_contract_spec.rb
@@ -38,6 +38,7 @@ RSpec.describe CustomFields::Hierarchy::UpdateItemContract do
     let!(:vader) { create(:hierarchy_item) }
     let!(:luke) { create(:hierarchy_item, label: "luke", short: "ls", parent: vader) }
     let!(:leia) { create(:hierarchy_item, label: "leia", short: "lo", parent: vader) }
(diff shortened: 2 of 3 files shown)


app/contracts/custom_fields/hierarchy/update_item_contract.rb
```diff
@@ -52,6 +52,7 @@ class UpdateItemContract < Dry::Validation::Contract
 
       rule(:short) do
         next if schema_error?(:item)
+        next unless key?
```
NobodysNightmare, line 55: Semantical question (note that I am new to `Dry::Validation::Contract`): I understood that our problem was that the value `""` (empty string) could collide in the uniqueness check, though those should be ignored. I.e. two items with an empty short should not trigger a validation error.

However, the rule as specified now, doesn't care about `value` (e.g. `next if value.blank?`) but instead it checks the presence of `key`.

I assume this means something like "if an argument named `:short` was given, run this rule, otherwise skip it".

Is this what we want?

In other words: Would this implementation still work properly, if we defined `starkiller` in our tests as

```
# short is mentioned explicitly as an empty string
let!(:starkiller) { create(:hierarchy_item, label: "starkiller", short: "", parent: vader) }
```

(I might be missing a lot of relevant context here, but I'll post the question at least for the sake of my own education)

NobodysNightmare: Meh, I think I found [the answer](https://dry-rb.org/gems/dry-validation/1.8/rules/#checking-if-the-value-is-present) to my own question:

> When you're not sure if the value is actually available, you can use key? method. It returns true when a value under rule's key is present, false otherwise.

NobodysNightmare (approved):

app/contracts/custom_fields/hierarchy/update_item_contract.rb
Kharonus: Yep, in the persistence layer we actually filter out empty strings and such, so that on the contract side, the hash won't contain the key `short` anymore.

But, of course you can evaluate it on the hash with `item[:short]` and would receive `nil`, which is why `value` is `nil`, and this is equal to siblings without a short. ;)

PR #17356, opened 2024-12-04
update release image for 15.1

ulferts:
# Ticket

https://community.openproject.org/wp/58211

Replace the teaser image for 15.1

## Screenshots

image


corinnaguenther:
@ulferts Not sure if a dev has to review this, but it looks good to me!

ulferts:
@corinnaguenther I think your review is just as good in this case. There isn't the risk of breaking anything apart from what the screenshot shows.

PR #17357, opened 2024-12-04
[OP#59913] Add couple missing locale keys

top4ek:
[OP#59913](https://community.openproject.org/projects/openproject/work_packages/59913/activity)

Add missing attribute in en locale so it could be translated via crowdin.

### DIFF
--- a/app/views/admin/settings/api_settings/show.html.erb
+++ b/app/views/admin/settings/api_settings/show.html.erb
@@ -70,7 +70,7 @@ See COPYRIGHT and LICENSE files for more details.
   </section>
   <section class="form--section">
     <fieldset class="form--fieldset">
-      <legend class="form--fieldset-legend"><%= I18n.t("setting_apiv3_cors_title")</legend>
+      <legend class="form--fieldset-legend"><%= I18n.t("setting_apiv3_cors_title") %></legend>
     </fieldset>
     <div class="form--field">
       <%= setting_check_box :apiv3_cors_enabled %>
--- a/app/views/admin/settings/api_settings/show.html.erb
+++ b/app/views/admin/settings/api_settings/show.html.erb
@@ -70,7 +70,7 @@ See COPYRIGHT and LICENSE files for more details.
   </section>
   <section class="form--section">
     <fieldset class="form--fieldset">
-      <legend class="form--fieldset-legend">Cross-Origin Resource Sharing (CORS)</legend>
+      <legend class="form--fieldset-legend"><%= I18n.t("setting_apiv3_cors_title")</legend>
     </fieldset>
     <div class="form--field">
       <%= setting_check_box :apiv3_cors_enabled %>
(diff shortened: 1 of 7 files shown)


config/locales/en.yml
```diff
@@ -847,6 +847,7 @@ en:
         is_required: "Required"
         max_length: "Maximum length"
         min_length: "Minimum length"
+        content_right_to_left: "Left-to-Right content"
```
NobodysNightmare: Can you double check that the translation is correct? The key is "right-to-left", but the translation is "left-to-right".

config/locales/en.yml
```diff
@@ -4240,6 +4244,8 @@ en:
       builtin: Built-in instance application
       confidential: Confidential
       singular: "OAuth application"
+      scope: "Scopes:"
```
NobodysNightmare: Not an error introduced in this PR, but I noticed that it seemed inconsistent to see only scopes with a trailing `:` here. And indeed it's also inconsistent in the UI:

image

I'd suggest to drop the colon here, so that it doesn't slip into all translations before being fixed.

NobodysNightmare (changes requested):
Thank you for providing missing translation keys. I left two comments that would be nice to be addressed before we merge this change.

Also note that the tests are failing right now:

```
  1) I18n files are normalized
     Failure/Error: expect(non_normalized).to be_empty, error_message

       The following files need to be normalized:
         modules/storages/config/locales/en.yml
       Please run `i18n-tasks normalize' to fix
```

I hope that running this rake task will fix the failures.

Note: I will probably not be around to merge these changes until beginning of next year. Rest assured that I will keep this tab open in my browser and make sure to look at this PR again once I get back to work.

top4ek:
Thanks for findings — fixed.

> I hope that running this rake task will fix the failures.

Interesting feature — it just moved one key little bit higher inside yaml. But still has inconsistency with quote style.

> until beginning of next year

NP. We have saying used on whole December: "Let's do it after holidays" :) Some time it means May holidays :)

app/views/admin/settings/api_settings/show.html.erb
```diff
@@ -70,7 +70,7 @@ See COPYRIGHT and LICENSE files for more details.
   </section>
   <section class="form--section">
     <fieldset class="form--fieldset">
-      <legend class="form--fieldset-legend">Cross-Origin Resource Sharing (CORS)</legend>
+      <legend class="form--fieldset-legend"><%= I18n.t("setting_apiv3_cors_title")</legend>
```
NobodysNightmare: This is a small syntax error (ERB interpolation is not closed):

```suggestion
      <legend class="form--fieldset-legend"><%= I18n.t("setting_apiv3_cors_title") %></legend>
```

NobodysNightmare (changes requested):

NobodysNightmare (approved):

PR #17358, opened 2024-12-04
Feature/58161 global stage administration

toy:
# Ticket
[OP#58161](https://community.openproject.org/wp/58161)

Allow global stage administration

# Merge checklist

- [ ] Added/updated tests

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -278,6 +278,7 @@ group :test do
   gem "capybara_accessible_selectors", git: "https://github.com/citizensadvice/capybara_accessible_selectors", tag: "v0.12.0"
   gem "capybara-screenshot", "~> 1.0.17"
   gem "cuprite", "~> 0.15.0"
+  gem "ferrum", github: "toy/ferrum", ref: "mouse-events-buttons-property-0.15"
   gem "rspec-wait"
   gem "selenium-devtools"
   gem "selenium-webdriver", "~> 4.20"
--- a/app/components/projects/settings/life_cycle_steps/step_component.html.erb
+++ b/app/components/projects/settings/life_cycle_steps/step_component.html.erb
@@ -1,11 +1,11 @@
 <%=
   flex_layout(align_items: :center,
               justify_content: :space_between) do |step_container|
-    step_container.with_column(flex_layout: true) do |title_container|
-      title_container.with_column(pt: 1, mr: 3) do
+    step_container.with_column(flex_layout: true, mr: 2, classes: "min-width-0") do |title_container|
+      title_container.with_column(pt: 1, mr: 3, classes: "ellipsis") do
         render(Primer::Beta::Text.new(classes: 'filter-target-visible-text')) { definition.name }
       end
-      title_container.with_column(pt: 1) do
+      title_container.with_column(pt: 1, classes: "no-wrap") do
         render(Projects::LifeCycleTypeComponent.new(definition))
       end
     end
--- a/app/components/settings/project_life_cycle_step_definitions/form_header_component.html.erb
+++ b/app/components/settings/project_life_cycle_step_definitions/form_header_component.html.erb
@@ -27,10 +27,6 @@ See COPYRIGHT and LICENSE files for more details.
 
 ++#%>
 
-<% html_title t(:label_administration),
-              t("settings.project_life_cycle_step_definitions.heading"),
-              t("settings.project_life_cycle_step_definitions.#{heading_scope}.heading") %>
-
 <%= render Primer::OpenProject::PageHeader.new do |header|
   header.with_title { t("settings.project_life_cycle_step_definitions.#{heading_scope}.heading") }
   header.with_description { t("settings.project_life_cycle_step_definitions.new.description") }
--- a/app/components/settings/project_life_cycle_step_definitions/form_header_component.html.erb
+++ b/app/components/settings/project_life_cycle_step_definitions/form_header_component.html.erb
@@ -1,0 +1,38 @@
+<%#-- copyright
+OpenProject is an open source project management software.
+Copyright (C) the OpenProject GmbH
+
+This program is free software; you can redistribute it and/or
+modify it under the terms of the GNU General Public License version 3.
+
+OpenProject is a fork of ChiliProject, which is a fork of Redmine. The copyright follows:
+Copyright (C) 2006-2013 Jean-Philippe Lang
+Copyright (C) 2010-2013 the ChiliProject Team
+
+This program is free software; you can redistribute it and/or
+modify it under the terms of the GNU General Public License
+as published by the Free Software Foundation; either version 2
+of the License, or (at your option) any later version.
+
+This program is distributed in the hope that it will be useful,
+but WITHOUT ANY WARRANTY; without even the implied warranty of
+MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+GNU General Public License for more details.
+
+You should have received a copy of the GNU General Public License
+along with this program; if not, write to the Free Software
+Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+
+See COPYRIGHT and LICENSE files for more details.
+
+++#%>
+
+<% html_title t(:label_administration),
+              t("settings.project_life_cycle_step_definitions.heading"),
+              t("settings.project_life_cycle_step_definitions.#{heading_scope}.heading") %>
+
+<%= render Primer::OpenProject::PageHeader.new do |header|
+  header.with_title { t("settings.project_life_cycle_step_definitions.#{heading_scope}.heading") }
+  header.with_description { t("settings.project_life_cycle_step_definitions.new.description") }
+  header.with_breadcrumbs(breadcrumbs_items)
+end %>
--- a/app/components/settings/project_life_cycle_step_definitions/row_component.html.erb
+++ b/app/components/settings/project_life_cycle_step_definitions/row_component.html.erb
@@ -43,6 +43,9 @@ See COPYRIGHT and LICENSE files for more details.
       title_container.with_column(pt: 1) do
         render(Projects::LifeCycleTypeComponent.new(definition))
       end
+      title_container.with_column(pt: 1, ml: 3) do
+        render(Primer::Beta::Text.new) { t("project.count", count: definition.project_count) }
+      end
     end
   end
 %>
--- a/app/components/settings/project_life_cycle_step_definitions/row_component.html.erb
+++ b/app/components/settings/project_life_cycle_step_definitions/row_component.html.erb
@@ -28,10 +28,9 @@ See COPYRIGHT and LICENSE files for more details.
 ++#%>
 
 <%=
-  flex_layout(align_items: :center,
-              justify_content: :space_between) do |step_container|
-    step_container.with_column(flex_layout: true) do |title_container|
-      title_container.with_column(pt: 1, mr: 3) do
+  flex_layout(align_items: :center, justify_content: :space_between) do |row_container|
+    row_container.with_column(flex_layout: true) do |title_container|
+      title_container.with_column do
         render(Primer::Beta::Link.new(
           classes: 'filter-target-visible-text',
           href: edit_admin_settings_project_life_cycle_step_definition_path(definition),
@@ -40,10 +39,10 @@ See COPYRIGHT and LICENSE files for more details.
           definition.name
         end
       end
-      title_container.with_column(pt: 1) do
+      title_container.with_column(ml: 2) do
         render(Projects::LifeCycleTypeComponent.new(definition))
       end
-      title_container.with_column(pt: 1, ml: 3) do
+      title_container.with_column(ml: 2) do
         render(Primer::Beta::Text.new) { t("project.count", count: definition.project_count) }
       end
     end
--- a/app/components/settings/project_life_cycle_step_definitions/row_component.html.erb
+++ b/app/components/settings/project_life_cycle_step_definitions/row_component.html.erb
@@ -27,7 +27,22 @@ See COPYRIGHT and LICENSE files for more details.
 
 ++#%>
 
-<% html_title t(:label_administration), t("settings.project_life_cycle_step_definitions.heading") %>
-
-<%= render Settings::ProjectLifeCycleStepDefinitions::IndexHeaderComponent.new %>
-<%= render Settings::ProjectLifeCycleStepDefinitions::IndexComponent.new(definitions: @definitions) %>
+<%=
+  flex_layout(align_items: :center,
+              justify_content: :space_between) do |step_container|
+    step_container.with_column(flex_layout: true) do |title_container|
+      title_container.with_column(pt: 1, mr: 3) do
+        render(Primer::Beta::Link.new(
+          classes: 'filter-target-visible-text',
+          href: edit_admin_settings_project_life_cycle_step_definition_path(definition),
+          font_weight: :bold
+        )) do
+          definition.name
+        end
+      end
+      title_container.with_column(pt: 1) do
+        render(Projects::LifeCycleTypeComponent.new(definition))
+      end
+    end
+  end
+%>
--- a/app/components/settings/project_life_cycle_step_definitions/row_component.html.erb
+++ b/app/components/settings/project_life_cycle_step_definitions/row_component.html.erb
@@ -37,7 +37,7 @@ See COPYRIGHT and LICENSE files for more details.
           ))
         end
       end
-      title_container.with_column(classes: "ellipsis") do
+      title_container.with_column(classes: "ellipsis", test_selector: "project-life-cycle-step-definition-name") do
         render(
           if allowed_to_customize_life_cycle?
             Primer::Beta::Link.new(
--- a/app/components/settings/project_life_cycle_step_definitions/row_component.html.erb
+++ b/app/components/settings/project_life_cycle_step_definitions/row_component.html.erb
@@ -31,6 +31,9 @@ See COPYRIGHT and LICENSE files for more details.
   flex_layout(align_items: :center, justify_content: :space_between) do |row_container|
     row_container.with_column(flex_layout: true, classes: "gap-2") do |title_container|
       title_container.with_column do
+        render(Primer::OpenProject::DragHandle.new)
+      end
+      title_container.with_column do
         render(Primer::Beta::Link.new(
           classes: 'filter-target-visible-text',
           href: edit_admin_settings_project_life_cycle_step_definition_path(definition),
--- a/app/components/settings/project_life_cycle_step_definitions/row_component.html.erb
+++ b/app/components/settings/project_life_cycle_step_definitions/row_component.html.erb
@@ -46,5 +46,40 @@ See COPYRIGHT and LICENSE files for more details.
         render(Primer::Beta::Text.new) { t("project.count", count: definition.project_count) }
       end
     end
+    row_container.with_column do
+      render(Primer::Alpha::ActionMenu.new) do |menu|
+        menu.with_show_button(icon: "kebab-horizontal", "aria-label": t(:button_actions), scheme: :invisible)
+
+        menu.with_item(
+          label: t(:label_edit),
+          href: edit_admin_settings_project_life_cycle_step_definition_path(definition)
+        ) do |item|
+          item.with_leading_visual_icon(icon: :pencil)
+        end
+
+        unless first?
+          move_action(menu:, move_to: :highest, label: t("label_agenda_item_move_to_top"), icon: "move-to-top")
+          move_action(menu:, move_to: :higher, label: t("label_agenda_item_move_up"), icon: "chevron-up")
+        end
+        unless last?
+          move_action(menu:, move_to: :lower, label: t("label_agenda_item_move_down"), icon: "chevron-down")
+          move_action(menu:, move_to: :lowest, label: t("label_agenda_item_move_to_bottom"), icon: "move-to-bottom")
+        end
+
+        menu.with_item(
+          label: t(:text_destroy),
+          scheme: :danger,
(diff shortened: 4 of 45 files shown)


config/initializers/menus.rb
```diff
@@ -337,10 +337,16 @@
 
   menu.push :admin_projects_settings,
             { controller: "/admin/settings/project_custom_fields", action: :index },
-            if: Proc.new { User.current.admin? },
+            if: Proc.new { User.current.admin? && OpenProject::FeatureDecisions.stages_and_gates_active? },
```
ulferts: This is at the wrong menu item, @toy. It should be at `:project_life_cycle_step_definitions_settings`

toy: Indeed, thank you, I changed this menu item to also lead to life cycle and apparently changed the wrong condition before rolling back the change

app/models/application_record.rb
```diff
@@ -34,22 +34,17 @@ def read_attribute_for_validation(attribute)
...
+      table.project(table[column_name].maximum.as("max_updated_at")).to_sql
     end
       .join(" UNION ")
 
-    union_query = <<~SQL
-      SELECT MAX(union_query.max_updated_at)
+    union_query = <<~SQL.squish
+      SELECT MAX(max_updated_at)
       FROM (#{queries})
-      AS union_query
```
ulferts, line 44: I needed to readd this line locally to stop PG complaining about subqueries needing an alias.

app/components/settings/project_life_cycle_step_definitions/form_header_component.html.erb
```diff
@@ -0,0 +1,38 @@
...
+along with this program; if not, write to the Free Software
+Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+
+See COPYRIGHT and LICENSE files for more details.
+
+++#%>
+
+<% html_title t(:label_administration),
+              t("settings.project_life_cycle_step_definitions.heading"),
+              t("settings.project_life_cycle_step_definitions.#{heading_scope}.heading") %>
```
ulferts: Not a major issue but since you moved the breadcrumbs_items into a method of the component (which makes sense to me), why not do the same to the items of the html_title?

ulferts: But on testing the functionality, it seems that setting the `html_title` within the component does not have the desired effect. I did not check why. But moving the code into the `form.html.erb` solves the problem.

app/controllers/admin/settings/project_life_cycle_step_definitions_controller.rb
```diff
@@ -0,0 +1,138 @@
...
+# Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+#
+# See COPYRIGHT and LICENSE files for more details.
+#++
+
+module Admin::Settings
+  class ProjectLifeCycleStepDefinitionsController < ::Admin::SettingsController
+    menu_item :project_life_cycle_step_definitions_settings
+
+    helper_method :allowed_to_customize_life_cycle?
```
ulferts: Why is this code placed in the controller? The "normal rails way" would be to have it in an actual helper file. Another way would be to have a module that is included into the components. That would avoid passing the value around (at the cost of having a less visible dependency to the EnterpriseToken)

app/components/settings/project_life_cycle_step_definitions/row_component.html.erb
```diff
@@ -0,0 +1,103 @@
...
+            item.with_leading_visual_icon(icon: :pencil)
+          end
+
+          unless first?
+            move_action(menu:, move_to: :highest, label: t("label_agenda_item_move_to_top"), icon: "move-to-top")
+            move_action(menu:, move_to: :higher, label: t("label_agenda_item_move_up"), icon: "chevron-up")
+          end
+          unless last?
+            move_action(menu:, move_to: :lower, label: t("label_agenda_item_move_down"), icon: "chevron-down")
+            move_action(menu:, move_to: :lowest, label: t("label_agenda_item_move_to_bottom"), icon: "move-to-bottom")
```
ulferts, line 83: This references a string introduced for meetings and the key is also defined in the meeting's i18n file. Searching for the key that this is not the first time this was done. In fact, the whole implementation of the action items for moving by now exists 5 times. Probably not as part of this PR but generally, extracting those lines into some shared code along with moving the i18n keys seems prudent.

app/components/settings/project_life_cycle_step_definitions/row_component.html.erb
```diff
@@ -0,0 +1,103 @@
...
+            href: admin_settings_project_life_cycle_step_definition_path(definition),
+            form_arguments: {
+              method: :delete,
+              data: {
+                confirm: t("text_are_you_sure_with_project_life_cycle_step")
+              }
+            }
+          ) do |item|
+            item.with_leading_visual_icon(icon: :trash)
+          end
```
ulferts, line 98: The lookbook defines a delete dialog that is applicable here. But it also says that browser's confirm dialogs should be used for now. I could imagine using the new [Danger zone dialog](https://community.openproject.org/wp/58680) in this case once it is done. Again, not part of this PR then.

app/controllers/admin/settings/project_life_cycle_step_definitions_controller.rb
```diff
@@ -0,0 +1,138 @@
...
+
+    def drop
+      if @definition.update(params.permit(:position))
+        flash[:notice] = I18n.t(:notice_successful_update)
+      else
+        # TODO: handle better
+        flash[:error] = I18n.t(:notice_bad_request)
+      end
+
+      redirect_to action: :index, status: :see_other
```
ulferts: This redirect leads to problems since the drop action is called with a turbo stream request which results in the index action also being called with `Accept: text/vnd.turbo-stream.html` on the redirected request. 

It might be that the problem simply vanishes by not doing a redirect at all. The only reason for that redirect I currently see is to have the flash message displayed. And this opens up the question to ask: Do we actually want those flash messages to be displayed when drag&drop is used? 

It seems that on other pages where we do drag&drop, no flash is displayed. I don't know if that has been explicitly agreed upon as a UX pattern, @HDinger. Other interactions like the ordering via menu item and also using the toggle buttons on e.g. project attributes does not trigger a flash message. So it seems to me that we differentiate between different kinds of interactions. 

If those smaller ones don't need to have a flash message, then not redirecting here would be fine. What to do in an error case might be a different story.

app/components/settings/project_life_cycle_step_definitions/row_component.html.erb
```diff
@@ -0,0 +1,103 @@
...
+          definition.name
+        end
+      end
+      title_container.with_column do
+        render(Projects::LifeCycleTypeComponent.new(definition))
+      end
+      title_container.with_column do
+        render(Primer::Beta::Text.new) { t("project.count", count: definition.project_count) }
+      end
+    end
```
ulferts, line 63: Probably an edge case but this layout struggles with very long names for the definition:

image

Admittedly, the project specific lifecycle administration suffers from the same limitations.

app/models/project/life_cycle_step_definition.rb
```diff
@@ -42,15 +44,19 @@ class Project::LifeCycleStepDefinition < ApplicationRecord
```
ulferts, line 45: Only with the administration now existing it becomes apparent, that the user is not prevented from creating two steps with the same name.

app/forms/projects/life_cycle_step_definitions/form.rb
```diff
@@ -0,0 +1,58 @@
...
+        name: :color,
+        required: true
+      )
+
+      f.submit(
+        scheme: :primary,
+        name: :submit,
+        label: submit_label
+      )
+    end
```
ulferts, line 55: The field for the color is longer than the input above. 

image

That is due to a wrapper applied to the primer input field `text_field` which is not applied to the op input field `color_select_list`. I suggest to fix this in a separate PR.

app/helpers/errors_helper.rb
```diff
@@ -33,6 +33,12 @@ def render_400(options = {}) # rubocop:disable Naming/VariableNumber
     false
   end
 
+  def render_402(options = {}) # rubocop:disable Naming/VariableNumber
+    unset_template_magic
+    render_error({ message: :notice_requires_enterprise_token, status: 402 }.merge(options))
+    false
+  end
```
ulferts, line 40: I like this addition. Makes sense to use that status code.

app/helpers/highlighting_helper.rb
```diff
@@ -4,6 +4,6 @@ def highlight_css_version_tag(max_updated_at = highlight_css_updated_at)
   end
 
   def highlight_css_updated_at
-    ApplicationRecord.most_recently_changed Status, IssuePriority, Type, UserPreference
+    ApplicationRecord.most_recently_changed Status, IssuePriority, Type, UserPreference, Project::LifeCycleStepDefinition
```
ulferts, line 7: 👍  thanks for fixing

app/models/application_record.rb
```diff
@@ -34,22 +34,17 @@ def read_attribute_for_validation(attribute)
...
     SQL
 
-    ActiveRecord::Base
-      .connection
-      .select_all(union_query)
-      .rows
-      &.first # first result row
-      &.first # max column
+    ActiveRecord::Base.connection.select_value(union_query)
```
ulferts, line 48: Thanks for the cleanup

app/views/admin/settings/project_life_cycle_step_definitions/form.html.erb
```diff
@@ -0,0 +1,40 @@
...
+along with this program; if not, write to the Free Software
+Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+
+See COPYRIGHT and LICENSE files for more details.
+
+++#%>
+
+<%= render Settings::ProjectLifeCycleStepDefinitions::FormHeaderComponent.new(@definition) %>
+
+<%= error_messages_for @definition %>
```
ulferts: For now at least, this is not necessary to have as all errors can be attributed to an input.

config/locales/en.yml
```diff
@@ -3309,9 +3316,9 @@ en:
       are_you_sure: "Are you sure you want to archive the project '%{name}'?"
       archived: "Archived"
     count:
-      zero: "0 Projects"
-      one: "1 Project"
-      other: "%{count} Projects"
+      zero: "0 projects"
+      one: "1 project"
+      other: "%{count} projects"
```
ulferts, line 3356: 👍 for fixing

config/locales/en.yml
```diff
@@ -3729,6 +3736,27 @@ en:
       new:
         heading: "New attribute"
         description: "Changes to this project attribute will be reflected in all projects where it is enabled. Required attributes cannot be disabled on a per-project basis."
+    project_life_cycle_step_definitions:
+      heading: "Project lifecycle"
+      heading_description: "Project lifecycle define the project stages and gates that can be used for your project planning and will appear in the overview page of each project. These attributes can be enabled or disabled but not re-ordered at a project level."
```
ulferts: There is a grammar error in here. It would have to be "Project lifecycle defines" (with the s). 

In any case, I suggest to rephrase the text a bit:

"The project lifecycle consists of stages and gates that structure your projects. They will appear in the overview page of a project and be available in the projects list. Individual stages and gates can be selected to customize a project but reordering is not possible on a project level."

app/models/application_record.rb
toy: Good catch, I accidentally stayed on 16 since some tests not long ago, and exactly in version 16 requirement for naming subquery was removed, will switch back to 13

ulferts (changes requested):
Hey @toy, 

I am aware that the PR is not yet done so sorry if I criticised anything you would have done anyway.

But the general direction is already looking good so I thought I could start on the details :).

app/components/settings/project_life_cycle_step_definitions/form_header_component.html.erb
toy: I already fixed it yesterday, but didn't push till few minutes ago. I had to move `html_title` from component to the action view, as `html_title` sets an instance variable, so inside component will have no effect on variable used to set page title

app/controllers/admin/settings/project_life_cycle_step_definitions_controller.rb
toy: This was also already changed yesterday, but not pushed

app/controllers/admin/settings/project_life_cycle_step_definitions_controller.rb
toy: The method is used both in the controller (in `require_enterprise_token` method) and components. It still can be extracted into a module to not pass it around, I'll see if it looks better

app/components/settings/project_life_cycle_step_definitions/row_component.html.erb
toy: Fixed for both:
image
image

config/locales/en.yml
toy: @marcalcobe

app/controllers/admin/settings/project_life_cycle_step_definitions_controller.rb
toy: Looks better with the module, if not for the copyright header it would even be a net negative change

app/components/settings/project_life_cycle_step_definitions/row_component.html.erb
toy: I agree, but better in a separate PR
https://community.openproject.org/wp/60231

app/models/project/life_cycle_step_definition.rb
toy: Added constraint

app/forms/projects/life_cycle_step_definitions/form.rb
ulferts: I added a [PR that enables setting the `input_width` on a color select](https://github.com/opf/openproject/pull/17482). This could then be applied here later on.

db/migrate/20241217190533_add_uniqueness_index_to_project_life_cycle_step_definitions_name.rb
```diff
@@ -0,0 +1,5 @@
+class AddUniquenessIndexToProjectLifeCycleStepDefinitionsName < ActiveRecord::Migration[7.1]
+  def change
+    add_index :project_life_cycle_step_definitions, :name, unique: true
```
ulferts: 👍 for being thorough.

config/locales/en.yml
marcalcobe: Okay for me.

app/components/projects/settings/life_cycle_steps/step_component.html.erb
```diff
@@ -1,11 +1,11 @@
...
               justify_content: :space_between) do |step_container|
-    step_container.with_column(flex_layout: true) do |title_container|
-      title_container.with_column(pt: 1, mr: 3) do
+    step_container.with_column(flex_layout: true, mr: 2, classes: "min-width-0") do |title_container|
+      title_container.with_column(pt: 1, mr: 3, classes: "ellipsis") do
         render(Primer::Beta::Text.new(classes: 'filter-target-visible-text')) { definition.name }
       end
-      title_container.with_column(pt: 1) do
+      title_container.with_column(pt: 1, classes: "no-wrap") do
         render(Projects::LifeCycleTypeComponent.new(definition))
```
ulferts, line 9: Merci beaucoup 🙇

lookbook/previews/open_project/enterprise_edition/banner_component_preview.rb
```diff
@@ -57,7 +57,7 @@ class BannerComponentPreview < Lookbook::Preview
       def default
         render(
           ::EnterpriseEdition::BannerComponent
-            .new(:form_configuration,
+            .new(:customize_life_cycle,
```
ulferts, line 60: Definitely reads better in the preview now 👍

app/controllers/admin/settings/project_life_cycle_step_definitions_controller.rb
```diff
@@ -0,0 +1,143 @@
...
+        flash[:notice] = I18n.t(:notice_successful_update)
+        redirect_to action: :index, status: :see_other
+      else
+        render :form, status: :unprocessable_entity
+      end
+    end
+
+    def destroy
+      if @definition.destroy
+        update_flash_message_via_turbo_stream(message: I18n.t(:notice_successful_delete))
```
ulferts: This should be a success message so adding `scheme: :success` is appropriate. 

To make it simpler for the next dev, you could add a `render_success_flash_message_via_turbo_stream` similar to the error case.

ulferts (changes requested):
I only found one thing that actually needs fixing and that one is tiny.

db/migrate/20241217190533_add_uniqueness_index_to_project_life_cycle_step_definitions_name.rb
toy: Extracted to #17494 and already merged

app/controllers/admin/settings/project_life_cycle_step_definitions_controller.rb
toy: I was not sure if it should be success or informative message, but you are right, it is a success one

lib/open_project/static/links.rb
```diff
@@ -268,6 +268,9 @@ def static_links
               },
               status_read_only: {
                 href: "https://www.openproject.org/docs/system-admin-guide/manage-work-packages/work-package-status/#create-a-new-work-package-status"
+              },
+              customize_life_cycle: {
+                href: "https://www.openproject.org/enterprise-edition" # TODO: update
```
EinLama, line 273: Just a reminder that the target page is still pending and should be inserted here later.

Gemfile
```diff
@@ -278,6 +278,7 @@ group :test do
   gem "capybara_accessible_selectors", git: "https://github.com/citizensadvice/capybara_accessible_selectors", tag: "v0.12.0"
   gem "capybara-screenshot", "~> 1.0.17"
   gem "cuprite", "~> 0.15.0"
+  gem "ferrum", github: "toy/ferrum", ref: "mouse-events-buttons-property-0.15"
```
EinLama, line 281: Great job fixing the bug within ferrum. As soon as a new version with [your merged PR](https://github.com/rubycdp/ferrum/pull/509) is released, we can use the original gem once again. 🚀 Do you happen to know when they plan to release a new version?

config/locales/en.yml
```diff
@@ -1906,6 +1909,8 @@ en:
     upsale:
       title: "Enterprise add-on"
       link_title: "More information"
+      customize_life_cycle:
+        description: "Create and organise different project stages and gates than the ones provided by PM2 project cycle planning."
```
EinLama: Using American English:

```suggestion
        description: "Create and organize different project stages and gates than the ones provided by PM2 project cycle planning."
```

app/models/project/life_cycle_step_definitions/scopes/with_project_count.rb
```diff
@@ -0,0 +1,50 @@
...
+        project_counts = Project::LifeCycleStep
+          .where(active: true)
+          .group(:definition_id)
+          .select(:definition_id, "COUNT(project_id) AS count")
+
+        Project::LifeCycleStepDefinition
+          .with(project_counts:)
+          .joins("LEFT OUTER JOIN project_counts ON #{quoted_table_name}.id = project_counts.definition_id")
+          .select("#{quoted_table_name}.*")
+          .select("COALESCE(project_counts.count, 0) AS project_count")
```
EinLama, line 46: Very elegant, nice! 💪🏻

spec/features/admin/project_life_cycle_step_definitions_spec.rb
```diff
@@ -0,0 +1,142 @@
...
+# You should have received a copy of the GNU General Public License
+# along with this program; if not, write to the Free Software
+# Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+#
+# See COPYRIGHT and LICENSE files for more details.
+# ++
+
+require "spec_helper"
+
+RSpec.describe "Projects life cycle settings", :js, :with_cuprite, with_flag: { stages_and_gates: true } do
```
EinLama, line 33: This is a very nice spec. Easy to read, yet testing thoroughly 👍🏻 

Did you consider adding one spec that checks what happens when the feature flag is not set? Not sure if it really adds a lot here.

EinLama (approved):
Great PR, works like a charm and offers beautiful, readable code 👏🏻 very well done! I learned a thing or two going through it.

I have dropped some reminders for open TODOs that you are already aware of, just for good measure.

The rest is very small nitpicky stuff. Fix or ignore at your own leisure and feel free to merge once you're done.

lib/open_project/static/links.rb
toy: Already created https://community.openproject.org/wp/60343 to not forget

cc @MayaBerd

Gemfile
toy: No idea, I only saw that allowing ferrum dependency to be 0.16.* was merged in cuprite.

Not required for this PR, but I created rubycdp/cuprite#288 to scroll to destination when dragging.

spec/features/admin/project_life_cycle_step_definitions_spec.rb
toy: I did consider and thought just a bit more on the side that it is not important enough to justify adding it, but given that you mention it, I just added it

ulferts (approved):

PR #17366, opened 2024-12-05
Make it possible to use workers in TLS setup

NobodysNightmare:
The example overrides file was missing overrides
for the workers, which means their network was not configured to include the gateway network, thus they didn't resolve IP addresses to the Traefik proxy.

Even if they had, it was also necessary to setup environment variables the same way that the backend service gets them, because otherwise the development TLS certificates would not be recognized and configuration of KeyCloak would not be taken into account for workers, which may be an issue in the future.

# Ticket
No ticket

Extracted environment variables into a shared section so they only need to be edited once (e.g. KeyCloak credentials) for both the backend and the worker.

I was considering to do the same for certificate volume mounts, because they are kinda repetitive, but there may be use cases where one wants to include additional volume mounts next to the certificate. As far as I can tell, this would not be easily possible, since the volume mounts are an array, not a map.

# Merge checklist

- [ ] ~~Added/updated tests~~
- [ ] ~~Tested major browsers (Chrome, Firefox, Edge, ...)~~

### DIFF
--- a/docker/dev/tls/docker-compose.core-override.example.yml
+++ b/docker/dev/tls/docker-compose.core-override.example.yml
@@ -1,27 +1,44 @@
+# Some containers (e.g. backend and workers) need some variables to be configured properly
+x-op-env-override: &environment
+  OPENPROJECT_CLI_PROXY: "${OPENPROJECT_DEV_URL}"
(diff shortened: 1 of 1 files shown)


brunopagno (approved):
Seems correct to me.

I needed to do the same, but I did the dumb version of copy+paste between `backend` and `worker` :grin:. This looks better.

PR #17369, opened 2024-12-05
[#59883] Pagination for recurring meetings

mrmir:
# Ticket

https://community.openproject.org/work_packages/59883

# Open points
I changed the schedule text in the footer to be the same as the header subtitle as I don't think it makes sense to have two slightly different strings for that. 

Does this work?

# Merge checklist

- [x] Added/updated tests


oliverguenther (approved):
LGTM, nice 👏

PR #17370, opened 2024-12-05
Simplify provider name shortening

NobodysNightmare:
Instead of using a regular expression to parse the name, we try to find the provider name inside the already existing constant that knows providers short names.

# Ticket
No ticket

Primarily: Simplify code, increase readability.

Secondarily: Side step a glitch in my VS Code that breaks syntax highlighting around the Regex :sunglasses: 

Instead of using a regex to parse the provider name, we perform a lookup in a hash that already existed before.

This changes behaviour for cases that are not covered by specs so far: Previously the method would have returned a result for unknown stores (say `Storages::DropboxStorage`), while the new method will only work for storages that are "registered" via the `PROVIDER_TYPE_SHORT_NAMES` constant.

I think this is consistent behaviour, but still it's a change in behaviour. Looking at the `StoragesController`, there seems to be a check that a selected provider is included in the constant, so it's unlikely that the change will affect any storage provider that exists in the wild.

# Merge checklist

- [x] Added/updated tests
- [ ] ~~Tested major browsers (Chrome, Firefox, Edge, ...)~~

### DIFF
--- a/modules/storages/app/models/storages/storage.rb
+++ b/modules/storages/app/models/storages/storage.rb
@@ -92,14 +92,12 @@ module Storages
     }.freeze, _prefix: :health
 
     def self.shorten_provider_type(provider_type)
-      case /Storages::(?'provider_name'.*)Storage/.match(provider_type)
-      in provider_name:
-        provider_name.underscore
-      else
-        raise ArgumentError,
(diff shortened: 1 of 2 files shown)


Kharonus (approved):
Nice improvement

PR #17374, opened 2024-12-05
Bug/59458 activity tab does not scroll to the bottom or correct comment on mobile

jjabari-op:
[Ticket](https://community.openproject.org/work_packages/59458)


akabiru (approved):
Manual QA looks good 👍🏾  In this instance, a brief write-up of your approach would have been helpful- we have this in the _"What approach did you take and why"_ in the pull request template

PR #17375, opened 2024-12-05
Storages new folder API endpoint

brunopagno:
# Ticket
https://community.openproject.org/wp/59900

The goal is to expose an API for our frontend to call to create new folders on a storage.

It's an API endpoint, and a service that call commands that already exist

# Merge checklist

- [X] Added/updated tests

### DIFF
--- a/docs/api/apiv3/components/schemas/storage_folder_write_model.yml
+++ b/docs/api/apiv3/components/schemas/storage_folder_write_model.yml
@@ -1,0 +1,13 @@
+# Schema: StorageFolderWriteModel
+---
+type: object
+required:
+  - name
+  - parentId
+properties:
+  name:
+    type: string
+    description: Name of the folder to be created
+  parentId:
+    type: string
+    description: Unique identifier of the parent folder in which the new folder should be created in
--- a/docs/api/apiv3/paths/storage_folders.yml
+++ b/docs/api/apiv3/paths/storage_folders.yml
@@ -1,0 +1,68 @@
+# /api/v3/storages/{id}/folders
+---
+post:
+  summary: Creation of a new folder
+  operationId: create_storage_folder
+  tags:
+    - File links
+  description: Creates a new folder under the given parent
+  parameters:
+    - name: id
+      description: Storage id
+      in: path
+      required: true
+      schema:
+        type: integer
+      example: 1337
+  requestBody:
+    content:
+      application/json:
+        schema:
+          $ref: '../components/schemas/storage_folder_write_model.yml'
+        examples:
+          'Valid example':
+            $ref: '../components/examples/storage-create-folder-request-body.yml'
+  responses:
+    '201':
+      description: Created
+      content:
+        application/hal+json:
+          schema:
+            $ref: '../components/schemas/storage_file_model.yml'
+    '400':
+      content:
+        application/hal+json:
+          schema:
+            $ref: '../components/schemas/error_response.yml'
+          example:
+            _type: Error
+            errorIdentifier: urn:openproject-org:api:v3:errors:InvalidQuery
+            message: The given parent is not a directory.
+      description: |-
+        Returned if the request is missing a required parameter.
+    '403':
+      content:
+        application/hal+json:
+          schema:
+            $ref: '../components/schemas/error_response.yml'
+          example:
+            _type: Error
+            errorIdentifier: urn:openproject-org:api:v3:errors:MissingPermission
+            message: You are not authorized to access this resource.
+      description: |-
+        Returned if the client does not have sufficient permissions.
+
+        **Required permission:** manage file links
+    '404':
+      content:
+        application/hal+json:
+          schema:
+            $ref: '../components/schemas/error_response.yml'
+          example:
+            _type: Error
+            errorIdentifier: urn:openproject-org:api:v3:errors:NotFound
+            message: The requested resource could not be found.
+      description: |-
+        Returned if the storage does not exist or the client does not have sufficient permissions to see it.
+
+        **Required permission:** view file links
--- a/modules/storages/app/services/storages/create_folder_service.rb
+++ b/modules/storages/app/services/storages/create_folder_service.rb
@@ -1,0 +1,77 @@
+# frozen_string_literal: true
+
+#-- copyright
+# OpenProject is an open source project management software.
+# Copyright (C) the OpenProject GmbH
+#
+# This program is free software; you can redistribute it and/or
+# modify it under the terms of the GNU General Public License version 3.
+#
+# OpenProject is a fork of ChiliProject, which is a fork of Redmine. The copyright follows:
+# Copyright (C) 2006-2013 Jean-Philippe Lang
+# Copyright (C) 2010-2013 the ChiliProject Team
+#
+# This program is free software; you can redistribute it and/or
+# modify it under the terms of the GNU General Public License
+# as published by the Free Software Foundation; either version 2
+# of the License, or (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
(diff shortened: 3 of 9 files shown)


docs/api/apiv3/components/schemas/storage_folder_write_model.yml
```diff
@@ -0,0 +1,13 @@
+# Schema: StorageFolderWriteModel
+---
+type: object
+required:
+  - name
+  - path
```
Kharonus: 🔴 rename to `parentId`

docs/api/apiv3/components/schemas/storage_folder_write_model.yml
```diff
@@ -0,0 +1,13 @@
...
+properties:
+  name:
+    type: string
+    description: Folder name
+  parentId:
+    type: integer
```
Kharonus: 🔴 type must be string

docs/api/apiv3/components/schemas/storage_folder_write_model.yml
```diff
@@ -0,0 +1,13 @@
+# Schema: StorageFolderWriteModel
+---
+type: object
+required:
+  - name
+  - path
+properties:
+  name:
+    type: string
+    description: Folder name
```
Kharonus: 🟡 Being more specific woudl be nice, like "The name of the folder to be created" and for parent id something like "The unique identifier of the parent folder the new folder is created in"

docs/api/apiv3/paths/storage_folders.yml
```diff
@@ -0,0 +1,28 @@
+# /api/v3/storages/{id}/folders
+---
+post:
+  summary: Creation of a new folder
+  operationId: storage_new_folder
```
Kharonus: 🟡  opoeration id pattern is: `create_storage_folder`

docs/api/apiv3/paths/storage_folders.yml
```diff
@@ -0,0 +1,28 @@
...
+  tags:
+    - File links
+  description: Creates a new folder on the given path
```
Kharonus: 🟡 description is still written for "path", should be mentioning a specific parent folder.

docs/api/apiv3/paths/storage_folders.yml
```diff
@@ -0,0 +1,28 @@
...
+      application/json:
+        schema:
+          $ref: '../components/schemas/storage_folder_write_model.yml'
+  responses:
+    '201':
+      description: Created
+      content:
+        application/hal+json:
+          schema:
+            $ref: '../components/schemas/storage_file_model.yml'
```
Kharonus, line 31: 🔴 error cases are missing. I'd assume 400, 404, 403.

docs/api/apiv3/paths/storage_folders.yml
```diff
@@ -0,0 +1,28 @@
...
+      in: path
+      required: true
+      schema:
+        type: integer
+      example: 1337
+  requestBody:
+    content:
+      application/json:
+        schema:
+          $ref: '../components/schemas/storage_folder_write_model.yml'
```
Kharonus, line 21: 🟡 add a small, simple, named example as a separate file and $ref it here. Same for response

modules/storages/lib/api/v3/storage_files/storage_folders_api.rb
```diff
@@ -0,0 +1,80 @@
...
+        helpers ::Storages::Peripherals::StorageFileInfoConverter,
+                ::Storages::Peripherals::StorageErrorHelper
+
+        helpers do
+          def parent_path_from_parent_id(parent_id)
+            ::Storages::StorageFileService
+              .call(storage: @storage, user: current_user, file_id: parent_id)
+              .match(
+                on_success: lambda { |folder_info|
+                  path = URI.decode_uri_component(folder_info.location)
```
brunopagno: Not sure if I am missing something here, but I needed to decode URI characters for the fetched information from the provider

modules/storages/lib/api/v3/storage_files/storage_folders_api.rb
```diff
@@ -0,0 +1,80 @@
...
+                  ::Storages::Peripherals::ParentFolder.new(path)
+                },
+                on_failure: ->(error) { raise_error(error) }
+              )
+          end
+        end
+
+        resources :folders do
+          params do
+            requires :name, type: String, desc: "Folder name"
```
brunopagno, line 41: self reminder to double check parameters

modules/storages/app/services/storages/create_folder_service.rb
```diff
@@ -0,0 +1,43 @@
...
+module Storages
+  class CreateFolderService < BaseService
+    def self.call(storage:, user:, name:, parent_location:)
+      new.call(storage:, user:, name:, parent_location:)
+    end
+
+    def call(storage:, user:, name:, parent_location:)
+      auth_strategy = Peripherals::Registry.resolve("#{storage}.authentication.user_bound").call(user: user)
+
+      ::Storages::Peripherals::Registry
```
mereghost: Since you are already on the `Storages` module you can call just `Peripherals::Registry`.

modules/storages/lib/api/v3/storage_files/storage_folders_api.rb
mereghost: Also on OneDrive, `FileInfo#location` isn't what you would send up to `CreateFolder`, it should be `id`

Annoying? Yes. Terribly so.

mereghost: WDYT of moving this method to the create folder service under another name like `CreateFolderService.with_parent_id(folder_name, parent_id)`, so that the resource ends up with no special business logic?

modules/storages/spec/requests/api/v3/storages/storage_folders_api_spec.rb
```diff
@@ -0,0 +1,133 @@
...
+        name: "/",
+        location: "/"
+      )
+    end
+
+    before do
+      Storages::Peripherals::Registry.stub(
+        "nextcloud.queries.file_info",
+        ->(_) { ServiceResult.success(result: file_info) }
+      )
```
mereghost: An alternative to just replacing it with a lambda, now that we have validating mocks in place, would be to stub the command/query itself.

It is a bit more verbose, but can help with detecting bad argument passing.

```ruby
file_info = class_double(Storages::Peripherals::StorageInteraction::Nextcloud::FileInfoQuery)
allow(file_info).to receive(:call).with(valid: :arguments).and_return(ServiceResult.success)

Storages::Peripherals::Registry.stub('nextcloud.queries.file_info', file_info)
```

mereghost (changes requested):
Looks good overall. It might need some more tests using OneDrive as the base storage.

modules/storages/spec/requests/api/v3/storages/storage_folders_api_spec.rb
brunopagno: Oh, I like this :clap: 
Will give it a try

modules/storages/lib/api/v3/storage_files/storage_folders_api.rb
brunopagno: Moved the logic to the service!

modules/storages/spec/requests/api/v3/storages/storage_folders_api_spec.rb
brunopagno: I adjusted the specs here. It actually instantly caught a small issue with my setup. Looking better I think :)

mereghost (approved):
Nice work!

PR #17376, opened 2024-12-05
Allow editing and auto-discovering supported grant types of OIDC providers

NobodysNightmare:
This is in preparation to enable authentication flows based on token exchange in the future. Using the `grant_types_supported` we can determine whether the IDP is capable of exchanging tokens and then use this information to indicate to users whether they are able to set-up token exchange based authentication for their storage provider.

# Ticket
[OP#58862](https://community.openproject.org/projects/cross-application-user-integration-stream/work_packages/58862)

Allow to decide for a given IDP, whether it's able to perform token exchange or not.

We store the list of grant types available on the IDP, since this is a feature agnostic way to store information about the provider. This information is then used to implement the `Provider#token_exchange_capable?` method, which is specific to the token exchange feature.

For editing the discovered information, I kept the UI simple by using a text input that accepts the grants in a space-delimited fashion. This solution was merely chosen to keep implementation complexity low, but is probably not the best UX we can offer. Assuming that most users will fill this information through a discovery endpoint, the solution might be sufficient, though.

# Screenshots

image

# Merge checklist

- [x] Added/updated tests
- [x] Tested ~~major browsers (Chrome, Firefox, Edge, ...)~~ in one browser

### DIFF
--- a/config/initializers/feature_decisions.rb
+++ b/config/initializers/feature_decisions.rb
@@ -61,3 +61,8 @@ end
 
 OpenProject::FeatureDecisions.add :stages_and_gates,
                                   description: "Enables the under construction feature of stages and gates."
+
+OpenProject::FeatureDecisions.add :oidc_token_exchange,
+                                  description: "Enables the under construction OAuth2 token exchange, allowing " \
+                                               "users to interact with storage providers without consenting " \
+                                               "in OAuth screens before first use."
--- a/modules/openid_connect/app/models/openid_connect/provider.rb
+++ b/modules/openid_connect/app/models/openid_connect/provider.rb
@@ -3,12 +3,12 @@ module OpenIDConnect
     include HashBuilder
 
     OIDC_PROVIDERS = %w[google microsoft_entra custom].freeze
-    DISCOVERABLE_ATTRIBUTES_MANDATORY = %i[authorization_endpoint
-                                           userinfo_endpoint
-                                           token_endpoint
-                                           issuer].freeze
-    DISCOVERABLE_ATTRIBUTES_OPTIONAL = %i[end_session_endpoint jwks_uri].freeze
-    DISCOVERABLE_ATTRIBUTES_ALL = DISCOVERABLE_ATTRIBUTES_MANDATORY + DISCOVERABLE_ATTRIBUTES_OPTIONAL
+    DISCOVERABLE_STRING_ATTRIBUTES_MANDATORY = %i[authorization_endpoint
+                                                  userinfo_endpoint
+                                                  token_endpoint
+                                                  issuer].freeze
+    DISCOVERABLE_STRING_ATTRIBUTES_OPTIONAL = %i[end_session_endpoint jwks_uri].freeze
+    DISCOVERABLE_STRING_ATTRIBUTES_ALL = DISCOVERABLE_STRING_ATTRIBUTES_MANDATORY + DISCOVERABLE_STRING_ATTRIBUTES_OPTIONAL
 
     MAPPABLE_ATTRIBUTES = %i[login email first_name last_name admin].freeze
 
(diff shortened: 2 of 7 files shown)


NobodysNightmare:
As noted in the original story: I am not sure whether we should merge this PR. Now that I understand why the feature was requested, I think it's preferrable to stick with "Custom", also for Keycloak providers.

NobodysNightmare:
Re: Rubocop/Reviewdog

> Spec path should end with `openid_connect/provider*_spec.rb`.

Is this intended or wrongly configured? I see that the spec right next to this one ignores this rule as well (it's not placed inside an `openid_connect` folder. Considering that all our specs are inside `modules/openid_connect`, this cop should probably be adapted.

Though, I am surprised that it's not already adapted correctly, so I assume there's a reason for that.

config/initializers/feature_decisions.rb
```diff
@@ -61,3 +61,8 @@
 
 OpenProject::FeatureDecisions.add :stages_and_gates,
                                   description: "Enables the under construction feature of stages and gates."
+
+OpenProject::FeatureDecisions.add :oidc_token_exchange,
```
NobodysNightmare, line 65: I just realized that some feature flags are defined on the scope of modules. Though I wouldn't know where to put this, since this flag will be crossing boundaries between `storages` and `openid_connect`.

Opinions?

modules/openid_connect/app/models/openid_connect/provider.rb
```diff
@@ -3,26 +3,28 @@ class Provider < AuthProvider
     include HashBuilder
 
     OIDC_PROVIDERS = %w[google microsoft_entra custom].freeze
-    DISCOVERABLE_ATTRIBUTES_MANDATORY = %i[authorization_endpoint
-                                           userinfo_endpoint
-                                           token_endpoint
-                                           issuer].freeze
-    DISCOVERABLE_ATTRIBUTES_OPTIONAL = %i[end_session_endpoint jwks_uri].freeze
-    DISCOVERABLE_ATTRIBUTES_ALL = DISCOVERABLE_ATTRIBUTES_MANDATORY + DISCOVERABLE_ATTRIBUTES_OPTIONAL
+    DISCOVERABLE_STRING_ATTRIBUTES_MANDATORY = %i[authorization_endpoint
```
brunopagno, line 6: Throwing a useless opinion here, but I don't feel the change in name here makes it any more meaningful. Do we somehow differentiate `string` from non `string` attributes?

config/initializers/feature_decisions.rb
brunopagno: I could see an argument for both in the module or in global space. Would be useful to know of a previous occurrence of such problem :sweat_smile: 

But I think it's okay to keep it here since it's already written.

NobodysNightmare: I think following a recent discussion on Matrix, I'd now go for **this** location, because the feature is crossing boundaries and apparently putting all things into modules isn't of utmost importance.

brunopagno (approved):
This looks pretty cool, although I have limited understanding of the OIDC topic :grimacing: 

Would be nice to have some visual information on the PR to understand better the UI changes, but from the code I think it's okay as is.

Left some ignorable comments

:heavy_check_mark:

modules/openid_connect/app/models/openid_connect/provider.rb
NobodysNightmare: Yes we do, this was the reason that I added this naming change. So far these constants were impacting all attributes equally and the constants are used fot things like "create string accessors for all of those".

However, the attribute that I just added was an array of strings, not a string. Thus I did the renaming to indicate "that's not all discoverable attributes, but just discoverable attributes of type string".

brunopagno: Fair enough. Sounds correct to me. :clap:

NobodysNightmare:
@brunopagno I added a screenshot to the PR description. I think the UI is an important limitation in this PR, because I put simplicity over the best possible UX here.

I still think that's fair, because I expect least people to have to manually change this stuff, because auto-discovery is how this is usually set up.

brunopagno:
@NobodysNightmare nice! I can think of two possible alternatives if we want to make this more defensive/explicit to the user.

1. Put a hint text below (or somewhere where it makes sense) saying something like "you should only change this if you know what you're doing".
2. Make the field disabled by default and have a button to explicit enable editting it.

Ultimately this falls into the same discussion we were having the other day about hand holding users. I can see someone accidentally break things by editting this field. But I would also be okay with pushing the current implementation and keeping track of improvements for the future so the :heavy_check_mark: stays :sweat_smile:.

...

All the above said, I think this should be a product decision. I don't have all the OIDC context to be able to say confidently who is going to use this and how so I might be guessing the wrong direction :grimacing:

NobodysNightmare:
Cool. I am going to merge this MR as-is for now. The "only change if you know what you are doing" comment is actually part of the existing UI already (kind of):

image

PR #17387, opened 2024-12-06
Adds handling of Hierarchy fields to the JournalFormatters

mereghost:
⚠️ Related WP: [OP#59855](https://community.openproject.org/projects/document-workflows-stream/work_packages/59855)

The CustomFieldFormatter didn't know how to handle the hierarchy type fields, therefore rendered the incorrect representation of the changes.

Here is a screenshot of the updated behaviour
image

### DIFF
--- a/app/models/custom_field/hierarchy/item.rb
+++ b/app/models/custom_field/hierarchy/item.rb
@@ -37,4 +37,8 @@ class CustomField::Hierarchy::Item < ApplicationRecord
   scope :including_children, -> { includes(children: :children) }
 
   def to_s = short.nil? ? label : "#{label} (#{short})"
+
+  def ancestry_path
+    self_and_ancestors.filter_map(&:to_s).reverse.join(" / ")
+  end
 end
--- a/lib/open_project/journal_formatter/custom_field/plain.rb
+++ b/lib/open_project/journal_formatter/custom_field/plain.rb
@@ -40,9 +40,7 @@ class OpenProject::JournalFormatter::CustomField::Plain < JournalFormatter::Base
     label = label_for_custom_field(custom_field)
 
     old_value, value = if custom_field
-                         coisa = get_formatted_values(custom_field, values)
-                         Rails.logger.error "[MARCELLOGGER] #{coisa.inspect}"
-                         coisa
+                         get_formatted_values(custom_field, values)
                        else
                          [values.first, values.last]
                        end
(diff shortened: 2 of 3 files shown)


app/models/custom_field/hierarchy/item.rb
```diff
@@ -37,4 +37,8 @@ class CustomField::Hierarchy::Item < ApplicationRecord
   scope :including_children, -> { includes(children: :children) }
 
   def to_s = short.nil? ? label : "#{label} (#{short})"
+
+  def ancestry_path
+    self_and_ancestors.filter_map(&:to_s).reverse.join(" / ")
```
Kharonus, line 42: 🟡 Let's extract the gem methods to the persistence layer, so, instead of calling `self_and_ancestors` directly from model, I'd say we introduce a simple method in the service wrapping it, like we did with descendants:

```ruby
      # Gets all descendant nodes in a tree starting from the item/node.
      # @param item [CustomField::Hierarchy::Item] the node
      # @param include_self [Boolean] flag
      # @return [Success(Array<CustomField::Hierarchy::Item>)]
      def get_descendants(item:, include_self: true)
        if include_self
          Success(item.self_and_descendants)
        else
          Success(item.descendants)
        end
      end
```

Kharonus (changes requested):
I'd have a single request. :D otherwise: nice finds, and quick fix. Cool!

app/models/custom_field/hierarchy/item.rb
mereghost: Hmmm... I don't see the advantage in this case. 

We might differ on how we see the `Service`, tho. I see it as a wrapper to hide the implementation details of the tree from the outside code a mix of wrapper and syntactic sugar over it.

This here is no different of asking the label or short or ID of the record.

Kharonus: I'd describe the difference as follows:

The service is the abstraction of the persistence from the domain layer. In this case, you'd be right, finding an item by ID is also a direct access.
(Now the famous) BUT: to be pragmatic, I do not intend to root EVERYTHING through it, but actually everything closure tree related. This way, we could replace the dependency without touching dozens of code files, but just the one. 

WDYT?

mereghost: I agree with you. In this case, we would need to change the actual model anyway so moot point. =)

Kharonus (approved):
True, convincing argument

app/models/custom_field/hierarchy/item.rb
Kharonus: Yeah, eyes opened now.... The model is actually more on persistence layer, sadly :D Makes no sense, using the service in here, as actually the service is using the model and the closure tree methods on it. So, If we would have imports, we would have a circular dependency even.

You are right, I'm wrong

PR #17388, opened 2024-12-06
[#59853] allow editing of hierarchy cfs in wp table

Kharonus:
# Ticket
[OP#59853](https://community.openproject.org/wp/59853)

- enable editing of multi-select custom fields of type hierarchy in the work package table

- revert change to TextDisplayField
- add correct types of custom field values to display field registry service


NobodysNightmare (approved):
Thanks for the thorough explanation. The change makes sense and LGTM.

PR #17400, opened 2024-12-09
[#58160] project stage columns on project list

EinLama:
# Ticket

https://community.openproject.org/wp/58160

All `Project::LifeCycleStepDefinition` instances show up in the **Configure View** modal. For each row/project with an active `Project::LifeCycleStep` configured, its cell will display the date values of the stage or gate fitting that definition.

Please note that as of now, sorting and filtering is not supported for these new columns.

In order to see and test this locally, you must first set up some life cycle step definitions and then populate some projects with the corresponding life cycle steps.

## Screenshots

project_list
Above, you can see that life cycle columns are now visible in the project list. They must be active and assigned to any of the projects. Additionally, they must be added via the configure view modal. Only then will they show up. Cells with no active, associated life cycle step are left empty. The appropriate icon and color for the specific definition is displayed.

configure_view
Above, you can see the columns when some life cycle definitions have been added.

quick_action_table_header
Above, you can see the action menu for the new columns. For now, they cannot be used for sorting or filtering. This will be done in later work packages.

# Merge checklist

- [x] Added/updated tests


app/components/projects/row_component.rb
```diff
@@ -94,6 +96,18 @@ def custom_field_column(column)
       end
     end
 
+    def life_cycle_step_column(column)
+      return nil unless user_can_view_project?
+
+      ls = Project::LifeCycleStep
```
dombesz: This query is a bit inefficient, because it is causing N*M+1 queries, where N is the number of rows and M is the number of columns. (LifeCycleSteps)

I would prefer to preload all the results into a hash in the Projects::TableComponent and access it in the Projects::RowComponent. We do something similar with the `favored_project_ids` here:

https://github.com/opf/openproject/pull/17400/files#diff-063ac5c72e515456a01c09c74b91b6d82a35e4dc5475215252cdd32a79409ac0R32

Basically it could be something like:

```ruby
  delegate :project_life_cycle_steps_by_definition, to: :table

  project_life_cycle_step_by_definition(column.life_cycle, project)
```

```ruby
# Projects::TableComponent

def project_life_cycle_step_by_definition(definition, project)
  @project_life_cycle_steps_by_definition ||=
    Project::LifeCycleStep.where(active: true).group_by{ |s| [s.definition_id, s.project_id]}
  @project_life_cycle_steps_by_definition[definition, project]
end
```

app/models/queries/projects/selects/life_cycle_step.rb
```diff
@@ -0,0 +1,90 @@
...
+  def caption
+    life_cycle.name
+  end
+
+  def life_cycle
+    return @life_cycle if defined?(@life_cycle)
+
+    @life_cycle = Project::LifeCycleStepDefinition
+                    .find_by(id: attribute[KEY, 1])
+  end
```
dombesz: I would suggest calling this method `life_cycle_step_definition`, although it is a bit longer, but there is a distinction between the `life_cycle_step` and the `life_cycle_step_definition`. I'm also just wondering if this declaration has any advantages over using just  a conditional assignment? `||=`

```suggestion
  def life_cycle_step_definition
    @life_cycle_step_definition ||= Project::LifeCycleStepDefinition.find_by(id: attribute[KEY, 1])
  end
```

frontend/src/global_styles/content/_projects_list.sass
```diff
@@ -104,19 +104,3 @@ $content-padding: 10px
     padding-left: 1em
   li
     list-style-type: none
-
```
dombesz, line 107: Nice catch, thank you!

frontend/src/global_styles/content/_table.sass
```diff
@@ -76,6 +76,7 @@ $input-elements: input, 'input.form--text-field', select, 'select.form--select',
     &.ng-leave
       @include animation(0.5s fade-out)
 
+
```
dombesz: There's an extra new line here.

app/models/queries/projects/selects/life_cycle_step.rb
```diff
@@ -0,0 +1,90 @@
...
+
+  def action_menu_classes
+    "leading-visual-icon-header"
+  end
+
+  private
+
+  def helpers
+    @helpers ||= Object.new.extend(ColorsHelper)
```
dombesz, line 88: Is there an advantage of doing this instead of just declaring `include ColorsHelper` in the class?

app/helpers/sort_helper.rb
```diff
@@ -389,35 +392,43 @@ def find_filter_for_column(column, filter_mapping)
...
                           underline: false, display: :inline_flex,
-                          classes: "generic-table--action-menu-button") do |button|
+                          classes: additional_menu_classes) do |button|
       if favorite
         # This column only shows an icon, no text.
         render Primer::Beta::Octicon.new(icon: "star-fill", color: :subtle, "aria-label": I18n.t(:label_favorite))
+      elsif column.respond_to?(:action_menu_header)
+        column.action_menu_header(button)
       else
         button.with_trailing_action_icon(icon: :"triangle-down")
```
dombesz: The last 2 branches of the condition have a lot in common. I think with a little modification on the `column.action_menu_header` one branch can be removed.  As a bonus, we will have the `h(caption).to_s` correctly html escaped.

```suggestion
      else
        button.with_leading_visual_icon(**column.visual_icon) if column.respond_to?(:visual_icon)
        button.with_trailing_action_icon(icon: :"triangle-down")
```
I also added the [accompanying suggestion](https://github.com/opf/openproject/pull/17400#discussion_r1882738982) to the `def action_menu_header(button)`.

app/models/queries/projects/selects/life_cycle_step.rb
```diff
@@ -0,0 +1,90 @@
...
+           end
+
+    classes = helpers.hl_inline_class("life_cycle_step_definition", life_cycle)
+    button.with_leading_visual_icon(icon:, classes:)
+
+    # As all other action menu headers, we will show an action icon and the caption:
+    button.with_trailing_action_icon(icon: :"triangle-down")
+
+    caption.to_s
+  end
```
dombesz: Just a preference, but I would rather keep the button and caption in the sort_helper and just return the `{ icon:, classes: }` from this method. The [sort_helper](https://github.com/opf/openproject/pull/17400#discussion_r1882735110) would also get simpler.

```suggestion
  def visual_icon(button)
    # Show the proper icon for the definition with the correct color.
    icon = case life_cycle
           when Project::StageDefinition
             :"git-commit"
           when Project::GateDefinition
             :diamond
           else
             raise "Unknown life cycle definition for: #{life_cycle}"
           end

    classes = helpers.hl_inline_class("life_cycle_step_definition", life_cycle)
    { icon:, classes: }
  end
```

app/helpers/sort_helper.rb
```diff
@@ -389,35 +392,43 @@ def find_filter_for_column(column, filter_mapping)
...
 
-      move_column_actions(menu, column, table_columns, content_args:, allowed_params:, **html_options)
-      add_and_remove_column_actions(menu, column, table_columns, content_args:, allowed_params:, **html_options)
+      move_column_actions(menu, attribute, table_columns, content_args:, allowed_params:, **html_options)
+      add_and_remove_column_actions(menu, attribute, table_columns, content_args:, allowed_params:, **html_options)
     end
   end
 
-  def action_button(menu, caption, favorite: false)
+  def action_button(menu, column, caption, favorite: false)
```
dombesz, line 418: This method seems to not have any tests, I would suggest adding some test just for the newly introduced functionality. Alternatively we could integraion test that all the icons are showing up correctly in the project headers.

dombesz (changes requested):
The functionality works nicely @EinLama !👏
I just left a few notes, but overall it looks good.

app/models/queries/projects/selects/life_cycle_step.rb
EinLama: The `ColorsHelper` has a lot of methods. When we include a Module, the current class inherits all of these. I found it a bit messy to pollute the namespace with many methods when I'm only interested in one of those (`ColorsHelper#hl_inline_class`).

So, instead of including the module, I figured we could also construct a little helper object that will include the module and keep the namespace clean, so that is what I did. I am not 100 % sure about this, though. It seems messy in its own right.

app/models/queries/projects/selects/life_cycle_step.rb
EinLama: Yes, you are right about the method name. I will adjust it 👍🏻 

About `defined?`, there is actually a slight difference:

1. When we use the `||=` operator here, the right-hand side of the operator (DB query) will be executed as long as the return value is `falsy`. This means when no model can be found and returned, the query would still be executed every time this method is called.
2. When using the `defined?` approach, the DB query is executed exactly once. Even if it returns `nil`, that value is saved in the instance variable and will be returned in future method calls.

I think in this use case it is preferable to not query the DB needlessly, so I went with `defined?`. I usually prefer `||=` because of its elegance, but here it makes sense to invest the extra syntax for `defined?`. What do you think?

frontend/src/global_styles/content/_table.sass
EinLama: Thank you, fixed! :)

app/components/projects/row_component.rb
EinLama: Amazing suggestion, thank you 👍🏻

app/models/queries/projects/selects/life_cycle_step.rb
EinLama: That's neat! It bothered me that the Select got too involved in the UI. It's much cleaner like this. Nice!

app/helpers/sort_helper.rb
EinLama: Good catch, I provided some specs 👍🏻

app/models/queries/projects/selects/life_cycle_step.rb
dombesz: I don't have a strong opinion about this, I just find this method a bit unusual.

app/models/queries/projects/selects/life_cycle_step.rb
dombesz: Thanks for the thorough explanation, yes it makes sense to do that and this is a very good caution agains using `||=` all the time.

app/components/projects/table_component.rb
```diff
@@ -195,6 +195,14 @@ def favored_project_ids
       @favored_project_ids ||= Favorite.where(user: current_user, favored_type: "Project").pluck(:favored_id)
     end
 
+    def project_life_cycle_step_by_definition(definition, project)
+      @project_life_cycle_steps_by_definition ||= Project::LifeCycleStep
+                                                    .where(active: true)
+                                                    .group_by { |s| [s.definition_id, s.project_id] }
+
+      @project_life_cycle_steps_by_definition[[definition.id, project.id]]&.first
```
dombesz: Sorry, I forgot that we have only one LifeCycleStep per Project and per Definition. In this case, we can use `index_by` instead of `group_by`. `index_by` just takes the latest value for the key and assigns that, instead of aggregating them into an array. That way the accessing looks a bit less ugly.

```suggestion
                                                    .index_by { |s| [s.definition_id, s.project_id] }

      @project_life_cycle_steps_by_definition[[definition.id, project.id]]
```

dombesz (approved):
Good job @EinLama , thank you! 👏
I just made a minor comment on  a suggestion on using `index_by` instead of `group_by`.

dombesz (changes requested):
Oops, I just realised that we have to put the whole implementation behind the feature flag `:stages_and_gates`.
We cannot show the stages and gates as selectable fields in the table configuration, nor show the columns.

EinLama:
Thank you for pointing out the necessity for the feature flag. I thought the new columns would be hidden by them being set to inactive by default, but that wasn't true :)

dombesz (approved):
Well done @EinLama ! 👏

PR #17403, opened 2024-12-09
[59468] Removed embedded file links colelction from work package resource

apfohl:
# Ticket
[59468](https://community.openproject.org/projects/document-workflows-stream/work_packages/59468/activity)

Remove embedded resource `FileLinkCollection` from work package resource.

### DIFF
--- a/spec/requests/api/v3/work_packages/create_resource_spec.rb
+++ b/spec/requests/api/v3/work_packages/create_resource_spec.rb
@@ -113,7 +113,7 @@ RSpec.describe "API v3 Work package resource",
     end
 
     it "creates a work package" do
-      expect(WorkPackage.all.count).to eq(1)
+      expect(WorkPackage.count).to eq(1)
     end
 
     it "uses the given parameters" do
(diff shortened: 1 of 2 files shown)


spec/requests/api/v3/work_packages/create_resource_spec.rb
```diff
@@ -310,7 +310,7 @@
           expect(last_response.body).to be_json_eql(
             api_v3_paths.file_links(work_package.id).to_json
           ).at_path("_links/fileLinks/href")
-          expect(last_response.body).to be_json_eql("1").at_path("_embedded/fileLinks/total")
+          expect(last_response.body).not_to have_json_path("_embedded/fileLinks")
```
Kharonus: 🟡 I'd rather remove this line, instead of checking for absence. As we no longer CAN embed it, why should we check for it to not be embedded?

Also, looking at the test here, should we maybe add an expectation, that the link `addFileLinks` is available? Equivalent to what we do with the link for the collection in 310.

Kharonus (changes requested):

spec/requests/api/v3/work_packages/create_resource_spec.rb
apfohl: It's tricky to test against something, not to be added in the future. But makes sense.

Kharonus (approved):
LGTM

PR #17412, opened 2024-12-09
attempt to stabilize flickering wp ancestor spec

ulferts:
Stabilize flickering

`rspec ./spec/features/work_packages/update_ancestors_spec.rb:174`

This is done by specifying an order for the query. Before that, it might have happend that 'second child' which first is outdented ends up at the top of the table. When the spec then tries to indent it again, it can't since that option is only available in case there is a work package above the manipulated entry.

## Screenshots

The screenshot shows how second child is the first element in the table. Taken from a failing test run.

Dec 9 Screenshot from OpenProject


cbliard (approved):
Sounds good 👏

PR #17415, opened 2024-12-10
Using I18n save text

Eric-Guo:

Add missing I18n in the button.

### Before

image

### After

image

I don't see other approach available here.

# Merge checklist


oliverguenther (approved):
Thanks for the fix 👍

PR #17420, opened 2024-12-10
Add docker-compose.yml for nextcloud development

NobodysNightmare:
This is similar to what we already have for Keycloak and GitLab in providing a minimal docker-compose file for integration work.

# Ticket
https://community.openproject.org/projects/document-workflows-stream/work_packages/59937/activity

### DIFF
--- a/docker/dev/nextcloud/README.md
+++ b/docker/dev/nextcloud/README.md
@@ -1,0 +1,15 @@
+A minimal setup to run a Nextcloud inside the TLS-enabled docker development stack.
+
+# First installation steps
+
+1. Allow accessing OP through `openproject.local`:
+    * `docker compose exec --user www-data nextcloud php occ config:system:set allow_local_remote_servers --value 1`
+2. Import Dev CA cert into Nextcloud's own certificate store:
+    * `docker compose cp /path/to/your/OpenProject_Development_Root_CA.crt nextcloud:/tmp/root.crt`
+    * `docker compose exec nextcloud chown www-data /tmp/root.crt`
+    * `docker compose exec --user www-data nextcloud php occ security:certificates:import /tmp/root.crt`
+3. Following [docs](https://docs.nextcloud.com/server/30/admin_manual/configuration_server/reverse_proxy_configuration.html#defining-trusted-proxies) configure Traefik as trusted proxy
+    * e.g. add `'trusted_proxies' => ['172.0.0.0/8'],` to `config/config.php` for a pretty broad allowance for most docker services (verify that your `gateway` network uses IPs in the given address range)
+4. Download/Activate Nextcloud plugins:
+    * `integration_openproject`
+    * `groupfolders` (dependency for certain OP functions)
--- a/docker/dev/nextcloud/docker-compose.yml
+++ b/docker/dev/nextcloud/docker-compose.yml
@@ -1,0 +1,35 @@
+services:
+  nextcloud:
+    image: nextcloud:stable
+    restart: unless-stopped
+    networks:
+      - gateway
+    volumes:
+      - nextcloud:/var/www/html
+      #   If you want to use a local version of a plugin, mount a local folder
+      #   more detailed instructions available at https://github.com/nextcloud/integration_openproject/blob/master/docs/setup_via_docker.md
+      # - ../nextcloud_apps:/var/www/html/custom_apps
+    labels:
+      - "traefik.enable=true"
+      - "traefik.http.routers.nextcloud.rule=Host(`nextcloud.local`)"
+      - "traefik.http.routers.nextcloud.entrypoints=websecure"
+
+  cron:
+    image: nextcloud:stable
+    restart: unless-stopped
+    volumes:
+      - nextcloud:/var/www/html
+      #   If you want to use a local version of a plugin, mount a local folder
+      #   more detailed instructions available at https://github.com/nextcloud/integration_openproject/blob/master/docs/setup_via_docker.md
+      # - ../nextcloud_apps:/var/www/html/custom_apps
+    networks:
+      - gateway
+    entrypoint: /cron.sh
+
+networks:
+  gateway:
+    external: true
+    name: gateway
+
+volumes:
+  nextcloud:


docker/dev/nextcloud/README.md
```diff
@@ -0,0 +1,15 @@
...
+# First installation steps
+
+1. Allow accessing OP through `openproject.local`:
+    * `docker compose exec --user www-data nextcloud php occ config:system:set allow_local_remote_servers --value 1`
+2. Import Dev CA cert into Nextcloud's own certificate store:
+    * `docker compose cp /path/to/your/OpenProject_Development_Root_CA.crt nextcloud:/tmp/root.crt`
+    * `docker compose exec nextcloud chown www-data /tmp/root.crt`
+    * `docker compose exec --user www-data nextcloud php occ security:certificates:import /tmp/root.crt`
+3. Following [docs](https://docs.nextcloud.com/server/30/admin_manual/configuration_server/reverse_proxy_configuration.html#defining-trusted-proxies) configure Traefik as trusted proxy
+    * e.g. add `'trusted_proxies' => ['172.0.0.0/8'],` to `config/config.php` for a pretty broad allowance for most docker services (verify that your `gateway` network uses IPs in the given address range)
```
NobodysNightmare, line 12: I would have preferred to add a magic `sed` call to be used for this, but I am lacking sed expertise to make this work I believe.

If anyone wants to chime in, feel free.

docker/dev/nextcloud/docker-compose.yml
```diff
@@ -0,0 +1,23 @@
+services:
+  nextcloud:
```
brunopagno, line 2: I needed to include the `cron` override on mine as well

```
  cron:
    volumes:
      - ${APP_DIR:-./../../custom_apps}:/var/www/html/custom_apps
    networks:
      - external
```

docker/dev/nextcloud/docker-compose.yml
```diff
@@ -0,0 +1,23 @@
...
+      - gateway
+    volumes:
+      - nextcloud:/var/www/html
+      #   If you want to use a local version of a plugin, mount a local folder
+      #   more detailed instructions available at https://github.com/nextcloud/integration_openproject/blob/master/docs/setup_via_docker.md
+      # - ../nextcloud_apps:/var/www/html/custom_apps
+    labels:
+      - "traefik.enable=true"
+      - "traefik.http.routers.nextcloud.rule=Host(`nextcloud.local`)"
+      - "traefik.http.routers.nextcloud.entrypoints=websecure"
```
brunopagno, line 15: There's also an option to using TLS here using

```
      - "traefik.http.routers.nextcloud.tls=true"
      - "traefik.http.routers.nextcloud.tls.certresolver=step"
```

But then we need to include and env variable

```
    environment:
      - OVERWRITEPROTOCOL=https
```

brunopagno (changes requested):
I find this briliant. It will change lives :tada: 

Only thing I believe we should change is the filename from `docker-compose.yml` to `docker-compose.override.yml`.

The rest are a few observations based on my own setup, but it's okay if they are not included.

docker/dev/nextcloud/docker-compose.yml
NobodysNightmare: Which purpose do these labels serve? Coming from the TLS configuration, I'd expect that using the `entrypoints=websecure` already doesn everything we need. For sure my nextcloud is served through HTTPS and step is providing the certificate for that with the config given in this PR.

As far as I can tell the settings you mention are set for the entrypoint already: https://github.com/opf/openproject/blob/dev/docker/dev/tls/docker-compose.yml#L14-L15

>  environment:
>    - OVERWRITEPROTOCOL=https

This is not necessary when the step about trusting Traefik has been done, because then NCs Apache will trust the `X-Forwarded-Proto` header provided by Traefik, which indicates the currently used protocol.

However, if you find a way to overwrite the trusted proxy with an env var, I'd be more than happy.

docker/dev/nextcloud/docker-compose.yml
NobodysNightmare: Can you post your full `cron` service definition? I only have the `nextcloud` service running locally for me.

NobodysNightmare:
> I find this briliant. It will change lives 🎉
> 
> Only thing I believe we should change is the filename from `docker-compose.yml` to `docker-compose.override.yml`.
> 
> The rest are a few observations based on my own setup, but it's okay if they are not included.

I am calling this `docker-compose.yml` because it's not an override file, but a base file. Otherwise: What's the base file we'd be overriding? I would want to avoid providing an override for a file that's not part of _this_ repository, because then the original file could change without us noticing and thus the override would override wrongly.

docker/dev/nextcloud/docker-compose.yml
brunopagno: I stopped tweaking the file when I got it working so I did not get to fine details. But fair point about the TLS. I assumed the wrong thing.

About the OVERWRITEPROTOCOL, I might have missed a previous step then. Cause I could only actually "upload" after adding it, otherwise the requests would always go to openproject.local by http. In any case it's all good.

brunopagno:
> What's the base file we'd be overriding? I would want to avoid providing an override for a file that's not part of this repository, because then the original file could change without us noticing and thus the override would override wrongly.

I guess that makes sense. It's not necessarily related to the [openproject-nextcloud-integration](https://github.com/nextcloud/integration_openproject), is it? I wonder if it would become more confusing in the future. But I see the last steps of the readme are already installing the whole thing.

brunopagno (approved):

docker/dev/nextcloud/docker-compose.yml
brunopagno: I was referring to the [cron definition in the integration project](https://github.com/nextcloud/integration_openproject/blob/master/docker-compose.yml#L35-L45).

NobodysNightmare: I see. Makes sense, if we need cron, we should add it to this compose file. Do you know if cron is needed for our integration purposes?

brunopagno: I do not know. [Briefly reading does not sound mandatory](https://docs.nextcloud.com/server/latest/admin_manual/configuration_server/background_jobs_configuration.html). But probably useful to have.

NobodysNightmare: I added a cron service as well now. Looking at my local installation, it seems to be working.

PR #17424, opened 2024-12-10
Implementation/59970 introduce enterprise banner viewcomponent

ulferts:
# Ticket

https://community.openproject.org/wp/59970

Introduce a ViewComponent to be used when rendering a banner. It will be used first within the project lifecycle administration but should ideally be used throughout the application eventually.

## Screenshots

image

The component enforces i18n and link href to be at a standard location so that it is easier to find the appropriate place for this kind of information.

# Merge checklist

- [x] Added/updated tests

### DIFF
--- a/app/components/enterprise_edition/banner_component.html.erb
+++ b/app/components/enterprise_edition/banner_component.html.erb
@@ -1,0 +1,19 @@
+<%=
+  grid_layout("op-ee-banner", **@system_arguments) do |grid|
+    grid.with_area(:'icon-container') do
+      content_tag :div, class: "op-ee-banner--shield" do
+        render(Primer::Beta::Octicon.new(icon: 'op-enterprise-addons',
+                                         size: :medium,
+                                         classes: "op-ee-banner--icon"))
+      end
+    end
+    grid.with_area(:'title-container') { render(Primer::Beta::Text.new) { title } }
+    grid.with_area(:'description-container') { render(Primer::Beta::Text.new) { description } }
+    grid.with_area(:'link-container') do
+      render(Primer::Beta::Link.new(href: href)) do |link|
+        link.with_trailing_visual_icon(icon: 'link-external')
+        link_title
+      end
+    end
+  end
+%>
--- a/app/components/enterprise_edition/banner_component.rb
+++ b/app/components/enterprise_edition/banner_component.rb
@@ -1,0 +1,107 @@
+# frozen_string_literal: true
+
+# -- copyright
+# OpenProject is an open source project management software.
+# Copyright (C) 2024 the OpenProject GmbH
+#
+# This program is free software; you can redistribute it and/or
+# modify it under the terms of the GNU General Public License version 3.
+#
+# OpenProject is a fork of ChiliProject, which is a fork of Redmine. The copyright follows:
+# Copyright (C) 2006-2013 Jean-Philippe Lang
+# Copyright (C) 2010-2013 the ChiliProject Team
+#
+# This program is free software; you can redistribute it and/or
+# modify it under the terms of the GNU General Public License
+# as published by the Free Software Foundation; either version 2
+# of the License, or (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program; if not, write to the Free Software
+# Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+#
+# See COPYRIGHT and LICENSE files for more details.
+# ++
+
+module EnterpriseEdition
+  # Add a general description of component here
+  # Add additional usage considerations or best practices that may aid the user to use the component correctly.
+  # @accessibility Add any accessibility considerations
+  class BannerComponent < ApplicationComponent
+    include OpPrimer::ComponentHelpers
+
+    # @param feature_key [Symbol, NilClass] The key of the feature to show the banner for.
+    # @param title [String] The title of the banner.
+    # @param description [String] The description of the banner.
+    # @param href [String] The URL to link to.
+    # @param skip_render [Boolean] Whether to skip rendering the banner.
+    # @param system_arguments [Hash] <%= link_to_system_arguments_docs %>
+    def initialize(feature_key,
+                   title: nil,
+                   description: nil,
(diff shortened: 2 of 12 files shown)


app/components/enterprise_edition/banner_component.html.erb
```diff
@@ -0,0 +1,19 @@
+<%=
+  render(Primer::OpenProject::GridLayout.new(css_class: "op-ee-banner", **@system_arguments)) do |grid|
```
HDinger: You can use the `grid_layout` helper for this.

app/components/tables/column_header_component.rb
```diff
@@ -0,0 +1,131 @@
...
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program; if not, write to the Free Software
+# Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+#
+# See COPYRIGHT and LICENSE files for more details.
+# ++
+
+class Tables::ColumnHeaderComponent < ApplicationComponent
```
HDinger: Should that file be part of the PR? 🤔

app/components/enterprise_edition/banner_component.rb
```diff
@@ -0,0 +1,93 @@
...
+
+    attr_reader :skip_render,
+                :feature_key
+
+    def title
+      @title || I18n.t("ee.upsale.#{feature_key}.title", default: I18n.t("ee.upsale.title"))
+    end
+
+    def description
+      @description || I18n.t("ee.upsale.#{feature_key}.description")
```
HDinger: There should be a default for the description I guess, similar to how title and link_title have one.

app/components/enterprise_edition/banner_component.sass
```diff
@@ -0,0 +1,65 @@
...
+    height: 40px
+    border-width: 10px 5px 10px 5px
+    border-radius: 0 0 10px 10px
+    border-style: solid
+
+  &--icon
+    width: 32px
+    height: 32px
+    margin-top: -6px
+    margin-left: -5px
```
HDinger: This is somehow bad style, as these numbers are completly random and will break as soon as we change the layout of the shield in any way. You could get rid of that by using a flexBox on the shield with `align_items: center` and `justify_content: center`

lookbook/previews/open_project/enterprise_edition/banner_component_preview.rb
```diff
@@ -0,0 +1,79 @@
...
+      # ```
+      #
+      # The href is inferred from `OpenProject::Static::Links.enterprise_docs[feature_key][:href]`.
+      #
+      # The value of `EnterpriseToken.show_banners?` is used to determine whether the banner should be shown. For this
+      # example, that value is overwritten as the banner might otherwise not show up in the preview.
+      def default
+        render(
+          ::EnterpriseEdition::BannerComponent
+            .new(:form_configuration,
```
HDinger, line 60: `form_configuration` is maybe a bad example key, as the content expects a list afterwards and it looks to have a colon but nothing coming afterwards.
Bildschirmfoto 2024-12-11 um 13 14 24

app/components/enterprise_edition/banner_component.sass
```diff
@@ -0,0 +1,65 @@
...
+    grid-template-areas: "icon-container title-container title-container" "icon-container description-container link-container"
+
+  &--icon-container
+    @extend .upsale-colored
+    align-self: start
+    justify-self: center
+
+  &--shield
+    @extend .upsale-border-colored
+    width: 32px
```
HDinger: Since that width is used three times in that file (and it is important that it is always the same value), I'd like to extract that into a variable

HDinger (changes requested):
I like the idea of using a key and thus enforcing the same structure of localisation for each feature. This should also make it easier to use. 👍 
I am wondering however, whether we should allow it at all to overwrite the key-based strings, but probably there are use cases where that makes sense.. 🤷‍♀️ 

One minor thing: It looks like the text is slighly not centred compared to the shield. This is due to the shield being smaller than the grid next to it. 
Bildschirmfoto 2024-12-11 um 13 39 47

app/components/tables/column_header_component.rb
ulferts: Sorry for that. You are right of course, that this found its way in here by mistake.

app/components/enterprise_edition/banner_component.html.erb
ulferts: Thanks for the pointer

app/components/enterprise_edition/banner_component.sass
ulferts: Done, thanks.

app/components/enterprise_edition/banner_component.sass
ulferts: :+1:

lookbook/previews/open_project/enterprise_edition/banner_component_preview.rb
ulferts: Yes, that looks bad. I wanted to replace it with a different one. But the one that was available didn't have an href set so I failed first. Now I find out that this alternative is also outdated so I removed it in the PR.

Ideally, it will be replaced by the first feature that actually uses this component, @toy. So when the component is applied to the project lifecycle administration, we can use that as the example.

app/components/enterprise_edition/banner_component.rb
ulferts: I believe that while title and link text will more or less always stay the same, the description needs to be different for every feature we apply it to. 

To manifest that believe, the component now raises an error with a pointer on how to fix this.

ulferts:
I believe to have addressed all your remarks, @HDinger .

Since the PR included an erroneously added file and the PR in total isn't big, I force pushed. I hope that does not make the rereview too strenuous.

> I am wondering however, whether we should allow it at all to overwrite the key-based strings, but probably there are use cases where that makes sense.. 🤷‍♀️

We'll see. I hope that it is not used but it allowed to better show the component in lookbook.

HDinger (approved):

app/components/enterprise_edition/banner_component.rb
```diff
@@ -0,0 +1,107 @@
...
+
+    def title
+      @title || I18n.t("ee.upsale.#{feature_key}.title", default: I18n.t("ee.upsale.title"))
+    end
+
+    def description
+      @description || begin
+        I18n.t("ee.upsale.#{feature_key}.description")
+      rescue StandardError
+        I18n.t("ee.upsale.#{feature_key}.description_html")
```
toy, line 76: @ulferts Why rescue instead of falling back using default as it is done for title?

PR #17425, opened 2024-12-10
Make keycloak setup more consistent

NobodysNightmare:
Giving operator control of restart by using unless-stopped. Previously it was in two extremes at the same time: The db would always restart, even if stopped, while the app server would never restart, even if it was running.

Also opting into a feature that's necessary to properly configure token exchange according to Keycloak docs at https://www.keycloak.org/securing-apps/token-exchange (While the docs say it's not needed for the internal to internal flow, I literally couldn't follow the instructions without this feature).

# Ticket
none

Make the default setup for Keycloak more consistent.

### DIFF
--- a/docker/dev/keycloak/docker-compose.yml
+++ b/docker/dev/keycloak/docker-compose.yml
@@ -3,7 +3,7 @@ version: "3.9"
 services:
   db-keycloak:
     image: postgres:13
-    restart: always
+    restart: unless-stopped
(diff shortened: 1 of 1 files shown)


Kharonus (approved):
I see no reason not to change those configs. But, alas, I was not the one writing those in the beginning. xD

But, if you are optimistic, that it improves the setup, :shipit:

PR #17430, opened 2024-12-11
[60025] Wrong locale of error message while token creation

top4ek:
[Ticket](https://community.openproject.org/projects/openproject/work_packages/60025/activity)

### DIFF
--- a/app/contracts/api_tokens/create_contract.rb
+++ b/app/contracts/api_tokens/create_contract.rb
@@ -32,7 +32,7 @@ module APITokens
   class CreateContract < BaseContract
     attribute :token_name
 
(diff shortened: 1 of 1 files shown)


NobodysNightmare:
FYI: I am going to try and reproduce this locally quickly, also to validate that passing a proc to a message is possible (I haven't seen/done this before).

I also want to check whether this then affects other places, where we call `I18n` on the class level, because I don't see why it shouldn't. E.g. [here](https://github.com/opf/openproject/blob/dev/app/models/queries/projects/selects/custom_field.rb#L30)

top4ek:
Agree, it's good idea to check other models for that.

NobodysNightmare (approved):
Thanks a lot for your contribution. The PR works as advertised. I am going to address other cases in a follow-up PR.

PR #17431, opened 2024-12-11
attempt to stabilize flickering relations spec

ulferts:
The spec used to check for the non existence of an .ng-option element with the text being the id of the current work package. Depending on the order the specs are run in it could happen that another work package ended up having that number as part of its subject. In such a case, that work package was returned.
Looking for the subject avoids the unspecificity.

## Screenshots

This screenshot shows the case where the searched for id is matching parts of the subject of the other work package. 

image


bsatarnejad (approved):
👍

PR #17434, opened 2024-12-11
[#59985] add validation check to custom field controller

Kharonus:
# Ticket
[OP#59985](https://community.openproject.org/work_packages/59985)

- do not allow creation of custom fields of type hierarchy

- add check to controller, as there is no contract for custom fields

### DIFF
--- a/app/controllers/custom_fields_controller.rb
+++ b/app/controllers/custom_fields_controller.rb
@@ -30,14 +30,17 @@ class CustomFieldsController < ApplicationController
   include CustomFields::SharedActions # share logic with ProjectCustomFieldsControlller
   layout "admin"
 
+  # rubocop:disable Rails/LexicallyScopedActionFilter
   before_action :require_admin
   before_action :find_custom_field, only: %i(edit update destroy delete_option reorder_alphabetical)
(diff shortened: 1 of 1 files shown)


app/controllers/custom_fields_controller.rb
```diff
@@ -30,14 +30,17 @@ class CustomFieldsController < ApplicationController
...
   before_action :find_custom_field, only: %i(edit update destroy delete_option reorder_alphabetical)
   before_action :prepare_custom_option_position, only: %i(update create)
   before_action :find_custom_option, only: :delete_option
+  before_action :validate_enterprise_token, only: %i(create)
+  # rubocop:enable Rails/LexicallyScopedActionFilter
 
   def index
     # loading wp cfs exclicity to allow for eager loading
-    @custom_fields_by_type = CustomField.all
+    @custom_fields_by_type = CustomField
```
brunopagno, line 43: Nice catch :clap:

brunopagno (approved):
lookin pretty good :tada:

app/controllers/custom_fields_controller.rb
Kharonus: rubocop did that for me ^^

PR #17435, opened 2024-12-11
Changing the where into a find to avoid random weird ordering

mereghost:
We need to enforce a certain order on the returned items from the Plain formatter

### DIFF
--- a/lib/open_project/journal_formatter/custom_field/plain.rb
+++ b/lib/open_project/journal_formatter/custom_field/plain.rb
@@ -96,7 +96,7 @@ class OpenProject::JournalFormatter::CustomField::Plain < JournalFormatter::Base
     items = CustomField::Hierarchy::Item.where(id: ids).index_by(&:id)
 
     ids.map do |id|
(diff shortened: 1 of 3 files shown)


lib/open_project/journal_formatter/custom_field/plain.rb
```diff
@@ -93,10 +93,12 @@ def find_user_value(value, _custom_field)
   def find_item_value(value, _custom_field)
     ids = value.split(",").map(&:to_i)
 
-    CustomField::Hierarchy::Item.where(id: ids).map do |item|
-      next I18n.t(:label_deleted_custom_option) unless ids.include?(item.id)
+    items = CustomField::Hierarchy::Item.where(id: ids).index_by(&:id)
 
-      item.ancestry_path
+    ids.map do |id|
+      next I18n.t(:label_deleted_custom_option) unless items[id]
```
Kharonus: Can we have a new key? as `option` is not the correct domain term for this. 

"(deleted item)"

Kharonus (changes requested):

Kharonus (approved):
LGTM

PR #17438, opened 2024-12-11
Fix docker testing documentation

NobodysNightmare:
Tests can't be executed using

    docker compose run --rm backend-test

anymore, instead the backend-test service is intended to be started and then tests need to be `exec`ed inside that container. It's also described like that below.

Provide documentation that allows to run tests after following it ;-)

### DIFF
--- a/docs/development/development-environment/docker/README.md
+++ b/docs/development/development-environment/docker/README.md
@@ -45,7 +45,7 @@ docker compose up -d frontend
 Optional: In case you want to develop on the OpenProject *BIM Edition* you need
 to install all the required dependencies and command line tools to convert IFC
 files into XKT files, so that the BIM models can be viewed via the *Xeokit*
-BIM viewer. As the conversions are done by background jobs you need install 
+BIM viewer. As the conversions are done by background jobs you need install
 those tools within the `worker` service:
 
 ```shell
@@ -61,11 +61,14 @@ Once the containers are done booting you can access the application under `http:
 You can run tests inside the `backend-test` container. You can run specific tests, too.
 
 ```shell
+# Required once before running tests
+docker compose up -d backend-test
+
 # Run all tests (not recommended)
-docker compose run --rm backend-test bundle exec rspec
+docker compose exec backend-test bundle exec rspec
 
 # Run the specified test
-docker compose run --rm backend-test bundle exec rspec spec/features/work_package_show_spec.rb
+docker compose exec backend-test bundle exec rspec spec/features/work_package_show_spec.rb
 ```
 
 ***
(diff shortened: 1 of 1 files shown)


docs/development/development-environment/docker/README.md
```diff
@@ -384,15 +387,15 @@ TLS support. Basically, you need to tell `traefik` for which docker compose serv
...
+In addition, we need to alter the environmental variables used in the new overrides. So we need to amend the `.env.development.local` file
 like that:
 
 ```shell
 OPENPROJECT_DEV_HOST=openproject.local
 OPENPROJECT_DEV_URL=https://${OPENPROJECT_DEV_HOST}
 ```
 
-After amending the override file and the `.env`, ensure that you restart the stack.
+After amending the override file and the `.env.development.local`, ensure that you restart the stack.
```
NobodysNightmare: TBD: I just noticed that my local web worker setup is borked. Not sure if my changes here are the cause.

NobodysNightmare: I think my suggestion to move this stuff to `env.development.local` is entirely wrong. The problem I had was that I overwrite `OPENPROJECT_HOST_NAME` as well, which will then break tests.

However, I can't find any instructions to do that, so I am wondering why I did it in the first place. As far as I can tell, by just following the original instructions here, both tests and web server should work fine.

Kharonus:
Hmm, why?

What is wrong with `run`? No one should be forced to `exec` on an upped service?

NobodysNightmare:
The important thing is

> Tests can't be executed using

I.e. if you try, it doesn't work. Let me show you:

```
$ docker compose run --rm backend-test bundle exec rspec
[+] Creating 7/0
 ✔ Container openproject-chrome-1          Running                                       0.0s 
 ✔ Container openproject-firefox-1         Running                                       0.0s 
 ✔ Container openproject-frontend-test-1   Ru...                                         0.0s 
 ✔ Container openproject-cuprite-chrome-1  R...                                          0.0s 
 ✔ Container openproject-db-test-1         Running                                       0.0s 
 ✔ Container openproject-opera-1           Running                                       0.0s 
 ✔ Container selenium-hub                  Running                                       0.0s 
Waiting for the Grid...
Bundle complete! 209 Gemfile dependencies, 416 gems now installed.
Use `bundle info [gemname]` to see where a bundled gem is installed.
8 installed gems you directly depend on are looking for funding.
  Run `bundle fund` for details
$
```

Does using `run` work for you at all? Maybe my setup is borked, wouldn't be the first time :D

brunopagno:
@NobodysNightmare can you try the same command with quotes?

`docker compose run --rm backend-test 'bundle exec rspec'`

I've noticed that some docker commands require us to pass the whole command as a string, and others explicitly cannot be passed in quotes.

But I have no problem using `docker run --rm` on my tests

NobodysNightmare:
Thanks Bruno. That indeed works. I don't understand why it's necessary to pass the full command as a single argument, but whatever.

This means nothing really remains of this pull request, going to close it.

PR #17439, opened 2024-12-11
Adds support of storing patterns on Type

mereghost:
## What?

Introduces the `patterns` attribute to `Type`.

## How?

I created a new text field, with a custom ActiveRecord type on Type, to represent the patterns to be applied on the certain fields of the associated WPs.

### DIFF
--- a/app/models/type.rb
+++ b/app/models/type.rb
@@ -34,6 +34,8 @@ class Type < ApplicationRecord
 
   include ::Scopes::Scoped
 
+  attribute :patterns, Types::PatternCollectionType.new
+
   before_destroy :check_integrity
 
   has_many :work_packages
@@ -50,16 +52,11 @@ class Type < ApplicationRecord
                           join_table: "#{table_name_prefix}custom_fields_types#{table_name_suffix}",
                           association_foreign_key: "custom_field_id"
 
-  belongs_to :color,
-             optional: true,
-             class_name: "Color"
+  belongs_to :color, optional: true, class_name: "Color"
 
   acts_as_list
 
-  validates :name,
-            presence: true,
-            uniqueness: { case_sensitive: false },
-            length: { maximum: 255 }
+  validates :name, presence: true, uniqueness: { case_sensitive: false }, length: { maximum: 255 }
 
   validates :is_default, :is_milestone, inclusion: { in: [true, false] }
 
@@ -67,10 +64,7 @@ class Type < ApplicationRecord
 
   default_scope { order("position ASC") }
 
-  scope :without_standard, -> {
-    where(is_standard: false)
-      .order(:position)
-  }
+  scope :without_standard, -> { where(is_standard: false).order(:position) }
 
   def to_s; name end
 
--- a/app/models/types/pattern.rb
+++ b/app/models/types/pattern.rb
@@ -1,0 +1,42 @@
+# frozen_string_literal: true
+
+#-- copyright
+# OpenProject is an open source project management software.
+# Copyright (C) the OpenProject GmbH
+#
+# This program is free software; you can redistribute it and/or
+# modify it under the terms of the GNU General Public License version 3.
+#
+# OpenProject is a fork of ChiliProject, which is a fork of Redmine. The copyright follows:
+# Copyright (C) 2006-2013 Jean-Philippe Lang
+# Copyright (C) 2010-2013 the ChiliProject Team
+#
+# This program is free software; you can redistribute it and/or
+# modify it under the terms of the GNU General Public License
+# as published by the Free Software Foundation; either version 2
+# of the License, or (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program; if not, write to the Free Software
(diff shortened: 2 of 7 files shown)


app/models/types/pattern.rb
```diff
@@ -0,0 +1,12 @@
+# frozen_string_literal: true
+
+#-- copyright
+#++
+
+module Types
+  Pattern = Data.define(:pattern) do
```
brunopagno: Naming questions because I'm that guy:

1. Pattern for the type name sounds okay to me, but I ask myself if it's specific enough. I don't want to call it `SubjectPattern`, but maybe there's a universe where we can call `NamePattern` or `AttributePattern`.
2. We might going to have situations where we say `pattern.pattern`. Should we name the attribute `value` or something generic?

mereghost: 1. I suck at naming, so feedback and ideas are always welcome. `SomethingPattern` sounds great. Pinpointing the `Something` is the trick. =/
2. My idea is to never have to call the `#pattern` directly, maybe even make it private. It would eventually have a method called `#apply` or `resolve` or something else (see point 1) that just runs the pattern somehow and returns the resulting string.

brunopagno: Fair enough. So, about the name of the type first. The something in `SomethingPattern` could be:
- **Attribute**: because we are assigning attributes to a pattern. But not sure if it's meaningful enough
- **WorkPackageNamePattern**: more specific, but probably too enterprisified to be meaningful.

Could not think of anything else now. I'll be happy if we stay with Pattern as well, just trying to bounce random ideas.

brunopagno: About the attribute ` :pattern` inside the type `Pattern`. Absolutely fine. I don't like calling `thing.thing`, but at the same time it's not a big deal.

mereghost: @brunopagno one good thing is that those objects are created by the PatternType, so we can make changes to their internals at will, even after they get created.

brunopagno: That works well for me <3

app/models/types/pattern_collection.rb
```diff
@@ -0,0 +1,53 @@
...
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program; if not, write to the Free Software
+# Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+#
+# See COPYRIGHT and LICENSE files for more details.
+#++
+
+module Types
```
Kharonus, line 31: 🟡 ouf, indeed, this wording ambiguity sucks....

Looks like some module which is collecting types for everything, not only work package types

app/models/type.rb
```diff
@@ -50,27 +52,19 @@ def copy_from_type(source_type)
...
-             class_name: "Color"
+  belongs_to :color, optional: true, class_name: "Color"
 
   acts_as_list
 
-  validates :name,
-            presence: true,
-            uniqueness: { case_sensitive: false },
-            length: { maximum: 255 }
+  validates :name, presence: true, uniqueness: { case_sensitive: false }, length: { maximum: 255 }
```
Kharonus, line 59: ❓ seeing this, does it mean, we have no work package type contract? We have to do the validations on the model creation?

mereghost: Gooooood Question. I have no idea, I'll check the code and get back to you

app/models/types/pattern_collection.rb
mereghost: Yes. =/ It was around already, but we should raise the discussion of renaming the model to `WorkPackageType` or something.

app/models/type.rb
mereghost: @Kharonus that's all we got: 

[Types::BaseContract](https://github.com/opf/openproject/blob/4911b8a14921efe0c552ab2029389ac12bcaefcb/app/contracts/types/base_contract.rb) is used on [BaseTypeService](https://github.com/opf/openproject/blob/247254495d059709e74ded6966976e3d9929c6b2/app/services/base_type_service.rb) and is the parent of `CreateTypeService` and `UpdateTypeService`

Kharonus: Ah, yeah, I checked the class, and I'd vote to move validation in there. Your call, if you want to do this with the next step (of pattern string validation itself).

Kharonus (approved):
Like written in the comment, I generally approve the changes.

The work of moving the validation to the contract is open, but merging this code to `dev` can enable the next steps on the implementation chunks (namely, connecting to UI via a controller).

app/models/type.rb
mereghost: Oh, this validation was already there, I just re-formatted the line. XD

Kharonus: true, then it is even less a reason to do this in this PR, but maybe worth a though, when adding pattern validation.

db/migrate/20241211152749_introduce_patterns_to_types.rb
```diff
@@ -0,0 +1,37 @@
...
+# along with this program; if not, write to the Free Software
+# Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+#
+# See COPYRIGHT and LICENSE files for more details.
+#++
+
+class IntroducePatternsToTypes < ActiveRecord::Migration[7.1]
+  def change
+    change_table :types do |t|
+      t.text :patterns, null: true
```
Kharonus, line 34: ❓ does this line already contain `default: null`?

mereghost: Yup

PR #17450, opened 2024-12-12
[#59914] Add times for labor costs to the cost report and export

as-op:
# Ticket

https://community.openproject.org/work_packages/59914

Display the start and end times of a TimeEntry in the costs view and the excel export. (If activated by `Setting.allow_tracking_start_and_end_times`)

## Screenshots

If the unit "Labor" is selected and no grouping is selected
cost_query

In exported excel sheet "Labor"
excel

* Moved some code from PDF Timesheet to helper for reuse
* Split up table code to insert the column(s) after the spent_on column.
* In the excel sheet there are separate columns for each time.

# Merge checklist

- [x] Added/updated tests


klaustopher (approved):
very good!

klaustopher (changes requested):
Sorry, rolling back ... There's a bug when you have set the language to German. In this case, the `unit` number is returned formatted to the german format (i.e. `4,5`) and when adding this to the start time, we get a wrong result because it only adds `4 hours` not 4 hours and 30 minutes:

grafik

as-op:
@klaustopher Waaah, very good catch! Thanks, will fix it.

klaustopher (approved):

PR #17458, opened 2024-12-13
bug/59458-activity-tab-does-not-scroll-to-the-bottom-or-correct-comment-on-mobile-fixed

jjabari-op:
[Ticket](https://community.openproject.org/work_packages/59458)

Fixed branch for OP#59458

Use a MutationObserver so that we can continuously watch for changes on the DOM (specifically on the activity tab container), and when we detect no more changes after a given timeout, fire the scroll event because now we know that the element is fully rendered and we can reliably calculate how far to scroll in pixels.


frontend/src/stimulus/controllers/dynamic/work-packages/activities-tab/index.controller.ts
```diff
@@ -380,11 +380,42 @@ export default class IndexController extends Controller {
...
+          top: scrollableContainer.scrollHeight,
+          behavior: 'smooth',
+        });
+      }, 100);
+    });
+
+    observer.observe(scrollableContainer, {
+      childList: true,
+      subtree: true,
+      attributes: true,
```
akabiru, line 413: 🟢 Seems we can exclude observing the attributes- childList & subtree from turbo frame & streams mutations should suffice I think

```suggestion
    observer.observe(scrollableContainer, {
      childList: true,
      subtree: true,
```

akabiru (approved):
Looks good to me 👍🏾 Thank you for the brief blurb on the approach as well!

PR #17459, opened 2024-12-13
feature: add work package number from the current branch to commit msg

EinLama:
A small lefthook extension that executes a bash script when the commit message is prepared.

Will try to extract the work package number from the name of the current branch.

If found, the commit message will automatically be prefixed with the current wp number, in the following format:

    [#58160] <your message here>

If the WP number cannot be found for any reason, the commit message will remain unchanged. Existing commit message preparations are respected. The prefix is only added when the commit message would be otherwise blank AND the WP number could be found.

### DIFF
--- a/.lefthook/prepare-commit-msg/commit-msg-prefix-from-branch.sh
+++ b/.lefthook/prepare-commit-msg/commit-msg-prefix-from-branch.sh
@@ -1,0 +1,35 @@
+#!/bin/bash
+# Looks at the current branch name and tries to extract the correct work package number from it.
+# If found, the commit message will automatically be prefixed with the work package number, following this format:
+#
+#   [#58160] <your text here>
+#
+
+# File path for the commit message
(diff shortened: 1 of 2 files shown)


cbliard (commented):
I find it useful, but a bit too intrusive as it may be of personal taste to add this number or not. Currently we do not have rules enforcing having the work package id in each and every commit message.

What about putting it in a `.lefthook.local.example` with an explanatory comment?

Also, maybe a matter of taste, but I prefix without the `#` personnally (`[12345] message` rather than `[#12345] message`.

EinLama:
Thank you, Christophe. It is true that the WP number is not enforced. I personally think it's useful in most cases.

In other channels, people have also stated that they have another personal taste or existing hook setup. I will close this PR and create a local configuration for myself so that I don't cause trouble for anyone else :) If you find this useful, feel free to copy it for yourself, too.

PR #17469, opened 2024-12-16
Discover access token audiences

NobodysNightmare:
# Ticket

https://community.openproject.org/projects/cross-application-user-integration-stream/work_packages/60162

This PR is an extension of previous work in https://github.com/opf/openproject/pull/16940. We want to be able to use tokens stored in the corresponding database model for access to third party services, such as Nextcloud.

There are different ways that we can use these existing tokens for that. The case handled in this PR is that the token might already be immediately usable for use in certain services, which we can discover from the token's audience.

We are expecting the access token to be a JWT that we can parse and verify using the metadata we have configured for the corresponding OIDC provider. While there is no guarantee that access tokens can be parsed as JWTs, it's very common to find when dealing with an OIDC IDP, since those are required to provide their ID tokens as JWTs, so the "infrastructure" for signing JWTs exists anyways.

To perform the parsing I extracted previously existing code from the `JwtOidc` warden strategy into a new parser service _and adapted it_ to the common needs of the warden strategy and our new code.

# Merge checklist

- Added/updated tests
    -  [x] For JWT parsing
    - [x] For audience discovery

### DIFF
--- a/modules/openid_connect/spec/services/openid_connect/associate_user_token_spec.rb
+++ b/modules/openid_connect/spec/services/openid_connect/associate_user_token_spec.rb
@@ -30,7 +30,7 @@
 require "spec_helper"
 
 RSpec.describe OpenIDConnect::AssociateUserToken do
-  subject { described_class.new(user).call(**args) }
+  subject { described_class.new(user, jwt_parser: parser).call(**args) }
 
   let(:user) { create(:user) }
 
@@ -38,6 +38,9 @@ RSpec.describe OpenIDConnect::AssociateUserToken do
 
   let(:access_token) { "access-token-foo" }
   let(:refresh_token) { "refresh-token-bar" }
+
+  let(:parser) { instance_double(OpenIDConnect::JwtParser, parse: [parsed_jwt, nil]) }
+  let(:parsed_jwt) { { "aud" => ["aud1", "aud2"] } }
 
   before do
     allow(Rails.logger).to receive(:error)
@@ -50,12 +53,50 @@ RSpec.describe OpenIDConnect::AssociateUserToken do
     expect(token.user_id).to eq user.id
     expect(token.access_token).to eq access_token
     expect(token.refresh_token).to eq refresh_token
-    expect(token.audiences).to eq ["io"]
+    expect(token.audiences).to contain_exactly("io", "aud1", "aud2")
   end
 
   it "logs no error" do
     subject
     expect(Rails.logger).not_to have_received(:error)
+  end
+
+  context "when the JWT encodes aud as a string" do
+    let(:parsed_jwt) { { "aud" => "aud1" } }
+
+    it "creates a correct user token", :aggregate_failures do
+      expect { subject }.to change(OpenIDConnect::UserToken, :count).by(1)
+
+      token = OpenIDConnect::UserToken.last
+      expect(token.access_token).to eq access_token
+      expect(token.refresh_token).to eq refresh_token
+      expect(token.audiences).to contain_exactly("io", "aud1")
+    end
+
+    it "logs no error" do
+      subject
+      expect(Rails.logger).not_to have_received(:error)
+    end
+  end
+
+  context "when the access token is not a valid JWT" do
+    before do
+      allow(parser).to receive(:parse).and_raise("Oops, not a JWT!")
+    end
+
+    it "creates a correct user token", :aggregate_failures do
+      expect { subject }.to change(OpenIDConnect::UserToken, :count).by(1)
+
+      token = OpenIDConnect::UserToken.last
+      expect(token.access_token).to eq access_token
+      expect(token.refresh_token).to eq refresh_token
+      expect(token.audiences).to contain_exactly("io")
+    end
+
+    it "logs no error" do
+      subject
+      expect(Rails.logger).not_to have_received(:error)
+    end
   end
 
   context "when there is no refresh token" do
@@ -68,7 +109,7 @@ RSpec.describe OpenIDConnect::AssociateUserToken do
       expect(token.user_id).to eq user.id
       expect(token.access_token).to eq access_token
       expect(token.refresh_token).to be_nil
-      expect(token.audiences).to eq ["io"]
+      expect(token.audiences).to contain_exactly("io", "aud1", "aud2")
     end
 
     it "logs no error" do
@@ -106,13 +147,31 @@ RSpec.describe OpenIDConnect::AssociateUserToken do
   context "when there is no audience" do
     let(:args) { { access_token:, refresh_token:, known_audiences: [] } }
 
-    it "does not create a user token" do
-      expect { subject }.not_to change(OpenIDConnect::UserToken, :count)
+    it "creates a correct user token", :aggregate_failures do
+      expect { subject }.to change(OpenIDConnect::UserToken, :count).by(1)
+
+      token = OpenIDConnect::UserToken.last
+      expect(token.access_token).to eq access_token
+      expect(token.refresh_token).to eq refresh_token
+      expect(token.audiences).to contain_exactly("aud1", "aud2")
     end
 
     it "logs no error" do
       subject
       expect(Rails.logger).not_to have_received(:error)
+    end
+
+    context "and the token has no audience defined" do
+      let(:parsed_jwt) { { "sub" => "ject" } }
+
+      it "does not create a user token" do
+        expect { subject }.not_to change(OpenIDConnect::UserToken, :count)
+      end
+
+      it "logs no error" do
+        subject
+        expect(Rails.logger).not_to have_received(:error)
+      end
     end
   end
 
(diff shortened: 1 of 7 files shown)


modules/openid_connect/app/services/openid_connect/provider_token_parser.rb
```diff
@@ -0,0 +1,90 @@
...
+#
+# You should have received a copy of the GNU General Public License
+# along with this program; if not, write to the Free Software
+# Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+#
+# See COPYRIGHT and LICENSE files for more details.
+# +
+
+module OpenIDConnect
+  class ProviderTokenParser
```
NobodysNightmare: I am starting to become unhappy about names.

I called the `UserToken` like that, because it's a token that grants access in the name of the user.

I called the `ProviderTokenParser` like that, because it only allows proper parsing of JWTs that were issued by a configured provider (they are not parseable without knowing the provider).

But now looking at both names next to each other, it doesn't seem to make sense, because there are no tokens to do stuff in the name of a provider. It seems counter intuitive, that the `ProviderTokenParser` would parse a `UserToken` and yet that's what it does...

mereghost: Names are hard as hell. This could easily become just `OIDC::JWTParser` and the `UserToken` could just be `OIDC::Token`.

I suck at naming. XD

modules/openid_connect/app/services/openid_connect/provider_token_parser.rb
```diff
@@ -0,0 +1,95 @@
...
+      @verify_audience = verify_audience
+      @required_claims = required_claims
+    end
+
+    def parse(token)
+      issuer, alg, kid = parse_unverified_iss_alg_kid(token)
+      raise Error, "Token signature algorithm #{alg} is not supported" if SUPPORTED_JWT_ALGORITHMS.exclude?(alg)
+
+      provider = fetch_provider(issuer)
+      raise Error, "The access token issuer is unknown" if provider.blank?
```
mereghost: 🟡 My guess here from the code, is that this exception is being used as flow control to represent an invalid but not exceptional state.

That's were we usually would rely on a `ServiceResult`, `Dry::Monad::Result` or similar object to represent a failed state without all the costs of raising an exception.

modules/openid_connect/app/services/openid_connect/provider_token_parser.rb
```diff
@@ -0,0 +1,95 @@
...
+        unverified_payload["iss"],
+        unverified_header.fetch("alg"),
+        unverified_header.fetch("kid")
+      ]
+    end
+
+    def fetch_provider(issuer)
+      return nil if issuer.blank?
+
+      OpenIDConnect::Provider.where(available: true).find { |p| p.issuer == issuer }
```
mereghost: 🔴 This will load all the providers in memory then run `Array#find`, can't the block become part of the `where` clause?

modules/openid_connect/spec/services/openid_connect/associate_user_token_spec.rb
```diff
@@ -48,14 +52,60 @@
...
     subject
     expect(Rails.logger).not_to have_received(:error)
   end
 
+  it "correctly tries parsing the access token" do
+    subject
+
+    expect(OpenIDConnect::ProviderTokenParser).to have_received(:new)
+      .with(verify_audience: false, required_claims: ["aud"])
+    expect(parser).to have_received(:parse).with(access_token)
```
mereghost: 🟢 isn't this testing the internals of the service, checking if a forcebly injected spy actually is being called with the correct params? 

Can't we use dependency injection here so that we don't need to stub the `.new` method? Is there a easy way to not have to mock/spy at all? Wouldn't be tested on the other examples below?

mereghost (changes requested):
Overall seems solid just one point that'd like to check viability.

modules/openid_connect/app/services/openid_connect/provider_token_parser.rb
NobodysNightmare: It's definitely not a `JWTParser`, because the token being a **JSON Web** Token is optional. Though I'll give `TokenParser` and `Token` a second thought.

mereghost: You could have multiple parsers being chosen at runtime for each type of token, might even clean up the code as each parser gets more and more specialized.

modules/openid_connect/app/services/openid_connect/provider_token_parser.rb
NobodysNightmare: That's a fair comment. My excuse is that all of this started from refactoring previously existing code and extracting it into a common class.

But that excuse is bad. I'll look into getting friends with `Dry::Monad::Result` once more ;-)

mereghost: Feel free to enlist me if you get stuck at using it. =)

modules/openid_connect/app/services/openid_connect/provider_token_parser.rb
NobodysNightmare: `issuer` is part of the options. I'll have to check, but assuming that `options` is a `JSONB` this should still be possible.

Now that I see this code, I also want to double check whether the new code circumvents a cache that the previous code may have been using (via `OpenProject::OpenIDConnect.providers`).

modules/openid_connect/spec/services/openid_connect/associate_user_token_spec.rb
NobodysNightmare: One way avoid the mocking is to pass a real JWT that has certain properties. We do this in `spec/requests/api/v3/authentication_spec.rb` where I think this approach is feasible, because it's an integrative spec.

Here I wanted to stay encapsulated, while still checking whether the tested unit properly behaves at its boundaries (e.g. whether it correctly sets-up token parsing). E.g. the token parser should not raise an error when the received token is for a foreign audience, but still it should ensure that an audience claim is part of the token.

Let me try to follow along with your suggestion: Your idea would be to hand in the token parser into the `AssociateUserToken` service. This would allow us to replace it with a mock during testing more easily (no need to overwrite new).

However, I'd still ask myself where we'd test that the "production parser" is configured correctly. The correct configuration is very much related to the use case of the service under test.

modules/openid_connect/app/services/openid_connect/provider_token_parser.rb
NobodysNightmare: I was even mixing things up. The token parser completely expects a JWT. Just other parts of the code (`AssociateUserToken`) treat JWT as optional. Thanks for the input :)

modules/openid_connect/spec/services/openid_connect/associate_user_token_spec.rb
mereghost: > However, I'd still ask myself where we'd test that the "production parser" is configured correctly. The correct configuration is very much related to the use case of the service under test.

This could be done by the request specs you mentioned. I'm okay with how it is now, but every time I see a stub of `.new`, I think of it as a code smell.

modules/openid_connect/app/services/openid_connect/provider_token_parser.rb
NobodysNightmare: Renamed to `OpenIDConnect::JwtParser`.

modules/openid_connect/app/services/openid_connect/provider_token_parser.rb
NobodysNightmare: Fixed. Now using:

```
OpenIDConnect::Provider.where(available: true).where("options->>'issuer' = ?", issuer).first
```

This is still completely uncached, so every request with a JWT will cause us to fetch the provider once. However, I expect this to be a generally cheap operation and not much more expensive than making the same request to a network cache (a file-based cache would have lower latency).

modules/openid_connect/app/services/openid_connect/provider_token_parser.rb
NobodysNightmare: My latest commit modanizes the `JwtParser` and does so in a way that I think is a net win in readability :)

Let me know what you think.

modules/openid_connect/spec/services/openid_connect/associate_user_token_spec.rb
NobodysNightmare: Dependency injecting the parser now.

modules/openid_connect/app/services/openid_connect/provider_token_parser.rb
NobodysNightmare: My latest commit monadizes the `JwtParser` and does so in a way that I think is a net win in readability :)

Let me know what you think.

mereghost (approved):

PR #17472, opened 2024-12-16
WIP Pattern input field

brunopagno:
# Ticket

https://community.openproject.org/projects/document-workflows-stream/work_packages/59999

[Fancy input](https://www.figma.com/design/I2IVNb97j1fNNCs32H6p2m/Automatic-generated-subjects)

## Screenshots

https://github.com/user-attachments/assets/efcea6d8-1d78-4a42-a762-f860e6ac2d14

The component was built with rails leveraging stimulus + primer where available. There is minimal custom styling. That said, the component does not look like Figma because we wanted to favour primer components.

A few decisions were made here:
- Using a contenteditable field for the input
- A few hacks were necessary (e.g. adding spaces around the editable field so that they are still editable)

### DIFF
--- a/app/components/work_packages/types/pattern_autocompleter.html.erb
+++ b/app/components/work_packages/types/pattern_autocompleter.html.erb
@@ -1,0 +1,119 @@
+<%#-- copyright
+OpenProject is an open source project management software.
+Copyright (C) the OpenProject GmbH
+
+This program is free software; you can redistribute it and/or
+modify it under the terms of the GNU General Public License version 3.
+
+OpenProject is a fork of ChiliProject, which is a fork of Redmine. The copyright follows:
+Copyright (C) 2006-2013 Jean-Philippe Lang
+Copyright (C) 2010-2013 the ChiliProject Team
+
+This program is free software; you can redistribute it and/or
+modify it under the terms of the GNU General Public License
+as published by the Free Software Foundation; either version 2
+of the License, or (at your option) any later version.
+
+This program is distributed in the hope that it will be useful,
+but WITHOUT ANY WARRANTY; without even the implied warranty of
+MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+GNU General Public License for more details.
+
+You should have received a copy of the GNU General Public License
+along with this program; if not, write to the Free Software
+Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+
+See COPYRIGHT and LICENSE files for more details.
+
+++#%>
+
+<%=
+  content_tag(
+    :div,
+    class: "pattern-autocompleter",
+    "data-controller": "pattern-autocompleter",
+    "data-pattern-autocompleter-pattern-initial-value": @value
+  ) do
+%>
+  <%=
+    @input.builder.hidden_field(
+      name,
+      value: @value,
+      data: {
+        "pattern-autocompleter-target": "formInput"
+      }
+    )
+  %>
+
+  <template data-pattern-autocompleter-target="tokenTemplate">
+    <%=
+      render(
+        Primer::Beta::Label.new(
+          contenteditable: false,
+          tag: :span, display: :inline, scheme: :accent,
+          "data-role": :token
+        )
+      ) do
+    %>
+      <%= content_tag(:span, style: "cursor: default;", "data-role": "token-text") { "__VALUE__" } %>
+      <%=
+        render(
+          Primer::Beta::Octicon.new(
+            icon: :x, size: :xsmall, style: "cursor: pointer;margin: 0.2em;",
+            "data-action": "click->pattern-autocompleter#remove_token"
+          )
+        )
+      %>
+    <% end %>
+  </template>
+
+  <%= content_tag(:div, style: "position: relative;") do %>
+      <%=
+        render(
+          Primer::Beta::Octicon.new(
+            icon: "triangle-down",
+            size: :small,
+            style: "cursor: pointer;position: absolute;right: 1em;top: 0.5em;",
+            "data-action": "click->pattern-autocompleter#suggestions_toggle"
+          )
+        )
+      %>
+      <%=
+        render(
+          Primer::Box.new(
+            contenteditable: true,
+            border: true, border_radius: 2, p: 1,
+            style: "white-space: pre-wrap;",
+            "data-pattern-autocompleter-target": "content",
+            data: {
+              action: "keydown->pattern-autocompleter#input_keydown
+                       input->pattern-autocompleter#input_change
+                       mouseup->pattern-autocompleter#input_mouseup
+                       blur->pattern-autocompleter#input_blur"
+            }
+          )
+        )
+      %>
+  <% end %>
+  <%= render(Primer::Box.new(box_shadow: :medium, border_radius: 2)) do %>
+    <%=
+      render(
+        Primer::Alpha::ActionList.new(
+          role: :list,
+          hidden: true,
+          show_dividers: false,
+          "data-pattern-autocompleter-target": "suggestions"
+        )
+      ) do |component|
+        @suggestions.each_key do |key|
+          component.with_divider_content(key.to_s.humanize)
+          entries = @suggestions[key]
+          entries.each do |prop, label|
+            component.with_item(label:, data: select_item_action.merge({ value: prop }))
+          end
+          component.with_divider
+        end
+      end
+    %>
+  <% end %>
+<% end %>
--- a/app/components/work_packages/types/pattern_autocompleter.rb
+++ b/app/components/work_packages/types/pattern_autocompleter.rb
@@ -1,0 +1,51 @@
+# frozen_string_literal: true
+
+# -- copyright
+# OpenProject is an open source project management software.
+# Copyright (C) 2024 the OpenProject GmbH
+#
+# This program is free software; you can redistribute it and/or
+# modify it under the terms of the GNU General Public License version 3.
+#
+# OpenProject is a fork of ChiliProject, which is a fork of Redmine. The copyright follows:
+# Copyright (C) 2006-2013 Jean-Philippe Lang
+# Copyright (C) 2010-2013 the ChiliProject Team
+#
+# This program is free software; you can redistribute it and/or
+# modify it under the terms of the GNU General Public License
+# as published by the Free Software Foundation; either version 2
+# of the License, or (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program; if not, write to the Free Software
+# Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+#
+# See COPYRIGHT and LICENSE files for more details.
+# ++
+
+module WorkPackages
+  module Types
+    class PatternAutocompleter < Primer::Forms::BaseComponent
+      delegate :name, to: :@input
+
+      def initialize(input:, value:, suggestions:)
+        super()
+        @input = input
+        @value = value
+        @suggestions = suggestions
+      end
+
+      def select_item_action
+        {
+          action: "click->pattern-autocompleter#suggestions_select",
+          role: "suggestion-item"
+        }
+      end
+    end
+  end
+end
--- a/app/components/work_packages/types/pattern_autocompleter.sass
+++ b/app/components/work_packages/types/pattern_autocompleter.sass
@@ -1,0 +1,5 @@
+.pattern-autocompleter
+  .selected
+    color: var(--list-item-hover--color)
+    background-color: var(--control-transparent-bgColor-hover)
+
--- a/frontend/src/stimulus/controllers/pattern-autocompleter.controller.ts
+++ b/frontend/src/stimulus/controllers/pattern-autocompleter.controller.ts
@@ -1,0 +1,308 @@
+/*
+ * -- copyright
+ * OpenProject is an open source project management software.
+ * Copyright (C) the OpenProject GmbH
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU General Public License version 3.
+ *
+ * OpenProject is a fork of ChiliProject, which is a fork of Redmine. The copyright follows:
+ * Copyright (C) 2006-2013 Jean-Philippe Lang
+ * Copyright (C) 2010-2013 the ChiliProject Team
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU General Public License
+ * as published by the Free Software Foundation; either version 2
+ * of the License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ *
+ * See COPYRIGHT and LICENSE files for more details.
+ * ++
+ */
+
+import { Controller } from '@hotwired/stimulus';
+
+export default class PatternAutocompleterController extends Controller {
+  static targets = [
+    'tokenTemplate',
+    'content',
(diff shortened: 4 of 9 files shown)


frontend/src/stimulus/controllers/pattern-autocompleter.controller.ts
```diff
@@ -0,0 +1,305 @@
...
+
+  input_change() {
+    // browsers insert a `<br>` tag on empty contenteditable elements so we need to cleanup
+    if (this.contentTarget.innerHTML === '<br>') {
+      this.contentTarget.innerHTML = ' ';
+    }
+
+    this.ensureSpacesAround();
+
+    // show suggestions for the current word
```
NobodysNightmare, line 100: While using the component, I realized that suggestions do not change when I navigate with the mouse or the arrow keys. E.g. when I start typing `Aut`, a suggestion for `Author` pops up.

Now when I either click somewhere else or use arrow keys to move the cursor away, I'd expect the auto-completion to disappear or possibly update for the next selected word. Instead the auto completion proposal stays expanded.

frontend/src/stimulus/controllers/pattern-autocompleter.controller.ts
```diff
@@ -0,0 +1,305 @@
...
+
+  input_mouseup() {
+    this.setRange();
+  }
+
+  input_focus() {
+    this.setRange();
+  }
+
+  input_blur() {
```
NobodysNightmare, line 123: I think blur is also a good moment to close the auto-completion. I noticed that I automatically (i.e. through muscle memory) clicked next to the input when I wanted the dropdown to close and then I was surprised that it remained open.

frontend/src/stimulus/controllers/pattern-autocompleter.controller.ts
```diff
@@ -0,0 +1,305 @@
...
+
+      const selectedItem = this.suggestionsTarget.querySelector('.selected') as HTMLElement;
+      if (selectedItem) {
+        this.insertToken(this.createToken(selectedItem.dataset.value!));
+        this.clearSuggestionsFilter();
+      }
+    }
+
+    // move up and down the suggestions selection
+    if (event.key === 'ArrowUp') {
```
NobodysNightmare, line 73: What I see is that we implement the behaviour of a select box manually here. This is tricky. For example I noticed that when using a screen reader, the current implementation does not read out my selection when I change it using the arrow keys. A normal select box (e.g. the color selection in the settings tab) does this.

That's an accessibility concern as far as I can tell. **However**, I realize that the same problem can be observed for two other places in our UI: The project selection in the top left and our emoji autocompleter. So maybe we never solved this so far. (for comparison, the GitHub emoji autocompleter behaves fine)

Sadly I am by no means an expert in _implementing_ accessible custom components. As far as I can tell, the usually best practice is to start with a standard component that is already accessible, e.g. using a real select box in the background.

NobodysNightmare (changes requested):
I've done some testing mostly from the UI side (i.e. not looking at the code first, but checking behaviour first) and left some comments.

Feel free to give me feedback on whether you agree or disagree to the findings, but also whenever you think that addressing a finding takes too much effort and should rather be extracted to a separate PR.

frontend/src/stimulus/controllers/pattern-autocompleter.controller.ts
brunopagno: I had trouble deciding when to close or not the autocomplete. But this sounds correct. I might have stayed a bit too long staring at that same screen.

frontend/src/stimulus/controllers/pattern-autocompleter.controller.ts
brunopagno: This was the case at some point. I don't remember why I removed it, but I'll add again.

frontend/src/stimulus/controllers/pattern-autocompleter.controller.ts
brunopagno: I am not sure how to handle this, though. I will reach out to other people on the dev team to see if it helps.

But I was *more afraid* of attempting to change focus/force navigate with something that is supposed to be a shortcut.

Maybe there's an obvious solution.

frontend/src/stimulus/controllers/pattern-autocompleter.controller.ts
```diff
@@ -0,0 +1,308 @@
...
+      event.preventDefault();
+      this.selectSuggestionAt(this.selectedSuggestion.index - 1);
+    }
+    if (event.key === 'ArrowDown') {
+      event.preventDefault();
+      this.selectSuggestionAt(this.selectedSuggestion.index + 1);
+    }
+
+    // close the suggestions
+    if (event.key === 'Escape' || event.key === 'ArrowLeft' || event.key === 'ArrowRight') {
```
NobodysNightmare, line 83: (Very optional)

Small suggestion after realizing that using `Esc` we can close the auto-completer: What do you think about adding `Ctrl+Space` as a combo to open the list again?

After the latest change this can also be useful for cases where the menu accidentially closed during keyboard navigation.

frontend/src/stimulus/controllers/pattern-autocompleter.controller.ts
brunopagno: Adjusted the hiding autocomplete behaviour

NobodysNightmare: Yeah, I also realize that my actual user expectation is not easy.

First things first: **I like the current behaviour** (closing on any arrow L/R input) more than the previous behaviour of never closing.

However, if I wanted to iterate on this (and we can legitimately do that outside the scope of this PR), I'd probably check whether the currently selected word changes as a consequence of the cursor navigation, and only close the input if the selected word changes.

Main reason for writing this down is to provoke denying or approving reactions, i.e. checking whether others would agree.

frontend/src/stimulus/controllers/pattern-autocompleter.controller.ts
```diff
@@ -0,0 +1,308 @@
...
+      this.updateFormInputValue();
+    }
+  }
+
+  // internal methods
+  private updateFormInputValue():void {
+    this.formInputTarget.value = this.toBlueprint();
+  }
+
+  private ensureSpacesAround():void {
```
NobodysNightmare, line 171: I remember that we talked about this in person and pointed out that it is a hack/workaround. That's perfectly fine!

However, since this code is not self-explanatory, we should make sure the intention is clearly documented. What do you think about adding a comment that briefly explains why these spaces are necessary? Something along the lines of:

> If there is no editable text before or after an uneditable node, it's impossible to move the cursor there. As a workaround to ensure that the user can always move the cursor anywhere, we ensure that there is at least a whitespace character at the beginning and end of our text box.

frontend/src/stimulus/controllers/pattern-autocompleter.controller.ts
```diff
@@ -0,0 +1,308 @@
...
+    }
+  }
+
+  private insertToken(tokenElement:HTMLElement) {
+    if (this.currentRange) {
+      const targetNode = this.currentRange.startContainer;
+      const targetOffset = this.currentRange.startOffset;
+
+      let pos = targetOffset - 1;
+      while (pos > -1 && targetNode.textContent?.charAt(pos) !== ' ') { pos-=1; }
```
NobodysNightmare: Mh, so maybe it's because I spent the past 10 minutes staring at the [Wikipedia page about whitespace](https://en.wikipedia.org/wiki/Whitespace_character), but seeing this code I am now wondering, whether we should not match for the space character and instead match for any kind of unicode whitespace.

I.e.

```suggestion
      while (pos > -1 && !isWhitespace(targetNode.textContent?.charAt(pos))) { pos-=1; }
```

where I leave the implementation of `isWhitespace` to the imagination of the reader.

brunopagno: I found two implementations interesting :sweat_smile: 

```javascript
/\s/.test(ch)
// or
str.trim() === ''
```

This is fun~

frontend/src/stimulus/controllers/pattern-autocompleter.controller.ts
```diff
@@ -0,0 +1,308 @@
...
+    const target = this.tokenTemplateTarget.content?.cloneNode(true) as HTMLElement;
+    const contentElement = target.firstElementChild as HTMLElement;
+    (contentElement.querySelector('[data-role="token-text"]') as HTMLElement).innerText = value;
+    return contentElement;
+  }
+
+  private toHtml(blueprint:string):string {
+    let htmlValue = blueprint.replace(/{{([0-9A-Za-z_]+)}}/g, (_, token:string) => this.createToken(token).outerHTML);
+    if (htmlValue.startsWith('<')) { htmlValue = ` ${htmlValue}`; }
+    if (htmlValue.endsWith('>')) { htmlValue = `${htmlValue} `; }
```
NobodysNightmare: These two lines seem to be duplicated with `ensureSpacesAround`. Can we make sure to only implement this once?

E.g. the caller of `toHtml(...)` could then call `ensureSpacesAround()` afterwards. I.e. what I'd try to achieve is to "localize" the workaround, making sure that it's not too far spread out.

NobodysNightmare:
Because the PR template always reminds me about it: Should we add our new component also to the Lookbook?

I am not quite sure how this works, but I guess we need to add a file like `lookbook/previews/open_project/common/pattern_autocomplete_preview.rb`, maybe not in "common", because our component has a pretty narrow use case. But unsure.

frontend/src/stimulus/controllers/pattern-autocompleter.controller.ts
brunopagno: Yeah, this is fair. Things that I started doing in different times and did not clean up properly. Thanks for spotting.

app/components/work_packages/types/pattern_autocompleter.rb
```diff
@@ -0,0 +1,51 @@
...
+      def initialize(input:, value:, suggestions:)
+        super()
+        @input = input
+        @value = value
+        @suggestions = suggestions
+      end
+
+      def select_item_action
+        {
+          action: "click->pattern-autocompleter#suggestions_select",
```
NobodysNightmare, line 45: I just spotted this one and noticed that I didn't try using the mouse to select an autocomplete entry yet.

For me it doesn't work to click on an autoselect entry. The pop-up closes, but nothing gets selected.

app/components/work_packages/types/pattern_autocompleter.html.erb
```diff
@@ -0,0 +1,119 @@
...
+    <%=
+      render(
+        Primer::Beta::Label.new(
+          contenteditable: false,
+          tag: :span, display: :inline, scheme: :accent,
+          "data-role": :token
+        )
+      ) do
+    %>
+      <%= content_tag(:span, style: "cursor: default;", "data-role": "token-text") { "__VALUE__" } %>
```
HDinger, line 58: Please do  not use inline styles. Either use system_arguments or if the style is not supported, add a class and write the styles in the CSS file.

app/components/work_packages/types/pattern_autocompleter.html.erb
```diff
@@ -0,0 +1,119 @@
...
+      <%=
+        render(
+          Primer::Beta::Octicon.new(
+            icon: :x, size: :xsmall, style: "cursor: pointer;margin: 0.2em;",
```
HDinger, line 62: Same, please do not use inline styles. Further, 0.2 is somehow a magic number, that does not even align the cross to the center:
Bildschirmfoto 2025-01-15 um 07 56 45

Please try to resolve that with CSS alignment mechanism (e.g. `vertica-align` or a `flex`)

app/components/work_packages/types/pattern_autocompleter.html.erb
```diff
@@ -0,0 +1,119 @@
...
+            "data-action": "click->pattern-autocompleter#remove_token"
+          )
+        )
+      %>
+    <% end %>
+  </template>
+
+  <%= content_tag(:div, style: "position: relative;") do %>
```
HDinger, line 70:
```suggestion
  <%= render(Primer::BaseComponent.new(tag: :div, position: :relative)) do %>
```

app/components/work_packages/types/pattern_autocompleter.html.erb
```diff
@@ -0,0 +1,119 @@
...
+      <%=
+        render(
+          Primer::Beta::Octicon.new(
+            icon: "triangle-down",
+            size: :small,
+            style: "cursor: pointer;position: absolute;right: 1em;top: 0.5em;",
```
HDinger, line 76: Position absolute has some drawbacks in this case. You are enforcing it to lay at that specific place no matter what the content of the input is. Thus the content and the arrow can overlap:

Bildschirmfoto 2025-01-15 um 08 18 12

app/components/work_packages/types/pattern_autocompleter.sass
```diff
@@ -0,0 +1,5 @@
+.pattern-autocompleter
+  .selected
+    color: var(--list-item-hover--color)
```
HDinger, line 3: Looking at the ActionList component, they use `--fgColor-default` for hovered elements. As we try to mimic that style, I'd suggest that we use the same here.

frontend/src/stimulus/controllers/pattern-autocompleter.controller.ts
HDinger: The idea with using Primer was to fully spport accessibility again, as the Primer components are claimed to be accessible. However, the ActionList in Primer is not meant to be used standalone afaik, so we are more or less misusing it here. Looking at other [components](https://primer-lookbook.github.com/view-components/lookbook/inspect/primer/alpha/action_menu/single_select) where the actionList is used, it is possible to navigate via arrows. Primer has an internal helper called `focus-group` for that, see:

* https://github.com/primer/view_components/blob/main/app/components/primer/focus_group.ts
* https://github.com/primer/view_components/blob/main/app/components/primer/alpha/action_menu.html.erb#L2

Maybe, we simply have to add that around our ActionList as well and we can get rid of the custom selection logic. But that is just an educated guess, I don't know what that really works.

HDinger (changes requested):
Hi @brunopagno 

I don't know which detail of review you want at that point. I will just write down the issues I found, feel free to ignore when you already aware of that.

- [ ] When loading the page, the field is shortly shown. Sometimes, it there so long that I even could start typing before it vanished
Bildschirmfoto 2025-01-15 um 07 54 54

- [ ] The search results are only selectable via keyboard, not via mouse click

- [ ] The label text and closing "x" are not alligned
Bildschirmfoto 2025-01-15 um 07 56 45

- [ ] The labels can overflow the input
Bildschirmfoto 2025-01-15 um 08 00 57

- [ ] When there are no search results, the headlines are still shown, which feels weird
Bildschirmfoto 2025-01-15 um 07 56 22

- [ ] I can't reliably reproduce it, but sometimes there are issues with the cursor position which is jumping around. For example when you select a label as the first element, it jumps to the right after pressing space.
Jan-15-2025 08-09-38

In general it works nicely. Looking at the code, it still feels however as if we should have solved that differently from the UX side from the beginning. We are implementing a lot of custom stuff here and misusing components for that. I know that it is too late now, but we should keep that in mind for later similar cases.

app/components/work_packages/types/pattern_autocompleter.html.erb
NobodysNightmare: I didn't see that arrow down at all until now :see_no_evil:

Devils advocate: Do we need this arrow at all? It makes the component look like a select, when it is mostly like a text input. E.g. we'd also not show an arrow down in a text input that allows autocompleting emojis (which I think is a good analogy for what we are doing here).

app/components/work_packages/types/pattern_autocompleter.rb
NobodysNightmare: New learning: Selecting with the mouse works after opening the suggestion list via the arrow down button, but not when using the mouse after starting autocompletion via typing.

brunopagno: I shall investigate. Maybe the focus group solution suggested above fixes all things and I can move on with life :eyes:

app/components/work_packages/types/pattern_autocompleter.html.erb
brunopagno: The arrow allows for a mouse only interaction of adding tokens so I think it might be necessary. :thinking:

HDinger:
Some other things:

- [ ] When I selected and item from the list via Enter, and the press Enter again, the element is added again to the input
Jan-15-2025 09-24-18


- [ ] In Safari, after I added an item, the cursor jumps to the beginning of the input
Jan-15-2025 09-24-10


- [ ] Currently it is not possible to delete a label via keyboard, only with a mouse. That is also an accessibility issue.

brunopagno:
Closing this PR in favour of https://github.com/opf/openproject/pull/17634

More information at https://community.openproject.org/projects/document-workflows-stream/work_packages/60625/activity

PR #17482, opened 2024-12-17
wrap form autocompleters into primer form classes

ulferts:

Wraps the OP specific form inputs for autocompletion (work package, project, color and the generic autocompleter) into a div just like the input fields provided by primer. That div then receives the css classes the primer inputs receive as well. Additionally, it allows providing the `input_width` parameter so that the width of the input is limited when needed. 

Before the change, the dom structure for the color autocompleter looks like this:

image

After, it looks like this:

image

That change is then applied to the status form where the color select now has the same width as the input for the name attribute:

image

The PR also adds  a preview for the color select which was missing before.

### DIFF
--- a/app/forms/statuses/form.rb
+++ b/app/forms/statuses/form.rb
@@ -106,7 +106,8 @@ module Statuses
       statuses_form.color_select_list(
         label: attribute_name(:color_id),
         name: :color_id,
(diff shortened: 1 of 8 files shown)


HDinger (approved):
Nice 👍 Thanks for the effort

PR #17496, opened 2024-12-18
[#60291] Apply view permissions on the Project List page

dombesz:
# Ticket
https://community.openproject.org/work_packages/60291

Apply stages and gates view permissions on the project list page. 
 - Do not display the Stage and Gate definitions, if the user does not have the `:view_project_stages_and_gates` permission in any projects.
 - Do not display Stages and Gates value entries for projects without the `:view_project_stages_and_gates`.
# Merge checklist

- [X] Added/updated tests

### DIFF
--- a/app/components/projects/row_component.rb
+++ b/app/components/projects/row_component.rb
@@ -98,7 +98,7 @@ module Projects
     end
 
     def life_cycle_step_column(column)
-      return nil unless user_can_view_project?
+      return nil unless user_can_view_project_stages_and_gates?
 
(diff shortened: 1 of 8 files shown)


ulferts (changes requested):
@dombesz I fear it is a bit more complicated. It might be that the user has the permission to see the dates in one project but doesn't have it in the next. So if the user lists projects, the dates set for the former should be visible but at the same time, they should be invisible on the later.

dombesz:
Thanks @ulferts , I'll add the permissions verification on the row level.

dombesz:
Hi @ulferts I addressed your CR comments.

oliverguenther (approved):

PR #17502, opened 2024-12-19
Optimizing the life cycle dialog

ulferts:
# Ticket

https://community.openproject.org/wp/60330

Improve the UX of date pickers in modals i.e. on the lifecycle modal:
* The date picker can now reach out of the modal which prevents the modal needing to increase and the user to scroll to the added content

## Screenshots

Kapture 2024-12-19 at 21 21 14

The approach is to append the date picker to the modal. However, that is not enough. The date picker dom element also needs to be leveraged to the top-layer which is done right after opening the date picker. Positioning can easily be fixed by removing their top margin. 

Because of the backdrop added when moving the date picker to the top layer, this move to the top layer needs to be repeated also when the input is clicked on again as that is treated as a click on the backdrop which removes an item from the top layer.

# Merge checklist

- [ ] Added/updated tests

### DIFF
--- a/app/forms/projects/life_cycles/form.rb
+++ b/app/forms/projects/life_cycles/form.rb
@@ -53,7 +53,7 @@ module Projects::LifeCycles
         label: "#{icon} #{text}".html_safe, # rubocop:disable Rails/OutputSafety
         leading_visual: { icon: :calendar },
         datepicker_options: {
-          inDialog: true,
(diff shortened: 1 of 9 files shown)


ulferts:
I would welcome hints on how to avoid the code duplication between the single and the range date picker.

bsatarnejad (approved):
Code looks well 👍

PR #17506, opened 2024-12-19
Add an enterprise banner on top of automatic subject configuration

NobodysNightmare:
We are making sure that users can revert back to manual subjects at any time and even keep using the feature if it's still enabled after their enterprise token expired. They will merely not be able to set new types to automatic and have to live with an enterprise banner being shown to them whenever they change the subject config.

# Ticket
https://community.openproject.org/projects/document-workflows-stream/work_packages/59929

Show users that this feature is enterprise-exclusive, while retaining their ability to use the feature (or stop using it) once their EE license has expired

## Screenshots

image

Our new standard component for EE banners, because standardization is great!

# Merge checklist

- [x] Added/updated tests

### DIFF
--- a/app/forms/work_packages/types/subject_configuration_form.rb
+++ b/app/forms/work_packages/types/subject_configuration_form.rb
@@ -45,6 +45,7 @@ module WorkPackages
             checked: has_pattern?,
             label: I18n.t("types.edit.subject_configuration.automatically_generated_subjects.label"),
             caption: I18n.t("types.edit.subject_configuration.automatically_generated_subjects.caption"),
+            disabled: !EnterpriseToken.active? && !has_pattern?,
             data: { action: "admin--subject-configuration#showPatternInput" }
           )
         end
--- a/lib/open_project/static/links.rb
+++ b/lib/open_project/static/links.rb
@@ -251,6 +251,9 @@ module OpenProject
               href: "https://www.openproject.org/docs/user-guide/time-and-costs/progress-tracking/"
             },
             enterprise_docs: {
+              automatic_subject_generation: {
+                href: "https://www.openproject.org/docs/system-admin-guide/manage-work-packages/work-package-types/#work-package-subject-configuration-enterprise-add-on"
+              },
               form_configuration: {
                 href: "https://www.openproject.org/docs/system-admin-guide/manage-work-packages/work-package-types/#work-package-form-configuration-enterprise-add-on"
               },
--- a/app/components/enterprise_edition/banner_component.rb
+++ b/app/components/enterprise_edition/banner_component.rb
@@ -29,9 +29,9 @@
 # ++
 
 module EnterpriseEdition
-  # Add a general description of component here
-  # Add additional usage considerations or best practices that may aid the user to use the component correctly.
-  # @accessibility Add any accessibility considerations
+  # A banner indicating that a given feature requires the enterprise edition of OpenProject.
+  # This component uses conventional names for translation keys or URL look-ups based on the feature_key passed in.
+  # It will only be rendered if necessary.
   class BannerComponent < ApplicationComponent
     include OpPrimer::ComponentHelpers
 
@@ -50,6 +50,7 @@ module EnterpriseEdition
                    **system_arguments)
       @system_arguments = system_arguments
       @system_arguments[:tag] = "div"
+      @system_arguments[:test_selector] = "op-ee-banner-#{feature_key.to_s.tr('_', '-')}"
       super
 
       @feature_key = feature_key
--- a/app/components/work_packages/types/subject_configuration_component.html.erb
+++ b/app/components/work_packages/types/subject_configuration_component.html.erb
@@ -28,6 +28,10 @@ See COPYRIGHT and LICENSE files for more details.
 ++#%>
 
 <%=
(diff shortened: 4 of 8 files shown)


lib/open_project/static/links.rb
```diff
@@ -251,6 +251,9 @@ def static_links
               href: "https://www.openproject.org/docs/user-guide/time-and-costs/progress-tracking/"
             },
             enterprise_docs: {
+              automatic_subject_generation: {
+                href: "https://www.openproject.org/docs/todo/do-not-merge"
```
NobodysNightmare: Can we already know/guess the correct URL this will have in the future? Otherwise we'll need to remember to revisit this link and fix it before public release.

NobodysNightmare:
@ulferts Not sure, but this might be the first time that we use the new banner component :tada: 

As can be seen in the screenshot above, there should be additional space between the banner and the form.

Do you think this is something that I should fix on a "this usage only" basis (e.g. by figuring out how to use `system_arguments` or if all else fails adding a `<br>` tag :wink: ) or should we rather introduce a default bottom-margin to the banner?

NobodysNightmare:
I talked to @HDinger in your absence. She expressed a preference in leaving margins/paddings out of the component and defining this on a case-by-case basis

lib/open_project/static/links.rb
Kharonus: This needs to be discussed with Maya. Maybe she can predict it, if not, we need to put a placeholder.

NobodysNightmare: Aligned with @MayaBerd and added a small placeholder into the docs already and pointed to the correct place.

Note: We should now not merge this branch before todays feature freeze (so that the docs are not updated before 15.3).

(Alternatively if we need to merge this today because of dependencies or what not --> just remove the text from the docs and we are safe as well)

app/forms/work_packages/types/subject_configuration_form.rb
```diff
@@ -36,13 +36,15 @@ class SubjectConfigurationForm < ApplicationForm
           group.radio_button(
             value: "manual",
             checked: !has_pattern?,
+            disabled: !EnterpriseToken.active? && !has_pattern?,
```
bsatarnejad: Based on the criteria:

> If the current mode is automatically generated subjects the admin must be allowed to enable mode manually editable subjects.

**Manually editable subjects** button shouldn't be disabled.

bsatarnejad (changes requested):

app/forms/work_packages/types/subject_configuration_form.rb
NobodysNightmare: It isn't :thinking: 

Maybe all the negation is irritating (sadly there is no `enabled`, which would be much easier to read):

```
disabled = !EnterpriseToken.active? && !has_pattern?

# assuming the enterprise token is not active and we have automatically generated subjects
disabled = !false && !true
disabled = true && false
disabled = false # not disabled
```

Would you find it more readable if I applied de morgan's law?

```suggestion
            disabled: !(EnterpriseToken.active? || has_pattern?),
```

(I found that less readable, which is why I wrote it the other way around)

bsatarnejad: I think that there are no conditions specified for disabling the **Manually editable subjects** button, so it should always remain enabled. What's your opinion on it?

NobodysNightmare: In general I'd prefer to enable both radio buttons or none of them, because otherwise I'd find the UI irritating. 

Maybe I should see whether it's possible to disable the entire form (instead of disabling per radio button).

NobodysNightmare: > Maybe I should see whether it's possible to disable the entire form (instead of disabling per radio button).

I've now visibly disabled the radio button group based on the same condition instead of individual buttons. So as soon as the user _saves_ their preference to not use a pattern anymore, the radio buttons disable, but as long as the selection is not saved, they can toggle back and forth.

> so it should always remain enabled. What's your opinion on it?

To me this is pretty clearly left open as an implementation detail. You will find that both suggestions (yours and mine) will not contradict acceptance criteria.

edit: I reflected on it once more. I am now going to change it to always enable the manual mode radio button and enable the automatic mode based on the condition previously used. I.e. users in automatic mode will still be able to toggle back and forth, but users in manual mode will see that their current mode is their only choice.

image

bsatarnejad: Hi @NobodysNightmare 
It looks much better now, I just see one issue here: when There is no active enterprise token and the current mode is **Automatically generated subjects**, **Subject pattern** input is not visible. The Criteria say that:

> If the current mode is automatically generated subjects the admin must be allowed to still edit and save the current pattern.

Screenshot 2025-01-08 at 15 46 49

NobodysNightmare: Yes, that's an issue outside the scope of this PR. The input is never visible when the code is currently `auto` and you refresh the page, even if there is an enterprise token.

A fix of this will be part of https://github.com/opf/openproject/pull/17560

bsatarnejad (approved):
LGTM 👍

PR #17514, opened 2024-12-20
Wrap reminder note so that it is always visible

akabiru:
# Ticket

https://community.openproject.org/work_packages/60196

When a notification entry has a reminder note, unset the height allowing the reminder note to expand the height of the notification entry. By default, *_reminder notes are limited to 128 chars_* so we shouldn't have overly extensive heights.

## Screenshots

https://github.com/user-attachments/assets/e52f4ea5-60c9-4c60-aa3c-f91f34303ebf

Screenshot 2024-12-20 at 2 11 58 PM

Define the reminder alert layout as a multi-line flex container

# Merge checklist

- [ ] Added/updated tests

### DIFF
--- a/frontend/src/app/features/in-app-notifications/entry/reminder-alert/in-app-notification-reminder-alert.component.sass
+++ b/frontend/src/app/features/in-app-notifications/entry/reminder-alert/in-app-notification-reminder-alert.component.sass
@@ -1,16 +1,15 @@
 @import "helpers"
 
+.op-ian-item
+  &:has(.op-ian-reminder-alert--note)
+    height: unset
+
 .op-ian-reminder-alert
-  display: grid
-  grid-template-columns: auto 1fr
-  grid-column-gap: $spot-spacing-0_25
-  align-items: center
+  display: flex
+  gap: $spot-spacing-0_25
+  flex-wrap: wrap
   color: var(--fgColor-muted)
 
   &--note
-    @include text-shortener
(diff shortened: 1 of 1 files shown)


HDinger (changes requested):
Hi @akabiru 

on the notification center we have a virtual scrolling implemented. We use the `cdk` viruall scrolling for that, which needs a fixed item height to know when it has to load more items. So when you are unsetting that height, the virtual scrolling will not work correctly any more. 

Bildschirmfoto 2024-12-20 um 12 30 03

I have not looked into the cdk virtual scrolling, and thus don't know whether it can by now work with flexible heights, but I honestly doubt it. That being said, I fear the feature can't really be implemented as desired.

akabiru:
> Hi @akabiru
> 
> on the notification center we have a virtual scrolling implemented. We use the `cdk` viruall scrolling for that, which needs a fixed item height to know when it has to load more items. So when you are unsetting that height, the virtual scrolling will not work correctly any more.
> Bildschirmfoto 2024-12-20 um 12 30 03
> 
> I have not looked into the cdk virtual scrolling, and thus don't know whether it can by now work with flexible heights, but I honestly doubt it. That being said, I fear the feature can't really be implemented as desired.

Thanks @HDinger ! I just realised that and removed your review request, but you were too fast ⚡ 😄 I fear the same as well, happy to spend some bit of time to see if there's a workaround 🤔

PR #17516, opened 2024-12-20
Update Work Package subject using a Type defined blueprint

mereghost:
Related WP: [OP#60003](https://community.openproject.org/projects/document-workflows-stream/work_packages/60003)

## [Wat?](https://www.destroyallsoftware.com/talks/wat)

This PR aims to make possible the definition/replacement of a `WorkPackage#subject` based on blueprint defined on its `Type`.

## How?

There are a lot of moving parts here. The `PatternCollection`, `Pattern` and their ActiveRecord value types where already in place.

Now comes the `PatternResolver` that is capable of taking the blueprint from a `Pattern` parsing it and resolving it into a string for replacement.

To update the Work Packages, changes where necessary on the `SetAttributeService`, `CreateService` and `UpdateService`.

## Details

`PatternResolver` resolves the tokens (basically fragments on double curly braces, i.e. `{{token}}`) using `TokenPropertyMapper`.

The `TokenPropertyMapper` is the source of all `property -> token` information, including which tokens are available. When called tries to find the key on its predefined attributes and if it is not found, it tries to call the method on the passed context and get its value. It also implements the I18n layer for the names of all the properties.

Things get more complicated when we get to custom fields, as those are dynamically added to work packages and can vary from installation to installation.

But all custom field assigned values can be accessed via the `custom_field_ID` method, so we first check the context:

- `{{custom_field_123}}` is considered part of the Work Package
- `{{parent_custom_field_123}}` will try to get `work_package.parent.custom_field_123`
- `{{project_custom_field_123}}` will try to get the project attribute 123 in the same way above.

The `SetAttributesService` needed to be updated so that it overrides whatever subject was there in the first place with a message saying that it will be updated later. This is important in case something fails later on.

Update and Create services do their respective updates very late in the process, to make sure all the changes that might impact the subject are already assigned.

On Create specifically we save first, just in case we need some auto-generated field like `created_at`.

### DIFF
--- a/app/models/type.rb
+++ b/app/models/type.rb
@@ -1,3 +1,5 @@
+# frozen_string_literal: true
+
 #-- copyright
 # OpenProject is an open source project management software.
 # Copyright (C) the OpenProject GmbH
@@ -34,9 +36,11 @@ class Type < ApplicationRecord
 
   include ::Scopes::Scoped
 
-  attribute :patterns, Types::PatternCollectionType.new
+  attribute :patterns, Types::Patterns::CollectionType.new
 
   before_destroy :check_integrity
+
+  belongs_to :color, optional: true, class_name: "Color"
 
   has_many :work_packages
   has_many :workflows, dependent: :delete_all do
@@ -52,12 +56,9 @@ class Type < ApplicationRecord
                           join_table: "#{table_name_prefix}custom_fields_types#{table_name_suffix}",
                           association_foreign_key: "custom_field_id"
 
-  belongs_to :color, optional: true, class_name: "Color"
-
   acts_as_list
 
   validates :name, presence: true, uniqueness: { case_sensitive: false }, length: { maximum: 255 }
-
   validates :is_default, :is_milestone, inclusion: { in: [true, false] }
 
   scopes :milestone
@@ -65,8 +66,9 @@ class Type < ApplicationRecord
   default_scope { order("position ASC") }
 
   scope :without_standard, -> { where(is_standard: false).order(:position) }
+  scope :default, -> { where(is_default: true) }
 
-  def to_s; name end
+  delegate :to_s, to: :name
 
   def <=>(other)
     name <=> other.name
@@ -81,26 +83,20 @@ class Type < ApplicationRecord
   end
 
   def self.standard_type
-    ::Type.where(is_standard: true).first
-  end
-
-  def self.default
-    ::Type.where(is_default: true)
+    where(is_standard: true).first
   end
 
   def self.enabled_in(project)
-    ::Type.includes(:projects).where(projects: { id: project })
+    includes(:projects).where(projects: { id: project })
   end
 
   def statuses(include_default: false)
     if new_record?
       Status.none
     elsif include_default
-      ::Type
-        .statuses([id])
-        .or(Status.where_default)
+      self.class.statuses([id]).or(Status.where_default)
     else
-      ::Type.statuses([id])
+      self.class.statuses([id])
     end
   end
 
@@ -108,9 +104,24 @@ class Type < ApplicationRecord
     object.types.include?(self)
   end
 
+  def replacement_patterns_defined?
+    return false if patterns.blank?
+
+    patterns.all_enabled.any?
+  end
+
+  def enabled_patterns
+    return {} if patterns.blank?
+
+    patterns.all_enabled
+  end
+
   private
 
   def check_integrity
-    raise "Can't delete type" if WorkPackage.where(type_id: id).any?
+    throw :abort if is_standard?
+    throw :abort if WorkPackage.exists?(type_id: id)
+
+    true
   end
 end
--- a/app/models/type.rb
+++ b/app/models/type.rb
@@ -107,7 +107,7 @@ class Type < ApplicationRecord
   def replacement_patterns_defined?
     return false if patterns.blank?
 
-    patterns.all_enabled.any?
+    enabled_patterns.any?
   end
 
   def enabled_patterns
--- a/app/models/types/pattern_resolver.rb
+++ b/app/models/types/pattern_resolver.rb
@@ -29,29 +29,14 @@
 #++
 
 module Types
-  class PatternMapper
+  class PatternResolver
     TOKEN_REGEX = /{{[0-9A-Za-z_]+}}/
-
-    MAPPING = {
-      type: ->(wp) { wp.type.name },
-      assignee: ->(wp) { wp.assigned_to&.name },
-      created: ->(wp) { wp.created_at },
-      author: ->(wp) { wp.author.name },
-      parent_id: ->(wp) { wp.parent&.id },
-      project_name: ->(wp) { wp.project.name }
-    }.freeze
-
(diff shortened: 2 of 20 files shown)


app/models/type.rb
```diff
@@ -81,36 +83,45 @@ def self.statuses(types)
...
+      self.class.statuses([id])
     end
   end
 
   def enabled_in?(object)
     object.types.include?(self)
   end
 
+  def replacement_patterns_defined?
+    return false if patterns.blank?
```
brunopagno, line 108: Do we need this condition here? I don't know if it adds anything to the one below

app/models/types/pattern_resolver.rb
```diff
@@ -0,0 +1,71 @@
...
+    def stringify(value)
+      case value
+      when Date, Time, DateTime
+        value.strftime("%Y-%m-%d")
+      when NilClass
+        "N/A"
+      else
+        value.to_s
+      end
+    end
```
brunopagno: I'm having ideas here, but we might want to build something clever to define default values for types alongside this structure here, so we only need to parse all types in one place.

app/models/types/patterns/token_property_mapper.rb
```diff
@@ -0,0 +1,75 @@
...
+        project: ->(wp) { wp.project_id },
+        project_active: ->(wp) { wp.project&.active? },
+        project_name: ->(wp) { wp.project&.name },
+        project_status: ->(wp) { wp.project&.status_code },
+        project_parent: ->(wp) { wp.project&.parent_id },
+        project_public: ->(wp) { wp.project&.public? },
+        start_date: ->(wp) { wp.start_date },
+        status: ->(wp) { wp.status&.name },
+        type: ->(wp) { wp.type&.name }
+      }.freeze
```
brunopagno: :clap:

app/services/work_packages/set_attributes_service.rb
```diff
@@ -43,6 +45,15 @@ def set_attributes(attributes)
...
 
     set_custom_attributes(attributes)
+    mark_templated_subject
+  end
+
+  def mark_templated_subject
+    return true unless work_package.type&.replacement_patterns_defined?
+
+    if work_package.type.patterns.all_enabled[:subject]
+      work_package.subject = "Templated by #{work_package.type.name}"
```
brunopagno, line 55: Not sure I follow. Is this a temporary subject? Does this get persisted?

brunopagno (commented):
It's looking pretty good. I can take another look after it is moved from draft.

app/models/types/pattern_resolver.rb
mereghost: Yeah... this is very naive right now on purpose.

app/services/work_packages/create_service.rb
```diff
@@ -67,25 +67,25 @@ def create(attributes, work_package)
...
   end
 
-  def set_attributes(attributes, wp)
-    attributes_service_class
-      .new(user:,
-           model: wp,
-           contract_class:)
-      .call(attributes)
+  def set_templated_subject(work_package)
+    return true unless work_package.type&.replacement_patterns_defined?
```
brunopagno, line 76: We're doing this a few times. Does it make sense to add a helper in the work_package for syntax sugar?

```ruby
work_package.type&.replacement_patterns_defined?
work_package.replacement_patterns_defined?
```

... probably no :sweat_smile:

app/services/work_packages/set_attributes_service.rb
mereghost: Yes. If there's no subject the contract will fail as it is a required field, so I fill in with something for now so that we can overwrite it later on.

It will get persisted as I might need "persisted" info (like `created_at`) for the pattern, but all is done within the context of the same request, so the user won't (hopefully) see it.

app/models/type.rb
mereghost: Not really, but it would avoid more method calls in case of patterns being `nil`.

app/services/work_packages/set_attributes_service.rb
brunopagno: Fair enough. Makes sense to me.

... I thought we might want to somehow tag these so that we can query any lingering/stuck cases later. But honestly `Templated by ` is probably good enough for that.

app/services/work_packages/create_service.rb
mereghost: I wondered the same thing. Ended up not doing it... for now. :P

app/models/types/pattern_resolver.rb
```diff
@@ -0,0 +1,67 @@
...
+# You should have received a copy of the GNU General Public License
+# along with this program; if not, write to the Free Software
+# Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+#
+# See COPYRIGHT and LICENSE files for more details.
+#++
+
+module Types
+  class PatternResolver
+    TOKEN_REGEX = /{{[0-9A-Za-z_]+}}/
```
brunopagno, line 33: Is this the final regex? Don't we need to support `.` or any other symbols?

brunopagno: I am thinking we might need to concat field names and context, like `{{parent.assignee}}`

app/models/types/pattern_resolver.rb
```diff
@@ -0,0 +1,67 @@
...
+
+      stringify(@mapper[token.context_key].call(context))
+    end
+
+    def stringify(value)
+      case value
+      when Date, Time, DateTime
+        value.strftime("%Y-%m-%d")
+      when NilClass
+        "NA"
```
brunopagno, line 61: This one is also potential to change. I wonder if we need to handle this as a translation.

But probably can tackle this separately

app/models/types/patterns/token_property_mapper.rb
```diff
@@ -0,0 +1,118 @@
...
+        TOKEN_PROPERTY_MAP.dig(key, :fn) || DEFAULT_FUNCTION.call(key)
+      end
+
+      alias :[] :fetch
+
+      def tokens_for_type(type)
+        base = default_tokens
+        base[:work_package].merge!(tokenize(work_package_cfs_for(type)))
+        base[:project].merge!(tokenize(project_attributes, "project_"))
+        base[:parent].merge!(tokenize(all_work_package_cfs, "parent_"))
```
brunopagno, line 79: I think this answers the question I posted above about the regex. :+1:

brunopagno (commented):
I think it looks good. It's still marked as draft, but looks right to me.

Left a few less important comments

app/models/types/pattern_resolver.rb
mereghost: Yeah, I went with just `prefix_` as it proved easier to read and less of hassle to parse.

mereghost:
> I think it looks good. It's still marked as draft, but looks right to me.
> 
> Left a few less important comments

I frigging could swear I clicked the "Ready for Review" button. 🤦🏼

brunopagno (approved):
:ship:

PR #17523, opened 2024-12-27
Chore/add frozen string literal true

EinLama:
Following #17481 , make a bunch of our files use `frozen_string_literal: true`

To do so, I filtered out all files that only violate one or both of `Layout/EmptyLineAfterMagicComment` and `Style/FrozenStringLiteralComment`. Files violating more rules would trigger a CI fail.

Then, I rejected changes that caused our pipeline to fail. Those folders/files will slowly be adjusted over time as we further develop OP.

## Script for finding files violating only the frozen string literal cop

To filter for files matching this criterion, I executed

`rubocop --format=json > rubocop_results.json`

Afterwards, I ran this Ruby script to filter out the matching files. `xargs` the file list and toss them into `rubocop -A`.

```ruby
require 'json'

file = File.read('rubocop_results.json')
results = JSON.parse(file)

matching_files = []

results['files'].each do |file|
  path = file['path']
  offenses = file['offenses']

  offense_types = offenses.map { |offense| offense['cop_name'] }.uniq

  if offense_types.all? { |type| %w[Layout/EmptyLineAfterMagicComment Style/FrozenStringLiteralComment].include?(type) } &&
     offense_types.size <= 2 # Ensure no extra offenses
    matching_files << path
  end
end

matching_files.each { |file| puts file }
```

### DIFF
--- a/Dangerfile
+++ b/Dangerfile
@@ -1,3 +1,5 @@
+# frozen_string_literal: true
+
 CORE_OR_MODULE_MIGRATIONS_REGEX = %r{(modules/.*)?db/migrate/.*\.rb}
 
 def added_or_modified_migrations?
--- a/Gemfile
+++ b/Gemfile
@@ -1,3 +1,5 @@
+# frozen_string_literal: true
+
 #-- copyright
 # OpenProject is an open source project management software.
 # Copyright (C) the OpenProject GmbH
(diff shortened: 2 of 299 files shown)


EinLama:
Abandoning the idea of batch-updating our files as some of the files include string operations that are not caught by the specs (usage of `gsub` in `app/models/journable/historic_active_record_relation.rb`). It is indeed risky to update too much at once. Careful inspection of one file at a time is advised.

PR #17524, opened 2024-12-30
Fix/60384 Adjust create_folder_command response

brunopagno:
Create folder has some paths that have issues with encoding. This should fix the behaviour and guarantee that it's properly tested.

# Ticket
https://community.openproject.org/projects/openproject/work_packages/60384

Fix an issue where creating a new folder on NextCloud causes the user to see a broken location picker. The problem was related to the returned path where the location picker should redirect to.

Make escaping follow the same rules on request and response from NextCloud.

# Merge checklist

- [x] Added/updated tests

### DIFF
--- a/modules/storages/spec/common/storages/peripherals/storage_interaction/nextcloud/create_folder_command_spec.rb
+++ b/modules/storages/spec/common/storages/peripherals/storage_interaction/nextcloud/create_folder_command_spec.rb
@@ -32,46 +32,74 @@ require "spec_helper"
 require_module_spec_helper
 
 RSpec.describe Storages::Peripherals::StorageInteraction::Nextcloud::CreateFolderCommand, :webmock do
-  let(:user) { create(:user) }
-  let(:storage) do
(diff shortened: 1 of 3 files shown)


modules/storages/spec/common/storages/peripherals/storage_interaction/nextcloud/create_folder_command_spec.rb
```diff
@@ -32,46 +32,74 @@
...
+  # For the VCR tests in this block it's necessary to
+  # create a user in NextCloud with the account name `member@example1`
+  # and use it's oauth access & refresh tokens on .env.test.local
+  describe "user with custom origin name" do
+    let(:user) { create(:user) }
+    let(:storage) do
+      create(
+        :nextcloud_storage_with_local_connection,
+        :as_not_automatically_managed,
+        origin_user_id: "member@example1",
```
brunopagno, line 88: I tried a few different things on NextCloud but it seems to be validating for special characters, so we cannot pass completely arbitrary usernames.

mereghost (approved):
Looks good. This will take care of the issue.

PR #17527, opened 2025-01-02
Removes custom fields with formats: Text, Bool, Link and Empty from the allowed custom field lists for patterns

mereghost:
This PR adds the finishing touches needed for work packages OP#60001, OP#60016 and OP#60017


brunopagno (approved):
:+1:

PR #17536, opened 2025-01-06
[#59533] Changing an image size is not reflected in the export

as-op:
# Ticket

https://community.openproject.org/work_packages/59533

Fix images sizes in "Generate PDF"

## Screenshots

Before
before

After
after

Remove wrong image styling


machisuji (approved):
👍

PR #17537, opened 2025-01-06
Introduces the position_cache

mereghost:
### Related WP: [OP#60393](https://community.openproject.org/projects/openproject/work_packages/60393/)

The source of the issue is the way that Queries are set. To be able to group and sort hierarchical items we have to issue 2 different queries and then feed the raw SQL. BUT we instantiate Queries on almost every view.

This PR introduces a cache for the calculations we had to do beforehand, eliminating the need to run the queries at all.


db/migrate/20250102161733_adds_position_cache_to_hierarchy_items.rb
```diff
@@ -0,0 +1,47 @@
...
+    reversible do |dir|
+      dir.up do
+        root_node = Struct.new(:id)
+        service = CustomFields::Hierarchy::HierarchicalItemService.new
+        sql = "select id from hierarchical_items where custom_field_id is not NULL;"
+        ActiveRecord::Base.connection.execute(sql).each do |row|
+          service.send :update_position_cache, root_node.new(**row)
+        end
+      end
+    end
```
mereghost: @machisuji This is the migration I mentioned. I can transpose the SQL here if the need be.

brunopagno (approved):
I took a careful look and it seems to be cool. Nice job :clap:

db/migrate/20250102161733_adds_position_cache_to_hierarchy_items.rb
machisuji: @mereghost - One day, in a future far away, this migration will start failing because we changed a signature in the `HierarchicalItemService` or even removed it all together. We've had many instances of this sort of thing happening.

Which is why I always nag people to avoid using services and/or AR model code in migrations.
Always think about if this will still work in 5+ years with new code on the old schema.

With complex migrations like that, I would like to have tests for the migration itself. You can find some examples for these kinds of specs [here](https://github.com/opf/openproject/tree/dev/spec/migrations).

But having a spec on its own is not enough if it isn't done right.
I see two options here.

1) Keep using the service in the spec, but make sure the spec is running on the db schema at the time of writing of the migration, not the latest migrated schema (which it would by default).

This means replicating the hierarchical_items table in its current form for the migration.

Otherwise, it might happen that the table is changed in the future along with the service code, and the migration spec will keep working even though the actual migration wouldn't work on the old schema.

Using the old schema will make the spec fail in that case. Which would then prompt us to fix the migration (probably by using option 2 below).

2) Use plain SQL in the migration (to begin with)

This means effectively copying the definition of `update_position_cache` inside the migration and using that.
This will keep working even if the schema changes in the future.

The spec on the other hand may very well start failing in the future if the schema has changed to be incompatible. 
The fix for this then would again be part of option 1, re-defining the schema of the hierarchical_items table in the old version for the spec.

***

There are discussions to be had about how we manage migrations and if we should have cut-offs at some point, requiring people to migrate one major version at a time, for instance. This would make this whole discussion obsolete.

But at this time we don't have a decision on that yet. So we'll have to be careful here with the migrations.

***

**tl;dr;** I'd like to see a spec for this migration, please. Also, we need to make sure this doesn't fail in the future with new code on an old schema.

Does that all make sense? We can talk about it in person too.

machisuji (changes requested):

mereghost:
@machisuji went with moving the SQL to the migration and also adding a test while I was at it. Better safe than sorry.

Can you take another look at it?

PS: Got an Ok from Markus via DM.

machisuji (approved):
👍 🔥

PR #17538, opened 2025-01-06
Add documentation links to error messages

NobodysNightmare:
Linking to two places in the documentation from the two connection checks we have for storages.

# Ticket
https://community.openproject.org/projects/openproject/work_packages/57337

Adding cross references between error messages and 

## Screenshots
image

The error message for manually triggered connection checks was changed as requested. However, in contrast to acceptance criteria two changes were made:

* The format of the manual check error message was not changed
* The doc link for the periodic check was added into a separate link

Both are because the error message of the periodic check is much less structured. That check purely shows a string stored in the database as an error message. Thus it was easy to wrap the link around the error code for the manual check, but much harder to do the same for the periodic check.

### DIFF
--- a/lib/open_project/static/links.rb
+++ b/lib/open_project/static/links.rb
@@ -44,8 +44,10 @@ module OpenProject
           @links ||= static_links.merge(dynamic_links)
         end
 
-        def url_for(item)
-          links.dig(item, :href)
+        def url_for(*items, anchor: nil)
+          url = links.dig(*items, :href)
+          url += "##{anchor}" if anchor.present?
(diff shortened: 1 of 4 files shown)


NobodysNightmare:
Something I unsuccessfully tried was adding manual anchors to the docs as pure HTML, e.g. `<a name="ERR_CLIENT_INVALID"></a>`, so that we could not only link to the section with the error codes, but to a specific error code.

However, rendering the corresponding document via opf/openproject-website seems to always filter the HTML anchors.

PR #17545, opened 2025-01-07
Subject pattern improvements

NobodysNightmare:
# Ticket
There is no ticket directly associated to this cleanup.

Reducing the error potential when working with patterns. Right now one could believe that `replacement_patterns_defined? == has_subject_replacement_pattern?`.

By requiring to pass an attribute into the method, we can't ask the (arguably meaningless) question whether _any_ patterns are defined, but we are forced to ask whether a pattern is defined for a specific attribute. This prevents certain kinds of errors, while hopefully there will not be code that even needs to ask whether _any_ patterns are defined.

Also starting to more consistently use helpers that (probably) became available while developing the larger feature of automatic subject generation (e.g. `enabled_patterns`, which never returns `nil`) and removing some mock implementations.

### DIFF
--- a/app/services/work_packages/set_attributes_service.rb
+++ b/app/services/work_packages/set_attributes_service.rb
@@ -49,9 +49,7 @@ class WorkPackages::SetAttributesService < BaseServices::SetAttributes
   end
 
   def mark_templated_subject
-    return true unless work_package.type&.replacement_patterns_defined?
-
-    if work_package.type.patterns.all_enabled[:subject]
+    if work_package.type&.replacement_pattern_defined_for?(:subject)
       work_package.subject = "Templated by #{work_package.type.name}"
     end
   end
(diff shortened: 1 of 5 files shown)


app/services/work_packages/set_attributes_service.rb
```diff
@@ -49,9 +49,7 @@ def set_attributes(attributes)
   end
 
   def mark_templated_subject
-    return true unless work_package.type&.replacement_patterns_defined?
-
-    if work_package.type.patterns.all_enabled[:subject]
+    if work_package.type&.replacement_pattern_defined_for?(:subject)
```
NobodysNightmare, line 52: This is the only place where there is a (small?) semantical change.

Previously  the method could have returned `true`, `nil` or `"Templated by ..."`. Now it will only return `nil` or `"Templated by ..."`. I am not sure if this is a meaningful difference that will break something, but the return value definitely escapes from the `set_attributes` method...

NobodysNightmare: I just checked, because this made me feel uneasy. The return value from `set_attributes` is not checked, so this change is meaningless.

Though this also means the `return true` was meaningless before and a simple `return` would have been sufficient.

mereghost (approved):

PR #17546, opened 2025-01-07
Disable subject editing

NobodysNightmare:
# Ticket

This PR covers two tickets:
* https://community.openproject.org/projects/document-workflows-stream/work_packages/59911
* https://community.openproject.org/projects/document-workflows-stream/work_packages/59910

Block users from changing subjects that were automatically set because of the WP type's configuration.

## Screenshots
image

Changing the schema response for `writable` dynamically, based on whether the WP type has a subject pattern defined or not. This is already respected by our own table views and also allows third party clients to recognize that the subject can't be changed.

Admittedly, I was unsure about how "deep" I should define the attribute as non-writable. I've now chosen the `WorkpackageSchemaRepresenter`, where my other options would have been to do it on the work package schema (probably `SpecificWorkPackageSchema`, because that one knows the type) or maybe even the create and update contracts.

# Merge checklist

- [x] Added/updated tests
- [x] Tested Firefox

### DIFF
--- a/lib/api/v3/work_packages/schema/base_work_package_schema.rb
+++ b/lib/api/v3/work_packages/schema/base_work_package_schema.rb
@@ -53,6 +53,7 @@ module API
 
           def writable?(property)
             property = property.to_s
+            return false if property == "subject" && type&.replacement_pattern_defined_for?(:subject)
 
             # Special case for milestones + date property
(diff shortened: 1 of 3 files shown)


mereghost (approved):
LGTM. Nice work.

PR #17554, opened 2025-01-08
fix PermissionAdder to only add permissions that could also be assigned via the contract

klaustopher:
# Ticket
https://community.openproject.org/projects/gmbh/work_packages/60500

We are using the `PermissionAdder` to add permissions in a migration, when a role already has a given other permission.

In the contract to add permissions to a role, we have a check that we do not allow to add roles that have `require: :loggedin` to the _anonymous role_ and roles that have `require: :member` to the _anonymous role_ or the _non-member role_.

Added checks to the `PermissionAdder` that compares the `require:` attribute with the `builtin` of the role, so that we only add those permissions that make sense.

I also added tests for this and the existing behavior.

# Merge checklist

- [x] Added/updated tests

### DIFF
--- a/db/migrate/migration_utils/permission_adder.rb
+++ b/db/migrate/migration_utils/permission_adder.rb
@@ -32,17 +32,24 @@ module Migration
       module_function
 
       def add(having, add)
-        Role
+        role_scope = Role
           .joins(:role_permissions)
           .where(role_permissions: { permission: having.to_s })
           .references(:role_permissions)
-          .find_each do |role|
+
+        role_scope.find_each do |role|
           # Check if the add-permission already exists before adding
-          already_exists = RolePermission
-                             .exists?(role_id: role.id, permission: add.to_s)
-          unless already_exists
-            role.add_permission! add
-          end
+          next if RolePermission.exists?(role_id: role.id, permission: add.to_s)
+
+          added_permission = OpenProject::AccessControl.permission(add)
+
+          # we cannot add permissions that require a member to a non-member role
+          next if added_permission.require_member? && role.builtin == Role::BUILTIN_NON_MEMBER
+
+          # we cannot add permissions that require a logged in user to an anonymous role
+          next if added_permission.require_loggedin? && role.builtin == Role::BUILTIN_ANONYMOUS
+
+          role.add_permission! add
         end
       end
     end
--- a/db/migrate/migration_utils/permission_adder.rb
+++ b/db/migrate/migration_utils/permission_adder.rb
@@ -32,6 +32,8 @@ module Migration
       module_function
 
       def add(having, add)
+        added_permission = OpenProject::AccessControl.permission(add)
+
         role_scope = Role
           .joins(:role_permissions)
           .where(role_permissions: { permission: having.to_s })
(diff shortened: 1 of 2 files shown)


klaustopher:
Warning is noted ... As this is not a fix for the bug itself, it's fine IMHO to add it to dev

spec/migrations/permission_adder_spec.rb
```diff
@@ -0,0 +1,88 @@
...
+  let(:roles) { [role, non_member_role, anonymous_role] }
+
+  context "with a permission without special requirements" do
+    it "adds the permission to all roles" do
+      expect(role.permissions).not_to include(:view_project_attributes)
+      expect(non_member_role.permissions).not_to include(:view_project_attributes)
+      expect(anonymous_role.permissions).not_to include(:view_project_attributes)
+
+      described_class.add(:view_project, :view_project_attributes)
+      roles.map(&:reload)
```
NobodysNightmare: Nit: Since we are not using the return value, this should be `each`, shouldn't it?

```suggestion
      roles.each(&:reload)
```

spec/migrations/permission_adder_spec.rb
```diff
@@ -0,0 +1,88 @@
...
+
+RSpec.describe Migration::MigrationUtils::PermissionAdder, type: :model do # rubocop:disable RSpec/SpecFilePathFormat
+  let!(:role) { create(:project_role) }
+  let!(:non_member_role) { create(:non_member) }
+  let!(:anonymous_role) { create(:anonymous_role) }
+
+  let(:roles) { [role, non_member_role, anonymous_role] }
+
+  context "with a permission without special requirements" do
+    it "adds the permission to all roles" do
```
NobodysNightmare, line 42: Consider this comment a nit, because this is the first time we are adding specs to this class, but I think we could also test the behaviour of this `PermissionsAdder` to only add a permission where another permission already exists. I was at first confused when reading "to all roles", when we only want to add permissions to some roles usually.

In other words: Ideally there would be a role that doesn't have the `view_project` permission and thus also doesn't receive the `view_project_attributes`.

db/migrate/migration_utils/permission_adder.rb
```diff
@@ -32,17 +32,24 @@ module PermissionAdder
...
+        role_scope.find_each do |role|
           # Check if the add-permission already exists before adding
-          already_exists = RolePermission
-                             .exists?(role_id: role.id, permission: add.to_s)
-          unless already_exists
-            role.add_permission! add
-          end
+          next if RolePermission.exists?(role_id: role.id, permission: add.to_s)
+
+          added_permission = OpenProject::AccessControl.permission(add)
```
NobodysNightmare: Also nit: Shouldn't we move this assignment out of the loop, because `add` is not changed between iterations?

NobodysNightmare (approved):
Generally LGTM and nice that we added tests to a class that's used during migrations.

I left some optional comments.

spec/migrations/permission_adder_spec.rb
klaustopher: Copied this over from the old implementation. But makes sense

db/migrate/migration_utils/permission_adder.rb
klaustopher: It's a local lookup, so it does not really matter wher it's done but sure

PR #17555, opened 2025-01-08
[#60480] Bug: Having a custom logo for the PDF export leads to the timesheet export failing

as-op:
# Ticket
https://community.openproject.org/work_packages/60480

Fix the bug

Include forgotten module


ulferts (approved):
The fix itself looks good. When looking into the bug I was wondering how to write a spec from it and refrained from doing so given that I am not familiar with the code structure. I also accept the fix as is but it would be good to have a spec for this as well to avoid regressions.

as-op:
@ulferts agreed. I omitted a test as I thought the code and actual functionality would have been tested for one of the other PDF exports with a custom logo. 
But after having a look I couldn't find any. I will add a test to this PR.

PR #17556, opened 2025-01-08
bug/60458 The 'set reminder' action is available to users who are not signed in

akabiru:
# Ticket

https://community.openproject.org/work_packages/60458

Set reminder action should not be visible anonymous users. This PR performs the data correction by removing `manage_own_reminders` permissions from non member and anonymous roles.

## Screenshots

| Before | After |
|--------|--------|
| Screenshot 2025-01-08 at 3 04 51 PM | Screenshot 2025-01-08 at 3 05 49 PM |

17b94f6f55432d105cf5e7129ae1fdcc77d1bf4e introduced a seed migration that added the `manage_own_reminders` permission to roles with `view_project` permission via `Migration::MigrationUtils::PermissionAdder`.

However, the permission added did not correctly exclude non member and anonymous roles. This has been corrected in https://github.com/opf/openproject/pull/17554

# Merge checklist

- [x] Added/updated tests

### DIFF
--- a/db/migrate/20250108100511_remove_incorrect_manage_own_reminders_permission.rb
+++ b/db/migrate/20250108100511_remove_incorrect_manage_own_reminders_permission.rb
@@ -1,14 +1,31 @@
 class RemoveIncorrectManageOwnRemindersPermission < ActiveRecord::Migration[7.1]
   def up
     # Remove manage_own_reminders permission from non member and anonymous roles
-    RolePermission
-      .where(
-        role: Role.where(builtin: [Role::BUILTIN_NON_MEMBER, Role::BUILTIN_ANONYMOUS]),
-        permission: :manage_own_reminders
-      ).destroy_all
+    execute <<-SQL.squish
+      DELETE FROM role_permissions
+      WHERE role_id IN (
+        SELECT id FROM roles WHERE builtin IN (#{Role::BUILTIN_NON_MEMBER}, #{Role::BUILTIN_ANONYMOUS})
+      )
+      AND permission = 'manage_own_reminders'
+    SQL
 
-    # Remove all reminders created by anonymous user
-    Reminder.where(creator_id: AnonymousUser.first.id).destroy_all
+    # Remove all reminders created by anonymous user and cascade delete related records
+    execute <<-SQL.squish
+      WITH deleted_reminders AS (
+        DELETE FROM reminders
+        WHERE creator_id IN (
+          SELECT id FROM users WHERE type = 'AnonymousUser'
+        )
+        RETURNING id
+      ),
+      deleted_reminder_notifications AS (
+        DELETE FROM reminder_notifications
+        WHERE reminder_id IN (SELECT id FROM deleted_reminders)
+        RETURNING notification_id
+      )
+      DELETE FROM notifications
+      WHERE id IN (SELECT notification_id FROM deleted_reminder_notifications)
+    SQL
   end
 
   # No-op
--- a/db/migrate/20250108100511_remove_incorrect_manage_own_reminders_permission.rb
+++ b/db/migrate/20250108100511_remove_incorrect_manage_own_reminders_permission.rb
@@ -1,10 +1,14 @@
 class RemoveIncorrectManageOwnRemindersPermission < ActiveRecord::Migration[7.1]
   def up
+    # Remove manage_own_reminders permission from non member and anonymous roles
     RolePermission
       .where(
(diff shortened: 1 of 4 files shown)


db/migrate/20250108100511_remove_incorrect_manage_own_reminders_permission.rb
```diff
@@ -0,0 +1,16 @@
...
+  def up
+    # Remove manage_own_reminders permission from non member and anonymous roles
+    RolePermission
+      .where(
+        role: Role.where(builtin: [Role::BUILTIN_NON_MEMBER, Role::BUILTIN_ANONYMOUS]),
+        permission: :manage_own_reminders
+      ).destroy_all
+
+    # Remove all reminders created by anonymous user
+    Reminder.where(creator_id: AnonymousUser.first.id).destroy_all
```
akabiru: @machisuji I presume low risk in using AR models as these are unlikely to change and break migrations, plus the tests should help I think. Pls shout if you disagree 🍺

klaustopher (approved):

db/migrate/20250108100511_remove_incorrect_manage_own_reminders_permission.rb
machisuji: @akabiru Thanks for writing a test for the migration! And yet I'm afraid I'd still prefer it if you didn't use AR in the migration code. However unlikely the permissions are to change, they could. Or something could change with the Role model, or the Reminder.

This should be 2 fairly simple SQL statements. Could I please ask you to use SQL?

machisuji (changes requested):

db/migrate/20250108100511_remove_incorrect_manage_own_reminders_permission.rb
machisuji: P.S. And thanks for flagging me! <3

klaustopher: With the Reminders I could kind of agree, but with the role and permission stuff, I'd personally say that it is pretty set in stone. The test will tell us if we break things, which we will probably neve will

machisuji: Unless the test is running on the schema at the time of writing, not on the latest schema (which it does), it actually won't tell us if the migration breaks.

akabiru: Thanks for the review @machisuji, happy to update to raw SQL and thanks for providing context yet again :)

db/migrate/20250108100511_remove_incorrect_manage_own_reminders_permission.rb
```diff
@@ -0,0 +1,33 @@
+class RemoveIncorrectManageOwnRemindersPermission < ActiveRecord::Migration[7.1]
+  def up
+    # Remove manage_own_reminders permission from non member and anonymous roles
+    execute <<-SQL.squish
+      DELETE FROM role_permissions
+      WHERE role_id IN (
+        SELECT id FROM roles WHERE builtin IN (#{Role::BUILTIN_NON_MEMBER}, #{Role::BUILTIN_ANONYMOUS})
```
machisuji: @akabiru Thanks a lot for turning this into SQL! Looks pretty easy to understand.

About this line in particular, I know it's pretty much guaranteed not to change. I know it's tedious when I nitpick like this. But could you please also get rid of the used constants and just use their values directly (non member => 1 and anonymous => 2)? We might as well go all the way here.

machisuji (changes requested):

db/migrate/20250108100511_remove_incorrect_manage_own_reminders_permission.rb
akabiru: @machisuji , I do see a risk in hard-coding these if the enums are ever reassigned for any reason. It would be more straightforward to find all references of these "constants"- which by definition should never change. But if we use the actual values then I worry this can fail silently.

akabiru: Ok, I think I found a sensible middle ground- added references in comments. Not ideal IMO, but should be good enough.

https://github.com/opf/openproject/pull/17556/commits/0e77873f52fd0db2008339305d407991d0a085c0

machisuji (approved):
👍

PR #17560, opened 2025-01-08
Save pattern changes

NobodysNightmare:
# Ticket
https://community.openproject.org/projects/document-workflows-stream/work_packages/60498

Replace the current mock implementation of the subject configuration form with one that is actually storing the values we enter into it.

The `Type::BaseService` was extended to take the new parameters and update the type accordingly. To keep aligned with the structure of the `patterns` collection, I made sure that we'd only change the `subject` portion of that collection and not any other (future) elements in it, since it's already designed to carry more than just a subject pattern.

To simplify code that relies on `type.patterns`, I made sure that the collection would always be present and can't be `nil` anymore. This allows us to avoid nil-checks and avoids branching code in the form of "if patterns exists update, else create".

# Merge checklist

- [x] Updating implemented
- [x] Added/updated tests

### DIFF
--- a/app/contracts/types/base_contract.rb
+++ b/app/contracts/types/base_contract.rb
@@ -42,6 +42,7 @@ module Types
     attribute :project_ids
     attribute :description
     attribute :attribute_groups
+    attribute :patterns
 
     validate :validate_current_user_is_admin
     validate :validate_attribute_group_names
--- a/app/models/types/patterns/collection.rb
+++ b/app/models/types/patterns/collection.rb
@@ -34,6 +34,10 @@ module Types
       extend Dry::Monads[:result]
       private_class_method :new
 
+      def self.empty
+        new(patterns: {})
+      end
+
       def self.build(patterns:, contract: CollectionContract.new)
         contract.call(patterns).to_monad.fmap { |success| new(success.to_h) }
       rescue ArgumentError => e
@@ -46,12 +50,16 @@ module Types
         super(patterns: transformed)
       end
 
+      def subject
+        patterns[:subject]
+      end
+
       def all_enabled
         patterns.select { |_, pattern| pattern.enabled? }
       end
 
-      def [](value)
-        patterns[value]
+      def update_pattern(name, blueprint:, enabled:)
+        self.class.build(patterns: patterns.to_h.merge(name => { blueprint:, enabled: }))
       end
 
       def to_h
--- a/app/models/types/patterns/collection.rb
+++ b/app/models/types/patterns/collection.rb
@@ -51,7 +51,7 @@ module Types
       end
 
       def [](value)
-        patterns.fetch(value)
+        patterns[value]
       end
 
       def to_h
--- a/app/models/types/patterns/collection_type.rb
+++ b/app/models/types/patterns/collection_type.rb
@@ -36,7 +36,9 @@ module Types
       end
 
       def cast(value)
-        Collection.build(patterns: value).value_or { nil }
+        return value if value.is_a?(Collection)
+
+        Collection.build(patterns: value).value_or { Collection.empty }
       end
 
       def serialize(pattern)
@@ -46,7 +48,7 @@ module Types
       end
 
       def deserialize(value)
-        return if value.blank?
+        return Collection.empty if value.blank?
 
         data = YAML.safe_load(value)
         cast(data)
--- a/app/services/base_type_service.rb
+++ b/app/services/base_type_service.rb
@@ -1,3 +1,5 @@
+# frozen_string_literal: true
+
 #-- copyright
 # OpenProject is an open source project management software.
 # Copyright (C) the OpenProject GmbH
@@ -106,12 +108,9 @@ class BaseTypeService
   end
 
   def set_subject_pattern(params)
-    type.patterns = {
-      subject: {
-        blueprint: params[:subject_pattern],
-        enabled: params[:subject_configuration] == "auto"
-      }
-    }
+    type.patterns = type.patterns.update_pattern(:subject,
+                                                 blueprint: params[:subject_pattern],
+                                                 enabled: params[:subject_configuration] == "auto").value!
   end
 
   def parse_attribute_groups_params(params)
--- a/app/services/base_type_service.rb
+++ b/app/services/base_type_service.rb
@@ -71,7 +71,9 @@ class BaseTypeService
   def set_params_and_validate(params)
     # Only set attribute groups when it exists
     # (Regression #28400)
-    set_attribute_groups(params) unless params[:attribute_groups].nil?
+    set_attribute_groups(params) if params[:attribute_groups]
+
+    set_subject_pattern(params) if params[:subject_configuration]
 
     # This should go before `set_scalar_params` call to get the
     # project_ids, custom_field_ids diffs from the type and the params.
@@ -92,7 +94,7 @@ class BaseTypeService
   end
 
   def set_scalar_params(params)
-    type.attributes = params.except(:attribute_groups)
+    type.attributes = params.except(:attribute_groups, :subject_configuration, :subject_pattern)
   end
 
   def set_attribute_groups(params)
@@ -101,6 +103,15 @@ class BaseTypeService
     else
       type.attribute_groups = parse_attribute_groups_params(params)
     end
+  end
+
+  def set_subject_pattern(params)
+    type.patterns = {
+      subject: {
+        blueprint: params[:subject_pattern],
+        enabled: params[:subject_configuration] == "auto"
+      }
+    }
   end
 
   def parse_attribute_groups_params(params)
--- a/app/components/work_packages/types/subject_configuration_component.rb
+++ b/app/components/work_packages/types/subject_configuration_component.rb
@@ -36,8 +36,8 @@ module WorkPackages
 
       def form_options
         {
-          url: "https://example.com",
-          method: :put,
+          url: update_type_tab_path(id: model.id, tab: "subject_configuration"),
+          method: :patch,
           model:,
           data: {
             application_target: "dynamic",
--- a/app/components/work_packages/types/subject_configuration_component.rb
+++ b/app/components/work_packages/types/subject_configuration_component.rb
@@ -42,7 +42,8 @@ module WorkPackages
           data: {
             application_target: "dynamic",
             controller: "admin--subject-configuration",
-            admin__subject_configuration_hide_pattern_input_value: true
+            admin__subject_configuration_hide_pattern_input_value:
+              !WorkPackages::Types::SubjectConfigurationForm.has_pattern?(model)
           }
         }
       end
--- a/app/forms/work_packages/types/subject_configuration_form.rb
+++ b/app/forms/work_packages/types/subject_configuration_form.rb
@@ -58,7 +58,8 @@ module WorkPackages
 
         subject_form.group(data: { "admin--subject-configuration-target": "patternInput" }) do |toggleable_group|
           toggleable_group.text_field(
-            name: :pattern,
+            name: :subject_pattern,
+            value: subject_pattern,
             label: I18n.t("types.edit.subject_configuration.pattern.label"),
             caption: I18n.t("types.edit.subject_configuration.pattern.caption"),
             required: true,
@@ -73,6 +74,12 @@ module WorkPackages
 
       def has_pattern?
         self.class.has_pattern?(model)
+      end
+
+      def subject_pattern
+        return "" if model.patterns.nil? || model.patterns[:subject].nil?
+
+        model.patterns[:subject].blueprint
       end
     end
   end
--- a/app/forms/work_packages/types/subject_configuration_form.rb
+++ b/app/forms/work_packages/types/subject_configuration_form.rb
@@ -31,6 +31,12 @@
 module WorkPackages
   module Types
     class SubjectConfigurationForm < ApplicationForm
+      class << self
+        def has_pattern?(type)
+          type.replacement_pattern_defined_for?(:subject)
+        end
+      end
+
       form do |subject_form|
         subject_form.radio_button_group(name: :subject_configuration) do |group|
           group.radio_button(
@@ -66,7 +72,7 @@ module WorkPackages
       private
 
       def has_pattern?
-        model.replacement_pattern_defined_for?(:subject)
+        self.class.has_pattern?(model)
       end
     end
   end
--- a/app/forms/work_packages/types/subject_configuration_form.rb
+++ b/app/forms/work_packages/types/subject_configuration_form.rb
@@ -77,9 +77,9 @@ module WorkPackages
       end
 
       def subject_pattern
-        return "" if model.patterns.nil? || model.patterns[:subject].nil?
+        return "" if model.patterns.subject.nil?
 
-        model.patterns[:subject].blueprint
+        model.patterns.subject.blueprint
       end
     end
   end
--- a/app/models/permitted_params.rb
+++ b/app/models/permitted_params.rb
@@ -614,6 +614,8 @@ class PermittedParams
           :color_id,
           :default,
           :description,
+          :subject_configuration,
+          :subject_pattern,
           { project_ids: [] }
         ],
         user: %i(
--- a/app/models/type.rb
+++ b/app/models/type.rb
@@ -109,8 +109,6 @@ class Type < ApplicationRecord
   end
 
   def enabled_patterns
-    return {} if patterns.blank?
-
     patterns.all_enabled
   end
 
--- a/app/models/types/patterns/collection_contract.rb
+++ b/app/models/types/patterns/collection_contract.rb
@@ -32,7 +32,7 @@ module Types
   module Patterns
     class CollectionContract < Dry::Validation::Contract
       params do
-        required(:subject).hash do
+        optional(:subject).hash do
           required(:blueprint).filled(:string)
           required(:enabled).filled(:bool)
         end
--- a/spec/controllers/types_controller_spec.rb
+++ b/spec/controllers/types_controller_spec.rb
@@ -312,6 +312,57 @@ RSpec.describe TypesController do
           expect(Type.find_by(name: "My type").projects.count).to eq(0)
         end
       end
+
+      describe "when updating the subject pattern" do
+        let(:params) do
+          { "id" => type.id,
+            "type" => { subject_pattern: blueprint, subject_configuration: "auto" },
+            "tab" => "subject_configuration" }
+        end
+        let(:blueprint) { "I choose you {{assignee}}!" }
+
+        before do
+          patch :update, params:
+        end
+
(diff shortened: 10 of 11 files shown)


app/models/types/patterns/collection_type.rb
```diff
@@ -36,7 +36,7 @@ def assert_valid_value(value)
       end
 
       def cast(value)
-        Collection.build(patterns: value).value_or { nil }
+        Collection.build(patterns: value.to_h).value_or { Collection.null }
```
NobodysNightmare: Preemptively summoning @mereghost here: These are two changes we talked about in one line.

1. `value.to_h` ensures that we can perform assignments such as `type.patterns = type.patterns` without failing back to the null value (whichever that is)
    * you said that you might have preferred a type check here, feel free to suggest what you had in mind :)
2. Instead of returning `nil` we are now failing back to `Collection.null` in case of errors (also once more during `deserialize` below)
    * I don't see that we are violating any expectations set by [`ActiveRecord::Type::Value`](https://api.rubyonrails.org/classes/ActiveModel/Type/Value.html) here, but feel free to chime in
    * I'd still argue that it's fair to choose the best fitting nil value based on the usage scenario, e.g. I think it's fair that an attribute like an optional `start_date` falls back to `nil`, while another attribute (e.g. this one here) chooses a more fitting null value (especially for collection-like types, I try to avoid `nil` very strongly, because it's nearly never needed, when you can easily iterate over empty collections or check for emptiness instead of checking for `nil?`

app/models/types/patterns/collection.rb
```diff
@@ -32,8 +32,13 @@ module Types
   module Patterns
     Collection = Data.define(:patterns) do
       extend Dry::Monads[:result]
+      include Dry::Monads[:result]
```
mereghost: We needed the extend due to the rescue block returning a `Failure` as anything that relies on `.build` expects.

If you want to return the null value, it needs to come wrapped in a `Success`.

app/models/types/patterns/collection.rb
```diff
@@ -50,8 +55,12 @@ def all_enabled
         patterns.select { |_, pattern| pattern.enabled? }
       end
 
-      def [](value)
-        patterns.fetch(value)
+      def [](name)
+        patterns[name]
```
mereghost: I've used fetch as rather than returning `nil` on a missing key, it raises a `KeyError` not failing silently.

app/models/types/patterns/collection_type.rb
mereghost: Let me address the 2 points:

1. I'd say a type check is preferred as it is cheaper, and the point of this object is type casting. We don't need to cast a type that is what we expect.

2. I don't like that AR when unable to cast assigns `nil`, but that's the convention. I don't mind it, if it is properly documented as it kinda breaks the Principle of Least Surprise.

app/models/types/patterns/collection.rb
```diff
@@ -50,8 +55,12 @@ def all_enabled
...
+      end
+
+      def updated(name, pattern)
```
mereghost: WDYT of the name `update_pattern`? We need the `:enabled` value too.

app/contracts/types/base_contract.rb
```diff
@@ -42,6 +42,7 @@ def self.model
     attribute :project_ids
     attribute :description
     attribute :attribute_groups
+    attribute :patterns
```
mereghost, line 45: I've been playing with validating this. It is kinda involved:
1. Get all the available tokens for that type (impossible on a type before saving)
2. Iterate through the defined patterns comparing tokens and add errors for all unknown tokens.

For 1 we can ask the `TokenPropertyMapper` (or add a method to `Type`)

```ruby
Types::Patterns::TokenPropertyMapper.new.tokens_for_type(model).values.flat_map(&:keys) 
```

Part 2, we need a resolver (that knows how to parse the pattern) to do it

```ruby
def unknown_keys(all_token_keys)
  @tokens.map(&:key) - all_token_keys
end
```

So validation would look like:

```ruby
def validate_patterns
  # probably only after a save as we need the type id or we could change it to rely on `custom_field_ids` but we would miss some filters =/
  tokens = model.all_pattern_tokens
  
  model.patterns.each do |key, pattern|
    unknown = pattern.unknown_tokens(all_tokens)
    next if unknown.empty?
    
    model.errors.add(:patterns, "Pattern for #{key} has unknown attributes #{unknown.join(',') }")
  end
end
```

app/models/types/patterns/collection.rb
NobodysNightmare: As you can see in this PR, I am using it to be able to check whether a subject is defined (in a place where I don't care whether it's _disabled_). Also now that subject is optional, there are actually legitimate use cases where I ask for a certain key, but it can be missing.

On a higher level note: I am a fan of using fetch on hashes, because often an error is preferable than receiving nil.
However, I am critical about using it in a method called `[]`, because that's pretty surprising. Now our collection quacks like a duck, but when I come to feed it some bread it bites me in the finger! (in other words: It looks like the harmless hash accessor, but it behaves like the error raising hash accessor)

app/models/types/patterns/collection.rb
NobodysNightmare: The enabled value is part of the pattern. I.e. `pattern` is expected to be a hash in the form `{ enabled:, blueprint: }`.

`update_pattern` sounds good to me. I was honestly unsure about this method name and also about whether I should make it passive (`updated` vs `update`), because of the immutability. But to be fair, none of the immutable Ruby methods are in passive form (e.g. `merge`).

app/contracts/types/base_contract.rb
NobodysNightmare: To be honest, that's a validation that I wanted to leave out of the scope of this PR :)

However, what I want to include is probably the backend-side half of https://github.com/opf/openproject/pull/17506. I.e. checking whether the user is allowed to enable a pattern.

app/models/types/patterns/collection_type.rb
NobodysNightmare: Point 2 is where I still disagree pretty strongly. Subjectively I don't think that we create a huge surprise for anyone here. But I want to be able to rationally explain that, so let me try.

## This is about ensuring a required value which reduces surprise

I think I want my main argument to be, that the net amount of surprise reduces. We can now properly ensure a required value on the attribute, which eliminates the worst kind of surprise: "Oh, I didn't expect this thing to be nil".

It's not that there is someone expecting to call `t.patterns = "bob"` to ensure that patterns become `nil`. The point where we **do** create a surprise is when someone tried assigning `t.patterns = nil`, because also this is impossible now. So I agree, that this is a surprise, but also what the hell are you doing to assign nil to a required value?

Everyone else can't become surprised by this really, because they are not on the critical surprise path (which is a real term that I didn't just make up).

## Active Record does not always assign nil when I do wrong things

Since this argument is partially about us bending to the "bad conventions" of AR, I just tested doing stuff I shouldn't be doing to see how AR behaves. **It doesn't always assign nil!**

Let's start with this example, that's a lot like the case I see with `patterns`:

```ruby
pry(main)> wp = WorkPackage.last
...
pry(main)> wp.id = "@mereghost"
=> "@mereghost"
pry(main)> wp.id
=> 0
```

I tried assigning a randomly chosen string to a column that AR will **cast** to integer, but that column is **required** to have a value. And indeed AR didn't assign `nil`, but instead it chose to assign a default value of the target type.

Now to be fair, that's where the similarities end. For example when assigning a hash, the result will be `nil` as you said. So maybe someone used `to_i` without realizing that it doesn't crash and burn on non-numbers.
Or maybe someone didn't care about that, because there are no rules and expectations what happens when you put garbage in, so there is no convention, just conincidences.

Casting of booleans is also pretty... interesting... Whatever I put in, turns out to be true, which arguably makes sense, because everything that's not `false` or `nil` is _truthy_. But then when I pass `nil`, it doesn't become `false`, because nil is _falsy_, but instead `nil` is passed vanilla. So apparently AR casts everything very permissively, but has an exception built-in for nil.

## Conclusion

I am not sure about your conclusion, as I said I am fine with the current implementation. But honestly, want I wanted to achieve was to ensure that this is a required value and we don't have to perform `nil` checks all the time.

An alternative that I could think of: Set a default database value for the column that ensures we will not deserialize `nil` for existing types, but a null object. Then add a contract that ensures that we can't store invalid/nil data (structurally invalid, I don't care about the blueprint content). This would also ensure that there are no cases where `patterns` would be nil and my goal is achieved.

That said, I personally find the current way to ensure that much cleaner.

app/models/types/patterns/collection.rb
NobodysNightmare: I am pretty new to the concept of monads. Does it mean that as soon as I claim to be a monad all my methods need to respond in `Success` and `Failure` wrappers?

I thought it's fine to only do that for certain methods, because e.g. `self.null` has no way of failing, so it's a matter of always calling it like `Patterns.null.value!`, which looks like busy work to me.

(also there will be lots of methods defined that still don't abide to this... let's talk about `to_s` or `nil?` :D)

app/contracts/types/base_contract.rb
NobodysNightmare: Thinking about it. I'll create an implementation chunk for that. This PR is already big enough as it is right now.

app/models/types/patterns/collection.rb
mereghost: I"d say that it looks like a collection accessor, but this no Hash so I wouldn't have Hash expectations on it, as it isn't even Enumerable. =)

For all I care we could have both `[]` and `fetch`. :P

app/models/types/patterns/collection.rb
mereghost: Hmm... I'd pass the blueprint and enable rather than kind of a "options". The signature becomes clearer and what the possible side effects could be.

app/contracts/types/base_contract.rb
mereghost: It was left out as this was about updating the Type and it was tied to the glueing of FE and BE, since it goes through the contract and services.

app/models/types/patterns/collection.rb
NobodysNightmare: Agreed.

app/models/types/patterns/collection.rb
NobodysNightmare: > I"d say that it looks like a collection accessor, but this no Hash so I wouldn't have Hash expectations on it

Are you telling me this doesn't qualify for principle of least surprise? :P 

But in all seriousness, things that I could imagine:

* renaming `[]` to `fetch` and adding at least `key?` for the use case of checking presence
* replacing `[](value)` by `subject()`, which returns a pattern or `nil`
    * we make wrong calls like `patterns["potato"]` or even `patterns["subject"]` impossible, which is probably the main motivation of using `fetch`
    * at the same time we allow checking presence through the same accessor

The second one is my favorite. WDYT?

app/models/types/patterns/collection_type.rb
mereghost: Again, I'm not against it as I hate that AR likes to ignore parsing errors. :P
 
AR tends to be as permissible as possible: in your integer example passing a string `"123"` will convert it correctly to the number (thanks SQLite) but other strings will be coerced to 0. Passing an `Object.new` makes it `nil`. Passing a `Time` will result in a UNIX timestamp.

The whole point is to coerce into something that would make sense for that datatype, and since we had `subject` as required everything else would be invalid.

So let's go for it and see how it ends up 👍🏼 . Also, I'd like to you check for/think about a couple of things:

1. Ensure that loading a `Type` where `patterns` is `NULL` and saving it, should be a no-op.
2. What do you think of the renaming `.null` to `Collection.empty`?

NobodysNightmare: 1. I think we are fine:

```
pry(main)> Type.first.attributes_before_type_cast["patterns"]
=> nil
pry(main)> Type.first.changed?
=> false
pry(main)> Type.first.save!
  Type Load (0.9ms)  SELECT "types".* FROM "types" ORDER BY position ASC LIMIT $1  [["LIMIT", 1]]
  TRANSACTION (0.3ms)  BEGIN
  Type Exists? (0.6ms)  SELECT 1 AS one FROM "types" WHERE LOWER("types"."name") = LOWER($1) AND "types"."id" != $2 LIMIT $3  [["name", "Task"], ["id", 1], ["LIMIT", 1]]
  TRANSACTION (0.1ms)  COMMIT
=> true
```

There is some database access going on related to the `save` but as far as I can tell none is related to my change. Definitely there is no `UPDATE` call.

2. Good idea, done

app/models/types/patterns/collection.rb
NobodysNightmare: Resolving this thread. We talked about this IRL.

Outcome:
* The `include` is a left-over from using `Success`/`Failure` from an instance method and could be removed by now
* It's totally fine that `null`/`empty` returns a value directly, without a wrapper, when it's always succeeding

app/models/types/patterns/collection.rb
NobodysNightmare: After our IRL discussion I chose the second option from the ones above.

I didn't go for the monad approach we discussed, but if you want I can give it a try in a separate commit and see how it would look.

app/contracts/types/base_contract.rb
NobodysNightmare: As discussed IRL, let's consider this out of scope (for this PR only) for two reasons:

* Ability to merge partial progress and soft-unblock testing things
* Possibly there is already conflict in this w.r.t. the work that @Kharonus did, which also makes smaller chunks of work preferrable

app/models/types/patterns/collection.rb
```diff
@@ -46,12 +50,16 @@ def initialize(patterns:)
         super(patterns: transformed)
       end
 
+      def subject
+        patterns[:subject]
```
mereghost, line 54: 🟢 We would need to do `nil` checks here in case no subject pattern is defined.

Returning a `Result` could make it elegant by the client code going like

```ruby
def subject = patterns.fetch(:subject) { Failure() }

collection.subject.bind { |pattern| do_stuff }
```

app/services/base_type_service.rb
```diff
@@ -103,6 +107,12 @@ def set_attribute_groups(params)
     end
   end
 
+  def set_subject_pattern(params)
+    type.patterns = type.patterns.update_pattern(:subject,
```
mereghost, line 111: 🟢 WDYT of adding a method on `Type` that does this? It feels that we are kinda leaking the internals here.

mereghost (approved):
Most of my reviews go along the traffic light system:

🟢 Something nice or an idea that shouldn't block merging at all
🟡 A point I wish to discuss further, but can be addressed later
🔴 Merge-blocking rare occurrences

Nice work. I left a couple of comments/suggestions that you can safely ignore or deal later. =)

app/models/types/patterns/collection.rb
mereghost: I think it would simplify the outside code, left an example as a comment on the review.

app/services/base_type_service.rb
NobodysNightmare: I am pretty indifferent on that. I think it's fair for the `BaseTypeService` to be more tightly coupled to the `Type` than other classes in the system and in the end it's mostly an attribute assignment we perform here.

But I see your point as well. Will move it over.

app/models/types/patterns/collection.rb
NobodysNightmare: I will introduce this in a separate commit. However, two thoughts on your code proposal:

1. Wouldn't `Some(...)`/`None()` be preferable here? There is a subject pattern or there is none
2. In your example, I can see that in the error case we return a `Failure`, but as far as I can tell, we never return a `Success`, so I am pretty sure that the `bind` call would fail
    * I noticed because what I am about to commit looks way less elegant than your code proposal ^^

NobodysNightmare: General thought:

I agree that there is some elegance in writing

```ruby
model.patterns.subject.bind(&:blueprint).or("")
```

as opposed to

```ruby
return "" if model.patterns.subject.nil?

model.patterns.subject.blueprint
```

On the other hand, even pure Ruby would have allowed me to write:

```ruby
model.patterns.subject&.blueprint || ""
```

If I had to express a fear, it would be that Monads add another layer of things, that a new contributor has to learn and unwrap. `bind(&:blueprint).or("")` might become easy to read once you are used to monads, but at first it looks arcane.

I thought I'd write this thought down, before I am too deep in to see this myself (which is probably some time next week).

edit:

It also doesn't seem to mix well with RSpec's DSL...

```
pattern = Type.find(type.id).patterns.subject

# before
expect(pattern).to be_enabled

# after A
expect(pattern.value!).to be_enabled

# after B
pattern.bind { |p| expect(p).to be_enabled }
```

~~The diff of monadization can be explored in the [last commit of this PR](https://github.com/opf/openproject/pull/17560/commits/0ee93cf6801ca294324671e363c093ec99a3719d)~~

edit 2: The "elegant" version I mentioned above doesn't even work... Let me see how unelegant it has to get...

```ruby
model.patterns.subject.maybe(&:blueprint).or(Some("")).value!
```

This seems to do the trick (if we included `Dry::Monads[:maybe]` in the client as well), but it's really starting to get ugly and I'd go towards :yellow_circle:  for that last change... The idea was to improve readability of client code and that decreased IMO. For now I reverted those changes.

app/models/types/patterns/collection.rb
NobodysNightmare: Changed accordingly. I am going to resolve this discussion here, because we can continue in the other threads :)

app/services/base_type_service.rb
NobodysNightmare: I am now deciding against this change as well.

Rationale: While moving the method, I noticed that I can't move it as-is, because the whole `params` interaction doesn't belong to the type (`Type` should not know how to parse `params`).

However, it felt wrong keep the call to `value!` inside the `Type` implementation, because now we'd not offer the caller in the service class any option to check the result of the call in a non-rescue way.

I thought the solution would be easy using monads and I did this:

```ruby
  #type.rb
  def set_subject_pattern(blueprint:, enabled:)
    patterns.update_pattern(:subject, blueprint:, enabled:).bind do |value|
      self.patterns = value
    end
  end

  # base_type_service.rb
  def set_subject_pattern(params)
    type.set_subject_pattern(blueprint: params[:subject_pattern], enabled: params[:subject_configuration] == "auto")
        .or { |e| raise e }
  end
```

I was happy with how that looked. I was happy with the separation of `type.rb` working on the happy path, while the service would make sure the sad path was also covered.

However, it didn't work, because `bind` doesn't return a monad, but rather an unwrapped value, so `bind {...}.or` doesn't work without rewrapping inside the bind-block. Which would mean that I'd have to include dry monads in `Type` and I wasn't in a mood to do that.

PR #17561, opened 2025-01-09
[60512] Prevent child relation creation for milestones

cbliard:
And correctly handle side-effects when adding a child relation: the parent was set without calling the service, meaning side-effects like rescheduling or updating ancestors properties was not done.

# Ticket

https://community.openproject.org/wp/60512

* Prevent creating a child relation from relations tab for a milestone
* Add all checks and side-effects for child creation by doing the service call (that was missing)

## Screenshots

Remove the menu entry for milestone work package type of course, but also change controller action to err when adding a child for a milestone.

# Merge checklist

- [x] Added/updated tests

### DIFF
--- a/app/controllers/work_package_children_controller.rb
+++ b/app/controllers/work_package_children_controller.rb
@@ -49,9 +49,9 @@ class WorkPackageChildrenController < ApplicationController
     target_work_package_id = params[:work_package][:id]
     target_child_work_package = WorkPackage.find(target_work_package_id)
 
-    target_child_work_package.parent = @work_package
-
-    if target_child_work_package.save
+    service_result = WorkPackages::UpdateService.new(user: current_user, model: target_child_work_package)
+                                                .call(parent: @work_package)
+    if service_result.success?
       @children = @work_package.children.visible
       @relations = @work_package.relations.visible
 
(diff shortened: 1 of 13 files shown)


app/controllers/work_package_children_controller.rb
```diff
@@ -49,9 +49,9 @@ def create
     target_work_package_id = params[:work_package][:id]
     target_child_work_package = WorkPackage.find(target_work_package_id)
 
-    target_child_work_package.parent = @work_package
-
-    if target_child_work_package.save
+    service_result = WorkPackages::UpdateService.new(user: current_user, model: target_child_work_package)
+                                                .call(parent: @work_package)
+    if service_result.success?
```
ulferts: The mistake that existed here before was repeated in the destroy action. In there, the service should be used as well. Otherwise a parent with two children with dates will keep the min/max of both children even if one is later on removed as a child.
So your hunch that there might be more amis was correct.

ulferts (changes requested):
The fix itself is good but I uncovered a similar bug next to the fixed issue which should be fixed at the same time.

app/controllers/work_package_children_controller.rb
cbliard: Good catch.

cbliard: Fixed

ulferts (approved):
👍

PR #17567, opened 2025-01-09
Bug/60540 adding a work package as a child via the activity tab broken

ulferts:
# Ticket

https://community.openproject.org/wp/60540

While working on this, the original error has been fixed but without correctly fixing the scrolling behaviour.

Fix scrolling to the newly added child work package which got broken in when [merging 15.1 into 15.2](https://github.com/opf/openproject/commit/de565583b2c81691c2eba71e5636004a2f72dfe6#diff-f7a3f6ceb8a1b0173a5e80dea088a0937596ae4ac26179f9182dbddab832dcc5L49-L52). This fails because it is now comparing an int to a string. This is fixed in ec4a8b2465254bdb16d530a6471f05a6d8ab45bc. The second commit tries to improve the expressiveness of the scrolling functionality and is not related to the bug fix.

### DIFF
--- a/app/components/work_package_relations_tab/index_component.rb
+++ b/app/components/work_package_relations_tab/index_component.rb
@@ -1,5 +1,11 @@
 # frozen_string_literal: true
 
+# Component for rendering the relations tab content of a work package
+#
+# This includes:
+# - Controls for adding new relations if the user has permission
+# - Related work packages grouped by relation type (follows, precedes, blocks, etc.)
+# - Child work packages
 class WorkPackageRelationsTab::IndexComponent < ApplicationComponent
   FRAME_ID = "work-package-relations-tab-content"
   NEW_RELATION_ACTION_MENU = "new-relation-action-menu"
@@ -11,6 +17,12 @@ class WorkPackageRelationsTab::IndexComponent < ApplicationComponent
 
   attr_reader :work_package, :relations, :children, :directionally_aware_grouped_relations, :relation_to_scroll_to
 
+  # Initialize the component with required data
+  #
+  # @param work_package [WorkPackage] The work package whose relations are being displayed
+  # @param relations [Array<Relation>] The relations associated with this work package
+  # @param children [Array<WorkPackage>] Child work packages
+  # @param relation_to_scroll_to [Relation, WorkPackage, nil] Optional relation or child to scroll to when rendering
   def initialize(work_package:, relations:, children:, relation_to_scroll_to: nil)
     super()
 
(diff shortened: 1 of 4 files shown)


app/controllers/work_package_children_relations_controller.rb
```diff
@@ -46,7 +46,7 @@ def create
     child = WorkPackage.find(params[:work_package][:id])
     service_result = set_relation(child:, parent: @work_package)
 
-    respond_with_relations_tab_update(service_result, scroll_to_id: child.id)
+    respond_with_relations_tab_update(service_result, relation_to_scroll_to: service_result.result)
```
cbliard, line 49: I found using `child` directly easier to reason about, but I'm ok with this choice.

app/components/work_package_relations_tab/index_component.rb
```diff
@@ -74,24 +86,31 @@ def render_header(border_box, title, items)
...
+        "work-packages--relations-tab--scroll-target": "scrollToRow"
+      }
+    end
+  end
+
+  def scroll_to?(item)
+    relation_to_scroll_to \
+      && item.id == relation_to_scroll_to.id \
+      && item.instance_of?(relation_to_scroll_to.class)
+  end
```
cbliard, line 112: As `relation_to_scroll_to` could be either a `Relation` or a `WorkPackage` (child relation), the class had to be checked too to prevent scrolling to a wrong element because of a similar id (unlikely but possible).

I added a commit directly to fix it (and extract methods).

app/components/work_package_relations_tab/index_component.rb
```diff
@@ -9,16 +15,22 @@ class WorkPackageRelationsTab::IndexComponent < ApplicationComponent
...
+  attr_reader :work_package, :relations, :children, :directionally_aware_grouped_relations, :relation_to_scroll_to
 
-  def initialize(work_package:, relations:, children:, scroll_to_id: nil)
+  # Initialize the component with required data
+  #
+  # @param work_package [WorkPackage] The work package whose relations are being displayed
+  # @param relations [Array<Relation>] The relations associated with this work package
+  # @param children [Array<WorkPackage>] Child work packages
+  # @param relation_to_scroll_to [Relation, WorkPackage, nil] Optional relation or child to scroll to when rendering
+  def initialize(work_package:, relations:, children:, relation_to_scroll_to: nil)
```
cbliard, line 26: Added this doc to emphasize that `relation_to_scroll_to` is either a `Relation` or a `WorkPackage` (child relation).

cbliard (approved):
Nice catch 👏 

I fixed one additional issue. I'll merge once the tests pass.

PR #17569, opened 2025-01-09
[#59885] Cancellation dialogs for one-time and recurring meetings

myabc:
# Ticket

https://community.openproject.org/wp/59885

- This PR targets the meetings module.
- Replaces standard browser `confirm` dialogs used to confirm deletion with new Primerized `DangerConfirmationDialog` component.

### 1. One-time (`StructuredMeeting`)

image

**Invoked by:**

- meeting actions on meeting show page (URL `http://localhost:5000/meetings/35`)
- OR row actions on meetings index page (URL `http://localhost:5000/meetings?upcoming=false`)

Screenshot 2025-01-30 at 17 41 05
image

### 2. Instantiated occurrences (`StructuredMeeting` with associated `RecurringMeeting`)

image

**Invoked by:**

- meeting actions on meeting show page (URL `http://localhost:5000/projects/brand-new-project/meetings/32`)
- OR row actions on meetings index page (URL `http://localhost:5000/meetings`)
- OR row actions on recurring meeting show page (URL `http://localhost:5000/projects/brand-new-project/recurring_meetings/7`)

image
Screenshot 2025-01-30 at 17 48 15
image

#### with either zero or infinite meetings in the series

Screenshot 2025-02-03 at 13 39 31

#### with one remaining meeting in the series

Screenshot 2025-02-03 at 13 42 11

#### with more than one remaining meeting in the series

Screenshot 2025-02-03 at 13 40 06

**Invoked by:**

- actions on recurring meeting show page (URL `http://localhost:5000/projects/brand-new-project/recurring_meetings/7`)

image

### 4. Scheduled occurrences (`ScheduledMeeting`)

image

**Invoked by:**

- row actions on recurring meeting show page (URL `http://localhost:5000/projects/brand-new-project/recurring_meetings/7`)

image

**This PR implements three DeleteDialog components, along with three accompanying `delete_*dialog` Controller actions.**  

- `Meetings::DeleteDialogComponent` (see screenshots 1 & 2 above) this renders different text depending on whether the meeting is a one-off meeting or an instantiated occurrence.
- `RecurringMeetings::DeleteDialogComponent` (see screenshot 3)
- `RecurringMeetings::DeleteScheduledDialogComponent` (see screenshot 4)

`Meetings::DeleteDialogComponent` could also have been implemented as two separate dialogs. My rule of thumb however was to stick to a one-to-one mapping with controller actions. (FWIW I also considered the opposite approach - having one or two polymorphic delete dialog implementations for the entire meetings module - but quickly discarded for reasons of complexity).

A significant part of work on this PR has been to ensure that the end user stays within either the global or project context when deleting a meeting. 

I configured routes for the `delete_*dialog` and `destroy` Controller actions within the top-level `meetings` resource and the nested `project/meetings` resource. An alternative strategy might have been to define a single global route for `delete_*dialog`/`destroy` actions and somehow pass in a redirect URL on dialog creation.

Ultimately though, I felt that creating separate routes was the simplest (and most Rails-like) strategy. That said, a large number of the tests that I wrote concern basically just sanity-check that project or global-scope is not lost - either on invoking the dialog, or on form submission.

:warning: This PR (specifically [96308af](https://github.com/opf/openproject/pull/17569/commits/96308af5121445d7b916360fe9a113c55303632c)) _may fix_ scenarios where a user is switched from global scope to project-scope. **However this needs some further manual testing.** While I've made sure the current spec suite passes, I felt that adding additional request specs/features for non-deletion actions was out-of-scope for this (already long-running) PR. _I'd recommend we create a separate work package to look into this._

This Pull Request adds basic specs for the `DangerDialog` components as well as the related `RowComponents`. Again, these specs could probably be DRY-ed up (sanity-checking project vs global scope might be something we could do with `shared_examples`).

### Danger Dialog 

There is a [draft PR to implement Danger Dialog warning variant, without second step confirmation](https://github.com/opf/primer_view_components/pull/234).

Taking into account the UX guidelines in #17789, my understanding is that the following end-user presented dialogs (referencing screenshots above) will be candidates for the simpler "warning" variant of the Danger Dialog:

1. One-time (`StructuredMeeting`)
2. Instantiated occurrences (`StructuredMeeting` with associated `RecurringMeeting`)
4. Scheduled occurrences (`ScheduledMeeting`)

We will only require the second-step confirmation check box for this dialog:

3. Meeting series (`RecurringMeeting`)

This PR can go ahead and be merged as-is, but given that there will be a forthcoming "warning" variant, I did not customise the check box label text for 1, 2 & 4. I also kept the confirmation button text as the default "Delete permanently"**, except in the case of 4, where we always refer to the action as "Cancel" and not "Delete".

@psatyal Please let me know if this looks ok to you?

** I plan to change the default button text from "Delete permanently" to "Delete" for the warning variant of the Danger Dialog / @HDinger 

# Merge checklist

- [X] Delete structured meetings
- [X] Delete recurring meetings
- [X] Delete scheduled recurring meetings
- [X] Added/updated tests


modules/meeting/spec/features/recurring_meetings/recurring_meeting_crud_spec.rb
```diff
@@ -89,13 +89,17 @@
...
+    show_page.delete_meeting_series
+    show_page.in_delete_dialog do
+      page.check "I understand that this deletion cannot be reversed"
 
-    accept_confirm(I18n.t("text_are_you_sure")) do
-      click_on "Delete meeting series"
+      click_on "Delete permanently"
     end
 
     expect(page).to have_current_path meetings_path # check path
```
myabc: Hi @oliverguenther. I saw you added this comment. Is that a `TODO` (i.e. do we still need to check if this is the correct path), or just an explanation of what the line does? - it seems unclear to me if we always want to forward to the global meetings path after meeting deletion

myabc: `RecurringMeetinsController#destroy` redirects to `polymorphic_path([@project, :meetings])`, so either the assertion or the implementation must be wrong? 😕

modules/meeting/app/controllers/recurring_meetings_controller.rb
```diff
@@ -130,6 +132,13 @@ def update
     end
   end
 
+  def delete_dialog
+    respond_with_dialog RecurringMeetings::DeleteDialogComponent.new(
+      meeting: @recurring_meeting,
+      project: @project
```
myabc, line 166: (I based my `delete_dialog` controller action on `details_dialog`) ❓  Not sure we need to pass in project here? - see other comments on nested (project scoped) routes.

modules/meeting/app/components/meetings/delete_dialog_component.rb
```diff
@@ -0,0 +1,67 @@
...
+        I18n.t("text_are_you_sure")
+      end
+    end
+
+    def recurring?
+      @meeting.is_a?(RecurringMeeting)
+    end
+
+    def recurring_meeting
+      return if recurring?
```
myabc: I'm not sure if this check is necessary (need to familiarise myself properly with `Meeting`/`StructuredMeeting`/`RecurringMeeting` domain models and their relationships)

config/locales/en.yml
```diff
@@ -3938,6 +3939,7 @@ en:
     For more information on what the check provides, what data is needed to provide available updates, and how to disable this check, please visit <a href="%{more_info_url}">the configuration documentation</a>.
   text_own_membership_delete_confirmation: "You are about to remove some or all of your permissions and may no longer be able to edit this project after that.\nAre you sure you want to continue?"
   text_plugin_assets_writable: "Plugin assets directory writable"
+  text_permanent_delete_confirmation_checkbox_label: "I understand that this deletion cannot be reversed"
```
myabc: @HDinger I think this belongs in the global I18n strings. (Not sure though if we want to reconsider in specifying a `DangerDialogComponent` default checkbox label?)

modules/meeting/app/components/recurring_meetings/delete_dialog_component.html.erb
```diff
@@ -0,0 +1,46 @@
...
+See COPYRIGHT and LICENSE files for more details.
+
+++#%>
+
+<%= 
+  render(Primer::OpenProject::DangerConfirmationDialog.new(
+    id:,
+    title:,
+    form_arguments: {
+      action: polymorphic_path([@project, @meeting]),
```
myabc: This was lifted from here: https://github.com/opf/openproject/pull/17569/files#diff-d10668a4a54d66743190f51acff89d1022705eeee43b368c366e7cabf564526aL42 However there is no project-scoped `destroy` route.

config/locales/en.yml
oliverguenther: I agree and I try to put generalizable strings into the main en.ml so they can be reused.

modules/meeting/app/components/meetings/delete_dialog_component.rb
oliverguenther: It's a bit convoluted due to the existence of the old meeting.

`Meeting` is the deprecated type with two text areas stored separately from another (MeetingAgenda and MeetingMinute).

`StructuredMeeting` is the new primerized meeting with multiple primerized

`RecurringMeeting` is a meeting series, which consists of multiple `StructuredMeeting`. Every `StructuredMeeting`, which is connected through the `belongs_to :recurring_meeting`.

On top of that, there are also `ScheduledMeeting`, which is a record of a of a meeting that has been rescheduled / moved its actual date (probably not relevant for this PR), or if an occurrence has been cancelled.

modules/meeting/app/components/meetings/delete_dialog_component.rb
```diff
@@ -0,0 +1,67 @@
...
+    end
+
+    private
+
+    def id = "delete-meeting-dialog"
+
+    def title = I18n.t("meeting.delete_dialog.title")
+
+    def confirmation_message
+      if recurring_meeting.present?
```
oliverguenther, line 50: I would assume this message has three cases:

1. The meeting is a one-off StructuredMeeting (it is neither a `RecurringMeeting` series nor does it have a  `recurring_meeting` assocation).
2. The meeting is part of a meeting series. (It is a `StructuredMeeting`, but has an assocation `recurring_meeting` set. The message would contain something like "This meeting is a part of the series XYZ. This will only delete this occurrence"
4. It is a meeting series object itself (it is a `RecurringMeeting`). "Deleting XYZ series will delete all occurrences..."

modules/meeting/app/components/meetings/delete_dialog_component.rb
```diff
@@ -0,0 +1,67 @@
...
+        I18n.t(:label_recurring_meeting_delete_confirmation, name: recurring_meeting.title)
+      else
+        I18n.t("text_are_you_sure")
+      end
+    end
+
+    def recurring?
```
oliverguenther: The naming is really hard to get right. In code, recurring meeting could mean two things:

1. It is a meeting series consisting of multiple meeting occurrences
2. It is the occurrence of a meeting series

Do you have suggestions to better separate this in English? In German this works better (_Terminserie_ vs _Wiederkehrende Besprechung_).

modules/meeting/app/components/recurring_meetings/delete_dialog_component.html.erb
oliverguenther: I guess having only the global destroy makes sense, as long as you do a `redirect_back` as to not accidentally leave the project scope.

We have this issue in a few places in meeting, where some actions in a project context incorrectly redirect you to the global level. This results in you losing your project sidebar, etc.

modules/meeting/app/controllers/recurring_meetings_controller.rb
oliverguenther: If you don't need to create routes depending on the current project context, then you can omit  `@project`.

modules/meeting/spec/features/recurring_meetings/recurring_meeting_crud_spec.rb
oliverguenther: We want to keep the current project context, if you are in one.

So if you are looking at `/projects/my-identifier/meetings/1234`, and start deleting from there, I would expect you to be redirected to `/projects/my-identifier/meetings`.

Likewise when destroying the meeting from the global meeting page `/meetings/1234`, you should be redirected to `/meetings`. It could be that we're missing the correct implementation here, or that the spec is incorrect. I will double check the current behavior

modules/meeting/app/models/structured_meeting.rb
```diff
@@ -34,6 +34,10 @@ class StructuredMeeting < Meeting
            inverse_of: :meeting
   accepts_nested_attributes_for :agenda_items
 
+  def self.model_name
+    @model_name ||= ActiveModel::Name.new(self, nil, "Meeting")
+  end
```
myabc: ⚠️ This may have unintended consequences.

This ensures `polymorphic_path` resolves the correct route when passed a `StructuredMeeting` (i.e. `ActiveModel::Naming.singular_route_key(StructuredMeeting) => "meeting"`).

config/locales/en.yml
HDinger: Hi @myabc 

@Kharonus added the DangerDialogs to the hierarchy CF in the scope of https://github.com/opf/openproject/pull/17717. He there added the exact same strings for the button and the checkbox label. So you can rebase your PR and remove the strings here.

myabc: @HDinger great, will do!

modules/meeting/app/components/meetings/delete_dialog_component.rb
```diff
@@ -0,0 +1,67 @@
...
+      @meeting.is_a?(RecurringMeeting)
```
myabc: This component doesn't need to be polymorphic - `@meeting` should always be a `Meeting` or a `StrucuturedMeeting`.

modules/meeting/app/components/meetings/delete_dialog_component.rb
myabc: As I've modelled it, `Meetings::DialogComponent` will always be called with a `Meeting` or `StructuredMeeting`.

modules/meeting/app/components/recurring_meetings/delete_dialog_component.html.erb
myabc: > I guess having only the global destroy makes sense, as long as you do a redirect_back as to not accidentally leave the project scope.

I ended up going with the opposite approach - have documented why in the PR description.

myabc:
@oliverguenther I noticed that [#17762](https://github.com/opf/openproject/pull/17762)) was merged a couple days ago. It looks like we're structuring our I18n strings in slightly different ways - e.g `text_meeting_series_end_early_heading` vs `recurring_meeting/delete_dialog` with nested keys. I don't feel particularly strongly about which is best but it would be good to agree on a convention.

oliverguenther:
Deletion fails for meetings that are `end_after=never` 
Bildschirmfoto 2025-02-03 um 14 21 47

psatyal (changes requested):
Hi @myabc,


I based my review on the screenshots instead of the code (since I don't trust my ability to read code erb syntax), but they look good to me. I would only make on minor modification:

> ### One-time (`StructuredMeeting`)
> Screenshot 2025-01-30 at 17 41 13
> 


For this first case (one-time, basic `StructuredMeeting`), I would change the second line to the default text so it reads:

> **Delete this meeting?**  
> This action is not reversible. Please proceed with caution.
> {checkbox} I understand that this deletion cannot be reversed/
> [Cancel] [Delete permanently]

This way, when we later go back to using the non-confirmation variant, the copy for that second line does not have to be updated.

Cheers
Parimal

myabc:
> Deletion fails for meetings that are `end_after=never` Bildschirmfoto 2025-02-03 um 14 21 47

@oliverguenther - thanks for pointing this out. Fixed.

oliverguenther (approved):
LGTM now 👍

myabc:
@psatyal Cheers for the feedback!

> I based my review on the screenshots instead of the code (since I don't trust my ability to read code erb syntax), but they look good to me. I would only make on minor modification:

No worries! - thanks for looking at the screenshots over here. I was going to cross-post to OP, but it's easier with a single source of truth.
 
> > ### One-time (`StructuredMeeting`)
> 
> For this first case (one-time, basic `StructuredMeeting`), I would change the second line to the default text so it reads:
> 
> > **Delete this meeting?**
> > This action is not reversible. Please proceed with caution.
> > {checkbox} I understand that this deletion cannot be reversed/
> > [Cancel] [Delete permanently]
> 
> This way, when we later go back to using the non-confirmation variant, the copy for that second line does not have to be updated.
> 

image

Done!

psatyal (approved):
Thanks! Looks good to me.

modules/meeting/app/components/recurring_meetings/delete_dialog_component.rb
```diff
@@ -0,0 +1,55 @@
...
+    private
+
+    def id = "delete-recurring-meeting-dialog"
+
+    def title = I18n.t("recurring_meeting.delete_dialog.title")
+
+    def confirmation_message
+      t("recurring_meeting.delete_dialog.confirmation_message_html",
+        title: @recurring_meeting.title,
+        count: @recurring_meeting.remaining_occurrences&.count || 0) # actually ♾️, but we show same message as for 0
```
myabc, line 52: for @klaustopher 😄

modules/meeting/config/locales/en.yml
```diff
@@ -284,6 +290,30 @@ en:
         The following meetings are planned in the recurring meeting schedule but are not open yet. 
         Every time a planned meeting starts, the next one will automatically be opened for you. 
         You can also open planned meetings manually to import the template and start editing the agenda.
+    delete_dialog:
+      title: "Delete meeting series"
+      heading: "Permanently delete this meeting series?"
+      confirmation_message_html:
+        zero: >
+          Deleting <strong>%{title}</strong> will delete any meetings in this series.
+          This action is not reversible. Please proceed with caution.
```
myabc, line 299: FYI @psatyal You might also be interested in this change (I've updated the visuals above under **3. Meeting series (`RecurringMeeting`)**)

We will display a slightly different confirmation message depending on the number of remaining meetings. Because of technical constraints with Rails pluralisation rules, We're also treating zero and infinite as the same for now.

I just thought you might be interested in why conflating these two scenarios doesn't work so well in other European languages:

> with regards to translating: "any" is vaguer than "all" in English (implying there 
> might not be any at all), but in German or French you'd probably use alle or toutes for both.
> 
> _löscht alle Meetings dieser Serie_
> _supprimera toutes les réunions de cette séries_

(cross-posting from an Element developer channel that I don't think you're in)

modules/meeting/spec/features/recurring_meetings/recurring_meeting_crud_spec.rb
```diff
@@ -90,7 +90,7 @@
 
     show_page.delete_meeting_series
     show_page.within_modal "Delete meeting series" do
-      page.check "I understand that this deletion cannot be reversed"
+      check "I understand that this deletion cannot be reversed", allow_label_click: true
```
cbliard, line 92: If all check boxes will need this, maybe it would be better to enable it globally with [`automatic_label_click`](https://www.rubydoc.info/gems/capybara/3.12.0/Capybara#configure-class_method)
```ruby
Capybara.configure do |config|
  config.automatic_label_click = true
end
```

myabc: @cbliard yes, I was thinking we could try enabling globally.

PR #17575, opened 2025-01-10
Also check static links for consistency

NobodysNightmare:
Previously we only checked internal linking inside the documentation for consistency. Now there is also a basic test to see whether links from OpenProject into the documentation are pointing to a valid page.

I used this opportunity to move the static links into a Yaml file. This made it easier to load the list of static links for the check as well and it removed a long data object from the Ruby code.

Ensure that we only link out to existing documentation pages from the OpenProject application.

I extended the existing link checker for our documentation, because it already built a map of all docs-pages. To make sharing the list of static links easier, I extracted them from Ruby code into a Yaml file.

# Screenshots

GitHub annotating broken link


as-op (approved):
Thanks a lot for this PR! Works like a charm.

PR #17591, opened 2025-01-13
[60423] Changed error message for when a folder can not be created

apfohl:
# Ticket
https://community.openproject.org/projects/document-workflows-stream/work_packages/60423/activity

Make the experince better for the user if the folder they wish to create can not be created.

## Screenshots
image

I chose to change the error text, as changing the error handling itself need more careful thought. I suggest do a code maintanance to rethink the error handling.


brunopagno (approved):
Yeah, I think this is a quick win. Nice simplification :bow:

mereghost (approved):

PR #17593, opened 2025-01-13
Introduce relations tab mediator

cbliard:
Introduce a mediator for the relations tab index component rendering. It is responsible for providing the relations and children from a work package for the relations tab. 

This moves the responsibility of relations and children visibility in one central place: the component asks for relations and children, and the mediator replies with the relations and children the user is allowed to see.


ulferts (approved):
👍

cbliard:
The failing test is passing locally. Merging.

PR #17595, opened 2025-01-13
Feature/59178 journalize changes to project stages and gates

toy:
# Ticket
[OP#59178](https://community.openproject.org/wp/59178)

* Journal activation/deactivation of project life cycle steps and setting/changing/removing their dates.
* Show those journals when «Project details» is selected in activities event type filter

* Not using feature flag, as both activation and editing of project life cycle steps are already hidden behind it
* Not creating initial journals, as project stages and gates should not yet be used by anyone
* Renamed event type itself from `project_attributes` to `project_details`, as project attributes is the name for project custom fields. Name `project_details` is also used in few migrations, but that is much less then usage of `project_attributes`. Created a compatibility alias for users that saved the link or have event types stored in session. [OP#61285](https://community.openproject.org/wp/61285) to cleanup.

# Merge checklist

- [x] Added/updated tests


app/models/activities/base_activity_provider.rb
```diff
@@ -187,6 +187,7 @@ def event_params(event_data)
     params
   rescue StandardError => e
     Rails.logger.error "Failed to deduce event params for #{event_data.inspect}: #{e}"
+    {}
```
dombesz: Was this a bug?

app/models/activities/fetcher.rb
```diff
@@ -72,22 +72,20 @@ def events(from: nil, to: nil, limit: nil)
...
     protected
 
-    # Sets the scope
     # Argument can be :all, :default or an array of event types
-    def scope=(scope)
+    def parse_scope(scope)
       case scope
-      when :all
-        @scope = event_types
+      when :all, false, nil
```
dombesz: Not sure about these values, should we also include other possible values, like `true`?

app/models/journal/project_life_cycle_step_journal.rb
```diff
@@ -0,0 +1,41 @@
...
+
+class Journal::ProjectLifeCycleStepJournal < Journal::AssociatedJournal
+  self.table_name = "project_life_cycle_step_journals"
+
+  belongs_to :life_cycle_step, class_name: "Project::LifeCycleStep"
+
+  alias_attribute :date, :start_date
+
+  def date_range
+    start_date..end_date if start_date || end_date
```
dombesz, line 39: Is this a typo, should it be `if start_date && end_date` ? This will generate open ended ranges.

app/services/journals/create_service/association.rb
```diff
@@ -0,0 +1,55 @@
...
+    include Helpers
+
+    def self.for(journable)
+      [
+        AgendaItemable,
+        Attachable,
+        Customizable,
+        ProjectLifeCycleStep,
+        Storable
+      ]
```
dombesz: Should this be moved into a constant?

app/services/journals/create_service/helpers.rb
```diff
@@ -0,0 +1,66 @@
...
+#++
+
+class Journals::CreateService
+  module Helpers
+    private
+
+    delegate :sanitize,
+             to: ::OpenProject::SqlSanitization
+
+    def journable_id = journable.id
```
dombesz: It seems quite arbitrary to move the `journable_id` into the helper and not move the `journable_type` method.
I found a few more methods that could also be part of the helper: `data_sink_columns`, `data_source_columns`, `journable_data_sql_addition`, `text_column_names`, `journable_timestamp`, `journable_type`, `journable_table_name`, `data_table_name`, `cause_sql`.

lib_static/plugins/acts_as_journalized/lib/acts/journalized/differ/association.rb
```diff
@@ -0,0 +1,124 @@
...
+# See COPYRIGHT and LICENSE files for more details.
+# ++
+
+module Acts::Journalized::Differ
+  class Association
+    # @param original [ActiveRecord::Base] the original object containing the association.
+    # @param changed [ActiveRecord::Base] the changed object containing the association.
+    # @param association [Symbol] the name of the association to be compared.
+    # @param id_attribute [Symbol] the attribute used to correlate association models.
+    # @param multiple_values [Boolean] whether there are multiple values per id_attribute:
```
dombesz: Just a small note, the type here should be `[Boolean, :joined]`

lib_static/plugins/acts_as_journalized/lib/acts/journalized/differ/association.rb
```diff
@@ -0,0 +1,124 @@
...
+    #         a hash with keys being attributes names and values being two element array with changes.
+    def multiple_attributes_changes(attributes, key_prefix:, grouped: false)
+      attributes.each_with_object({}) do |attribute, result|
+        attribute_changes(attribute).each do |id, change|
+          if grouped
+            result["#{key_prefix}_#{id}"] ||= {}
+            result["#{key_prefix}_#{id}"][attribute] = change
+          else
+            result["#{key_prefix}_#{id}_#{attribute}"] = change
+          end
```
dombesz: I believe introducing 2 different key formats adds unnecessary complexity, I would rather prefer to have one key format only.

lib/open_project/journal_formatter/project_life_cycle_step.rb
```diff
@@ -0,0 +1,90 @@
...
+# Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+#
+# See COPYRIGHT and LICENSE files for more details.
+#++
+
+class OpenProject::JournalFormatter::ProjectLifeCycleStep < JournalFormatter::Base
+  def render(key, values, options = { html: true })
+    step = Project::LifeCycleStep.find(key[/\d+/])
+
+    name = step.definition.name
```
dombesz: These will create 2 * M * N + 1 queries. I would suggest taking a look at the [JournalFormatterCache](https://github.com/opf/openproject/blob/dev/lib_static/plugins/acts_as_journalized/lib/journal_formatter_cache.rb) and perhaps at the [JournalFormatter::NamedAssociation](https://github.com/opf/openproject/blob/dev/lib_static/plugins/acts_as_journalized/lib/journal_formatter/named_association.rb) to eliminate it.

spec/features/projects/life_cycle/active_in_project_spec.rb
```diff
@@ -56,10 +56,32 @@
 
   let(:project_life_cycle_page) { Pages::Projects::Settings::LifeCycle.new(project) }
 
+  let(:activity_page) { Pages::Projects::Activity.new(project) }
```
dombesz, line 59: I'm not very firm on this one, but having the activity related specs under the `/features/activities/` would make it more consistent with the existing specs.

spec/features/projects/life_cycle/overview_page/dialog/update_spec.rb
```diff
@@ -31,10 +31,13 @@
 
 RSpec.describe "Edit project stages and gates on project overview page", :js, with_flag: { stages_and_gates: true } do
   include_context "with seeded projects and stages and gates"
+
   shared_let(:overview) { create :overview, project: }
 
   let(:overview_page) { Pages::Projects::Show.new(project) }
 
+  let(:activity_page) { Pages::Projects::Activity.new(project) }
```
dombesz, line 39: Same here, I would prefer the activity related specs under the `/features/activities/`.

spec/lib/acts/journalized/differ/association_spec.rb
```diff
@@ -0,0 +1,257 @@
...
+        let(:multiple_values) { :joined }
+
+        it "returns the flat changes" do
+          expect(result)
+            .to eq(
+              "project_life_cycle_steps_1_active" => ["false", "true"],
+              "project_life_cycle_steps_2_active" => [nil, "true"],
+              "project_life_cycle_steps_3_active" => ["true", "false"],
+              "project_life_cycle_steps_3_end_date" => ["", "2024-01-18"],
+              "project_life_cycle_steps_3_start_date" => ["", "2024-01-17"],
```
dombesz: Is there any reason the `end_date` comes first in all the expectations? It makes it a bit harder to read IMO.

spec/lib/acts/journalized/differ/association_spec.rb
```diff
@@ -0,0 +1,257 @@
...
+              "project_life_cycle_steps_4_end_date" => ["2024-01-17", "2024-01-18"],
+              "project_life_cycle_steps_4_start_date" => ["2024-01-16", "2024-01-17"]
+            )
+        end
+      end
+
+      describe "requesting single value change" do
+        let(:multiple_values) { false }
```
dombesz: The `let(:multiple_values) { true }` case does not seem to be covered.

dombesz (changes requested):
I found a few minor issues and left comments about them. Notably the N+1 issues would require attention, and the scenario below also produces interesting results:

1. Enable a stage
2. Set the date interval
3. Disable the stage

Result: The activities show only the date interval update from step 2. The activation and deactivation does not show up as they cancel each other out.
This looks misleading as the value can only be updated if the step is activated, and the logs show nothing about the field being active. One would assume the field was activated in a previous activity, but that is incorrect.

I propose to update the [`def aggregatable?`](https://github.com/opf/openproject/blob/21b775d1237ed5561341cdf981dab785e4929912/app/services/journals/create_service.rb#L883-L890) method, so that journals are not aggregated when a field is activated (I think deactivation can be aggregated).

We do a similar journal split in case there are multiple notes ("comments") on the same journal. The aggregation is skipped when the user makes multiple comments within the aggregation window.

On a side note, I would rather have the PR unrelated changes in separate PR's in the future, because it makes reviewing a lot easier.

Overall it's a terrific job, thanks @toy 👏

app/models/activities/base_activity_provider.rb
toy: It was intentional, but I'm not sure if there is much sense in it. Normally the method returns a `Hash`, but in case of exception it logs it and returns `true` from `Rails.logger.error` which is unexpected by the only use in `fill_events`. So returning empty `Hash` prevents rescuing one exception to cause another. And I am not sure about it all, as this just hides the exception, and I don't know if there are maybe legacy cases for errors that should be hidden. So maybe the correct way is just to remove rescue and logging altogether.

app/models/activities/fetcher.rb
toy: This handles the case when scope is falsey from `options[:scope] || :all`

spec/lib/acts/journalized/differ/association_spec.rb
toy: Alphabetical :) But I agree that it is easier to read the other way

spec/lib/acts/journalized/differ/association_spec.rb
toy: I skipped that combination, as it is not used and supposedly doesn't need to be used.
The `:joined` is reproducing the behaviour before this PR which is used by all other formatters, where expectation is that all values are strings and multiple and joined with comma. So I added a way to not convert dates to string only to parse them back. But from what I saw multiple values are only needed when getting single attribute for custom fields, so I skipped it for `multiple_attributes_changes` tests.

spec/features/projects/life_cycle/active_in_project_spec.rb
toy: This is here only to check that activity is created after activating life cycles

spec/features/projects/life_cycle/overview_page/dialog/update_spec.rb
toy: And here only to check that activity is created after changing life cycle dates

app/models/journal/project_life_cycle_step_journal.rb
toy: This was intended, maybe not really warranted, but I'm trying to prevent hiding data if in some conditions only one of dates is assigned

app/services/journals/create_service/association.rb
toy: I agree, just rails dev reloading makes it wrong in any way: assigning classes to a constant causes constant to have outdated classes if they are reloaded, assigning names makes it ugly especially with namespaces, and keeping them in a method warrants a valid question why aren't they in a constant

lib_static/plugins/acts_as_journalized/lib/acts/journalized/differ/association.rb
toy: This was really necessary until I introduced diffing virtual attributes instead of attributes only, but now I think it should be possible to adapt the code to not need grouped diff

app/services/journals/create_service/helpers.rb
toy: I extracted only methods that needed to be shared between classes handling associations sql and `CreateService`, but looking again it will be cleaner if delegation is used

toy:
@dombesz Thank you for thorough review

app/models/activities/base_activity_provider.rb
toy: Removed it

lib/open_project/journal_formatter/project_life_cycle_step.rb
toy: The problem is present for all formatters requiring to load additional models, so I've extracted solving it into https://community.openproject.org/wp/62330

dombesz (approved):
Works nicely, thanks @toy !

PR #17599, opened 2025-01-14
Eager load data required for timesheet

ulferts:
# Ticket

https://community.openproject.org/wp/60591

Avoid n+1 queries when exporting timesheets:

## Screenshots

Turns:
image
and more

into:

image


as-op (approved):
Nice catch! 
The speed-up is noticeable for my local test projects. Thanks for the fix!

PR #17604, opened 2025-01-14
[58427] Indicate that there are work package relations which the user has no permission to see

bsatarnejad:
# Ticket
https://community.openproject.org/wp/58427
https://community.openproject.org/wp/60957
https://community.openproject.org/wp/60958
https://community.openproject.org/wp/61044

We want to show ghost relations of a WP in relations tab, but they will look different from visible relations.

## Screenshots
Screenshot 2025-01-15 at 14 01 13

Return both visible and ghost relations to index component and then call relation component with different values for visible and ghost relations.

# Merge checklist

- [ ] Added/updated tests


app/components/work_package_relations_tab/relations_mediator.rb
```diff
@@ -29,21 +29,67 @@
...
+  def visible_relations
+    @visible_relations ||= work_package.relations.includes(:to, :from).visible
+  end
+
+  def visible_children
+    @visible_children ||= work_package.children.visible
+  end
+
+  def ghost_relations
+    @ghost_relations = work_package.relations.includes(:to, :from).where.not(id: @visible_relations.select(:id))
```
cbliard: If `visible_relations` has not been called yet, then `@visible_relations` would be `nil`, and that method would not work properly.

```suggestion
    @ghost_relations = work_package.relations.includes(:to, :from).where.not(id: visible_relations.select(:id))
```

It should use `visible_relations` directly, so that it works even if `visible_relations` has never been called before.

app/components/work_package_relations_tab/relations_mediator.rb
```diff
@@ -29,21 +29,67 @@
...
   end
 
-  def children
-    @children ||= work_package.children.visible
+  def ghost_children
+    @ghost_children ||= work_package.children.where.not(id: @visible_children.select(:id))
```
cbliard: Same: if `visible_children` has not been called yet, then `@visible_children` would be `nil`, and that method would not work properly.

```suggestion
    @ghost_children ||= work_package.children.where.not(id: visible_children.select(:id))
```

config/initializers/menus.rb
```diff
@@ -692,7 +692,7 @@
             { tab: :relations },
             skip_permissions_check: true,
             badge: ->(work_package:, **) {
-              work_package.relations.visible.count + work_package.children.visible.count
+              WorkPackageRelationsTab::RelationsMediator.new(work_package: work_package).all_relations_count
```
cbliard, line 695: Nice catch 👍 
It's nice that you harmonized the usage everywhere.

cbliard (changes requested):
Nice work 👏 

I only have two tiny code change requests.

I have also some 2 other questions / remarks about things that are not specified:
- For milestone work packages, I think it should show only one date like on the full page view
- For predecessor / successor ghost relation, when the dates are not set, I think the date line should still be shown, similar to what happens for ghost children without dates.

I asked both questions to Parimal in the [related work package (#58427)](https://community.openproject.org/wp/58427).

cbliard (changes requested):
Thanks, I found some issues:

1. In the bottom of the page, there is the number of relations being displayed. Not sure why it happens, but it's a bug. I simply load the page with the relations, and this gimmick appears.

image

2. For the dates widget, when there are no dates, a dash should be displayed. Currently nothing is displayed

image

It happens because of the condition `start_date == due_date` to display a single date. The condition is also true when both are `nil`.

Relevant text from the spec in 58427:
> if a date missing (start date and/or finish date) we will nevertheless show the relevant icon with a dash (-) to indicate what is missing.

So I'll mark "review failed" for [Feature: Indicate that there are work package relations which the user has no permission to see (#58427)](https://community.openproject.org/projects/stream-planning-and-reporting/work_packages/58427/activity) for these two reasons.


3. The icon for manually/automatically scheduled is displayed correctly for predecessors / successors, but not for children where the calendar icon is always displayed. Both should behave the same. I think the code to handle is currently duplicated for successors/predecessors and children and could be extracted into a method or a component so that it always behaves the same.

So I'll mark "review failed" for [Bug: Icon beside dates of a relation is wrong in relation row (#60958)](https://community.openproject.org/projects/stream-planning-and-reporting/work_packages/60958/activity)

And I'll mark "review approved" for both [Feature: Relations tab: Show a single date when start and finish dates are the same (#61044)](https://community.openproject.org/projects/stream-planning-and-reporting/work_packages/61044/activity) and [Bug: Show dates of Children of a work package in relations tab (#60957)](https://community.openproject.org/projects/stream-planning-and-reporting/work_packages/60957/activity).

bsatarnejad:
> Thanks, I found some issues:
> 
> 1. In the bottom of the page, there is the number of relations being displayed. Not sure why it happens, but it's a bug. I simply load the page with the relations, and this gimmick appears.
> 
> image
> 
> 2. For the dates widget, when there are no dates, a dash should be displayed. Currently nothing is displayed
> 
> image
> 
> It happens because of the condition `start_date == due_date` to display a single date. The condition is also true when both are `nil`.
> 
> Relevant text from the spec in 58427:
> 
> > if a date missing (start date and/or finish date) we will nevertheless show the relevant icon with a dash (-) to indicate what is missing.
> 
> So I'll mark "review failed" for [Feature: Indicate that there are work package relations which the user has no permission to see (#58427)](https://community.openproject.org/projects/stream-planning-and-reporting/work_packages/58427/activity) for these two reasons.
> 
> 3. The icon for manually/automatically scheduled is displayed correctly for predecessors / successors, but not for children where the calendar icon is always displayed. Both should behave the same. I think the code to handle is currently duplicated for successors/predecessors and children and could be extracted into a method or a component so that it always behaves the same.
> 
> So I'll mark "review failed" for [Bug: Icon beside dates of a relation is wrong in relation row (#60958)](https://community.openproject.org/projects/stream-planning-and-reporting/work_packages/60958/activity)
> 
> And I'll mark "review approved" for both [Feature: Relations tab: Show a single date when start and finish dates are the same (#61044)](https://community.openproject.org/projects/stream-planning-and-reporting/work_packages/61044/activity) and [Bug: Show dates of Children of a work package in relations tab (#60957)](https://community.openproject.org/projects/stream-planning-and-reporting/work_packages/60957/activity).

Thanks @cbliard , I'll fix them 👍

cbliard (approved):
Perfect 👌

PR #17617, opened 2025-01-15
Bug/59916 on narrow screens (including mobile) the view always scrolls to the activity

jjabari-op:
[Ticket](https://community.openproject.org/projects/communicator-stream/work_packages/59916)

- the described issue is resolved and covered with a spec
- I've encountered another scroll behavior on screens below 550px, which seems not to be related to the implementation of the activity tab (-> happens for files/other tabs as well): The page scrolls down around 700px on page load on these small screens. I didn't find the reason for that

### DIFF
--- a/spec/features/activities/work_package/activities_spec.rb
+++ b/spec/features/activities/work_package/activities_spec.rb
@@ -1,3 +1,5 @@
+# frozen_string_literal: true
+
 #-- copyright
 # OpenProject is an open source project management software.
 # Copyright (C) 2012-2024 the OpenProject GmbH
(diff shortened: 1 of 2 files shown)


spec/features/activities/work_package/activities_spec.rb
```diff
@@ -922,24 +922,68 @@
...
+        # this one is actually failing, but it's not caused by the activity tab
+        # the scroll position is at around 700, some other part of the frontend code seems to trigger a scroll
+        # happens for the files tab as well for example
+        #
+        # it "does not scroll to the bottom when the newest journal entry is on the bottom", :aggregate_failures do
+        #   sleep 1 # wait for a potential auto scrolling to finish
+        #   # expect activity tab not to be visibe, as the page is not scrolled to the bottom
+        #   scroll_position = page.evaluate_script("document.querySelector(\"#content-body\").scrollTop")
+        #   expect(scroll_position).to eq(0)
+        # end
```
akabiru: Can we instead mark the context with an RSpec `pending` statement - linking to the work package to resolve this: https://community.openproject.org/work_packages/60638

akabiru (changes requested):
Looks good 👍🏾 Just one resolution and should be good after.

spec/features/activities/work_package/activities_spec.rb
jjabari-op: True! Done :)

akabiru (approved):
🚀

PR #17623, opened 2025-01-16
Fix flickering SharesController spec

NobodysNightmare:
The spec expected a specific order of the items being passed, when the items had no strongly defined order with which they were requested from the database, so tests success was dependent on internal database state always leading to results being returned in the same order.

I've seen this spec fail at least once due to wrong result order.


Kharonus (approved):
make totally sense even without testing it locally.

PR #17624, opened 2025-01-16
[#60183] add pattern contract validation to subject configuration

Kharonus:
# Ticket
[OP#60183](https://community.openproject.org/work_packages/60183)

- subject configuration form for work package type must react on data contract validation
- allow persistence of pattern blueprint if configuration is set to `manual`

- added data contract validation to service
- use specific controller & action to update form configuration
- introduce form object model
- harmonized route definition

### DIFF
--- a/app/controllers/work_packages/types/subject_configuration_controller.rb
+++ b/app/controllers/work_packages/types/subject_configuration_controller.rb
@@ -63,19 +63,29 @@ module WorkPackages
       def tab_path = edit_tab_type_path(id: @type.id, tab: :subject_configuration)
 
       def pattern_collection_update(form_params)
-        case form_params
-        in { subject_configuration: "generated", pattern: String => blueprint }
-          { subject: { blueprint:, enabled: true } }
-        in { subject_configuration: "manual", pattern: String => blueprint }
-          if blueprint.empty?
-            # Submitting the form with an empty blueprint and manual subject configuration will
-            # remove the subject pattern from the collection
-            nil
+        patterns = @type.patterns.to_h
+
+        subject_pattern =
+          case form_params
+          in { subject_configuration: "generated", pattern: String => blueprint }
+            { subject: { blueprint:, enabled: true } }
+          in { subject_configuration: "manual", pattern: String => blueprint }
+            if blueprint.empty?
+              # Submitting the form with an empty blueprint and manual subject configuration will
+              # remove the subject pattern from the collection
+              nil
+            else
+              { subject: { blueprint:, enabled: false } }
+            end
           else
-            { subject: { blueprint:, enabled: false } }
+            nil
           end
+
+        if subject_pattern.nil?
+          patterns.delete("subject")
+          patterns
         else
-          nil
+          patterns.merge(subject_pattern.stringify_keys)
         end
       end
     end
--- a/app/controllers/work_packages/types/subject_configuration_controller.rb
+++ b/app/controllers/work_packages/types/subject_configuration_controller.rb
@@ -1,0 +1,83 @@
+# frozen_string_literal: true
+
+#-- copyright
+# OpenProject is an open source project management software.
+# Copyright (C) the OpenProject GmbH
+#
+# This program is free software; you can redistribute it and/or
+# modify it under the terms of the GNU General Public License version 3.
+#
+# OpenProject is a fork of ChiliProject, which is a fork of Redmine. The copyright follows:
+# Copyright (C) 2006-2013 Jean-Philippe Lang
+# Copyright (C) 2010-2013 the ChiliProject Team
+#
+# This program is free software; you can redistribute it and/or
+# modify it under the terms of the GNU General Public License
+# as published by the Free Software Foundation; either version 2
+# of the License, or (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program; if not, write to the Free Software
+# Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+#
+# See COPYRIGHT and LICENSE files for more details.
+#++
+
+module WorkPackages
+  module Types
+    class SubjectConfigurationController < ApplicationController
+      layout "admin"
+
+      before_action :require_admin
+      before_action :find_type, only: %i[update_subject_configuration]
+
+      def update_subject_configuration
+        form_params = params.require(:types_forms_subject_configuration_form_model)
+                            .permit(:subject_configuration, :pattern)
+                            .to_h
+
+        UpdateTypeService.new(@type, current_user)
+                         .call({ patterns: pattern_collection_update(form_params) }) do |call|
+          call.on_success do
+            redirect_to tab_path, notice: I18n.t(:notice_successful_update)
+          end
+
+          call.on_failure do
+            @default_tab = "subject_configuration"
+            render template: "types/edit", status: :unprocessable_entity
+          end
+        end
+      end
+
+      private
+
+      def find_type
+        @type = ::Type.find(params[:id])
+      end
+
+      def tab_path = edit_tab_type_path(id: @type.id, tab: :subject_configuration)
+
+      def pattern_collection_update(form_params)
+        case form_params
+        in { subject_configuration: "generated", pattern: String => blueprint }
+          { subject: { blueprint:, enabled: true } }
+        in { subject_configuration: "manual", pattern: String => blueprint }
+          if blueprint.empty?
+            # Submitting the form with an empty blueprint and manual subject configuration will
+            # remove the subject pattern from the collection
+            nil
+          else
+            { subject: { blueprint:, enabled: false } }
+          end
+        else
+          nil
+        end
+      end
+    end
+  end
+end
--- a/app/forms/work_packages/types/subject_configuration_form.rb
+++ b/app/forms/work_packages/types/subject_configuration_form.rb
@@ -31,60 +31,55 @@
 module WorkPackages
   module Types
     class SubjectConfigurationForm < ApplicationForm
-      class << self
-        def has_pattern?(type)
-          type.replacement_pattern_defined_for?(:subject)
-        end
-      end
-
       form do |subject_form|
         subject_form.radio_button_group(name: :subject_configuration) do |group|
           group.radio_button(
-            value: "manual",
-            checked: !has_pattern?,
+            value: :manual,
+            checked: subject_configuration_manual?,
             label: I18n.t("types.edit.subject_configuration.manually_editable_subjects.label"),
             caption: I18n.t("types.edit.subject_configuration.manually_editable_subjects.caption"),
             data: { action: "admin--subject-configuration#hidePatternInput" }
           )
           group.radio_button(
-            value: "auto",
-            checked: has_pattern?,
+            value: :generated,
+            checked: !subject_configuration_manual?,
             label: I18n.t("types.edit.subject_configuration.automatically_generated_subjects.label"),
             caption: I18n.t("types.edit.subject_configuration.automatically_generated_subjects.caption"),
-            disabled: !EnterpriseToken.active? && !has_pattern?,
+            disabled: !enterprise? && subject_configuration_manual?,
             data: { action: "admin--subject-configuration#showPatternInput" }
           )
         end
 
         subject_form.group(data: { "admin--subject-configuration-target": "patternInput" }) do |toggleable_group|
           toggleable_group.text_field(
-            name: :subject_pattern,
-            value: subject_pattern,
+            name: :pattern,
             label: I18n.t("types.edit.subject_configuration.pattern.label"),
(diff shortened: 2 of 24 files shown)


app/forms/work_packages/types/subject_configuration_form.rb
```diff
@@ -31,39 +31,33 @@
...
           )
           group.radio_button(
-            value: "auto",
-            checked: has_pattern?,
+            value: :generated,
+            checked: !subject_configuration_manual?,
             label: I18n.t("types.edit.subject_configuration.automatically_generated_subjects.label"),
             caption: I18n.t("types.edit.subject_configuration.automatically_generated_subjects.caption"),
-            disabled: !EnterpriseToken.active? && !has_pattern?,
+            disabled: !EnterpriseToken.active?,
```
NobodysNightmare: :yellow_circle: I am pretty sure that the `&& !has_pattern` was required to fulfill our requirements around OP#59929. Specifically: If the token is removed, but the pattern is enabled, we should keep editing enabled.

app/helpers/tabs_helper.rb
```diff
@@ -37,8 +39,8 @@ def render_tabs(tabs, form = nil, with_tab_nav: true)
     end
   end
 
-  def selected_tab(tabs)
-    tabs.detect { |t| t[:name] == params[:tab] } || tabs.first
+  def selected_tab(tabs, default_tab = nil)
+    tabs.detect { |t| t[:name] == params[:tab] || t[:name] == default_tab } || tabs.first
```
NobodysNightmare: :yellow_circle: I didn't yet see where this is called, but won't this cause the default tab to be returned, even though there might be a tab that would fit the `params[:tab]`?

E.g.

```ruby
tabs = %w[a b c d].map { |name| { name: name } }
params = { tab: "c" }
selected_tab(tabs, "b")
=> "b"
```

I would expect that you want the above example to return "c", but the implementation will return "b".

app/services/base_type_service.rb
```diff
@@ -75,13 +75,17 @@ def set_params_and_validate(params)
...
-
     # This should go before `set_scalar_params` call to get the
     # project_ids, custom_field_ids diffs from the type and the params.
     # For determining the active custom fields for the type, it is necessary
     # to know whether the type is a milestone or not.
     set_milestone_param(params) unless params[:is_milestone].nil?
+
+    # Set patterns includes a data validation before assigning the value to the attribute.
+    # A failure should return a failure.
+    set_patterns(params) unless params[:patterns].nil?
```
NobodysNightmare: :green_circle: I'd prefer to avoid negations. Therefore I'd suggest either of

```suggestion
    set_patterns(params) if params[:patterns]
```

or

```suggestion
    set_patterns(params) if params[:patterns].present?
```

None of the two is exactly equal to your code w.r.t. semantics, thought I'd assume that we either expect a hash with some keys or `nil`, in which case the behaviour should be the same.

app/controllers/work_packages/types/subject_configuration_controller.rb
```diff
@@ -0,0 +1,83 @@
...
+      before_action :require_admin
+      before_action :find_type, only: %i[update_subject_configuration]
+
+      def update_subject_configuration
+        form_params = params.require(:types_forms_subject_configuration_form_model)
+                            .permit(:subject_configuration, :pattern)
+                            .to_h
+
+        UpdateTypeService.new(@type, current_user)
+                         .call({ patterns: pattern_collection_update(form_params) }) do |call|
```
NobodysNightmare, line 45: :red_circle: I think this code is not forward-compatible enough for the data structure we prepared. Let me explain...

Context: The `patterns` attribute was written in a way, that it already expects us to handle more than just the patterns of a subject (the focus of our current story). You can see that, because it's not the `subject_pattern` attribute, but rather it's an attribute that looks like a hash and has a key called `subject`. This means we can expect that the `patterns` attribute is supposed to be used for other kinds of patterns in the future and `subject` is just one of them.

Now this code is calling the `UpdateTypeService` with a full representation of `patterns`. The `UpdateTypeService` will take this to overwrite the `patterns` attribute in full. On the other hand, we source the value from `pattern_collection_update`, which only takes the `form_params` into account, which will only carry subject-specific state here (our controller is the **subject** configuration controller).

In other words: This controller is written in a way that it will overwrite any and all non-subject-patterns, if they existed on the `patterns` collection. IMO that's an error.

The two high-level solutions that come to my mind are:
1. Write this update in a way that non-subject keys will not be harmed
2. Change the structure of `patterns` back to something that does not claim to be future-proofed for other things than subjects

My preferred solution would be 1, but I remember that @mereghost talked to me about not future-proofing too much, that's why I am suggesting 2 as well. The current solution is IMO "half future-proof", which is the worst of both worlds. We now have a data structure that can hold more than just a subject, but we treat it like a data structure that **only ever** holds a subject.

app/services/base_type_service.rb
```diff
@@ -107,12 +120,6 @@ def set_attribute_groups(params)
     end
   end
 
-  def set_subject_pattern(params)
-    type.patterns = type.patterns.update_pattern(:subject,
```
NobodysNightmare, line 111: :yellow_circle: Assuming we remain at the current implementation, the `update_pattern` method is now unused and can be removed entirely.

NobodysNightmare (changes requested):
I left some comments, most of them minor, though one (marked with :red_circle:) I deem important to properly discuss. Looks good to me in general.

One thing I was wondering about: I would have expected this PR to contain further validations on the blueprint (e.g. validating the tokens inside for existence). Was this planned for a separate pull request?

app/forms/work_packages/types/subject_configuration_form.rb
Kharonus: Yes, I removed it, because I introduced the form object and no longer worked on top of a wp type model.

I added now the check again in a slightly changed way, but expressing a similar logic for disabling the radio button.

app/helpers/tabs_helper.rb
Kharonus: true.... why is that happening? I will look into it. Thanks for finding it.

app/services/base_type_service.rb
Kharonus: yep, agreed

app/helpers/tabs_helper.rb
NobodysNightmare: Explanation-wise: Because both conditions are in a single `detect` (aka `find`) block. I guess what you intended to do was `tabs.find { /* by params */ } || tabs.find { /* by default */ } || tabs.first`

app/forms/work_packages/types/subject_configuration_form.rb
```diff
@@ -31,60 +31,55 @@
...
-            input_width: :large
+            input_width: :large,
+            validation_message: validation_message_for(:patterns)
           )
         end
 
         subject_form.submit(
           name: :submit,
           label: I18n.t(:button_save),
-          disabled: !EnterpriseToken.active? && !has_pattern?,
```
NobodysNightmare, line 73: This was part of a fix that I specifically performed after QA noticed that the submit button remained enabled, even if the rest of the form is disabled.

I agreed that it's confusing to be able to submit a form that allows no changes to be made.

PR introducing this: https://github.com/opf/openproject/pull/17610

app/controllers/work_packages/types/subject_configuration_controller.rb
Kharonus: hmm, let me think a step ahead of option 1:

would you prefer adding this logic to the controller or to the service? The controller could merge the `patterns` data hash before sending it to the service which then does the data validation and assigns it to the object before persitence.

On the other hand, one could argue, the controller should not care about such trifles. in the end it is now a specific controller for subject generation configuration. So, the update type service could be the one merging the data. 

For the second approach I fear, the cut may seem clean, but it is not, when it comes to pattern deletion. So, if the update is a removal of a certain pattern. The first approach can be easily written, as the controller knows everything about the context and purpose.

WDYT? @NobodysNightmare @mereghost

app/forms/work_packages/types/subject_configuration_form.rb
Kharonus: I talked it through with dominic just today. The application of the form model makes the disabling of the submit button obsolete. I can explain, if need be :D

NobodysNightmare (changes requested):
`permitted_params.rb` still contains my previous code that allows `subject_pattern` and `subject_configuration` on type.

You can clean this up.

app/controllers/work_packages/types/subject_configuration_controller.rb
NobodysNightmare: As we discussed IRL: What sounds like a plausible approach is for the controller to take the `patterns` and inside the patterns only assign to the `:subject` key and then pass the whole thing along to the service.

This means the service does not have to guess the intention of its caller and the controller is not _really_ concerned with state-management too much. It just carefully changes the bits it's interested in.

app/forms/work_packages/types/subject_configuration_form.rb
NobodysNightmare: As discussed IRL: I can kinda live with most outcomes, but it would make sense to align our wireframes to match what we want to see, so that QA also knows what to expect.

app/helpers/tabs_helper.rb
Kharonus: solved

app/controllers/work_packages/types/subject_configuration_controller.rb
Kharonus: Tried to implement this approach.

app/forms/work_packages/types/subject_configuration_form.rb
Kharonus: Yes, I will talk to Dominic and cecile.

spec/controllers/work_packages/types/subject_configuration_controller_spec.rb
```diff
@@ -0,0 +1,117 @@
...
+RSpec.describe WorkPackages::Types::SubjectConfigurationController do
+  let(:user) { create(:admin) }
+  let(:wp_type) { create(:type) }
+
+  current_user { user }
+
+  context "when the user is not logged in" do
+    let(:user) { User.anonymous }
+
+    it "responds with forbidden" do
```
NobodysNightmare: :green_circle: That's not technically what we expect.

The expectation is about a redirect to the login, not an HTTP 403 Forbidden.

Maybe:
```suggestion
    it "requires login" do
```

spec/controllers/work_packages/types/subject_configuration_controller_spec.rb
```diff
@@ -0,0 +1,117 @@
...
+  context "when the user is not an admin" do
+    let(:user) { create(:user) }
+
+    it "responds with forbidden" do
+      put :update_subject_configuration, params: { id: wp_type.id }
+      expect(response).to have_http_status :forbidden
+    end
+  end
+
+  describe "PUT #update_subject_configuration" do
```
NobodysNightmare, line 58: :yellow_circle: Both contexts (when the user is not an admin and when the user is not logged in) above are testing the behaviour of the `update_subject_configuration` action when receiving a PUT. I think those contexts should semantically be placed inside this `describe` as well.

As far as I can tell, this should also not conflict with the additional setup peformed in the current describe block.

app/forms/work_packages/types/subject_configuration_form.rb
Kharonus: As discussed IRL: We leave submit button active and I update the AC.

spec/controllers/work_packages/types/subject_configuration_controller_spec.rb
Kharonus: While this is true, that it probably won't clash, the two contexts above are about controller before_actions. So, technically, those are not about this PUT, we just need any action to trigger the before action. Would you be fine to leave them outside?

NobodysNightmare: Let me add this perspective: That these things are `before_action`s is an implementation detail. A `before_filter` can take an `only` or `except` argument, there is no guarantee that it runs for each action.

So I'd probably ask a proper unit test for a controller to repeat this expectation for all of the controller actions, to ensure that all actions behave the way they should. That's also the reason why I saw this context inside the action `describe`.

I am definitely not going to die on that hill, i.e. I'd be fine to leave the context where it is (especially considering that we only have one action so far).

app/services/update_type_service.rb
```diff
@@ -35,4 +37,24 @@ def call(params)
...
+
+    super
+  end
+
+  def set_patterns(params)
+    Types::Patterns::Collection
+      .build(patterns: params[:patterns])
+      .either(
+        ->(collection) { type.patterns = collection },
+        ->(result) { type.errors.add(:patterns, result.errors(full: true).messages.join(", ")) }
```
NobodysNightmare, line 57: :green_circle: Sorry that I didn't see this one before. Shouldn't we loop and add all the errors, so that formatting (i.e. the `join(', ')`) is left as a choice to the final renderer?

I.e.
```suggestion
        ->(result) { result.errors(full: true).each { |e| type.errors.add(:patterns, e.message) }
```

NobodysNightmare (approved):
As I said, none of my remaining comments is blocking a merge. LGTM.

app/services/update_type_service.rb
Kharonus: Solved in #17644

PR #17636, opened 2025-01-16
Do not focus custom fields in the create child dialog

cbliard:
# Ticket

PR #17625
Late review revealed an issue. This PR is a follow-up about it.

In the create child dialog, displayable from the relations tab of a work package, custom fields using the autocompleter input field are getting the focus by default. This is visible only if the work package type has some specific kind of custom fields like `hierarchy`, `version` or `list`.

This is not correct: the field to focus is the subject field.

Also add a test to ensure that all different kind of custom fields are displayed in the create child dialog and no error is reported.

Pass in `focusDirectly: false` so that the focus is not stolen. This may benefit some other areas of the app.

# Merge checklist

- [X] Added/updated tests

### DIFF
--- a/spec/features/work_packages/tabs/relations_children_spec.rb
+++ b/spec/features/work_packages/tabs/relations_children_spec.rb
@@ -5,6 +5,8 @@ require "spec_helper"
 require "support/edit_fields/edit_field"
 
 RSpec.describe "Relations children tab", :js, :with_cuprite do
+  include CustomFieldsHelpers
+
   shared_let(:normal_cf) { create(:string_wp_custom_field, is_required: false) }
   shared_let(:required_cf) { create(:string_wp_custom_field, is_required: true) }
   shared_let(:type_milestone) { create(:type_milestone) }
@@ -118,20 +120,13 @@ RSpec.describe "Relations children tab", :js, :with_cuprite do
     let!(:user) { create(:admin) }
 
     before do
-      # Introspect FactoryBot to find all traits used to create work package custom fields
-      traits = FactoryBot.factories[:wp_custom_field].defined_traits
-      traits = traits.reject { |t| t.name == "multi_value" }
-      traits = traits.map { |t| t.name.to_sym }
-
-      traits.each do |trait|
-        [true, false].each do |required| # rubocop:disable Performance/CollectionLiteralInLoop
-          cf = create(:wp_custom_field,
-                      trait,
-                      is_required: required)
+      factory_bot_custom_field_traits_for("WorkPackage")
+        .product([true, false])
+        .each do |trait, is_required|
+          cf = create(:wp_custom_field, trait, is_required:)
           project.types.first.custom_fields << cf
           project.work_package_custom_fields << cf
         end
-      end
     end
 
     it "displays a field for each required custom field" do
--- a/spec/features/work_packages/tabs/relations_children_spec.rb
+++ b/spec/features/work_packages/tabs/relations_children_spec.rb
@@ -113,4 +113,48 @@ RSpec.describe "Relations children tab", :js, :with_cuprite do
       relations_tab.expect_no_new_relation_type("New child")
     end
   end
+
+  context "when all possible custom fields are there" do
+    let!(:user) { create(:admin) }
+
+    before do
+      # Introspect FactoryBot to find all traits used to create work package custom fields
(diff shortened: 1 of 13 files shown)


spec/features/work_packages/tabs/relations_children_spec.rb
```diff
@@ -113,4 +113,48 @@
       relations_tab.expect_no_new_relation_type("New child")
     end
   end
+
+  context "when all possible custom fields are there" do
+    let!(:user) { create(:admin) }
+
+    before do
+      # Introspect FactoryBot to find all traits used to create work package custom fields
+      traits = FactoryBot.factories[:wp_custom_field].defined_traits
```
NobodysNightmare: :yellow_circle: I like the idea of having each custom field type in a single test very much. It essentially mimics what one might do when testing this manually as well.

However, I am not a big fan of iterating over FactoryBot traits for that to be honest. The reason is that it's IMO easy for these to misalign with custom field types:
* I might create a custom field type, but forget to create a trait for it (e.g. the `multi_hierarchy` option that you just added in this PR)
    * no tests will fail, but this one will then be incomplete
* I might create a trait for the factory that makes sense, but would now (unknowingly) change this spec
    * if that makes this test fail it's annoying, but at least I will know
    * if that doesn't make it fail, then I wouldn't even notice that I changed this spec

Can't we instead use `OpenProject::CustomFieldFormat.available_formats` to iterate the custom field formats that are defined and then look for factory traits with a corresponding name?

This would make us iterate over an authoritative source of custom field types (one that _has_ to be changed when adding a new one) and it would as a side-effect also encourage to set-up proper factory traits when introducing a new type.

NobodysNightmare (changes requested):
Thanks for the quick fix. I had one comment on the way that the custom field types are iterated, let me know what you think :)

spec/features/work_packages/tabs/relations_children_spec.rb
cbliard: Actually, I used `OpenProject::CustomFieldFormat.available_formats` at first, and it did not work out of the box because some custom fields need additional options to be valid. For instance, `list` needs options, and `hierarchy` needs root item. Creating multi-value custom fields for types that support it was also a bit weird to code: calling `#multi_value_possible?` on a custom field instance to know if a multi-value version should be instantiated.

That's why I ended up using `FactoryBot`. I wanted something better, like selecting only traits which have redefined `field_format`, but did not manage to do it. That why it manually excludes the `multi_value` trait.

I still think that's an acceptable trade-off, as our factories should reflect what we have in code, and they should also work out of the box.

I share your concerns that they may be misaligned with the defined custom field types. What I can do is encapsulating these traits in a method, so that the logic can be reused somewhere else, and have that method tested, so that we continuously check that all custom field types with their multi-value variant are covered.

What do you think?

NobodysNightmare: Your idea sounds good and would alleviate my concerns.

Let me quickly share what I had in mind, because this was probably not clear enough: I don't have a huge problem with assuming that certain factory traits exist, my main concern was that the factory bot traits were the _source_ of possible CF types to iterate from.

So if it helps, something along those lines would work for me, **too**:

```ruby
traits = OpenProject::CustomFieldFormat.available_formats.flat_map { |f| [f, "multi_#{f}".to_sym] }
traits.each do |trait|
  # what you already did before
end
```

cbliard: Do you know what the format `"empty"` is used for?

cbliard: Ah, found it in `config/initializers/custom_field_format.rb`

```ruby
  # This is an internal formatter used as a fallback in case a value is not found.
  # Setting the label to nil in order to avoid it becoming available for selection as a custom value format.
  fields.register OpenProject::CustomFieldFormat.new("empty",
                                                     label: nil,
                                                     order: 11,
                                                     formatter: "CustomValue::EmptyStrategy")
```

cbliard: @NobodysNightmare I changed it the way we discussed. It now relies on `OpenProject::CustomFieldFormat` information to build up the traits list.

I also stumbled upon an order-dependent test. I fixed it.

Could you review it again please?

NobodysNightmare (approved):
LGTM

PR #17640, opened 2025-01-17
Wrapper of autocompleter is a grid box without setting a column

bsatarnejad:
# Ticket
https://community.openproject.org/wp/60619

Avoid the selected value of autocompleter from overflowing.

## Screenshots
Before:
https://github.com/user-attachments/assets/3a75d437-9814-492f-bfc4-3be23a715f25

After:
https://github.com/user-attachments/assets/a2f559d5-9630-4e59-9769-67058dd26a45

With our new implementation for form input fields we are wrapping them in a 'FormControl-input-wrap'. It works for all inputs  except autocompleter. We could use 'FormControl-select-wrap' class, but it doesn't work either in ng-select. By setting a minmax(0, auto) on a column, the column will not cause layout issues by expanding beyond the container.

# Merge checklist

- [ ] Added/updated tests

### DIFF
--- a/frontend/src/global_styles/content/_autocomplete.sass
+++ b/frontend/src/global_styles/content/_autocomplete.sass
@@ -172,3 +172,7 @@ div.autocomplete
 
 .work-package-table--container .ng-dropdown-panel
   z-index: auto !important
+
+
+.FormControl-input-wrap:has(ng-select)
+  grid-template-columns: minmax(0, auto)


HDinger (changes requested):
This is not ideal because of multiple reasons:

* It relys on a Primer class. If they change the class name that will break
* This is really specific to the implementation of the `.FormControl-input-wrap` class. If the Primer team changes the implementation of it even slightly, this will most likely break.
* Looking at the DOM, the `div` below that is actually the overflowing one so I guess the solution should be targeted there (if possible)

Bildschirmfoto 2025-01-17 um 09 36 06

PR #17641, opened 2025-01-17
Fix flaky test

cbliard:
The test is opening a dialog and setting text in an autocompleter. That triggers the dropdown of the autocompleter options. But as the dialog is still in its opening animation, the dropdown is not displayed at the right place: it is over the input. When the test then tries to click on that input, it fails because the dropdown is overlapping the input. That's the flakiness

This commit adds a helper `wait_for_size_animation_completion(selector)` to wait for the animation of an element to finish. It's used to wait for the dialog animation to complete. Once the animation is finished, it is safe to set the text in the input. It triggers the display of the dropdown as before, but it is now displayed at the right place and the test can continue without failing.

The flaky test was `./spec/features/projects/project_custom_fields/overview_page/dialog/inputs_spec.rb:252`
It was failing in https://github.com/opf/openproject/actions/runs/12807028970/job/35706720767

The `wait_for_size_animation_completion` helper works by getting the element size, waiting 100ms, get the size again and continue looping if the two sizes are different.

### DIFF
--- a/spec/support/capybara/wait_helpers.rb
+++ b/spec/support/capybara/wait_helpers.rb
@@ -1,0 +1,53 @@
+# frozen_string_literal: true
+
+#-- copyright
+# OpenProject is an open source project management software.
+# Copyright (C) the OpenProject GmbH
+#
+# This program is free software; you can redistribute it and/or
+# modify it under the terms of the GNU General Public License version 3.
+#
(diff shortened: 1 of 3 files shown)


cbliard:
And the tests suite fails for another flaky spec
`./spec/features/placeholder_users/placeholder_user_memberships_spec.rb:41` this time
But it passes locally.

PR #17643, opened 2025-01-17
UserTokens::FetchService

NobodysNightmare:
# Ticket
https://community.openproject.org/projects/cross-application-user-integration-stream/work_packages/60637

Providing a way to make `OpenIDConnect::UserToken` practically usable for authentication.

When you want to use such a token during authentication, you need to find the correct token for your target service and you also want to be able to ensure that it has been refreshed, if necessary.

This service is also intended to later be able to perform Token Exchange, which will be used to obtain an entirely new token, if none is yet available.

Providing a single service class that can be used to obtain an access token. For most practical use-cases the `access_token_for` method will be enough, because we will be able to parse an expiration date from the token.

However, if a client notices that the access token is rejected, they still might need to refresh it manually, before being able to make a successful call. For that case they can force refreshing the token via `refreshed_access_token_for`.

# Merge checklist

- [x] Added/updated tests

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -304,16 +304,16 @@ class User < Principal
     end
   end
 
-  # Return user's authentication provider for display
   def authentication_provider
-    return if identity_url.blank?
+    return nil if identity_url.blank?
 
-    identity_url.split(":", 2).first
+    slug = identity_url.split(":", 2).first
+    AuthProvider.find_by(slug:)
   end
 
   # Return user's authentication provider for display
   def human_authentication_provider
-    authentication_provider&.titleize
+    authentication_provider&.display_name
   end
 
   ##
--- a/modules/openid_connect/app/services/openid_connect/jwt_parser.rb
+++ b/modules/openid_connect/app/services/openid_connect/jwt_parser.rb
@@ -38,8 +38,9 @@ module OpenIDConnect
       RS512
     ].freeze
 
-    def initialize(verify_audience: true, required_claims: [])
+    def initialize(verify_audience: true, verify_expiration: true, required_claims: [])
       @verify_audience = verify_audience
+      @verify_expiration = verify_expiration
       @required_claims = required_claims
     end
 
@@ -56,6 +57,7 @@ module OpenIDConnect
           true,
           {
             algorithm: alg,
+            verify_expiration: @verify_expiration,
             verify_aud: @verify_audience,
             aud: provider.client_id,
             required_claims: all_required_claims
--- a/modules/openid_connect/app/services/openid_connect/user_tokens/fetch_service.rb
+++ b/modules/openid_connect/app/services/openid_connect/user_tokens/fetch_service.rb
@@ -1,0 +1,123 @@
+# frozen_string_literal: true
+
+#-- copyright
+# OpenProject is an open source project management software.
+# Copyright (C) 2012-2024 the OpenProject GmbH
+#
+# This program is free software; you can redistribute it and/or
+# modify it under the terms of the GNU General Public License version 3.
+#
+# OpenProject is a fork of ChiliProject, which is a fork of Redmine. The copyright follows:
+# Copyright (C) 2006-2013 Jean-Philippe Lang
+# Copyright (C) 2010-2013 the ChiliProject Team
+#
+# This program is free software; you can redistribute it and/or
+# modify it under the terms of the GNU General Public License
+# as published by the Free Software Foundation; either version 2
+# of the License, or (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program; if not, write to the Free Software
+# Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+#
+# See COPYRIGHT and LICENSE files for more details.
+#++
+
+module OpenIDConnect
+  module UserTokens
+    ##
+    # Provides APIs to obtain access tokens of a given user for use at a third-party
+    # application for which we know the audience name, which is typically the application's
+    # client_id at an identity provider that OpenProject and the application have in common.
+    class FetchService
+      include Dry::Monads[:result]
+
+      def initialize(user:, jwt_parser: JwtParser.new(verify_audience: false, verify_expiration: false))
+        @user = user
+        @provider = user.authentication_provider
+        @jwt_parser = jwt_parser
+      end
+
+      ##
+      # Obtains an access token that can be used to make requests in the user's name at the
+      # remote service identified by the +audience+ parameter.
+      #
+      # The access token will be refreshed before being returned by this method, if it can be
+      # identified as being expired. There is no guarantee that all access tokens will be properly
+      # recognized as expired, so client's still need to make sure to handle rejected access tokens
+      # properly. Also see #refreshed_access_token_for.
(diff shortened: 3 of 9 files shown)


app/models/user.rb
```diff
@@ -304,16 +304,16 @@ def name(formatter = nil)
...
 
-    identity_url.split(":", 2).first
+    slug = identity_url.split(":", 2).first
+    AuthProvider.find_by(slug:)
   end
 
   # Return user's authentication provider for display
   def human_authentication_provider
-    authentication_provider&.titleize
+    authentication_provider&.slug&.titleize
```
NobodysNightmare: While this change is the semantically equal refactoring of the previous code, I am wondering whether we should now use this instead:

```suggestion
    authentication_provider&.display_name
```

This seems more logical to me, now that we have loaded the correct provider. Though it would change what's displayed. E.g. if you have `Keycloak` as an OIDC provider, it would have previously been shown as `Oidc Keycloak`, while it's now shown as `Keycloak` :shrug:

modules/openid_connect/app/services/openid_connect/jwt_parser.rb
```diff
@@ -56,6 +57,7 @@ def parse(token)
           true,
           {
             algorithm: alg,
+            verify_expiration: @verify_expiration,
```
NobodysNightmare, line 60: TODO: Add tests

modules/openid_connect/app/services/openid_connect/user_tokens/fetch_service.rb
```diff
@@ -0,0 +1,104 @@
...
+module OpenIDConnect
+  module UserTokens
+    class FetchService
+      include Dry::Monads[:result]
+
+      def initialize(user:, jwt_parser: JwtParser.new(verify_audience: false, verify_expiration: false))
+        # TODO: fail if no provider? fail if user is not using OIDC?
+        # or optimistically assume that if there is a token, everything will be alright?
+        @user = user
+        @provider = user.authentication_provider
```
NobodysNightmare, line 42: I am going to remove the TODO comment above and choose the "optimistic" path, unless anyone stops me.

Rationale: For the `provider` to be missing, the user would either not be using OIDC or the provider would have to be deleted.

In the former case, the user should not have any `oidc_user_tokens`, so there is no way that we'd ever try to use the provider to refresh a token.

In the latter case, there would be no way for the user to log-in.

app/models/user.rb
NobodysNightmare: Looking at where/how this is used, I am now using the `display_name`.

modules/openid_connect/app/services/openid_connect/jwt_parser.rb
NobodysNightmare: done

modules/openid_connect/app/services/openid_connect/user_tokens/fetch_service.rb
```diff
@@ -0,0 +1,121 @@
...
+
+      private
+
+      def token_with_audience(aud)
+        token = @user.oidc_user_tokens.where("audiences ? :aud", aud:).first
+        token ? Success(token) : Failure("No token for given audience")
+      end
+
+      def refresh(token)
+        refresh_token_request(token.refresh_token).bind do |json|
```
NobodysNightmare, line 90: :red_circle: We should handle the case of having no `refresh_token` available, i.e. refresh being impossible.

For the scope of this PR, that's simply an error. For later scope, this might as well simply result in an attempt to exchange a fresh token.

NobodysNightmare: done

modules/openid_connect/app/services/openid_connect/user_tokens/fetch_service.rb
```diff
@@ -0,0 +1,123 @@
...
+        token = token.bind do |t|
+          if expired?(t.access_token)
+            refresh(t)
+          else
+            Success(t)
+          end
+        end
+
+        token.fmap(&:access_token)
+      end
```
Kharonus, line 65: 🟡 would this method be much more readable using the "do notation`?

It would look similar to:

```ruby
      def access_token_for(audience:)
        token = yield token_with_audience(audience)
     
        token = yield refresh(token) if expired?(token.access_token)
         
        token.fmap(&:access_token)
      end

Comment (Kharonus) [reply modules/openid_connect/app/services/openid_connect/user_tokens/fetch_service.rb:65]: for context: https://dry-rb.org/gems/dry-monads/1.6/do-notation/

File: modules/openid_connect/app/services/openid_connect/user_tokens/fetch_service.rb [hunk 6]
```diff
@@ -0,0 +1,123 @@
...
+
+      def refresh_token_request(refresh_token)
+        response = OpenProject.httpx
+                              .basic_auth(@provider.client_id, @provider.client_secret)
+                              .post(@provider.token_endpoint, form: {
+                                      grant_type: :refresh_token,
+                                      refresh_token:
+                                    })
+        response.raise_for_status
```
Kharonus, line 108: 🟡 Should we raise or actually return a failure here? Was it an option, while you designed it?

modules/openid_connect/app/services/openid_connect/user_tokens/fetch_service.rb
```diff
@@ -0,0 +1,123 @@
...
+
+      def token_with_audience(aud)
+        token = @user.oidc_user_tokens.where("audiences ? :aud", aud:).first
+        token ? Success(token) : Failure("No token for given audience")
+      end
+
+      def refresh(token)
+        return Failure("Can't refresh the access token") if token.refresh_token.blank?
+
+        refresh_token_request(token.refresh_token).bind do |json|
```
Kharonus, line 90: 🟡 Here you could also get rid of the `bind` and the indention with the do notation.

modules/openid_connect/app/services/openid_connect/user_tokens/fetch_service.rb
Kharonus: ah, you are actually doing that with the rescue, I see. NVM

Kharonus (changes requested):
So, I have only some code style remarks. So, if we conclude together, that those are not worth it, I'd be okay with approving this.

modules/openid_connect/app/services/openid_connect/user_tokens/fetch_service.rb
NobodysNightmare: We could have a vote on that :sweat_smile: 

While I started to get used to and even like monads a bit while writing this service, I am still not super warm with the do-notation. To be honest, I start to have zero feeling for the flow of execution anymore, when reading these methods (which block am I yielding to?).

So yeah, I avoided using this notation as a matter of personal reading preference.

modules/openid_connect/app/services/openid_connect/user_tokens/fetch_service.rb
NobodysNightmare: If we decide to **do** that, I'd suggest to make the change in the follow up PR, which moves the refreshing into a separate service.

modules/openid_connect/app/services/openid_connect/user_tokens/fetch_service.rb
NobodysNightmare: I think I accept defeat ;-)

https://github.com/opf/openproject/commit/77c203f9b75d171f3ba1433b4f24e0642be55e96

(Note that this commit is part of the follow-up PR https://github.com/opf/openproject/pull/17678)

Kharonus (approved):
Can we merge those implementations of the services? They are working, they are tested. And if we want to rework on top of them some interfaces, this is no blocker for a merge.

PR #17644, opened 2025-01-17
[#60518] validate enterprise action in backend 

Kharonus:
# Ticket
[OP#60519](https://community.openproject.org/work_packages/60518)

- if UI is tricked, update of subject generation must still be controlled by enterprise guard

- validate pattern update against enterprise action before parsing it

### DIFF
--- a/app/services/update_type_service.rb
+++ b/app/services/update_type_service.rb
@@ -42,19 +42,35 @@ class UpdateTypeService < BaseTypeService
 
   def set_params_and_validate(params)
     # Set patterns includes a data validation before assigning the value to the attribute.
-    # A failure should return a failure.
-    set_patterns(params) if params[:patterns].present?
-    return [false, type.errors] if type.errors.any?
+    # A validation failure should return a service call failure.
+    if params[:patterns].present?
+      validate_enterprise_action(params[:patterns])
+      set_patterns(params[:patterns])
+      return [false, type.errors] if type.errors.any?
+    end
 
     super
   end
 
-  def set_patterns(params)
+  def validate_enterprise_action(patterns)
+    change_from_manual_to_generated = !type.patterns.subject&.enabled? && patterns.dig(:subject, :enabled)
+    action = :work_package_subject_generation
+
+    if change_from_manual_to_generated && !EnterpriseToken.allows_to?(action)
+      type.errors.add(:patterns, :error_enterprise_only, action: action.to_s.titleize)
+    end
+  end
+
+  def set_patterns(patterns)
     Types::Patterns::Collection
-      .build(patterns: params[:patterns])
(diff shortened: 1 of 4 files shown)


app/services/update_type_service.rb
```diff
@@ -42,19 +42,36 @@ def call(params)
...
+      set_patterns(params[:patterns])
+      return [false, type.errors] if type.errors.any?
+    end
 
     super
   end
 
-  def set_patterns(params)
+  def validate_enterprise_action(patterns)
+    change_from_manual_to_generated = !type.patterns.subject.enabled? && patterns.dig(:subject, :enabled) == true
```
NobodysNightmare: :green_circle: Is the `== true` necessary?

```suggestion
    change_from_manual_to_generated = !type.patterns.subject.enabled? && patterns.dig(:subject, :enabled)
```

app/services/update_type_service.rb
```diff
@@ -42,19 +42,36 @@ def call(params)
...
+
+    if change_from_manual_to_generated && !EnterpriseToken.allows_to?(enterprise_action)
+      type.errors.add(:patterns, :error_enterprise_only, action: enterprise_action.to_s.titleize)
+    end
+  end
+
+  def enterprise_action = :work_package_subject_generation
```
NobodysNightmare: :yellow_circle: I find it surprising that the `UpdateTypeService` has a method `enterprise_action` that's specific to subject generation.

Couldn't there be multiple enterprise features on the `Type` that would need to be checked in the `UpdateTypeService`?

I'd suggest one of the following:
* Make `validate_enterprise_action` accept the action as an argument and pass it from `set_params_and_validate` (inside the if block it's clear about which feature we must be talking)
* Rename `enterprise_action` to something like `subject_generation_enterprise_action`

NobodysNightmare (changes requested):
I'd have expected to find this validation logic on a contract, though I also find it acceptable on the update service. Just a surprise for me. Is this because updating types is different or would you also have implemented it like this on a model that follows all our usual patterns? (create service, update service, contracts, etc)

:yellow_circle: Should we have specs for the new validation?

app/services/update_type_service.rb
Kharonus: Well, usually I'd say this is only important once the service actually DOES check for multiple enterprise actions. But I also have no big concern about using the proposed name.

app/services/update_type_service.rb
Kharonus: Now, that is an old question :D In a typed language, I'd say "No". But here you could have any value at key `enabled`.

Generally I agree with you though

Kharonus:
Added some unit tests

NobodysNightmare (approved):

PR #17646, opened 2025-01-17
[#60649] Change op-autocompleter debounce to 250ms

myabc:
⚠️ **This change affects all subclasses of `OpAutocompleterComponent`, not just the work package relations autocompleter `WorkPackageRelationsAutocompleteComponent`**

# Ticket

https://community.openproject.org/wp/60649

Change op-autocompleter debounce to 250ms

Increases the debounce to a more typical value (200-300ms) for a complex web application. The current value (50ms)* results in a large number of unnecessary API requests - on installatons with large numbers of work packages, this can lead to severe performance degradation.

* it's worth noting that most users don't even type this fast!

## Screenshots

https://github.com/user-attachments/assets/27c3f664-25e4-477f-8f0c-efafe248ab45

⚠️ **This PR also includes the patch from #17648** - without it, I cannot run the frontend tests locally.

# Merge checklist

- [ ] Added/updated tests

### DIFF
--- a/frontend/src/app/shared/components/autocompleter/op-autocompleter/op-autocompleter.component.ts
+++ b/frontend/src/app/shared/components/autocompleter/op-autocompleter/op-autocompleter.component.ts
@@ -23,8 +23,8 @@ import {
   ViewContainerRef,
 } from '@angular/core';
 import { DropdownPosition, NgSelectComponent } from '@ng-select/ng-select';
-import { BehaviorSubject, merge, NEVER, Observable, of, Subject, timer } from 'rxjs';
-import { debounce, distinctUntilChanged, filter, switchMap, tap, map } from 'rxjs/operators';
+import { BehaviorSubject, merge, NEVER, Observable, of, Subject } from 'rxjs';
+import { debounceTime, distinctUntilChanged, filter, switchMap, tap, map } from 'rxjs/operators';
 import { AddTagFn, GroupValueFn } from '@ng-select/ng-select/lib/ng-select.component';
 
 import { HalResource } from 'core-app/features/hal/resources/hal-resource';
@@ -228,6 +228,8 @@ export class OpAutocompleterComponent<T extends IAutocompleteItem = IAutocomplet
 
   @Input() public relations?:boolean = false;
 
+  @Input() public debounceTimeMs:number = 250;
+
   @Output() public open = new EventEmitter<unknown>();
 
   @Output() public close = new EventEmitter<unknown>();
@@ -283,8 +285,6 @@ export class OpAutocompleterComponent<T extends IAutocompleteItem = IAutocomplet
   projectedFooterTemplate:TemplateRef<Element>;
 
   footerTemplate:TemplateRef<Element>;
-
-  initialDebounce = true;
 
   private opAutocompleterService = new OpAutocompleterService(this.apiV3Service);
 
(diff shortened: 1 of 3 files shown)


frontend/src/app/shared/components/autocompleter/op-autocompleter/op-autocompleter.component.ts
```diff
@@ -522,14 +522,6 @@ export class OpAutocompleterComponent<T extends IAutocompleteItem = IAutocomplet
       );
   }
 
-  private getDebounceTimeout():number {
-    if (this.initialDebounce) {
-      this.initialDebounce = false;
-      return 0;
-    }
```
myabc, line 529: Removing the logic that disables the debounce on first user interaction - I don't think this behaviour makes sense (do we really want to make API calls for single letter query strings - in fact it would make more sense to me to only make requests when a minimum number of characters have been entered by the user)

cbliard: I vaguely remember that I added that to stabilize end-to-end tests. It was done in this commit: https://github.com/opf/openproject/commit/9aa36e54dbc4706041ce95badc6f859dfd69d429

Sadly, I don't remember much of it and I did not add a lot of comments or information in the commit message.

oliverguenther: could it have been related to the way `send_keys` work? on cuprite I think this directly sends a JS event

frontend/src/app/shared/components/autocompleter/op-autocompleter/op-autocompleter.component.ts
```diff
@@ -228,6 +228,8 @@ export class OpAutocompleterComponent<T extends IAutocompleteItem = IAutocomplet
 
   @Input() public relations?:boolean = false;
 
+  @Input() public debounceTimeMs:number = 250;
+
```
cbliard, line 232: Do you think you could make it configurable for feature tests? like make it 10 ms so that feature tests do not wait for the debounce?

cbliard (approved):
Seems good for me.
I reran the test workflow as a lot of them were failing.

Also wondering if the debounce value could be set to a lower value when running the feature tests.

frontend/src/app/shared/components/autocompleter/op-autocompleter/op-autocompleter.component.ts
cbliard: No, can't be. Cuprite was added on 21 Jun 2023 in 4ce2ebacd0b. The initial debounce was added one month earlier on 24 May 2023 in 9aa36e5.

I mean maybe it would help, but it was not the reason why it was added back then.

cbliard: Found it!

The loading indicator is set only after the first debounce completes. And in tests, we enter some text and wait for this loading indicator to disappear.

As it does not appear immediately, we do not wait for it, and so the test is out of sync with what is happening on screen.

That's probably why so many tests are failing on this PR.

cbliard: Found the reason in an old thread with Benjamin:

image
image
image

And that's what I did: I set it to 0ms, but only for the first character being entered, so that the debouncing behavior is preserved.

frontend/src/app/shared/components/autocompleter/op-autocompleter/op-autocompleter.component.ts
oliverguenther: Yes, this is possible by checking the `environment` of the initializer: https://github.com/opf/openproject/blob/dev/frontend/src/app/core/setup/globals/openproject.ts#L65-L67

`window.openProject.environment === 'test'`

myabc:
@cbliard I've made the change you suggested - the feature specs should now run without a debounce.

@oliverguenther I went ahead and fixed the ng tests for the autocompleter component - I've commented over on #17677 though.

cbliard:
@myabc thank you so much for setting debounce to 0ms in test env 🙇 

We may need to harmonize that debounce value throughout the app. It varies from component to component. (out of the scope of this PR)

oliverguenther (approved):

PR #17647, opened 2025-01-17
[#60666] User can't save lifecycle modal if project is invalid

dombesz:
# Ticket
https://community.openproject.org/work_packages/60666

Avoid validating the project custom fields when saving the project with a different contract than the `Projects::BaseContract` and its descendants. This decoupling resolves issues where custom field validations blocked the saving of objects associated with a project when they were being saved through the project.

A few example of such cases:
- Saving the life cycle steps happens via the project, but having an invalid custom field will prevent saving them. See https://community.openproject.org/wp/60666 .
- Saving the project settings `Show attachments in the work packages files tab` on the `/projects/<project_id>/settings/project_storages/attachments` page is also fails when there is an invalid custom field on the project. See https://community.openproject.org/wp/55789 .

Since the project custom field validations are happening on the model level, there was no way to avoid them even when using different contracts than `Projects::BaseContract`. Ideally the solution would be to move all the custom field validations to the contract layer of the application, this way we can have more control over where the validation is activated for each entity. Unfortunately that is a massive overhaul, because the `acts_as_customizable` plugin is used across all the application. Additionally, some models using the plugin are not even updated via a service or contract, meaning the model layer validations are mandatory for them.

A more feasible solution is to introduce context based validation on the custom fields. This way we can easily turn on and off custom field validations on the project model when it is being saved.

The custom field validations will remain enabled on all the models using the `acts_as_customizable` plugin. The following declaration `acts_as_customizable validate_on: :saving_custom_fields` will enable custom validation context on a model basis. This means calling the custom field validations are not triggered when calling `valid?` on the model, but they are triggered when calling `valid?(:saving_custom_fields)`.

Since we are saving custom fields only via the `Projects::BaseContract` descendants, we only have to enable the context based validations in this contract. This code will enable it:

```ruby
    def valid?(context = :saving_custom_fields)
      super
    end
```

As a result, the project custom field validations are only triggered when the project is saved via the `Projects::BaseContract` and they are not triggered when using other contracts such as the `Projects::SettingsContract` or the `ProjectLifeCycleSteps::BaseContract`.

# Merge checklist

- [x] Added/updated tests

### DIFF
--- a/app/contracts/project_life_cycle_steps/base_contract.rb
+++ b/app/contracts/project_life_cycle_steps/base_contract.rb
@@ -31,9 +31,7 @@ module ProjectLifeCycleSteps
     validate :select_custom_fields_permission
     validate :consecutive_steps_have_increasing_dates
 
-    def valid?(context = :saving_life_cycle_steps)
-      super
-    end
+    def valid?(context = :saving_life_cycle_steps) = super
 
     def select_custom_fields_permission
       return if user.allowed_in_project?(:edit_project_stages_and_gates, model)
--- a/app/models/project.rb
+++ b/app/models/project.rb
@@ -100,7 +100,8 @@ class Project < ApplicationRecord
 
   acts_as_favorable
 
-  acts_as_customizable # extended in Projects::CustomFields in order to support sections
+  acts_as_customizable validate_on: :saving_custom_fields
+  # extended in Projects::CustomFields in order to support sections
   # and project-level activation of custom fields
 
   acts_as_searchable columns: %W(#{table_name}.name #{table_name}.identifier #{table_name}.description),
--- a/app/models/project.rb
+++ b/app/models/project.rb
@@ -110,8 +110,8 @@ class Project < ApplicationRecord
   # Providing a custom context while creating the project, will not execute the callbacks on the
   # `:create` or `:update` contexts, meaning the identifier will not get initialised.
   # In order to initialise the identifier, the `default_validation_context` (`:create`, or `:update`)
-  # should be included when validating via the `:saving_custom_fields`. This way ever create or update
-  # callback will also be executed alongside the `:saving_custom_fields` callbacks.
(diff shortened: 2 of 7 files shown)


app/models/project.rb
```diff
@@ -100,9 +100,29 @@ class Project < ApplicationRecord
...
+  # extended in Projects::CustomFields in order to support sections
   # and project-level activation of custom fields
 
+  # Override the `validation_context` getter to include the `default_validation_context` when the
+  # context is `:saving_custom_fields`. This is required, because the `acts_as_url` plugin from
+  # `stringex` defines a callback on the `:create` context for initialising the `identifier` field.
+  # Providing a custom context while creating the project, will not execute the callbacks on the
+  # `:create` or `:update` contexts, meaning the identifier will not get initialised.
+  # In order to initialise the identifier, the `default_validation_context` (`:create`, or `:update`)
+  # should be included when validating via the `:saving_custom_fields`. This way ever create or update
```
ulferts: typo 'ever' -> 'every'

ulferts (approved):
All good and looking fine.

Super tiny remark on a typo, @dombesz . Feel free to merge after addressing it.

app/contracts/project_life_cycle_steps/base_contract.rb
```diff
@@ -31,9 +31,7 @@ class BaseContract < ::ModelContract
     validate :select_custom_fields_permission
     validate :consecutive_steps_have_increasing_dates
 
-    def valid?(context = :saving_life_cycle_steps)
-      super
-    end
+    def valid?(context = :saving_life_cycle_steps) = super
```
myabc, line 34: @dombesz genial! - perhaps we can add a Rubocop cop to prefer endless functions?

dombesz: Thanks @myabc ! Added it here https://github.com/opf/openproject/pull/17707 .

PR #17667, opened 2025-01-20
Bump Primer to 0.52.2 version

bsatarnejad:
# Ticket
https://community.openproject.org/wp/59860

Bump Primer to 0.52.2 version 
Use trailing units in WP create relation modal and meeting add and edit form.


modules/meeting/app/forms/meeting/time_group.rb
```diff
@@ -64,7 +64,7 @@ class Meeting::TimeGroup < ApplicationForm
         label: Meeting.human_attribute_name(:duration),
         visually_hide_label: false,
         required: true,
-        caption: I18n.t("text_in_hours"),
+        trailing_visual: { text: { text: I18n.t("text_hours") } },
```
myabc: Hi @bsatarnejad! There is already a translation in core that we can use here, I think: `datetime.units.hour`:

```yaml
    units:
      hour:
        one: "hour"
        other: "hours"
```

So you should be able to call `I18n.t("datetime.units.hour", count: 2)` or `I18n.t("datetime.units.hour.other")` (I'm not sure which is the more correct use of the I18n API though)

bsatarnejad: Hi @myabc 
Thanks for the hint 👍 
We may use units more in the future, so it would be better to define them generally in one place, not for a specific case.

modules/meeting/app/forms/meeting/time_group.rb
```diff
@@ -64,7 +64,7 @@ class Meeting::TimeGroup < ApplicationForm
         label: Meeting.human_attribute_name(:duration),
         visually_hide_label: false,
         required: true,
-        caption: I18n.t("text_in_hours"),
+        trailing_visual: { text: { text: I18n.t("datetime.units.hour.other") } },
```
HDinger: I see two issues here: 

Bildschirmfoto 2025-01-21 um 10 28 11

* The unit is shown twice (one time after the value, and one time at the end of the field)
* The unit at the end is truncated. 

So I'd suggest, to get rid of the unit after the value, and at the end use an abbreviated form like "h". We may have to ask product about that however

HDinger (changes requested):
The syntax `trailing_visual: { text: { text: ..` looks somehow weird, but since it is coming from Primer, we have to stick with that 🤷‍♀️

modules/meeting/app/forms/meeting/time_group.rb
myabc: 👍🏻 for "h"

bsatarnejad: Hi @myabc as discussed [here](https://community.openproject.org/projects/stream-planning-and-reporting/work_packages/59860/activity#activity-16), We continue with 'in hours' as caption for the input field.

PR #17671, opened 2025-01-20
use former version of chrome until chromedriver fixes issues with 132 in selenium

akabiru:
See https://github.com/opf/openproject/pull/17663


dombesz (approved):
Thanks Kabiru! 🙏

PR #17675, opened 2025-01-21
Bug/60718 activity field populates automatically

klaustopher:
# Ticket
https://community.openproject.org/projects/stream-time-and-costs/work_packages/60718

The `SetAttributeService` used to set a default activity. As we do not consistently use that and a default feels odd according to UX, we should remove it in that place as well

# Merge checklist

- [x] Added/updated tests

### DIFF
--- a/modules/costs/app/services/time_entries/set_attributes_service.rb
+++ b/modules/costs/app/services/time_entries/set_attributes_service.rb
@@ -58,7 +58,6 @@ module TimeEntries
     def set_default_attributes(*)
       set_default_user
       set_default_hours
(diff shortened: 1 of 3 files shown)


as-op (changes requested):
@klaustopher The activity is no longer set by default.

It seems that TimeEntry.activity was automatically set/mandatory before, as there are a few places without a check.

E.g.: 
search

The failing spec is one of them.

klaustopher:
Tests are green

as-op (approved):

as-op:
Approved. So I assume the specs without checks make sure themselves the activity is set.

PR #17678, opened 2025-01-21
Add token exchange capability to FetchService

NobodysNightmare:
We will exchange tokens if there is no token for
the given audience or if it's the only way to perform a refresh.

# Ticket
https://community.openproject.org/work_packages/60152/activity

Allowing to perform an OAuth 2.0 Token Exchange when that's necessary to obtain a token to be used at the target audience.

I integrated this into the existing `FetchService`, which now performs 2-3 things, depending on how you count:
* Looking up a token (a glorified `where` statement)
* Refreshing the access token if expired
* Exchanging the token if it can't be found **or** it's expired, but there is no refresh token

I chose to integrate all of these aspects into a single service, because it makes sense from a callers perspective, where you just want to have a token and you want to use it. However, I also found it hard to separate these aspects, because they depend on each other. Token Exchange is not only needed for tokens that can't be found, but also for tokens that need to be refreshed. And to perform a token exchange, we need to fetch the correct token first.

# Merge checklist

- [x] Added/updated tests for the entire behaviour
- [ ] Splitting tests according to units, after splitting into 3 services

### DIFF
--- a/modules/openid_connect/app/services/openid_connect/user_tokens/exchange_service.rb
+++ b/modules/openid_connect/app/services/openid_connect/user_tokens/exchange_service.rb
@@ -32,6 +32,7 @@ module OpenIDConnect
   module UserTokens
     class ExchangeService
       include Dry::Monads[:result]
+      include Dry::Monads::Do.for(:call)
 
       class Disabled
         class << self
@@ -50,19 +51,17 @@ module OpenIDConnect
       def call(audience)
         return Failure("Provider does not support token exchange") unless supported?
 
-        FetchService.new(user: @user, token_exchange: Disabled)
-                    .access_token_for(audience: UserToken::IDP_AUDIENCE)
-                    .bind do |idp_token|
-                      exchange_token_request(idp_token, audience).bind do |json|
-                        access_token = json["access_token"]
-                        refresh_token = json["refresh_token"]
-                        break Failure("Token exchange response invalid") if access_token.blank?
+        idp_token = yield FetchService.new(user: @user, token_exchange: Disabled)
+                            .access_token_for(audience: UserToken::IDP_AUDIENCE)
 
-                        token = store_exchanged_token(audience:, access_token:, refresh_token:)
+        json = yield exchange_token_request(idp_token, audience)
 
-                        Success(token)
-                      end
-                    end
+        access_token = json["access_token"]
+        refresh_token = json["refresh_token"]
+        return Failure("Token exchange response invalid") if access_token.blank?
+
+        token = store_exchanged_token(audience:, access_token:, refresh_token:)
+        Success(token)
       end
 
       def supported?
--- a/modules/openid_connect/app/services/openid_connect/user_tokens/exchange_service.rb
+++ b/modules/openid_connect/app/services/openid_connect/user_tokens/exchange_service.rb
@@ -1,0 +1,109 @@
+# frozen_string_literal: true
+
+#-- copyright
+# OpenProject is an open source project management software.
+# Copyright (C) 2012-2024 the OpenProject GmbH
+#
+# This program is free software; you can redistribute it and/or
+# modify it under the terms of the GNU General Public License version 3.
+#
+# OpenProject is a fork of ChiliProject, which is a fork of Redmine. The copyright follows:
+# Copyright (C) 2006-2013 Jean-Philippe Lang
+# Copyright (C) 2010-2013 the ChiliProject Team
+#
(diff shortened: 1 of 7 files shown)


NobodysNightmare:
Since the separation of concerns topic already came up as a question in the daily today, I'll make another shot at separating involved services more cleanly. Let's see how this will work out.

edit: This has been done

modules/openid_connect/app/services/openid_connect/user_tokens/fetch_service.rb
```diff
@@ -37,9 +37,13 @@ module UserTokens
     class FetchService
       include Dry::Monads[:result]
 
-      def initialize(user:, jwt_parser: JwtParser.new(verify_audience: false, verify_expiration: false))
+      def initialize(user:,
+                     jwt_parser: JwtParser.new(verify_audience: false, verify_expiration: false),
+                     token_exchange: ExchangeService.new(user:),
+                     token_refresh: RefreshService.new(user:, token_exchange:))
```
Kharonus, line 44: 🟡 Seeing constructors like this makes me feel itchy. Should we consider any kind of IoC here? like Dry::AutoInject?

Also peeking into the direction of @mereghost

modules/openid_connect/app/services/openid_connect/user_tokens/exchange_service.rb
```diff
@@ -0,0 +1,109 @@
...
+#
+# See COPYRIGHT and LICENSE files for more details.
+#++
+
+module OpenIDConnect
+  module UserTokens
+    class ExchangeService
+      include Dry::Monads[:result]
+
+      class Disabled
```
Kharonus, line 37: 🟡 This also hints into the direction of dependency injection.

So, depending on the condition, either the `ExchangeService` or the `DisabledExchangeService` is injected.

modules/openid_connect/app/services/openid_connect/user_tokens/exchange_service.rb
```diff
@@ -0,0 +1,109 @@
...
+      end
+
+      def initialize(user:)
+        @user = user
+      end
+
+      def call(audience)
+        return Failure("Provider does not support token exchange") unless supported?
+
+        FetchService.new(user: @user, token_exchange: Disabled)
```
Kharonus: 🟡 a do notation would make this piece of code less indented and IMHO better readable

Kharonus (changes requested):
In combination with the base branch I'd like to raise the discussion of IoC and a bit polishing on the monad coding style.

both topics certainly can be extracted out of this PR into a following refactoring.

modules/openid_connect/app/services/openid_connect/user_tokens/fetch_service.rb
NobodysNightmare: To be fair: I would have probably not added `token_exchange` and `token_refresh` as parameters at all, but implemented them as private methods and then in specs I'd have mocked the `new` method of them, but @mereghost told me having a dependency injecting constructor would be preferrable to that.

In this specific case, I was at least able to use the DI for production code as well, by implementing `ExchangeService::Disabled`, which is a way to avoid passing an additional `enable_token_exchange` parameter.

NobodysNightmare: I just checked the code example of [dry-autoinject](https://dry-rb.org/gems/dry-auto_inject/0.6/). I understand that we use it in certain places (e.g. our storage queries), but to me it feels like a too big gun to use here ("Kanonen auf Spatzen").

If the problem we wanted to solve was, that it's somehow bad practice to mock internal implementation details of a class (e.g. knowing that it would internally call `OtherService.new`), then I can understand that making the injection of said class a part of the class's public interface (e.g. by passing `OtherService.new`  as an argument).

However, I start to become skeptical, when the solution to now having to pass a class in via an argument, is to use a library that makes the internally used class a secret again (it would be "hidden" in a private method again), but allows to change it from another place (e.g. by changing the outside "container").

That at least does not contribute to clarity IMO. So the question is, which problem we are solving by doing that.

Kharonus: Don't get me wrong, the constructor already is prepared for ctor injection. I was just confused with the default parameters.

Okay, so IoC was discussed before, which makes me rest easier, that we this code is prepared to maybe even go another step into that direction.

Kharonus (approved):
Same here, should we merge this as our first shot at the service layer?

PR #17692, opened 2025-01-22
Feature/59925 show user popover on the name

EinLama:
# Ticket
https://community.openproject.org/wp/59925

Enable the hover card not only on avatars, but also when hovering on names.

- [x] patch AvatarComponent so that names rendered there will display a hover card.
- [x] search application for usernames that show up outside of the AvatarComponent
  - [x] linked names in WP journal
  - [x] within auto completers
  - [x] linked names in WP description and comments
  - [x] news
  - [x] wiki
  - [x] admin: users
  - [x] project: members

**Notable technical exceptions** where the hover card will not render due to technical limitations of our spot modal usage:

- work package share dialog
- global search
- global invitation modal

I am currently working on removing these limitations in a follow-up PR, but wanted to make the current state available for review.

**Notable other exceptions** where the required effort to patch in hover cards was not worth it in my opinion:

- the news block on the my-page: it uses a very different way of constructing links. I skipped it since the necessary increase in complexity is not worth it for a single widget.
- some username entries in WP journals that are not links, but plain text. This happens for example when the activity describes someone setting a new assignee on a work package. The new assignee will not be a link to a user, it will just be the username instead. Again skipped due to complexity vs. gain.

## Screenshots

https://github.com/user-attachments/assets/0a084df8-7bf5-4d58-b7c7-555ca5f787d5

I applied the same logic to usernames as I did on the avatars. ~For some user names (e.g. within auto completers), there was no easy way to inject the necessary parameters into the data attributes of the user name span, so I instead introduced a new CSS class: `.op-hover-card--data-on-sibling-principal`
If a hover card is triggered, but the source element has this class, the hover card data is sourced from sibling elements if possible.~

To cover most other links, I extended `ObjectLinking#link_to_user` to add hover card info to the link.

Fixed some timing and consistency bugs while on it, too.

# Merge checklist

- [x] Added/updated tests

### DIFF
--- a/frontend/src/app/core/setup/globals/global-listeners/hover-card-trigger.service.ts
+++ b/frontend/src/app/core/setup/globals/global-listeners/hover-card-trigger.service.ts
@@ -58,9 +58,13 @@ export class HoverCardTriggerService {
     jQuery(document.body).on('mouseover', '.op-hover-card--preview-trigger', (e) => {
       e.preventDefault();
       e.stopPropagation();
+
       // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment
       const el = e.target as HTMLElement;
       if (!el) { return; }
+
+      // For some elements, the hover card data is on another element. We need to find that element.
+      const dataSource = this.getDataSource(el);
 
       if (this.previousTarget && this.previousTarget === el) {
         // Re-entering the trigger counts as hovering over the card:
@@ -73,8 +77,7 @@ export class HoverCardTriggerService {
       this.close(true);
       this.previousTarget = el;
 
-      // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment
-      const turboFrameUrl = el.getAttribute('data-hover-card-url');
+      const turboFrameUrl = this.parseHoverCardUrl(dataSource);
       if (!turboFrameUrl) { return; }
 
       // Reset close timer for when hovering over multiple triggers in quick succession.
@@ -87,7 +90,7 @@ export class HoverCardTriggerService {
 
       // Set a delay before showing the hover card
       this.hoverTimeout = window.setTimeout(() => {
-        this.showHoverCard(el, turboFrameUrl, e);
+        this.showHoverCard(dataSource, turboFrameUrl, e);
       }, this.OPEN_DELAY_IN_MS);
     });
 
@@ -159,6 +162,45 @@ export class HoverCardTriggerService {
       // Allow opening this target once more, since it has been orderly closed
       this.previousTarget = null;
     }
+  }
+
+  private getDataSource(el:HTMLElement) {
+    if (el.className.includes('op-hover-card--data-on-sibling-principal')) {
+      const candidate = this.findSiblingElementWithHoverCardData(el);
+      if (candidate) {
+        return candidate as HTMLElement;
+      }
+    }
+
+    return el;
+  }
+
+  private findSiblingElementWithHoverCardData(el:HTMLElement) {
+    let sibling = el.previousElementSibling;
+    while (sibling) {
+      if (sibling.classList.contains('op-principal')) {
+        return this.childElementWithHoverCardTrigger(sibling);
+      }
+      sibling = sibling.previousElementSibling;
+    }
+
+    sibling = el.nextElementSibling;
+    while (sibling) {
+      if (sibling.classList.contains('op-principal')) {
+        return this.childElementWithHoverCardTrigger(sibling);
+      }
+      sibling = sibling.nextElementSibling;
+    }
+
+    return null;
+  }
+
+  private childElementWithHoverCardTrigger(principal:Element) {
+    return principal.querySelector('.op-hover-card--preview-trigger');
+  }
+
+  private parseHoverCardUrl(el:HTMLElement) {
+    return el.getAttribute('data-hover-card-url');
   }
 
   private parseHoverCardOptions(el:HTMLElement) {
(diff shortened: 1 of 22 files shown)


frontend/src/app/core/setup/globals/global-listeners/hover-card-trigger.service.ts
```diff
@@ -154,13 +164,50 @@ export class HoverCardTriggerService {
...
+    return null;
+  }
+
+  private childElementWithHoverCardTrigger(principal:Element) {
+    return principal.querySelector('.op-hover-card--preview-trigger');
+  }
+
+  private parseHoverCardUrl(el:HTMLElement) {
+    return el.getAttribute('data-hover-card-url');
+  }
```
HDinger: You rely heavily on a certain structure for this to work, but that is not clear when you just see the class. So whenever somewhen changes the structure in the autocompleter and e.g wrap the principal in a div, that will not work any more without the person noticing it.

I don't really get why all of this is necessary. You wrote that you could not add the hoverCard information directly on the element but why not?

HDinger (changes requested):
Hi @EinLama 

I don't quite get the necessity for the sibling logic. Maybe we can shortly talk about that. Other than that, I found some issues while testing it:

- [ ]  I managed quite often to get to a state where either the name or the avatar does not trigger a hoverCard any more
Jan-31-2025 09-07-10

- [ ] There seems to an issue with the helper, but I could not quite get my finger on it. The WP info line shortly shows the author as a link before it switches to a text. That does not happen with all users however, I only saw it when the author is my current logged in user. However, even when there is a link shown, there is no hoverCard when hovering it.

Jan-31-2025 09-02-19

- [x] In the acceptance criteria it is said, that there should be no flicker when switching between name and avatar. The card is however disappearing when I leave the name and reopens when hovering the avatar. Due to the delay it does not feel flickering but I'd neverless like to hear the opinion of the product team on this.
Jan-31-2025 09-04-03

frontend/src/app/core/setup/globals/global-listeners/hover-card-trigger.service.ts
EinLama: We had a call about this and Henriette suggested a great way to overcome the obstacles I found while trying to place hovercard data directly onto the span. With these suggestions, all of the complicated methods above could be removed 🎉

EinLama:
Hey @HDinger,

thank you for your review. You already helped me get rid of the unnecessary bloat fetching data from a sibling element. So that one is done.

> I managed quite often to get to a state where either the name or the avatar does not trigger a hoverCard any more

I could reproduce this by "jittering" the mouse cursor over the trigger. The hover card trigger service has a safe-guard that - once shown - will not re-show the same hover card again. This safe-guard was too restrictive and did not consider that you could move the mouse cursor away before the hover card is actually shown. So the trigger service thought it had to safe guard a hover card that wasn't there yet. I moved the safe-guard to the appropriate position and that fixed it. If the cause for your observed behavior was the same, that should be fixed, too.

>  There seems to an issue with the helper, but I could not quite get my finger on it. The WP info line shortly shows the author as a link before it switches to a text. That does not happen with all users however, I only saw it when the author is my current logged in user. However, even when there is a link shown, there is no hoverCard when hovering it.

I could not reproduce this, nor have I ever seen it 🤔 

>  In the acceptance criteria it is said, that there should be no flicker when switching between name and avatar. The card is however disappearing when I leave the name and reopens when hovering the avatar. Due to the delay it does not feel flickering but I'd neverless like to hear the opinion of the product team on this.

Will discuss later today and report back.

EinLama:
The hover card disappearing briefly when alternating between username and avatar [has been discussed ](https://community.openproject.org/meetings/4251#item-9333)with product and is deemed okay 👍🏻

frontend/src/app/shared/components/autocompleter/user-autocompleter/user-autocompleter-template.component.sass
```diff
@@ -11,3 +11,8 @@
       display: block
       margin-left: var(--control-small-size)
       margin-top: var(--control-xsmall-gap)
+
+// For hover cards to work within the auto completer, we have to ensure that the hover event can be properly
+// triggered when mousing over the name. The easiest way to do this is to set it to inline-block.
+.op-user-autocompleter--name.op-hover-card--preview-trigger
+  display: inline-block
```
HDinger: I don't quite understand why this is needed at all but it causes problems with long names which are simply cut off:

Bildschirmfoto 2025-02-04 um 08 59 32

HDinger (changes requested):
I only have small remarks left:

* The author of a WP is now correctly shown as link (in the little info row next to the status). However, hovering over it does not trigger a HoverCard
* When opening a user autocompleter (e.g to set an assignee), I can also see groups. Hovering over a group name triggers an empty hovercard which lays behind the dropdown. There should be no card at all

Bildschirmfoto 2025-02-04 um 08 58 47

frontend/src/app/shared/components/autocompleter/user-autocompleter/user-autocompleter-template.component.sass
EinLama: I added this property since I encountered weird behavior. When you hovered over a name in an autocompleter, the hover card would show up. But as soon as you traced over a space between words, it would disappear - only to re-appear shortly thereafter when you leave the space and enter the next word.

However, this no longer happens now. So there must have been another in-development-state present back then that was the cause for this buggy behavior. Thanks for questioning this, I can remove this hack!

EinLama:
> The author of a WP is now correctly shown as link (in the little info row next to the status). However, hovering over it does not trigger a HoverCard

Thank you, fixed!

> When opening a user autocompleter (e.g to set an assignee), I can also see groups. Hovering over a group name triggers an empty hovercard which lays behind the dropdown. There should be no card at all

That was a side-efffect of me removing the sibling logic. Fixed now so that hover cards are only shown for principals of type user.

EinLama:
The red spec is flaky, have seen this a couple of times 😵

HDinger (approved):
Nice job 👍 🌟

PR #15779, opened 2024-06-07
[55466] Display progress units in days and hours

cbliard:
https://community.openproject.org/wp/55466

To reduce problems with displaying too many units in too narrow columns, the only units being displayed will be days and hours.
As weeks and months are not displayed anymore, in administration > calendar and dates > working days and hours, the settings "days per week" and "days per month" have been removed as well.

PR #17694, opened 2025-01-22
[#60911] Rehabilitate frontend component test suite

myabc:
# Ticket

https://community.openproject.org/wp/60911 (part of Feature https://community.openproject.org/wp/60910)

**Goals:**

- Get the current frontend test suite running green once again.
- CI: Create an initial GitHub workflow for the frontend test suite.

This is an alternative to #17677, which removes the frontend test suite entirely.

## Screenshots

Test suite results with code coverage

This Pull Request initially started out as a spike - to see how easy it would be to restore CI runs for the suite (i.e. setting up a GitHub workflow action). However, once #17648 and #17646 were merged, only ~10 specs out of 450 were failing. As such, I went ahead and have attempted to fix these failures.

## GitHub workflow for frontend component test suite

This is as a very simple, "vanilla" workflow action. It relies on the included software in GitHub's Ubuntu 24.04 runner image + `setup-node@v4` action to setup the environment and execute the spec suite.

The most significant bit of work was to port the Rake `openproject:plugins:register_frontend` task to a node script. This was done so that we don't need a step that sets up ruby, installs OpenProject dependencies. 

Things to note:

- The port does not implement `generate_plugin_sass`, as AFAICT global styling isn't necessary for the frontend test suite to run successfully.
- For now **this hardcodes the list of plugins that we test** - _I am not terribly familiar with how we're doing plugin management atm, so not sure if this is problematic?_.

## Spec fixes

The failures in the frontend test suite are a result of the test suite having become out-of-sync with the actual implementation - which is understandable since it hasn't been running on CI for a number of years, and therefore overlooked.

The reasons for the failures and how I fixed them can basically be summarised as follows:

1. **stub data that was out-of-sync with the implementation/didn't confirm to TS interfaces**: updated stubs, added explicit types where possible.
2. **CSS selector changes due to component template changes**: updated CSS selectors.
3. **CSS selector changes due to upgrading Formly from 5.0 to 6.0** (see https://formly.dev/docs/guide/migration): updated CSS selectors. This PR also replaces the  deprecated use of `FormlyTemplateOptions` with `FormlyFieldProps` - _however there is some additional work that needs to be done to properly deal with Formly 6.x deprecations. I have broken this out into separate work package, [op#60912](https://community.openproject.org/wp/60912)._
4. **an issue with `CKEditorSetupService#load()` use of dynamic ESM imports**: I spent a fair amount of time playing around with `angular.json`/`karma.config.js` settings, even going as far as debugging the Webpack config that `ng test` generates under-the-hood. However I struggled to get past "chunk load errors", so ended up replacing replacing `CKEditorSetupService#create()` with a spy that returns a stub `ICKEditor` implementation. I think this is probably a better solution going forward.

To make things easier for review - I have referenced the commits that most likely caused the spec failures in several of this PR's commit messages. I'll also list the related PRs in the checklist below - it would be really fantastic if you could have a quick look at the associated spec changes.

# Merge checklist

- [x] Added/updated tests
- [ ] Review from related PR authors:
  - [ ] #12375 (@oliverguenther, reviewed by @cbliard)
  - [ ] #15098 (@HDinger, reviewed by @oliverguenther)
  - [ ] #15779 (@cbliard, reviewed by @dombesz)
  - [ ] #12834 (@oliverguenther), which also bumps Formly
- [ ] Make time-zone service spec more robust across timezones (currently failing locally in UTC-3)
- [ ] Fix dynamic-form component spec warnings ([listed in this gist](https://gist.github.com/myabc/ff686dea44641013747303428c356fdd)) (strangely these only appear when the entire suite and not when running the individual spec with `fdescribe`)
- [ ] Check code coverage excludes to ensure accurate stats.
- [ ] Decide how to report code coverage in a useful way, e.g. appending stats to PR? (ChatGPT has [some ideas over here](https://chatgpt.com/share/67954e28-a6e0-8006-80e5-609dfb2afbcd))

Additional tasks (might be out-of-scope)

- [ ] Support testing additional browsers (Firefox) (?)
- [ ] Decide if we want to skip long-running "Test suite" (`test-core.yml`) workflow if frontend test suite check fails*

----

_* I believe this is possible with `workflow_run`, but @crohr / @machisuji probably know more? If it is technically possible, we should also figure out if introducing workflow dependencies makes sense (as part of [op#60910](https://community.openproject.org/wp/60910)), especially given that the frontend test suite does not currently test any non-Angular TS/JS code._


.github/workflows/test-frontend-unit.yml
```diff
@@ -0,0 +1,54 @@
...
+on:
+  workflow_dispatch:
+  push:
+    branches:
+      - dev
+      - release/*
+    paths:
+      - '**/frontend/**/*.ts'
+      - '**/frontend/**/*.js'
+      - '**/frontend/**/*.json'
```
myabc, line 12: It would be great if someone could check these filter patterns. I was caught out at first by GH Actions lack of support for extended globbing (see [[Filter pattern cheat sheet](https://docs.github.com/en/actions/writing-workflows/workflow-syntax-for-github-actions#filter-pattern-cheat-sheet)).

.github/workflows/test-frontend-unit.yml
```diff
@@ -0,0 +1,54 @@
...
+    runs-on: ubuntu-latest
+  
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+          
+      - uses: actions/setup-node@v4
+        with:
+          node-version: '18.13'
```
myabc: I specified `18.13` because it's the same version we're using in `eslint-core.yml`. However we should probably switch both to `20.9.0`, the node version specified in `docker/ci/Dockerfile`, as well as the root `package.json` in the repo.

.github/workflows/test-frontend-unit.yml
```diff
@@ -0,0 +1,54 @@
...
+  contents: read
+
+defaults:
+  run:
+    working-directory: ./frontend
+
+jobs:
+  units:
+    name: Units
+    runs-on: ubuntu-latest
```
myabc, line 31: `ubuntu-latest` resolves to GitHub's Ubuntu 24.04 runner image, which [includes Chrome, chromedriver amongst other browsers and drivers](https://github.com/actions/runner-images/blob/main/images/ubuntu/Ubuntu2404-Readme.md#browsers-and-drivers).

.github/workflows/test-frontend-unit.yml
```diff
@@ -0,0 +1,54 @@
...
+  
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
```
myabc, line 36: I copied and pasted this from `eslint-core.yml`, but I think this isn't what we want - according to the docs this [fetches all history for all tags and branches](https://github.com/actions/checkout?tab=readme-ov-file#fetch-all-history-for-all-tags-and-branches). / cc @machisuji

.github/workflows/test-frontend-unit.yml
```diff
@@ -0,0 +1,54 @@
...
+        id: npm-i
+        run: npm i
+
+      - name: Register plugins
+        id: npm-run-ci-plugins-register-frontend
+        run: npm run ci:plugins:register_frontend
+
+      - name: Test (Angular)
+        id: npm-test
+        run: npm test -- --code-coverage
```
myabc, line 54: Although test coverage results are written to `stdout` (so visible in the job output), we should probably find a way to append the report to the PR somehow - @machisuji @cbliard @crohr any ideas what our options might be?

frontend/karma.conf.js
```diff
@@ -8,28 +8,36 @@ module.exports = function (config) {
...
       // For testing, shim its functionality
-      'src/test/i18n-shim.ts',
       'node_modules/jquery/dist/jquery.js',
       // 'node_modules/angular-mocks/angular-mocks.js'
     ],
     frameworks: ['jasmine', '@angular-devkit/build-angular'],
     plugins: [
       require('karma-jasmine'),
       require('karma-chrome-launcher'),
+      require('karma-coverage'),
```
myabc, line 18: From glancing at the source for `@angular-devkit/build-webpack`, `ng test` should append this plugin automatically if `"test"`/`"options"`/`"codeCoverage": true` is specified. However I couldn't get it to work.

frontend/src/app/shared/components/dynamic-forms/components/dynamic-inputs/formattable-textarea-input/components/formattable-control/formattable-control.component.ts
```diff
@@ -43,13 +42,13 @@ export class FormattableControlComponent implements ControlValueAccessor, OnInit
...
-      type: this.templateOptions.editorType as ICKEditorType,
-      field: this.templateOptions.name as string,
+      type: this.templateOptions.editorType!,
+      field: this.templateOptions.name,
       // This is a very project resource specific hack to allow macros on description and statusExplanation but
       // disable it for custom fields. As the formly based approach is currently limited to projects, and that is to be removed,
       // such a "pragmatic" approach should be ok.
-      macros: (this.templateOptions.property as string).startsWith('customField') ? 'none' : 'resource',
-      options: { rtl: this.templateOptions?.rtl },
+      macros: this.templateOptions.property?.startsWith('customField') ? 'none' : 'resource',
```
myabc, line 50: safe navigation operator is necessary here!

frontend/src/app/shared/components/dynamic-forms/typings.d.ts
```diff
@@ -15,16 +16,34 @@ export interface IOPFormlyFieldSettings extends FormlyFieldConfig {
...
+  noWrapLabel?:boolean;
+  virtualScroll?:boolean;
+  clearOnBackspace?:boolean;
+  clearSearchOnAdd?:boolean;
+  hideSelected?:boolean;
+  text?:Record<string, string>;
+  typeahead?:boolean;
+  inlineLabel?:boolean;
+  clearable?:boolean;
+  multiple?:boolean;
```
myabc, line 46: `FormlyFieldConfig` is now a generic interface with a `Props` type parameter. As stated in the associated commit, this could (and perhaps should) be reworked as a an interface hierarchy (+ discriminated unions), defining only the applicable properties for a given `OPInputType` or template type. Could possibly be done as part of [op#60912](https://community.openproject.org/wp/60912).

modules/github_integration/frontend/module/tab-prs/tab-prs.component.spec.ts
```diff
@@ -109,11 +149,11 @@ describe('TabPrsComponent', () => {
     fixture = TestBed.createComponent(TabPrsComponent);
     component = fixture.componentInstance;
     element = fixture.debugElement;
-    halResourceService = fixture.debugElement.injector.get(HalResourceService) as jasmine.SpyObj<HalResourceService>;
     changeDetectorRef = fixture.debugElement.injector.get(ChangeDetectorRef) as jasmine.SpyObj<ChangeDetectorRef>;
     // @ts-ignore
     component.workPackage = { id: 'testId' };
 
+    changeDetectorRef.markForCheck();
```
myabc, line 158: manually invoking change detection gets this spec passing - not sure if this is the canonical way of testing asynchronous in Angular tests though?

modules/gitlab_integration/frontend/module/git-actions/git-actions.service.ts
```diff
@@ -41,8 +41,8 @@ export class GitActionsService {
     return str
       .replace(/&/g, 'and ') // & becomes and
       .replace(/ +/g, '-') // Spaces become dashes
-      .replace(/[\000-\039]/g, '') // ASCII control characters are out
```
myabc, line 44: `\039` isn't a valid Octal - cool TS 5.5 validates regexen!

frontend/src/app/shared/helpers/drag-and-drop/drag-and-drop.service.ts
```diff
@@ -4,6 +4,7 @@ import {
 import { DOCUMENT } from '@angular/common';
 import { DomAutoscrollService } from 'core-app/shared/helpers/drag-and-drop/dom-autoscroll.service';
 import { findIndex, reinsert } from 'core-app/shared/helpers/drag-and-drop/drag-and-drop.helpers';
+import dragula from 'dragula';
```
myabc: Need to double-check this - Dragula is required with the `expose-loader` in `init-vendors.ts`, which we're not currently importing in `test.ts`. We are also not importing `typings/shims.d.ts`.

.github/workflows/test-frontend-unit.yml
myabc: It might be better to explicitly specify `ubuntu-24.04`:
image

oliverguenther:
@myabc  Feature tests are broken due to 

> Error: Failed to initialize Angular compilation - The Angular Compiler requires TypeScript >=5.2.0 and <5.5.0 but 5.5.4 was found instead.

.github/workflows/test-frontend-unit.yml
oliverguenther: We had this limited to something like ~50, which would fail for longer running PRs. There is a magic cutoff probably, but the impact is not that big as the checkout action is cached.

frontend/src/app/shared/components/dynamic-forms/typings.d.ts
oliverguenther: We don't want to spend any time into Formly, as the project form will be replaced by Primer in the near future. As we now have custom fields support in Primer forms, doing that would be very easy already

.github/workflows/test-frontend-unit.yml
crohr: @myabc you can at least append it as a [job summary](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/workflow-commands-for-github-actions#adding-a-job-summary). Or use one of the existing actions that posts/edits a comment to include any body you would like.

myabc:
@oliverguenther
 
> > Error: Failed to initialize Angular compilation - The Angular Compiler requires TypeScript >=5.2.0 and <5.5.0 but 5.5.4 was found instead.

as just discussed, I've split off the TypeScript 5.5-related commits into a separate branch/PR: #17733.

.github/workflows/test-frontend-unit.yml
oliverguenther: I don't think we depend on the specific version here, so let's keep this as is

.github/workflows/test-frontend-unit.yml
oliverguenther: Let's merge this as is, we can improve the output when extending test suite to include stimulus (which might result in us using a different test runner overall)

oliverguenther (approved):
Looking great! Thanks for the efforts

PR #17696, opened 2025-01-23
bug/60719 disable wp comment submitter while the form request is in progress

akabiru:
# Ticket

https://community.openproject.org/work_packages/60719

Prevent possible repeat submissions that can arise from slow network connections, and the user double clicking the comment submit button

## Screenshots

_*Clicking the submit button twice (quickly)_

_Before_

https://github.com/user-attachments/assets/a83a580e-1d9b-4e46-a88f-2557d34188ea

_After_

https://github.com/user-attachments/assets/39333886-6209-41e9-8097-b47021b00d39

Add submit button targets via StimulusJS to manually enable and disable the submit button onSubmit.

[Turbo does automate this when using vanilla forms](https://turbo.hotwired.dev/reference/attributes#automatically-added-attributes) natively, but since we bypass the submit request to a turbo stream- it does not apply.

# Merge checklist

- [ ] ~Added/updated tests~ I'm not sure the tests are easy to automate here

### DIFF
--- a/frontend/src/stimulus/controllers/dynamic/work-packages/activities-tab/index.controller.ts
+++ b/frontend/src/stimulus/controllers/dynamic/work-packages/activities-tab/index.controller.ts
@@ -24,13 +24,16 @@ export default class IndexController extends Controller {
     showConflictFlashMessageUrl: String,
   };
 
-  static targets = ['journalsContainer', 'buttonRow', 'formRow', 'form', 'reactionButton'];
+  static targets = ['journalsContainer', 'buttonRow', 'formRow', 'form', 'formSubmitButton', 'reactionButton'];
 
   declare readonly journalsContainerTarget:HTMLElement;
   declare readonly buttonRowTarget:HTMLInputElement;
   declare readonly formRowTarget:HTMLElement;
   declare readonly formTarget:HTMLFormElement;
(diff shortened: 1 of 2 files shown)


frontend/src/stimulus/controllers/dynamic/work-packages/activities-tab/index.controller.ts
```diff
@@ -661,10 +664,18 @@ export default class IndexController extends Controller {
         console.error('Error saving activity:', error);
       })
       .finally(() => {
-        this.saveInProgress = false;
+        this.formSubmitInProgress = false;
       });
   }
 
+  private set formSubmitInProgress(inProgress:boolean) {
```
brunopagno: :green_circle: I think semantically it would make more sense to me if we used a regular function instead of a setter here. The reason being that it's not obvious that when we do the `formSubmitInProgress = value` that something else is happening. It looks like a flag.

No biggie though, feel free to ignore.

brunopagno (approved):
Looks good to me :clap: 

Nice video captures <3

frontend/src/stimulus/controllers/dynamic/work-packages/activities-tab/index.controller.ts
akabiru: Ah, makes sense to me! Updating...

PR #17703, opened 2025-01-23
Make migration reversible

NobodysNightmare:
Allowing to revert the migration, in case it's needed (e.g. because you want to rollback the migration before it).

This migration itself is something that you probably don't need to revert, but it can make life easier to be able to revert it.


myabc (approved):
LGTM! - thanks!

PR #17704, opened 2025-01-23
[#60144] primerize wp types settings form

Kharonus:
# Ticket
[OP#60144](https://community.openproject.org/work_packages/60144)

- harmonize look and feel of work package type forms

- remove old form for new and edit
- add primer form component


app/forms/work_packages/types/settings_form.rb
```diff
@@ -0,0 +1,108 @@
...
+# along with this program; if not, write to the Free Software
+# Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+#
+# See COPYRIGHT and LICENSE files for more details.
+#++
+
+module WorkPackages
+  module Types
+    class SettingsForm < ApplicationForm
+      form do |subject_form|
```
NobodysNightmare: :yellow_circle: Probably a copy & paste leftover, but this is the `settings_form`, not the `subject_form`.

NobodysNightmare (approved):
Changes look (literally) nice.

(Obviously assuming the failing spec will be fixed)

mereghost (approved):

app/forms/work_packages/types/settings_form.rb
Kharonus: ❗ nooo, how can this even happen ❗

app/views/types/new.html.erb
```diff
@@ -39,8 +39,10 @@ See COPYRIGHT and LICENSE files for more details.
...
-<%= error_messages_for 'type' %>
-
-<%= form_for controller.type, builder: TabularFormBuilder, lang: current_language  do |f| %>
-  <%= render partial: 'types/form/settings', locals: { f: f } %>
-<% end %>
+<%=
+  render WorkPackages::Types::SettingsComponent.new(
+    @type,
+    copy_workflow_from: params.dig(:type, :copy_workflow_from),
+    params:
```
NobodysNightmare: :red_circle: Passing `params` into the component is a debugging left-over, isn't it?

app/components/work_packages/types/settings_component.rb
```diff
@@ -0,0 +1,52 @@
...
+#++
+
+module WorkPackages
+  module Types
+    class SettingsComponent < ApplicationComponent
+      include OpPrimer::ComponentHelpers
+      include OpTurbo::Streamable
+
+      def initialize(model, **options)
+        @copy_workflow_from = options[:copy_workflow_from].to_i
```
NobodysNightmare: :green_circle: Maybe it's just me, but this `to_i` feels wrong to me. In the end it's just a matter of passing the value through from the outside, but maybe I'd prefer if the conversion took place outside of `SettingsComponent` (in the place that fills this from `params`, where we know that things are only _stringly_ typed.

Also as an aside: We _could_ make use of named arguments in this method. Something along the lines of:

```ruby
def initialize(model, copy_workflow_from: nil, **opts)
  # <--- use copy_workflow_from here
  super(model, **opts) # bonus: The superclass constructor will not receive the named argument that we consumed
end
```

NobodysNightmare (changes requested):

app/components/work_packages/types/settings_component.rb
mereghost: I'd add that I'd prefer to use `Integer` but that will raise an error rather than failing gracefully.

I'm with @NobodysNightmare here. I'd prefer to have the input sanitized beforehand.

app/components/work_packages/types/settings_component.rb
```diff
@@ -0,0 +1,52 @@
...
+        super
+      end
+
+      def form_options
```
mereghost, line 42: 🟢 I suspect that I probably was the first to make this option hashes with ternaries, but I think that having them as separate methods will make the intent a bit clearer.

```ruby
def form_options
  return create_form_options if model.new_record?
  
  update_form_options
end

private

attr_reader :copy_workflow_from

def update_form_options = { url: type_path(id: model.id), method: :patch, model:, copy_workflow_from: }
```

app/controllers/types_controller.rb
```diff
@@ -62,19 +62,17 @@ def edit
     end
   end
 
-  def create # rubocop:disable Metrics/AbcSize
+  def create
     CreateTypeService
       .new(current_user)
-      .call(permitted_type_params, copy_workflow_from: params[:copy_workflow_from]) do |call|
+      .call(permitted_type_params, copy_workflow_from: params.dig(:type, :copy_workflow_from)) do |call|
```
mereghost, line 68: 🟡 Am I missing something here (most probably yes) but since we are feeding the all the permitted type params, why do we need to the `copy_workflow_from` as a special ❄️?

app/controllers/types_controller.rb
```diff
@@ -62,19 +62,17 @@ def edit
...
       @type = call.result
 
       call.on_success do
         redirect_to_type_tab_path(@type, t(:notice_successful_create))
```
mereghost, line 72: 🟡 If you are dealing with Turbo Frames, it is recommended to redirect with a `:see_other` (303) status

mereghost (approved):
Left a couple of questions but nothing that should block the merge. :shipit:

app/views/types/new.html.erb
Kharonus: yes, definitely, good catch!

app/controllers/types_controller.rb
Kharonus: `copy_workflow_from` is no property from the type record.

previously (as you can see from the change) it was passed not even inside `params|type` but directly on top level params.

Usually, having something like this, I'd rework the whole form to use a form model, as this is apparently what the create service wants. But for sake of simplicity I tried to keep the previous approach alive. But with the primer forms I no longer was able to define the form input field outside the "model".

app/components/work_packages/types/settings_component.rb
Kharonus: Maybe there is something wrong a bit deeper. When starting to develop this functionality I started using strings for the type id. from params we also get strings, other records use string as their id. but for some reason, the `type.id` is an integer.

so, I have two options
- sanitizing this when first reading from params
- have the comparison function work to safely compare strings with integer ids

WDYT?

NobodysNightmare: I'd prefer to sanitize when reading from `params`. The latter is what a Rails method would probably try to do (e.g. `Type.find(...)`. But if I can choose, I try to write code that uses the correct types early.

NobodysNightmare (approved):

PR #17705, opened 2025-01-23
Draft: Allow to configure authentication method through UI

NobodysNightmare:
Done:
* Authentication method
* Nextcloud Audience
* Properly choose second step during create

TODO:
* Properly choose step after NC audience
* Automatically update next steps during select

# Ticket
https://community.openproject.org/projects/openproject/work_packages/60154/

## Screenshots

TODO

Some usages of Turbo Streams were replaced with updates to a parent Turbo Frame instead, which also allows to change larger parts of the UI with simpler code (e.g. we don't need `update.turbo_stream.erb` anymore).

TODO: What else? Introducing `current_steps`, making form state management more explicit...

# Merge checklist

- [ ] Added/updated tests


modules/storages/app/services/storages/one_drive_storage_wizard.rb
```diff
@@ -0,0 +1,65 @@
...
+    step :access_management,
+         completed_if: ->(storage) { !storage.automatic_management_unspecified? },
+         preparation: :prepare_storage_for_access_management_form
+
+    step :oauth_client,
+         completed_if: ->(storage) { storage.oauth_client.present? },
+         preparation: ->(storage) { storage.build_oauth_client }
+
+    step :redirect_uri,
+         completed_if: ->(storage) { storage.oauth_client && storage.oauth_client.created_at < 3.seconds.ago } # HACK HACK HACK
```
NobodysNightmare: This should obviously be done nicer... Not sure how.

modules/storages/app/controllers/storages/admin/oauth_clients_controller.rb
```diff
@@ -65,25 +65,12 @@ def create # rubocop:disable Metrics/AbcSize
     service_result.on_failure do
       update_via_turbo_stream(component: Storages::Admin::Forms::OAuthClientFormComponent.new(oauth_client: @oauth_client,
                                                                                               storage: @storage))
```
NobodysNightmare: Last(?) problem to be solved: This will not render the `turbo-frame: "page-content"` anymore. So after showing an error, we will not update the next step in the wizard.

PR #17708, opened 2025-01-23
Code maintenance/60714 ci broken with chromedriver for chrome 132

ulferts:
# Ticket

https://community.openproject.org/wp/60714

Rely on the current chrome again by switching to the new headless mode as:

>  Old Headless mode has been removed from the Chrome binary. Please use the new Headless mode (https://developer.chrome.com/docs/chromium/new-headless) or the chrome-headless-shell which is a standalone implementation of the old Headless mode (https://developer.chrome.com/blog/chrome-headless-shell).

Additionally, this fixes the flickering spec for `spec/features/work_packages/table/edit_work_packages_spec.rb`. The spec tried to clear an input field with the value 'Foobar'. But because the click to activate the field was happening in the middle of the string and then backspace was used, only the 'Foo' part was removed leaving 'bar'. Now the browser first goes to the end of the input before backspacing.

### DIFF
--- a/docker/ci/Dockerfile
+++ b/docker/ci/Dockerfile
@@ -11,8 +11,7 @@ ENV PGVERSION=13
 RUN wget --quiet -O- https://www.postgresql.org/media/keys/ACCC4CF8.asc | apt-key add - && \
   echo "deb http://apt.postgresql.org/pub/repos/apt bookworm-pgdg main" > /etc/apt/sources.list.d/pgdg.list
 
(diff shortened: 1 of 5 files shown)


ulferts:
The version is adapted by now.

oliverguenther (approved):
Excellent, thank you for finding the cause 👏

cbliard:
Ah, the old headless mode was explicitly used with selenium (used `--headless=old`). Good catch!

And with Cuprite/Ferrum, headless was specified without explicitly asking for new or old mode (used `--headless`), which is why it continued working.

That all makes sense now.

PR #17717, opened 2025-01-24
[#58637] replaced deletion confirmation with danger dialog

Kharonus:
# Ticket
[OP#58637](https://community.openproject.org/work_packages/58637)

- make deletion better guarded

- deleting a custom field hierarchy item now is guarded by a danger dialog

image


HDinger (approved):
Nice 👍 Thanks for taking care of that 🙇 

I think the failing test is unrelated, but please double check.

mereghost (approved):

PR #17721, opened 2025-01-24
Feature/60875 Amend work package comment href from `#activity-<journal-sequence>` to `#comment-<journal-id>`

akabiru:
# Ticket

https://community.openproject.org/wp/60875

Amend comment "copy link" id from `#activity-<journal-sequence>` to `#comment-<journal-id>`

 - [x] Old links with `#activity-<journal-sequence-number>` still work.
 - [x] Convert timestamps to anchors
    - [x] Add tests for mobile screens
    - [x] Add tests for filter states
 - [ ] ~Cover changesets~ _changesets do not include anchors_
 - [x] Copying the URL of a comment will have a new URL scheme #comment-<journal-id>
 - [x] When that URL is opened the browser will auto-scroll to have the comment in the visible viewport.

Feature flag: `OPENPROJECT_FEATURE_WORK_PACKAGE_COMMENT_ID_URL_ACTIVE`

## Screenshots

_New comment id anchor with backwards compatibility for activity anchors_

https://github.com/user-attachments/assets/4c086c0c-a21f-4f57-853c-39c2c73ab5d0

_Feature flag OFF_

https://github.com/user-attachments/assets/91a5ec04-81fa-4fc8-a43b-fc05cb8c382d

Index the activity elements with both `journal-{id,sequence-version}` via `data-anchor-{comment,activity}-id` so that both formats are supported

# Merge checklist

- [x] Added/updated tests


app/components/work_packages/activities_tab/journals/item_component.html.erb
```diff
@@ -1,14 +1,17 @@
 <%=
   component_wrapper(data: wrapper_data_attributes, class: "work-packages-activities-tab-journals-item-component") do
-    flex_layout(data: { 
+    flex_layout(data: {
       test_selector: "op-wp-journal-entry-#{journal.id}"
     }) do |journal_container|
       if show_comment_container?
         journal_container.with_row do
           render(border_box_container(
-                  id: "activity-anchor-#{journal.sequence_version}",
```
brunopagno, line 9: For a moment I thought we used native browser anchor behaviour, but I guess that has never been the case.

So I am assuming that removing the ID here has no consequence.

brunopagno (approved):
Looks pretty good. Nice job :tada: 

I tested locally and it worked as described.

app/components/work_packages/activities_tab/journals/item_component.html.erb
akabiru: Good eye, the navigation was done by hand as it was reported to cause ["native positioning issues"](https://github.com/opf/openproject/blob/dev/frontend/src/stimulus/controllers/dynamic/work-packages/activities-tab/index.controller.ts#L430) - in any case, since we need to support two formats, I moved the anchors to data attributes.

PR #17725, opened 2025-01-26
[60863] The arrows for lag filed in add-relation modal is visible in firefox

bsatarnejad:
# Ticket
https://community.openproject.org/wp/60863

- [x] Hide the stepper on Firefox
- [x] Avoid saving lag with a very big value

# Merge checklist

- [ ] Added/updated tests

### DIFF
--- a/app/models/relation.rb
+++ b/app/models/relation.rb
@@ -103,7 +103,7 @@ class Relation < ApplicationRecord
   scope :follows_with_lag,
         -> { follows.where("lag > 0") }
 
-  validates :lag, numericality: { allow_nil: true }
+  validates :lag, numericality: { allow_nil: true, less_than_or_equal_to: 2_147_483_647 }
 
   validates :to, presence: true, uniqueness: { scope: :from }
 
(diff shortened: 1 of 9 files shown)


app/models/relation.rb
```diff
@@ -103,7 +103,7 @@ class Relation < ApplicationRecord
   scope :follows_with_lag,
         -> { follows.where("lag > 0") }
 
-  validates :lag, numericality: { allow_nil: true }
+  validates :lag, numericality: { allow_nil: true, less_than_or_equal_to: 2_147_483_647 }
```
dombesz: I would also add `greater_than_or_equal_to: 0`, because if the input validation is bypassed, it is possible to save negative lag values.

dombesz (commented):
The hiding of the arrows works as expected, however when I submit the form with a large lag value, the error message is not displayed, but the form is not saved.

validation message not shown

dombesz (approved):
Thanks @bsatarnejad, I just corrected the error messages, otherwise LGTM!

PR #17738, opened 2025-01-27
Refactor the step-by-step storage setup

NobodysNightmare:
The idea is that now there is a central class (a Wizard) managing the steps necessary to guide the user through creation of a new storage.

This avoids problems such as the update action of step A being required to prepare some parts of step B already, so that the view/component for step B can be rendered correctly.

Since the same components are used inside and outside of the step-by-step wizard, we are explicitly transferring the state on whether we are currently in a wizard via parameters. This results in some degree of boilerplate, but is hopefully more clear than guessing based on context, whether or not we should show the next step after submission or go back to regular edit mode.

# Ticket
https://community.openproject.org/projects/cross-application-user-integration-stream/work_packages/60930

Reducing the coupling between steps in the storage setup, so that the exchange of steps is easy, without influencing other steps in the form.

The idea is that there is a wizard that is responsible for managing the necessary steps, while the state is fully stored on the model. This means the wizard should at any point be able to look at the model and decide which steps still need to be done and which steps have been completed already.

This means a controller does not need to know the next step in the process to prepare that already. However, a controller will still need to distinguish between the normal "edit" case (redirecting back to edit) and the wizard/step-by-step case, where it will redirect back to a wizard url (the URL itself does not encode state, except for the model id).

# Merge checklist

- [x] Added/updated tests


modules/storages/app/services/storages/wizard.rb
```diff
@@ -0,0 +1,103 @@
...
+#
+# You should have received a copy of the GNU General Public License
+# along with this program; if not, write to the Free Software
+# Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+#
+# See COPYRIGHT and LICENSE files for more details.
+#++
+
+module Storages
+  class Wizard
```
NobodysNightmare, line 32: This class was written pretty generically, not with only storages in mind. I could for example imagine to also use it for the OIDC provider setup, which from the UX point of view looks pretty similar.

Any opinions on moving this class to the core and calling it `Wizard` instead of `Storages::Wizard`?

modules/storages/app/services/storages/one_drive_storage_wizard.rb
```diff
@@ -0,0 +1,65 @@
...
+    step :access_management,
+         completed_if: ->(storage) { !storage.automatic_management_unspecified? },
+         preparation: :prepare_storage_for_access_management_form
+
+    step :oauth_client,
+         completed_if: ->(storage) { storage.oauth_client.present? },
+         preparation: ->(storage) { storage.build_oauth_client }
+
+    step :redirect_uri,
+         completed_if: ->(storage) { storage.oauth_client && storage.oauth_client.created_at < 3.seconds.ago } # HACK HACK HACK
```
NobodysNightmare: We are running into a limitation with the approach of expecting all state on the model.

The purpose of this last step in the UI is as follows: Previously you configured the `oauth_client` and now we want to show you the corresponding redirect URL, because you need to copy it into the OAuth 2.0 AS.

However, there is no change happening on the `storage` that we could observe to confirm that this step is done.

The current (hacky!!!) solution is to check whether the client was _recently_ created, if yes you need to see the URL, else not (if your browser takes too long for the redirect, bad luck).
Alternatively we could store something in the options of the storage to confirm that this piece of UX was shown, but then we'd have something UI-specific saved on the storage. Doing that also feels wrong.

Good ideas are welcome.

NobodysNightmare:
One more thought: I was considering to replace the `Storage#configuration_checks` with the `completed_if` conditions of the wizard as well. In places like the frontend UI this would probably work well and ensure that what we show (complete/incomplete) is in-sync with the wizard's state.

However, I decided against it (for now), because also checks like `Storage#completed?` depend on it and those are also used outside of the UI. This would effectively move the wizard away from the UI and we'd need to integrate it closer with the storage class. This is something I didn't want to do for now, but I am also interested in opinions on that.

modules/storages/app/services/storages/wizard.rb
mereghost: I can't wait to see the Fighter, Rogue and Cleric joining the party 🤣

modules/storages/app/services/storages/one_drive_storage_wizard.rb
mereghost: Can´t we pass down some flag around to represent a newly created OAuth Client?

NobodysNightmare: So the whole idea is that our code can safely come from a redirect to (e.g.) `.../new?continue_wizard=:id`. We explicitly wanted to achieve independence between steps, so redirecting to something like `.../new?continue_wizard=:id&did_i_just_create_a_client=yes` would defeat the purpose, because then the client creating step would need to know, that the next step is also interested in the client (which it only is for OneDrive, but not for Nextcloud).

Also, depending on information that's not stored on the `model` would mean, that the wizard state is not reproducible. E.g. after completing the wizard, I might see `pending_steps` again. This would practically work for our current use case, but looks like a design flaw to me. It would definitely prevent use cases such as sourcing the "complete"/"incomplete" labels from a wizard.

modules/storages/app/components/storages/admin/storage_view_component.html.erb
```diff
@@ -16,21 +16,25 @@
...
         end
       end
 
       component.with_row(scheme: :default) do
-        render(Storages::Admin::OAuthClientInfoComponent.new(oauth_client: storage.oauth_client, storage:))
+        if wizard_step == "oauth_client"
+          render(Storages::Admin::Forms::OAuthClientFormComponent.new(oauth_client: storage.oauth_client, storage:, in_wizard: true))
+        else
+          render(Storages::Admin::OAuthClientInfoComponent.new(oauth_client: storage.oauth_client, storage:))
+        end
```
mereghost: You probably tried, but let me ask anyway: have you tried to move all these conditionals to the component itself (the Object)?

NobodysNightmare: I added another commit that drives the refactoring further. While that keeps the conditional inside the `StorageViewComponent`, it reduces the number of conditionals for that purpose to one instead of many.

modules/storages/app/services/storages/wizard.rb
akabiru: > This class was written pretty generically, not with only storages in mind. I could for example imagine to also use it for the OIDC provider setup, which from the UX point of view looks pretty similar.
> 
> Any opinions on moving this class to the core and calling it `Wizard` instead of `Storages::Wizard`?

Really nice one Jan! 👏🏾  I would imagine this as a (view) component concern `app/components/concerns` ? The bare namespace `Wizard` might be too generic IMO- I suggest `MultiStepWizard`

NobodysNightmare: Two challenges to that.

Strongest one first: I don't think that this belongs into `concerns` anywhere, because as far as I can tell this directory is intended for modules that enhance/change the behaviour of classes including them, taking the name from [`ActiveSupport::Concern`](https://api.rubyonrails.org/classes/ActiveSupport/Concern.html). Since the wizard is not an includable module, but an inheritable base class, it doesn't belong into concerns.

I'd also go as far as not wanting to put it into `components`, because it doesn't implement a component itself. But I see that after the latest iteration it at least influences components strongly, so I might be convinced on that. Which other places would there be? `services` is my "catch-all" for classes that do things. I think we will not have enough to warrant the creation of a `wizards` directory.

Second challenge: In naming things, I see the desire to clarify by adding words, though this leads to long class names. Consistently `MultiStepWizard` would lead to `NextcloudStorageMultiStepWizard` etc. But then, the term wizard already [implies multiple steps](https://en.wikipedia.org/wiki/Wizard_(software)), I didn't make that term up, I just think it describes what we are doing here :)

modules/storages/app/services/storages/one_drive_storage_wizard.rb
NobodysNightmare: What do you think about persisting the user's acknowledgement of copying the redirect URL into the database?

No-no? Or acceptable?

modules/storages/app/common/storages/peripherals/one_drive_registry.rb
```diff
@@ -51,6 +51,22 @@ module Peripherals
         register(:set_permissions, StorageInteraction::OneDrive::SetPermissionsCommand)
       end
 
+      namespace("components") do
+        namespace("forms") do
+          register(:access_management, ::Storages::Admin::Forms::AccessManagementFormComponent)
+          register(:general_information, ::Storages::Admin::Forms::GeneralInfoFormComponent)
```
Kharonus, line 57: ❓ Is the follow up here to have two different form components? Your first step makes sense, so that we then can have a form component for OD and one for NC.

modules/storages/app/common/storages/peripherals/one_drive_storage_wizard.rb
```diff
@@ -0,0 +1,75 @@
...
+           completed_if: ->(storage) { storage.oauth_client.present? },
+           preparation: ->(storage) { storage.build_oauth_client }
+
+      step :redirect_uri,
+           section: :oauth_configuration,
+           completed_if: ->(storage) {
+             # Working around the fact that there is nothing changed on the storage after showing
+             # the redirect url. The redirect URL step only exists to show the oauth client's redirect
+             # URL to the user right after the client was created.
+             storage.oauth_client&.created_at && storage.oauth_client.created_at < 10.seconds.ago
```
Kharonus: 🟡 I don't understand yet, what this workaround actually does. Something like, if I refresh the page this step is "completed" and hence has this fancy label and is collapsed?

Kharonus: What is opposing the possibility of simply having the same completed check as the step before `completed_if: ->(storage) { storage.oauth_client.present? }`?

Kharonus (approved):
I have two things open, but in general this looks pretty mergeable.

modules/storages/app/common/storages/peripherals/nextcloud_registry.rb
```diff
@@ -57,6 +57,22 @@ module Peripherals
...
+          register(:oauth_client, ::Storages::Admin::Forms::OAuthClientFormComponent)
+        end
+
+        register(:setup_wizard, NextcloudStorageWizard)
+
+        register(:automatically_managed_folders, ::Storages::Admin::AutomaticallyManagedProjectFoldersInfoComponent)
+        register(:general_information, ::Storages::Admin::GeneralInfoComponent)
+        register(:oauth_application, ::Storages::Admin::OAuthApplicationInfoComponent)
+        register(:oauth_client, ::Storages::Admin::OAuthClientInfoComponent)
+      end
```
mereghost, line 74: <3

modules/storages/app/common/storages/peripherals/one_drive_registry.rb
mereghost: Well... it at least allows to.

modules/storages/app/common/storages/peripherals/one_drive_storage_wizard.rb
mereghost: 🟢 WDYT of making the test: `storage.oauth_client.persisted? && storage.oauth_client.created_at < 10.seconds.ago`

mereghost: `present` won't mean `persisted` which is what he is checking for.

modules/storages/app/components/storages/admin/forms/oauth_client_form_component.rb
```diff
@@ -33,16 +33,17 @@ class OAuthClientFormComponent < ApplicationComponent
...
-    def initialize(oauth_client:, storage:, **)
-      super(oauth_client, **)
-      @storage = storage
-    end
+    options in_wizard: false
 
     def self.wrapper_key = :storage_oauth_client_section
 
+    def form_url
+      query = { continue_wizard: storage.id } if in_wizard
```
mereghost, line 36: 💚 It as been a while since I saw this. That's a very ruby-esque line, making query nil in case in_wizard is false. =)

modules/storages/app/components/storages/admin/oauth_application_info_component.rb
```diff
@@ -34,13 +34,8 @@ class OAuthApplicationInfoComponent < ApplicationComponent
...
 
-    attr_reader :storage
-    alias_method :oauth_application, :model
-
-    def initialize(oauth_application:, storage:, **)
-      super(oauth_application, **)
-      @storage = storage
-    end
+    alias_method :storage, :model
+    delegate :oauth_application, to: :storage
```
mereghost: 🟢 I noticed a lot of components include the same modules and have the same aliases, delegates set. Would it be worth create a base `StorageConfigComponent` to inherit from?

modules/storages/app/services/storages/wizard.rb
```diff
@@ -0,0 +1,104 @@
...
+    def initialize(model:, user: nil)
+      @model = model
+      @user = user
+    end
+
+    def prepare_next_step
+      next_step = pending_steps.first
+      return nil if next_step.nil?
+
+      self.class.steps_by_name.fetch(next_step).prepare_on(self)
```
mereghost: 🟢 Since you access `self.class.steps_by_name` in a couple of places, would be worth to introduce a instance level access (not sure at which level of visibility)?

mereghost (approved):
Great work! It will help us a lot by AoE'ing bugs all around. 🧙🏼 

I left a couple of questions/suggestions, but feel free to ignore them as none should warrant blocking the merge.

modules/storages/app/common/storages/peripherals/one_drive_storage_wizard.rb
Kharonus: Okay, then using `persisted?`, would be fine, too. I'm just challenging the usage of `10s` here.

modules/storages/app/common/storages/peripherals/one_drive_registry.rb
Kharonus: I think, it should be the next step. Getting rid of any if statements in the forms

modules/storages/app/common/storages/peripherals/one_drive_storage_wizard.rb
mereghost: I figured out that the 10s was a random number guaranteed by a d20 roll. :P

modules/storages/app/common/storages/peripherals/one_drive_registry.rb
NobodysNightmare: I didn't look into the forms too much, but I agree. If there are differences in those, this allows to split them into different forms.

Though I also agree that "next step" is the right point in time here, because this PR is already too huge xD

modules/storages/app/common/storages/peripherals/one_drive_storage_wizard.rb
NobodysNightmare: Marcello kinda nails it.

The problem we need to solve here is that the step needs to be visible _at least once_. If I used the same completed condition as the step before, then this step would never be visible.

The way it's currently written, the step is visible for 10 seconds after the previous step finished, so even on a slow connection you would pretty surely be guaranteed to see it. The second half of the workaround is that there is no next step afterwards and the form goes out of "step-by-step mode" anyways, so there is not really a "too large" value here, except if we wanted to reuse the wizard in more places (such as to determine the "completed" / "incomplete" labels).

As I wrote in one of my comments: This goes down to the fact that the basic assumption of the wizard is, that the correct current step can be read from the state of the model, which is not really possible here. One other way out would be to explicitly store the information that the user has seen the redirect URL. Though then we'd store UI-specific info on the storage model, which is also very debatable.

NobodysNightmare: I think my first test was

```ruby
storage.oauth_client.present? && storage.oauth_client.created_at < 10.seconds.ago
```

This failed when there were validation errors for the oauth_client IIRC, because then we'd have a client present with a nil creation date.

If I add one ampersand to your proposal, it might work as well as the current code and is probably more readable than it:

```ruby
storage.oauth_client&.persisted? && storage.oauth_client.created_at < 10.seconds.ago
```

modules/storages/app/components/storages/admin/oauth_application_info_component.rb
NobodysNightmare: Actually a good question. I purposefully made sure that all components respond to the same initializer signature (since I am creating them through a registry), I guess one of the side effects was that they all now look very similar as well. I am going to look into this, but I might prefer to defer this to "later".

NobodysNightmare: I added a new commit that introduces a superclass for info components and another one for the form components which collect the low-hanging fruits at least.

modules/storages/app/services/storages/wizard.rb
NobodysNightmare: done

modules/storages/app/components/storages/admin/oauth_application_info_component.rb
mereghost: Nice. And later is more than fine.

PR #17739, opened 2025-01-27
improve performance of involving_user scope which improves date alert…

ulferts:
# Ticket

https://community.openproject.org/wp/60932

The performance of the `CreateDateAlertsNotificationsJob` job on instances with a larger set of work packages and users is suboptimal. This can be traced down to the way the `WorkPackage.involving_user` scope is performing which is included in the SQL for fetching all work packages for which a user should get notified.

The SQL before the start is like this:

```SQL
SELECT
	"work_packages".*
FROM
	"work_packages"
	LEFT OUTER JOIN "watchers" ON "watchers"."watchable_type" = 'WorkPackage'
	AND "watchers"."watchable_id" = "work_packages"."id"
WHERE
	(
		"watchers"."user_id" = 31032
		OR "work_packages"."assigned_to_id" = 31032
		OR "work_packages"."assigned_to_id" IN (
			SELECT
				"group_users"."group_id"
			FROM
				"group_users"
			WHERE
				"group_users"."user_id" = 31032
		)
		OR "work_packages"."responsible_id" = 31032
		OR "work_packages"."responsible_id" IN (
			SELECT
				"group_users"."group_id"
			FROM
				"group_users"
			WHERE
				"group_users"."user_id" = 31032
		)
	)
```

This leads to the following execution plan:

image

The statement does a full scan on work packages and it does so three times. The reason for this is are the OR statements on three different attributes where indices are not applied

By changing the statement to use UNIONS the OR can be avoided. Additionally using a CTE to only do the repetitive filtering for users and groups once, the statement now looks like this:

```SQL
WITH
	"user_and_groups" AS (
		SELECT
			"group_users"."group_id"
		FROM
			"group_users"
		WHERE
			(USER_ID = 31032)
		UNION
		SELECT
			31032
	)
SELECT
	"work_packages".*
FROM
	"work_packages"
WHERE
	"work_packages"."id" IN (
		(
			(
				SELECT
					"work_packages"."id"
				FROM
					"work_packages"
					LEFT OUTER JOIN "watchers" ON "watchers"."watchable_type" = 'WorkPackage'
					AND "watchers"."watchable_id" = "work_packages"."id"
				WHERE
					"watchers"."user_id" = 31032
			)
			UNION
			(
				SELECT
					"work_packages"."id"
				FROM
					"work_packages"
				WHERE
					(
						ASSIGNED_TO_ID IN (
							SELECT
								*
							FROM
								USER_AND_GROUPS
						)
					)
			)
			UNION
			(
				SELECT
					"work_packages"."id"
				FROM
					"work_packages"
				WHERE
					(
						RESPONSIBLE_ID IN (
							SELECT
								*
							FROM
								USER_AND_GROUPS
						)
					)
			)
		)
	)
```

which admittedly is more complicated. 

So is the execution plan:

image

But that plan also shows that index scans are used now.

This approach might not be best in scenarios where the majority of the work packages are assigned to/watched by/responsible for by the same user. But that is a rather unlikely scenario.

The performance measurements vary especially for the former statement. Measurements have shown execution times between 400ms and 1.7s. But the result for the reworked statement seems to constantly be below 10ms.

### DIFF
--- a/app/models/work_packages/scopes/involving_user.rb
+++ b/app/models/work_packages/scopes/involving_user.rb
@@ -36,19 +36,45 @@ module WorkPackages::Scopes
       # for, or watcher, via a group or themself
       #
       # @param user User the user involved in work packages.
-      def involving_user(user)
-        WorkPackage.left_joins(:watchers)
-          .where(watchers: { user: })
-          .or(WorkPackage.where(assigned_to: user))
-          .or(WorkPackage.where(assigned_to: group_having(user)))
-          .or(WorkPackage.where(responsible: user))
-          .or(WorkPackage.where(responsible: group_having(user)))
+      def involving_user(user) # rubocop:disable Metrics/AbcSize
+        # Using CTEs and UNION for performance reasons
+        # The alternative would be the simpler:
+        # ```
+        # WorkPackage
+        #   .left_joins(:watchers)
+        #   .where(watchers: { user: })
+        #   .or(WorkPackage.where(assigned_to: user))
+        #   .or(WorkPackage.where(assigned_to: group_having(user)))
+        #   .or(WorkPackage.where(responsible: user))
+        #   .or(WorkPackage.where(responsible: group_having(user)))
+        # ```
+        # That however prevents indices from being used because of the or clauses on different columns.
+        watched = WorkPackage
+                    .left_joins(:watchers)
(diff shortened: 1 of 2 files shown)


cbliard (approved):
Very nice!

PR #17740, opened 2025-01-27
Use Caddy for local TLS with docker instead of Traefik

brunopagno:
This is just exploratory work for now, feel free to ignore.

Trying to simplify TLS setup

Caddy has some out of the box functionality to run HTTPs on localhost environment

https://caddyserver.com/docs/automatic-https#local-https

With this we should be able to reduce the setup a bit.


brunopagno:
Yeah, this is going to take some work, probably not worth the effort now

PR #17763, opened 2025-01-29
Make Keycloak remember things

NobodysNightmare:
The default configuration for keycloak didn't attach any kind of volume to its database, so after

    docker compose down

everything was gone. I learned this today.


Kharonus (approved):
LGTM

PR #17778, opened 2025-01-30
add start time & end time to time entry to api docs

klaustopher:
# Ticket
https://community.openproject.org/projects/stream-time-and-costs/work_packages/60633

Also add api docs for the start and end times

## Screenshots

# Merge checklist

- [ ] Added/updated tests

### DIFF
--- a/docs/api/apiv3/components/schemas/time_entry_model.yml
+++ b/docs/api/apiv3/components/schemas/time_entry_model.yml
@@ -8,7 +8,7 @@ properties:
     minimum: 1
   comment:
     allOf:
-      - $ref: './formattable.yml'
+      - $ref: "./formattable.yml"
       - description: A comment to the time entry
   spentOn:
     type: string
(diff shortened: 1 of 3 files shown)


docs/api/apiv3/components/schemas/time_entry_model.yml
```diff
@@ -25,6 +25,24 @@ properties:
     type: string
     format: date-time
     description: The time the time entry was created
+  startTime:
+    type: string
```
Kharonus: 🟡 if this thing can be `null` it is encouraged to use something like:

```yml
    type:
      - 'string'
      - 'null'
```

But, will it actually render the `null` in the JSON? Or will it be skipped in the render (the representer has `allow_nil: true` in the getter, what is the effect of this?

klaustopher: It will not render the field, when the setting is disabled. And if it is not enforced to track times, then it can be null

Kharonus (approved):
LGTM

PR #17779, opened 2025-01-30
Do not implicitly handle errors on notification center

brunopagno:
# Ticket

https://community.openproject.org/wp/59328

There's an error notification that is confusing to users and useless to developers. The idea is to hide it.

Not expose backend errors on the notifications page.


brunopagno:
A note for the future, but disabling this alert also disables this error message when filtering by date alerts (which is an enterprise feature). As far as I know the user would need to find a link to the page, when trying to navigate throught the menu they would see the upsale page

image

oliverguenther (approved):

PR #17785, opened 2025-01-31
Bug/61081 time entry creation via the api is not correctly handling user timezones

klaustopher:
# Ticket
https://community.openproject.org/projects/stream-time-and-costs/work_packages/61081

Fixes handling of the `startTime` timestamp to work with the correct timezone. Also adding tests for that behavior

## Screenshots

# Merge checklist

- [x] Added/updated tests

### DIFF
--- a/modules/costs/lib/api/v3/time_entries/time_entry_representer.rb
+++ b/modules/costs/lib/api/v3/time_entries/time_entry_representer.rb
@@ -88,16 +88,6 @@ module API
                    datetime_formatter.format_duration_from_hours(represented.hours) if represented.hours
                  end
 
-        date_time_property :start_time,
-                           exec_context: :decorator,
-                           getter: ->(*) { datetime_formatter.format_datetime(represented.start_timestamp, allow_nil: true) },
-                           if: ->(*) { TimeEntry.can_track_start_and_end_time? }
-
-        date_time_property :end_time,
-                           exec_context: :decorator,
-                           getter: ->(*) { datetime_formatter.format_datetime(represented.end_timestamp, allow_nil: true) },
-                           if: ->(*) { TimeEntry.can_track_start_and_end_time? }
-
         date_time_property :created_at
         date_time_property :updated_at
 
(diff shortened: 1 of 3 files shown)


modules/costs/lib/api/v3/time_entries/time_entry_representer.rb
```diff
@@ -108,6 +98,20 @@ class TimeEntryRepresenter < ::API::Decorators::Single
 
         associated_resource :user
 
+        date_time_property :start_time,
```
Kharonus, line 101: 🟢 uh, so the order has an influence in the parsing part of the representer? `start_time` needs to be defined after `user` to have access?

modules/costs/spec/requests/api/time_entries/create_resource_spec.rb
```diff
@@ -0,0 +1,124 @@
...
+          post path, parameters.to_json
+
+          expect(subject).to have_http_status(:created)
+          time_entry_id = json_response["id"]
+          time_entry = TimeEntry.find(time_entry_id)
+
+          expect(time_entry.spent_on).to eq(Date.new(2024, 12, 24))
+          expect(time_entry.hours).to eq(2)
+          expect(time_entry.start_time).to eq(13 * 60) # 12:00 UTC = 13:00 Berlin on December 24th, 2024
+          expect(time_entry.time_zone).to eq("Europe/Berlin")
```
Kharonus, line 88: ❓ maybe stupid question, but how does the lib differentiate between DST and standard time for `Europe/Berlin`? Is there another identifier for DST? Or might this test break in April?

Kharonus (approved):
Seems legit in every aspect. I have a question in there, which probably doesn't block the merge.

modules/costs/lib/api/v3/time_entries/time_entry_representer.rb
klaustopher: I understand it to be parsed sequentially, which makes sense

modules/costs/spec/requests/api/time_entries/create_resource_spec.rb
klaustopher: Europe/Berlin is a fixed identifier for a timezone that has rules attached to it with information when DST starts and when the offset to UTC changes. It then applies the correct offset when you pass it the date and time:

```
[1] pry(main)> tz = ActiveSupport::TimeZone['Europe/Berlin']
=> #<ActiveSupport::TimeZone:0x00000001628dbc70 @name="Europe/Berlin", @tzinfo=#<TZInfo::DataTimezone: Europe/Berlin>, @utc_offset=nil>

[2] pry(main)> tz.local(2024, 12, 24, 12, 0, 0)
=> Tue, 24 Dec 2024 12:00:00.000000000 CET +01:00

[3] pry(main)> tz.local(2024, 6, 18, 12, 0, 0)
=> Tue, 18 Jun 2024 12:00:00.000000000 CEST +02:00
```

As we use a fixed date here in the test (24.12.2024), we will always get a timestamp using Central European Time, never Central European Summer Time.

klaustopher: That's also why it is always a good idea to use the local identifiers like Europe/Berlin, Asia/Seoul or America/Mexico_City instead of things like `CET`, `KST` or `CST`

PR #17791, opened 2025-01-31
bug/60976 Setting the user display format without lastname breaks user custom fields with Group values

akabiru:
# Ticket

https://community.openproject.org/wp/60976

Fix an internal server error that occurs when using custom fields of type 'user' with user format settings that exclude 'lastname'. The error manifests when Groups are included in the possible values and the user format setting doesn't include 'lastname' (e.g., when set to 'username' only).

The root cause was that Groups rely on `lastname` (aliased as `name`) for comparison operations, but this column wasn't always being selected when loading possible values for custom fields. This would happen specifically when the user format setting didn't include lastname (e.g., `:username` or `:firstname` formats).

The solution ensures we always include the `lastname` column when selecting possible values for user custom fields, regardless of the user format setting. This maintains Group functionality while respecting user format preferences.

Follows #17350

# Merge checklist

- [x] Added/updated tests

### DIFF
--- a/app/models/custom_field.rb
+++ b/app/models/custom_field.rb
@@ -333,8 +333,11 @@ class CustomField < ApplicationRecord
                   .in_visible_project_or_me(User.current)
               end
 
-      scope
-        .select(User::USER_FORMATS_STRUCTURE[Setting.user_format].map(&:to_s), "id", "type")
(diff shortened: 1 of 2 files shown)


oliverguenther (approved):
Good catch, thanks for looking into this! 👍

PR #17793, opened 2025-01-31
Do not allow time tracking for users that are not visible to you

klaustopher:
# Ticket
https://community.openproject.org/projects/stream-time-and-costs/work_packages/61084

Prevent logging time for users that cannot be accessed by the user. This is currently only prevented by the auto completer, but not by the backend.

## Screenshots

# Merge checklist

- [ ] Added/updated tests

### DIFF
--- a/modules/costs/app/contracts/time_entries/base_contract.rb
+++ b/modules/costs/app/contracts/time_entries/base_contract.rb
@@ -46,6 +46,7 @@ module TimeEntries
     validate :validate_hours_are_in_range
     validate :validate_project_is_set
     validate :validate_work_package
+    validate :validate_user
 
     validates :spent_on,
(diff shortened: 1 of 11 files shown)


modules/costs/app/contracts/time_entries/base_contract.rb
```diff
@@ -117,8 +127,8 @@ def work_package_not_in_project?
       model.work_package && model.project != model.work_package.project
     end
 
-    def validate_logged_by_current_user
-      errors.add :logged_by_id, :not_current_user if model.logged_by != logged_by
```
klaustopher, line 121: this validation was never run, and `logged_by` is not even defined, so using it will throw a NoMethodError

oliverguenther (approved):
Thanks for the fix 👍 Some API tests are still failing

PR #17794, opened 2025-01-31
Enable erb-linting on lefthook and the CI

dombesz:
# Ticket
N/A

Enabling the erb-linter for the CI to avoid common mistakes in the erb files.

The enabling unfortunately has to be done on the file level because `erb_lint` cannot lint only the modifications from a file.

# Merge checklist

- [ ] Added/updated tests


.github/workflows/rubocop-core.yml
```diff
@@ -26,3 +29,11 @@ jobs:
             rubocop-rspec_rails:gemfile
           reporter: github-pr-check
           only_changed: true
+      - name: Run bundle install
```
dombesz, line 32: For some reason  the `erblint` command tries to run the ruby environment in a way that it requires bundle install, otherwise it fails. https://github.com/opf/openproject/actions/runs/13077602831/job/36494149828

The reason is, the `use_bundler:` env variable is not forwarded to the script, meaning it will always try to run the `erblint` command with `bundle exec`. This raise an error, as `bundle install` wasn't ran previously.

The fix is to run `bundle install`. This will slow down the action a bit, but it's not considerable.

klaustopher (approved):

PR #17795, opened 2025-01-31
fix unexpected passing of US-ASCII string into Commonmarker on search

ulferts:
# Ticket

https://community.openproject.org/wp/61110

Avoid passing a US-ASCII string into Commonmarker which leads to an error therein. 

By compacting, nil values are returned from the set of fetched attachment fulltexts. Otherwise a 

`[nil, "Some text"].join` 

leads to a US-ASCII encoded string.

The same problem does not happen for filenames as those have a default value of `''` defined on the table and shouldn't be empty to begin with.

# Merge checklist

- [x] Added/updated tests

### DIFF
--- a/app/helpers/search_helper.rb
+++ b/app/helpers/search_helper.rb
@@ -124,13 +124,19 @@ module SearchHelper
 
   def attachment_fulltexts(event)
     only_if_tsv_supported(event) do
-      Attachment.where(id: event.attachment_ids).pluck(:fulltext).join(" ")
+      # The compact is important here as the fulltext can be nil.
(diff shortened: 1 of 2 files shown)


toy:
This fixes particular case, but maybe better to ensure that `Commonmarker.to_html` in `OpenProject::TextFormatting::Filters::MarkdownFilter` gets a UTF-8 string?

Linking created issues: https://bugs.ruby-lang.org/issues/21102 and https://github.com/gjtorikian/commonmarker/issues/351

klaustopher (approved):

PR #17801, opened 2025-02-03
[60615] The right-side pane scrolls with the rest of the page after adding a child via the Relations tab

bsatarnejad:
# Ticket
https://community.openproject.org/wp/60615
https://community.openproject.org/wp/60629

After creating a new child and setting as a new relation for parent, in relations tab, we should scroll into it.

## Screenshots

keep-open-on-submit set to false (its default value). If it is true, it keeps the dialog element but hide it. If the Child is created in relations tab, then we should scroll into this child.

This solution fixes the issue in Gantt view too.

# Merge checklist

- [ ] Added/updated tests

### DIFF
--- a/app/components/work_package_relations_tab/index_component.rb
+++ b/app/components/work_package_relations_tab/index_component.rb
@@ -107,7 +107,7 @@ class WorkPackageRelationsTab::IndexComponent < ApplicationComponent
     if helpers.current_user.allowed_in_project?(:add_work_packages, work_package.project)
       menu.with_item(
         label: t("work_package_relations_tab.relations.new_child"),
-        href: new_project_work_packages_dialog_path(work_package.project, parent_id: work_package.id),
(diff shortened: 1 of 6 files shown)


HDinger (changes requested):
This correctly fixes the white space bug that magically appeared after creating the new WorkPackage. However, the scrolling issue described in the ticket still remains.

HDinger (changes requested):
I don't think that the generic `work_packages/dialogs_controller` should know or care about where it is called. Otherwise, we'd need to introduce a param for every place this is called in order to replace the correct component after a successful call. 
So please make the more generic and handle relations specific code in the respective controller.

bsatarnejad:
I'm closing this PR because this bug will be fixed in [WP for auto scrolling to the new created child](https://community.openproject.org/projects/stream-planning-and-reporting/work_packages/61143/activity)

PR #17803, opened 2025-02-03
Use relative path instead of url for activities refresh

cbliard:
When using accessing the instance using the wrong host name (different from `Setting.host_name` configured in the administration or through env vars), the app should still work.

The activities tab component is using a url to know where the activities should be fetched from. This will trigger a CORS error when using the wrong host name, and the activities will not be updated. By using a relative path, the request will be relative to the current host, and thus work correctly even when using the wrong host name.

I detected that during local development while using port 4200 instead of 3000 as we usually do.


dombesz (approved):

akabiru:
Great find @cbliard and thanks for fixing it! 🙇🏾

PR #17807, opened 2025-02-03
Bug/61134 no error message when user tries to save time log without work packages

klaustopher:
# Ticket
https://community.openproject.org/work_packages/61134

We have 3 cases how we can log time
- from the work package
- from the spent time widget in a project
- from the my spent time widget on the my page

> [!NOTE]  
> Currently on the `TimeEntry` only the project is a mandatory field, it is not required to be linked with a work package. We might want to change this in the future. See [WP #61136](https://community.openproject.org/projects/stream-time-and-costs/work_packages/61136/)

We do not offer selecting the project itself on the modal (not the old one and not the new one), so in the case when we log time entries from the my page, we do not know the project and thus the WP becomes mandatory in the dialog because we need it to set the project through the WP. But since the validation is run on the `project_id` field, we do not show it anywhere in the dialog. So we need to display the error from the `project_id` field on the `work_package_id` autocompleter.

## Screenshots
Bildschirmfoto 2025-02-03 um 15 28 38

# Merge checklist

- [ ] Added/updated tests

### DIFF
--- a/modules/costs/app/components/time_entries/work_package_form.rb
+++ b/modules/costs/app/components/time_entries/work_package_form.rb
@@ -41,7 +41,8 @@ module TimeEntries
       if show_work_package_field?
         f.work_package_autocompleter name: :work_package_id,
                                      label: TimeEntry.human_attribute_name(:work_package),
-                                     required: true,
+                                     required: work_package_required?,
+                                     validation_message: work_package_validation_error,
                                      autocomplete_options: {
                                        defaultData: false,
                                        component: "opce-time-entries-work-package-autocompleter",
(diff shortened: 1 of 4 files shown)


modules/costs/app/components/time_entries/work_package_form.rb
```diff
@@ -41,7 +41,8 @@ def initialize(visible: true)
       if show_work_package_field?
         f.work_package_autocompleter name: :work_package_id,
                                      label: TimeEntry.human_attribute_name(:work_package),
-                                     required: true,
+                                     required: work_package_required?,
```
dombesz, line 44: I believe `required: true` is sufficient here, because logging time directly on a project is not possible, so there is no such case when the work package field is shown and not required.

dombesz (approved):
I left one comment, but it works as expected.👍
On a side note, the form has too much space when the work package field is not shown.

image

modules/costs/app/components/time_entries/work_package_form.rb
klaustopher: True. But I still believe we gonna switch to have the WP mandatory, so we won't really need it in the future, that's why the comment above the 2 methods

PR #17809, opened 2025-02-03
[#60923] Display sums in Timesheet PDF

as-op:
# Ticket

https://community.openproject.org/work_packages/60923

* Add an overview sums table for the timesheet PDF export (if export contains more than one user)
* Add a sum row below each user table

# Merge checklist

- [x] Added/updated tests


klaustopher (approved):
looked at it yesterday. very good. Approving since i believe all tests will be green ;)

PR #17810, opened 2025-02-03
Bump Keycloak to latest version

NobodysNightmare:
The version we've been using was outdated by 5 major releases.

A quick smoke check looks like everything is still working as intended with the new version and it helps in creation of bug reports when you use the latest version of a software ;-)


ba1ash (approved):
LGTM

PR #17811, opened 2025-02-03
Skip broken recurring meetings spec

mrmir:
Skipping the broken spec

WP to fix it - https://community.openproject.org/wp/61144


myabc (changes requested):
I think we can close now that #17569 (with @oliverguenther's [2a38e89](https://github.com/opf/openproject/pull/17569/commits/2a38e890d5de2fb7b3c05ca4931dac9cf83531df)) is merged

PR #17826, opened 2025-02-04
Fix/61049 activity tab scroll to bottom

brunopagno:
# Ticket
https://community.openproject.org/projects/communicator-stream/work_packages/61049
https://community.openproject.org/projects/communicator-stream/work_packages/61074

Make "scrolling to the end" of the activity tab when there are more than 30 comments more consistent.

The reason the scrolling ends up on the wrong place seems to be that we do the turbo frame DOM update simultaneously with the scrolling behaviour, thus putting the user in the wrong position.

## Screenshots

https://github.com/user-attachments/assets/8c325774-bb3c-42bc-aeb5-b31e8c9ef51b

Tried to rely on [Turbo Frame's autoscroll properties](https://turbo.hotwired.dev/reference/frames#html-attributes) but we have too much dependency on javascript custom behaviour, checking the viewport and handling custom events, that in the end just "forcing" auto scroll seems to be a way simpler approach.

# Merge checklist

### DIFF
--- a/frontend/src/stimulus/controllers/dynamic/work-packages/activities-tab/index.controller.ts
+++ b/frontend/src/stimulus/controllers/dynamic/work-packages/activities-tab/index.controller.ts
@@ -393,7 +393,7 @@ export default class IndexController extends Controller {
     this.tryScroll(activityAnchorName, activityId, 0, maxAttempts);
   }
 
-  private tryScrollToBottom(attempts:number = 0, maxAttempts:number = 20) {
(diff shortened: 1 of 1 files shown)


akabiru (approved):
:shipit:

akabiru:
> Caution
> 
> The provided work package version does not match the core version
> 
> Details:
> 
>     * Work package URL: https://community.openproject.org/projects/communicator-stream/work_packages/61049
> 
>     * Work package version: not set
> 
>     * Core version: 15.3.0
> 
> 
> Please make sure that:
> 
>     * The work package version OR your pull request target branch is correct

I've set the version on the WP to 15.3 👍🏾

PR #17828, opened 2025-02-05
Fix open redirect vulnerability by enforcing internal URL validation for back_url

anisjellazi:
**Ticket:**  

---

**What are you trying to accomplish?**  
This PR fixes an open redirect vulnerability caused by improper validation of user-supplied `back_url` parameters. In our current implementation (in files like `account_controller.rb` and `application_controller.rb`), attackers could manipulate the `back_url` parameter to redirect users to external malicious sites. This PR ensures that only internal, relative URLs are used for redirection.

---

**What approach did you choose and why?**  

- **Internal URL Validation:**  
  We added a helper method `valid_internal_url` which parses the provided URL and verifies:
  - The URL does not include a scheme (e.g., `http://` or `https://`) or host.
  - The URL path starts with a single `/` and does not begin with `//` (to avoid protocol-relative URLs).
  - If the URL fails these checks, a safe fallback is returned (e.g., `home_url` or `nil`).

- **Usage Update:**  
  We updated all redirection calls that use `params[:back_url]` or HTTP_REFERER to pass through `valid_internal_url`. This change was made in both `account_controller.rb` and `application_controller.rb` to ensure consistency and prevent external redirects.

**Screenshots:**  

---

**Merge checklist:**

- [x] Added/updated tests  

### DIFF
--- a/app/controllers/account_controller.rb
+++ b/app/controllers/account_controller.rb
@@ -77,8 +77,6 @@ class AccountController < ApplicationController
 
   # Log out current user and redirect to welcome page
   def logout
-    # Keep attributes from the session
-    # to identify the user
     previous_session = session.to_h.with_indifferent_access
     previous_user = current_user
 
@@ -97,7 +95,7 @@ class AccountController < ApplicationController
 
       @user = @token.user
       if request.post?
-        call = ::Users::ChangePasswordService.new(current_user: @user, session:).call(params)
+        call = ::Users::ChangePasswordService.new(current_user: @user, session: session).call(params)
         call.apply_flash_message!(flash) if call.errors.empty?
 
         if call.success?
@@ -112,50 +110,27 @@ class AccountController < ApplicationController
       mail = params[:mail]
       user = User.find_by_mail(mail) if mail.present?
 
-      # Ensure the same request is sent regardless of which email is entered
-      # to avoid detecability of mails
       flash[:notice] = I18n.t(:notice_account_lost_email_sent)
 
       unless user
-        # user not found in db
         Rails.logger.error "Lost password unknown email input: #{mail}"
         return
       end
 
       unless user.change_password_allowed?
-        # user uses an external authentication
         UserMailer.password_change_not_possible(user).deliver_later
         Rails.logger.warn "Password cannot be changed for user: #{mail}"
         return
       end
 
-      # create a new token for password recovery
       token = Token::Recovery.new(user_id: user.id)
       if token.save
         UserMailer.password_lost(token).deliver_later
         flash[:notice] = I18n.t(:notice_account_lost_email_sent)
-        redirect_to action: "login", back_url: home_url
+        # Use the safe URL here:
+        redirect_to action: "login", back_url: valid_internal_url(home_url)
         nil
       end
-    end
-  end
-
-  # User self-registration
-  def register
-    return self_registration_disabled unless allow_registration?
-
-    @user = invited_user
-
-    if request.get?
-      registration_through_invitation!
-    else
-      if Setting.email_login?
-        params[:user][:login] = params[:user][:mail]
-      end
-
-      self_registration!
-
-      call_hook :user_registered, { user: @user } if @user.persisted?
     end
   end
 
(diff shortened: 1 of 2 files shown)


app/controllers/account_controller.rb
```diff
@@ -112,53 +110,30 @@ def lost_password
...
       end
 
-      # create a new token for password recovery
       token = Token::Recovery.new(user_id: user.id)
       if token.save
         UserMailer.password_lost(token).deliver_later
         flash[:notice] = I18n.t(:notice_account_lost_email_sent)
-        redirect_to action: "login", back_url: home_url
+        # Use the safe URL here:
+        redirect_to action: "login", back_url: valid_internal_url(home_url)
```
klaustopher, line 131: `home_url` is not based on any user input, why do we need to put it through the helper?

app/controllers/account_controller.rb
```diff
@@ -77,8 +77,6 @@ def internal_login
 
   # Log out current user and redirect to welcome page
   def logout
-    # Keep attributes from the session
-    # to identify the user
```
klaustopher, line 81: Is there a good reason to get rid of the comments in this PR?

app/controllers/account_controller.rb
```diff
@@ -284,179 +214,52 @@ def activate_through_ldap(user)
...
-    appends = flash[:_csp_appends]
-    return unless appends
-
-    append_content_security_policy_directives(appends)
-  end
-
-  def check_internal_login_enabled
-    render_404 unless omniauth_direct_login?
-  end
+  # ... other actions and methods remain unchanged ...
```
klaustopher, line 238: You removed many controller actions like the `register` and `change_password` actions and many helpers. You then added this comment. What's the intent here?

app/controllers/application_controller.rb
```diff
@@ -527,9 +514,34 @@ def action_hooks
...
+    return nil unless url.present?
+    begin
+      uri = URI.parse(url)
+    rescue URI::InvalidURIError
+      return nil
+    end
+    if uri.scheme.nil? && uri.host.nil? && uri.path.present? && uri.path.start_with?('/') && !url.start_with?('//')
+      uri.to_s
+    else
+      nil
```
klaustopher, line 544: Returning `nil` here and passing the result to the `redirect_to` method yields an error:


```
ActionController::ActionControllerError: Cannot redirect to nil!
```

What would the intended result be? The current code would present a 500 Internal Server errors to our users.

klaustopher (changes requested):
Hi @anisjellazi,

first of all, thanks for the contribution and helping to make OpenProject more secure.

Unfortunately we cannot merge your PR in the current state.

- First of all, you have removed vital functionality of the `AccountController` like the actions to regsiter and change the password. It was replaced with the comment `# ... other actions and methods remain unchanged ...`. Did you maybe use an AI code assistant to generate the fix and then just copied over the generated file? Otherwise we would need a strong argument why for the sake of this fix we would need to remove functionality to change passwords or self register.
- We already have the well tested [`RedirectPolicy`](https://github.com/opf/openproject/blob/dev/app/policies/redirect_policy.rb) code in place that we use to validate URLs that we want to redirect to. Maybe it would be the better approach to use that helper instead of creating a new one.
- You are applying your `valid_internal_url` helper to the HTTP_REFERRER. By definition the referer **could** be a relative URL, in reality we almost never see that. It will almost always be an absolute URL. So at the very minimum we need to add a check that validates if an absolute URL is to the same host that the application runs on and also allow those.
- If a URL does not match your check, you are returning `nil`. When passing `nil` to the `redirect_to` method, an exception is raised that would show an error page to all of our users. We should have a better handling than that.
- There are currently no tests for your newly introduced behavior. Especially with security related issues it would be very great if you could provide a test case that demonstrates the currently problematic behavior and then shows that the newly introduced code is going to fix it.

So, there's a few problems that we would need to address to go forward with this PR. First and foremost, the unrelated code that was removed would need to be re-added and we should use our already existin policy method instead of introducing a new helper.

If you need further assistance with this, let us know.

PR #17834, opened 2025-02-05
Autocorrect from erb_lint

toy:
```ruby
#!/usr/bin/env ruby

require 'json'
require 'set'
require 'time'

# there are less than 100, no need to go through pages
data = `curl --silent -H "Accept: application/vnd.github.v3+json" 'https://api.github.com/repos/opf/openproject/pulls?state=open&per_page=100'`

threshold = Date.today - 100 # only updated during last 100 days
pr_ids = JSON.parse(data)
  .filter{ Time.iso8601(_1['updated_at']).to_date >= threshold }
  .map{ _1['number'] }

system *%W[git fetch origin dev], *pr_ids.map{ "+refs/pull/#{_1}/head:refs/remotes/origin/pr/#{_1}" }

refs = pr_ids.map{ |pr_id| "origin/pr/#{pr_id}" }
refs += `git for-each-ref --format='%(refname:short)' 'refs/remotes/origin/release/*'`.split("\n")

pr_touched_paths = Set.new

refs.each do |ref|
  pr_touched_paths.merge IO.popen(%W[git diff --name-only ...#{ref}], &:read).split("\n")
end

target_paths = Dir['**/*.erb']

paths = target_paths - pr_touched_paths.to_a

html_paths, non_html_paths = paths.partition{ _1.end_with?('.html.erb', '.turbo_stream.erb') }

system *%w[erb_lint --autocorrect] + html_paths

system *%w[erb_lint --autocorrect --enable-linters rubocop] + non_html_paths
```

Following rubocops were additionally disabled: `Rails/Date`, `Rails/FindEach`, `Rails/LinkToBlank`, `Rails/TimeZone`, `Style/RedundantConstantBase`, `Style/StringConcatenation`


app/components/admin/custom_fields/custom_field_projects/new_custom_field_projects_form_modal_component.html.erb
```diff
@@ -30,20 +30,20 @@ See COPYRIGHT and LICENSE files for more details.
...
       method: :post
     ) do |form|
       concat(render(Primer::Alpha::Dialog::Body.new(
-        id: dialog_body_id, test_selector: dialog_body_id, aria: { label: title },
-        classes: "Overlay-body_autocomplete_height"
-      )) do
+                      id: dialog_body_id, test_selector: dialog_body_id, aria: { label: title },
+                      classes: "Overlay-body_autocomplete_height"
+                    )) do
         render(Projects::CustomFields::CustomFieldMappingForm.new(form, project_mapping: @custom_field_project_mapping))
```
NobodysNightmare: I am wondering whether we need alignment on "yes the rules make sense" before we start enforcing them.

I am commenting this here, because this place looks odd to me. The code inside the block is now to the left of the apparent start of the block.

In the end, this is probably caused (and fixed) by the way that we choose to set the parenthesis here.

E.g.

```
concat(
  render(...)
) do
```

would break acceptably w.r.t. the position of the `do`.

toy: I assume this is rubocop linter inside erb_lint, and I would even prefer it to auto format as you suggest

cbliard: Your example is not correct: the `do` is associated with `render`, so all these are valid from our current rubocop rules
```ruby
concat(render(Something.new(
                ...
              )) do
    # do something
end)
```

or 
```ruby
concat(
  render(Something.new(
           ...
        )) do
    # do something
  end
)
```

or 
```ruby
concat(
  render(
    Something.new(
      ...
    )
  ) do
    # do something
  end
)
```

The third one looks like what you would like to achieve. I don't know if there is a rule that automatically indents the first arg to the next line but that would do it.

AI suggest looking at Layout/FirstArgumentIndentation and Layout/MultilineMethodCallIndentation like this
```
Layout/FirstArgumentIndentation:
  EnforcedStyle: consistent

Layout/MultilineMethodCallIndentation:
  EnforcedStyle: indented
```

When trying with the above additional rules, it formats like this:
```ruby
concat(render(Something.new(
  ...
)) do
  # do something
end)
```

or 
```ruby
concat(
  render(Something.new(
    ...
  )) do
    # do something
  end
)
```

or (this one is the same for both)
```ruby
concat(
  render(
    Something.new(
      ...
    )
  ) do
    # do something
  end
)
```

That really looks like how you'd like it to be.

Regarding my preference, I'm fine with any as long as it's consistent everywhere.

NobodysNightmare: Thanks for the hint about `do` being associated to the `render`. I definitely misread that.

I'd be fine with all of the three formattings you added at the end (which I understand are the auto-format results of the first three after applying the new rules), all of them help in distinguishing inside block vs outside block.

cbliard: I do not know which other implication these new rules would have on our existing `.rb` files, or if everyone is aligned with this change, and I do not have a lot of time to spend on it either.

Could you handle it?

dombesz: @cbliard there is no implication on the `.rb` files, we can safely override rules in `.erb-lint.yml` without affecting `rubocop.yml`.

klaustopher:
Good thing to see all rules in place. Makes it a lot easier to argue about them and some changes. God job 👍

app/components/admin/custom_fields/custom_field_projects/new_custom_field_projects_form_modal_component.html.erb
cbliard: Just looking at some files around, it would reformat this
```ruby
      params.require(:work_package)
            .slice("schedule_manually_touched",
                   "ignore_non_working_days_touched",
                   "start_date_touched",
                   "due_date_touched",
                   "duration_touched")
            .transform_values { _1 == "true" }
            .permit!
```
into this:
```ruby
      params.require(:work_package)
        .slice("schedule_manually_touched",
               "ignore_non_working_days_touched",
               "start_date_touched",
               "due_date_touched",
               "duration_touched")
        .transform_values { _1 == "true" }
        .permit!
```

+++

also this
```ruby
  def moving_work_packages
    @moving_work_packages ||= WorkPackage
                                .for_scheduling(moved_work_packages, switching_to_automatic_mode:)
  end
```

to this
```ruby
  def moving_work_packages
    @moving_work_packages ||= WorkPackage
      .for_scheduling(moved_work_packages, switching_to_automatic_mode:)
  end
```

+++

or also this
```ruby
    attributes = work_package
                   .attributes
                   .slice(*writable_work_package_attributes(work_package))
                   .merge("custom_field_values" => work_package.custom_value_attributes)
                   .merge(overwritten_attributes)
```

to this
```ruby
    attributes = work_package
      .attributes
      .slice(*writable_work_package_attributes(work_package))
      .merge("custom_field_values" => work_package.custom_value_attributes)
      .merge(overwritten_attributes)
```

+++

I find the first one better, but for the two others I find it a bit harder to scan.

For this reason, I would remove `MultilineMethodCallIndentation` (and it was already disabled 5 years ago, so that makes sense... I think).

Only `FirstArgumentIndentation` seems to be needed for what you'd like to do.

Here are the relevant doc links:
- https://rubydoc.info/gems/rubocop/RuboCop/Cop/Layout/FirstArgumentIndentation
- https://rubydoc.info/gems/rubocop/RuboCop/Cop/Layout/MultilineMethodCallIndentation

cbliard: > @cbliard there is no implication on the `.rb` files, we can safely override rules in `.erb-lint.yml` without affecting `rubocop.yml`.

Ha nice! I didn't know. That's nice.
Edit: or rather I forgot, because I created that file :/

NobodysNightmare: OMG. Seeing how many options with weird names this cop already accepts is a good indicator that it is hard to intuitively judge which behaviour one would expect :see_no_evil: 

From the examples I've seen so far, I think `consistent` seems like a good choice, but limiting the change to erb linting for now sounds like the safer choice to me.

That said, I'd really like to hear opinions on my initial statement. I.e. is it only me that finds

```
concat(render(Primer::Alpha::Dialog::Body.new(
                id: dialog_body_id, test_selector: dialog_body_id, aria: { label: title },
                classes: "Overlay-body_autocomplete_height"
              )) do
  render(Projects::CustomFields::CustomFieldMappingForm.new(form, project_mapping: @custom_field_project_mapping))
```

weirdly indented around the `do`? If it's only a me-problem, I'd also be fine not changing anything.

In other words: It's cool that we discussed already that changing this is possible in the linter, but who agrees to the premise?

toy: > In other words: It's cool that we discussed already that changing this is possible in the linter, but who agrees to the premise?

I understand how indent works in that case, but I think it is horrible. For me following would be easily readable, but I'm fine with most variations that @cbliard listed

```ruby
concat(
  render(
    Primer::Alpha::Dialog::Body.new(
      id: dialog_body_id,
      test_selector: dialog_body_id,
      aria: { label: title },
      classes: "Overlay-body_autocomplete_height"
    )
  ) do
    render(
      Projects::CustomFields::CustomFieldMappingForm.new(
        form,
        project_mapping: @custom_field_project_mapping
      )
    )
```

cbliard: Then for what @toy proposed I think we need `Layout/FirstMethodArgumentLineBreak` cop too.

```yaml
Layout/FirstArgumentIndentation:
  EnforcedStyle: consistent

Layout/FirstMethodArgumentLineBreak:
  Enabled: true
```

I like it too. It may span over a lot of lines, but at least it's extra clear.

Could you amend the erb config file with it and rerun the linting to see if that looks better?

toy: I didn't apply everything that can be done automatically by rubocop, like constant resolution was getting broken

toy:
Unexpected fix of `<input type="text"></input>` to `<input type="text"><input>` was shown by feature specs

NobodysNightmare (approved):
I think this is reasonable. I didn't look through the entire diff, but rather did some spot checking of the auto-correct results.

I am not sure whether all the changes (even the ones I suggested) lead to pure improvement, but I think the main thing we achieve now is consistency. This means when we discuss the next rule changes, we will be able to compare what only such a change will mean to the code base, without the noise from all the other fixes that are necessary as well.

**tl;dr:** LGTM, but maybe not the last PR of its kind :)

PR #17835, opened 2025-02-05
Fix js error when time entry has no activity

cbliard:
Noticed while logging time on community on the "My spent time" widget in "My page".

image

This happens for each logged time entry which does not link to an activity. Not harmful, but creates noise, and lead me in the wrong direction while looking for another issue.

### DIFF
--- a/frontend/src/app/features/calendar/te-calendar/te-calendar.component.ts
+++ b/frontend/src/app/features/calendar/te-calendar/te-calendar.component.ts
@@ -632,7 +632,7 @@ export class TimeEntryCalendarComponent implements AfterViewInit, OnDestroy {
           </li>
           <li class="tooltip--map--item">
             <span class="tooltip--map--key">${schema.activity.name}:</span>
-            <span class="tooltip--map--value">${this.sanitizedValue(entry.activity.name)}</span>
+            <span class="tooltip--map--value">${this.sanitizedValue(entry.activity?.name || '')}</span>
           </li>
           <li class="tooltip--map--item">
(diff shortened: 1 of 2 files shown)


myabc (approved):
Good spot! - this fixes the issue on my page!

frontend/src/app/features/calendar/te-calendar/te-calendar.component.ts
```diff
@@ -632,7 +632,7 @@ export class TimeEntryCalendarComponent implements AfterViewInit, OnDestroy {
           </li>
           <li class="tooltip--map--item">
             <span class="tooltip--map--key">${schema.activity.name}:</span>
-            <span class="tooltip--map--value">${this.sanitizedValue(entry.activity.name)}</span>
+            <span class="tooltip--map--value">${this.sanitizedValue(entry.activity?.name || '')}</span>
```
myabc, line 635: I guess we could also update `sanitizedValue` to handle `undefined` and `null`. It's calling `DomSanitizer.sanitize`, which does permit `null` as an value argument: `DomSanitizer.sanitize(context: SecurityContext, value: SafeValue | string | null): string | null`

(very minor point though!)

cbliard: I tried, but it felt more complex than necessary.
`entry.activity?.name` is actually `string | undefined`, not `string | null`, and `DomSanitizer.sanitize` does not accept `undefined` for parameter `value`, only `SaveValue | string | null`.

So I chose the simpler option: ` || ''`.

PR #17837, opened 2025-02-05
Draft: Choose auth method including token exchange

Kharonus:
- Depending on the configuration, the code must be able to authenticate with exchanged tokens over the authorization code flow from standard OAuth2

# And now?
Please check the inline comments I prefixed with `#README`. Those are the thoughts I had. Those should serve as a basis for discussion.

### DIFF
--- a/modules/storages/app/common/storages/peripherals/storage_interaction/authentication_strategies/nextcloud_strategies.rb
+++ b/modules/storages/app/common/storages/peripherals/storage_interaction/authentication_strategies/nextcloud_strategies.rb
@@ -37,10 +37,24 @@ module Storages
             ::Storages::Peripherals::StorageInteraction::AuthenticationStrategies::BasicAuth.strategy
           end
 
-          UserBound = ->(user:) do
-            ::Storages::Peripherals::StorageInteraction::AuthenticationStrategies::OAuthUserToken
-              .strategy
-              .with_user(user)
+          UserBound = ->(user:, storage:) do
+            # README: This check does not match for user that are only hijacked by an IDP based login.
+            token_exchange_preferred = storage.audience.present? && user.authentication_provider.present?
+
+            if token_exchange_preferred
+              ::Storages::Peripherals::StorageInteraction::AuthenticationStrategies::SsoUserToken
+                .strategy
+                .with_user(user)
+            elsif storage.oauth_configuration.present?
+              # README: This allows for potentially optional configuration of OAuth authentication method.
+              ::Storages::Peripherals::StorageInteraction::AuthenticationStrategies::OAuthUserToken
+                .strategy
+                .with_user(user)
+            else
+              # README: If for a user - storage combination no possible authentication method is found, the
+              # noop strategy will result in non authorized requests.
+              ::Storages::Peripherals::StorageInteraction::AuthenticationStrategies::Noop.strategy
+            end
           end
         end
       end
--- a/modules/storages/app/common/storages/peripherals/storage_interaction/authentication_strategies/sso_user_token.rb
+++ b/modules/storages/app/common/storages/peripherals/storage_interaction/authentication_strategies/sso_user_token.rb
@@ -32,17 +32,34 @@ module Storages
   module Peripherals
     module StorageInteraction
       module AuthenticationStrategies
-        module OneDriveStrategies
-          UserLess = -> do
-            ::Storages::Peripherals::StorageInteraction::AuthenticationStrategies::OAuthClientCredentials.strategy
+        class SsoUserToken
+          def self.strategy
+            Strategy.new(:sso_user_token)
           end
 
-          UserBound = ->(user:, storage:) do
-            # README: OneDrive storages won't support any authentication method switching - related to the fact,
-            # that we don't yet plan to rewrite the configuration form yet.
-            ::Storages::Peripherals::StorageInteraction::AuthenticationStrategies::OAuthUserToken
-              .strategy
-              .with_user(user)
+          def initialize(user)
+            @user = user
+          end
+
+          def call(storage:, http_options: {}, &)
+            # README: important remark: this code will NOT automatically refresh, if the token does not have
+            # information about the expiration date. The question is, IF we want to support this case.
+            # In the case above, the token will be used without refresh, which will lead to an unauthorized response
+            # once the token is no longer valid. This can only be solved right now by logging out
+            # and in again - which will replace the account token itself.
+            # Open question: Do we want to support this case?
+            OpenIDConnect::UserTokens::FetchService
+              .new(user: @user)
+              .access_token_for(audience: storage.audience)
+              .either(
+                ->(token) do
+                  opts = http_options.deep_merge({ headers: { "Authorization" => "Bearer #{token}" } })
+                  yield OpenProject.httpx.with(opts)
+                end,
+                ->(error) do
+                  Failures::Builder.call(code: :unauthorized, log_message: error.message)
+                end
+              )
           end
         end
       end
--- a/modules/storages/app/common/storages/peripherals/storage_interaction/nextcloud/util.rb
+++ b/modules/storages/app/common/storages/peripherals/storage_interaction/nextcloud/util.rb
@@ -70,7 +70,9 @@ module Storages
               case auth_strategy.key
               when :basic_auth
                 ServiceResult.success(result: storage.username)
-              when :oauth_user_token
+              when :oauth_user_token || :sso_user_token
+                # README: So, this might lead to problems. How can we ensure, that at this point in time,
+                # the remote identity is already created for token exchange?
                 origin_user_id = RemoteIdentity.where(user_id: auth_strategy.user, oauth_client: storage.oauth_client)
                                                .pick(:origin_user_id)
                 if origin_user_id.present?
--- a/modules/storages/app/models/storages/nextcloud_storage.rb
+++ b/modules/storages/app/models/storages/nextcloud_storage.rb
@@ -70,6 +70,10 @@ module Storages
       authentication_method == "oauth2_sso"
     end
 
+    def audience
+      nextcloud_audience
+    end
+
     def configuration_checks
       {
         storage_oauth_client_configured: oauth_client.present?,
--- a/docker/dev/keycloak/docker-compose.yml
+++ b/docker/dev/keycloak/docker-compose.yml
@@ -1,5 +1,3 @@
-version: "3.9"
-
 services:
   db-keycloak:
     image: postgres:13
--- a/modules/storages/app/common/storages/peripherals/storage_interaction/authentication.rb
+++ b/modules/storages/app/common/storages/peripherals/storage_interaction/authentication.rb
@@ -42,6 +42,8 @@ module Storages
             AuthenticationStrategies::BasicAuth.new
           when :oauth_user_token
             AuthenticationStrategies::OAuthUserToken.new(strategy.user)
+          when :sso_user_token
+            AuthenticationStrategies::SsoUserToken.new(strategy.user)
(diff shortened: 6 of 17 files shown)


modules/storages/app/common/storages/peripherals/storage_interaction/authentication_strategies/nextcloud_strategies.rb
```diff
@@ -37,10 +37,24 @@ module NextcloudStrategies
             ::Storages::Peripherals::StorageInteraction::AuthenticationStrategies::BasicAuth.strategy
           end
 
-          UserBound = ->(user:) do
-            ::Storages::Peripherals::StorageInteraction::AuthenticationStrategies::OAuthUserToken
-              .strategy
-              .with_user(user)
+          UserBound = ->(user:, storage:) do
+            # README: This check does not match for user that are only hijacked by an IDP based login.
+            token_exchange_preferred = storage.audience.present? && user.authentication_provider.present?
```
NobodysNightmare, line 42: I'd call this `sso_preferred`, since the underlying strategy does not solely work with Token Exchange, but with all kinds of tokens obtained through SSO.

modules/storages/app/common/storages/peripherals/storage_interaction/authentication_strategies/nextcloud_strategies.rb
```diff
@@ -37,10 +37,24 @@ module NextcloudStrategies
...
+                .with_user(user)
+            elsif storage.oauth_configuration.present?
+              # README: This allows for potentially optional configuration of OAuth authentication method.
+              ::Storages::Peripherals::StorageInteraction::AuthenticationStrategies::OAuthUserToken
+                .strategy
+                .with_user(user)
+            else
+              # README: If for a user - storage combination no possible authentication method is found, the
+              # noop strategy will result in non authorized requests.
+              ::Storages::Peripherals::StorageInteraction::AuthenticationStrategies::Noop.strategy
```
NobodysNightmare, line 56: Will this result in us making actual requests that will then be rejected?

IIRC we talked about how we should avoid sending out requests that we know will fail eventually.

Would it be better to cause an error/exception here that can be caught before a request goes out?

modules/storages/app/common/storages/peripherals/storage_interaction/authentication_strategies/sso_user_token.rb
```diff
@@ -0,0 +1,68 @@
...
+            @user = user
+          end
+
+          def call(storage:, http_options: {}, &)
+            # README: important remark: this code will NOT automatically refresh, if the token does not have
+            # information about the expiration date. The question is, IF we want to support this case.
+            # In the case above, the token will be used without refresh, which will lead to an unauthorized response
+            # once the token is no longer valid. This can only be solved right now by logging out
+            # and in again - which will replace the account token itself.
+            # Open question: Do we want to support this case?
```
NobodysNightmare, line 50: My opinion: If we want to keep our implementation simple here by enforcing that we know an expiration time, we should at least try to carry the expiration time communicated by the token endpoint along. This should give us an expiration time for the majority of OAuth 2.0 implementations.

If you agree I'd write a feature for that and would probably exchange with Wieland whether this limitation is acceptable.

modules/storages/app/common/storages/peripherals/storage_interaction/nextcloud/util.rb
```diff
@@ -70,7 +70,9 @@ def origin_user_id(caller:, storage:, auth_strategy:)
               case auth_strategy.key
               when :basic_auth
                 ServiceResult.success(result: storage.username)
-              when :oauth_user_token
+              when :oauth_user_token || :sso_user_token
+                # README: So, this might lead to problems. How can we ensure, that at this point in time,
+                # the remote identity is already created for token exchange?
                 origin_user_id = RemoteIdentity.where(user_id: auth_strategy.user, oauth_client: storage.oauth_client)
```
NobodysNightmare, line 76: Can't we "simply" turn this into a "find or create" style method for the SSO case?

I.e. if we find a remote identity, we take it. If we don't find one, we make an API call to figure it out and store it.

This means the first time that an SSO user needs a remote identity, they will fetch it via an API call. At all subsequent times, we will have it "cached" in our remote identities.

modules/storages/app/models/storages/nextcloud_storage.rb
```diff
@@ -70,6 +70,10 @@ def authenticate_via_idp?
       authentication_method == "oauth2_sso"
     end
 
+    def audience
+      nextcloud_audience
```
NobodysNightmare, line 74: It's not yet too late to rename the `nextcloud_audience` property to `audience`, making this alias unnecessary.

modules/storages/app/common/storages/peripherals/storage_interaction/authentication_strategies/nextcloud_strategies.rb
Kharonus: yep, naming is hard :D

modules/storages/app/common/storages/peripherals/storage_interaction/nextcloud/util.rb
NobodysNightmare: From our IRL discussion: The current plan is that there might an an event triggered from the token exchange or from the request of a token for a certain audience.

Since those events are handled synchronously, there is no race condition that can happen here. Once we arrive at this point, the event should have created the `RemoteIdentity`.

Essentially, we can keep the code as-is.

modules/storages/app/models/storages/nextcloud_storage.rb
Kharonus: the method gives us another possibility: having control about which type of storage CAN have an audience.

I have no big pain in keeping it like this for now. But open for anything.

modules/storages/app/common/storages/peripherals/storage_interaction/authentication_strategies/sso_user_token.rb
Kharonus: Not sure I understand, what is this feature proposal :D

NobodysNightmare: Let me back up: Right now we only know of an expiration time if we can parse the token as JWT, read it successfully and extract the `exp` claim. In all other situations, we don't know the expiration and would try to use the token forever.

You are suggesting here, that we never try to refresh the token, unless we know that it's expired.

My opinion: Right now, this is not good enough, because we would fail too often.

My suggestion: We should know the expiration time in more cases, to make your proposal feasible. That's the feature I want to write. Important remark is that we will not be able to cover 100% of cases according to the OAuth 2.0 specification. That's where I'd clarify with Wieland, that we are "good enough" from a product perspective.

modules/storages/app/common/storages/peripherals/storage_interaction/authentication_strategies/nextcloud_strategies.rb
```diff
@@ -37,10 +37,24 @@ module NextcloudStrategies
             ::Storages::Peripherals::StorageInteraction::AuthenticationStrategies::BasicAuth.strategy
           end
 
-          UserBound = ->(user:) do
-            ::Storages::Peripherals::StorageInteraction::AuthenticationStrategies::OAuthUserToken
-              .strategy
-              .with_user(user)
+          UserBound = ->(user:, storage:) do
```
mereghost, line 40: Should we, instead of passing the whole storage, just sending its audience?

mereghost, line 40: Would if be worth making `UserBound` into a full blown object? 🤔

mereghost (commented):
Seems fine overall.

modules/storages/app/common/storages/peripherals/storage_interaction/authentication_strategies/nextcloud_strategies.rb
NobodysNightmare: You mean `UserBound = ->(user:, audience:) { ... }`?

I'd prefer the current way, as it seems to make more sense from a caller's perspective. You hand in a storage and a user to figure out how the user might authenticate.

Otherwise the caller would have to be concerned with how to get an audience from a storage and what that might mean. It also feels like the more brittle interface to offer.

mereghost: More akin to `sso_supported`.

The thing is, we spread the knowledge of a storage to a thing that has no business knowing that it even exists.

modules/storages/app/common/storages/peripherals/storage_interaction/authentication_strategies/sso_user_token.rb
Kharonus: You have my support in raising this as a feature proposal ;)

modules/storages/app/common/storages/peripherals/storage_interaction/authentication_strategies/nextcloud_strategies.rb
Kharonus: That is a question we should answer together. I can understand both approaches.

modules/storages/app/common/storages/peripherals/storage_interaction/authentication_strategies/nextcloud_strategies.rb
Kharonus: @mereghost I think sending just the audience is something very specific. We might fall into the same pit we had for requests, that were build for specific nextcloud and then we had trouble in writing those for OneDrive.

My thoughts are coming from a yesterdays experiment, were Jan and me created an Azure SSO, and we were able  to make server to server request with fully obfuscated audiences, it was a weird UUID.

So, point is, maybe we should have a better abstraction here. What is the entity, that holds the information about all configured authentication methods? right now, it is the storage object.

> we spread the knowledge of a storage

true, but not here. this adapter is called `nextcloud_strategies`. it is bound to a storage type.

Kharonus: How would it look like? Do you mind writing down in pseudo code some of the interface you have in mind?

NobodysNightmare: I'd +1 what Eric said.

Also note that even accepting a parameter like `sso_supported` will also spread knowledge, because the caller of this thing, will suddenly have to _know_ about SSO, to pick an auth strategy, when previously it just needed to know that an auth strategy is broadly related to a user and a storage.

mereghost: > true, but not here. this adapter is called nextcloud_strategies. it is bound to a storage type.

I'd say that it has to know about the "authentication bits" of nextcloud, not the storage itself.

I get the feeling that we are missing an actual abstraction here. We started to draft this with the `OAuthConfiguration` but maybe that was the wrong name.

mereghost: Not very different than it looks like now, but since it is getting smarter and smarter, the fact that it is a lambda kinda bothers me. We can even keep the signature.

```ruby
Very::Good::Name::UserBound.call(user:, storage:)
```

It would probably lead to better test-ability in the end too.

NobodysNightmare:
@Kharonus I just found this PR again while looking for open PRs.

I think this can be closed, correct?

PR #17838, opened 2025-02-05
Authentication method UI

NobodysNightmare:
# Ticket
https://community.openproject.org/projects/cross-application-user-integration-stream/work_packages/60154

# Merge checklist

- [x] Added/updated tests

### DIFF
--- a/app/contracts/composed_contract.rb
+++ b/app/contracts/composed_contract.rb
@@ -1,0 +1,94 @@
+# frozen_string_literal: true
+
+#-- copyright
+# OpenProject is an open source project management software.
+# Copyright (C) the OpenProject GmbH
+#
+# This program is free software; you can redistribute it and/or
+# modify it under the terms of the GNU General Public License version 3.
+#
+# OpenProject is a fork of ChiliProject, which is a fork of Redmine. The copyright follows:
+# Copyright (C) 2006-2013 Jean-Philippe Lang
+# Copyright (C) 2010-2013 the ChiliProject Team
+#
+# This program is free software; you can redistribute it and/or
+# modify it under the terms of the GNU General Public License
+# as published by the Free Software Foundation; either version 2
+# of the License, or (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program; if not, write to the Free Software
+# Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+#
+# See COPYRIGHT and LICENSE files for more details.
+#++
+
+##
+# A contract that allows to compose itself from other contracts like:
+#
+#     class MyContract < ComposedContract
+#       include_contract MyOtherContractA
+#       include_contract MyOtherContractB
+#     end
+#
+# This allows to run the smaller contracts separately, but also to have one contract
+# that validates the model as a whole. The composed contract will make sure that all attributes
+# defined as writable in one contract are also considered writable in other contracts.
+#
+# It's not possible to add validations to the composed contract directly. If you need additional validations,
+# extract them into a separate contract that you include in the composed contract.
+class ComposedContract
+  class << self
+    def included_contract_classes
+      @included_contract_classes ||= []
+    end
+
+    def include_contract(contract_class)
+      included_contract_classes << contract_class
+    end
+  end
+
+  attr_reader :errors, :model, :writable_attributes
+
+  def initialize(model, user, options: {})
+    @model = model
+    @user = user
+    @options = options
+
+    @errors = ActiveModel::Errors.new(model)
+    @writable_attributes = []
+  end
+
+  def validate
+    errors.clear
+    included_contracts.each do |subcontract|
+      subcontract.validate
+      errors.merge!(subcontract.errors)
+    end
+
+    errors.empty?
+  end
+
+  def valid?
+    validate
+  end
+
+  private
+
+  def included_contracts
+    @included_contracts ||= begin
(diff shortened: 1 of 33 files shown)


modules/storages/app/forms/storages/admin/nextcloud_authentication_method_input_form.rb
```diff
@@ -0,0 +1,48 @@
...
+    form do |storage_form|
+      storage_form.select_list(
+        name: :authentication_method,
+        label: I18n.t("activerecord.attributes.storages/storage.authentication_method"),
+        visually_hide_label: false,
+        required: true,
+        caption: I18n.t("storages.instructions.authentication_method"),
+        input_width: :large
+      ) do |list|
+        ::Storages::NextcloudStorage::AUTHENTICATION_METHODS.each do |m|
```
NobodysNightmare, line 42: TODO: We should hide the select for authentication method behind a feature flag

modules/storages/spec/common/storages/peripherals/nextcloud_connection_validator_spec.rb
```diff
@@ -255,11 +255,11 @@
...
     it "returns a validation failure if storage audience isn't set" do
-      storage.update(nextcloud_audience: nil)
+      storage.update!(nextcloud_audience: nil)
 
-      expect(subject.type).to eq(:error)
-      expect(subject.error_code).to eq(:oidc_audience_missing)
-      expect(subject.description).to eq(I18n.t("storages.health.connection_validation.oidc_audience_missing"))
+      expect(subject.type).to eq(:none)
+      expect(subject.error_code).to eq(:wrn_not_configured)
+      expect(subject.description).to eq(I18n.t("storages.health.connection_validation.wrn_not_configured"))
```
NobodysNightmare: @mereghost FYI: I removed the separate audience check from the validator, because after adding a configuration check for the audience to the storage (which was necessary for a complete/incomplete label), the generic configuration check of the validator took control of the error result.

I left this spec here intact, to define the new expected behaviour.

modules/storages/app/common/storages/peripherals/nextcloud_storage_wizard.rb
```diff
@@ -33,16 +33,28 @@ module Peripherals
...
            completed_if: ->(storage) { storage.oauth_client.present? },
            preparation: ->(storage) { storage.build_oauth_client }
 
+      # OAuth 2.0 SSO
+
+      step :nextcloud_audience,
+           section: :oauth_configuration,
+           if: ->(storage) { storage.authenticate_via_idp? },
+           completed_if: ->(storage) { storage.nextcloud_audience.present? },
+           preparation: :prepare_oauth_application
```
NobodysNightmare: copy-pasta

app/contracts/composed_contract.rb
```diff
@@ -0,0 +1,94 @@
...
+  end
+
+  def valid?
+    validate
+  end
+
+  private
+
+  def included_contracts
+    @included_contracts ||= begin
```
ba1ash, line 84: :yellow_circle: Let's use different names for class and instance level variables and their getters to avoid possible confusion. I suggest: 
`included_contract_classes` and `included_contracts` or 
`contract_classes` and `contracts`. Wdyt?

modules/storages/app/components/storages/admin/storage_Info_component.rb
ba1ash, line 1: :yellow_circle: Should `Info` be capitalized in the file name?

modules/storages/app/models/storages/nextcloud_storage.rb
```diff
@@ -78,7 +78,8 @@ def configuration_checks
       {
```
ba1ash, line 78: :yellow_circle: In case of `oauth2_sso` configuration checks report the following, but for user it seems like everything is configured and there is no way to configure `oauth_client` and `oauth_application`.  Should configuration checks be reworked in the scope of this PR?

modules/storages/spec/features/storages/admin/create_storage_spec.rb
```diff
@@ -57,8 +57,9 @@
       end
 
       expect(page).to have_current_path(new_admin_settings_storage_path(provider: "nextcloud"))
+      # TODO: expect(page).to be_axe_clean.within "#content"
```
ba1ash: :green_circle: Should this comment be removed? Let's leave it if you plan to fulfill this TODO in the near future otherwise I am up for removing it.

ba1ash (commented):
The PR seems and works fine for me. I have a couple of questions only.

app/contracts/composed_contract.rb
NobodysNightmare: Makes sense.

modules/storages/app/components/storages/admin/storage_Info_component.rb
NobodysNightmare: Well spotted :D

modules/storages/app/models/storages/nextcloud_storage.rb
NobodysNightmare: Uh, good point. I didn't notice this. Yes, we need to change this, otherwise the storage will not be usable I guess. I will fix it here.

modules/storages/spec/features/storages/admin/create_storage_spec.rb
NobodysNightmare: My plan is to address it soon, but maybe that's too optimistic and I should just write a bug...

modules/storages/app/models/storages/nextcloud_storage.rb
NobodysNightmare: Actually... Let me fix this in the [follow-up PR](https://github.com/opf/openproject/pull/17979), because that one clarifies when which auth method should be used anyways.

ba1ash: Works for me. Feel free to fix it in the follow-up.

ba1ash (approved):
:+1:

PR #17840, opened 2025-02-05
[#60562] Consolidate "Download PDF" with "Generate PDF" 

as-op:
# Ticket

https://community.openproject.org/work_packages/60562

Consolidate "Download PDF" feature with the "Generate PDF" feature in one export modal and add an admin section for work package types to enable/disable/sort PDF export templates

### Modal
Bildschirmfoto 2025-02-18 um 15 05 30

### Settings
Bildschirmfoto 2025-02-18 um 15 04 38

* Use turbo & primer for the modal
* Use JSONB for the storage of the pdf export template settings for a type
   * a list of disabled templates (so no need for a migration to enable existing and future new ones)
   * a sorted list (only if the order has been modified at least once)

# Merge checklist

- [x] Remove the "Download PDF" menu entry
- [x] Add template selection to the export modal dialog
- [x] template selection displays a caption depending on the selected 
- [x] Split hyphenation "Off" from selection into a checkbox
- [x] Auto-select hyphenation language from the current language if available
- [x] Remove paper size select   
- [x] Apply footer text settings to work package exporter
- [x] Remove the feature flag 
- [x] Add template selection to work package types admin section
   - [x] Add data to Type
   - [x] Allow manual sorting of templates in type
   - [x] Allow deactivating templates in type   
- [x] Filter and order template selection in the dialog by work package type
- [x] Added/updated tests


app/models/type/pdf_export_templates.rb
```diff
@@ -0,0 +1,93 @@
...
+    indexes = order.each_with_index.to_a.to_h
+    templates.sort_by { |template| indexes[template.id] }
+  end
+
+  def list_enabled
+    list.filter(&:enabled)
+  end
+
+  def find(template_id)
+    build_in_template = build_in_templates.find { |t| t[:id] == template_id }
```
klaustopher: Just a naming suggestion: You mean the templates that are built in? Then it should be `built_in_templates` otherwise it would be the templates that we do a build in :D

klaustopher (approved):
Added one remark about naming. Tried it out in the browser, works as intended, tried to break it by adding weird stuff to the database. All good 👍

PR #14460, opened 2023-12-21
[#50622] Alias `Storages::NextcloudStorage#automatically_managed?` as `Storages::NextcloudStorage#automatic_management_enabled?` for clarity

akabiru:
### https://community.openproject.org/work_packages/50622

See https://github.com/opf/openproject/pull/13940/files#r1363723710

The predicate `Storages::NextcloudStorage#automatically_managed?` was misleading as a storage itself is not automatically managed, rather the a Project Folder Storage is.

Deprecate `Storages::NextcloudStorage#automatically_managed?` in favour of `Storages::NextcloudStorage#automatic_management_enabled?`

```ruby
[4] pry> storage.automatically_managed?
DEPRECATION WARNING: `#automatically_managed?` is deprecated. Use `#automatic_management_enabled?` instead. NOTE: The new method name better reflects the actual behavior of the storage. It's not the storage that is automatically managed, rather the Project (Storage) Folder is. A storage only has this feature enabled or disabled.
```

PR #17853, opened 2025-02-06
Follow advice of deprecation warning

NobodysNightmare:
I noticed those deprecation warnings while running some tests. I was a bit surprised to see them, because the deprecation was introduced by us, but apparently not yet fixed.


akabiru (approved):
🥇 Thanks Jan! Seems we forgot these traces in https://github.com/opf/openproject/pull/14460

PR #17865, opened 2025-02-07
[#60151] enable nextcloud storages with sso strategy

Kharonus:
# Ticket
[OP#60151](https://community.openproject.org/work_packages/60151)

- have a possibility to authenticate via token exchange

- add decision logic, which strategy to use for nextcloud storages
- add sso user token logic with token exchange


modules/storages/app/common/storages/peripherals/storage_interaction/authentication_strategies/one_drive_strategies.rb
```diff
@@ -37,11 +37,13 @@ module OneDriveStrategies
             ::Storages::Peripherals::StorageInteraction::AuthenticationStrategies::OAuthClientCredentials.strategy
           end
 
-          UserBound = ->(user:) do
+          # rubocop:disable Lint/UnusedBlockArgument
+          UserBound = ->(user:, storage:) do
```
NobodysNightmare: :green_circle: The shorter/less annoying way of disable rubocop here might be to move the disable to the end of this line, because then it will "auto-disable" on following lines.

This would also make sure that if anyone added a block inside here (e.g. a `bind`), that one would be linted again.

```ruby
          UserBound = ->(user:, storage:) do # rubocop:disable Lint/UnusedBlockArgument
```

modules/storages/app/models/storages/storage.rb
```diff
@@ -147,6 +147,13 @@ def available_project_folder_modes
       raise Errors::SubclassResponsibility
     end
 
+    # Returns a value of an audience, if configured for this storage.
+    # The presence of an audience signals that this storage prioritizes
+    # remote authentication over token exchange if possible.
+    def audience
+      raise Errors::SubclassResponsibility
```
NobodysNightmare, line 154: The interface already defines that returning an audience means "I support SSO authentication" and not returning one means that no SSO authentication is supported.

What do you think about implementing the default behaviour already here as `nil` and documenting that subclasses that want to support SSO authentication need to return an audience?

This would effectively allow us to drop the empty implementation over at OneDrive and reduce the amount of boilercode required to implement a minimal storage.

modules/storages/app/models/storages/storage.rb
```diff
@@ -147,6 +147,13 @@ def available_project_folder_modes
       raise Errors::SubclassResponsibility
     end
 
+    # Returns a value of an audience, if configured for this storage.
+    # The presence of an audience signals that this storage prioritizes
+    # remote authentication over token exchange if possible.
```
NobodysNightmare: :green_circle: 

```suggestion
    # remote authentication via Single-Sign-On if possible.
```

modules/storages/app/common/storages/peripherals/storage_interaction/authentication_strategies/nextcloud_strategies.rb
```diff
@@ -37,10 +37,20 @@ module NextcloudStrategies
...
+            if sso_preferred
+              ::Storages::Peripherals::StorageInteraction::AuthenticationStrategies::SsoUserToken
+                .strategy
+                .with_user(user)
+            elsif storage.oauth_configuration.present?
+              ::Storages::Peripherals::StorageInteraction::AuthenticationStrategies::OAuthUserToken
+                .strategy
+                .with_user(user)
+            else
+              ::Storages::Peripherals::StorageInteraction::AuthenticationStrategies::Noop.strategy
```
NobodysNightmare: :yellow_circle:  As discussed in the Draft PR, we should still agree on whether we want to keep using the `Noop` strategy here, which will eventually result in a request being made without any authentication and subsequently fail or whether we want to cause an error for this case.

My preference would be to fail loudly, since we do not expect to come into this situation and thus should fail early and explicitly.

NobodysNightmare (changes requested):
I left a few comments. In addition to them, a _basic_ unit test for `modules/storages/app/common/storages/peripherals/storage_interaction/authentication_strategies/sso_user_token.rb` would make sense IMO.

modules/storages/app/models/storages/storage.rb
Kharonus: I see the thought behind it, yet I would disagree because of one detail.

In a perfect world, this one would be an abstract class and then default behaviour could be implemented here. Yet, in ruby you CAN instantiate a `storage`, so we need to "break" some of his methods.

There is another, even more important paradigm for this, and this is unintended functionality. If the base class implements default behaviour, this can shadow the situation, where a subclass functionality was never defined.

This is why I really like to raise errors on places in a base class where I want to force any developer - even me - to think about, how this needs to be implemented by each subclass.

modules/storages/app/common/storages/peripherals/storage_interaction/authentication_strategies/nextcloud_strategies.rb
Kharonus: Yes, I remember. So, we have the noop strategy, which basically just yields the request block. This is needed for other places as well. This method here needs to return a strategy, and somehow it feels weird to raise an error here. So....

What about the idea of having one more strategy that MUST fail without doing a request? It is a common pattern. We could name it `AuthenticationStrategies::Error` or `AuthenticationStrategies::Failure`.

NobodysNightmare: Sounds good to me, naming wise I'd choose `Failure`.

Kharonus:
Yep, @NobodysNightmare , I was thinking about how to test the new code.

Testing the new strategies makes definitely sense. Should we have a unit test for the `nextcloud_strategies`, too? There is some business logic now.

modules/storages/app/common/storages/peripherals/storage_interaction/authentication_strategies/failure.rb
```diff
@@ -0,0 +1,52 @@
...
+# Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+#
+# See COPYRIGHT and LICENSE files for more details.
+#++
+
+module Storages
+  module Peripherals
+    module StorageInteraction
+      module AuthenticationStrategies
+        class Failure
```
mereghost, line 35: 🟢 The aim here to have a strategy that always fails? 🤔

modules/storages/app/common/storages/peripherals/storage_interaction/authentication_strategies/nextcloud_strategies.rb
```diff
@@ -33,23 +33,28 @@ module Peripherals
...
+                ::Storages::Peripherals::StorageInteraction::AuthenticationStrategies::SsoUserToken
+                  .strategy
+                  .with_user(user)
+              elsif storage.oauth_client.present?
+                ::Storages::Peripherals::StorageInteraction::AuthenticationStrategies::OAuthUserToken
+                  .strategy
+                  .with_user(user)
+              else
+                error "No user-bound authentication strategy applicable for file storage #{storage.id}."
+                ::Storages::Peripherals::StorageInteraction::AuthenticationStrategies::Failure.strategy
```
mereghost: 🟡 Not sure if I like the direction here, but I have no other idea to offer right now.

spec/factories/user_factory.rb
```diff
@@ -34,6 +36,7 @@
     sequence(:mail) { |n| "bobmail#{n}.bobbit@bob.com" }
     password { "adminADMIN!" }
     password_confirmation { "adminADMIN!" }
+    identity_url { nil }
```
mereghost, line 39: 🟢 Is there a case where this is filled and the user wasn't provided by a 3rd party (or the other way around)?

mereghost (approved):
Seems reasonable. I left a couple of questions more out of curiosity / not knowing enough than actual blocking issues.

modules/storages/app/common/storages/peripherals/storage_interaction/authentication_strategies/failure.rb
NobodysNightmare: Exactly, see comment at the :yellow_circle: remark below.

modules/storages/app/common/storages/peripherals/storage_interaction/authentication_strategies/nextcloud_strategies.rb
NobodysNightmare: The alternatives we discussed were:

* Using the `Noop` strategy (original implementation), this would cause us to try a request that we know will fail, because we "forgot" to authenticate
* Raising returning an error right here on the spot

I was personally indifferent between the `Failure` strategy (solution chosen now) and raising an error on the spot, I just didn't like the idea of performing a web request that's guaranteed to fail.

spec/factories/user_factory.rb
NobodysNightmare: > or the other way around)

I think if a user logs in through LDAP, the `identity_url` is empty/nil. Not sure if this counts for what you have in mind, because in that case we present the authentication form, but we forward the password check to LDAP.

modules/storages/spec/common/storages/peripherals/storage_interaction/authentication_strategies/nextcloud_strategies/user_bound_spec.rb
```diff
@@ -0,0 +1,86 @@
...
+
+  context "if file storage is not fully configured" do
+    let(:user) { create(:user) }
+    let(:storage) { create(:nextcloud_storage, nextcloud_audience: nil) }
+
+    it "must return the failure strategy" do
+      strategy = described_class.call(user:, storage:)
+      expect(strategy.key).to eq(:failure)
+    end
+  end
```
NobodysNightmare, line 106: One case that I see missing is local user on SSO-only storage.

What I am wondering: Would the test be more readable if we used nested contexts?

```ruby
context "when the user is local" do
  context "and when the storage is configured for SSO only" do
    it "must return the failure strategy"
  end

  context "and when the storage is configured for OAuth only" do
    it "must use an OAuthUserToken strategy"
  end

  context "and when the storage is configured for both" do
    it "must use an OAuthUserToken strategy"
  end
end

context "when the user is provided by an IDP" do
  context "and when the storage is configured for SSO only" do
    it "must use an SsoUserToken strategy"
  end

  context "and when the storage is configured for OAuth only" do
    it "must use an OAuthUserToken strategy"
  end

  context "and when the storage is configured for both" do
    it "must use an SsoUserToken strategy"
  end
end
```

(at least that would make it easier for me to check that we covered all combinations :D)

modules/storages/spec/common/storages/peripherals/storage_interaction/authentication_strategies/sso_user_token_spec.rb
```diff
@@ -0,0 +1,79 @@
...
+    it "must not yield and return failure" do
+      was_yielded = false
+      result = strategy.call(storage:) { was_yielded = true }
+
+      expect(was_yielded).to be_falsy
+      expect(result).to be_failure
+      expect(result.result).to eq(:unauthorized)
+      expect(result.errors).to be_a(Storages::StorageError)
+      expect(result.errors.data).to eq(error)
+      expect(result.errors.data).to eq(error)
```
NobodysNightmare: Duplicate expectation.

modules/storages/spec/common/storages/peripherals/storage_interaction/authentication_strategies/nextcloud_strategies/user_bound_spec.rb
Kharonus: as we have a 2x3 matrix here, I'm doubting, nested contexts make it easier to read. But I can try to nest it around the user.

modules/storages/app/common/storages/peripherals/storage_interaction/authentication_strategies/nextcloud_strategies.rb
Kharonus: Well, to be precise, we do not know if it would fail with `Noop` 😁 

There might be weird cases, where someone would call `user-bound` on a not fully configured storage but being aware, that the result would be an unauthorized request.

But, TBH, I do not see the use case. So, I´d be fine with the design of having a failure without doing the outbound request.

spec/factories/user_factory.rb
Kharonus: Right now, our whole check for `authentication_provider` on a user goes through this property. It was missing in the factory, hence I added it.

modules/storages/spec/common/storages/peripherals/storage_interaction/authentication_strategies/sso_user_token_spec.rb
Kharonus: oups

spec/factories/user_factory.rb
mereghost: Let me put this other way: can we track the IdP of the user from this? Let's say we have 2 different OIDC IdPs or a mix of OIDC/SAML?

SAML also fills this value and would be absolutely worthless for us. =/ We may need to parse it somehow as it follows the `slug:idp-user-id` format.

modules/storages/app/common/storages/peripherals/storage_interaction/authentication_strategies/nextcloud_strategies.rb
mereghost: Yeah, as I said, it feels weird to have a strategy that will fail instead of failing immediately but I'm out of ideas. :P

spec/factories/user_factory.rb
Kharonus: yep, this is how it is parsed in the user model:

```ruby
  def authentication_provider
    return nil if identity_url.blank?

    slug = identity_url.split(":", 2).first
    AuthProvider.find_by(slug:)
  end
```

mereghost: And we verify that the provider is a `OpenIDConnect::Provider` somewhere else? I think I saw it but for some reason I can't find it now. 🤣

Kharonus: Not, that I know of. Maybe we should?

modules/storages/app/common/storages/peripherals/storage_interaction/authentication_strategies/nextcloud_strategies.rb
Kharonus: Ah, here this is born out of my head. The signature of this method is about to resolve a strategy. So, how would we fail? in here we only could raise an error. but, this might not be nice for the caller, he might want to have proper error handling without rescueing from exceptions.

that is the long and short of this approach.

modules/storages/spec/common/storages/peripherals/storage_interaction/authentication_strategies/nextcloud_strategies/user_bound_spec.rb
NobodysNightmare: Yeah, I am optimistic because one of the two numbers is a "2". That's also why I suggested having the user in the outer context.

spec/factories/user_factory.rb
mereghost: Oh, yeah. We are dealing specifically with OIDC not all SSO solutions so things like SAML/LDAP.

NobodysNightmare: I guess right now we'd want SAML users to use two-way-oauth2 instead of SSO, correct?

mereghost: Yes. Same with LDAP.

NobodysNightmare: I think LDAP does not fill the identity_url.

mereghost (approved):
Noice. :shipit:

PR #17869, opened 2025-02-07
Fix reactions style

toy:
# Ticket
[OP#61402](https://community.openproject.org/wp/61402), related to [OP#40437](https://community.openproject.org/wp/40437)

Make icons square, a bit increase icon/font size and reduce distance between icons.

## Screenshots
|was|now|
|---|---|
|image|image|

# Merge checklist

- [ ] Added/updated tests


marcalcobe:
Hi @toy, these were never my designs. @psatyal worked on this, so I suppose is it better if he does the review on this.

toy:
@marcalcobe Certainly, thank you!

app/components/work_packages/activities_tab/journals/item_component/add_reactions.sass
```diff
@@ -2,3 +2,6 @@
   row-gap: var(--base-size-4, 4px)
   @media screen and (max-width: $breakpoint-sm)
     max-width: 200px
+
+.op-add-reactions-button
+  font-size: 15px
```
akabiru: 🟠 are there any primer specific sizes we can use?

ref: https://primer.style/css/storybook/?path=/story/utilities-typography--font-size , https://primer.style/css/storybook/?path=/story/utilities-typography--font-size

app/components/work_packages/activities_tab/journals/item_component/reactions.sass
```diff
@@ -1,6 +1,10 @@
 .op-emoji-reactions--gap
   row-gap: var(--base-size-4, 4px)
 
+.op-reactions-button
+  padding: 0 7.5px
+  font-size: 15px
```
akabiru: ditto on primer specific sizes

akabiru (commented):
Nice one Ivan, I like the improvements 🎨  - I left some questions on using primer specific font sizes. It would also be good to link to a ticket as this does change the UI (original FEATURE#40437)

app/components/work_packages/activities_tab/journals/item_component/add_reactions.sass
toy: Increasing emoji size was one of my goals, so 16px (`--h4-size`) instead of 15px is also fine

akabiru: Nice, I think [`.f4`](https://primer.style/css/storybook/?path=/story/utilities-typography--font-size) should have you covered. "h4 size" sounds "un-semantic" 

But perhaps @HDinger has better ideas 🙂 

https://github.com/opf/primer_view_components/blob/54762ad996a26b24185a08804b5b96b34039812d/app/lib/primer/css/utilities.css#L6720

app/components/work_packages/activities_tab/journals/item_component/reactions.sass
```diff
@@ -1,6 +1,9 @@
 .op-emoji-reactions--gap
   row-gap: var(--base-size-4, 4px)
 
+.op-reactions-button .Button-label
```
akabiru: 🟢 One thing to watch out for is breaking changes from primer changing class names- I reckon the odds are low but IIRC we try to avoid styling against primer internal class names for this reason.

akabiru (approved):
Top work, again- love the new design! 🥇

akabiru:
> Top work, again- love the new design! 🥇

@toy I like the font size adjustment but I'm having second thoughts about the rounded emoji counters- one point @psatyal made earlier in the design was avoid drawing attention to emojis reactions - the rounded nature calls them out visually more than I'd like and they can be perceived as notifications as that's what we use for the notification center- add the tab counters and it seems like too much visual noise.

Screenshot 2025-02-10 at 10 24 52 PM

psatyal:
Hi @toy,

These are very interesting suggestions for improvement. Thank you. 

Could I ask you what triggered this work? I don't see a work package for this feature, so I don't have context on what motivated it or why we are priotising this. Generally, we don't make direct changes to the design without first writing a feature with acceptance criteria. Having a work packge for work we're doing is important so so we can:

- prioritise them in relation to other features and optimisations
- document the changes we make inn OpenProject
- have properly defined specs for testing/QA 
- notify relevant people (marketing who does screenshots/documentation)

 As for the changes themselves, I think the reduced padding is definitely an improvement. 

- Is this implementation using the trailing counter on invisible buttons for the number? I'm not sure the extra visual complexity introduced by the badges is an improvement, especially for emojis of other people, which are of the 'invisible' variety (without a border).
- What was the reason for the larger font size?

emoji-design-changes-user-and-outside

app/components/work_packages/activities_tab/journals/item_component/reactions.sass
HDinger: I'd strongly vote against this whole block. As @akabiru said before, if Primer changes the class name, this will break without us noticing it. Further, what is even more important to me: You are overwriting the Primer default behaviour here which in itself is something that we should avoid. There are defined styles for the gap between the label and the trailing content. We should stick with that.

app/components/work_packages/activities_tab/journals/item_component/reactions.html.erb
```diff
@@ -3,29 +3,32 @@
...
+                tag: :a,
+                href: href(reaction:),
+                data: {
+                  turbo_stream: true,
+                  turbo_method: :put,
+                  "work-packages--activities-tab--index-target": "reactionButton"
+                },
+                aria: { label: aria_label_text(reaction, data[:users]) },
+                disabled: current_user_cannot_react?,
+                classes: %w[op-reactions-button f4 px-2]
```
HDinger: Please do not use these classes directly. Instead use the `system_arguments` of Primer. Thus, we can be sure that the code works even if they change the class names in the future

app/components/work_packages/activities_tab/journals/item_component/reactions.html.erb
```diff
@@ -3,29 +3,32 @@
...
+              )
+            ) do |button|
+              button.with_tooltip(
+                text: number_of_user_reactions_text(data[:users]),
+                test_selector: "reaction-tooltip-#{reaction}"
+              )
+              button.with_trailing_visual_counter(count: data[:count])
```
HDinger: I like that we are using the default Primer slot here now 👍

HDinger (changes requested):
Hi @toy 

I appreaciate your proactiveness 👍 I agree with @psatyal, it is hard to evaluate that change without a ticket assigned to it which is specified and accepted. Of course we can all give our two scents to it, but in the end we have a product team whose task it is to think about a consistent and thourough UI. We should not simply skip that. 
I added some remarks to the code but would like to wait for the feedback of the product team.

psatyal:
I created a new work package for this:
https://community.openproject.org/projects/communicator-stream/work_packages/61402/activity 

@toy: I can think you can address my points in the work package (I've copied my comment over) Henriette's ones here since they are more code-related.

toy:
> Could I ask you what triggered this work?

Just an urge, I felt that buttons should be square instead of rectangular and that icons/emojis were a bit smallish.
Also showing is more efficient then describing and mocking in html is easier for me.

toy:
@psatyal Thank you for creating the work package

app/components/work_packages/activities_tab/journals/item_component/add_reactions.html.erb
```diff
@@ -34,7 +34,8 @@
                     turbo_method: :put,
                     "work-packages--activities-tab--index-target": "reactionButton"
                   },
-                  aria: { label: I18n.t("reactions.react_with", reaction: reaction.to_s.humanize(capitalize: false)) }
+                  aria: { label: I18n.t("reactions.react_with", reaction: reaction.to_s.humanize(capitalize: false)) },
+                  classes: %w[op-add-reactions-button f4]
```
HDinger: Please do not use these classes (`f4`) directly but rather the system_arguments, e.g `font_size: 4`

app/components/work_packages/activities_tab/journals/item_component/reactions.html.erb
```diff
@@ -3,29 +3,31 @@
...
+                tag: :a,
+                href: href(reaction:),
+                data: {
+                  turbo_stream: true,
+                  turbo_method: :put,
+                  "work-packages--activities-tab--index-target": "reactionButton"
+                },
+                aria: { label: aria_label_text(reaction, data[:users]) },
+                disabled: current_user_cannot_react?,
+                classes: %w[op-reactions-button px-2]
```
HDinger: Same as before, please use the system_arguments

app/components/work_packages/activities_tab/journals/item_component/add_reactions.sass
```diff
@@ -2,3 +2,7 @@
   row-gap: var(--base-size-4, 4px)
   @media screen and (max-width: $breakpoint-sm)
     max-width: 200px
+
+.op-add-reactions-button
+  width: var(--control-medium-size)
+  padding: unset
```
HDinger, line 8: `var(--control-medium-size)` is intended to define the height of a control. The width is usally defined by the content itself. Actually, you don't need all of that. Just set `px: 1` on the Button directly and it should get squared.

app/components/work_packages/activities_tab/journals/item_component/reactions.html.erb
```diff
@@ -3,29 +3,31 @@
...
+              )
+            ) do |button|
+              button.with_tooltip(
+                text: number_of_user_reactions_text(data[:users]),
+                test_selector: "reaction-tooltip-#{reaction}"
+              )
+              render(Primer::Beta::Text.new(classes: "f4")) { EmojiReaction.emoji(reaction) } + " #{data[:count]}"
```
HDinger:
```suggestion
              render(Primer::Beta::Text.new(font_size: 4)) { "#{EmojiReaction.emoji(reaction)} #{data[:count]}" }
```

HDinger (changes requested):
The outcome looks nice 👍 I left some code remarks

app/components/work_packages/activities_tab/journals/item_component/add_reactions.sass
toy: `px: 1` is too little:
image
`px: 2` is also not sqaure - 2px too much (32×30):
image
And `var(--control-medium-size)` for width sets it to exact same value as height

app/components/work_packages/activities_tab/journals/item_component/reactions.html.erb
toy: The text component is needed to have increased font size only for emoji, but not count

HDinger: 👍

app/components/work_packages/activities_tab/journals/item_component/add_reactions.sass
HDinger: Got it, thanks 👍

HDinger (approved):
Thanks for investing in this 👍 🙇

HDinger:
Merging as the failing tests are unrelated.

PR #17872, opened 2025-02-08
improve performance of wp mentionable users filter

ulferts:
# Ticket

https://community.openproject.org/wp/61365

Improve the performance of the `Queries::Principals::Filters::MentionableOnWorkPackageFilter` which is central for returning the users available when trying to mention somebody in a work package comment. On some SaaS instances, the `api/v3/principals` endpoint using this filter currently has an impact of 50%. 

image

This is caused in part by the filter, when checking for the allowed values fetching the ids of all work packages the requesting user can see. 

image

This is done to validate that the filter is not used for probing for information. While this is necessary to be done, it can be done in a more performant way. Instead of fetching all the ids and then comparing those to the filter's values, the filter can check within the DB that the values are within the visible scope using ARs `exists?` functionality.    

This mechanism should be applied to other `HugeList` filters as well but this is out of scope for this drive by fix.

### DIFF
--- a/app/models/queries/filters/strategies/huge_list.rb
+++ b/app/models/queries/filters/strategies/huge_list.rb
@@ -33,10 +33,18 @@ module Queries::Filters::Strategies
 
     def validate
       unique_values = values.uniq
-      allowed_and_desired_values = allowed_values_subset & unique_values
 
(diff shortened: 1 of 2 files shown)


klaustopher (approved):
👏

PR #17881, opened 2025-02-10
Rework of Storages Registry based approach

mereghost:
#### Yes, this is a large chunk of code. No, it can't be done piecemeal as this would increase complexity tenfold.

It can be reviewed piecemeal tho. :P

# What is this?

For a long time we experimented and introduced ideas to the `Storages` code-base that were left half-baked, half implemented or outright abandoned.

This work tries to extrapolate these ideas and move forward with this type of implementation.

# Why?

`Storages` code will forever get more complexity as we support more and more features and should, by all means, be invisible to the "outside". Outside code should just call a couple of services and work with their results using the already familiar `ServiceResult` abstraction.

Module internal code should need to keep this level of promises. Besides we came up with a bunch of other abstractions (input data, result objects) that we never ended up implementing for all the cases.

# What has changed?

For storages client code? Hopefully nothing. `Storages` module has been radically transformed tho.

### Adapters

First, we call the different Storage Provider implementations by what they are `Adapters`. Most of the code that lived under `Peripherals`, has been moved to the `Adapters` namespace.

The current structure looks like so:

1. Adapters
  2. Authentication
  3. Input
  4. Results
  5. Providers
     1. Provider
        1. Commands
        2. Queries
        3. Validators
        4. ProviderResults

The entirety of the adapters namespace makes heavy use of `dry-rb`, specifically `monads` and `validation`. Then we have Services that will make those into `ServiceResult` with `ActiveModel::Errors` as its `error` value.

#### Input

Those are `Data` subclasses that provide a `build` method that runs a validation on the args, guaranteeing a modicum of structure. The result will be either a `Success(Input)` or a `Failure(Dry::Validation::Result)` in all cases.

This allows the caller to handle parameter errors before even calling the the `Provider`. This also ensures that the commands will have exactly the same input structure. 

#### Results

Same idea with the `Input` , all structured results will be wrapped in a `Data` subclass to ensure compatibility between providers.

\<MORE INFO SOON... check back later...\>


modules/storages/app/common/storages/adapters/input/strategy_contract.rb
```diff
@@ -0,0 +1,44 @@
...
+# See COPYRIGHT and LICENSE files for more details.
+#++
+
+module Storages
+  module Adapters
+    module Input
+      class StrategyContract < Dry::Validation::Contract
+        params do
+          # The included list need to be made dynamic
+          required(:key).filled(:symbol, included_in?: %i[noop basic_auth oauth_client_credentials oauth_user_token])
```
Kharonus: 🟡 how so "dynamic"?

modules/storages/app/common/storages/adapters/authentication.rb
```diff
@@ -0,0 +1,66 @@
...
+        when :oauth_client_credentials
+          AuthenticationStrategies::OAuthClientCredentials.new(auth.use_cache)
+        else
+          raise Errors::UnknownAuthenticationStrategy, "Unknown #{auth.key} authentication scheme"
+        end
+      end
+
+      # TODO: Add tests for this - 2025-01-15 @mereghost
+      def self.authorization_state(storage:, user:)
+        auth_strategy = Input::Strategy.build(user:, key: :oauth_user_token)
```
Kharonus: this is still a result at this point. this is missing error handling, right?

Or, you could now probably build a real monadic chain, right? As I see two results - the `strategy.build` and the `auth_check.call`.

modules/storages/app/common/storages/adapters/authentication_strategies/authentication_strategy.rb
Kharonus, line 1: 🟢 the rework of the strategies looks quite nice

modules/storages/app/common/storages/adapters/providers/nextcloud/base.rb
```diff
@@ -0,0 +1,81 @@
...
+module Storages
+  module Adapters
+    module Providers
+      module Nextcloud
+        class Base
+          include TaggedLogging
+          include Dry::Monads::Result(Results::Error)
+
+          def self.call(storage:, auth_strategy:, input_data:)
+            new(storage).call(auth_strategy: auth_strategy, input_data: input_data)
```
Kharonus: 🟡 shorthand for parameters in `.call`?

modules/storages/app/common/storages/adapters/providers/nextcloud/base.rb
Kharonus, line 1: 🟢 nice, so the base class replaces the util class

modules/storages/app/common/storages/adapters/providers/nextcloud/queries/propfind_query.rb
Kharonus, line 1: ❓ Why did you dissolve the idea of the `internal` namespace? Now I'd stumble across the fact, that the propfind query has a fully different interface then its brethren. It even overrides `self.call` and stuff like this.

Kharonus (commented):
Some first comments

modules/storages/app/common/storages/adapters/input/strategy_contract.rb
mereghost: Goood question. Ideally we could have this somewhere and just reference it. XD

modules/storages/app/common/storages/adapters/authentication.rb
mereghost: That's the idea, the monadic chain.

The code that's asking for this will be touched soon, so I'll have to take a better look at it.

modules/storages/app/common/storages/adapters/providers/nextcloud/base.rb
mereghost: 🤦🏼 Somehow I missed this.

modules/storages/app/common/storages/adapters/providers/nextcloud/queries/propfind_query.rb
mereghost: It wasn't a conscious choice... it kinda happened. I played with the idea of making the interface the same but never fully committed to it. 

They aren't exposed by the registry, so the interface is a lesser issue to me. I might still reintroduce the namespace.

app/models/remote_identity.rb
```diff
@@ -35,4 +35,7 @@ class RemoteIdentity < ApplicationRecord
 
   validates :user, uniqueness: { scope: %i[auth_source integration] }
   validates :origin_user_id, :user, :auth_source, :integration, presence: true
+
+  # FIXME: This needs a better name - 2025.03.18 @mereghost
+  scope :of_user_and_client, ->(user, client, integration) { find_by(user:, auth_source: client, integration:) }
```
NobodysNightmare, line 40: The very least is `of_user_auth_source_and_integration`, because it's by now requiring different arguments.

In the end, remote identities are defined on the tuple of user, auth method and integration, because changing any of the three could result in a different remote identity:

* different users, obviously have different identities
* the same user can have different identities at different integrations (e.g. nextcloud uses your username, XYZWiki may use an integer identifier)
* even at the same integration, one user could have different identities based on auth method (e.g. your OAuth identity might be linked to a different remote user than for SSO login)

So is there a name for that? What about a simple, but parameter-safe:

```
scope :for, ->(user:, auth_source:, integration:) { find_by(user:, auth_source:, integration:) }
```

The simple advantage over `find_by` would be that all kwargs are required to be passed, so you can't forget a required arg and thus get a wrong result.

app/models/user.rb
```diff
@@ -320,6 +320,10 @@ def human_authentication_provider
     authentication_provider&.display_name
   end
 
+  def provided_by_oidc?
+    authentication_provider.is_a?(OpenIDConnect::Provider)
```
NobodysNightmare, line 333: Future-proofing hints:

1. We'll soon have the possibility of more than one provider associated to the same user (logging in through multiple was always possible, but until recently we only stored the most recent provider)
2. The fact that an `authentication_provider` is present, doesn't even mean that the user used it to log-in

The more I think about this, the more I believe that these problems are out of scope for your changes here. So take this as a mere FYI ^^

lib/api/errors/outbound_request_forbidden.rb
```diff
@@ -35,7 +35,7 @@ class OutboundRequestForbidden < ErrorBase
       code 500
 
       def initialize(message = I18n.t("api_v3.errors.code_500_outbound_request_failure", status_code: 403))
-        super
+        super(message || I18n.t("api_v3.errors.code_500_outbound_request_failure", status_code: 403))
```
NobodysNightmare, line 38: This change looks incomplete. The default value in the method signature is the same as the fallback here, so should the default value be changed to `message = nil` for conciseness?

It's also surprising that the initializer allows overwriting the `status_code`, but we 100% ignore this.

(same same for other errors defined below, I'll not repeat this comment to keep discussions focussed)

modules/storages/app/common/storages/adapters/authentication.rb
```diff
@@ -0,0 +1,88 @@
...
+            AuthenticationStrategies::OAuthClientCredentials.new(auth.use_cache)
+          when :sso_user_token
+            AuthenticationStrategies::SsoUserToken.new(auth.user)
+          else
+            raise Errors::UnknownAuthenticationStrategy, "Unknown #{auth.key} authentication scheme"
+          end
+        end
+        # rubocop:enable Metrics/AbcSize
+
+        # TODO: Needs update for OIDC. Add tests for this.
```
NobodysNightmare, line 60: TODO still relevant? Quick glimpse from my side would say "yes", unless you changed more UI code etc. to make this work.

app/models/user.rb
mereghost: Yeah, this will need to be updated once we have the multiple auth providers in. =/

modules/storages/app/common/storages/adapters/authentication.rb
mereghost: It is a maybe. It seems to work and tests pass, but maybe we would need to see if all cases are covered.

ba1ash (approved):
LGTM! Thanks for your efforts, Marcello :heart:

PR #17882, opened 2025-02-10
Replace with_merged_former_errors by validate_and_merge_errors

NobodysNightmare:
The main problem with the previous method was that it only worked if the contract being validated inside the block was sharing the same `errors` object as the contract calling it. The old implementation was incapable of incorporating errors from a contract that had its own `errors` object. Sharing of the `errors` object is pretty common among our contracts, because most of them just delegate the `errors` method (among others) to the model being validated.

In my opinion it's a code smell that our contracts use the `errors` of the model they are validating as the place to store their validation result, however I didn't dare touching this yet, because changing that would certainly be a huge change (e.g. because it means that validating the contract will change the model being inspected). I can also imagine that some places rely on it for historic reasons (because using "classic" validations also puts the errors on the model). However, I think that no two contracts should RELY on using the SAME errors object, just because the way we implemented them happens to cause them to share the same object.

This PR was coupled out from #17838 to be able to discuss and test this change separately. In #17838 I want to add a `ComposedContract` which is essentially a way of combining multiple contracts into one.

That class is not reusing the errors object of the model under test, because that would introduce an unintended coupling between the contracts that I'd need to work around.

### DIFF
--- a/app/contracts/base_contract.rb
+++ b/app/contracts/base_contract.rb
@@ -283,11 +283,29 @@ class BaseContract < Disposable::Twin
     end
   end
 
-  def with_merged_former_errors
+  def validate_and_merge_errors(contract)
+    # Duplicating errors beforehand, in case the other contract shares
+    # the errors with us (because by default contracts reuse the errors
+    # of the model they validate) to prevent that existing errors are cleaned
+    # during validation
     former_errors = errors.dup
 
-    yield
+    contract.validate
 
-    errors.merge!(former_errors)
+    # Another side effect of sharing errors is that they may have already been added to
+    # our list of errors, after the validation was executed, so we manually filter out
+    # duplicates.
+    # To avoid looping over all the errors we just added (forever), we copy the errors of the
+    # contract into an array first
+    contract_errors = contract.errors.each.to_a
+    merge_errors_deduplicated(former_errors.chain(contract_errors))
+  end
+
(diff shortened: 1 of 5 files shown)


NobodysNightmare:
What makes me sad is that the implementation of the new method _should_ be

```
contract.validate
errors.merge!(contract.errors)
```

This is not possible, mostly because we need to deal with the fact that we _might_ be sharing the `errors` object with the merged contract (or a contract _called by_ the merged contract). That I discovered a bug in `added?` along the way doesn't help the amount of necessary method comments either.

mereghost:
I think that this was on purpose due to how dependent services worked. You might want @ulferts or @oliverguenther context on this.

ulferts (approved):
The change should be fine.

It is a bit harder to tell since the PR does not provide any specs which makes understanding the purpose of the change harder.

The move to using the model's error object actually is a fairly new development that was made necessary by shifting to Primer forms. They rely on the errors being in the AR `errors` object. While it would have been possible to have a decorator or some other mechanism to wrap the model and the error into a single object to work with Primer forms, that would have been necessary to be done in all the places that use Primer forms which in the long run should be a large portion of the application.

NobodysNightmare:
The comment about missing specs is fair. I am going to address that, because the implementation actually caused some headaches for me.

PR #17884, opened 2025-02-10
Storage Health Check: Check for presence of an OIDC IdP

mereghost:
### Related WP: [OP#61339](https://community.openproject.org/projects/cross-application-user-integration-stream/work_packages/61339/activity)

This will check the presence of an OIDC IdP only if the `oauth_sso` auth method is selected.

### DIFF
--- a/modules/storages/app/common/storages/peripherals/nextcloud_connection_validator.rb
+++ b/modules/storages/app/common/storages/peripherals/nextcloud_connection_validator.rb
@@ -54,7 +54,7 @@ module Storages
 
       private
 
-      def has_oidc_configuration_error?
+      def sso_misconfigured
         return None() unless @storage.authenticate_via_idp?
 
         openid_connect_idp_absent?
@@ -73,7 +73,7 @@ module Storages
         host_url_not_found
           .or { missing_dependencies }
           .or { version_mismatch }
-          .or { has_oidc_configuration_error? }
+          .or { sso_misconfigured }
           .or { with_unexpected_content }
           .or { capabilities_request_failed_with_unknown_error }
       end
--- a/modules/storages/app/common/storages/peripherals/nextcloud_connection_validator.rb
+++ b/modules/storages/app/common/storages/peripherals/nextcloud_connection_validator.rb
@@ -54,10 +54,26 @@ module Storages
 
       private
 
+      def has_oidc_configuration_error?
+        return None() unless @storage.authenticate_via_idp?
+
+        openid_connect_idp_absent?
+      end
+
+      def openid_connect_idp_absent?
+        return None() if OpenIDConnect::Provider.any?
+
+        Some(ConnectionValidation.new(type: :error,
+                                      error_code: :oidc_provider_missing,
+                                      timestamp: Time.current,
+                                      description: I18n.t("storages.health.connection_validation.oidc_provider_missing")))
+      end
+
       def has_base_configuration_error?
         host_url_not_found
           .or { missing_dependencies }
           .or { version_mismatch }
+          .or { has_oidc_configuration_error? }
           .or { with_unexpected_content }
           .or { capabilities_request_failed_with_unknown_error }
       end
--- a/modules/storages/app/common/storages/peripherals/nextcloud_connection_validator.rb
+++ b/modules/storages/app/common/storages/peripherals/nextcloud_connection_validator.rb
@@ -57,10 +57,10 @@ module Storages
       def sso_misconfigured
         return None() unless @storage.authenticate_via_idp?
 
-        openid_connect_idp_absent?
+        openid_connect_idp_absent
       end
 
-      def openid_connect_idp_absent?
+      def openid_connect_idp_absent
         return None() if OpenIDConnect::Provider.any?
 
         Some(ConnectionValidation.new(type: :error,
--- a/modules/storages/spec/common/storages/peripherals/nextcloud_connection_validator_spec.rb
+++ b/modules/storages/spec/common/storages/peripherals/nextcloud_connection_validator_spec.rb
@@ -230,7 +230,7 @@ RSpec.describe Storages::Peripherals::NextcloudConnectionValidator do
     end
   end
 
-  context "when OpenID Connect is available" do
+  context "when OpenID Connect is enabled" do
     let(:storage) { create(:nextcloud_storage_configured, :oidc_enabled) }
     let(:app_version) { Storages::SemanticVersion.parse("2.6.3") }
     let(:capabilities_response) do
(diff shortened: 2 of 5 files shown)


modules/storages/app/common/storages/peripherals/nextcloud_connection_validator.rb
```diff
@@ -54,10 +54,26 @@ def validate
 
       private
 
+      def has_oidc_configuration_error?
```
NobodysNightmare: :yellow_circle: Since the method name ends in a `?`, I'd expect that I can interpret the result of this method as a boolean. However, as far as I can tell, both `None` and `Some` are truthy values.

Since this is used in the configuration checks below, what do you think about

```suggestion
      def sso_is_misconfigured
```

1. Rephrased to be part of a sentence ("or SSO is misconfigured")
2. Replacing OIDC with SSO to try and achieve consistency in naming with regards to that (Nit)

modules/storages/spec/common/storages/peripherals/nextcloud_connection_validator_spec.rb
```diff
@@ -230,6 +230,25 @@
     end
   end
 
+  context "when OpenID Connect is available" do
```
NobodysNightmare: :yellow_circle: Context name sounds misleading to me. From what I read, OpenID Connect (aka _SSO_ Authentication) is **enabled**, but there is no provider **available**.

```suggestion
  context "when OpenID Connect is enabled" do
```

modules/storages/spec/common/storages/peripherals/nextcloud_connection_validator_spec.rb
```diff
@@ -230,6 +230,25 @@
...
+    let(:capabilities_response) do
+      ServiceResult.success(result: Storages::NextcloudCapabilities.new(
+        app_enabled?: true,
+        app_version:,
+        group_folder_enabled?: false,
+        group_folder_version: nil
+      ))
+    end
+
+    it "errors if no Provider is available" do
```
NobodysNightmare, line 251: Should we add

```ruby
it "succeeds if a Provider is available"
```

?

(depending on taste in a nested context or immediately, like written above)

modules/storages/spec/factories/storage_factory.rb
```diff
@@ -92,7 +92,12 @@
     provider_type { Storages::Storage::PROVIDER_TYPE_NEXTCLOUD }
     sequence(:host) { |n| "https://host#{n}.example.com/" }
     authentication_method { "two_way_oauth2" }
-    nextcloud_audience { "nextcloud" }
+    nextcloud_audience { nil }
+
+    trait :oidc_enabled do
```
NobodysNightmare, line 97: :green_circle: Naming nit: I'd repeat the `sso` terminology here, especially since the authentication method does not try to imply OIDC

```suggestion
    trait :sso_authentication do
```

```suggestion
    trait :sso_enabled do
```

NobodysNightmare (changes requested):

modules/storages/app/common/storages/peripherals/nextcloud_connection_validator.rb
mereghost: About the the query method, I was just following the pattern already existing in the object itself.

modules/storages/spec/common/storages/peripherals/nextcloud_connection_validator_spec.rb
mereghost: Oh, good catch!

modules/storages/spec/factories/storage_factory.rb
mereghost: Hmmmmm... I'm starting to wonder if going with SSO won't be a mistake as we have other SSO solutions supported on OpenProject, like SAML.

modules/storages/spec/common/storages/peripherals/nextcloud_connection_validator_spec.rb
mereghost: Added!

modules/storages/app/common/storages/peripherals/nextcloud_connection_validator.rb
NobodysNightmare: I see, because there are two methods that also end in `?` while returning some or none. IMO we should remove the question mark from those two methods as well, all the others already behave nicely.

(happy to have this out-of-scope for this PR)

mereghost: Ended up going with your suggestion. Let's see how this shapes up as I implement the rest of the checks. =)

modules/storages/spec/factories/storage_factory.rb
NobodysNightmare: To be fair, that's why it was called `oauth2_sso` and not `sso` alone :D

I see what you mean, but right now it sounds as likely to me that adding OIDC to names is a mistake, as it could be that dropping OIDC is a mistake.

I.e. the implementation could develop to be super duper tightly coupled to OIDC and we'd need to do something entirely different for SAML. Or the implementation develops in a direction where the storage just cares about the thing being "some kind of SSO", but which kind exactly is not important.

Right now we don't know, so we only guess which name will be better. I am mostly advocating for having the term "SSO" included, because the distinguishing factor between the solutions we have today, is that one is SSO and the other is not SSO. Whether or not we qualify the term "SSO" with OIDC, I have less strong feelings about, because we don't know how that will develop.

modules/storages/app/common/storages/peripherals/nextcloud_connection_validator.rb
```diff
@@ -54,10 +54,26 @@ def validate
 
       private
 
+      def sso_misconfigured
+        return None() unless @storage.authenticate_via_idp?
+
+        openid_connect_idp_absent?
+      end
+
+      def openid_connect_idp_absent?
```
NobodysNightmare: Can we get rid of this question mark as well? :D As I wrote, I see where you got them, but so far only two out of ... eight previously existing methods had a question mark appended. Let's make sure we kill the disease before it spreads :D

NobodysNightmare (approved):

modules/storages/spec/factories/storage_factory.rb
mereghost: `audience` and `authentication_method` are already pretty tied to OIDC as is implementation wise and we have no plans in the short/medium term of supporting anything else than OIDC for this.

We could even extend this trait to supply a `OpenIDConnect::Provider` (which I think I'll need down the line).

NobodysNightmare: `oidc_sso_enabled`?

mereghost: Sounds good. I'll change it on the next PR unless this CI run fails. XD

PR #17886, opened 2025-02-10
Use @-mention of user when quote replying to a comment

brunopagno:
# Ticket
https://community.openproject.org/projects/communicator-stream/work_packages/37093/activity

Make it so that quoting a message from a user, actually mentions them

## Screenshots

https://github.com/user-attachments/assets/366a5660-73c4-4185-b424-c0eb8c217dfc

I found out what text the CodeMirror would produce and manually inject the same text into the editor. It seems to translate accurately into a mention and even persists it correctly.

# Merge checklist

- [X] Added/updated tests

### DIFF
--- a/frontend/src/stimulus/controllers/dynamic/work-packages/activities-tab/index.controller.ts
+++ b/frontend/src/stimulus/controllers/dynamic/work-packages/activities-tab/index.controller.ts
@@ -620,10 +620,12 @@ export default class IndexController extends Controller {
   quote(event:Event) {
     event.preventDefault();
     const target = event.currentTarget as HTMLElement;
+    const userId = target.dataset.userIdParam as string;
     const userName = target.dataset.userNameParam as string;
+    const textWrote = target.dataset.textWroteParam as string;
     const content = target.dataset.contentParam as string;
 
-    const quotedText = this.quotedText(content, userName);
+    const quotedText = this.quotedText(content, userId, userName, textWrote);
     const formVisible = !this.formRowTarget.classList.contains('d-none');
     if (formVisible) {
       this.insertQuoteOnExistingEditor(quotedText);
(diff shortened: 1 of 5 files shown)


frontend/src/stimulus/controllers/dynamic/work-packages/activities-tab/index.controller.ts
```diff
@@ -632,12 +634,13 @@ export default class IndexController extends Controller {
...
 
-  private quotedText(rawComment:string, userName:string) {
+  private quotedText(rawComment:string, userId:string, userName:string, textWrote:string) {
     const quoted = rawComment.split('\n')
       .map((line:string) => `\n> ${line}`)
       .join('');
 
-    return `${userName}\n${quoted}`;
+    // if we ever change CKEditor or how @mentions work this will break
+    return `<mention class="mention" data-id="${userId}" data-type="user" data-text="@${userName}">@${userName}</mention> ${textWrote}:\n\n${quoted}`;
```
akabiru, line 643: Ah, interesting. I'd imagine CKEditor has an internal builder class we can call? _*asking co-pilot, see if it can figure it out_ :)

akabiru (approved):
🥇

frontend/src/stimulus/controllers/dynamic/work-packages/activities-tab/index.controller.ts
brunopagno: I thought it was an internal thing, did not think it was CKEditor magic.

https://ckeditor.com/docs/ckeditor5/latest/api/module_mention_mentioncommand-MentionCommand.html

I'll try again a bit more.

brunopagno: Somehow I cannot get this one to work, so I might go with the current implementation, and if needed we refactor as a next step. Don't think it's worth investing much more time into this one.

akabiru: Makes sense, thanks for looking into it! Non-blocking IMO 💯

PR #17900, opened 2025-02-11
Storage Health Check: Adds check for token exchange capabilities

mereghost:
### Related Work Package:  [OP#61341](https://community.openproject.org/projects/cross-application-user-integration-stream/work_packages/61341)

Adds the capability check.

### DIFF
--- a/modules/storages/app/common/storages/peripherals/nextcloud_connection_validator.rb
+++ b/modules/storages/app/common/storages/peripherals/nextcloud_connection_validator.rb
@@ -57,7 +57,16 @@ module Storages
       def sso_misconfigured
         return None() unless @storage.authenticate_via_idp?
 
-        openid_connect_idp_absent
+        openid_connect_idp_absent.or { openid_connect_token_exchange }
+      end
+
+      def openid_connect_token_exchange
+        return None() if OpenIDConnect::Provider.any? { it.token_exchange_capable? }
+
+        Some(ConnectionValidation.new(type: :error,
+                                      error_code: :oidc_token_exchange_missing,
+                                      timestamp: Time.current,
+                                      description: I18n.t("storages.health.connection_validation.oidc_token_exchange_missing")))
       end
 
       def openid_connect_idp_absent
--- a/modules/storages/config/locales/en.yml
+++ b/modules/storages/config/locales/en.yml
@@ -224,6 +224,7 @@ en:
         missing_dependencies: 'A required dependency is missing on the file storage. Please add the following dependency: %{dependency}.'
         not_configured: The connection could not be validated. Please finish configuration first.
         oidc_provider_missing: There is no OpenID Connect Identity Provider configured.
+        oidc_token_exchange_missing: There is no OpenID Connect Identity Provider with the capability to do Token Exchange.
         placeholder: Check your connection against the server.
         subtitle: Connection validation
         tenant_id_wrong: The configured directory (tenant) id is invalid. Please check the configuration.
--- a/modules/storages/spec/common/storages/peripherals/nextcloud_connection_validator_spec.rb
+++ b/modules/storages/spec/common/storages/peripherals/nextcloud_connection_validator_spec.rb
@@ -231,7 +231,7 @@ RSpec.describe Storages::Peripherals::NextcloudConnectionValidator do
   end
 
   context "when OpenID Connect is enabled" do
-    let(:storage) { create(:nextcloud_storage_configured, :oidc_enabled) }
+    let(:storage) { create(:nextcloud_storage_configured, :oidc_sso_enabled) }
     let(:app_version) { Storages::SemanticVersion.parse("2.6.3") }
     let(:capabilities_response) do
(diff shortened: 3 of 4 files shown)


NobodysNightmare (changes requested):
The check provided is not enough. According to acceptance criteria the actual token of the admin should be tested for suitability (i.e. does it include the correct audience), if the token is suitable no error should be shown, but if the token is unsuitable an actual token exchange should be attempted to validate that this yields a token that would be usable.

In other words:
* the current test fails unnecessarily, if the IDP issues a token with the correct audience to begin with
* the current test succeeds incorrectly, if the IDP is assumed to be token exchange capable, but would later refuse to perform the desired token exchange

mereghost:
> The check provided is not enough. According to acceptance criteria the actual token of the admin should be tested for suitability

That's the next check. =)
Once we know we have a capable OIDC IdP I'll check if the audience is set on the storage (another step) and then validate the exchange (last step over OIDC checks).

NobodysNightmare:
But now the current check would also fail unnecessarily because it statically assumes that there must be an exchange-capable IDP. Is the plan to remove that check in an upcoming PR?

mereghost:
> But now the current check would also fail unnecessarily because it statically assumes that there must be an exchange-capable IDP. Is the plan to remove that check in an upcoming PR?

It will only run if an IdP provider is available and the storage is set for IdP usage.

NobodysNightmare:
But even then token exchange is not always necessary (╯°□°）╯︵ ┻━┻

If the storage is configured for SSO/IDP usage and the IDP issues a token to you that includes the correct audience already, then no token exchange is necessary.

That's why the AC are formulated the way they are. Essentially the first check is "is the token usable?" and only if not, further checks are necessary to see whether the token could be made usable.

mereghost:
I see. So this is a matter of moving this check to later in the chain. After the audience and existing token validation.

As the check and PR title implies this is testing just the IdP capabilities. We can hook it at any point in the checks. ;)

modules/storages/app/common/storages/peripherals/nextcloud_connection_validator.rb
```diff
@@ -57,7 +57,16 @@ def validate
       def sso_misconfigured
         return None() unless @storage.authenticate_via_idp?
 
-        openid_connect_idp_absent
+        openid_connect_idp_absent.or { openid_connect_token_exchange }
+      end
+
+      def openid_connect_token_exchange
+        return None() if OpenIDConnect::Provider.any? { it.token_exchange_capable? }
```
Kharonus, line 64: 🟡 I still get the chills, when looking at the assumption, that we have only a single SSO Provider. It falls back to checks like here, where we write `any?`.

Will it hide the falsy configurations too well? Will we get too many false positives - positive in the meaning, that the connection test is successful?

mereghost: If there is one that is capable and, as @NobodysNightmare noted above, it is necessary that should be the one we want, right?

mereghost: You could make a point that it is actually impossible to validate OIDC config due to:
- Multiple IdP available that may or may not require `Token Exchange`
- Non JWT tokens may lack audience
- Admin might be logged through the one that doesn't give access to Nextcloud
- Or worse, be a local account

NobodysNightmare: > Will it hide the falsy configurations too well? 

Not really, because the main check is supposed to be that you obtain a token that's usable.

We will eventually obtain an actual token and also try to **really** exchange it.

However, I'd still agree. When writing the AC I assumed that we would test the current user's IDP for exchange capability, which I thought would be clear from the context of testing an actual token and an actual token exchange, but now that this test was implemented first, the current implementation is a valid interpretation of that single line of AC.

NobodysNightmare: > You could make a point that it is actually impossible to validate OIDC config

I'd want to disagree on that notion. I think the main limitation with the check that's suggested in the feature is that we except the admin to be logged in with a suitable account and that configuration of the admin on the IDP's side is transferrable to other users.

Responding to @mereghost 's list:

* Multiple IdP available that may or may not require Token Exchange
    * that's "solved" by using the admin's SSO, this is a limitation in some way, but also allows the admin to log in through all available SSO's and perform the test with all SSO accounts, not awesome, but doable
* Non JWT tokens may lack audience
    * That would be a true negative, correct? I.e. if the token is opaque and we don't know that it has the correct audience, we will not use it if exchange is not possible; I don't see how validation would be different from reality here
* Admin might be logged through the one that doesn't give access to Nextcloud
    * correct, that's the main limitation
* Or worse, be a local account
    * yes, hopefully we have good error messages :)

But yeah, IMO that doesn't make validation _impossible_, but it adds _constraints_ to it. We should make sure to communicate those constraints to our admins.

mereghost:
I'll close this PR and open a new one following the results of these convos.

PR #17912, opened 2025-02-12
[#61178] Replace AuthCheckQuery with UserQuery.

ba1ash:
# Ticket

https://community.openproject.org/work_packages/61178

I try to get a way to find out user name on storage side by only having an access token. OpenProject has to know user name on Nexcloud side to construct proper WebDAV paths, at least, and in OIDC scenario at the moment there is no way to have it.

I try to make AuthCheckQuery more "general". It has name UserQuery now and responds with additional data on success(id of the user on storage side). The query has been modified for both Nextcloud and OneDrive storage modules. Next step will be to use this query after token exchange to update RemoteIdentity.

# Merge checklist

- [x] Added/updated tests


modules/storages/app/common/storages/peripherals/oauth_configurations/configuration_interface.rb
```diff
@@ -41,6 +43,21 @@ def to_httpx_oauth_config = raise ::Storages::Errors::SubclassResponsibility
...
+          Registry
+            .resolve("#{storage_name}.queries.user")
+            .call(auth_strategy:, storage: @storage)
+            .match(
+              on_success: ->(user) { user[:id] },
+              on_failure: ->(error) { raise "UserQuery responed with #{error}" }
+            )
+        end
+
+        def storage_name = raise ::Storages::Errors::SubclassResponsibility
```
NobodysNightmare: Defining this method might not be necessary. As far as I can tell, we always have a `@storage` when making these calls. At some point `Storage#to_s` was rewritten to return the provider name short form that can be used in registry resolve calls.

So what you should be able to do is `resolve("#{@storage}.your.stuff.here")`.

ba1ash: Yes, true. That's absolutely unnecessary. Thanks!

modules/storages/spec/common/storages/peripherals/storage_interaction/nextcloud/user_query_spec.rb
```diff
@@ -0,0 +1,74 @@
...
+#++
+
+require "spec_helper"
+require_module_spec_helper
+
+RSpec.describe Storages::Peripherals::StorageInteraction::Nextcloud::UserQuery, :webmock do
+  let(:user) { create(:user) }
+  # let(:storage) do
+  #   create(:nextcloud_storage_with_local_connection, :as_not_automatically_managed, oauth_client_token_user: user)
+  # end
```
NobodysNightmare: :yellow_circle: Before merging, we should remove outcommented code like this one (I guess a debugging left-over?).

NobodysNightmare (changes requested):
Small comments aside, this looks good. Nice to see that we can now treat Nextcloud and OneDrive so similar.

Pretty nice to imagine that the SSO case could effectively follow the same path, after further adjustments are done :)

app/models/oauth_client.rb
```diff
@@ -29,6 +31,7 @@
 class OAuthClient < ApplicationRecord
   belongs_to :integration, polymorphic: true
   has_many :remote_identities, dependent: :destroy
+  has_many :oauth_client_tokens, dependent: :destroy
```
mereghost, line 34: Good catch. DB already has a ON DELETE CASCADE set anyway. =)

modules/storages/app/common/storages/peripherals/oauth_configurations/nextcloud_configuration.rb
```diff
@@ -47,8 +47,8 @@ def initialize(storage)
 
         # rubocop:enable Lint/MissingSuper
 
-        def extract_origin_user_id(rack_access_token)
-          rack_access_token.raw_attributes[:user_id]
+        def storage_name
+          "nextcloud"
         end
```
mereghost: As per @NobodysNightmare comment, you can remote this. We made `Storage#to_s` to map to the `short_provider_name`. =)

modules/storages/app/common/storages/peripherals/oauth_configurations/one_drive_configuration.rb
```diff
@@ -51,11 +51,8 @@ def initialize(storage)
...
         # rubocop:enable Lint/MissingSuper
 
-        def extract_origin_user_id(rack_access_token)
-          OpenProject.httpx.get(
-            UrlBuilder.url(@storage.uri, "/v1.0/me"),
-            headers: { "Authorization" => "Bearer #{rack_access_token.access_token}", "Accept" => "application/json" }
-          ).raise_for_status.json["id"]
+        def storage_name
+          "one_drive"
         end
```
mereghost: Same as previous comment

mereghost (commented):
Looks good so far. Just a couple of comments. =)

app/models/oauth_client.rb
ba1ash: Yes, I saw on db level. I added the statement just for convenience.

Gemfile
```diff
@@ -235,7 +235,7 @@ gem "factory_bot_rails", "~> 6.4.4", require: false
 gem "turbo_power", "~> 0.7.0"
 gem "turbo-rails", "~> 2.0.0"
 
-gem "httpx"
+gem "httpx", "~> 1.3.4"
```
NobodysNightmare, line 240: IMO we should add a comment here, why we restrict the version and (probably much harder) we should try to create an upstream issue with reproduction instructions.

Did you get any idea how/why the newer version of HTTPX does that? Because it's working fine for most of our tests, so it surprises me that for one test it's showing this behaviour...

ba1ash: I have no idea why yet. But it seems to be the case for all failures when VCR or WebMock stub can not be found. Then there is an infinite hang happening. I'll add a clarification comment.

NobodysNightmare: The new comment is good, because it explains what to look out for to update.

In case you write/find an issue in their issue tracker, this would also be a great place to cross link it.

spec/services/remote_identities/create_service_spec.rb
```diff
@@ -4,23 +4,31 @@
...
+    create(:oauth_client_token,
+           user:,
+           oauth_client: oauth_config.oauth_client)
+  end
+
+  subject(:service) { described_class.new(user:, oauth_config:, oauth_client_token:) }
 
-  subject(:service) { described_class.new(user:, oauth_config:, oauth_token:) }
+  before do
+    stub_request(:get, File.join(storage.host, "/ocs/v1.php/cloud/user")).to_return(status: 200, body: File.read("modules/storages/spec/support/payloads/nextcloud_user_query_success.xml"), headers: {})
```
NobodysNightmare: Seeing that there are quite a bunch of line length offenses for essentially the same stubbed line. What do you think about offering a centralized helper to mock this?

E.g. `stub_nextcloud_user_query(storage.host)`

NobodysNightmare (approved):
Looks good to me. Approving already, but expecting that Rubocop is happy before merging ;-)

Gemfile
ba1ash: Yes, sure, That is my plan.

spec/services/remote_identities/create_service_spec.rb
ba1ash: Yes, good idea. Will do.

PR #17914, opened 2025-02-12
Storage Health Check: Adds check for storage audience

mereghost:
Related Work Package: [OP#61340](https://community.openproject.org/projects/cross-application-user-integration-stream/work_packages/61340)

Check to make sure audience is defined.


ba1ash (approved):
LGTM!

PR #17919, opened 2025-02-12
[#61306] Dialog text should use "Cancel occurrence"

myabc:
# Ticket

https://community.openproject.org/wp/61306

Updates the title, description and confirmation button text of `Meetings::DeleteDialogComponent` to use the term "cancel" rather than "delete" for removing occurrences.

**Danger Dialogs for cancelling open/instantiated occurrences and planned/scheduled occurrences should now be visually identical.**

## Screenshots

| Scenario                      | Before | After |
|-------------------------------|--------|--------|
| open/instantiated occurrences | Screenshot 2025-02-12 at 11 54 21 | Screenshot 2025-02-12 at 11 54 54 |
| planned/scheduled occurrences | Screenshot 2025-02-12 at 11 54 30 | Screenshot 2025-02-12 at 11 55 04 | 

I stuck with the 1-to-1 controller action to dialog approach introduced in #17569. It is not the DRY-est approach, but it reduces complexity. It might be possible to make `Meetings::DeleteDialogComponent` polymorphic in the future - thereby allowing us to drop `RecurringMeetings::ScheduledDeleteDialogComponent`.

Additionally, I've kept the I18n string keys separate for now. Although we are making the UI for cancelling open/instantiated occurrences and planned/scheduled occurrences visually identical for moment, I could envisage us making adjustments to the copy in the future.

# Merge checklist

- [X] Added/updated tests

### DIFF
--- a/modules/meeting/app/components/meetings/delete_dialog_component.html.erb
+++ b/modules/meeting/app/components/meetings/delete_dialog_component.html.erb
@@ -31,6 +31,8 @@ See COPYRIGHT and LICENSE files for more details.
   render(Primer::OpenProject::DangerDialog.new(
     id:,
     title:,
+    confirm_button_text:,
+    cancel_button_text:,
     form_arguments: {
       action: polymorphic_path([@project, @meeting.becomes(Meeting)]),
       method: :delete,
--- a/modules/meeting/app/components/meetings/delete_dialog_component.rb
+++ b/modules/meeting/app/components/meetings/delete_dialog_component.rb
@@ -64,9 +64,25 @@ module Meetings
 
     def confirmation_message
       if recurring_meeting.present?
-        t("meeting.delete_dialog.occurrence.confirmation_message_html", title: recurring_meeting.title)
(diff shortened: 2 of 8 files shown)


myabc:
> Additionally, I've kept the I18n string keys separate for now. Although we are making the UI for cancelling open/instantiated occurrences and planned/scheduled occurrences visually identical for moment, I could envisage us making adjustments to the copy in the future.

I might be wrong here - let me know what you think.

mrmir:
I just took a quick look here, but FYI we don't need to update strings for all locales I don't think. Only the English ones are necessary as the rest get overwritten by the crowdin action anyway

myabc:
> I just took a quick look here, but FYI we don't need to update strings for all locales I don't think. Only the English ones are necessary as the rest get overwritten by the crowdin action anyway

I wasn't sure if Crowdin synced both ways. I'll update the PR to only touch `en.yml`.

oliverguenther (approved):

PR #17920, opened 2025-02-12
[62708] Implementing ARIA live regions to communicate contextual changes

bsatarnejad:
# Ticket
https://community.openproject.org/wp/62708

Implement a generic method for ARIA live regions to communicate contextual changes
Add an example of its usage in date picker

- [ ] Added/updated tests


app/components/work_packages/date_picker/dialog_content_component.html.erb
```diff
@@ -98,6 +102,8 @@
           end
         end
       end
-    end
-  end
-%>
+      end%>
```
HDinger: Wrong indentation

frontend/src/stimulus/controllers/dynamic/work-packages/date-picker/preview.controller.ts
```diff
@@ -364,7 +364,6 @@ export default class PreviewController extends DialogPreviewController {
     }
   }
 
-  // called from inputs defined in the date_picker/date_form.rb
```
HDinger, line 356: Why did you remove that comment?

app/controllers/work_package_children_relations_controller.rb
```diff
@@ -101,7 +101,11 @@ def respond_with_relations_tab_update(service_result, **)
       component = WorkPackageRelationsTab::IndexComponent.new(work_package: @work_package, **)
       replace_via_turbo_stream(component:)
       render_success_flash_message_via_turbo_stream(message: I18n.t(:notice_successful_update))
-
+      render_aria_update_message(
+        message: I18n.t("work_package_relations_tab.relations.create_child_aria_live_message"),
+        type: "polite",
+        role: "alert"
```
HDinger: `role: "alert"` is the default, so this line can be ommited.

app/controllers/work_package_relations_controller.rb
```diff
@@ -103,7 +103,11 @@ def respond_with_relations_tab_update(service_result, **)
       component = WorkPackageRelationsTab::IndexComponent.new(work_package: @work_package, **)
       replace_via_turbo_stream(component:)
       render_success_flash_message_via_turbo_stream(message: I18n.t(:notice_successful_update))
-
+      render_aria_update_message(
+        message: I18n.t("work_package_relations_tab.relations.create_relation_aria_live_message"),
+        type: "polite",
+        role: "alert"
```
HDinger: Same here, can be omitted

lookbook/docs/patterns/18-aria-live.md.erb
```diff
@@ -0,0 +1,93 @@
...
+- `aria-live="assertive"`: Announces updates immediately, interrupting the user.
+- `aria-atomic="true"`: Ensures the entire content of the region is read when it changes.
+
+## Implementing ARIA Live Regions in Primer
+Primer provides a built-in component named live-region to implement ARIA Live Regions easily.
+We have decided to use a single global ARIA live region in base.html to ensure live regions are present in the DOM before updates occur. This approach is better than having multiple live regions because it ensures that announcements are delivered in a clear and consistent manner. By centralizing updates, we avoid potential conflicts between overlapping live regions and provide a more seamless experience for screen reader users.
+There is an exception for modal or dialog, especially if that dialog traps focus. When a dialog is open, screen readers often limit their reading and focus to within the dialog. If aria-live region is in base.html (outside the dialog), it becomes “invisible” to the screen reader while the modal is open — it's still updated in the DOM, but not announced until the dialog is closed, and the screen reader returns to the full DOM scope.
```
HDinger:
```suggestion
There is an exception for modals or dialogs, especially if that dialog traps focus. When a dialog is open, screen readers often limit their reading and focus to the content of the dialog. Aria-live regions that are part of the base.html and thus outside of the dialog become “invisible” to the screen reader as long as the modal is open.
```

lookbook/docs/patterns/18-aria-live.md.erb
```diff
@@ -0,0 +1,93 @@
...
+The `@primer/live-region-element` package provides methods to trigger announcements programmatically:
+
+```ts
+import { announce } from '@primer/live-region-element';
+
+announce('Example polite message', { politeness: 'polite' });
+```
+
+### Using Turbo Streams to Trigger ARIA Live Updates
+To use it in a Rails Turbo response, and send updates to the screen reader after any action, we create a new Turbo stream component that triggers an ARIA action:
```
HDinger:
```suggestion
To use it in a Rails Turbo response, and send updates to the screen reader after any action, we created a new Turbo stream component that triggers an ARIA action:
```

lookbook/docs/patterns/18-aria-live.md.erb
```diff
@@ -0,0 +1,93 @@
...
+  - `type: "polite"` → Ensures the message is read when the user is idle
+  - `role: "status"` → Helps screen readers recognize this as a status update
+  - `target: "nil"` → Ensures the update isn't inserted into the DOM but is handled as an announcement
+
+#### How to use it?
+In controller:
+```ruby
+    def render_aria_update_message(message:, type:, role: "alert")
+      turbo_streams << OpTurbo::StreamComponent.new(action: :aria, message:, type:, role:, target: "nil").render_in(view_context)
+    end
```
HDinger: I'd split that example into two, as it currently suggests, that people have to add that method themselves. However, they should just use the method we already provided.

lookbook/docs/patterns/18-aria-live.md.erb
```diff
@@ -0,0 +1,93 @@
+## What are ARIA Live Regions?
```
HDinger: Please create a folder called "Accessibility" and move that file in there.

lookbook/docs/patterns/18-aria-live.md.erb
HDinger: Please add an empty line between the paragraphs, otherwise it is really hard to read.

lookbook/docs/patterns/18-aria-live.md.erb
```diff
@@ -0,0 +1,93 @@
...
+      if (type === 'assertive') {
+        void announce(message, {
+          politeness: 'assertive',
+        });
+      } else {
+        void announce(message, {
+          politeness: 'polite',
+          delayMs: 5000,
+        });
+      }
```
HDinger: I really don't understand what this paragraph and the method is about or what it is used for. The headline says "Turbo Stream Action for ARIA Live" but the previous headline is also for Turbo stream actions with headlines..

lookbook/docs/patterns/18-aria-live.md.erb
```diff
@@ -0,0 +1,93 @@
...
+      turbo_streams << OpTurbo::StreamComponent.new(action: :aria, message:, type:, role:, target: "nil").render_in(view_context)
+    end
+
+    def update
+      render_success_flash_message_via_turbo_stream(message: I18n.t(:notice_successful_update))
+      render_aria_update_message(message: I18n.t("work_package_relations_tab.relations.create_child_aria_live_message"),type: "polite",role: "alert")
+      respond_with_turbo_streams
+    end
+```
+When updating via Turbo Frame:
```
HDinger: Missing newline

frontend/src/turbo/aria-stream-action.ts
```diff
@@ -0,0 +1,19 @@
+import { StreamActions, StreamElement } from '@hotwired/turbo';
+import { announce } from '@primer/live-region-element';
+
+export function registerAriaStreamAction() {
+  StreamActions.aria = function dialogStreamAction(this:StreamElement) {
```
HDinger: What is that needed for?

app/controllers/concerns/op_turbo/component_stream.rb
```diff
@@ -83,6 +83,10 @@ def render_error_flash_message_via_turbo_stream(**)
       render_flash_message_via_turbo_stream(**, scheme: :danger, icon: :stop)
     end
 
+    def render_aria_update_message(message:, type:, role: "alert")
+      turbo_streams << OpTurbo::StreamComponent.new(action: :aria, message:, type:, role:, target: "nil").render_in(view_context)
```
HDinger: The target is weird. Should that really be a string "nil" or rather the object `nil`?

lookbook/docs/patterns/18-aria-live.md.erb
```diff
@@ -0,0 +1,93 @@
...
+
+## Implementing ARIA Live Regions in Primer
+Primer provides a built-in component named live-region to implement ARIA Live Regions easily.
+We have decided to use a single global ARIA live region in base.html to ensure live regions are present in the DOM before updates occur. This approach is better than having multiple live regions because it ensures that announcements are delivered in a clear and consistent manner. By centralizing updates, we avoid potential conflicts between overlapping live regions and provide a more seamless experience for screen reader users.
+There is an exception for modal or dialog, especially if that dialog traps focus. When a dialog is open, screen readers often limit their reading and focus to within the dialog. If aria-live region is in base.html (outside the dialog), it becomes “invisible” to the screen reader while the modal is open — it's still updated in the DOM, but not announced until the dialog is closed, and the screen reader returns to the full DOM scope.
+In this case, we should add the live-region component to the dialog itself, like what we have implemented for primerized date picker.
```
HDinger: Maybe you should make this a separate section with its own heading (e.g. "Live regions in Dialogs") and push it a bit down, maybe even at the end. It feels very confusing and I am missing an example.

app/components/work_packages/date_picker/dialog_content_component.html.erb
```diff
@@ -98,6 +102,8 @@
...
+  <% if @live_region_message %>
+    <%= render OpTurbo::StreamComponent.new(action: :aria, message: @live_region_message, type:"assertive", target: "nil") %>
+  <% end %>
```
HDinger: Could that be moved to the top below the actual `live-region`?

HDinger (changes requested):

frontend/src/stimulus/controllers/dynamic/work-packages/date-picker/preview.controller.ts
bsatarnejad: This file doesn't exist anymore. I will change it to date_picker/date_form_component.rb 👍

frontend/src/turbo/aria-stream-action.ts
bsatarnejad: I want to define aria action for turbo, I must explicitly register it by assigning a function to StreamActions.aria. Therefore, when a stream with action="aria" arrives, run this custom function I’m registering as StreamActions.aria.

lookbook/docs/patterns/accessibility/18-aria-live.md.erb
```diff
@@ -0,0 +1,117 @@
+## What are ARIA Live Regions?
+ARIA Live Regions are a crucial accessibility feature that allows screen readers to announce dynamic content updates without requiring user focus changes.
```
myabc:
```suggestion
ARIA Live Regions are a crucial accessibility feature that allow screen readers to announce dynamic content updates without requiring user focus changes.
```

frontend/src/turbo/aria-stream-action.ts
```diff
@@ -0,0 +1,19 @@
+import { StreamActions, StreamElement } from '@hotwired/turbo';
+import { announce } from '@primer/live-region-element';
+
+export function registerAriaStreamAction() {
+  StreamActions.aria = function ariaStreamAction(this:StreamElement) {
+      const message = this.getAttribute('message') ?? '';
+      const type = this.getAttribute('type') ?? 'polite';
```
myabc: Logical OR here is probably better here - so that we also handle blank attribute values, e.g. `type=""`.

```suggestion
      const message = this.getAttribute('message') || '';
      const type = this.getAttribute('type') || 'polite';
```

frontend/src/turbo/aria-stream-action.ts
```diff
@@ -0,0 +1,19 @@
...
+      if (type === 'assertive') {
+        void announce(message, {
+          politeness: 'assertive',
+        });
+      } else {
+        void announce(message, {
+          politeness: 'polite',
+          delayMs: 5000,
```
myabc: Magic number. Where does this value come from? - it'd be good to replace with a constant, and ideally also document why 5 seconds (as opposed to the [default 150ms](https://github.com/primer/live-region-element/blob/main/packages/live-region-element/src/live-region-element.ts#L49C6-L49C38)) is an appropriate value.

💡 If 5 seconds is an appropriate value for the Date Picker, but not elsewhere, we could also make this configurable - as an attribute on the Turbo Stream element.

lookbook/docs/patterns/accessibility/18-aria-live.md.erb
```diff
@@ -0,0 +1,117 @@
...
+
+### ARIA Live Attributes
+ARIA Live Regions rely on specific attributes to control how updates are announced:
+- `aria-live="polite"`: Announces updates when the user is idle.
+- `aria-live="assertive"`: Announces updates immediately, interrupting the user.
+- `aria-atomic="true"`: Ensures the entire content of the region is read when it changes.
+
+## Implementing ARIA Live Regions in Primer
+Primer provides a built-in component named live-region to implement ARIA Live Regions easily.
```
myabc: "built-in" a little ambiguous - it's not a built-in View Component, but rather another library from the Primer project.

```suggestion
The Primer project provides a custom element (Web Component) to ease working with ARIA Live Regions.
```

----

For separate discussion with @camertron, @HDinger and team:

We should differentiate between the "Primer project" (i.e. what's online at https://github.com/primer), PVC upstream and OpenProject's fork of PVC..

lookbook/docs/patterns/accessibility/18-aria-live.md.erb
```diff
@@ -0,0 +1,117 @@
...
+We have decided to use a single global ARIA live region in base.html to ensure live regions are present in the DOM before updates occur. This approach is better than having multiple live regions because it ensures that announcements are delivered in a clear and consistent manner. By centralizing updates, we avoid potential conflicts between overlapping live regions and provide a more seamless experience for screen reader users.
```
myabc:
```suggestion
We have decided to use a single global ARIA live region in `base.html` to ensure live regions are present in the DOM before updates occur. This approach is better than having multiple live regions because it ensures that announcements are delivered in a clear and consistent manner. By centralizing updates, we avoid potential conflicts between overlapping live regions and provide a more seamless experience for screen reader users.
```

app/components/work_packages/date_picker/dialog_content_component.html.erb
```diff
@@ -1,5 +1,13 @@
 <%=
   content_tag("turbo-frame", id: "wp-datepicker-dialog--content") do
+%>
+  <live-region>
+    <div id="assertive" aria-live="assertive" aria-atomic="true" class="hidden-for-sighted"></div>
+  </live-region>
+  <% if @live_region_message %>
+    <%= render OpTurbo::StreamComponent.new(action: :aria, message: @live_region_message, type: "assertive", target: "nil") %>
```
myabc: Is `target: "nil"` correct here?

app/controllers/work_packages/date_picker_controller.rb
```diff
@@ -157,7 +157,29 @@ def datepicker_modal_component
                                                          focused_field:,
                                                          triggering_field: params[:triggering_field],
                                                          touched_field_map:,
-                                                         date_mode:)
+                                                         date_mode:,
+                                                         live_region_message: live_region_message)
```
myabc: Nitpick. You can use shorthand Hash syntax here:

```suggestion
                                                         live_region_message:)
```

app/controllers/work_packages/date_picker_controller.rb
```diff
@@ -157,7 +157,29 @@ def datepicker_modal_component
...
+  end
+
+  def live_region_message
+    I18n.t(
+      "work_packages.datepicker_modal.update_inputs_aria_live_message",
+      start_date: @work_package.start_date,
+      due_date: @work_package.due_date,
+      duration: @work_package.duration,
+      working_days_only: working_days_label,
+      scheduling: scheduling_label
```
myabc: Nitpick (i.e.don't feel obliged to change) - if you name the keys and values the same, you can take advantage of shorthand Hash syntax here too. e.g.

```suggestion
      working_days_label:,
      scheduling_label:
```

app/controllers/work_packages/date_picker_controller.rb
```diff
@@ -157,7 +157,29 @@ def datepicker_modal_component
...
+    )
+  end
+
+  def working_days_label
+    key = @work_package.ignore_non_working_days ? "true" : "false"
+    I18n.t("activerecord.attributes.work_package.include_non_working_days.#{key}")
```
myabc: You can skip the intermediate variable and just use interpolation:

```suggestion
    I18n.t("activerecord.attributes.work_package.include_non_working_days.#{@work_package.ignore_non_working_days}")
```

frontend/src/turbo/setup.ts
```diff
@@ -30,6 +31,7 @@ addTurboEventListeners();
 addTurboGlobalListeners();
 registerDialogStreamAction();
 registerFlashStreamAction();
+registerAriaStreamAction();
```
myabc: I'm not convinced by the naming - ARIA Live Stream? (or perhaps that sounds too much like Twitch 😄 )

frontend/src/turbo/aria-stream-action.ts
myabc: Perhaps we could also call this "politeness" rather than the much more generic "type"?

lookbook/docs/patterns/accessibility/18-aria-live.md.erb
```diff
@@ -0,0 +1,117 @@
...
+### Using Turbo Streams to Trigger ARIA Live Updates
+To use it in a Rails Turbo response, and send updates to the screen reader after any action, we created a new Turbo stream component that triggers an ARIA action:
+
+```ruby
+OpTurbo::StreamComponent.new(action: :aria, message: "Form submission successful!", type: "polite", role: "status", target: nil)
+```
+
+- This creates a **Turbo Stream update** with:
+  - `action: :aria` → Calls the `aria` action defined in `aria.stream.action.ts`
+  - `message: "Form submission successful!"` → The text announced by the screen reader
```
myabc:
```suggestion
  - `message: "Form submission successful!"` → The text to be announced to user
```

lookbook/docs/patterns/accessibility/18-aria-live.md.erb
```diff
@@ -0,0 +1,117 @@
...
+  - `type: "polite"` → Ensures the message is read when the user is idle
+  - `role: "status"` → Helps screen readers recognize this as a status update
```
myabc: AFAICT we are not actually doing anything with `role`?

lookbook/docs/patterns/accessibility/18-aria-live.md.erb
```diff
@@ -0,0 +1,117 @@
...
+          politeness: 'polite',
+          delayMs: 5000,
+        });
+      }
+  };
+}
+```
+
+## Live Regions in Dialogs and Modals
+There is an important exception to our use of the global ARIA live region in base.html.
```
myabc:
```suggestion
There is an important exception to our use of the global ARIA Live Region in `base.html`.
```

lookbook/docs/patterns/accessibility/18-aria-live.md.erb
```diff
@@ -0,0 +1,117 @@
...
+
+When a modal or dialog is open, screen readers often restrict focus and reading context to the content inside the dialog. In this case, ARIA live regions outside the dialog (such as those defined in base.html) become inaccessible to the screen reader.
```
myabc:
```suggestion
When a modal or dialog is open, screen readers often restrict focus and reading context to the content inside the dialog. In this case, live regions outside the dialog (such as those defined in `base.html`) become inaccessible to the screen reader.
```

lookbook/docs/patterns/accessibility/18-aria-live.md.erb
```diff
@@ -0,0 +1,117 @@
...
+
+To ensure that announcements are still read aloud in this situation, we should include a live-region component within the dialog itself, like what we have implemented for primerized date picker.
+
+```erb
+  <live-region>
+    <div id="assertive" aria-live="assertive" aria-atomic="true" class="hidden-for-sighted"></div>
+  </live-region>
+  <%= render OpTurbo::StreamComponent.new(action: :aria, message: "Date picker is updated.", type: "assertive", target: "nil") %>
```
myabc: ~~I don't understand why we suggest rendering the Stream Element inline like this?~~

lookbook/docs/patterns/accessibility/18-aria-live.md.erb
```diff
@@ -0,0 +1,117 @@
...
+```ruby
+OpTurbo::StreamComponent.new(action: :aria, message: "Form submission successful!", type: "polite", role: "status", target: nil)
+```
+
+- This creates a **Turbo Stream update** with:
+  - `action: :aria` → Calls the `aria` action defined in `aria.stream.action.ts`
+  - `message: "Form submission successful!"` → The text announced by the screen reader
+  - `type: "polite"` → Ensures the message is read when the user is idle
+  - `role: "status"` → Helps screen readers recognize this as a status update
+  - `target: "nil"` → Ensures the update isn't inserted into the DOM but is handled as an announcement
```
myabc:
```suggestion
  - `target: nil` → Ensures the update isn't inserted into the DOM but is handled as an announcement
```

app/controllers/concerns/op_turbo/component_stream.rb
```diff
@@ -83,6 +83,10 @@ def render_error_flash_message_via_turbo_stream(**)
       render_flash_message_via_turbo_stream(**, scheme: :danger, icon: :stop)
     end
 
+    def render_aria_update_message(message:, type:, role: "alert")
```
myabc: It would be good to have a sensible default for `type`[^1] here, i.e.

```suggestion
    def render_aria_update_message(message:, type: :polite, role: :alert)
```

Also symbols are slightly more idiomatic Ruby[^2].

[^1]: see other comment about naming - `type` vs `politeness`.
[^2]: see other comment about `role` not being used the custom Turbo Stream Action.

lookbook/docs/patterns/accessibility/18-aria-live.md.erb
```diff
@@ -0,0 +1,117 @@
...
+
+### How to Trigger an ARIA Update from the Controller
+You can trigger a screen reader announcement from your Rails controller by calling the `render_aria_update_message` helper, which is already provided:
```
myabc:
```suggestion
You can trigger a screen reader announcement from your Rails controller by calling the provided `render_aria_update_message` helper:
```

app/controllers/concerns/op_turbo/component_stream.rb
```diff
@@ -83,6 +83,10 @@ def render_error_flash_message_via_turbo_stream(**)
...
+      turbo_streams << OpTurbo::StreamComponent.new(action: :aria, message:, type:, role:, target: nil).render_in(view_context)
```
myabc: nitpick - formatting

```suggestion
      turbo_streams << OpTurbo::StreamComponent
        .new(action: :aria, message:, type:, role:, target: nil)
        .render_in(view_context)
```

lookbook/docs/patterns/accessibility/18-aria-live.md.erb
```diff
@@ -0,0 +1,117 @@
...
+
+This helper internally renders a Turbo Stream using the aria action to announce the message via ARIA live regions.
+For reference only — you do not need to define this yourself.
+Here is the internal implementation of `render_aria_update_message`:
+
+```ruby
+    def render_aria_update_message(message:, type:, role: "alert")
+      turbo_streams << OpTurbo::StreamComponent.new(action: :aria, message:, type:, role:, target: nil).render_in(view_context)
+    end
+```
```
myabc: I'm not sure you need to go into this level of detail about the implementation.

What is important is that calling controllers also call `respond_to_with_turbo_streams`/`respond_with_turbo_streams` (or the equivalent `respond_to` + block).

app/components/work_packages/date_picker/dialog_content_component.html.erb
```diff
@@ -1,5 +1,13 @@
 <%=
   content_tag("turbo-frame", id: "wp-datepicker-dialog--content") do
+%>
+  <live-region>
+    <div id="assertive" aria-live="assertive" aria-atomic="true" class="hidden-for-sighted"></div>
```
myabc: 1. we should use a different `id` here - we are already declaring a `div` with this `id` in `base.html.erb`.
2. Can we start to use Primer's `sr-only` rather than `hidden-for-sighted`? / @HDinger  (se. #18837)

lookbook/docs/patterns/accessibility/18-aria-live.md.erb
myabc: Ok, I think I understand now - it hadn't occurred to me that you can render `<turbo-stream>` as part of regular HTML response and that those actions would still be processed (this may come in handy in #18716, as we might want to use Turbo Frames with flash messages).

config/locales/en.yml
```diff
@@ -744,6 +744,7 @@ en:
         automatic: "Automatic"
         manual: "Manual"
       show_relations: "Show relations"
+      update_inputs_aria_live_message: "Date picker updated. Scheduling mode: %{scheduling}, %{working_days_only}, Start date: %{start_date}, Finish date: %{due_date}, Duration: %{duration} days."
```
myabc: This doesn't work well when fields (e.g start date, duration) are empty. We might want to construct this message string in Ruby.

Screenshot 2025-05-06 at 22 59 58

app/components/work_packages/date_picker/dialog_content_component.html.erb
```diff
@@ -1,5 +1,13 @@
...
+  </live-region>
+  <% if @live_region_message %>
+    <%= render OpTurbo::StreamComponent.new(action: :aria, message: @live_region_message, type: "assertive", target: "nil") %>
+  <% end %>
```
myabc: Is there a good reason (apart from the programmatic delay) that we can't forgo the `live-region` here and just update the live region directly (with Turbo Frames + morphing)?

```suggestion
<div id="assertive" aria-live="assertive" aria-atomic="true" class="hidden-for-sighted">
  <%= @live_region_message %>
</div>
```

It would be one fewer layer of indirection.

(I realise we might have discussed this before, so apologies if I'm forgetting something)

myabc (changes requested):
Hi @bsatarnejad. I'm really sorry that it's taken so long for me to give feedback on this PR.

I have only tested on macOS with VoiceOver so far - but the good news is that, for the most part, everything works as expected, and that this is a nice enhancement for users of assistive technology.

There are various things I think we should address though:

- **assertiveness**: while the "assertive" behaviour with the date picker seems to work well, the "polite" updates on adding a relation don't work so well, with VoiceOver announcing the change long after the flash message disappears.

image

https://github.com/user-attachments/assets/1f237bd0-cc0b-41dd-8c60-96a7d291e53a

(this might beg the question as to whether we should look into integrating flash messages + live regions)

- **commit hygiene**: this PR now consists of 54 commits, despite not being huge in terms of changes (`16 files changed, 240 insertions(+), 11 deletions(-)`). I realise that this is because it is a long-running piece of work that has suffered from requirement churn and changes in prioritisation along the way. That said, although OP policy AFAICT is to try to maintain commit history, I think this PR would benefit from some cleanup (namely rebasing + squashing of related commits). I'm happy to help with this.

- **documentation**: this is generally fine, but I think we should prefer the more generic term "asssistive technology" (and "users of A.T.") to "screen readers" where possible.

  - we should try to be consistent with capitalisation: "ARIA Live Regions" vs. "live regions".
  - With Turbo Streams, I think we should encourage the use of the `render_aria_update_message` helper over the component directly.

- **naming**: We should call the custom Turbo Stream Action something more specific than "ARIA". Perhaps `LiveRegionStreamAction`?

- **design**: Should we introduce a specialization of the `OpTurbo::StreamComponent` that enforces correct arguments - rather than passing necessary args for the `aria` action as `**turbo_stream_args`?

- **tests**: there are currently no tests for these changes. We should discuss how best to test (both with Capy features and specs for components/helpers).

If anything is unclear, I'd be happy to explain! - in any case, it's probably worth us setting aside some time to discuss in more detail. Cheers!

lookbook/docs/patterns/accessibility/18-aria-live.md.erb
```diff
@@ -0,0 +1,117 @@
...
+        });
+      } else {
+        void announce(message, {
+          politeness: 'polite',
+          delayMs: 5000,
+        });
+      }
+  };
+}
+```
```
myabc, line 103: I think you should be able to do the following:

```ts
<%= render file: "frontend/src/turbo/aria-stream-action.ts" %>
```

lookbook/docs/patterns/accessibility/18-aria-live.md.erb
```diff
@@ -0,0 +1,117 @@
...
+          politeness: 'assertive',
+        });
+      } else {
+        void announce(message, {
+          politeness: 'polite',
+          delayMs: 5000,
+        });
+      }
+  };
+}
```
myabc: I think you should be able to do the following:

```suggestion
<%= render file: "frontend/src/turbo/aria-stream-action.ts" %>
```

app/components/work_packages/date_picker/dialog_content_component.html.erb
myabc: I think `<live-region></live-region>` should suffice. `LiveRegionElement` [defines its own template](https://github.com/primer/live-region-element/blob/main/packages/live-region-element/src/live-region-element.ts#L259-L276).

myabc:
it would be great if we could look into how this fits in with the [Declarative Shadow DOM](https://github.com/primer/live-region-element?tab=readme-ov-file#declarative-shadow-dom) (my Shadow DOM knowledges is shaky at best - @HDinger and @camertron might be able to shed some light!)

camertron:
Hey @myabc, thanks for tagging me 😄

> it would be great if we could look into how this fits in with the [Declarative Shadow DOM](https://github.com/primer/live-region-element?tab=readme-ov-file#declarative-shadow-dom) (my Shadow DOM knowledges is shaky at best)

I believe the biggest reason to use a shadow DOM with the `<live-region>` element is to prevent global CSS styles from accidentally making the element's contents visible to sighted users (live regions should typically only be visible to screen readers, which `<live-region>` element achieves with CSS). Other than that, the shadow DOM shouldn't affect any of the actual announcement behavior offered by the element.

I did also want to pop in here and mention that, generally speaking, there should only be one ARIA live region on the page at a time. Having two or more areas [can be flagged as errors or warnings](https://github.com/w3c/aria/issues/1689#issuecomment-1026892454) by automated accessibility checkers, and is [discouraged](https://github.com/w3c/aria/issues/1689#issuecomment-1026892454) by live-region-element itself:

> Having multiple live regions on a page is discouraged so we recommend having a single global live region that is available across every page of your application by embedding this live-region element as part of your page layout.

I think this is mostly due to various screen reader quirks and bugs rather than an inherent limitation of live regions, but it's worth considering to ensure an optimal user experience. In this PR it appears a new `<live-region>` will be constructed for each date picker instance. If I recall correctly, live-region-element's `announce` function (and its friends) will automatically find any `<live-region>` element that exist in the DOM, so adding one to the application's layout should work well.

I would suggest merging this PR first and consolidating live regions in a separate PR 😄

bsatarnejad:
> Hi @bsatarnejad. I'm really sorry that it's taken so long for me to give feedback on this PR.
> 
> I have only tested on macOS with VoiceOver so far - but the good news is that, for the most part, everything works as expected, and that this is a nice enhancement for users of assistive technology.
> 
> There are various things I think we should address though:
> 
> * **assertiveness**: while the "assertive" behaviour with the date picker seems to work well, the "polite" updates on adding a relation don't work so well, with VoiceOver announcing the change long after the flash message disappears.
> 
> image
>  Screen.Recording.2025-05-07.at.01.18.44.mov 
> (this might beg the question as to whether we should look into integrating flash messages + live regions)
> 
> * **commit hygiene**: this PR now consists of 54 commits, despite not being huge in terms of changes (`16 files changed, 240 insertions(+), 11 deletions(-)`). I realise that this is because it is a long-running piece of work that has suffered from requirement churn and changes in prioritisation along the way. That said, although OP policy AFAICT is to try to maintain commit history, I think this PR would benefit from some cleanup (namely rebasing + squashing of related commits). I'm happy to help with this.
> * **documentation**: this is generally fine, but I think we should prefer the more generic term "asssistive technology" (and "users of A.T.") to "screen readers" where possible.
>   
>   * we should try to be consistent with capitalisation: "ARIA Live Regions" vs. "live regions".
>   * With Turbo Streams, I think we should encourage the use of the `render_aria_update_message` helper over the component directly.
> * **naming**: We should call the custom Turbo Stream Action something more specific than "ARIA". Perhaps `LiveRegionStreamAction`?
> * **design**: Should we introduce a specialization of the `OpTurbo::StreamComponent` that enforces correct arguments - rather than passing necessary args for the `aria` action as `**turbo_stream_args`?
> * **tests**: there are currently no tests for these changes. We should discuss how best to test (both with Capy features and specs for components/helpers).
> 
> If anything is unclear, I'd be happy to explain! - in any case, it's probably worth us setting aside some time to discuss in more detail. Cheers!

Hi @myabc , Thanks for the feedback. The issue you mentioned for relation update is right, but it is an example of the aria-live usage. We can remove it, because I think we should give this kind of update in flash messages.

lookbook/docs/patterns/accessibility/18-aria-live.md.erb
bsatarnejad: Yes, we don't use it, its default value is alert, it can be removed.

app/components/work_packages/date_picker/dialog_content_component.html.erb
bsatarnejad: As we discussed, we need it for adding some delay.

frontend/src/turbo/live-region-stream-action.ts
```diff
@@ -0,0 +1,21 @@
+import { StreamActions, StreamElement } from '@hotwired/turbo';
+import { announce } from '@primer/live-region-element';
+
+export function registerLiveRegionStreamAction() {
+  StreamActions.liveRegion = function liveRegionStreamAction(this:StreamElement) {
+    const message = this.getAttribute('message') || '';
+    const politeness = this.getAttribute('politeness') || 'polite';
+    const delayAttr = this.getAttribute('delay');
+    const delay = delayAttr !== null && !Number.isNaN(Number(delayAttr)) ? Number(delayAttr) : 0;
```
myabc: You can simplify this as follows:

```suggestion
    const delay = parseInt(this.getAttribute('delay'), 10) || 0;
```

(`parseInt(null) => NaN`, which is `falsey`)

*updated to specify a radix)

frontend/src/turbo/live-region-stream-action.ts
```diff
@@ -0,0 +1,21 @@
+import { StreamActions, StreamElement } from '@hotwired/turbo';
+import { announce } from '@primer/live-region-element';
+
+export function registerLiveRegionStreamAction() {
+  StreamActions.liveRegion = function liveRegionStreamAction(this:StreamElement) {
+    const message = this.getAttribute('message') || '';
```
myabc: Is there any reason we would want to pass an empty message to `announce`?

We could just return early if `message` is falsey.

bsatarnejad:
I'm merging the PR, since the failing test doesn't happen locally.

PR #17921, opened 2025-02-12
fix double submit after refresh

brunopagno:
# Ticket
https://community.openproject.org/projects/communicator-stream/work_packages/60719

Fix a behaviour where a comment would get submitted twice if, after a page refresh, the user clicked the submit button multiple times quickly

First thing we do in the form is prevent it to do it's default behaviour. This way if anything triggers the "submit" method, while it's already doing something else it will not trigger a double submit.


akabiru (approved):
Nice one 👍🏾

PR #17922, opened 2025-02-12
[#57799] Agenda item outcomes

mrmir:
# Ticket

https://community.openproject.org/work_packages/57799

- Add in meeting outcomes
- Add in a new 'in progress' meeting status that allows editing outcomes

- I used a base component to switch between outcome components depending on the value of certain flags

# Merge checklist

- [x] Added/updated tests


frontend/src/stimulus/controllers/dynamic/meeting-outcome-form.controller.ts
```diff
@@ -0,0 +1,41 @@
...
+
+import { Controller } from '@hotwired/stimulus';
+
+export default class extends Controller {
+  connect():void {
+    this.focusInput();
+  }
+
+  focusInput():void {
+    this.element.scrollIntoView({ block: 'center' });
```
oliverguenther, line 39: I feel like this should be a generalized controller, with a value to define the scrolling behavior

oliverguenther:
*   [x] There is a new meeting state in between the existing _Open_ and _Closed_
    *   [x] This new state is &#39;In progress&#39;
    *   [x] When a meeting is _Open,_ the default action in the right-hand pane wiill be &#39;Mark as in progress&#39;, which will change state accordingly
    *   [x] When a meeting is _In progress,_ the default action in the right-hand pane wiill be &#39;Close meeting&#39;, which will change state accordingly
    *   [x] In the right hand pane, there is a new &#39;configure&#39; icon next to the &#39;Meeting status&#39; header that lets users switch directly from one state to another
*   [x] In the _In Progress_ state:
    *   [x] Agenda items remain editable but the note goes to a muted colour
    *   [x] An &quot;+ Outcome&quot; button is added to the bottom of each agenda item (regular and work packages)
        *   [x] This action is also added to the Agenda item More (⋯) action menu
    *   [x] A closed meeting takes the current state (with outcomes) and makes everything read-only, as is the case today
    *   [x] Going back to &#39;Open&#39; mode will change the colour of the note text to the default shade but keep any added outcomes as is, but they will not be editable until the state is changed to &#39;In progress&#39;
*   [x] Clicking on the &quot;+ Outcome&quot; button will display a CKEditor text area with label &#39;Outcome&#39; with two buttons below it: Cancel (secondary button), Save (primary button)
*   [x] A saved outcome is visible in a grey box (see Figma mockups)
*   [x] There is a More (⋯) icon button on outcomes with the following options:
    *   [x] Edit outcome
    *   [x] Copy link to clipboard **This links to the item by ID, which is not prefixed. It could happen that #item-123 is assigned twice, once for the outcome, once for the agenda item. it should be prefixed**
    *   [x] Remove outcome

oliverguenther (approved):

frontend/src/stimulus/controllers/dynamic/meeting-outcome-form.controller.ts
mrmir: Fair point, I've created https://community.openproject.org/wp/61923 for this

PR #17933, opened 2025-02-13
remove unnecessary spring startup from rubocop

ulferts:
# Ticket

N/A

Allow rubocop to run stable in my IDE.

The rubocop binstub included starting a spring server. Spring is notoriously unstable at least on my mac. From time to time, it just hangs. Additionally, its startup sometimes crashes because of it not finding a database connection (not sure why).

But moreover, starting a spring server should just not be necessary here. Rubocop does not need a running OpenProject application as it does static code analysis.

### DIFF
--- a/bin/rubocop
+++ b/bin/rubocop
@@ -1,4 +1,3 @@
 #!/usr/bin/env ruby
-load File.expand_path("spring", __dir__)
 require 'bundler/setup'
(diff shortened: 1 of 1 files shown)


toy (approved):

cbliard (approved):
Nice.
Never experienced this issue myself because I use `DISABLE_SPRING=1`.

PR #17940, opened 2025-02-13
Storage Health Check: OIDC Provisioned Current User 

mereghost:
### Related Work Package: [OP#61477](https://community.openproject.org/projects/cross-application-user-integration-stream/work_packages/61477)

Adds the check if the current user is provisioned by OIDC IdP.

### DIFF
--- a/modules/storages/app/common/storages/peripherals/nextcloud_connection_validator.rb
+++ b/modules/storages/app/common/storages/peripherals/nextcloud_connection_validator.rb
@@ -58,7 +58,7 @@ module Storages
       def sso_misconfigured
         return None() unless @storage.authenticate_via_idp?
 
-        audience_missing.or { non_provisioned_user }
+        audience_missing.or { non_provisioned_user }.or { oidc_provisioned_user }
       end
 
       def audience_missing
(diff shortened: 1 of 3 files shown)


modules/storages/app/common/storages/peripherals/nextcloud_connection_validator.rb
```diff
@@ -71,14 +71,23 @@ def audience_missing
...
-        return None() if @user.authentication_provider.present?
+        return None() if @user.identity_url.present?
 
         Some(ConnectionValidation.new(type: :warning,
                                       error_code: :oidc_non_provisioned_user,
                                       timestamp: Time.current,
                                       description: I18n.t("storages.health.connection_validation.oidc_non_provisioned_user")))
       end
 
+      def oidc_provisioned_user
```
NobodysNightmare: :yellow_circle: I think these methods are usually named after "what could be wrong here", but the user being OIDC provisioned means it's not wrong.

--> `not_oidc_provsioned_user`? `non_oidc_user`?

Not sure about a really good name for this, just noticed that the method had the name of the good case.

NobodysNightmare (approved):
After inverting the method name, this LGTM.

PR #17942, opened 2025-02-13
[#59540] highlight closest predecessor in relations tab

myabc:
⚠️ **This PR is based off of #17806** - please review/merge that work first.

# Ticket

https://community.openproject.org/wp/59540

Provide a visual indication of the closest predecessor in the Relations tab and Work Package date picker.

## Screenshots

| Relations tab | Date picker |
|---------------|-------------|
| Screenshot 2025-02-26 at 11 47 28 | Screenshot 2025-02-26 at 11 52 05 | 

https://github.com/user-attachments/assets/f01adfec-4f5f-4637-aaae-f58019bb0aaa

The logic to determine the closest relation is in UI (component) code rather than the model. Based on the requirements, I think this is UI logic rather than business logic.

`WorkPackageRelationsTab::ClosestRelation` is a custom comparator that does "the heavy lifting" by encapsulate the sorting logic.

**For discussion: a few ideas for possible improvements**

1. subclass `WorkPackageRelationsTab::RelationComponent`, creating `ChildRelationComponent` and `PredecessorRelationComponent` that accept only the necessary parameters. It would enable us to remove a bunch of branch logic (i.e. `if parent_child_relationship?`) in the base class.
2. We could also consider using the decorator/presenter pattern and making `ClosestRelation` delegate all missing methods to `relation`. `ClosestRelation` would take on some UI logic responsibility, rather than just comparison. (I prefer the first idea for improvement though).
3. Optimise iterating over relations and relation groups - there is scope for this, but it doesn't need to be done now (_premature optimisation and all that 😄_ )
4. Component specs for `WorkPackageRelationsTab::RelationComponent` itself.

# Merge checklist

- [x] Added/updated tests

### DIFF
--- a/app/components/work_package_relations_tab/closest_relation.rb
+++ b/app/components/work_package_relations_tab/closest_relation.rb
@@ -1,0 +1,68 @@
+# frozen_string_literal: true
+
+#-- copyright
+# OpenProject is an open source project management software.
+# Copyright (C) the OpenProject GmbH
+#
+# This program is free software; you can redistribute it and/or
+# modify it under the terms of the GNU General Public License version 3.
+#
+# OpenProject is a fork of ChiliProject, which is a fork of Redmine. The copyright follows:
+# Copyright (C) 2006-2013 Jean-Philippe Lang
+# Copyright (C) 2010-2013 the ChiliProject Team
+#
+# This program is free software; you can redistribute it and/or
+# modify it under the terms of the GNU General Public License
+# as published by the Free Software Foundation; either version 2
+# of the License, or (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program; if not, write to the Free Software
+# Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+#
+# See COPYRIGHT and LICENSE files for more details.
+#++
+module WorkPackageRelationsTab
+  class ClosestRelation
+    include Comparable
+
+    attr :relation
+
+    delegate :predecessor, to: :relation
+
+    def <=>(other)
+      comparison = compare_nilable_dates(soonest_start, other.soonest_start)
+      comparison = -compare_nilable_dates(predecessor.created_at, other.predecessor.created_at) if comparison.zero?
+      comparison
+    end
+
+    def initialize(relation)
+      @relation = relation
+    end
+
+    def soonest_start
+      return @soonest_start if defined?(@soonest_start)
+
+      @soonest_start = relation.successor_soonest_start(gap: 0.days)
+    end
+
+    def inspect
+      "#<#{self.class.name} soonest_start: #{soonest_start} relation: #{relation.inspect}>"
+    end
+
+    private
+
+    def compare_nilable_dates(date1, date2)
+      return -1 if date1.nil? && !date2.nil?
+      return 1 if !date1.nil? && date2.nil?
+      return 0 if date1.nil? && date2.nil?
+
+      date1 <=> date2
+    end
+  end
+end
--- a/app/components/work_package_relations_tab/relations_mediator.rb
+++ b/app/components/work_package_relations_tab/relations_mediator.rb
@@ -38,8 +38,28 @@ class WorkPackageRelationsTab::RelationsMediator
       visible_relations.any? || ghost_relations.any?
     end
 
+    def all_relations
+      visible_relations + ghost_relations
+    end
+
+    def closest_relation?(relation) = closest_relation == relation
+
+    def closest_relation = closest_relations.first
+
+    def closest_relations
+      return [] unless follows_type?
+
+      all_relations
+        .sort_by { WorkPackageRelationsTab::ClosestRelation.new(it) }
+        .reverse
+    end
+
     def children_type?
       type == "children"
+    end
+
+    def follows_type?
+      type == "follows"
     end
   end
 
--- a/app/components/work_package_relations_tab/relations_mediator.rb
+++ b/app/components/work_package_relations_tab/relations_mediator.rb
@@ -48,18 +48,15 @@ class WorkPackageRelationsTab::RelationsMediator
 
     def closest_relation?(relation) = closest_relation == relation
 
-    def closest_relation = closest_relations.first
-
-    def closest_relations
-      return [] unless type.follows?
+    def closest_relation
+      return nil unless type.follows?
 
       all_relations
         .lazy
         .map { WorkPackageRelationsTab::ClosestRelation.new(it) }
         .select(&:soonest_start)
-        .sort
-        .reverse
-        .map(&:relation)
+        .max
+        &.relation
     end
   end
 
--- a/app/components/work_package_relations_tab/relations_mediator.rb
+++ b/app/components/work_package_relations_tab/relations_mediator.rb
@@ -30,6 +30,10 @@
 
 class WorkPackageRelationsTab::RelationsMediator
   RelationGroup = Data.define(:type, :visible_relations, :ghost_relations) do
+    def initialize(type:, visible_relations:, ghost_relations:)
+      super(type: ActiveSupport::StringInquirer.new(type), visible_relations:, ghost_relations:)
+    end
+
     def count
       visible_relations.count + ghost_relations.count
     end
@@ -47,7 +51,7 @@ class WorkPackageRelationsTab::RelationsMediator
     def closest_relation = closest_relations.first
 
     def closest_relations
-      return [] unless follows_type?
+      return [] unless type.follows?
 
       all_relations
         .lazy
@@ -56,14 +60,6 @@ class WorkPackageRelationsTab::RelationsMediator
         .sort
         .reverse
         .map(&:relation)
-    end
-
-    def children_type?
-      type == "children"
-    end
-
-    def follows_type?
-      type == "follows"
     end
   end
 
--- a/app/components/work_package_relations_tab/relations_mediator.rb
+++ b/app/components/work_package_relations_tab/relations_mediator.rb
@@ -50,8 +50,12 @@ class WorkPackageRelationsTab::RelationsMediator
       return [] unless follows_type?
 
       all_relations
-        .sort_by { WorkPackageRelationsTab::ClosestRelation.new(it) }
+        .lazy
+        .map { WorkPackageRelationsTab::ClosestRelation.new(it) }
+        .select(&:soonest_start)
+        .sort
         .reverse
+        .map(&:relation)
     end
 
     def children_type?
--- a/spec/components/work_package_relations_tab/closest_relation_spec.rb
+++ b/spec/components/work_package_relations_tab/closest_relation_spec.rb
@@ -1,0 +1,204 @@
+# frozen_string_literal: true
+
+#-- copyright
+# OpenProject is an open source project management software.
+# Copyright (C) the OpenProject GmbH
+#
+# This program is free software; you can redistribute it and/or
+# modify it under the terms of the GNU General Public License version 3.
+#
+# OpenProject is a fork of ChiliProject, which is a fork of Redmine. The copyright follows:
+# Copyright (C) 2006-2013 Jean-Philippe Lang
+# Copyright (C) 2010-2013 the ChiliProject Team
+#
+# This program is free software; you can redistribute it and/or
+# modify it under the terms of the GNU General Public License
+# as published by the Free Software Foundation; either version 2
+# of the License, or (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program; if not, write to the Free Software
+# Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+#
+# See COPYRIGHT and LICENSE files for more details.
+#++
+
+require "rails_helper"
+
+RSpec.describe WorkPackageRelationsTab::ClosestRelation do
+  let(:work_package) { build_stubbed(:work_package) }
+  let(:today) { Time.zone.today }
+
+  def closest_relation(lag: 0, **wp_attributes)
+    predecessor = build_stubbed(:work_package, **wp_attributes)
+    relation = build_stubbed(:follows_relation, from: work_package, to: predecessor, lag:)
+    described_class.new(relation)
+  end
+
+  describe "#<=>" do
+    context "without a lag" do
+      context "when comparing two instances with different due dates" do
+        it "compares with the respective due dates" do
+          expect(closest_relation(due_date: 1.day.from_now)).to be < closest_relation(due_date: 2.days.from_now)
+          expect(closest_relation(due_date: today)).to be > closest_relation(due_date: 2.days.ago)
+          expect(closest_relation(due_date: today)).to be > closest_relation(due_date: nil)
+          expect(closest_relation(due_date: nil)).to be < closest_relation(due_date: 1.day.ago)
+        end
+      end
+
+      context "when comparing two instances with different start dates" do
+        it "compares with the respective start dates" do
+          expect(closest_relation(start_date: 1.day.from_now)).to be < closest_relation(start_date: 2.days.from_now)
+          expect(closest_relation(start_date: today)).to be > closest_relation(start_date: 2.days.ago)
+          expect(closest_relation(start_date: today)).to be > closest_relation(start_date: nil)
+          expect(closest_relation(start_date: nil)).to be < closest_relation(start_date: 1.day.ago)
+        end
+      end
+
+      context "when comparing two instances with same due dates" do
+        it "compares with the respective created at dates, oldest > newest" do
+          expect(closest_relation(due_date: 1.day.from_now)).to be > closest_relation(due_date: 1.day.from_now)
+          expect(closest_relation(due_date: 2.days.ago)).to be > closest_relation(due_date: 2.days.ago)
+          expect(closest_relation(due_date: nil)).to be > closest_relation(due_date: nil)
+        end
+      end
+
+      context "when comparing two instances with same start dates" do
+        it "compares with the respective created at dates, oldest > newest" do
+          expect(closest_relation(start_date: 1.day.from_now)).to be > closest_relation(start_date: 1.day.from_now)
+          expect(closest_relation(start_date: 2.days.ago)).to be > closest_relation(start_date: 2.days.ago)
+          expect(closest_relation(start_date: nil)).to be > closest_relation(start_date: nil)
+        end
+      end
+
+      context "when comparing two instances with both due and start dates set" do
+        it "compares with the respective due dates, ignoring start dates" do
+          expect(closest_relation(due_date: 10.days.from_now, start_date: today))
+            .to be < closest_relation(due_date: 12.days.from_now, start_date: 2.days.ago)
+          expect(closest_relation(due_date: 14.days.from_now, start_date: today))
+            .to be > closest_relation(due_date: 12.days.from_now, start_date: 2.days.ago)
+        end
+      end
+    end
+
+    context "with a lag" do
+      context "when comparing two instances with same soonest starts (due date and lag)" do
+        it "compares with the respective created at dates, oldest > newest" do
+          expect(closest_relation(lag: 3, due_date: 3.days.from_now))
+            .to be > closest_relation(lag: 4, due_date: 2.days.from_now)
+          expect(closest_relation(lag: 2, due_date: 3.days.ago))
+            .to be > closest_relation(lag: 1, due_date: 4.days.ago)
+          expect(closest_relation(lag: 2, due_date: nil))
+            .to be > closest_relation(lag: 2, due_date: nil)
+        end
+      end
+
+      context "when comparing two instances with same soonest starts (start date and lag)" do
+        it "compares with the respective created at dates, oldest > newest" do
+          expect(closest_relation(lag: 3, start_date: 3.days.from_now))
+            .to be > closest_relation(lag: 4, start_date: 2.days.from_now)
+          expect(closest_relation(lag: 2, start_date: 3.days.ago))
+            .to be > closest_relation(lag: 1, start_date: 4.days.ago)
+          expect(closest_relation(lag: 2, start_date: nil))
+            .to be > closest_relation(lag: 2, start_date: nil)
+        end
+      end
+
+      context "when comparing two instances with different soonest starts (due date and lag)" do
+        it "compares with the combined due dates and lag" do
+          expect(closest_relation(lag: 4, due_date: 3.days.from_now))
+            .to be > closest_relation(lag: 4, due_date: 2.days.from_now)
+          expect(closest_relation(lag: 1, due_date: 3.days.ago))
+            .to be < closest_relation(lag: 4, due_date: 3.days.ago)
+          expect(closest_relation(lag: 4, due_date: nil))
+            .to be < closest_relation(lag: 4, due_date: 2.days.from_now)
+          expect(closest_relation(lag: 1, due_date: 3.days.ago))
+            .to be > closest_relation(lag: 4, due_date: nil)
+        end
+      end
+
+      context "when comparing two instances with different soonest starts (start date and lag)" do
+        it "compares with the combined start dates and lag" do
+          expect(closest_relation(lag: 4, start_date: 3.days.from_now))
+            .to be > closest_relation(lag: 4, start_date: 2.days.from_now)
+          expect(closest_relation(lag: 1, start_date: 3.days.ago))
+            .to be < closest_relation(lag: 4, start_date: 3.days.ago)
+          expect(closest_relation(lag: 4, start_date: nil))
+            .to be < closest_relation(lag: 4, start_date: 2.days.from_now)
+          expect(closest_relation(lag: 1, start_date: 3.days.ago))
+            .to be > closest_relation(lag: 4, start_date: nil)
+        end
+      end
+    end
+  end
+
+  describe "#soonest_start" do
+    context "with a nil due and start date" do
+      it "returns nil" do
+        expect(closest_relation(due_date: nil, start_date: nil).soonest_start).to be_nil
+        expect(closest_relation(lag: 1, due_date: nil, start_date: nil).soonest_start).to be_nil
+      end
+    end
+
+    context "with a due date set" do
+      let(:due_date) { Date.new(2020, 7, 14) }
+
+      context "without a lag" do
+        it "returns the due date" do
+          expect(closest_relation(due_date:).soonest_start).to eq(Date.new(2020, 7, 14))
+        end
+      end
+
+      context "with a positive lag" do
+        it "returns the combined due date and lag" do
+          expect(closest_relation(lag: 11, due_date:).soonest_start).to eq(Date.new(2020, 7, 25))
+        end
+      end
+
+      context "with a negative lag" do
+        it "returns the combined due date and lag" do
+          pending "negative lags are not yet implemented"
+          expect(closest_relation(lag: -2, due_date:).soonest_start).to eq(Date.new(2020, 7, 12))
+        end
+      end
+    end
+
+    context "with a start date set" do
+      let(:start_date) { Date.new(2020, 7, 14) }
+
+      context "with a zero lag" do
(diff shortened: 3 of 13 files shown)


spec/components/work_package_relations_tab/relations_mediator_spec.rb
```diff
@@ -0,0 +1,48 @@
...
+    shared_let(:related_work_package1) { create(:work_package, subject: "WP #1", project:, due_date: today + 3.days) }
+    shared_let(:related_work_package2) { create(:work_package, subject: "WP #2", project:, due_date: today + 2.days) }
+    shared_let(:related_work_package3) { create(:work_package, subject: "WP #3", project:, due_date: today + 2.days) }
+    shared_let(:related_work_package4) { create(:work_package, subject: "WP #4", project:) }
+    shared_let(:relations) do
+      [create(:relation, from: work_package, to: related_work_package1, lag: 3, relation_type: Relation::TYPE_FOLLOWS),
+       create(:relation, from: work_package, to: related_work_package2, lag: 14, relation_type: Relation::TYPE_FOLLOWS),
+       create(:relation, from: work_package, to: related_work_package3, lag: 14, relation_type: Relation::TYPE_FOLLOWS),
+       create(:relation, from: work_package, to: related_work_package4, lag: 10, relation_type: Relation::TYPE_FOLLOWS)]
+    end
```
myabc: @cbliard I've seen that there is a `let_work_packages` helper that takes an ASCII table string - I didn't use it in this case, but it might be a good candidate?

spec/features/work_packages/details/relations/primerized_relations_spec.rb
```diff
@@ -148,7 +148,7 @@ def label_for_relation_type(relation_type)
 
       tabs.expect_counter("relations", 6)
 
-      relations_tab.expect_relation(relation_follows)
+      relations_tab.expect_closest_relation(relation_follows)
```
myabc: This isn't necessarily the right place to test for this - and this expectation isn't particularly robust, given that there is only one "follows" relation in this scenario.

myabc:
@cbliard as mentioned in the WP, this is ready for some initial feedback:

- is calling `successor_soonest_start` correct? - this is what I was most unsure about.

some more general questions:

- the work to calculate "closest" is in the component rather than the model (and also in Ruby rather than SQL - I've figured that we already have the models loaded in memory, so adding a custom comparator was the strategy I went with). I think there is scope for some later optimisation - from cursory glance it looks like we are iterating over relation groups unnecessary.
- this PR lacks feature specs - I am still trying to figure out where best to add.

myabc:
(Leaving this in "Draft" as I don't consider this mergeable, just ready for initial feedback)

app/components/work_package_relations_tab/closest_relation.rb
```diff
@@ -0,0 +1,58 @@
...
+      self.class.compare(relation, other.relation)
+    end
+
+    def initialize(relation)
+      @relation = relation
+    end
+
+    def inspect
+      @relation
+    end
```
cbliard: I think an instance should clearly reflect what it is to avoid confusion, here inspecting it would return something like `#<Relation id: 1013, from_id: 1012, to_id: 1006, lag: 0, description: nil, relation_type: "follows">` which looks like a `Relation` instance, but it's not and it does not behave like it either.

It should show what it is.

What about this?
```
    def inspect
      "#<#{self.class.name} relation: #{relation.inspect}>"
    end
```

It would display `#<WorkPackageRelationsTab::ClosestRelation relation: #<Relation id: 1013, from_id: 1012, to_id: 1006, lag: 0, description: nil, relation_type: "follows">>` which feels less confusing to me.

Or even
```ruby
    def inspect
      "#<#{self.class.name} soonest_start: #{soonest_start} relation: #{relation.inspect}>"
    end

    def soonest_start
      @soonest_start ||= relation.successor_soonest_start(gap: 0.days)
    end
```
that would help debugging: `#<WorkPackageRelationsTab::ClosestRelation soonest_start: 2025-02-26 relation: #<Relation id: 1013, from_id: 1012, to_id:...`

spec/components/work_package_relations_tab/closest_relation_spec.rb
```diff
@@ -0,0 +1,106 @@
...
+    context "with no lag" do
+      let(:lag_a) { 0 }
+      let(:lag_b) { 0 }
+
+      context "with different due_date" do
+        let(:related_work_package_a) { build_stubbed(:work_package, due_date: today + 3.days) }
+        let(:related_work_package_b) { build_stubbed(:work_package, due_date: today + 2.days) }
+
+        it_behaves_like "relation_a comes first"
+      end
```
cbliard: I find the unit tests a bit hard to read, and the example descriptions are not really describing what's going on.

```
WorkPackageRelationsTab::ClosestRelation
  <=>
    with no lag
      with different due_date
        behaves like relation_a comes first
          left operand is less than right
      with different start_date
        behaves like relation_a comes first
          left operand is less than right
      when both due_date and start_date set due_date takes precedence
        behaves like relation_b comes first
          left operand is greater than right
      with the same due_date relation with oldest work_package (created_at) comes first
        behaves like relation_a comes first
          left operand is less than right
```

What about something that looks like this?

```
WorkPackageRelationsTab::ClosestRelation
  <=>
    with no lag
      when comparing two instances with different due dates
        compares with the respective due dates
      when comparing two instances without any due date and with different start dates
        compares with the respective start dates
      when comparing two instances with the same due date
        compares with the respective creation dates
etc.
```

I feel like it better communicates the intent.

cbliard: Also I find it a bit counter intuitive that the one with the greater predecessor end date is not the one with the greater `ClosestRelation` when doing comparison.

To put it in other terms,
* yesterday < today

So I would have expected
* ClosestRelation(predecessor: WorkPackage(due_date: yesterday)) < ClosestRelation(predecessor: WorkPackage(due_date: today))

And actually it is the other way around.
Is there a good reason? Cause I would find it more natural to be the other way around.

spec/components/work_package_relations_tab/closest_relation_spec.rb
```diff
@@ -0,0 +1,106 @@
...
+
+      context "with different start_date" do
+        let(:related_work_package_a) { build_stubbed(:work_package, start_date: today + 3.days) }
+        let(:related_work_package_b) { build_stubbed(:work_package, start_date: today + 2.days) }
+
+        it_behaves_like "relation_a comes first"
+      end
```
cbliard: Have you considered using some helper methods rather than shared examples?

For instance, writing it like this:

```ruby
  def closest_relation(lag: 0, **wp_attributes)
    predecessor = build_stubbed(:work_package, **wp_attributes)
    relation = build_stubbed(:follows_relation, from: work_package, to: predecessor, lag:)
    described_class.new(relation)
  end

  context "when comparing two instances with different due dates" do
    it "compares with the respective due dates" do
      expect(closest_relation(due_date: today + 1.day)).to be > closest_relation(due_date: today + 2.days)
      expect(closest_relation(due_date: today)).to be < closest_relation(due_date: 2.days.ago)
    end
  end
```

More context about this:

One principle I like about rspec is to have a [self-contained test](https://thoughtbot.com/blog/the-self-contained-test), and using shared examples helps with brevity, but makes the reader jump back and forth to understand what the test does and how it works.

For instance, with "relation_a comes first", I'm not sure what "comes first" means, and I do not know how objects are built. Then when going to the implementation, I see `result`, and I have to jump to the subject definition, and then I have to check what `-1` means. Is it `a > b` or the other way around?

And then after reading, I understand that coming first means it is the smallest when comparing (a < b).

I find it easier to read about when the assertion is expressed in one line like this:
```ruby
expect(closest_relation(due_date: today + 1.day)).to be > closest_relation(due_date: today + 2.days)
```
And it's easy to add multiple different expectations in the same test case without much ceremony:
```ruby
expect(closest_relation(due_date: today + 1.day)).to be > closest_relation(due_date: today + 2.days)
expect(closest_relation(due_date: today + 1.day)).to be > closest_relation(due_date: nil)
expect(closest_relation(due_date: 5.years.ago)).to be > closest_relation(due_date: today + 2.days)
```

It's not always possible though.

What do you think?

spec/components/work_package_relations_tab/relations_mediator_spec.rb
cbliard: Yes, that would be a good candidate. 
The goal of this helper is to make it easier to scan and understand what's going on in the test.

Rewritten it could look like this:
```ruby
  describe "#relation_group", "follows" do
    shared_let_work_packages(<<~TABLE)
      hierarchy    | MTWTFSS    | scheduling mode | predecessors
      predecessor1 | XXX        | manual          |
      predecessor2 | XX         | manual          |
      predecessor3 | XX         | manual          |
      predecessor4 |            | manual          |
      work_package |          X | automatic       | predecessor1 with lag 2, predecessor2 with lag 7, predecessor3 with lag 7, predecessor4 with lag 10
    TABLE

    let(:group) { mediator.relation_group("follows") }

    it "returns the closest relation" do
      expect(group.closest_relation).to eq _table.relation(predecessor: predecessor2)
    end

    it "returns a boolean" do
      expect(group.closest_relation?(_table.relation(predecessor: predecessor2))).to be true
      expect(group.closest_relation?(_table.relation(predecessor: predecessor1))).to be false
    end

    it "returns the relations with the closest first" do
      expect(group.closest_relations).to eq [
        _table.relation(predecessor: predecessor2),
        _table.relation(predecessor: predecessor3),
        _table.relation(predecessor: predecessor1),
        _table.relation(predecessor: predecessor4)
      ]
    end
  end
```

I used smaller lags and names, because having all the predecessors on one line makes the line quite big. Maybe I should add a column `successors` to make it look better.

Also, as the examples are about different methods, the output looks a bit strange with `--format documentation`:

```
WorkPackageRelationsTab::RelationsMediator
  RelationGroup
    #relation_group follows
      returns the closest relation
      returns a boolean
      returns the relations with the closest first
```

I would add a describe block for each and change some wording to make it clearer, like so:

```ruby
  shared_let_work_packages(<<~TABLE)
    hierarchy    | MTWTFSS    | scheduling mode | predecessors
    predecessor1 | XXX        | manual          |
    predecessor2 | XX         | manual          |
    predecessor3 | XX         | manual          |
    predecessor4 |            | manual          |
    work_package |          X | automatic       | predecessor1 with lag 2, predecessor2 with lag 7, predecessor3 with lag 7, predecessor4 with lag 10
  TABLE

  describe "RelationGroup" do
    let(:group) { mediator.relation_group("follows") }

    describe "#closest_relation" do
      it "returns the closest follows relation" do
        expect(group.closest_relation).to eq _table.relation(predecessor: predecessor2)
      end
    end

    describe "#closest_relation?(relation)" do
      it "returns true if the given relation is the closest one, false otherwise" do
        expect(group.closest_relation?(_table.relation(predecessor: predecessor2))).to be true
        expect(group.closest_relation?(_table.relation(predecessor: predecessor1))).to be false
      end
    end

    describe "#closest_relations" do
      it "returns all follows relations with the closest first" do
        expect(group.closest_relations).to eq [
          _table.relation(predecessor: predecessor2),
          _table.relation(predecessor: predecessor3),
          _table.relation(predecessor: predecessor1),
          _table.relation(predecessor: predecessor4)
        ]
      end
    end
  end
```

The documentation output is nicer:

```
WorkPackageRelationsTab::RelationsMediator
  RelationGroup
    #closest_relation
      returns the closest relation
    #closest_relation?
      returns true if the given relation is the closest one, false otherwise
    #closest_relations
      returns all follows relations with the closest first```

File: spec/support/components/work_packages/relations.rb [hunk 6]
```diff
@@ -252,6 +252,15 @@ def expect_relation(relatable)
         find_row(relatable)
       end
 
+      def expect_closest_relation(relatable)
+        actual_relatable = find_relatable(relatable)
+        relation_row = find_row(actual_relatable)
+
+        within relation_row do
+          expect(page).to have_css(".Label", text: "Closest")
+        end
```
cbliard: With the new `it` of ruby 3.4, this can be rewritten like this:

```suggestion
        within relation_row do
          expect(it).to have_css(".Label", text: "Closest")
        end
```

Not that useful, but I find it neat in conjunction with `within`.

cbliard (commented):
It looks nice.

The feature works well, except when some of the predecessors do not have dates.
When that's the case, the one without dates is shown as the closest (if it has the oldest creation date), which does not makes sense.

Here is a screenshot:

image

I have not tested with all predecessors having no dates. I think that in this case it should not display any "Closest" label at all, but I think that's not specified yet.

I also made some other comments about the code and the tests.

spec/components/work_package_relations_tab/relations_mediator_spec.rb
cbliard: I'm hesitant to add a successors column.

What do you think?
Between

```
    shared_let_work_packages(<<~TABLE)
      hierarchy    | MTWTFSS    | scheduling mode | predecessors
      predecessor1 | XXX        | manual          |
      predecessor2 | XX         | manual          |
      predecessor3 | XX         | manual          |
      predecessor4 |            | manual          |
      work_package |          X | automatic       | predecessor1 with lag 2, predecessor2 with lag 7, predecessor3 with lag 7, predecessor4 with lag 10
    TABLE
```
and
```
    shared_let_work_packages(<<~TABLE)
      hierarchy    | MTWTFSS    | scheduling mode | successors
      work_package |          X | automatic       |
      predecessor1 | XXX        | manual          | work_package with lag 2
      predecessor2 | XX         | manual          | work_package with lag 7
      predecessor3 | XX         | manual          | work_package with lag 7
      predecessor4 |            | manual          | work_package with lag 10
    TABLE
```
which one reads better?

Also, only `X`, `[`, and `]` count as recognized characters. Any other character can be used for space, meaning we can represent lag visually with another character to make it more "readable". What about `-`?

It would look like this:
```
    shared_let_work_packages(<<~TABLE)
      hierarchy    | MTWTFSS    | scheduling mode | successors
      work_package |          X | automatic       |
      predecessor1 | XXX--      | manual          | work_package with lag 2
      predecessor2 | XX-------  | manual          | work_package with lag 7
      predecessor3 | XX-------  | manual          | work_package with lag 7
      predecessor4 |            | manual          | work_package with lag 10
    TABLE
```

Would it be better? (until we have negative lags...)

cbliard:
I realize I did not address your initial questions

> * is calling `successor_soonest_start` correct? - this is what I was most unsure about.

Yes, definitely. This method is the one used in the scheduling (through `WorkPackage#soonest_start` defined in `app/models/work_package/scheduling_rules.rb`).

> some more general questions:
> 
> * the work to calculate "closest" is in the component rather than the model (and also in Ruby rather than SQL - I've figured that we already have the models loaded in memory, so adding a custom comparator was the strategy I went with). I think there is scope for some later optimisation - from cursory glance it looks like we are iterating over relation groups unnecessary.

Yes, there is probably room for optimization, but let's focus on making it work first with reasonable performances.

> * this PR lacks feature specs - I am still trying to figure out where best to add.

Maybe in existing feature specs about relations? I saw you updated one. Not sure if there are other ones existing.

spec/features/work_packages/details/relations/primerized_relations_spec.rb
cbliard: It's fine for me: it tests that the relation is the closest one. That checks that everything it wired correctly.

When you say that it's not robust, I'm not sure to get it. In which ways could this assertion fail?

spec/components/work_package_relations_tab/relations_mediator_spec.rb
```diff
@@ -0,0 +1,48 @@
...
+
+  current_user { user }
+
+  subject(:mediator) { described_class.new(work_package:) }
+
+  describe "#relation_group", "follows" do
+    shared_let(:related_work_package1) { create(:work_package, subject: "WP #1", project:, due_date: today + 3.days) }
+    shared_let(:related_work_package2) { create(:work_package, subject: "WP #2", project:, due_date: today + 2.days) }
+    shared_let(:related_work_package3) { create(:work_package, subject: "WP #3", project:, due_date: today + 2.days) }
+    shared_let(:related_work_package4) { create(:work_package, subject: "WP #4", project:) }
```
cbliard: swapping the line with `related_work_package4` with the line `related_workPackage1` makes the test fail.

myabc:
@cbliard fantastic! - thanks for the detailed review and great suggestions. I'll fix the bug with predecessors without dates, and then begin incorporating your other suggestions!

spec/components/work_package_relations_tab/closest_relation_spec.rb
myabc: RSpec offers a very rich API*, and as such I sometimes succumb to the temptation to reach for a RSpec DSL methods like `shared_example`, forgetting that it is still just Ruby code. I can see that defining a `closest_relation` helper makes a lot more sense here - and that it'll express intent more concisely.

_* some would argue too comprehensive!_

> And then after reading, I understand that coming first means it is the smallest when comparing (a < b).

You're right. I'm using "coming first" somewhat out-of-context. Of course, the item would come first if the `#sort` were called on an Array of `ClosestRelation` instances (`[ClosestRelation.new(relation:), ClosestRelation.new(relation:)].sort`), but that's not what we're testing here. (tbh I managed to confuse myself when writing some of these examples - so that is probably a bad sign!)

spec/components/work_package_relations_tab/closest_relation_spec.rb
myabc: I guess calling `ClosestRelation.first` rather than `.last` felt more natural to me.

myabc: I've inverted the behaviour of `ClosestRelation#<=>` and instead call `reverse` after sorting in `RelationsGroup#closest_relations` - does this feel more intuitive?

spec/support/components/work_packages/relations.rb
myabc: the `it` implicit parameter is a feature that I make frequent use of in Kotlin/Android, so definitely nice to see it in Ruby.

spec/components/work_package_relations_tab/relations_mediator_spec.rb
myabc: This should be fixed - `RelationGroup#closest_relations` now excludes relations without dates.

spec/components/work_package_relations_tab/closest_relation_spec.rb
myabc: @cbliard I hope I've managed to take into account all your suggestions in this particular spec!

myabc:
@cbliard - regarding issues with the logic:

> The feature works well, except when some of the predecessors do not have dates.
When that's the case, the one without dates is shown as the closest (if it has the oldest creation date), which does not makes sense.

This should now be fixed.

> I have not tested with all predecessors having no dates. I think that in this case it should not display any "Closest" label at all, but I think that's not specified yet.

Even though we haven't finalised the specification, this _should_ now be the behaviour (I still need to do some manual testing Wed a.m.)

spec/components/work_package_relations_tab/relations_mediator_spec.rb
myabc: Although I like the idea of `-` representing lags I went ahead and implemented your first suggestion. I'll give it some more thought today (Wed)!

spec/features/work_packages/details/relations/primerized_relations_spec.rb
myabc: > When you say that it's not robust, I'm not sure to get it. In which ways could this assertion fail?

"Robust" might have been the wrong choice of words - a bit of an exaggeration. What I wanted to say was that, since there is a single "follows" relation in this scenario, we are not yet testing rendering multiple predecessors at a feature level (and any applicable logic regarding labelling, sorting - tbd - and grouping).

spec/components/work_package_relations_tab/closest_relation_spec.rb
cbliard: I think to get the closest, I would find the one with the maximum end + lag date, so I would call `relations.max` to find it, rather than `relations.first`.

And actually, is the list of closest relations needed?

spec/features/work_packages/details/relations/primerized_relations_spec.rb
cbliard: Ok got it. Well, I think that's cheap so it's fine.

app/components/work_package_relations_tab/closest_relation.rb
```diff
@@ -0,0 +1,69 @@
...
+    attr :relation
+
+    delegate :predecessor, to: :relation
+    delegate :created_at, to: :predecessor
+
+    def <=>(other)
+      comparison = compare_nilable_dates(soonest_start, other.soonest_start)
+      comparison = -compare_nilable_dates(created_at, other.created_at) if comparison.zero?
+      comparison
+    end
```
cbliard: Very clean and readable 👍

app/components/work_package_relations_tab/closest_relation.rb
```diff
@@ -0,0 +1,69 @@
...
+# See COPYRIGHT and LICENSE files for more details.
+#++
+module WorkPackageRelationsTab
+  class ClosestRelation
+    include Comparable
+
+    attr :relation
+
+    delegate :predecessor, to: :relation
+    delegate :created_at, to: :predecessor
```
cbliard: `delegate` supports multiple arguments:
```suggestion
    delegate :predecessor, :created_at, to: :relation
```

spec/components/work_package_relations_tab/closest_relation_spec.rb
```diff
@@ -0,0 +1,204 @@
...
+            .to be < closest_relation(lag: 4, start_date: 2.days.from_now)
+          expect(closest_relation(lag: 1, start_date: 3.days.ago))
+            .to be > closest_relation(lag: 4, start_date: nil)
+        end
+      end
+    end
+  end
+
+  describe "#soonest_start" do
+    context "with a nil due or start date" do
```
cbliard:
```suggestion
    context "with a nil due and start date" do
```

spec/components/work_package_relations_tab/closest_relation_spec.rb
```diff
@@ -0,0 +1,204 @@
...
+        it "returns the combined due date and lag" do
+          expect(closest_relation(lag: 11, due_date:).soonest_start).to eq(Date.new(2020, 7, 25))
+        end
+      end
+
+      context "with a negative lag" do
+        it "returns the combined due date and lag" do
+          pending "negative lags are not yet implemented"
+          expect(closest_relation(lag: -2, due_date:).soonest_start).to eq(Date.new(2020, 7, 12))
+        end
```
cbliard, line 167: Even if that should work as-is, negative lag is not officially implemented yet. It will be when this feature is done: [Feature: Allow users to enter negative lag (#58011) | Stream Planning and Reporting | OpenProject Community](https://community.openproject.org/projects/stream-planning-and-reporting/work_packages/58011/activity?query_id=5758)

Still, I think it's ok to keep this example.

cbliard (approved):
Nice!
Thanks for having incorporated my suggestions. I find the spec much easier to read and understand.

I tried it and could not find any bugs. Tried with ghost relations too.

Nice job 👏

spec/components/work_package_relations_tab/closest_relation_spec.rb
myabc: I'll append the WP number to the pending message.

app/components/work_package_relations_tab/closest_relation.rb
myabc: @cbliard as I've implemented it, we're falling back on `WorkPackage#created_at` and not `Relation#created_at` for cases where soonest starts are the same (or `nil`). 

As such, this does need to be `delegate :predecessor, to: :relation` (I've removed the second `delegate` line as inlining it in `#<=>` makes the intent clearer IMO).

Could you just double-check this correct?

spec/components/work_package_relations_tab/relations_mediator_spec.rb
myabc: @cbliard I'm erring for the third option - but unless I'm misunderstanding something, the specifying lag in the `successors` column is redundant for `predecessor1` - `predecessor3`?

Should I take a look at implementing this, or something you'd like to do? (either in this PR, or in a follow up?)

app/components/work_package_relations_tab/closest_relation.rb
cbliard: Sorry, I'm wrong. You're completly right.

spec/components/work_package_relations_tab/relations_mediator_spec.rb
cbliard: No that was just an idea. That's nice to have, but not mandatory at all.
For the values in it, I copied it from the original values:

```
    shared_let(:relations) do
      [create(:relation, from: work_package, to: related_work_package1, lag: 3, relation_type: Relation::TYPE_FOLLOWS),
       create(:relation, from: work_package, to: related_work_package2, lag: 14, relation_type: Relation::TYPE_FOLLOWS),
       create(:relation, from: work_package, to: related_work_package3, lag: 14, relation_type: Relation::TYPE_FOLLOWS),
       create(:relation, from: work_package, to: related_work_package4, lag: 10, relation_type: Relation::TYPE_FOLLOWS)]
    end
```

except that instead of lags of `[3, 14, 14, 10]` I used `[2, 7, 7, 10]`.

Or maybe you meant redundant with the ascii in the `MTWTFSS` column? Well, any character other than `X`, `[`, `]` can be there without being interpreted, so it's only helpful for us mere humans.

this table
```ruby
    shared_let_work_packages(<<~TABLE)
      hierarchy    | MTWTFSS    | scheduling mode | successors
      work_package |          X | automatic       |
      predecessor1 | XXX--      | manual          | work_package with lag 2
      predecessor2 | XX-------  | manual          | work_package with lag 7
      predecessor3 | XX-------  | manual          | work_package with lag 7
      predecessor4 |            | manual          | work_package with lag 10
    TABLE
```
is interpreted exactly in the same way as
```ruby
    shared_let_work_packages(<<~TABLE)
      hierarchy    | MTWTFSS    | scheduling mode | successors
      work_package | hello!!! X | automatic       |
      predecessor1 | X X....... | manual          | work_package with lag 2
      predecessor2 | XX.:.:.:.: | manual          | work_package with lag 7
      predecessor3 | XX.:|:.:|: | manual          | work_package with lag 7
      predecessor4 | !?!?!?!?!? | manual          | work_package with lag 10
    TABLE
```
I used that to represent non-working days in some specs like in [`spec/workers/work_packages/apply_working_days_change_job_spec.rb`](https://github.com/opf/openproject/blob/dev/spec/workers/work_packages/apply_working_days_change_job_spec.rb)
```ruby
      context "when a follower has a predecessor with lag covering a day that is now a non-working day" do
        let_work_packages(<<~TABLE)
          subject     | MTWTFSS | scheduling mode | predecessors
          predecessor | XX   ░░ | manual          |
          follower    |    X ░░ | automatic       | follows predecessor with lag 1
        TABLE

        before do
          set_non_working_week_days("wednesday")
        end

        it "moves the follower start date forward to keep the lag to 1 day" do
          subject
          expect(WorkPackage.all).to match_table(<<~TABLE)
            subject     | MTWTFSS |
            predecessor | XX░  ░░ |
            follower    |   ░ X░░ |
          TABLE
        end
      end
```

spec/components/work_package_relations_tab/closest_relation_spec.rb
myabc: > And actually, is the list of closest relations needed?

Now that we've agreed that we're going to keep sorting by creation date, no.

myabc: @cbliard Good point - I've applied your suggestion.

spec/support/capybara/primer_component_selectors.rb
```diff
@@ -0,0 +1,60 @@
...
+#
+# You should have received a copy of the GNU General Public License
+# along with this program; if not, write to the Free Software
+# Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+#
+# See COPYRIGHT and LICENSE files for more details.
+#++
+
+Capybara.add_selector :primer_label, locator_type: [String, Symbol] do
+  label "Primer Label"
```
cbliard, line 32: Nice selector! It makes the test more expressive.

app/components/work_package_relations_tab/relations_mediator.rb
```diff
@@ -48,18 +48,15 @@ def all_relations
...
 
       all_relations
         .lazy
         .map { WorkPackageRelationsTab::ClosestRelation.new(it) }
         .select(&:soonest_start)
-        .sort
-        .reverse
-        .map(&:relation)
+        .max
+        &.relation
```
cbliard: If it takes the `.max`, `.lazy` is useless.

cbliard (approved):
As before, all good for me. The code is expressive and working well.

I especially like the `:primer_label` Capybara selector.

cbliard:
Failing tests `./spec/features/work_packages/new/new_work_package_datepicker_spec.rb:51` and `./modules/costs/spec/features/time_entry_dialog_spec.rb:219` are passing locally.

Merging

PR #17949, opened 2025-02-14
Fix quote flaky test

brunopagno:
# Ticket
https://community.openproject.org/wp/61488

Fix flaky spec

`quote_comment` was submitting the comment as well. So there was a window where the ckeditor would still have the expected text. Now I removed the submit, so it should be more consistently finding the expected value inside the editor.


akabiru (approved):
🤼

akabiru:
Merging as you're OOO 🍵

PR #17951, opened 2025-02-14
Feature/59183 filters for stages and gates on project list

ulferts:
# Ticket

https://community.openproject.org/wp/59183

*   The project list offers filters on stages and gates    
    *   Available filters:        
        1.  Projects that are in any stage/gate            
            *   Name: "Lifecycle: Any stage or gate"                
        2.  Projects that are in a specific Stage/Gate            
            *   This will require having one filter per Stage/Gate that are dynamically added/removed when new stages/gates are added.                
            *   The activation/deactivation of a Stage/Gate within a project does not factor into the availability of the filter. As soon as the stage/gate is created in the administration, the filter is available.                
            *  Name for a stage: "Lifecycle stage: [Name of the stage]"
            *  Name for a gate: "Lifecycle gate: [Name of the gate]"               
    *   Operators for both filters:        
        *   today - no value required: Matches if the specified stage/gate (for 2.) or if any stage/gate (for 1.) has a start date lower and an end date higher than today.            
        *   this week - no value required: Matches if the specified stage/gate&#39;s (for 2.) or if any stage/gate&#39;s (for 1.) interval formed by start and end date overlap the date interval created by the current week.            
            *   The calculation for the current week takes the &quot;Week starts on&quot; Setting into account.                
        *   on &nbsp;- one value required: Matches if the specified stage/gate (for 2.) or if any stage/gate (for 1.) has a start date lower and an end date higher than the provided date.            
            *   \[open\] find term for this operator - the user would want to know if something happens on that day.                
        *   between - two values: Matches if the specified stage/gate&#39;s (for 2.) or if any stage/gate&#39;s (for 1.) interval formed by start and end date are within the date interval created by the two values. If only the lower value is provided, it matches if the stage/gate&#39;s values are higher than the value. If only the upper value is provided, it matches if the stage/gate&#39;s values are lower than the value.            
        *   is empty - no value required: Matches if the specified stage/gate (for 2.) or if any stage/gate (for 1.) has no value set.            
    *   Restrictions        
        *   In case a stage/gate is disabled in a project the value that stage/gate might have set in the project is not found.            
        *   The filters are only available if the user has the &quot;View project phases and gates&quot; permission in any project.            
        *   Only values set for stages/gates within a project are only filtered on if the user has the &quot;View project phases and gates&quot; in that project.            
*   The filters are also available via the API v3.    
    *   Those added filters need to be documented in the API documentation.
    

# Merge checklist

- [x] Added/updated tests

### DIFF
--- a/app/models/queries/projects.rb
+++ b/app/models/queries/projects.rb
@@ -28,45 +28,47 @@
 
 module Queries::Projects
   ::Queries::Register.register(ProjectQuery) do
-    filter Filters::AncestorFilter
-    filter Filters::AvailableProjectAttributesFilter
-    filter Filters::AvailableCustomFieldsProjectsFilter
-    filter Filters::TypeFilter
     filter Filters::ActiveFilter
-    filter Filters::TemplatedFilter
-    filter Filters::PublicFilter
-    filter Filters::NameFilter
-    filter Filters::NameAndIdentifierFilter
-    filter Filters::MemberOfFilter
-    filter Filters::TypeaheadFilter
-    filter Filters::CustomFieldFilter
+    filter Filters::AncestorFilter
+    filter Filters::AnyStageOrGateFilter
+    filter Filters::AvailableCustomFieldsProjectsFilter
+    filter Filters::AvailableProjectAttributesFilter
     filter Filters::CreatedAtFilter
-    filter Filters::LatestActivityAtFilter
-    filter Filters::PrincipalFilter
-    filter Filters::ParentFilter
+    filter Filters::CustomFieldFilter
+    filter Filters::FavoredFilter
     filter Filters::IdFilter
+    filter Filters::LatestActivityAtFilter
+    filter Filters::LifeCycleGateFilter
+    filter Filters::LifeCycleStageFilter
+    filter Filters::MemberOfFilter
+    filter Filters::NameAndIdentifierFilter
+    filter Filters::NameFilter
+    filter Filters::ParentFilter
+    filter Filters::PrincipalFilter
     filter Filters::ProjectStatusFilter
+    filter Filters::PublicFilter
+    filter Filters::TemplatedFilter
+    filter Filters::TypeFilter
+    filter Filters::TypeaheadFilter
     filter Filters::UserActionFilter
     filter Filters::VisibleFilter
-    filter Filters::FavoredFilter
-    filter Filters::AnyStageOrGateFilter
 
+    order Orders::CustomFieldOrder
     order Orders::DefaultOrder
     order Orders::LatestActivityAtOrder
-    order Orders::RequiredDiskSpaceOrder
-    order Orders::CustomFieldOrder
     order Orders::LifeCycleStepOrder
-    order Orders::ProjectStatusOrder
     order Orders::NameOrder
+    order Orders::ProjectStatusOrder
+    order Orders::RequiredDiskSpaceOrder
     order Orders::TypeaheadOrder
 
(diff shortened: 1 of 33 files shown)


app/models/queries/projects.rb
```diff
@@ -28,44 +28,47 @@
...
     order Orders::TypeaheadOrder
 
     select Selects::CreatedAt
     select Selects::CustomField
     select Selects::Default
+    select Selects::Favored
     select Selects::LatestActivityAt
     select Selects::LifeCycleStep
     select Selects::RequiredDiskSpace
     select Selects::Status
```
ulferts, line 72: Throughout the PR, a couple of lists were sorted.

ulferts:
Given that the functionality is behind a feature flag and not to be released in 15.4. the version is correct.

app/models/queries/projects/filters/life_cycle_gate_filter.rb
```diff
@@ -0,0 +1,81 @@
...
+
+  class << self
+    def key
+      /\Alcsd_gate_(\d+)\z/
+    end
+
+    private
+
+    def name_for_step(stage)
+      "lcsd_gate_#{stage.id}"
```
dombesz, line 43: Just a typo, but this should be gate.

```suggestion
    def name_for_step(gate)
      "lcsd_gate_#{gate.id}"
```

dombesz: Just a typo, the argument should be `gate`.

```suggestion
    def name_for_step(gate)
      "lcsd_gate_#{gate.id}"
```

lib/open_project/internationalization/date.rb
```diff
@@ -0,0 +1,57 @@
...
+# Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+#
+# See COPYRIGHT and LICENSE files for more details.
+# ++
+
+# The module is not called I18n as that leads to tons of conflicts
+# where I18n is called within the OpenProject module and rails I18n is meant.
+
+module OpenProject
+  module Internationalization
```
dombesz, line 35: I don't have anything agains this name. I have few more ideas like keeping it `I18n` or ditching the `Internationalization` and use `OpenProject::DateUtils` instead.

spec/models/queries/projects/project_query_results_life_cycle_gate_spec.rb
```diff
@@ -0,0 +1,446 @@
...
+      end
+    end
+
+    context "when only the upper value is provided and that one is before the gate's date" do
+      let(:values) { ["", (gate_date - 1.day).to_s] }
+
+      it "returns no project" do
+        expect(instance.results).to be_empty
+      end
+    end
```
dombesz, line 385: The opposite cases for the lower value seem to be missing.

Edit: since the stage spec has the 3 lower cases and 1 upper case, I assume this is intentional.

dombesz (approved):
Nice work Jens, I left a few notes and questions, but feel free to merge anyway.

lib/open_project/internationalization/date.rb
ulferts: @dombesz , I tried using I18n but that collided heavily with the global I18n module. Because now every call from code within `OpenProject` would need to clarify whether the OpenProject I18n or the global I18n is meant. That is why I abandoned it. But I am definitively open for suggestions. 

Will merge in the meantime until we have an agreed upon solution.

PR #17962, opened 2025-02-17
Wait for cost entry to update after save before test assertion

akabiru:
Allow the page to update after saving the cost entry. Avoid possible race condition where the test was checking the value before the page has finished updating

https://github.com/opf/openproject/actions/runs/13326139713/job/37219715506

```
modules/costs/spec/features/cost_entries/add_cost_entry_spec.rb:110 # Work Package cost fields with german locale creates the budget including the given cost items with german locale
```

### DIFF
--- a/modules/costs/spec/features/cost_entries/add_cost_entry_spec.rb
+++ b/modules/costs/spec/features/cost_entries/add_cost_entry_spec.rb
@@ -149,7 +149,9 @@ RSpec.describe "Work Package cost fields", :js do
       fill_in "cost_entry_overridden_costs", with: "55.000,55"
       click_on I18n.t(:button_save)
 
-      expect(page).to have_css("#cost_entry_costs", text: "55.000,55 EUR")
+      # Add explicit wait for the updated cost value
+      wait_for { page }.to have_css("#cost_entry_costs", text: "55.000,55 EUR")
+
       entry.reload
       expect(entry.units).to eq(1.42)
       expect(entry.costs).to eq(2.84)


modules/costs/spec/features/cost_entries/add_cost_entry_spec.rb
```diff
@@ -149,7 +149,9 @@
       fill_in "cost_entry_overridden_costs", with: "55.000,55"
       click_on I18n.t(:button_save)
 
-      expect(page).to have_css("#cost_entry_costs", text: "55.000,55 EUR")
+      # Add explicit wait for the updated cost value
+      wait_for { page }.to have_css("#cost_entry_costs", text: "55.000,55 EUR")
+
```
cbliard, line 154: I don't get why this would be different with `wait_for`. For me they are both waiting for the same thing.

> Avoid possible race condition where the test was checking the value before the page has finished updating

But it does not check the value, it tries to find a node matching css query `#cost_entry_costs` and having the text "55.000,55 EUR". If the text is something else, it does not match and continues waiting.

If the code was this instead:
```ruby
cost_entry_costs_element = page.find("#cost_entry_costs")
expect(cost_entry_costs_element.text).to eq("55.000,55 EUR")
```
then yes that would be checking the value before the page has finished updating.

So I'm a bit skeptical about that one. I suppose that's what you wanted to explain us at last dev weekly.

cbliard (approved):
As explained in the comment, I'm a bit skeptical that it behaves differently, but I see nothing wrong with the proposed code, so it's worth a try.

If you have time to explain me why the previous code fails and why the new code works, I'd be interested :)

modules/costs/spec/features/cost_entries/add_cost_entry_spec.rb
akabiru: > I don't get why this would be different with `wait_for`. For me they are both waiting for the same thing.
> 
> > Avoid possible race condition where the test was checking the value before the page has finished updating
> 
> But it does not check the value, it tries to find a node matching css query `#cost_entry_costs` and having the text "55.000,55 EUR". If the text is something else, it does not match and continues waiting.
> 
 
I meant "text" in this case, typo. On save, the assertion that the text is updated to `"55.000,55 EUR"` seems to race ahead before the actual response is received back in some cases, which resulted in the flakiness.

Using `wait_for` has it retry until the page is updated. Perhaps an even better option would be to reduce from the default timeout of 10s-  to sth link 2s so that we don't mask any slowness in response times.

cbliard: My point is that `expect(page).to have_css("#cost_entry_costs", text: "55.000,55 EUR")` is also waiting. If the element is there, but the text is not the one being specified, then it will continue trying until it matches entirely.

And I think you pointed to something: `wait_for` default timeout is 10s, and I don't think we change that setting.
Capybara default timeout has been configured to 4s (`Capybara.default_max_wait_time = 4` in `spec/support/capybara.rb`). 

If the actions being done is particularly long, it may be why there is a different behavior for both. If that's so, using `expect(page).to have_css("#cost_entry_costs", text: "55.000,55 EUR", wait: 10)` should behave the same.

But looking at the test, when "Save" button is clicked, it's really updating quickly, so I don't understand why a long delay should be needed.

When looking at the screenshot (https://openproject-ci-public-logs.s3-eu-west-1.amazonaws.com/screenshot_2025-02-14-09-25-28.017.png), the text is "0,00 €". No matter how long we wait or with which helper, the text is wrong and it won't change. Something did not update correctly or at the wrong time.

image

So I think this test will continue to flicker.

I remember that I already fixed some flickering on this test: https://github.com/opf/openproject/commit/dca3349fc59. There seems to be another edge case about it.

These flaky specs are so frustrating :(

cbliard: Actually, I think it's the same issue than the one I fixed: the test does `visit edit_cost_entry_path(entry)` but then does not wait for the spent_on date field to be fully loaded. I suspect that the issue occurs when it's not loaded yet, and the value is updated.

I'll try to reproduce it by making the loading taking longer. If I'm right, then it should always fail locally.

cbliard: The test is maybe not flickering: there is a real issue.

I was able to reproduce the issue quite consistently: when updating the costs, sometimes it resets to 0 for no reasons.

Here is a video:

[updating costs then saving does not work.webm](https://github.com/user-attachments/assets/3ba5cf31-afb1-4b89-a608-96d2ec951766)

cbliard: And there is a short window where it displays the correct value before being reset to "0.00 €" or some other values. I suppose that's why it passes sometimes, but actually it should fail because the text is reset.

It seems to happen when losing focus while clicking Save button directly. If losing focus first, then click Save, the issue does not occur.

I'll continue investigating.

cbliard: Yes, there is a real issue which comes from clicking on save and leaving the field at the same time. The good value is saved and then a request is sent to update the form values based on the fields being set.

@dombesz proposed to have a look at it and fix it.

akabiru: > My point is that expect(page).to have_css("#cost_entry_costs", text: "55.000,55 EUR") is also waiting. If the element is there, but the text is not the one being specified, then it will continue trying until it matches entirely.

Perhaps I'm missing something but I don't think that's the case, there was no capybara `wait:` option- it should not wait, if it finds the node `#cost_entry_costs` with a different text other than "55.000,55 EUR" it should fail immediately - but log that it found a different text on that node.

Ex. https://github.com/opf/openproject/actions/runs/13326139713/job/37219715506

```log
  2) Work Package cost fields with german locale creates the budget including the given cost items with german locale
     Failure/Error: example.run
       expected to find visible css "#cost_entry_costs" with text "55.000,55 EUR" but there were no matches. Also found "0,00 EUR", which matched the selector but not all filters. 
```

akabiru: > And I think you pointed to something: `wait_for` default timeout is 10s, and I don't think we change that setting. Capybara default timeout has been configured to 4s (`Capybara.default_max_wait_time = 4` in `spec/support/capybara.rb`).
> 
> If the actions being done is particularly long, it may be why there is a different behavior for both. If that's so, using `expect(page).to have_css("#cost_entry_costs", text: "55.000,55 EUR", wait: 10)` should behave the same.
> 
> But looking at the test, when "Save" button is clicked, it's really updating quickly, so I don't understand why a long delay should be needed.

We can actually modify the `wait_for` timeout properties- it becomes

```ruby
wait(timeout: 2).for { page }.to have_...
```

My hypothesis is that the save button is clicked quickly but the page doesn't always update quickly enough before the #cost_entry_costs assertion is ran- which is why the value is still "0,00 EUR" on the UI.

Let's observe it / sync up perhaps it's easier. But I agree that a long wait time should not be needed.

dombesz: The problem comes from the fact that the field updater script uses a get query and the browser caches the stale responses. Using a post query solves the issue, please see https://github.com/opf/openproject/pull/18020 .

cbliard: > > My point is that expect(page).to have_css("#cost_entry_costs", text: "55.000,55 EUR") is also waiting. If the element is there, but the text is not the one being specified, then it will continue trying until it matches entirely.
> 
> Perhaps I'm missing something but I don't think that's the case, there was no capybara `wait:` option- it should not wait, if it finds the node `#cost_entry_costs` with a different text other than "55.000,55 EUR" it should fail immediately - but log that it found a different text on that node.

That is the case. `expect(page).to have_css("css query", conditions)` waits, even without a `wait` parameter. And once it has waited and failed, it displays an error message explaining what it expected, and what it found. To get it to fail immediately, `wait: 0` must be used.

This is explained here: https://www.rubydoc.info/gems/capybara/3.12.0#asynchronous-javascript-ajax-and-friends

If you are interested by technical details, the heart of it in in [`Capybara::Node::Base#synchronize`](https://github.com/teamcapybara/capybara/blob/0480f90168a40780d1398c75031a255c1819dce8/lib/capybara/node/base.rb#L76). It tries to find with the given argument (css query and text in this case), fails, waits a little and tries again until the full `default_max_wait_time` is elapsed.

And in this case, it's good that it is failing: this test revealed an issue with the way the cost edition currently works.

akabiru: > The problem comes from the fact that the field updater script uses a get query and the browser caches the stale responses. Using a post query solves the issue, please see #18020 .

Ah, really nice find @dombesz - then perhaps we can revert this commit. 🧹

akabiru: > > > My point is that expect(page).to have_css("#cost_entry_costs", text: "55.000,55 EUR") is also waiting. If the element is there, but the text is not the one being specified, then it will continue trying until it matches entirely.
> > 
> > 
> > Perhaps I'm missing something but I don't think that's the case, there was no capybara `wait:` option- it should not wait, if it finds the node `#cost_entry_costs` with a different text other than "55.000,55 EUR" it should fail immediately - but log that it found a different text on that node.
> 
> That is the case. `expect(page).to have_css("css query", conditions)` waits, even without a `wait` parameter. And once it has waited and failed, it displays an error message explaining what it expected, and what it found. To get it to fail immediately, `wait: 0` must be used.
> 
> This is explained here: https://www.rubydoc.info/gems/capybara/3.12.0#asynchronous-javascript-ajax-and-friends
> 
> If you are interested by technical details, the heart of it in in [`Capybara::Node::Base#synchronize`](https://github.com/teamcapybara/capybara/blob/0480f90168a40780d1398c75031a255c1819dce8/lib/capybara/node/base.rb#L76). It tries to find with the given argument (css query and text in this case), fails, waits a little and tries again until the full `default_max_wait_time` is elapsed.
> 
> And in this case, it's good that it is failing: this test revealed an issue with the way the cost edition currently works.

Ah, thank you for clarifying this! I had the wrong understanding 😅

PR #17966, opened 2025-02-17
[#61466] show hover cards in WP multi select CFs

EinLama:
# Ticket
https://community.openproject.org/wp/61466

Show the hover card for multi user CF within work packages.

I did not find an easy way to display the popover for the tagged selections while the dropdown is open. See attached GIF. Since you can get the popover when the dropdown is not visible, there is an easy workaround for our users.

## Screenshots

multiuser_cf

# Merge checklist

- [ ] Added/updated tests

### DIFF
--- a/frontend/src/app/shared/components/fields/edit/field-types/multi-select-edit-field.component.ts
+++ b/frontend/src/app/shared/components/fields/edit/field-types/multi-select-edit-field.component.ts
@@ -243,7 +243,7 @@ export class MultiSelectEditFieldComponent extends EditFieldComponent implements
    * For multi-select fields that are of type User, we want to show a hover card when hovering over users in the
    * dropdown. For this to happen we must define a URL for the hover card.
    */
-  protected getHoverCardUrl(item:any) {
+  protected getHoverCardUrl(item:{ id?:string }) {
     if (item && item.id) {
       return this.pathHelperService.userHoverCardPath(item.id);
     }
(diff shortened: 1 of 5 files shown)


EinLama:
eslint is rather unhappy about my changes. But it was very unhappy with this file to begin with.

> ✖ 47 problems (47 errors, 0 warnings)
>  1 error and 0 warnings potentially fixable with the `--fix` option.

frontend/src/app/shared/components/fields/edit/field-types/multi-select-edit-field.component.ts
```diff
@@ -223,4 +226,30 @@ export class MultiSelectEditFieldComponent extends EditFieldComponent implements
...
+    }
+
+    return '';
+  }
+
+  /**
+   * For multi-select fields that are of type User, we want to show a hover card when hovering over users in the
+   * dropdown. For this to happen we must define a URL for the hover card.
+   */
+  protected getHoverCardUrl(item:any) {
```
dombesz: I believe the following type declaration should do the trick.
```suggestion
  protected getHoverCardUrl(item:{ id?:string }) {
```

Alternatively:
```typescript
protected getHoverCardUrl({ id }: { id?: string }) {
  if (id) {
  //...
  }
```

dombesz (approved):
Works as expected, I left one suggestion about making the eslint errors go away.

frontend/src/app/shared/components/fields/edit/field-types/multi-select-edit-field.component.ts
EinLama: It does, thank you! 👍🏻

PR #17973, opened 2025-02-17
60858 in add relation dialog selecting a wp that user doesnt have permission to edit raises an error

bsatarnejad:
# Ticket
https://community.openproject.org/wp/60858

Before adding a child, we check if the user has permission to edit the selected WP and set a parent for it.

## Screenshots
Screenshot 2025-02-17 at 16 51 35

Create a method in update contract of Workpackages to check if the user has permission to edit the WP and then call this method before setting a WP as Child in relations tab.

# Merge checklist

- [ ] Added/updated tests

### DIFF
--- a/app/contracts/work_packages/update_contract.rb
+++ b/app/contracts/work_packages/update_contract.rb
@@ -49,7 +49,7 @@ module WorkPackages
     attribute_permission :project_id, :move_work_packages
 
     def can_set_parent?
-      @can.allowed?(model, :edit)
+      @can.allowed?(model, :manage_subtasks)
     end
 
     private
--- a/app/contracts/work_packages/update_contract.rb
+++ b/app/contracts/work_packages/update_contract.rb
@@ -48,6 +48,10 @@ module WorkPackages
     default_attribute_permission :edit_work_packages
     attribute_permission :project_id, :move_work_packages
 
+    def can_set_parent?
+      @can.allowed?(model, :edit)
+    end
+
     private
 
     def user_allowed_to_edit
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -1410,6 +1410,7 @@ en:
               must_be_empty_when_work_is_empty_and_percent_complete_is_100p: >-
                 must be empty when Work is empty and % Complete is 100%.
           readonly_status: "The work package is in a readonly status so its attributes cannot be changed."
+          lack_of_permission: "Cannot add child because work package is read-only."
         type:
           attributes:
             attribute_groups:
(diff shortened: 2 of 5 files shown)


config/locales/en.yml
```diff
@@ -1362,6 +1362,10 @@ en:
         work_package:
           is_not_a_valid_target_for_time_entries: "Work package #%{id} is not a valid target for reassigning the time entries."
           attributes:
+            id:
+              format: "%{message}"
+              lack_of_permission: "Cannot add child because you don't have permissions to edit the selected work package."
+              blank: "ID can't be blank."
```
cbliard: I understand that you tweaked format and blank so that the lack of permission error can be displayed without the "ID" prefix, and blank with the "ID" prefix. But I do not quite like having to do this: it may have an impact on how current and future errors are displayed on other places.

Wouldn't it be better to simply do `child.errors.add(:id, I18n.t(:errors.cannot_add_child_because_of_lack_of_permission))`?

app/contracts/work_packages/update_contract.rb
```diff
@@ -48,6 +48,10 @@ class UpdateContract < BaseContract
     default_attribute_permission :edit_work_packages
     attribute_permission :project_id, :move_work_packages
 
+    def can_set_parent?
+      @can.allowed?(model, :edit)
+    end
```
cbliard: I tested more and it looks like it's not enough: when I test with the edit_work_packages permission, it still fails, because one additional permission is needed: `:permission_manage_subtasks` ("Manage work package hierarchies")

And I also tried with only `:permission_manage_subtasks`, and it works! So it should check for `:manage_subtasks` permission instead.

And I tried also by sharing it with "Edit" level. It also does not display any error, but does not work. I think it will be fixed once `:manage_subtasks` permission is used.

cbliard (changes requested):
When we paired together, I was wrong about using that `@can.allowed?(model, :edit)`. Actually the permission to check is `:manage_subtasks`. Without it, no parent can be set.

And also, the fact that something goes wrong and no error is displayed is something I don't like. It should display an error if `:base` or `:id` errors are not present.

Here is a code proposal to display any other error as a base error:

on line 52, intead of:
```ruby
base_errors: service_result.errors[:base]),
```
use
```ruby
base_errors: base_errors(service_result.errors),
```

and then
```ruby
  def base_errors(errors)
    if errors[:base].present?
      errors[:base]
    elsif errors[:id].present?
      nil
    else
      errors.full_messages
    end
  end
```

That will display an error if there is one on a field other than `:id`.

And while testing I also realized that the "Delete relation" action is proposed on a child even if the user is lacking the `:manage_subtasks` permission. I'll create another bug for it.

config/locales/en.yml
cbliard: Actually, I'm wrong: the attribute prefix is always added.
Better keep it like it is.

cbliard (approved):
All good 👏

PR #17974, opened 2025-02-17
[61551] Fix: make child creation from relations tab work in status-based progress calculation mode

cbliard:
In status-based mode, the done_ratio is automatically added to the work package, and it's read-only too. It should not be submitted as a hidden field in the create child dialog form.

Excluding all read-only attributes from being submitted as hidden fields fixes the issue.

# Ticket

https://community.openproject.org/wp/61551

In status-based mode, it's not possible to create child work package from the relations tab.

It's because hidden parameters are added to the form based on the work package default values, and in status based mode, the % complete value gets a default value, but it's also read-only so submitting the form sends a 400 error.

Excluding all read-only attributes from being added to the create form as hidden fields, so it fixes this issue, and it could fix potential similar issues.

# Merge checklist

- [x] Added/updated tests

### DIFF
--- a/app/forms/work_packages/dialogs/create_form.rb
+++ b/app/forms/work_packages/dialogs/create_form.rb
@@ -94,7 +94,10 @@ module WorkPackages::Dialogs
       render_custom_fields(form: f)
 
       # Keep hidden fields for relevant changes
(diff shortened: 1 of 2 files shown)


dombesz (approved):

dombesz:
Works as expected, feel free to merge when the CI is green.

PR #17979, opened 2025-02-18
Allow to configure SSO with fallback to two-way OAuth 2.0

NobodysNightmare:
Our code to select how we authenticate mostly accounted for this already, so it was just a matter of allowing admins to configure it.

We now focus on what's been explicitly configured, so that it is not necessary to clear the audience or the oauth client/application just when changing the settings. Otherwise stale configuration could lead to the wrong authentication to happen contrary to the configuration.

# Ticket
https://community.openproject.org/projects/cross-application-user-integration-stream/work_packages/61532

# Merge checklist

- [x] Added/updated tests
- [x] Tested Firefox

### DIFF
--- a/modules/storages/spec/common/storages/peripherals/storage_interaction/authentication_strategies/nextcloud_strategies/user_bound_spec.rb
+++ b/modules/storages/spec/common/storages/peripherals/storage_interaction/authentication_strategies/nextcloud_strategies/user_bound_spec.rb
@@ -46,7 +46,7 @@ RSpec.describe Storages::Peripherals::StorageInteraction::AuthenticationStrategi
     end
 
     context "if file storage is configured for sso and oauth" do
-      let(:storage) { create(:nextcloud_storage_configured, :oidc_sso_enabled) }
+      let(:storage) { create(:nextcloud_storage_configured, :oidc_sso_with_fallback) }
 
       it "must use an SsoUserToken strategy" do
         strategy = described_class.call(user:, storage:)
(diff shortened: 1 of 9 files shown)


modules/storages/spec/common/storages/peripherals/storage_interaction/authentication_strategies/nextcloud_strategies/user_bound_spec.rb
```diff
@@ -95,13 +95,13 @@
     end
   end
 
-  context "if file storage is not fully configured" do
+  context "if file storage is configured for OAuth 2.0, but client and app not fully configured" do
```
ba1ash: It seems like this use case should be under `if user is local` context, right?

modules/storages/spec/common/storages/peripherals/storage_interaction/authentication_strategies/nextcloud_strategies/user_bound_spec.rb
```diff
@@ -95,13 +95,13 @@
...
     let(:user) { create(:user) }
     let(:storage) { create(:nextcloud_storage) }
 
-    it "must return the failure strategy" do
+    it "returns the configured strategy" do
```
ba1ash: Why is it not failing in this case anymore? The returned strategy will not be usable, will it? If we do not care anymore about that, then I suppose this test case can be removed.

NobodysNightmare: True, the returned strategy will not be usable. Let me think, we could also use the failure strategy, but the end result would be the same: failure.

I think I chose the current approach for its simplicity in implementation, but maybe it's safer to return failure then, because we don't risk to run into a nil pointer or similar.

In either case I'd keep the spec, because I'd say that we always "care" about the behaviour of the user bound strategy, it's just that our opinion on this specific case changed.

modules/storages/spec/common/storages/peripherals/storage_interaction/authentication_strategies/nextcloud_strategies/user_bound_spec.rb
NobodysNightmare: true, I'll move it.

ba1ash (approved):
:+1:

PR #17992, opened 2025-02-19
Explicitly store expiration time of access tokens

NobodysNightmare:
Previously we didn't consider the `expires_in` from the token endpoint response, thus being unable to know whether a token was expired or not before using it. While we worked around that by inspecting tokens that seemed readable as JWTs, the better approach is to memoize the information returned by the server, as that also allows to know the expiration time of opaque access tokens.

We'll keep the JWT extraction around as a fallback mechanism, but prefer reading the info from the token endpoint, if it's present.

# Ticket
https://community.openproject.org/projects/openproject/work_packages/61344

# Merge checklist

- [x] Added/updated tests
- [x] Tested major browsers (Firefox)

### DIFF
--- a/spec/support/shared/timecop.rb
+++ b/spec/support/shared/timecop.rb
@@ -1,0 +1,36 @@
+# frozen_string_literal: true
+
+#-- copyright
+# OpenProject is an open source project management software.
+# Copyright (C) the OpenProject GmbH
+#
+# This program is free software; you can redistribute it and/or
+# modify it under the terms of the GNU General Public License version 3.
+#
+# OpenProject is a fork of ChiliProject, which is a fork of Redmine. The copyright follows:
+# Copyright (C) 2006-2013 Jean-Philippe Lang
+# Copyright (C) 2010-2013 the ChiliProject Team
+#
+# This program is free software; you can redistribute it and/or
+# modify it under the terms of the GNU General Public License
+# as published by the Free Software Foundation; either version 2
+# of the License, or (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program; if not, write to the Free Software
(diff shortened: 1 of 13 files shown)


spec/support/shared/timecop.rb
```diff
@@ -0,0 +1,40 @@
...
+#
+# You should have received a copy of the GNU General Public License
+# along with this program; if not, write to the Free Software
+# Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+#
+# See COPYRIGHT and LICENSE files for more details.
+#++
+
+RSpec.configure do |config|
+  config.around do |example|
```
mereghost: 🟡 Wouldn't `config.around(:example, :freeze_time)` be preferred here, so that the block only runs on examples that have the tag?

mereghost (commented):
🟡 Is it worth to add some VCR cassettes over the specs that do requests to Keycloak?

spec/support/shared/timecop.rb
NobodysNightmare: Uh, sweet. I didn't even know about that. This totally makes sense!

NobodysNightmare:
> 🟡 Is it worth to add some VCR cassettes over the specs that do requests to Keycloak?

You mean the ones where I manually set-up webmock to return a token response? I kinda preferred explicit mocking so far, because I considered the responses to be simple and standardized enough.

My concerns with VCR would be that cassettes turn out to be considered a magic black box (i.e. you don't look inside, you just expect them to work) and that we might encode Keycloak-specifics that are not strictly necessary/guaranteed by the RFCs.

Does that make sense? Do you think I should add other spec helpers instead then? (e.g. `stub_oauth2_token_response` or similar)

mereghost:
I have the opposite fear: that the mocked responses will be forever permanent even when API changes. XD

If you think it is fine, it is fine. =)

NobodysNightmare:
> I have the opposite fear: that the mocked responses will be forever permanent even when API changes. XD

I see. Nah, I don't have that fear, because all the mockings are sticking to the specs. But agreed: If we wanted to test Keycloak-specifics, then we should use VCR.

> If you think it is fine, it is fine. =)

**It's fine!** :D

ba1ash (approved):
LGTM :+1:

PR #17994, opened 2025-02-19
limit time entry comment to the current database value with a rails validation

klaustopher:
# Ticket
https://community.openproject.org/work_packages/61602

Currently we have a DB limit of 255 characters for the comment field. When entering more than 255 characters, the DB insert fails. Instead we add a validation error. For 15.4 we will increase the limit

## Screenshots
Bildschirmfoto 2025-02-19 um 10 26 53

Added a rails migration

# Merge checklist

- [x] Added/updated tests


oliverguenther (approved):

klaustopher:
WP changed to 15.3.1

PR #17999, opened 2025-02-19
Adds the preliminar check for an existing token

mereghost:
### Related WP: [OP#61608](https://community.openproject.org/projects/cross-application-user-integration-stream/work_packages/61608)

Implements the first step for checking the actual OIDC tokens. This is a preliminary check verifying if the current user already has a token with the required audiences.

The next step is to validate the token with a request and also try to grab a new one (or refresh... whatever FetchService wants to do).

### DIFF
--- a/modules/openid_connect/app/services/openid_connect/user_tokens/fetch_service.rb
+++ b/modules/openid_connect/app/services/openid_connect/user_tokens/fetch_service.rb
@@ -85,7 +85,7 @@ module OpenIDConnect
       private
 
       def token_with_audience(aud)
-        token = @user.oidc_user_tokens.where("audiences ? :aud", aud:).first
+        token = @user.oidc_user_tokens.with_audience(aud).first
         return Success(token) if token
 
         return @token_exchange.call(aud) if @token_exchange.supported?
--- a/modules/openid_connect/spec/factories/oidc_user_token_factory.rb
+++ b/modules/openid_connect/spec/factories/oidc_user_token_factory.rb
@@ -9,6 +9,6 @@ FactoryBot.define do
     user
     access_token { "INVALID_TOKEN" }
     refresh_token { "COOL_AID_TOKEN" }
-    audiences { ["__op-idp__"] + Array(extra_audiences) }
+    audiences { [OpenIDConnect::UserToken::IDP_AUDIENCE] + Array(extra_audiences) }
   end
 end
--- a/modules/openid_connect/spec/factories/oidc_user_token_factory.rb
+++ b/modules/openid_connect/spec/factories/oidc_user_token_factory.rb
@@ -1,0 +1,14 @@
+# frozen_string_literal: true
+
+FactoryBot.define do
+  factory :oidc_user_token, class: "OpenIDConnect::UserToken" do
+    transient do
+      extra_audiences { nil }
+    end
+
+    user
+    access_token { "INVALID_TOKEN" }
+    refresh_token { "COOL_AID_TOKEN" }
+    audiences { ["__op-idp__"] + Array(extra_audiences) }
+  end
+end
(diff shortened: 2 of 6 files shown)


modules/openid_connect/app/services/openid_connect/user_tokens/fetch_service.rb
```diff
@@ -85,7 +85,7 @@ def refreshed_access_token_for(audience:)
       private
 
       def token_with_audience(aud)
-        token = @user.oidc_user_tokens.where("audiences ? :aud", aud:).first
+        token = @user.oidc_user_tokens.with_audience(aud).first
```
NobodysNightmare, line 88: :green_heart: Totally makes sense to move this into a scope for this model, since it's a common query to expect.

modules/openid_connect/spec/factories/oidc_user_token_factory.rb
```diff
@@ -0,0 +1,14 @@
...
+FactoryBot.define do
+  factory :oidc_user_token, class: "OpenIDConnect::UserToken" do
+    transient do
+      extra_audiences { nil }
+    end
+
+    user
+    access_token { "INVALID_TOKEN" }
+    refresh_token { "COOL_AID_TOKEN" }
+    audiences { ["__op-idp__"] + Array(extra_audiences) }
```
NobodysNightmare: :green_circle:  Should we source the magic string `"__op-idp__"` from `OpenIDConnect::UserToken::IDP_AUDIENCE`?

modules/storages/app/common/storages/peripherals/nextcloud_connection_validator.rb
```diff
@@ -58,7 +58,7 @@ def validate
       def sso_misconfigured
         return None() unless @storage.authenticate_via_idp?
 
-        audience_missing.or { non_provisioned_user }.or { non_oidc_provisioned_user }
+        audience_missing.or { non_provisioned_user }.or { non_oidc_provisioned_user }.or { token_usability }
```
NobodysNightmare: :information_source: I gifted you a merge conflict here (see my comment in another PR): `audience_missing` was removed as a check entirely, because there is already a check on whether configuration is complete and since my change over there, we actually do check the audience as part of the configuration (i.e. `configured?`).

modules/storages/spec/common/storages/peripherals/nextcloud_connection_validator_spec.rb
```diff
@@ -275,6 +276,15 @@
       expect(subject.error_code).to eq(:oidc_non_oidc_user)
       expect(subject.description).to eq(I18n.t("storages.health.connection_validation.oidc_non_oidc_user"))
     end
+
+    it "returns an error if the user does not have a usable token" do
+      user.update(identity_url: "#{oidc_provider.slug}:UNIVERSALLY-DUPLICATED-IDENTIFIER")
```
NobodysNightmare: :green_circle: A hint that I usually give during tests is that `update!` would be preferrable here. If  `user` _had_ a validation added in the future that would make this update statement invalid, `update` would silently not be performed. In the best case this makes your test case fail later, but harder to debug. In the worst case it keeps your test case succeeding, but not testing what you intended.

In contrast `update!` would raise an error and tell you about the problem immediately. I think in OP this is less of an issue, because we have fewer validations on the model, but I think the general idea remains: If you don't check the return value of `update`, you should use `update!`, so that you don't have to.

NobodysNightmare (approved):
Some micro-suggestions from my side, but generally LGTM.

modules/openid_connect/spec/factories/oidc_user_token_factory.rb
mereghost: Oh! I missed the constant. Thanks for pointing it out.

modules/storages/spec/common/storages/peripherals/nextcloud_connection_validator_spec.rb
mereghost: Agreed. 

Usually I'd just create a new user for each special case as I think it communicates better but since I've been messing with the structure of the tests all the time, I have relied on updates. >.<

PR #18000, opened 2025-02-19
Task/61607 allow more characters for time entry comment

klaustopher:
# Ticket
https://community.openproject.org/projects/stream-time-and-costs/work_packages/61607

Allow comments up to 1000 characters for time entries. We do want to increase the limit from the 255 we have right now. But we also do not want to allow arbitrary lengths of text. So we limit to 1000 characters on the model. DB is uncapped

## Screenshots
grafik
grafik
grafik

DB column is text, limit comes from rails

# Merge checklist

- [x] Added/updated tests


as-op (approved):
Super! 👍

PR #18002, opened 2025-02-19
Bump rubocop_performance

ulferts:
# Ticket

N/A

Bump rubocop-performance and liste it as a rubocop plugin. This avoids the problem we faced e.g. with rubocop-rails


ulferts:
The flickering feature spec can be ignored. I attempted to fix it in dev. 

Locally I checked that the rubocop is still active.

klaustopher (approved):

PR #18009, opened 2025-02-20
Group AWS gems into one dependabot PR

klaustopher:
# Ticket
n/a

AWS gems are always released in bulk so this change groups their updates together in one PR and thus reduces the number of PRs we will see for AWS gems


ulferts (approved):
Good idea, @klaustopher.

PR #18015, opened 2025-02-20
Show an error when logging time on a work package you did not have permission to log for

klaustopher:
# Ticket
https://community.openproject.org/projects/stream-time-and-costs/work_packages/61635

For an admin, the autocompleter even lists work packages of projects that do not have the time&costs module enabled. When trying to log time for those work packages, the form submit returns an error but no error is shown. This PR adds a validation error to the work package that is presented to the user.

Additionally we also set the correct properties on the autocompleter so that we are using the custom URL that would prevent even selecting this work package

## Screenshots

# Merge checklist

- [x] Added/updated tests

### DIFF
--- a/modules/costs/app/components/time_entries/work_package_form.rb
+++ b/modules/costs/app/components/time_entries/work_package_form.rb
@@ -84,7 +84,11 @@ module TimeEntries
     end
 
     def work_package_validation_error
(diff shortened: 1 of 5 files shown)


klaustopher:
version in WP added

oliverguenther (approved):

PR #18018, opened 2025-02-20
Allow admins to choose not exchanging for specific SSO token 

NobodysNightmare:
This is intended for simple setups, where the token returned by the IDP can be immediately used for authentication at a storage. It is also required for integrating with Nextcloud as an IDP.

## Ticket
https://community.openproject.org/projects/cross-application-user-integration-stream/work_packages/61623

## What are you trying to accomplish?
It should be possible for admins to use the admin token obtained from the IDP directly for authentication at the storage.

## Screenshots

image

## Merge checklist

- [x] Added/updated tests
- [x] Tested firefox/chromium

### DIFF
--- a/frontend/src/stimulus/controllers/dynamic/storages/storage-audience.controller.ts
+++ b/frontend/src/stimulus/controllers/dynamic/storages/storage-audience.controller.ts
@@ -48,7 +48,6 @@ export default class StorageAudienceController extends Controller {
   }
 
   showAudienceInput() {
-    this.audienceInputTarget.value = '';
     this.audienceInputWrapperTarget.classList.remove('d-none');
   }
 
--- a/frontend/src/stimulus/controllers/dynamic/storages/storage-audience.controller.ts
+++ b/frontend/src/stimulus/controllers/dynamic/storages/storage-audience.controller.ts
@@ -1,0 +1,58 @@
+/*
+ * -- copyright
+ * OpenProject is an open source project management software.
+ * Copyright (C) the OpenProject GmbH
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU General Public License version 3.
+ *
+ * OpenProject is a fork of ChiliProject, which is a fork of Redmine. The copyright follows:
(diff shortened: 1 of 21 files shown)


frontend/src/stimulus/controllers/dynamic/storages/storage-audience.controller.ts
```diff
@@ -0,0 +1,58 @@
...
+  declare readonly idpRadioTarget:HTMLInputElement;
+
+  connect() {
+    if (this.idpRadioTarget.checked) {
+      this.hideAudienceInput();
+    }
+  }
+
+  showAudienceInput() {
+    this.audienceInputTarget.value = '';
```
ba1ash: :yellow_circle: 
Setting value to be empty on every show leads to user loosing its input which can be inconvenient. For example, in `Automatically managed folder` section of the same page password is not removed when user clicks on checkbox.
Therefore I suggest not clearing input value, but then `__op-idp__` can appear in the input field. I'm not sure if that is critical.

ba1ash (approved):
Well done!
I have one UX related remark.

frontend/src/stimulus/controllers/dynamic/storages/storage-audience.controller.ts
NobodysNightmare: > but then __op-idp__ can appear in the input field

This is exactly why I added the clearing, to prevent this "magic" audience from leaking through.

I understand the general inconvenience of losing an input, on the other hand we could at best prevent losing it when you click around in the form. As soon as you save, your input is lost anyways, because the radio button is just a way to set the same data.

Let me try something...

NobodysNightmare: Thanks for nudging me over. I found a solution that is much cleaner, because we never even send the internal identifier to the browser.

Feel free to have a look at my latest commit.

ba1ash: Looks nice to me! Thanks for taking care of it.

PR #18019, opened 2025-02-20
[#57618] If user removes all columns for their PDF export then default columns are used  

as-op:
# Ticket

https://community.openproject.org/work_packages/57618

* Show an error when a column selection is required for an export (pdf-table, xls, csv)
* Allow exporting a PDF-report with no columns selected at all, not using the default columns of a query

## Screenshots
error message with no columns selected

Bit more of a feature than a bug tbh, anyway:
* ng-select does not have a built-in form validator or _require_ parameter, nor is it a primer component, so I mimicked a validation error below the component with Angular. It is displayed immediately after the last column entry is removed and removed if there is at least one column.

* if there is no column selected where it has to, the stimulus controller blocks the form to be submitted and the modal dialog is not closed. 

# Implementation checklist
- [x] Mimic a primer validation error below the angular draggable ng-select
- [x] in the export dialog require a column selection of min 1 for pdf-table, csv, xls
- [x] do not use default columns in pdf-report if none are  selected by the user

# Merge checklist

- [x] Added/updated tests


EinLama (approved):
Works nicely for PDF (table and report), XLS and CSV 👍🏻 

Code reads good, too! Great job!

I noticed one minor flaw with the current solution: when the error message is displayed, it is impossible to submit the form (aka create the report). This is good and intended. However, the `Export` button looks like it is still active. You can click it, but nothing will happen as long as the form error is unresolved.

image

Depending on how we deal with this in other Angular forms, it might be nice to also set the button to `disabled` in case there is a form error - or whatever we do there 🤔 
As most users won't encounter this visual issue, it might not be worth it. I will let you be the judge of it 👨🏻‍⚖️

as-op:
Hi @EinLama, thanks for the kind review! You are totally right about the disabled button. But…

as we would need a listener from stimulus into Angular or the other way round calling stimulus from Angular, I weighed up it is not worth the complexity effort to do so. 
As you said, I agree this is a small issue only very few people will have a use case for, as it just makes no sense to export column-based documents and sheets without columns :D

PR #18020, opened 2025-02-20
Fix cost log edit form field updater

dombesz:
# Ticket
https://community.openproject.org/wp/61689

The Costs field display an invalid value, when the form is submitted while still focusing the cost field. This happens because the cost field updater logic runs after the form is submitted, overwriting the correctly rendered form content.
## Screenshots
<details><summary>Before</summary>
<p>

[updating costs then saving does not work.webm](https://github.com/user-attachments/assets/3ba5cf31-afb1-4b89-a608-96d2ec951766)

</p>
</details> 

<details><summary>After</summary>
<p>

fix

</p>
</details> 

Use a `post` request instead of a `get` so the updater script's stale response values are not cached. 

# Merge checklist

- [ ] Added/updated tests

### DIFF
--- a/frontend/src/app/shared/components/remote-field-updater/remote-field-updater.component.ts
+++ b/frontend/src/app/shared/components/remote-field-updater/remote-field-updater.component.ts
@@ -45,11 +45,7 @@ export class RemoteFieldUpdaterComponent implements OnInit, OnDestroy {
 
   private url:string;
 
-  private htmlMode:boolean;
-
   private form:HTMLFormElement;
-
-  private target:HTMLElement;
 
   private debouncedUpdaterBound:EventListener;
 
@@ -60,12 +56,10 @@ export class RemoteFieldUpdaterComponent implements OnInit, OnDestroy {
   ngOnInit():void {
     const element = this.elementRef.nativeElement as HTMLElement;
     this.form = element.closest('form') as HTMLFormElement;
-    this.target = this.form.querySelector('.remote-field--target') as HTMLElement;
     this.costTypeSelect = this.form.querySelector('#cost_entry_cost_type_id');
     this.unitsTextField = this.form.querySelector('#cost_entry_units');
 
     this.url = element.dataset.url as string;
-    this.htmlMode = element.dataset.mode === 'html';
 
     this.debouncedUpdaterBound = _.debounce(this.updater.bind(this), 500);
 
@@ -111,12 +105,7 @@ export class RemoteFieldUpdaterComponent implements OnInit, OnDestroy {
   private request(params:Record<string, string>) {
     const headers:Record<string, string> = {};
 
-    // In HTML mode, expect html response
-    if (this.htmlMode) {
-      headers.Accept = 'text/html';
-    } else {
-      headers.Accept = 'application/json';
-    }
+    headers.Accept = 'application/json';
 
     return this.http
       .get(
@@ -124,7 +113,7 @@ export class RemoteFieldUpdaterComponent implements OnInit, OnDestroy {
         {
           params,
           headers,
-          responseType: (this.htmlMode ? 'text' : 'json') as any,
+          responseType: 'json',
           withCredentials: true,
         },
       );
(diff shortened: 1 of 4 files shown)


cbliard (changes requested):
Thanks for looking into it.

I think it does not work as expected anymore: when updating the number of units, it should update the cost in real time. With the changes you introduced, this is no longer the case.

Before (working):

[before.webm](https://github.com/user-attachments/assets/d9ff2c72-d194-44c4-bdbb-6ccaec0e1458)

After (not working anymore):

[after.webm](https://github.com/user-attachments/assets/62a299fe-15ea-4399-b6d8-02a6de742967)

Could you please have a second look?

frontend/src/app/shared/components/remote-field-updater/remote-field-updater.component.ts
```diff
@@ -88,12 +88,12 @@ export class RemoteFieldUpdaterComponent implements OnInit {
...
-      .get(
+      .post(
         this.url,
         {
           params,
           headers,
-          responseType: (this.htmlMode ? 'text' : 'json') as any,
+          responseType: (this.htmlMode ? 'text' : 'json'),
           withCredentials: true,
         },
```
cbliard: The new code does not work as expected because the signature for post is different from the one for get.
* get is `get(url, options)`
* post is `post(url, body, options)`

Updating it makes it work again, but the bug is still there.

code to update it:
```
    return this.http
      .post(
        this.url,
        params,
        {
          headers,
          withCredentials: true,
        },
      );
```

cbliard (changes requested):
It's much better, but it still does not work as expected in one case: if a cost type has different rates, for instance these rates:
* 12 euros from 20 Feb
* 13 euros from 25 Feb

Then when changing dates between 21 Feb and 26 Feb, the resulting cost should be updated.
Currently it's not because the date field is an `HTMLInputElement`.

I have the feeling that the code lazily sets both 'change' and 'keyup' events on the whole form, but it does not really make sense. It would benefit from setting the right event for each relevant field:

* work_package_id: no listeners
  * Not involved in cost computation. BTW I hardly understand why this field is editable. The description below is not even updated when it's changed.
* date: change listener
* user: change listener
* cost type: change listener
* unit: keyup listener, or better: input listener
  * keyup is bad: it fires when pressing the arrows to move the cursor, or when pressing Ctrl or Alt, which is useless, but does not fire if the user clicks in the field and paste some content using "Edit" -> "Paste" (or right clicks and select Paste, or middle clicks to paste (with Linux))
  * input is better: it will fire as soon as the value is changed, no matter how it happens. It will not fire for key presses which do not update the field.
* costs: no listeners
  * not involved in cost computation: if the cost is manually entered, it does not need to be computed
* comment no listeners
  * not involved in cost computation

That would be better.

frontend/src/app/shared/components/remote-field-updater/remote-field-updater.component.ts
dombesz: Reverted it.

cbliard:
@dombesz I gave it a try.
The trick was to call cancel on the debounce function on destroy.
Then I did as proposed and added event listeners only on the relevant fields involved in the cost calculation.

It seems to now work as expected, even if I'm not sure to understand how the overridden cost works: as soon as it is set, it is always used. The user has to clear the field to use the calculated cost again. Yet, when the overridden cost is there, updating something (the units for instance) will recompute the cost but it will not be saved. The overridden cost will continue to be used. That's quite a confusing UI, but that was already like this.

I also removed the `fill_in_with_send_keys` you introduced. It's no longer needed now that it relies on `update` event instead of `keyup`.

And I also removed some dead code.

Could you please review my changes?

dombesz:
> Yet, when the overridden cost is there, updating something (the units for instance) will recompute the cost but it will not be saved. The overridden cost will continue to be used. That's quite a confusing UI, but that was already like this.

@cbliard I also find that behaviour confusing, I would assume if an overriden cost is entered, the cost calculation should not update if  the units are changed.

> The trick was to call cancel on the debounce function on destroy.
I like this solution better, thanks!

dombesz (commented):
@cbliard I cannot approve the changes, but they look good. Thanks!

cbliard (approved):

PR #18030, opened 2025-02-21
Update Gemfile.lock

NobodysNightmare:
This was apparently ~~forgotten~~ _missing_ after updating the Gemfile to 0.54.1.


NobodysNightmare:
FYI @myabc, as far as I can tell your change touched the `Gemfile.lock` but it didn't update the checksums. Do you need to update bundler?

NobodysNightmare:
I am daring to merge this without explicit approval...

PR #18033, opened 2025-02-21
[#61178] Update RemoteIdentity after token obtained.

ba1ash:
# Ticket

https://community.openproject.org/wp/61178
https://community.openproject.org/wp/61291

With this PR I try to make `RemoteIdentity` more flexible and being able to store user identities for combination of `auth_source: oauth_client or oidc provider` and `integration: storage`

The goal is accomplished by adding two polymorphic associations to `RemoteIdentity`: `integration` and `auth_source`.

# Merge checklist

- [x] Updated tests.
- [x] Tested with Chromium


app/models/remote_identity.rb
```diff
@@ -30,7 +30,8 @@
 
 class RemoteIdentity < ApplicationRecord
   belongs_to :user
-  belongs_to :oauth_client, class_name: "OAuthClient"
+  belongs_to :oauth_client, polymorphic: true
```
NobodysNightmare: I am wondering whether this should be renamed after being made polymorphic. One of the polymorphic options is for it to be an OAuth Client, so it would be confusing if the association is called `oauth_client` when it sometimes is an oauth client and sometimes not.

I'd suggest `auth_source`.

db/migrate/20250220123358_add_oauth_client_type_to_remote_identities.rb
```diff
@@ -0,0 +1,55 @@
...
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program; if not, write to the Free Software
+# Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+#
+# See COPYRIGHT and LICENSE files for more details.
+#++
+
+class AddOAuthClientTypeToRemoteIdentities < ActiveRecord::Migration[7.1]
```
NobodysNightmare: Actually more things happen in this migration than announced by the name.

db/migrate/20250220123358_add_oauth_client_type_to_remote_identities.rb
```diff
@@ -0,0 +1,55 @@
...
+  def up
+    add_column :remote_identities, :oauth_client_type, :string
+    add_column :remote_identities, :integration_type, :string
+    add_column :remote_identities, :integration_id, :bigint
+
+    add_index :remote_identities, %i[oauth_client_type integration_id integration_type]
```
NobodysNightmare: :red_circle: This index looks suspicious to me:

* I'd have expected separate indexes for `oauth_client` and `integration`. Right now the index is not usable for fetching identities of a given integration
* `oauth_client_id` is missing, making the index less useful to find identities of a specific oauth client (it can only filter identities for the correct auth type)
* The order of `integration_type` and `integration_id` should be inverted, since it never makes sense to just query for the id, but it can make sense to just query for the type (e.g. finding identities associated to nextcloud storages)

modules/openid_connect/app/services/openid_connect/user_tokens/exchange_service.rb
```diff
@@ -65,6 +67,11 @@ def call(audience)
         # A second reason is that at least Keycloak (an IDP we implement against), offers broken
         # refresh tokens after token exchange (see https://github.com/keycloak/keycloak/issues/37016)
         token = store_exchanged_token(audience:, access_token:, refresh_token: nil, expires_in:)
+
+        storage = Storages::Storage.all.find { |s| s.audience == audience }
```
NobodysNightmare: :red_circle: I think this code should be in the `FetchService`, otherwise it would only be called on token exchange and many times we do not perform a token exchange.

:yellow_circle: I think we should not hardwire this generic piece of authentication code to _storages_ specifically. We now have a direct dependency from OIDC to storages.

NobodysNightmare (changes requested):

db/migrate/20250220123358_add_oauth_client_type_to_remote_identities.rb
ba1ash: Of course, we do not need composite index here. It is a mistake. It should be 3 separate indexes.

NobodysNightmare:
Something that I just found during integration testing. We also need to update modules/storages/app/services/storages/nextcloud_managed_folder_sync_service.rb:

```diff
diff --git a/modules/storages/app/services/storages/nextcloud_managed_folder_sync_service.rb b/modules/storages/app/services/storages/nextcloud_managed_folder_sync_service.rb
index 78e27cc34ae..76e34852db6 100644
--- a/modules/storages/app/services/storages/nextcloud_managed_folder_sync_service.rb
+++ b/modules/storages/app/services/storages/nextcloud_managed_folder_sync_service.rb
@@ -318,7 +318,7 @@ def active_project_storages_scope
     end
 
     def remote_identities_scope
-      RemoteIdentity.includes(:user).where(oauth_client: @storage.oauth_client)
+      RemoteIdentity.includes(:user).where(integration: @storage)
     end
 
     def auth_strategy
@@ -326,7 +326,7 @@ def auth_strategy
     end
 
     def admin_remote_identities_scope
-      RemoteIdentity.includes(:user).where(oauth_client: @storage.oauth_client, user: User.admin.active)
+      RemoteIdentity.includes(:user).where(integration: @storage, user: User.admin.active)
     end
   end
 end
```

This makes sure that we add all users with a remote identity for the given storage to the Nextcloud group.

Probably the same for `OneDriveManagedFolderSyncService`, even though there we do not yet officially support anything else than oauth clients.

NobodysNightmare:
One more:

```diff
diff --git a/modules/storages/app/common/storages/peripherals/storage_interaction/nextcloud/util.rb b/modules/storages/app/common/storages/peripherals/storage_interaction/nextcloud/util.rb
index 2888c25fd8e..08b99c4f18c 100644
--- a/modules/storages/app/common/storages/peripherals/storage_interaction/nextcloud/util.rb
+++ b/modules/storages/app/common/storages/peripherals/storage_interaction/nextcloud/util.rb
@@ -71,7 +71,7 @@ def origin_user_id(caller:, storage:, auth_strategy:)
               when :basic_auth
                 ServiceResult.success(result: storage.username)
               when :oauth_user_token, :sso_user_token
-                origin_user_id = RemoteIdentity.where(user_id: auth_strategy.user, oauth_client: storage.oauth_client)
+                origin_user_id = RemoteIdentity.where(user_id: auth_strategy.user, integration: storage)
                                                .pick(:origin_user_id)
                 if origin_user_id.present?
                   ServiceResult.success(result: origin_user_id)
```

app/models/remote_identity.rb
ba1ash: Yes, of course, it needs to be renamed. `auth_source` works for me.

db/migrate/20250220123358_add_polymorphic_auth_source_and_integration_to_remote_identities.rb
```diff
@@ -0,0 +1,69 @@
...
+#++
+
+class AddPolymorphicAuthSourceAndIntegrationToRemoteIdentities < ActiveRecord::Migration[7.1]
+  def up
+    remove_foreign_key :remote_identities, :oauth_clients
+
+    rename_column :remote_identities, :oauth_client_id, :auth_source_id
+    add_column :remote_identities, :auth_source_type, :string
+    add_index :remote_identities, :auth_source_type
+    execute <<~SQL.squish
```
ba1ash: It is duplicated in the following update statement. Remove this one.

db/migrate/20250220123358_add_polymorphic_auth_source_and_integration_to_remote_identities.rb
```diff
@@ -0,0 +1,69 @@
...
+      UPDATE remote_identities SET auth_source_type = 'OAuthClient'
+    SQL
+
+    add_column :remote_identities, :integration_type, :string
+    add_column :remote_identities, :integration_id, :bigint
+    add_index :remote_identities, :integration_id
+    add_index :remote_identities, :integration_type
```
ba1ash: Create a composite index on type and id.

db/migrate/20250220123358_add_polymorphic_auth_source_and_integration_to_remote_identities.rb
```diff
@@ -0,0 +1,69 @@
...
+# See COPYRIGHT and LICENSE files for more details.
+#++
+
+class AddPolymorphicAuthSourceAndIntegrationToRemoteIdentities < ActiveRecord::Migration[7.1]
+  def up
+    remove_foreign_key :remote_identities, :oauth_clients
+
+    rename_column :remote_identities, :oauth_client_id, :auth_source_id
+    add_column :remote_identities, :auth_source_type, :string
+    add_index :remote_identities, :auth_source_type
```
ba1ash: Create a composite index on type and id.

modules/storages/app/models/storages/storage.rb
```diff
@@ -220,5 +223,18 @@ def health_reason_identifier
     def health_reason_description
       @health_reason_description ||= self.class.extract_part_from_piped_string(health_reason, 1)
     end
+
+    def extract_origin_user_id(token)
+      auth_strategy = ::Storages::Peripherals::Registry
```
ba1ash, line 230: Add a new strategy that will use passed to it token right away without trying to "fetch" it.

modules/openid_connect/app/services/openid_connect/user_tokens/fetch_service.rb
```diff
@@ -79,11 +82,27 @@ def access_token_for(audience:)
       def refreshed_access_token_for(audience:)
         token = yield token_with_audience(audience)
         token = yield @token_refresh.call(token)
+
+        # create_remote_identity(token, audience)
```
ba1ash: Emit an event and move creation to the event subscription.

db/migrate/20250220123358_add_oauth_client_type_to_remote_identities.rb
ba1ash: UPD: after the discussion we decided to use 2 composite indexes for new polymorphic associations.

modules/openid_connect/app/services/openid_connect/user_tokens/exchange_service.rb
ba1ash: Done. Moved to `FetchService` and decoupled using event pubsub.

app/models/remote_identity.rb
```diff
@@ -30,8 +30,9 @@
 
 class RemoteIdentity < ApplicationRecord
   belongs_to :user
-  belongs_to :oauth_client, class_name: "OAuthClient"
+  belongs_to :auth_source, polymorphic: true
+  belongs_to :integration, polymorphic: true
 
-  validates :user, uniqueness: { scope: :oauth_client }
-  validates :origin_user_id, :user, :oauth_client, presence: true
+  validates :user, uniqueness: { scope: %i[auth_source integration] }
```
NobodysNightmare, line 36: :yellow_circle:  I am really not sure what's the right thing to do here.

I am pretty sure that there will be code (possibly touched in this PR), that will _assume_ there is exactly one user per integration. Because uniqueness per storage is what we effectively had before.

Thus it would make sense to keep it that way:

```suggestion
  validates :user, uniqueness: { scope: :integration }
```

However, this would be a problem for users migrating to SSO if either of these is true:
* code relies on looking up the user with their current auth source
* the origin user id is different between SSO and the old authentication

On the other hand: Allowing to migrate has been extracted to a separate feature, so being more restrictive might be okay for now.

What do you think?

app/services/remote_identities/create_service.rb
```diff
@@ -32,35 +32,31 @@ module RemoteIdentities
...
       @result = ServiceResult.success(result: @model, errors: @model.errors)
     end
 
     def call
-      @model.origin_user_id = @oauth_config.extract_origin_user_id(@oauth_client_token)
+      @model.origin_user_id = @integration.extract_origin_user_id(@token)
       if @model.save
-        emit_event(@oauth_config.oauth_client.integration)
+        OpenProject::Notifications.send(
+          OpenProject::Events::REMOTE_IDENTITY_CREATED,
```
NobodysNightmare: :yellow_circle:  I am wondering whether this code won't be run every time that we fetch a token for the storage.

I see nothing limiting us to the `create` case here, so also all subsequent updates (possibly with no changes) will trigger the `REMOTE_IDENTITY_CREATED` event.

Am I missing something? Is this intended?

modules/openid_connect/app/services/openid_connect/user_tokens/fetch_service.rb
```diff
@@ -38,6 +38,10 @@ class FetchService
       include Dry::Monads[:result]
       include Dry::Monads::Do.for(:access_token_for, :refreshed_access_token_for)
 
+      TOKEN_OBTAINED = "access_token_obtained"
```
NobodysNightmare: :yellow_circle: I am wondering whether this should be defined on `OpenProject::Events`, because then it would read as `OpenProject::Events::TOKEN_OBTAINED` (like the other event we touched in this PR. Alternatively, I could imagine it being called `TOKEN_OBTAINED_EVENT`, which would in fully qualified form be called `OpenIDConnect::UserTokens::FetchService::TOKEN_OBTAINED_EVENT`.

modules/storages/app/common/storages/peripherals/nextcloud_registry.rb
```diff
@@ -88,6 +88,7 @@ module Peripherals
       namespace("authentication") do
         register(:userless, StorageInteraction::AuthenticationStrategies::NextcloudStrategies::UserLess, call: false)
         register(:user_bound, StorageInteraction::AuthenticationStrategies::NextcloudStrategies::UserBound)
+        register(:bearer_token, StorageInteraction::AuthenticationStrategies::NextcloudStrategies::BearerToken)
```
NobodysNightmare: :green_circle: Naming-wise I am not the biggest fan of `bearer_token` / `BearerToken` here, because the `UserBound` strategy can also be correctly described with the term "Bearer Token", because it uses the `Authorization` HTTP header with a token of type `Bearer`. It's even thinkable that a future implementation of `UserLess` would do the same.

I.e. the name doesn't quite indicate what makes the new strategy special. Proposals:

* `Passthrough`
* `SpecificBearerToken`

modules/storages/app/components/storages/project_storages/members/row_component.rb
```diff
@@ -100,7 +100,7 @@ def storage_connection_status
 
     def oauth_client_connected?
       storage.oauth_client.present? &&
-        member.principal.remote_identities.exists?(oauth_client: storage.oauth_client)
+        member.principal.remote_identities.exists?(auth_source: storage.oauth_client)
```
NobodysNightmare, line 103: I don't quite understand where this method is used (yet). I guess there is an overview showing who in a project has access to a storage?

This will not yet properly work for SSO users. Would you agree? I am happy to have this extracted into a bug/feature, but we should make this visible as to not forget it, unless we fix it immediately.

modules/storages/app/services/storages/nextcloud_managed_folder_sync_service.rb
```diff
@@ -318,15 +318,15 @@ def active_project_storages_scope
...
+      RemoteIdentity.includes(:user).where(integration: @storage, auth_source: @storage.oauth_client)
     end
 
     def auth_strategy
       @auth_strategy ||= userless.call
     end
 
     def admin_remote_identities_scope
-      RemoteIdentity.includes(:user).where(oauth_client: @storage.oauth_client, user: User.admin.active)
+      RemoteIdentity.includes(:user).where(integration: @storage, auth_source: @storage.oauth_client, user: User.admin.active)
```
NobodysNightmare: :red_circle:  I am pretty sure that this code (and the one above, line 321) will not work for SSO integrations, because they will not be findable through `auth_source: @storage.oauth_client`.

What was the motivation to introduce this additional query parameter compared to the suggestion I posted in this PR earlier?

The main weakness that I could currently see is, that there are _possibly_ two remote identities for the same user that _possibly_ result in a different `origin_user_id`.

On the other hand, in the given code this would even be the desired behaviour, because it would allow both origin user ids of the current user to have access to the corresponding files, wouldn't it? (though see my earlier comment about validating for exactly one identity per integration in the meantime)

modules/storages/app/services/storages/one_drive_managed_folder_sync_service.rb
```diff
@@ -223,11 +223,11 @@ def active_project_storages_scope
     end
 
     def client_remote_identities_scope
-      RemoteIdentity.includes(:user).where(oauth_client: @storage.oauth_client)
+      RemoteIdentity.includes(:user).where(integration: @storage, auth_source: @storage.oauth_client)
     end
 
     def admin_remote_identities_scope
-      RemoteIdentity.includes(:user).where(oauth_client: @storage.oauth_client, user: User.admin.active)
+      RemoteIdentity.includes(:user).where(integration: @storage, auth_source: @storage.oauth_client, user: User.admin.active)
```
NobodysNightmare: :red_circle: See `nextcloud_managed_folder_sync_service.rb`.

modules/storages/lib/open_project/storages/engine.rb
```diff
@@ -135,6 +135,21 @@ def self.external_file_permissions
...
+        ) do |payload|
+          audience = payload[:audience]
+          token = payload[:token]
+          storage = Storages::Storage.all.find { |s| s.audience == audience }
+          if storage
+            RemoteIdentities::CreateService
+              .call(user: token.user, integration: storage, token:)
+              .on_failure { raise "RemoteIdentity creation failed" }
+          else
+            puts "WARNING: no integration was found for audience: #{audience}"
```
NobodysNightmare: :green_circle: I'd leave this debug log out to be honest. Going forward it will be normal to see tokens being obtained for things that are not storages. We shouldn't expect that every exchange happens for a storage.

modules/storages/lib/open_project/storages/engine.rb
```diff
@@ -135,6 +135,21 @@ def self.external_file_permissions
         ) do |payload|
           ::Storages::HealthService.new(storage: payload[:storage]).healthy
         end
+
+        OpenProject::Notifications.subscribe(
+          ::OpenIDConnect::UserTokens::FetchService::TOKEN_OBTAINED
+        ) do |payload|
+          audience = payload[:audience]
+          token = payload[:token]
+          storage = Storages::Storage.all.find { |s| s.audience == audience }
```
NobodysNightmare: :green_circle: The number of storages will usually be low, so I don't care too much that we fetch all of them, just to be able to validate their audience. I guess the main problem is, that right now there is no audience "column" for one drive storages and thus we can't filter this on the database level?

modules/storages/spec/factories/storage_factory.rb
```diff
@@ -221,8 +222,11 @@
              refresh_token: ENV.fetch("ONE_DRIVE_TEST_OAUTH_CLIENT_REFRESH_TOKEN",
                                       "MISSING_ONE_DRIVE_TEST_OAUTH_CLIENT_REFRESH_TOKEN"),
              token_type: "bearer")
-      create(:remote_identity, oauth_client: storage.oauth_client, user: evaluator.oauth_client_token_user,
-                               origin_user_id: "33db2c84-275d-46af-afb0-c26eb786b194")
+      create(:remote_identity,
+             auth_source: storage.oauth_client,
+             user: evaluator.oauth_client_token_user,
+             integration: storage,
+             origin_user_id: "33db2c84-275d-46af-afb0-c26eb786b194")
```
NobodysNightmare: :yellow_circle: Shouldn't we be using changing values here? Otherwise all remote identities created through the factory would have colliding user names by default.

```suggestion
             origin_user_id: SecureRandom.uuid)
```

Alternatively: If there is a good reason to always use the same one, I'd go with the @mereghost approach and call it `UNIVERSALLY-DUPLICATED-IDENTIFIER`. This at least makes it transparent, that we are always dealing with the same thing and it's 5% funny ;-)

Alternatively: Just don't specify anything, because the factory for remote identities already takes care of generating sequential origin user ids.

NobodysNightmare (changes requested):

modules/storages/app/common/storages/peripherals/storage_interaction/nextcloud/util.rb
```diff
@@ -71,7 +71,8 @@ def origin_user_id(caller:, storage:, auth_strategy:)
               when :basic_auth
                 ServiceResult.success(result: storage.username)
               when :oauth_user_token, :sso_user_token
-                origin_user_id = RemoteIdentity.where(user_id: auth_strategy.user, oauth_client: storage.oauth_client)
+                origin_user_id = RemoteIdentity.where(user_id: auth_strategy.user, auth_source: storage.oauth_client,
```
NobodysNightmare: :red_circle: This will also not fly for `sso_user_token` (which is part of this `when`), because authentication through SSO will have a different `auth_source` than `storage.oauth_client`.

modules/storages/app/components/storages/project_storages/members/row_component.rb
NobodysNightmare: I just looked into this in the context of my PR. I am going to tackle it there.

modules/openid_connect/app/services/openid_connect/user_tokens/fetch_service.rb
ba1ash: I avoided `OpenProject::Events`, because this event is produced from within the oidc module. Let's go with `TOKEN_OBTAINED_EVENT`.

modules/storages/app/common/storages/peripherals/nextcloud_registry.rb
ba1ash: I'd go with `SpecificBearerToken` then. Thanks for suggesting!

modules/storages/app/components/storages/project_storages/members/row_component.rb
ba1ash: Yes, it does not work properly for SSO users at the moment.
The method is used on project storage members page. The page is used to give a members overview which show who have completed OAuth flow and who have not. The page can be found here:  https://openproject.local/projects/ppproject/settings/project_storages/6/members.
I filed a bug: https://community.openproject.org/projects/cross-application-user-integration-stream/work_packages/61880/activity

modules/storages/app/services/storages/nextcloud_managed_folder_sync_service.rb
ba1ash: You're absolutely right. I was not paying enough attention to the above snippet, my bad. The `auth_source` clause is obsolete here.

modules/storages/app/services/storages/one_drive_managed_folder_sync_service.rb
ba1ash: Yes, the same.

modules/storages/lib/open_project/storages/engine.rb
ba1ash: I did not do that, because at the moment of writing the field was called `nextcloud_audience` and migration was happening to `storage_audience`. We can filter through database now.

modules/storages/spec/factories/storage_factory.rb
ba1ash: This origin_user_id is used in a couple of vcr cassetes. I will try to move it specifically to the specs concerned with it.

app/services/remote_identities/create_service.rb
ba1ash: Yes, this code will run on every token fetch.  I do not see a reason for emitting an event if there is no changes to existing model. Would you agree?

app/models/remote_identity.rb
ba1ash: I think we should go with `auth_source integration` scope. I know one place to update where we should be explicit about auth_source.  That is `modules/storages/app/common/storages/peripherals/storage_interaction/nextcloud/util.rb`. I suggest there we have something like:
``` ruby
              when :oauth_user_token, :sso_user_token
                auth_source = if auth_strategy.key == :oauth_user_token
                                storage.oauth_client
                              else
                                user.authentication_provider
                              end
                origin_user_id = RemoteIdentity.where(user_id: auth_strategy.user,
                                                      auth_source:,
                                                      integration: storage)
                                               .pick(:origin_user_id)
                if origin_user_id.present?
                  ServiceResult.success(result: origin_user_id)
                else
                  failure(code: :error,
                          data: StorageErrorData.new(source: caller),
                          log_message:
                            "No origin user ID or user token found. Cannot execute query without user context.")
                end


```
Other places should be fine with new constraint. But maybe I am missing something.

modules/storages/app/common/storages/peripherals/storage_interaction/nextcloud/util.rb
ba1ash: Yes, that is right. I suggested above(in the uniqueness thread) to have two separate branches here with explicit auth_source set. Wdyt?

app/services/remote_identities/create_service.rb
NobodysNightmare: Exactly. I think under normal circumstances the event should be fired once upon creation.

app/models/remote_identity.rb
NobodysNightmare: Essentially I didn't think enough about edge cases right now, and I'd be happy to delay that.

But on a general level:
* `uniqueness: { scope: %i[auth_source integration] }` means that it's possible that `user.remote_identities.where(storage:).count > 1`
* This means there could be two different `origin_user_id` values for the same user at a storage (user.remote_identities.where(storage:).pluck(:origin_user_id).uniq.count > 1`)
* This means if we have code like `RemoteIdentity.where(user:, integration:).pick(:origin_user_id)` it could pick the wrong user id

However, you already properly addressed the one case I know of in your example above, by making sure we pick the correct `auth_source`. And other places I see in this PR actually work correctly by iterating over all identities, regardless of the auth source. So it's probably fine I guess?

ba1ash: That's my understanding as well.

modules/storages/app/common/storages/peripherals/storage_interaction/nextcloud/util.rb
NobodysNightmare: Yep, code suggested there will fix this. Alternatively (I didn't check what it would look like) we could also have two `when` cases, one for SSO and one for `oauth_user_token`. I am not sure what would be cleaner in the end.

modules/storages/app/services/storages/one_drive_managed_folder_sync_service.rb
NobodysNightmare: We still need to drop the `auth_source` here. Only `integration` and `user` are enough to identify which users need to be synced.

NobodysNightmare (approved):
Except for one simple but forgotten review comment (I added a reminder) and Rubocop being unhappy about the complexity of one method we made bigger, this LGTM.

modules/storages/spec/common/storages/peripherals/storage_interaction/nextcloud/util_spec.rb
```diff
@@ -0,0 +1,57 @@
...
+RSpec.describe(Storages::Peripherals::StorageInteraction::Nextcloud::Util) do
+  describe ".origin_user_id" do
+    it "responds with appropriate origin_user_id when user has two remote identities" \
+       "for one integration within two different auth_sources" do
+      oauth_client = create(:oauth_client)
+      oidc_provider = create(:oidc_provider)
+      integration = create(:nextcloud_storage, oauth_client:)
+      user = create(:user, identity_url: "#{oidc_provider.slug}:123123213")
+      create(:remote_identity, user:, auth_source: oauth_client, integration:, origin_user_id: "456")
+      create(:remote_identity, user:, auth_source: oidc_provider, integration:, origin_user_id: "123")
```
NobodysNightmare, line 43: :red_circle:  I had a flickering spec on my branch (which is based on this PR) that possibly uncovered an error in our data layer: https://github.com/opf/openproject/actions/runs/13626909831/job/38085999789?pr=18111

The spec failed with

```
duplicate key value violates unique constraint "index_remote_identities_on_user_id_and_auth_source_id"
```

I didn't realize this at first, but it means that we apparently have a unique index over `user_id` and `auth_source_id`. This means if the same user has remote identities for two auth sources **and** the auth sources happen to have the same ID, the index will conflict. The index should rather be built on `(user_id, auth_source_type, auth_source_id)`.

edit: Probably the last index you add in the migration already fulfills that job. So did we just forget to delete an old index?

edit 2: Thinking further, this index would also conflict if we had two integrations using the same SSO provider. There should probably be an explicit test case for that somewhere...

edit 3: I was surprised that this shows up on my branch but not on yours... but part of my changes is, that more test cases now create a `RemoteIdentity`. So I guess it's just more likely on my branch that such a conflict would occur.

ba1ash: Good catch. It is a leftover. It should have been removed by me, but I lost it during a couple of migrate/rollback cycles. I removed it in my next commit. There is a new uniq index over 5 columns: `user_id auth_source_type auth_source_id integration_id integration_type`.

NobodysNightmare: Specs on my PR succeeded after rebasing on top of the latest changes :tada:

NobodysNightmare (approved):

PR #18034, opened 2025-02-21
[#61465] users in lists miss popover additional fixes

EinLama:
# Ticket

https://community.openproject.org/wp/61465

Additional code changes for the CKEditor can be found in [commonmark-ckeditor-build#98](https://github.com/opf/commonmark-ckeditor-build/pull/98)

Fix occurrences where a hover card was expected, but did not show up:

* in the team planner, when adding a new assignee (within the auto completer)
* when adding a user board list
* when mentioning users within CKEditor (dropdown selection while typing)
* when mentioning work packages within CKEditor (dropdown selection while typing)

Additionally fix two issues:

* the user hover card was accidentally rendered when you hover over your own avatar in the top right
* when you hover over the avatar within another hover card, the system attempted to render another hover card on top. Since that is not possible, the existing hover card would close.

While I was at it, I fixed the way that hover card options are handled within the principal renderer. Hover cards have their own options-object now to configure whether they show up or not. Previously, this was crammed into the AvatarOptions. Since hover cards also render on usernames, it felt wrong to leave it like this.

This also allowed me to adjust the default values so that the additional two fixes from above were possible.

The work package lists two other occurrences where hover cards are missing:

1. time and costs user filter list
2. "move to project" user lists

Both of these were NOT added since the mentioned forms use native select elements. We cannot add popovers there without a major rework of the form.

## Screenshots

https://github.com/user-attachments/assets/e25d020b-3355-4076-aaaf-7b5b298e1ca6

https://github.com/user-attachments/assets/bb730ea2-f244-4162-9ae9-13eaf5634102

https://github.com/user-attachments/assets/f2b5d3cc-3eb4-4880-b527-688bd04cfed5

# Merge checklist

- [x] Added/updated tests


bsatarnejad (approved):
LGTM 👍

PR #18037, opened 2025-02-21
Fix encoding issue with client id and secret

NobodysNightmare:
We errorneously did not encode the client id and secret. According to the HTTP Basic Auth spec, certain characters such as colon (:) are simply not allowed as part of the username.

The OAuth 2.0 spec already considered this, by requiring client id and secret to be URL form-encoded, but we forgot to do that.

Since this would have required changes in two places that already looked very similar, the requests to the token endpoint have been extracted into a service class, where they can share the basic code to construct the request and we can centrally test behaviours, such as authentication.

# Ticket
https://community.openproject.org/projects/cross-application-user-integration-stream/work_packages/61694/activity

# Merge checklist

- [x] Added/updated tests
- [x] Tested changes manually


ba1ash (approved):
:+1:

PR #18040, opened 2025-02-21
[#61315] fix tokens for work packages

Kharonus:
# Ticket
[OP#61315](https://community.openproject.org/wp/61315)

- subject pattern input now provides all configured custom fields of the work package type

- fix relation for work package only cfs

### DIFF
--- a/app/models/types/patterns/token_property_mapper.rb
+++ b/app/models/types/patterns/token_property_mapper.rb
@@ -103,7 +103,7 @@ module Types
       end
 
       def work_package_cfs_for(type)
-        all_work_package_cfs.where(type: type)
+        all_work_package_cfs.merge(type.custom_fields)
       end
 
       def all_work_package_cfs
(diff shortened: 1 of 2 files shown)


NobodysNightmare (commented):
The added test makes me confident that the fix works as intended. But to be honest I have a hard time understanding why your fix fixes.

Going for an attempt. Do I see it correctly, that it is not enough to look for custom fields with the corresponding type for the work package, but rather we need to check the join table between type and custom fields (which we do via `merge(type.custom_fields`)?

But then, what does `custom_fields.type` even _mean_?

Kharonus:
Yeah, I think you tapped into the same trap, as Marcello did when implementing it. And it is no shame in that :P

point is, that the `type` column of a custom field does NOT include any reference to a work package type, but to a custom field type, which is one of `WorkPackageCustomField`, `UserCustomField`, and so on. So, the actual relation we want to have is the combination of two things: All available custom fields - which contain filtering based on custom field type - and all custom fields enabled for the specific work package type, which you get by `work_package_type.custom_fields`.

Does this suffice as an explanation?

NobodysNightmare:
Doh! Of course, in hindsight it's obvious that `type` is the single-table-inheritance type column xD

Thanks!

NobodysNightmare (approved):

PR #18042, opened 2025-02-21
[#61165] fixed separator spacing in pattern input

Kharonus:
# Ticket
[OP#61165](https://community.openproject.org/wp/61165)

- fix spacing of separators

- add system argument to custom component

### before

image

### after

image


Kharonus:
@HDinger please let me know, if you agree to the fix by using a system argument here.

We already talked about, that action list is not the correct component for the pattern input. So, my thought was, the customization for the spacing should happen here, where the action list is rendered - and now customized.

brunopagno (approved):
Honestly this looks good to me. As simple as can be :+1:

app/components/work_packages/types/pattern_input.rb
```diff
@@ -49,6 +49,7 @@ def suggestions_for_stimulus
       def suggestions_list_component
         @suggestions_list_component ||= Primer::Alpha::ActionList.new(
           role: :list,
+          classes: "mr-3",
```
HDinger: Please do not use the helper classes generated by the system arguments directly. Primer might change that, and it will not work any more. Please use the system arguments instead.

```suggestion
          mr: 3
```

HDinger (changes requested):
I have some remarks:

* The spacing on the right is smaller than the spacing to the left. Don't know whether this is accepted, but to me, it looked off.

Bildschirmfoto 2025-02-24 um 07 51 41

* The ticket is also about the width of the input. I assume that this is explicitly ignored for this PR, but will it be tackled afterwards?

Kharonus:
@HDinger those are brilliant questions :D

and, TBH, I though our meeting today will precede a review of you for this. The spacing is part of a short discussion about the composition of forms inside the bigger application.

Regarding the left side margin: there is an additional margin inherited from the usage of `ul` for the suggestions list. It comes from `_base.sass` and is 2rem. I didn't want to override it. And the left side margin was there before. 

A simple fix would be to use on the right side the same as on the left side. For my first shot, I simply took the margin we have for examples in the [lookbook](https://qa.openproject-edge.com/lookbook/inspect/primer/alpha/action_list/with_manual_dividers). Here you can see, that we also have this big margin on the left, while the right has none. Only the wrapping contianer covers the overflowing dividers.

So, another approach next to my fix would be to set a 1rem margin on a surrounding container of our suggestion list. Or, we could override `ml` in the same way :/

Kharonus:
Reworked the styling after talking to Marc slightly:

separator04

app/components/work_packages/types/pattern_input.rb
```diff
@@ -31,6 +31,8 @@
 module WorkPackages
   module Types
     class PatternInput < Primer::Forms::BaseComponent
+      prepend Primer::OpenProject::Forms::WrappedInput
```
Kharonus, line 34: 💡 Added wrapped input

app/components/work_packages/types/pattern_input.sass
```diff
@@ -2,3 +2,8 @@
   .input
     white-space: pre-wrap
 
+    &:focus,
+    &:focus-visible
+      border-color: var(--focus-outlineColor) !important
+      box-shadow: inset 0 0 0 1px var(--focus-outlineColor)
+      outline: none
```
Kharonus, line 9: 💡 This component used the spot outline color. I had to override the styles so that is uses the primer border recoloring.

🔴 This is a bit dangerous, as when the primer style changes, this one will be forgotten. How can I make it work to simply use primer border recoloring?

HDinger: Are you sure about the spot styles? Looking at the DOM, I can't see any spot related focus styles.. THe bright blue is the browser default highlighting.. 

As to the question: I thin, there is no way to achieve that via CSS (at least not with the options that we have). The only solution that comes to my mind now is to make the input a real Primer TextField with automatically gets all the classes and thus sets the correct styles. But I don't know which sideeffects that will have on the actual component.

brunopagno: I don't think we can make it an input because of the approach with `contenteditable`. Unless we can use a "div" tag with the input styles.

Kharonus: positive, I'm sure. The bright blue outline comes from:

```css
:focus-visible {
    outline: 2px solid rgb(0, 163, 255);
    outline-offset: -2px;
}
```

this is defined in [focus.sass](https://github.com/opf/openproject/blob/dev/frontend/src/app/spot/styles/sass/variables/focus.sass).

HDinger: Ah, good catch!

HDinger: I think given all of these information, we have to live with that. To be fair, we are using a lot of Primer variables already (and we usually encourage devs to use them in the sass file). 
So we have to do a sorrow check anyway when Primer updates their variables.

Kharonus:
So, what is the overall opinion? looks good enough as a fix?

frontend/src/stimulus/controllers/pattern-input.controller.ts
```diff
@@ -72,6 +72,7 @@ export default class PatternInputController extends Controller {
     this.contentTarget.innerHTML = this.toHtml(this.patternInitialValue) || ' ';
     this.extractValidTokens();
     this.tagInvalidTokens();
+    this.clearSuggestionsFilter();
```
brunopagno, line 75: :+1:

brunopagno (approved):
Looks absolutely briliant to me :clap:

HDinger (approved):
Lgtm 👍 Thanks for investing in that 🙇

PR #18053, opened 2025-02-24
bug/61708 Cannot delete users who used emoji reactions

akabiru:
https://community.openproject.org/work_packages/61708

Add missing active record user association with dependent destroy

### DIFF
--- a/spec/models/user_spec.rb
+++ b/spec/models/user_spec.rb
@@ -47,6 +47,10 @@ RSpec.describe User do
           status:)
   end
 
+  describe "Associations" do
+    it { is_expected.to have_many(:emoji_reactions).dependent(:destroy) }
+  end
+
   describe "with long but allowed attributes" do
     it "is valid" do
       user.firstname = "a" * 256
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -91,6 +91,7 @@ class User < Principal
            inverse_of: :user,
            dependent: :destroy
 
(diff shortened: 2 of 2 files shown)


klaustopher (approved):

spec/models/user_spec.rb
```diff
@@ -47,6 +47,10 @@
           status:)
   end
 
+  describe "Associations" do
+    it { is_expected.to have_many(:emoji_reactions).dependent(:destroy) }
```
akabiru, line 51: I only scoped to the current bug- but I'd like to expand these unit tests for other associations (separately)

myabc:
@akabiru N.B. according to Active Record Doctor (see PR #18075), `dependent: :delete_all` would suffice:

```
use `dependent: :delete_all` or similar on User.emoji_reactions - associated model EmojiReaction has no callbacks and can be deleted in bulk
```

I don't know if we want to revise this?

akabiru:
@myabc Sounds sensible with active record doctor integrated. Otherwise I'd worry about missed callbacks- which ar-doctor should take care of afaict.

Perhaps we run the ar-doctor proposals in one PR?

myabc:
> Perhaps we run the ar-doctor proposals in one PR?

@akabiru Yeah, that sounds like a good plan. Probably best as a follow up to #18075 (WP [#61776](https://community.openproject.org/wp/61776)), but I'll see what @machisuji thinks!

PR #18057, opened 2025-02-24
Make a few more files ready to handle frozen string literals

NobodysNightmare:
Affected files were previously modifying string literals in-place and have now been changed to either ensure they are declared as mutable or avoid modifying them at all.

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -475,8 +475,6 @@ module ApplicationHelper
 
   def password_complexity_requirements
     rules = OpenProject::Passwords::Evaluator.rules_description
-    # use 0..0, so this doesn't fail if rules is an empty string
-    rules[0] = rules[0..0].upcase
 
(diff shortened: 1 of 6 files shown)


app/helpers/application_helper.rb
```diff
@@ -469,8 +469,6 @@ def link_to_content_update(text, url_params = {}, html_options = {})
 
   def password_complexity_requirements
     rules = OpenProject::Passwords::Evaluator.rules_description
-    # use 0..0, so this doesn't fail if rules is an empty string
-    rules[0] = rules[0..0].upcase
```
NobodysNightmare, line 479: This in-place modification seems to have been used to title-ize the `rules_description`. However, `rules_description` already returns a string that originates from `I18n`, so all necessary titleization should already have been applied on the locale-level.

NobodysNightmare:
(I am absolutely going to ignore Rubocop complaints; all of them are about things that were not broken in this PR)

ba1ash (approved):

PR #18059, opened 2025-02-24
[#61619] unclear error message upon save on empty subject patterns

Kharonus:
# Ticket
[OP#61619](https://community.openproject.org/wp/61619)

- removed error message banners
- fixed controller and params when calling updates
- renamed subject configuration controller to emphasize usage in tab

### DIFF
--- a/.erb_lint.yml
+++ b/.erb_lint.yml
@@ -39,3 +39,5 @@ linters:
         Enabled: true
       Style/FrozenStringLiteralComment:
         Enabled: false
+      Style/RedundantConstantBase:
+        Enabled: false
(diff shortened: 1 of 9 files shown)


NobodysNightmare (approved):

NobodysNightmare:
> removed error message banners

It took me a moment to realize why removing the banners was the right thing to do. But we already show a nice error message. I just overlooked it, as long as the banner was there.

Kharonus:
Indeed, QA didn't overlook it though. :D

Yes, if we have full inline form validation I see no reason to maintain an error banner. Once we have errors, that are not bound to an input field, we can readd logic to show them.

Kharonus:
the failing test fail on `dev`, too, and are considered unrelated.

PR #18063, opened 2025-02-24
Implements the handling of token acquiral and refresh results.

mereghost:
### Related WorkPackage: [OP#61608](https://community.openproject.org/projects/cross-application-user-integration-stream/work_packages/61608)

This PR adds a check over the token usability. It uses the `FetchService` to get a usable token.

It handles the most common error states from that service.

### DIFF
--- a/modules/openid_connect/app/models/openid_connect/provider.rb
+++ b/modules/openid_connect/app/models/openid_connect/provider.rb
@@ -1,5 +1,33 @@
 # frozen_string_literal: true
 
+#-- copyright
+# OpenProject is an open source project management software.
+# Copyright (C) the OpenProject GmbH
+#
+# This program is free software; you can redistribute it and/or
+# modify it under the terms of the GNU General Public License version 3.
+#
+# OpenProject is a fork of ChiliProject, which is a fork of Redmine. The copyright follows:
+# Copyright (C) 2006-2013 Jean-Philippe Lang
+# Copyright (C) 2010-2013 the ChiliProject Team
+#
+# This program is free software; you can redistribute it and/or
+# modify it under the terms of the GNU General Public License
+# as published by the Free Software Foundation; either version 2
+# of the License, or (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program; if not, write to the Free Software
+# Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+#
+# See COPYRIGHT and LICENSE files for more details.
+#++
+
 module OpenIDConnect
   class Provider < AuthProvider
     include HashBuilder
--- a/modules/openid_connect/app/models/openid_connect/provider.rb
+++ b/modules/openid_connect/app/models/openid_connect/provider.rb
@@ -6,6 +6,8 @@ module OpenIDConnect
 
     has_many :remote_identities, as: :auth_source, dependent: :destroy
 
+    TOKEN_EXCHANGE_GRANT_TYPE = "urn:ietf:params:oauth:grant-type:token-exchange"
+
     OIDC_PROVIDERS = %w[google microsoft_entra custom].freeze
     DISCOVERABLE_STRING_ATTRIBUTES_MANDATORY = %i[authorization_endpoint
                                                   userinfo_endpoint
@@ -23,6 +25,7 @@ module OpenIDConnect
     DISCOVERABLE_STRING_ATTRIBUTES_ALL.each do |attribute|
       store_attribute :options, attribute, :string
     end
+
     MAPPABLE_ATTRIBUTES.each do |attribute|
       store_attribute :options, "mapping_#{attribute}", :string
     end
@@ -87,7 +90,7 @@ module OpenIDConnect
     def token_exchange_capable?
       return false if grant_types_supported.blank?
 
-      grant_types_supported.include?("urn:ietf:params:oauth:grant-type:token-exchange")
+      grant_types_supported.include?(TOKEN_EXCHANGE_GRANT_TYPE)
     end
 
     def icon
--- a/modules/openid_connect/app/services/openid_connect/user_tokens/exchange_service.rb
+++ b/modules/openid_connect/app/services/openid_connect/user_tokens/exchange_service.rb
@@ -60,7 +60,7 @@ module OpenIDConnect
         json = yield exchange_token_request(idp_token, audience)
 
         access_token, expires_in = json.values_at("access_token", "expires_in")
-        Failure(@error.with(code: :token_exchange_response_invalid, payload: json)) if access_token.blank?
+        return Failure(@error.with(code: :token_exchange_response_invalid, payload: json)) if access_token.blank?
 
         # We are explicitly opting to not store the refresh token for exchanged tokens
         # For one there is no need to store one, we can simply exchange a new token once the old expired.
@@ -82,7 +82,9 @@ module OpenIDConnect
       end
 
       def exchange_token_request(idp_token, audience)
-        TokenRequest.new(provider:).exchange(idp_token, audience)
+        TokenRequest.new(provider:).exchange(idp_token, audience).alt_map do
+          it.with(code: :"token_exchange_#{it.code}", source: self.class)
+        end
       end
 
       def store_exchanged_token(audience:, access_token:, refresh_token:, expires_in:)
--- a/modules/openid_connect/app/services/openid_connect/user_tokens/exchange_service.rb
+++ b/modules/openid_connect/app/services/openid_connect/user_tokens/exchange_service.rb
@@ -31,7 +31,7 @@
 module OpenIDConnect
   module UserTokens
     class ExchangeService
-      include Dry::Monads[:result]
+      include Dry::Monads::Result(TokenOperationError)
       include Dry::Monads::Do.for(:call)
 
       class Disabled
@@ -48,19 +48,19 @@ module OpenIDConnect
 
       def initialize(user:)
         @user = user
+        @error = TokenOperationError.new(source: self.class)
       end
 
       def call(audience)
-        return Failure(:token_exchange_not_supported) unless supported?
+        unless supported?
+          return Failure(@error.with(code: :token_exchange_not_supported, payload: provider&.grant_types_supported))
+        end
 
-        idp_token = yield FetchService.new(user:, token_exchange: Disabled)
-                                      .access_token_for(audience: UserToken::IDP_AUDIENCE)
+        idp_token = yield fetch_openproject_token
+        json = yield exchange_token_request(idp_token, audience)
 
-        json = yield request_exchange(idp_token, audience)
-
-        access_token = json["access_token"]
-        expires_in = json["expires_in"]
-        return Failure(:token_exchange_response_invalid) if access_token.blank?
+        access_token, expires_in = json.values_at("access_token", "expires_in")
+        Failure(@error.with(code: :token_exchange_response_invalid, payload: json)) if access_token.blank?
 
         # We are explicitly opting to not store the refresh token for exchanged tokens
         # For one there is no need to store one, we can simply exchange a new token once the old expired.
@@ -77,21 +77,24 @@ module OpenIDConnect
 
       private
 
-      def request_exchange(idp_token, audience)
+      def fetch_openproject_token
+        FetchService.new(user: @user, token_exchange: Disabled).access_token_for(audience: UserToken::IDP_AUDIENCE)
+      end
+
+      def exchange_token_request(idp_token, audience)
         TokenRequest.new(provider:).exchange(idp_token, audience)
       end
 
       def store_exchanged_token(audience:, access_token:, refresh_token:, expires_in:)
         token_data = { access_token:, refresh_token:, expires_at: expires_in&.seconds&.from_now }
         token = user.oidc_user_tokens.where("audiences ? :audience", audience:).first
-        if token
-          if token.audiences.size > 1
-            raise "Did not expect to update token with multiple audiences (#{token.audiences}) in-place."
-          end
 
-          token.update!(**token_data)
-        else
+        if token.nil?
           token = user.oidc_user_tokens.create!(audiences: [audience], **token_data)
+        elsif token.audiences.size > 1
+          raise "Did not expect to update token with multiple audiences (#{token.audiences}) in-place."
+        else
+          token.update!(**token_data)
         end
 
         token
--- a/modules/openid_connect/app/services/openid_connect/user_tokens/exchange_service.rb
+++ b/modules/openid_connect/app/services/openid_connect/user_tokens/exchange_service.rb
@@ -2,7 +2,7 @@
 
 #-- copyright
 # OpenProject is an open source project management software.
-# Copyright (C) 2012-2024 the OpenProject GmbH
+# Copyright (C) the OpenProject GmbH
 #
 # This program is free software; you can redistribute it and/or
 # modify it under the terms of the GNU General Public License version 3.
@@ -78,10 +78,7 @@ module OpenIDConnect
       private
 
       def request_exchange(idp_token, audience)
-        TokenRequest.new(provider:).exchange(idp_token, audience).or do |error|
-          Rails.logger.error("Failed to refresh token: #{error.inspect}")
-          Failure(:failed_token_exchange)
-        end
+        TokenRequest.new(provider:).exchange(idp_token, audience)
       end
 
       def store_exchanged_token(audience:, access_token:, refresh_token:, expires_in:)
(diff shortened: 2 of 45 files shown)


modules/openid_connect/app/models/openid_connect/provider.rb
```diff
@@ -1,7 +1,11 @@
+# frozen_string_literal: true
```
NobodysNightmare, line 1: :green_circle:  On the topic of missing comments at the start of the file: You might want to add a copyright header as well :)

modules/openid_connect/app/services/openid_connect/user_tokens/exchange_service.rb
```diff
@@ -49,16 +49,16 @@ def initialize(user:)
...
-                            .access_token_for(audience: UserToken::IDP_AUDIENCE)
+                                      .access_token_for(audience: UserToken::IDP_AUDIENCE)
 
-        json = yield TokenRequest.new(provider:).exchange(idp_token, audience)
+        json = yield request_exchange(idp_token, audience)
 
         access_token = json["access_token"]
         expires_in = json["expires_in"]
-        return Failure("Token exchange response invalid") if access_token.blank?
+        return Failure(:token_exchange_response_invalid) if access_token.blank?
```
NobodysNightmare: :yellow_circle: I am surprised that the `RefreshService` didn't get the same makeover w.r.t. this error code.

I noticed, because I wanted to ask the open question on whether we want to specifically separate the error codes for both cases (`token_exchange_response_invalid` vs. `token_refresh_response_invalid`) or whether we instead would want to unify them under `token_response_invalid` or `invalid_token_response`.

modules/openid_connect/app/services/openid_connect/user_tokens/exchange_service.rb
```diff
@@ -74,6 +74,13 @@ def supported?
 
       private
 
+      def request_exchange(idp_token, audience)
+        TokenRequest.new(provider:).exchange(idp_token, audience).or do |error|
+          Rails.logger.error("Failed to refresh token: #{error.inspect}")
+          Failure(:failed_token_exchange)
```
NobodysNightmare: :yellow_circle: As far as I can tell, what this change does, is making sure that we don't return `HTTPX` errors, but always symbols instead. I need to dig deeper into this, but an interesting side effect could be that we are now losing compatibility with [this existing code](https://github.com/opf/openproject/blob/dev/modules/storages/app/common/storages/peripherals/storage_interaction/authentication.rb#L69). Though compatibility was only partial so far, since our string errors would also have failed there.

I am not saying we should not do this change here, but _maybe_ we want to change more? (possibly in my PR)

(I only recently learned about this code, while working on my error handling ticket)

modules/openid_connect/app/services/openid_connect/user_tokens/refresh_service.rb
```diff
@@ -70,7 +70,10 @@ def exchange_instead_of_refresh(token)
       end
 
       def refresh_token_request(refresh_token)
```
NobodysNightmare, line 70: :green_circle: Do we want to rename this method to `request_refresh` or alternatively rename the newly introduced `request_exchange` to `exchange_token_request`? I like symmetry :D

modules/openid_connect/spec/services/openid_connect/user_tokens/token_request_spec.rb
```diff
@@ -120,7 +120,7 @@
     it "uses a properly formatted request body" do
       subject
       expect(WebMock).to have_requested(:post, provider.token_endpoint)
-        .with(body: { grant_type: "urn:ietf:params:oauth:grant-type:token-exchange", subject_token: token, audience: })
+        .with(body: { grant_type: OpenIDConnect::Provider::TOKEN_EXCHANGE_GRANT_TYPE, subject_token: token, audience: })
```
NobodysNightmare, line 124: :green_heart:  Looking at this diff (and the ones before) I suddenly realize, that this single string was used quite a bunch of times across the codebase already. Thanks for cleaning this up.

I think the name is well chosen, but still I am sad that we managed our constant name to be longer than the original - very unwieldy - string.

modules/storages/app/common/storages/peripherals/nextcloud_connection_validator.rb
```diff
@@ -70,29 +72,40 @@ def non_provisioned_user
...
+        error_code = case result.failure
+                     when :failed_refresh
+                       :oidc_cant_refresh_token
+                     when :failed_token_exchange, :token_exchange_response_invalid
+                       :oidc_cant_exchange_token
+                     else
+                       :unknown_error
+                     end
 
         # Temporary error as I implement the next steps
```
NobodysNightmare: :yellow_circle: Still temporary?

modules/storages/config/locales/en.yml
```diff
@@ -243,6 +243,8 @@ en:
         host_not_found: No Nextcloud server found at the configured host url. Please check the configuration.
         missing_dependencies: 'A required dependency is missing on the file storage. Please add the following dependency: %{dependency}.'
         not_configured: The connection could not be validated. Please finish configuration first.
+        oidc_cant_exchange_token: There seems to be problem with the Token Exchange setup on your OpenID Connect Provider. Please check its configuration and try again.
+        oidc_cant_refresh_token: There was an error while trying to check your access to the storage. Please check the server logs for further information.
         oidc_no_token_with_required_audience: There's no OpenID Connect token with the required %{audience} audience.
```
NobodysNightmare: :yellow_circle: Is this locale still being used? I see that we removed a hardcoded reference to it, while not adding a case that would still use an error code of this name.

If not: We still call `I18n` with the `audience` parameter, which is then also not needed anymore.

modules/storages/app/common/storages/peripherals/nextcloud_connection_validator.rb
```diff
@@ -70,29 +72,40 @@ def non_provisioned_user
...
       def token_usability
-        usable_token = @user.oidc_user_tokens.with_audience(@storage.audience).first
-        return None() if usable_token
+        service = OpenIDConnect::UserTokens::FetchService.new(user: @user)
+
+        result = service.access_token_for(audience: @storage.audience)
+        return None() if result.success?
+
+        error_code = case result.failure
+                     when :failed_refresh
```
NobodysNightmare: Comparing this `when` to the following `when`, I think the answer to my earlier question is: We should introduce a `token_refresh_response_invalid` error code as well.

NobodysNightmare (changes requested):
I think I like the direction in which we develop our code here.

As mentioned in one of the comments, it probably makes sense to apply the same pattern (masking HTTPX errors) in other places of the existing storages code as well, though that's out of scope for your PR, just a direction that I now see.

modules/storages/config/locales/en.yml
```diff
@@ -243,6 +243,8 @@ en:
         host_not_found: No Nextcloud server found at the configured host url. Please check the configuration.
         missing_dependencies: 'A required dependency is missing on the file storage. Please add the following dependency: %{dependency}.'
         not_configured: The connection could not be validated. Please finish configuration first.
+        oidc_cant_exchange_token: There seems to be problem with the Token Exchange setup on your OpenID Connect Provider. Please check its configuration and try again.
```
NobodysNightmare:
```suggestion
        oidc_cant_exchange_token: There seems to be a problem with the Token Exchange setup on your OpenID Connect Provider. Please check its configuration and try again.
```

NobodysNightmare: We should possibly also hint at the server logs here...

modules/openid_connect/app/models/openid_connect/provider.rb
mereghost: Oh... missed that. Will do!

modules/openid_connect/app/services/openid_connect/user_tokens/exchange_service.rb
mereghost: Yeah... I got conflicted on if I should address it now or on a later PR. =/

As for the error, I think we could return different error and let the caller reason what it should send the user.

modules/openid_connect/app/services/openid_connect/user_tokens/exchange_service.rb
mereghost: Hmmmm.... Cool, I can adapt it to make it compatible.

The thing is: which statuses are we expect to get from the TokenRequest? 200, 401, 403?

modules/storages/config/locales/en.yml
mereghost: I don´t think it is. I had deleted it, but it probably came back during one of the merges. >.<

modules/storages/app/common/storages/peripherals/nextcloud_connection_validator.rb
mereghost: Temporarily permanent xD Good catch.

modules/openid_connect/app/services/openid_connect/user_tokens/exchange_service.rb
NobodysNightmare: Good question. I think right now it is 200, 401 and others (unexpected). Though that differentiation is very obviously arbitrary.

Would be fine for me to fix this in my PR, because we probably also need to change what all the storage queries return...

modules/openid_connect/app/services/openid_connect/user_tokens/exchange_service.rb
```diff
@@ -46,19 +46,19 @@ def supported? = false
...
 
-        json = yield TokenRequest.new(provider:).exchange(idp_token, audience)
+        idp_token = yield fetch_openproject_token
+        json = yield exchange_token_request(idp_token, audience)
 
-        access_token = json["access_token"]
-        expires_in = json["expires_in"]
-        return Failure("Token exchange response invalid") if access_token.blank?
+        access_token, expires_in = json.values_at("access_token", "expires_in")
+        return Failure(@error.with(code: :token_exchange_response_invalid, payload: json)) if access_token.blank?
```
NobodysNightmare: :yellow_circle: Mh, so I assume that `with` will return a copy of the original where additional properties have been set. But to me it still feels weird to initialize an `@error` in the initializer that we will hopefully never use and then if we use it we create copies of it.

What do you think about one of these alternatives?

```
def error
  TokenOperationError.new(source: self.class)
end

# used like
error.with(code: :foo, payload: { foo: :bar })
```

This version looks like what you currently do during usage, but only instantiates an error if it occurs. However, it will still instantiate two errors (the original and the copy). We could optimize that like so:


```
def error(code:, payload:)
  TokenOperationError.new(source: self.class, code:, payload:)
end

# used like
error(code: :foo, payload: { foo: :bar })
```

This would probably be my favorite version. No useless instantiation, no instance variable that I might stumble across and still clear to read.

modules/openid_connect/app/services/openid_connect/user_tokens/exchange_service.rb
```diff
@@ -74,17 +74,26 @@ def supported?
 
       private
 
+      def fetch_openproject_token
```
NobodysNightmare: :green_circle: I think I'd avoid calling this `fetch_openproject_token`, since it could lead to the false impression that this is a token issued for the audience of OpenProject (our application), when we indeed fetch a token whose audience is the IDP that gave us the token.

To be consistent with existing variable names, what do you think about `fetch_idp_token`?

modules/openid_connect/app/services/openid_connect/user_tokens/exchange_service.rb
```diff
@@ -46,19 +46,19 @@ def supported? = false
...
+        end
 
-        json = yield TokenRequest.new(provider:).exchange(idp_token, audience)
+        idp_token = yield fetch_openproject_token
+        json = yield exchange_token_request(idp_token, audience)
 
-        access_token = json["access_token"]
-        expires_in = json["expires_in"]
-        return Failure("Token exchange response invalid") if access_token.blank?
+        access_token, expires_in = json.values_at("access_token", "expires_in")
```
NobodysNightmare, line 61: :green_heart: Sweet. That looks much nicer!

modules/openid_connect/app/services/openid_connect/user_tokens/fetch_service.rb
```diff
@@ -66,31 +67,17 @@ def access_token_for(audience:)
...
         token = @user.oidc_user_tokens.with_audience(aud).first
         return Success(token) if token
 
-        return @token_exchange.call(aud) if @token_exchange.supported?
-
-        Failure("No token for audience '#{aud}'")
+        if @token_exchange.supported?
+          @token_exchange.call(aud)
+        else
+          Failure(@error.with(code: :no_token_for_audience, payload: "No token for audience '#{aud}'"))
```
NobodysNightmare, line 88: TIL: Apparently our Rubocop is not configured to prefer guard clauses with no questions asked. I am pretty sure that a default-rubocop would opinionate against this new code in favor of the previous guard clause.

modules/storages/app/common/storages/peripherals/nextcloud_connection_validator.rb
```diff
@@ -93,6 +81,29 @@ def non_oidc_provisioned_user
...
       end
 
+      def token_usability
+        service = OpenIDConnect::UserTokens::FetchService.new(user: @user)
+
+        result = service.access_token_for(audience: @storage.audience)
+        return None() if result.success?
+
+        error_code = case result.failure
+                     in { code: /token_exchange/ | :unable_to_refresh_token }
```
NobodysNightmare: :green_circle: Should we rename this to `unable_to_exchange_token`? I see that it originates from the refresh service, but there it only occurs when we try to exchange instead of a refresh.

And here you decided to cause an error of `oidc_cant_exchange_token` instead of `oidc_cant_refresh_token`, so I guess it could make sense?

NobodysNightmare (approved):
Generally LGTM. After going through the services, I pretty clearly don't like the `@error = ...` initializers, but they are not a critical blocker for me, in case you prefer them for some reason :)

modules/openid_connect/app/services/openid_connect/user_tokens/exchange_service.rb
mereghost: Sounds good. I suck at naming, so any suggestion is more than welcome.

modules/openid_connect/app/services/openid_connect/user_tokens/fetch_service.rb
mereghost: I'm indifferent to it, but I do find it more legible when the return value is the result of a `if` to have it in full rather than a guard statement modifier.

If there were more computations besides returning the failure, I'd default to the guard clause. =)

modules/storages/app/common/storages/peripherals/nextcloud_connection_validator.rb
mereghost: I was torn at this: this errors comes up if:

- The existing token doesn't have the nextcloud storage audience
- No token exchange is possible.

I'm not sure how to capture that exact case in a name and specific error message. =/

modules/openid_connect/app/services/openid_connect/user_tokens/exchange_service.rb
mereghost: `with` returns a new `Data` instance with the new attributes.

The way I thought of it was similar to an ActiveRecord/Service where we have an `error` value tied to it.

modules/openid_connect/app/services/openid_connect/user_tokens/fetch_service.rb
NobodysNightmare: Yeah, I am also indifferent TBH. This was really just a TIL, not a request for change :)

modules/storages/app/common/storages/peripherals/nextcloud_connection_validator.rb
NobodysNightmare: IMO there are two ways to look at it.

1. Someone asks us to _refresh_ a token that has no `refresh_token` and our fallback plan of exchanging it instead doesn't work. So we could argue that the name `unable_to_refresh_token` is valid. However, then I'd also vote for moving it to the _refresh validation_. For example this error could legitimately occur if our IDP never hands out a refresh token and is incapable of token exchange as well.
2. We could argue that we are deep enough into trying an exchange, but then realize that an exchange is not possible. So the name could be `unable_to_exchange_token`

I think in both cases we will have the chance to misrepresent what the actual cause of the issue is. I think right now I'd keep the validation the way that you have it and rename to `unable_to_exchange_token`. But I could understand both ways to go about it.

modules/openid_connect/app/services/openid_connect/user_tokens/exchange_service.rb
NobodysNightmare: > The way I thought of it was similar to an ActiveRecord/Service where we have an error value tied to it.

Not sure if I understand this. In ways of the outside behaviour of our services, all three proposals should look the same. What I want to avoid is having an instance variable (state) that is not really being used for anything related to the state of our class, but purely for behaviour. That's what methods are for.

modules/openid_connect/app/services/openid_connect/user_tokens/fetch_service.rb
```diff
@@ -71,41 +72,21 @@ def access_token_for(audience:)
...
```
NobodysNightmare, line 88: Can we also get a `failure_with` here? :)

PR #18070, opened 2025-02-25
Stream the SQL dump file when backing up

crohr:
Previously this was loading the entire dump file in memory before adding to the zip, potentially leading to memory exhaustion.

https://community.openproject.org/work_packages/50637/activity


machisuji (approved):
👍

PR #18072, opened 2025-02-25
chore: bump md_to_pdf to 0.2.0

as-op:
This is bumping md_to_pdf to v0.2.0 which has the proper gems requirements for ruby >= 3.4.

This does not need to wait for a 15.4 release as it cannot be reviewed by QA and also fixes the bug in [61749](https://community.openproject.org/projects/openproject/work_packages/61749/activity) which enables a workaround for admins with PDF styling. So it would be nice to have in a patch level release (if there is one at all after 15.3.1)

# Merge checklist

- [x] Added/updated tests (in [md_to_pdf](https://github.com/opf/md-to-pdf/blob/main/spec/markdown_to_pdf/table_spec.rb#L245))


as-op:
Thanks @klaustopher!

PR #18076, opened 2025-02-25
Add `WorkPackage#comments_count` for counter cache

akabiru:
See: https://guides.rubyonrails.org/association_basics.html#counter-cache


brunopagno (approved):
It looks correct

PR #18079, opened 2025-02-25
make comments journable

brunopagno:

Checking how this would look like. Making comments into journable is good as an event ledger thing, but it's overkill as a snapshot changes tracker.

- I could not figure out a way of handling the user. Journals end up with a associated user and I don't know where it comes from, but the comments themselves have an associated user already. So we have some weird duplication here.


brunopagno:
Exploration was done and we decided not to continue with this refactor _now_

PR #18091, opened 2025-02-26
Introduce restricted journals for work package comments with restricted visibility

akabiru:
# Ticket

- [Database migration changes](https://community.openproject.org/work_packages/60980)
- [Restricted Comments UI](https://community.openproject.org/wp/60979)
- [Add feature flag](https://community.openproject.org/projects/communicator-stream/work_packages/62009/activity)
- [Add new permissions for restricted work package comments](https://community.openproject.org/projects/communicator-stream/work_packages/60978/activity)
- [Honor work package comment with restricted visibility permissions on the activity tab (in both notification center and wp full view)](https://community.openproject.org/projects/communicator-stream/work_packages/60989/activity)

This is a partial implementation of comments with restricted visibility. It is included here data and query changes on journals, so that we can hide/show restricted ones based on visibility rules.

https://github.com/user-attachments/assets/629aac33-c0a9-4e48-98a5-8315ef379786

# For reviewers

The code in this PR was mostly reviewed as other partial PRs that were merged in this branch. We would still very much appreciate a sanity check on database migrations and journal changes.


app/contracts/work_packages/create_note_contract.rb
```diff
@@ -28,14 +30,16 @@
...
-    end
+    def self.model = WorkPackage
 
     attribute :journal_notes do
       errors.add(:journal_notes, :error_unauthorized) unless can?(:comment)
     end
 
+    attribute :journal_restricted do
+      errors.add(:journal_restricted, :invalid) if model.journal_notes.blank? && model.journal_restricted
+    end
```
brunopagno: I think it makes things a bit simpler if we don't make a special case for notes. This check seems correct from the point of view of the "comments with restricted visibility" feature, but not really from journals as a db model

app/services/journals/create_service.rb
```diff
@@ -901,6 +904,10 @@ def only_one_note(predecessor, notes)
       predecessor.notes.empty? || notes.empty?
     end
 
+    def unrestricted_journal?(restricted)
+      !restricted
+    end
```
brunopagno: I think it would still be better to do this check as

```ruby
  predecessor.restricted == restricted
```

My thought is that if the previous and current updates are both restricted we should be okay aggregating them.

brunopagno (commented):
I'll say this looks absolutely briliant. Thanks a lot for digging into the journal behaviour :bow: 

From my side, I think we are okay pushing this forward, but I did leave a few comments with ideas that _in my insane mind_ might make things simpler for us in the future.

That said, I think it would probably worth getting the review of someone with a bit more mileage in this part of the code, just for sanity check of our assumptions.

app/contracts/work_packages/create_note_contract.rb
akabiru: Keeping in mind this contract is a `CreateNoteContract` which is only used in the `AddWorkPackageNoteService` specifically for wp journal notes and not in any other journal creation service it would seem fitting. Pls lmk if I'm missing some context

brunopagno: I see. That makes much more sense now. :thinking:
All good, we can mark it as resolved :bow:

brunopagno:
Hey @ulferts since you're the go to reference on journals we would apprciate your sanity check on this PR. The description should cover the current state of things, but let us know if anything.

config/initializers/permissions.rb
```diff
@@ -309,6 +309,34 @@
...
                      dependencies: :view_work_packages
 
+      wpt.permission :view_comments_with_restricted_visibility,
+                     {},
+                     permissible_on: %i[work_package project],
+                     require: :loggedin,
+                     dependencies: :view_work_packages,
+                     visible: -> { OpenProject::FeatureDecisions.comments_with_restricted_visibility_active? }
+
+      wpt.permission :write_comments_with_restricted_visibility,
```
akabiru: 🟢 _nitpick_ @brunopagno I wonder if `add_comments_with_restricted_visibility` is more fitting and conventional to OP, the term write can imply both add + edit actions which can be confusing

brunopagno: Good point. The name "write" came from product and I did not think about this. The actual _action_ would be `add`. Lets remember to adjust this before merging, then.

brunopagno: https://github.com/opf/openproject/pull/18253

app/contracts/work_packages/create_note_contract.rb
```diff
@@ -28,14 +30,16 @@
...
-      WorkPackage
-    end
+    def self.model = WorkPackage
 
     attribute :journal_notes do
       errors.add(:journal_notes, :error_unauthorized) unless can?(:comment)
     end
 
+    attribute :journal_restricted do
+      errors.add(:journal_restricted, :invalid) if model.journal_notes.blank? && model.journal_restricted
```
ulferts: Didn't look into this too deeply. But would it make sense to check for the feature flag in here as well?

app/controllers/work_packages/activities_tab_controller.rb
```diff
@@ -295,13 +295,14 @@ def update_index_component
...
+    to_boolean = ->(value, default) { ActiveRecord::Type::Boolean.new.cast(value.presence || default) }
+
     AddWorkPackageNoteService
       .new(user: User.current,
            work_package: @work_package)
       .call(journal_params[:notes],
-            send_notifications: !(params.has_key?(:notify) && params[:notify] == "false"))
-    ###
+            send_notifications: to_boolean.call(params[:notify], true),
+            restricted: to_boolean.call(journal_params[:restricted], false))
```
ulferts: Since the intend of the to_boolean proc doesn't seem to pass it around, it would be easier to just extract the functionality into a simple method of this controller.

app/models/journal.rb
```diff
@@ -113,6 +113,14 @@ class Journal < ApplicationRecord
   scope :for_wiki_page, -> { where(journable_type: "WikiPage") }
   scope :for_work_package, -> { where(journable_type: "WorkPackage") }
   scope :for_meeting, -> { where(journable_type: "Meeting") }
+  scope :restricted_visible, ->(work_package) {
+    if OpenProject::FeatureDecisions.comments_with_restricted_visibility_active? &&
+        User.current.allowed_in_work_package?(:view_comments_with_restricted_visibility, work_package)
+      all
+    else
+      where(restricted: false)
+    end
```
ulferts: This I fear is prone to producing errors. At least when used in a way that is not intended by this PR. 

Within this PR, the calls are like this:
```
work_package
  .journals
  .restricted_visible(work_package
```

Doing so will return the visible journals of this work package.

But nothing prevents somebody from calling

```
Journal
  .restricted_visible(work_package)
```

which, depending on the permissions, might translate to

```
Journal.all
```

And that, after the permission has been checked on a single work package.

This seems to be a prime use case for an [extension](https://guides.rubyonrails.org/association_basics.html#extensions) on the `journals` association.

That way, the scope would only be visible on the `journals` association and it would read nicer as the caller wouldn't have to provide the `work_package` again:

```
work_package
  .journals
  .restricted_visible
```

config/initializers/permissions.rb
```diff
@@ -309,6 +309,34 @@
...
+                     permissible_on: %i[work_package project],
+                     require: :loggedin,
+                     dependencies: %i[view_work_packages view_comments_with_restricted_visibility],
+                     visible: -> { OpenProject::FeatureDecisions.comments_with_restricted_visibility_active? }
+
+      wpt.permission :edit_others_comments_with_restricted_visibility,
+                     {},
+                     permissible_on: %i[work_package project],
+                     require: :loggedin,
+                     dependencies: %i[view_work_packages view_comments_with_restricted_visibility],
```
ulferts: Is the intend to be able to set the permissions on individual work packages? So far, I haven't seen this manifested in the PR, at least in the UI. I see also calls for `User.current.allowed_in_work_package?` which suggests that this is intended. But it seems to me that the permissions are currently only seeded for project permissions. And since there is no UI currently for the work package roles, users will not be able to assign the permission themselves. 

I could see valid use cases for doing this but the reason why I am asking is performance. Querying for work package permissions is way more costly than querying on the project level.

app/models/activities/fetcher.rb
```diff
@@ -114,6 +115,16 @@ def eager_load_associations(events)
...
     end
 
+    def filter_by_visibility(events)
+      events.reject! do |event|
+        if OpenProject::FeatureDecisions.comments_with_restricted_visibility_active?
+          event.journal.restricted? && !@user.allowed_in_project?(:view_comments_with_restricted_visibility, event.project)
+        else
+          event.journal.restricted?
+        end
+      end
```
ulferts: This here contradicts what is written in the `restricted_visible` scope where the check is on the work package level as well. 

Additionally, it will be slow without benefits I can see. Adapting the query for fetching the journals in the first place is the way to go, I think.

To this end, a subquery can be added when fetching the journals. That could either be

```
WorkPackage.allowed_to(User.current, :view_comments_with_restricted_visibility)
```

or

```
Project.allowed_to(User.current, :view_comments_with_restricted_visibility)
```

depending on whether the permission can or cannot be assigned to work package roles as well.

app/components/work_packages/activities_tab/journals/new_component.html.erb
```diff
@@ -53,28 +53,41 @@
...
-                  type: :submit,
-                  data: {
-                    test_selector: "op-submit-work-package-journal-form",
-                    "work-packages--activities-tab--index-target": "formSubmitButton"
-                  }
+          if adding_restricted_comment_allowed?
+            concat(
+              flex_layout do |flex|
+                flex.with_row(mb: 3, test_selector: "op-work-package-journal-restricted-comment-checkbox") do
+                  render(WorkPackages::ActivitiesTab::Journals::RestrictedNoteForm.new(f))
```
ulferts: Clicking on the checkbox closes the comment field unless the user inserts text first. So if the user has the flow of first selecting the visibility, this is annoying.

Kapture 2025-03-13 at 10 30 36

app/models/work_package.rb
```diff
@@ -382,8 +382,18 @@ def update_done_ratio_from_status
   # check if user is allowed to edit WorkPackage Journals.
   # see Acts::Journalized::Permissions#journal_editable_by
   def journal_editable_by?(journal, user)
-    user.allowed_in_project?(:edit_work_package_notes, project) ||
-      (user.allowed_in_work_package?(:edit_own_work_package_notes, self) && journal.user_id == user.id)
+    can_edit_journal_note = ->(edit_others_notes_permission, edit_own_notes_permission) {
+      user.allowed_in_project?(edit_others_notes_permission, project) ||
+        (user.allowed_in_work_package?(edit_own_notes_permission, self) && journal.user_id == user.id)
+    }
```
ulferts: This I would also extract into a simple method that is then called.

ulferts (changes requested):
I haven't completed my review yet. From what I have seen though, there are a number of contradictions and things to improve but the PR is heading into a direction that this can be merged. 

Personally, I am still convinced that the commenting functionality should be extracted from journalizing but within the constraints of excluding this possibility, this PR works.

Appart from the comments within the code, what I would see as a prerequisite for merging, are tests added on the `acts_as_journalized_level`, e.g. for [work packages](https://github.com/opf/openproject/blob/dev/spec/models/work_package/work_package_acts_as_journalized_spec.rb)

app/services/journals/create_service.rb
```diff
@@ -62,24 +62,28 @@ def call(notes: "", cause: CauseOfChange::NoCause.new)
...
     private
 
-    # If the journalizing happens within the configured aggregation time, is carried out by the same user, has an
-    # identical cause and only the predecessor or the journal to be created has notes, the changes are aggregated.
+    # Aggregatable journals must meet the following criteria:
+    #
+    # * The journalizing happens within the configured aggregation time
+    # * The journalizing is carried out by the same user
+    # * The cause is identical
+    # * ONLY the predecessor or the journal to be created has notes
```
ulferts: Aware that this was taken from the comment before, I would still rewrite this as:

'ONLY one note exists between the predecessor and the journal to be created'

app/services/journals/create_service.rb
```diff
@@ -62,24 +62,28 @@ def call(notes: "", cause: CauseOfChange::NoCause.new)
...
+    # * The predecessor and the journal to be created have the restriction
```
ulferts: This sentence seems to be incomplete. At least as of now, that would mean that either both are restricted or both are not.

In the broader context, I don't know if that actually needs to be checked as, at least currently, a restricted journal will always have a note. But I guess it is better to check this explicitly in case the requirements change.

app/components/work_packages/activities_tab/index_component.rb
```diff
@@ -80,6 +93,8 @@ def polling_interval
       def adding_comment_allowed?
         User.current.allowed_in_work_package?(:add_work_package_notes, @work_package)
       end
+
+      def activities_tab_index_stimulus_controller = "work-packages--activities-tab--index"
```
ulferts: Why extract this to a method and not the call in [`add_comment_wrapper_data_attributes`](https://github.com/opf/openproject/pull/18091/files#diff-51c7ddab80242dd6f1c64cf7c9f8fc05e7d239ffc0e53eef728117caf1f9137bR72)

Personally, I would also prefer having it extracted to a method one can then just call where the suffix is appended. I think it would read nicer, e.g.:

```
          index_stimulus_controller("-update-streams-path-value") => :update_streams_work_package_activities_path(work_package),
          index_stimulus_controller("-sorting-value") => journal_sorting,
          index_stimulus_controller("-filter-value") => filter,
          index_stimulus_controller("-user-id-value" => User.current.id,
          index_stimulus_controller("-work-package-id-value") => work_package.id,
          index_stimulus_controller("-polling-interval-in-ms-value") => polling_interval,
          index_stimulus_controller("-notification-center-path-name-value") => notifications_path,
          index_stimulus_controller("-show-conflict-flash-message-url-value") => show_conflict_flash_message_work_packages_path,
```

But admittedly, that is subjective, so feel free to ignore my comment.

app/components/work_packages/activities_tab/journals/new_component.rb
```diff
@@ -57,6 +57,11 @@ def button_row_display_value
         def form_row_display_value
           form_hidden_initially ? :none : :block
         end
+
+        def adding_restricted_comment_allowed?
+          OpenProject::FeatureDecisions.comments_with_restricted_visibility_active? &&
+            User.current.allowed_in_work_package?(:add_comments_with_restricted_visibility, work_package)
```
ulferts: This line of code again checks for the permission on work package level

app/components/work_packages/activities_tab/journals/new_component.html.erb
brunopagno: This is a known issue. We are making design changes to the whole comment compose input as follow ups.

config/initializers/permissions.rb
brunopagno: Good point. I believe the product requirement is only project permissions, not permission for individual work packages.

We can make the change.

app/models/work_package.rb
ulferts: This also has both calls in it. The one on the project and the one on the work package level.

ulferts (changes requested):
I finished my review. To be clear, I didn't do a complete review but rather focused on the backend changes and the structure therein.

It was surprising to see how little changes were necessary on the `Journals::CreateService`. That is good news insofar as that there should be less possibilities for errors.

app/components/work_packages/activities_tab/index_component.rb
akabiru: It's actually called in the [`action: "#{activities_tab_index_stimulus_controller}:onSubmit-end@window->#{stimulus_controller}#onSubmitEnd",`](https://github.com/opf/openproject/pull/18091/files#diff-51c7ddab80242dd6f1c64cf7c9f8fc05e7d239ffc0e53eef728117caf1f9137bR79) that passes events between the index and comments controller.

Good call, the proposed structure with suffix does read nicer. Will update! 👍🏾

app/components/work_packages/activities_tab/journals/new_component.html.erb
akabiru: Yes- covered in #18262, #18302

app/controllers/work_packages/activities_tab_controller.rb
akabiru: Makes sense- I was trying a new "localised" approach. Indeed a method is simpler. 👍🏾

app/contracts/work_packages/create_note_contract.rb
akabiru: I wasn't sure before- but it doesn't hurt to implement additional safety. I reckon it would mean that when the feature flag is off- and the user hacks in sending `retsricted=true`- the contract fails validation

app/models/journal.rb
akabiru: Oh wow, I see how this can be misused. The extensions idea is sound. We'll look into that! 👍🏾

app/services/journals/create_service.rb
akabiru: Ah, yes- typo! 👍🏾

config/initializers/permissions.rb
brunopagno: This was adjusted and merged into the feature branch. Resolving the comment

app/models/activities/fetcher.rb
brunopagno: @akabiru fyi, I'll look into this one, next

app/models/journal.rb
akabiru: @brunopagno looking into this one 👀

akabiru: https://github.com/opf/openproject/pull/18366

app/models/activities/fetcher.rb
brunopagno: The fix https://github.com/opf/openproject/pull/18367

brunopagno:
@akabiru in which scenario do we do squash&merge?

brunopagno:
@ulferts I think it's possible to review again. We should have covered the requested changes.

brunopagno:
Just noticed there are conflicts, I'll take a look

akabiru:
> @akabiru in which scenario do we do squash&merge?

I would merge in this case @brunopagno - I know it's a lot of commits but in this case I'd rather keep the history.

brunopagno (approved):

brunopagno:
We addressed the concerns on the PR so we are merging it.

@ulferts if you want to have another look and point out anything else that we might have missed, feel free to reach out.

PR #18099, opened 2025-02-27
[#61604] harden default function of property mapper

Kharonus:
# Ticket
[OP#61604](https://community.openproject.org/wp/61604)

- generated subject of wp without parent, while pattern contains parent custom field, should not render error

- context can be `nil` - e.g. if no parent exists - in this case `nil` must be returned

### DIFF
--- a/app/models/types/patterns/token_property_mapper.rb
+++ b/app/models/types/patterns/token_property_mapper.rb
@@ -31,7 +31,11 @@
 module Types
   module Patterns
     class TokenPropertyMapper
-      DEFAULT_FUNCTION = ->(key, context) { context.public_send(key.to_sym) }.curry
+      DEFAULT_FUNCTION = ->(key, context) do
+        return nil if context.nil?
+
+        context.public_send(key.to_sym)
+      end.curry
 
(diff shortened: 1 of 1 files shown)


app/models/types/patterns/token_property_mapper.rb
```diff
@@ -31,7 +31,11 @@
 module Types
   module Patterns
     class TokenPropertyMapper
-      DEFAULT_FUNCTION = ->(key, context) { context.public_send(key.to_sym) }.curry
+      DEFAULT_FUNCTION = ->(key, context) do
+        return nil if context.nil?
```
mereghost, line 35: I think we should raise an exception here. This should never happen.

mereghost (changes requested):

app/models/types/patterns/token_property_mapper.rb
Kharonus: Huh? of course this happens. If you have a wp without a parent. ;) the other option is to make this decision in the caller. You think, there is the better place?

mereghost (approved):

PR #18111, opened 2025-02-28
SSO error handling

NobodysNightmare:
# Ticket
* https://community.openproject.org/projects/cross-application-user-integration-stream/work_packages/61612
* https://community.openproject.org/projects/cross-application-user-integration-stream/work_packages/61880

Fixing errors related to showing the connection status of SSO users to their storage.

In the files tab, the following cases should now be indicated correctly:

* A "classic" (OAuth) user is not yet linked to the storage (existed before)
* A "classic" (OAuth) user could not successfully authenticate at the storage (new: indicated as an error)
* An SSO user could not successfully authenticate at the storage (indicated as an error)

In the project storages member overview (e.g. `/projects/:name/settings/project_storages/:id/members`):

* SSO members are now indicated as connected, once they obtained a remote identity
* SSO members have a specific not-connected message, when there is no remote identity yet

# Which approach did you choose and why?
It occured to me, that the question of which authentication method will really be used for a user at a storage has to be asked in many different places. Thus I extracted the relevant code into a helper called `AuthenticationMethodSelector`.

# Merge checklist

- [x] Added/updated tests

### DIFF
--- a/modules/storages/app/components/storages/project_storages/members/row_component.rb
+++ b/modules/storages/app/components/storages/project_storages/members/row_component.rb
@@ -51,19 +51,16 @@ module Storages::ProjectStorages::Members
 
     def status
       connection_result = storage_connection_status
-
-      if connection_result == :not_connected
-        ensure_connection_url = oauth_clients_ensure_connection_url(
-          oauth_client_id: storage.oauth_client.client_id,
-          storage_id: storage.id
-        )
+      case connection_result
+      when :not_connected
         helpers.op_icon("icon-warning -warning") +
           content_tag(
             :span,
             I18n.t("storages.member_connection_status.not_connected",
-                   link: link_to(I18n.t("link"), ensure_connection_url),
-                   class: "pl-2").html_safe
+                   link: link_to(I18n.t("link"), ensure_connection_url)).html_safe
           )
+      when :not_connected_sso
+        content_tag(:span, I18n.t("storages.member_connection_status.not_connected_sso"))
       else
         I18n.t("storages.member_connection_status.#{connection_result}")
       end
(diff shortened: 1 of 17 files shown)


modules/storages/app/components/storages/project_storages/members/row_component.rb
```diff
@@ -89,7 +86,12 @@ def principal_show_path
     end
 
     def storage_connection_status
-      return :not_connected unless oauth_client_connected?
+      unless oauth_client_connected?
```
ba1ash: :yellow_circle: I think this method name is bad now. Therefore I suggest to remove the method and just use its body in the condition here. It is expressive enough and no need to think about a better method name.
```suggestion
      unless member.principal.remote_identities.exists?(integration: storage)
```

modules/storages/app/components/storages/project_storages/members/row_component.rb
```diff
@@ -99,12 +101,18 @@ def storage_connection_status
     end
 
     def oauth_client_connected?
```
ba1ash: :yellow_circle: I suggest to remove this method in favor the suggestion above.

ba1ash (approved):
LGTM. It works fine for my local setup. I have only one code suggestion.

modules/storages/app/components/storages/project_storages/members/row_component.rb
NobodysNightmare: I changed the method name to `storage_connected?`. I preferred to keep the method, because it keeps the abstraction level inside `storage_connection_status` more consistent.

After renaming, I noticed that I could replace the `unless` with an `if` by switching the method order around, so I did that as well.

ba1ash: Okay, works for me. Thanks for taking care! :+1:

PR #18113, opened 2025-02-28
[#56694] transform Markdown to HTML in search results

EinLama:
# Ticket
https://community.openproject.org/wp/56694 

The search results should display proper markdown where applicable. This is more of a feature than a mere bugfix, I suppose.

I have used our text formatter to transform the event description from markdown to HTML. The previous plaintext implementation offered text abbreviation and search result highlighting. I preserved this behavior by carefully modifying the generated HTML DOM.

## Screenshots
image

# Merge checklist

- [x] Added/updated tests

### DIFF
--- a/app/helpers/search_helper.rb
+++ b/app/helpers/search_helper.rb
@@ -76,7 +76,9 @@ module SearchHelper
       doc.traverse do |node|
         next unless node.text?
 
-        highlighted_text = node.content.gsub(/(#{Regexp.escape(token)})/i, '<span class="search-highlight">\1</span>')
+        t = (tokens.index(node.content.downcase) || 0) % 4
+        highlighted_text = node.content.gsub(/(#{Regexp.escape(token)})/i,
+                                             "<span class='search-highlight token-#{t}'>\\1</span>")
         node.replace(Nokogiri::HTML::DocumentFragment.parse(highlighted_text))
       end
     end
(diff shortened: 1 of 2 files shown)


app/helpers/search_helper.rb
```diff
@@ -170,4 +204,44 @@ def abbreviated_text(words)
...
+  end
+
+  def process_text_node(content, current_length, max_length)
+    return content if current_length >= max_length
+
+    truncated_text = truncate_formatted_text(content, length: 100)
+
+    if current_length + truncated_text.length > max_length
+      t = truncated_text[0..(max_length - current_length - 1)]
+      truncated_text = "#{t}..."
```
dombesz: I believe we could use the [truncate](https://apidock.com/rails/v7.1.3.4/String/truncate) rails string extension:

```suggestion
      truncated_text.truncate(max_length - current_length - 1)
```
I am also wondering if the `truncate_formatted_text` is required, since we are processing the text nodes here.

dombesz (approved):
Works as expected, thanks Tobi.👏
I left one comment only.

app/helpers/search_helper.rb
EinLama: Thank you, I have added `truncate` 👏🏻 

The `truncate_formatted_text` call is required to shorten additional portions of the search results, so I left it in for now.

PR #18124, opened 2025-03-03
[#54378] show calendar widget date in configured format

EinLama:
# Ticket
https://community.openproject.org/wp/54378

The configured date format should be used in the calendar (and calendar widgets) if applicable. Also addressed some eslint-related errors and warnings.

After I implemented everything, I noticed that the date format strings always include the year. Rendering the year in the calendar header takes up a lot of unnecessary space, as the year is most often visible in a separate tag above the calendar. So I decided to strip the year from the format string before rendering the date.

## Screenshots

https://github.com/user-attachments/assets/76dac8e3-4477-46ff-9813-565fa96fbee3

Demonstration of applying different settings. Calendar widget on project overview page. Note that this video was recorded with an earlier code base, where the year is not yet stripped from the calendars format string.

Screenshot 2025-03-04 at 11 32 38
Calendar page, week grid.

Screenshot 2025-03-04 at 11 32 35
Calendar page, month grid (note that the date format changes will not be applied here, this is intentional)

Screenshot 2025-03-04 at 11 33 38
Calendar widget on "My page"

Using the `ConfigurationService` to fetch the currently configured date format and applying it in the [dayHeaderContent](https://fullcalendar.io/docs/day-header-render-hooks) hook of fullcalendar.

When the user has not configured the date, we default to the fullcalendar default settings for date formatting. This will also apply when the calendar is in dayGridMonth mode, which is visible on the calendar page, when you click on "month". Here, the column header will represent a day of the week - and no date is visible. See above screenshots for an example.

# Merge checklist

- [x] Added/updated tests

### DIFF
--- a/frontend/src/app/features/calendar/op-calendar.service.spec.ts
+++ b/frontend/src/app/features/calendar/op-calendar.service.spec.ts
@@ -1,0 +1,69 @@
+//-- copyright
+// OpenProject is an open source project management software.
+// Copyright (C) the OpenProject GmbH
+//
+// This program is free software; you can redistribute it and/or
+// modify it under the terms of the GNU General Public License version 3.
+//
+// OpenProject is a fork of ChiliProject, which is a fork of Redmine. The copyright follows:
+// Copyright (C) 2006-2013 Jean-Philippe Lang
+// Copyright (C) 2010-2013 the ChiliProject Team
+//
+// This program is free software; you can redistribute it and/or
+// modify it under the terms of the GNU General Public License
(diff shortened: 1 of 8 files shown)


ulferts (approved):
Looks good, @EinLama . 

I didn't merge since the unit specs report an error (unrelated- a merge might fix that) so I couldn't get the results of the feature specs.

EinLama:
Merging as the red spec passes locally. Additionally it's unrelated to any code changes in this PR.

PR #18128, opened 2025-03-03
[#61692] add validation for invalid tokens

Kharonus:
# Ticket
[OP#61692](https://community.openproject.org/work_packages/61692)

- prevent submitting subject patterns with invalid tokens

- added validation rule to update service
- display invalid tokens in validation error

### DIFF
--- a/app/services/update_type_service.rb
+++ b/app/services/update_type_service.rb
@@ -43,9 +43,11 @@ class UpdateTypeService < BaseTypeService
   def set_params_and_validate(params)
     # Set patterns includes a data validation before assigning the value to the attribute.
     # A validation failure should return a service call failure.
-    if params[:patterns].present?
-      validate_enterprise_action(params[:patterns])
-      set_patterns(params[:patterns])
+    patterns = params[:patterns]
+    if patterns.present?
+      validate_enterprise_action(patterns)
+      validate_pattern_content(patterns)
+      set_patterns(patterns)
       return [false, type.errors] if type.errors.any?
     end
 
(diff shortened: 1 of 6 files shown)


app/services/update_type_service.rb
```diff
@@ -61,6 +63,25 @@ def validate_enterprise_action(patterns)
...
+      valid_tokens.include?(token) ? acc : acc << token
+    end
+
+    if invalid_tokens.any?
+      message = I18n.t("types.edit.subject_configuration.errors.invalid_tokens", tokens: invalid_tokens.join(", "))
+      type.errors.add(:patterns, message)
+    end
+  end
+
+  def flat_valid_token_list = Types::Patterns::TokenPropertyMapper.new.tokens_for_type(type).values.map(&:keys).flatten
```
brunopagno: :clap:

brunopagno (approved):
Looks good to me :grin:

Kharonus:
Thanks for the very quick review, @brunopagno .

I still would like to push another commit, which moves the validation from service to the base contract of wp types. As I discussed with @mereghost, that the validation of the content is independent from the validation of the format (data structure) I realised that we actually CAN now have this as a validation rule in the base contract itself.

So, please have a look, once I pushed it. I write is as a separate commit, so that we can easily decide, which commit to merge.

Kharonus:
So, @brunopagno @mereghost, can you recheck?

brunopagno (approved):
Still looks beautiful. Even more so, I'd dare to say~

PR #18131, opened 2025-03-03
Reject restricted journals when user cannot see them

brunopagno:
# Ticket

https://community.openproject.org/wp/60989

Filter out restricted comments from activities on:
- Global Activities
- Project Activities
- Activity tab (Work Package activities)
- User Activities

## Screenshots

None yet

Not even sure yet, just exploring and trying out things

# Merge checklist

- [ ] Added/updated tests


app/components/work_packages/activities_tab/journals/index_component.rb
```diff
@@ -68,9 +68,12 @@ def base_journals
         end
 
         def fetch_journals
+          can_see_restricted = User.current.allowed_in_work_package?(:view_comments_with_restricted_visibility, work_package)
+
           API::V3::Activities::ActivityEagerLoadingWrapper.wrap(
             work_package
               .journals
+              .where(can_see_restricted ? nil : { restricted: false })
```
akabiru: 👋🏾 I think a `visibile` active record scope would be more fitting, it would also remove the need for nil checks as scopes pass through AR relations 

Ref: https://github.com/search?q=repo%3Aopf%2Fopenproject%20visible&type=code

brunopagno: Oh yeah, I'm still in the "get the thing to work" phase, before prettifying. But that's a good tip, will save me some time investigating. Thanks :bow:

app/models/activities/fetcher.rb
```diff
@@ -114,6 +115,12 @@ def eager_load_associations(events)
       end
     end
 
+    def filter_by_visibility(events)
+      events.reject! do |event|
+        event.journal.restricted? && !@user.allowed_in_project?(:view_comments_with_restricted_visibility, event.project)
+      end
+    end
+
```
akabiru, line 123: I half expected that we could reuse the `restricted_visible` scope but I see it's not exactly straightforward due to the permissions check 👍🏾

app/components/work_packages/activities_tab/journals/index_component.rb
```diff
@@ -71,6 +71,7 @@ def fetch_journals
           API::V3::Activities::ActivityEagerLoadingWrapper.wrap(
             work_package
               .journals
+              .restricted_visible(work_package)
```
akabiru, line 74: ✨

spec/models/journal_spec.rb
```diff
@@ -68,4 +68,48 @@
       end
     end
   end
+
+  describe "#restricted_visible scope" do
```
akabiru:
```suggestion
  describe ".restricted_visible scope" do
```

akabiru (approved):
Beautiful 👏🏾 👏🏾 :shipit:

app/models/activities/fetcher.rb
brunopagno: I thought about it but could not find an obvious way of doing it as we _start from events_ and include journals.

I also deemed this as not the most important since the fetcher seems to be the least important place where we query activities at the moment.

PR #18148, opened 2025-03-04
Show correct text when audience is nil

NobodysNightmare:
Also adding tests so that we cover these cases in the
future. While writing them, I also noticed that one
out of three translations didn't end in punctuation.
This has been made more consistent now.

# Ticket
https://community.openproject.org/projects/cross-application-user-integration-stream/work_packages/61855

# Merge checklist

- [x] Added/updated tests


ba1ash (approved):
LGTM

PR #18150, opened 2025-03-04
[#61446] PDF export of a single work package follows work package form configuration including related wp tables

as-op:
# Ticket

https://community.openproject.org/work_packages/61446

## Screenshots

# Development checklist

- [x] dedicated styles for each of the PDF exports
- [x] refactoring into folders per export type
- [x] follow the work package form configuration
- [x] implement export of related wp tables from the work package form configuration (query group)
   - [x] if there are no work packages, a "[No work packages]" message will be displayed instead
   - [x] no borders, no uppercase letters in the header row
- [x] insert long text custom fields according to the work package form configuration (not last in the attribute group)
- [x] link to the work package from the subject line
- [x] page orientation setting
- [x] less bold attribute labels, e.g. no uppercase
- [x] add type and ID in front of the subject
- [x] add status behind the subject 
- [x] do not split date-field (combined as in the web), remove duration
- [x] line below group section title

# Merge checklist

- [x] Added/updated tests


app/components/work_packages/exports/generate/modal_dialog_component.rb
```diff
@@ -66,6 +66,17 @@ def hyphenation_default
...
         end
 
+        def page_orientation_default
+          page_orientation_options[0]
+        end
+
+        def page_orientation_options
+          [
+            { label: "Portrait", value: "portrait" },
+            { label: "Landscape", value: "landscape" }
```
ulferts: These need to be i18ned

docs/system-admin-guide/design/pdf-export-styles/attributes-and-description/README.md
```diff
@@ -0,0 +1,928 @@
...
+  group_heading: {}
+  table: {}
+```
+
+| Key | Description | Data type |
+| - | - | - |
+| `group_heading` | **Overview group heading**<br/>Styling for the group label if grouping is activated<br/>See [Overview group heading](#overview-group-heading) | object |
+| `table` | **Overview table**<br/>See [Overview table](#overview-table) | object |
+
+## alert boxes (styled blockquotes)
```
ulferts: This is written in all lowercase when the other headings are written with the first char being in uppercase.

docs/system-admin-guide/design/pdf-export-styles/attributes-and-description/README.md
```diff
@@ -0,0 +1,928 @@
...
+
+## Alert
+
+Styling to denote a quote as alert box
+
+Key: `alert`
+
+Example:
+```yml
+ALERT:
```
ulferts, line 26: Why is this key in all uppercase? I guess it doesn't matter but it is confusing.

docs/system-admin-guide/design/pdf-export-styles/README.md
```diff
@@ -1,1226 +1,10 @@
...
-`mm` - Millimeter, `cm` - Centimeter, `dm` - Decimeter, `m` - Meter
-
-`in` - Inch, `ft` - Feet, `yr` - Yard
-
-`pt` - [Postscript point](https://en.wikipedia.org/wiki/Point_(typography)#Desktop_publishing_point) (default if no unit is used)
+| YAML file                                                  | PDF Export                                                 |
+|------------------------------------------------------------|------------------------------------------------------------|
+| [attributes-and-description](./attributes-and-description) | Single PDF Export template "Attributes and description"    |
+| [report](./report)                                         | PDF Export template "Report", "Overview table" and "Gantt" |
+| [timesheet](./timesheet)                                   | PDF Export of timesheets of the Cost module                |
```
ulferts: It would have helped me to provide some additional context about what the purpose of those styling files are and how they are built up. I guess someone could read all this out of the schema files but a text describing the purpose would be helpful.

docs/system-admin-guide/design/pdf-export-styles/report/README.md
```diff
@@ -0,0 +1,1179 @@
...
+  styles:
+    - bold
+```
+
+| Key | Description | Data type |
+| - | - | - |
+| … | See [Font properties](#font-properties) |  |
+| … | See [Margin properties](#margin-properties) |  |
+
+## alert boxes (styled blockquotes)
```
ulferts: Also written in lowercase here

app/models/work_package/pdf_export/export/report/schema.json
```diff
@@ -547,6 +547,54 @@
...
+                  "height": 1
+                }
+              },
+              "properties": {
+                "height": {
+                  "title": "Line height of the horizontal rule",
+                  "examples": [
+                    1
+                  ],
+                  "$ref": "#/$defs/measurement"
```
ulferts, line 580: Is it possible to write the default values that are present in the application into here as well?

ulferts (changes requested):
I couldn't do a proper review reading and studying line by line the changes herein. There are just too many lines of code in a codebase, I am not knowledgeable in. I remarked minor things that are probably easy to fix or can be ignored. 

From a user perspective, three things struck me as weird:
1) Description is still displayed last. That used to make sense with the table on top but now that the attempt is to have the printout follow the same structure as in the UI, it feels odd.
2) Empty query groups are hidden. Doing this made me as a user wonder if I did something wrong. Displaying an "empty" message would be cleaner to me.
3) Images are scaled to a point where they overflow the page easily. This probably existed before, but with the landscape mode, this will happen more often:

image

docs/system-admin-guide/design/pdf-export-styles/attributes-and-description/README.md
as-op: Unfortunately, this inconsistency is in the md-to-pdf gem, I will fix this outside of this PR

app/components/work_packages/exports/generate/modal_dialog_component.rb
as-op: Thanks! I added the strings already to the yml file, but forgot to use them

app/models/work_package/pdf_export/export/report/schema.json
as-op: While this would be nice, it would only work for top entries in the schema not for the schema components as they are to be reused and then have different example/default values.

as-op:
> 1. Description is still displayed last. That used to make sense with the table on top but now that the attempt is to have the printout follow the same structure as in the UI, it feels odd.

I've discussed this with Niels, and we kept the description last for now.

> 2. Empty query groups are hidden. Doing this made me as a user wonder if I did something wrong. Displaying an "empty" message would be cleaner to me.

Agreed, I will amend the code and acceptance criteria accordingly.

> 3. Images are scaled to a point where they overflow the page easily. This probably existed before, but with the landscape mode, this will happen more often:

Yes, there is already a bug ticket, I've increased the priority so I can address this soon (in the md-to-pdf gem). https://community.openproject.org/projects/14/work_packages/61732/activity

docs/system-admin-guide/design/pdf-export-styles/README.md
as-op: I added a much better description. These files are just temporary solutions, as they will be overwritten by every OP update. The plan is to have a dedicated [admin section](https://community.openproject.org/projects/14/work_packages/61743/activity).

PR #18151, opened 2025-03-04
[#61573] allow more mouse control on pattern input

Kharonus:
# Ticket
[OP#61573](https://community.openproject.org/wp/61573)

- enable better control with mouse on the pattern input

- clicking before the first token gives a whitespace before it
- clicking behind the last token gives a whitespace behind it

### DIFF
--- a/frontend/src/stimulus/controllers/pattern-input.controller.ts
+++ b/frontend/src/stimulus/controllers/pattern-input.controller.ts
@@ -87,12 +87,12 @@ export default class PatternInputController extends Controller {
       event.preventDefault();
     }
     if (event.key === 'ArrowLeft') {
-      if (this.contentTarget.innerHTML.startsWith('<')) {
+      if (this.startsWithToken()) {
         this.insertSpaceIfFirstCharacter();
       }
     }
     if (event.key === 'ArrowRight') {
-      if (this.contentTarget.innerHTML.endsWith('>')) {
+      if (this.endsWithToken()) {
         this.insertSpaceIfLastCharacter();
       }
     }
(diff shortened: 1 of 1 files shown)


Kharonus:
@brunopagno 

interested in some more small improvements on your baby? :D At least I hope, those are improvements.

frontend/src/stimulus/controllers/pattern-input.controller.ts
```diff
@@ -226,16 +235,24 @@ export default class PatternInputController extends Controller {
...
+    selection.removeAllRanges();
+    selection.addRange(postRange);
+  }
+
+  private endsWithToken():boolean {
+    return this.contentTarget.innerHTML.endsWith('>');
+  }
+
+  private startsWithToken():boolean {
+    return this.contentTarget.innerHTML.startsWith('<');
```
brunopagno, line 255: :clap:

frontend/src/stimulus/controllers/pattern-input.controller.ts
```diff
@@ -226,16 +235,24 @@ export default class PatternInputController extends Controller {
...
-    if (selection) {
-      const postRange = document.createRange();
-      if (position === 'after') {
-        postRange.setStartAfter(target);
-      } else {
-        postRange.setStartBefore(target);
-      }
-      selection?.removeAllRanges();
-      selection?.addRange(postRange);
+    if (selection === null) { return; }
```
brunopagno, line 238: This is semantically different than the original check, but it looks correct according to documentation. I think this is okay :+1: 

https://developer.mozilla.org/en-US/docs/Web/API/Window/getSelection

brunopagno (approved):
I think this is the way. I don't like that we need to do so many "manual interventions" to make the behaviour feel right, but I think this implementation is as straightforward as we can get.

Nice job :tada:

frontend/src/stimulus/controllers/pattern-input.controller.ts
Kharonus: yeah, typescript is way more specific about static typing. So we safely can use that `null` check.

PR #18152, opened 2025-03-04
[#62523] Work package datepicker incorrectly renders turbo frame response

dombesz:
# Ticket
https://community.openproject.org/wp/62523

The calendar Turbo Frame rendering is buggy and bloats the DOM when switching between automatic and manual mode. This occurs because the response updates an element inside the Turbo Frame instead of replacing the frame itself, leading to deeply nested frames. To reproduce, toggle between modes several times—each render adds a new Turbo Frame inside the previous one.

It additionally fixed some focus issues which are documented here: https://community.openproject.org/wp/65236

## Screenshots
image

- Use a plain turbo frame response to update turbo frames instead of a turbo stream response, because that is incorrect.
- Move the live region methods out of the datepicker controller and place it in the `WorkPackages::DatePicker::DialogContentComponent`
- Replace datepicker component during the morphing process when the Manual / Automatic switch is being toggled. The replacement will result in a re-initialisation too.
- Do not touch the datepicker component when turbo request does not switch between Manual / Automatic.
- As a future improvement, to avoid flickering, we could initialise the new angular tags in the shadow dom and morph them when they are complete.

# Merge checklist

- [X] Added/updated tests

### DIFF
--- a/app/components/work_packages/date_picker/date_form_component.rb
+++ b/app/components/work_packages/date_picker/date_form_component.rb
@@ -32,6 +32,9 @@ module WorkPackages
   module DatePicker
     class DateFormComponent < ApplicationComponent
       include OpPrimer::ComponentHelpers
+
+      FOCUSED_CLASS = "op-datepicker-modal--date-field_current"
+
       attr_reader :work_package
 
       def initialize(work_package:,
@@ -78,7 +81,7 @@ module WorkPackages
           disabled: disabled?(name),
           label:,
           show_clear_button: show_clear_button?(name),
-          classes: "op-datepicker-modal--date-field #{'op-datepicker-modal--date-field_current' if @focused_field == name}",
+          classes: "op-datepicker-modal--date-field #{FOCUSED_CLASS if focused?(name)}",
           validation_message: validation_message(name),
           type: field_type(name),
           placeholder: placeholder(name)
@@ -165,6 +168,10 @@ module WorkPackages
         @disabled
       end
 
+      def focused?(name)
+        @focused_field == name && !disabled?(name)
+      end
+
       def field_value(name)
         errors = @work_package.errors.where(name)
         if (user_value = errors.map { |error| error.options[:value] }.find { !it.nil? })
@@ -206,7 +213,7 @@ module WorkPackages
                          "focus->work-packages--date-picker--preview#onHighlightField",
                  test_selector: "op-datepicker-modal--#{name.to_s.dasherize}-field" }
 
-        if @focused_field == name && !disabled?(name)
+        if focused?(name)
           data[:qa_highlighted] = "true"
           data[:focus] = "true"
         end
--- a/app/components/work_packages/date_picker/date_form_component.rb
+++ b/app/components/work_packages/date_picker/date_form_component.rb
@@ -217,7 +217,7 @@ module WorkPackages
       def single_date_field_button_link(focused_field)
         permitted_params = params.merge(date_mode: "range", focused_field:).permit!
 
-        if params[:action] == "new"
+        if params[:action].in?(["new", "create"])
           new_date_picker_path(permitted_params)
         else
           work_package_date_picker_path(permitted_params)
--- a/app/components/work_packages/date_picker/date_form_component.rb
+++ b/app/components/work_packages/date_picker/date_form_component.rb
@@ -224,10 +224,10 @@ module WorkPackages
       def single_date_field_button_link(focused_field)
         permitted_params = params.merge(date_mode: "range", focused_field:).permit!
 
-        if params[:action].in?(["new", "create"])
-          new_date_picker_path(permitted_params)
+        if work_package.new_record?
+          preview_date_picker_path(permitted_params)
         else
-          work_package_date_picker_path(permitted_params)
+          preview_work_package_date_picker_path(work_package, permitted_params)
         end
       end
 
--- a/spec/features/work_packages/table/scheduling/manual_scheduling_spec.rb
+++ b/spec/features/work_packages/table/scheduling/manual_scheduling_spec.rb
@@ -65,9 +65,9 @@ RSpec.describe "Manual scheduling", :js do
       # Toggle to manual scheduling mode
       datepicker.toggle_scheduling_mode
 
-      # Expect editable in single mode with due date field visible
-      datepicker.expect_add_start_date_button_visible
-      datepicker.expect_due_date ""
+      # Expect editable in single mode with start date field visible
+      datepicker.expect_add_finish_date_button_visible
+      datepicker.expect_start_date ""
       datepicker.expect_cancel_button_enabled
       datepicker.expect_save_button_enabled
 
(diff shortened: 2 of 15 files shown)


spec/features/work_packages/table/scheduling/manual_scheduling_spec.rb
```diff
@@ -65,9 +65,9 @@
       # Toggle to manual scheduling mode
       datepicker.toggle_scheduling_mode
 
-      # Expect editable in single mode with due date field visible
-      datepicker.expect_add_start_date_button_visible
-      datepicker.expect_due_date ""
+      # Expect editable in single mode with start date field visible
+      datepicker.expect_add_finish_date_button_visible
+      datepicker.expect_start_date ""
```
cbliard, line 70: Nice change!

I think that's better that when switching to manual mode after having clicked on start date field in work packages list, it focuses start date field. It makes total sense.

One little thing I would change is the focused field when in automatic mode: before your changes, due date was always the one to be focused when switching from automatic to manual, so when in automatic it was the one being shown as focused, regardless of having clicked on start date field or finish date field first:

image

But now, when switching to manual, it feels weird because finish date was focused when automatic, and when switching to manual it's start date which is focused. Here is a demo:

swtiching to manual changes the focused field

(Note that it works well when finish date field is clicked first).

The code to modify is in `app/controllers/work_packages/date_picker_controller.rb:117`, and there is a twist:
- if the work package is in automatic mode because it's a parent, neither start or finish dates are editable, so it's ok to focus the clicked field and keep that field focused when switching to manual mode
- if the work package is in automatic mode because it's a successor, stat date is not editable but finish date is, so it should be focused when in automatic mode regardless of the field having been clicked first. And when switching to manual mode, honestly I'm not sure if the focus should be kept on finish date or if it's ok to switch focus to start date if it was clicked first. I think it's ok.

That would be nice to have, but it's not required either, we can also register a bug for it to keep a trace of possible improvements.

cbliard (approved):
Really nice improvement 👏 
I saw that additional tag being created each time the mode was switched, but we already spent so much time on it and it had no functional impact so we decided to let it like this.

I'm glad it's finally fixed, and that so much controller code has been removed at the same time. I also like that you were brave enough to dive into the date picker code. 

I tried to make it fail, but apart the bugs I already know about it, it works pretty much the same way as it did before. It just flickers when switching modes.

I have a small remark about focused field when in automatic mode, but for me it's ok to merge as-is.

Great work!

spec/features/work_packages/table/scheduling/manual_scheduling_spec.rb
dombesz: Thanks for the thorough review Christophe!

My proposed solution for the first case (Parent in automatic mode) is to not highlight any disabled fields, we do not and cannot apply native focus on it, so I would say it makes more sense to not highlight it either.

For the second issue, I'm not sure either, maybe we can consult the product team and solve it separately.

Please take a look at the first solution, thanks!

bsatarnejad:
Hi @dombesz 
I found an accessibility issue here:
When changing the scheduling mode and banner is changed as well, banner should be focused and read out by the screen reader, which is failing now:

https://github.com/user-attachments/assets/42154dc3-4b52-44ad-9836-044af67eea43

bsatarnejad (changes requested):
There is an accessibility failure here, while changing the scheduling mode, the banner should be read out by the screen reader.

spec/features/work_packages/table/scheduling/manual_scheduling_spec.rb
cbliard: > My proposed solution for the first case (Parent in automatic mode) is to not highlight any disabled fields, we do not and cannot apply native focus on it, so I would say it makes more sense to not highlight it either.

Totally agree, that makes more sense. I was too conservative here.

> For the second issue, I'm not sure either, maybe we can consult the product team and solve it separately.

Yep, I talked about it in [today's UI/UX Open door.](https://community.openproject.org/projects/gmbh/meetings/5649) The way you implemented it has been approved, and I think overall it makes more sense. Thanks for that.

As agreed with Parimal, I have to write some specs about it to document how it works now. I'll write them in another bug report and reference that bug in this PR if that's ok for you?

app/components/work_packages/date_picker/date_form_component.rb
```diff
@@ -78,7 +81,7 @@ def text_field_options(name:, label:)
           disabled: disabled?(name),
           label:,
           show_clear_button: show_clear_button?(name),
-          classes: "op-datepicker-modal--date-field #{'op-datepicker-modal--date-field_current' if @focused_field == name}",
+          classes: "op-datepicker-modal--date-field #{FOCUSED_CLASS if focused?(name)}",
```
cbliard, line 84: Neat 👌

cbliard (approved):

cbliard:
> There is an accessibility failure here, while changing the scheduling mode, the banner should be read out by the screen reader.

Would it be possible to write automated tests about this?

spec/features/work_packages/table/scheduling/manual_scheduling_spec.rb
dombesz: Sounds good, thank you.

cbliard: I created https://community.openproject.org/projects/openproject/work_packages/65236/activity and referenced it in this PR.

bsatarnejad:
> > There is an accessibility failure here, while changing the scheduling mode, the banner should be read out by the screen reader.
> 
> Would it be possible to write automated tests about this?

Yes, [we set tabindex attribute for the banner ](https://github.com/opf/openproject/blob/dev/frontend/src/stimulus/controllers/dynamic/work-packages/date-picker/preview.controller.ts#L532), connect method was called with any change on the date picker, but now it isn't.

bsatarnejad:
Hey @dombesz @cbliard ,
I found another issue that while opening the date picker, there is an update message read by the screen reader, while there is no update:

https://github.com/user-attachments/assets/dbff97c8-7972-413a-bcc4-6be03c724541

dombesz:
Hi @bsatarnejad,

I addressed the issues you mentioned, I had to introduce the preview actions on the datepicker and progress controller to solve the issue of sending live region updates unnecessarily. The preview actions required in order to differentiate between the calls of a dialog opened by the user, versus the dialog being updated by a change. Now if the dialog is updated by a change, the preview action is called instead of the corresponding new/edit. The progress controller implements the preview action too, just to make it compatible with the inherited stimulus preview controller.

bsatarnejad:
> Hi @bsatarnejad,
> 
> I addressed the issues you mentioned, I had to introduce the preview actions on the datepicker and progress controller to solve the issue of sending live region updates unnecessarily. The preview actions required in order to differentiate between the calls of a dialog opened by the user, versus the dialog being updated by a change. Now if the dialog is updated by a change, the preview action is called instead of the corresponding new/edit. The progress controller implements the preview action too, just to make it compatible with the inherited stimulus preview controller.

Hi @dombesz Thanks, it looks good now, I just found an issue that if it is a single date picker and you want to enter a date(in this example a start date), at first focus is correctly set on the text input, but after a few seconds its label is focused, and you should press tab again to trap the focus on input and enter a date.

https://github.com/user-attachments/assets/c96dc39e-c58a-4eef-9b6c-5e7c71dd3860

bsatarnejad (approved):
👍

PR #18163, opened 2025-03-05
[#61291] Extend AMPF specs with OIDC users.

ba1ash:
# Ticket

https://community.openproject.org/work_packages/61291

# Merge checklist

- [x] Added/updated tests


modules/storages/spec/services/storages/one_drive_managed_folder_sync_service_spec.rb
```diff
@@ -193,6 +218,7 @@
       expect(result.name).to match(/_=o=_ _ _Jedi_ Project Folder ___ \(\d+\)/)
     end
 
+    # rubocop:disable RSpec/ExampleLength
```
NobodysNightmare: I have to admit that I agree to Rubocop here.

The test was previously understandable, though arguably I am wondering whether even the original could have been split already (I can see two times that we do something and then expect on the result of that).

But for the new spec, I have no clue what's going on to be honest:

1. We make very specific expectations on the calls that `SetPermissionsCommand` receives
2. It's hard to follow-up with why exactly these calls are being expected and not different calls

Maybe it would already help to simply split the `it` statement into the different expectations we have, e.g.

```ruby
it "sets permissions for user A to X"
it "sets permissions for user B to Y"
it "sets permissions for user C to Z"
```

Then it could be clearer why we expect these different calls.

modules/storages/spec/services/storages/one_drive_managed_folder_sync_service_spec.rb
```diff
@@ -48,8 +48,31 @@
...
+  shared_let(:oidc_user) do
+    identity_url = "#{oidc_provider.slug}:qweqweqweqwe"
+    create(:user, identity_url:)
+  end
+  shared_let(:oidc_admin) do
+    identity_url = "#{oidc_provider.slug}:zxczxczxczxc"
+    create(:admin, identity_url:)
+  end
+  shared_let(:oidc_user_token) do
+    create(:remote_identity,
```
NobodysNightmare: Why do we create a `remote_identity` but name it `*_token`?

For a variable called `oidc_user_token` I would have expected something that is a `OpenIDConnect::UserToken`, but not a `RemoteIdentity`.

modules/storages/spec/services/storages/one_drive_managed_folder_sync_service_spec.rb
ba1ash: I am not happy about it as well. Looking for a better way...

mereghost: Hmmm... I see where the issue is. We have 2 different distinct tests running side by side.

The original, using VCR, that has actual ids and whatnot and an rely on exercising the entire codebase with no need for mocking.

We also have another one, based on OIDC that isn't covered by VCR (as we don't have it in place) that needs to mock the calls.

My gut reaction would be to isolate all the OIDC tests from the surrounding context and mock it as needed, but that's a lot of mocking of very important code.

My second reaction, is... why do we care? I see a point in making sure that we can get all the `RemoteIdentities` for a storage (a test on a scope of `RemoteIdentities`) but here? Behaviour-wise we don't care from where the remote identity comes from, only that it is found for that storage.

ba1ash: I'd say it is rather important to make sure remote_identites from both auth_sources are involved in AMPF sync. I changed the spec a bit without new users involved. It should be a bit better now. Kindly take a look...

NobodysNightmare: I see that something changed, but I still don't understand the spec and I am not sure whether I would be able to change it, understanding what it's supposed to tell me about our code behaviour.

Let me go back to the suggestion I made earlier: Can we split the spec? For example I see that we expect:

* 4 calls to be made in total to `Storages::Peripherals::StorageInteraction::OneDrive::SetPermissionsCommand`
* we expect four times one specific call to be made, but we don't care about call order or anything where these expectations depend on each other

So what I could imagine is something in this direction:

```ruby
context "with inactive project folder", vcr: "one_drive/sync_service_hide_inactive" do
  # maybe rename to :set_permissions_command or :set_permissions
  let(:clazz) { Storages::Peripherals::StorageInteraction::OneDrive::SetPermissionsCommand }
        
  before do
    original_folder = create_folder_for(inactive_project_storage)
    inactive_project_storage.update(project_folder_id: original_folder.result.id)

    set_permissions_on(original_folder.result.id,
                           [{ user_id: "2ff33b8f-2843-40c1-9a17-d786bca17fba", permissions: [:read_files] },
                            { user_id: "248aeb72-b231-4e71-a466-67fa7df2a285", permissions: [:write_files] },
                            { user_id: "33db2c84-275d-46af-afb0-c26eb786b194", permissions: [:write_files] }])
    allow(clazz).to receive(:call).and_call_original
  end

  it "hides (removes all permissions) from inactive project folders" do
    # essentially what you already have
  end

  it "updates 4 files" do # or do we update 4 remote identities? maybe I misunderstand!
    service.call
    expect(clazz).to have_received(:call).with(
          auth_strategy: an_instance_of(Storages::Peripherals::StorageInteraction::AuthenticationStrategies::Strategy),
          input_data: an_instance_of(Storages::Peripherals::StorageInteraction::Inputs::SetPermissions),
          storage: an_instance_of(Storages::OneDriveStorage)
        ).exactly(4).times
  end

  it "updates file X like this" do
    # one of the .once expectations goes here
  end

  it "updates file Y like that" do
    # another one of the .once expectations goes here
  end

  # ... and so on
end
```

For example right now I don't quite understand why we expect what we expect.
One example: Why does file `01AZJL5PPGH34OKD6VVRGKX47XXIIB2DB2` get `write_files` permissions for all three users, when we set `read_files` for the user with id `2ff33b8f-2843-40c1-9a17-d786bca17fba` before?

I bet there is an explanation for that, and it hopefully fits into the description of an `it`.

ba1ash: @NobodysNightmare kindly have a look at the changes again.

NobodysNightmare (approved):

PR #18164, opened 2025-03-05
Show proper error message even if fetching origin user id fails

NobodysNightmare:
I noticed an additional case where we do not show the intended error message for 61612. Reproduction is comparatively easy:

1. Have a working setup using token **exchange**
2. Misconfigure the storage to use the IDP token instead of exchanging for a token with the right audience (storage will respond with 401 on every request)

This way, when opening the files tab, we'd previously not have shown the files, but would have displayed a red error message about an HTTP 500. The error occured while fetching the remote identity, but it would even have happened if we already knew the remote identity for the user.

This PR fixes two issues:
1. We only fetch the remote identity information, if there is no remote identity yet (causing fewer requests to storage in general, but also making reproduction harder: you need to have **no** remote identity)
2. Fetch errors of the remote identity are now only logged, but do not raise an error across the `OpenProject::Notification` boundary. This allows the UI to show an appropriate error message.

# Ticket
https://community.openproject.org/projects/cross-application-user-integration-stream/work_packages/61986

### DIFF
--- a/app/services/remote_identities/create_service.rb
+++ b/app/services/remote_identities/create_service.rb
@@ -32,36 +32,45 @@ module RemoteIdentities
   class CreateService
     attr_reader :user, :model
 
-    def self.call(user:, integration:, token:)
-      new(user:, integration:, token:).call
+    def self.call(user:, integration:, token:, force_update: false)
+      new(user:, integration:, token:, force_update:).call
     end
 
-    def initialize(user:, integration:, token:)
+    def initialize(user:, integration:, token:, force_update: false)
       @user = user
       @integration = integration
       @token = token
+      @force_update = force_update
 
       @model = RemoteIdentity.find_or_initialize_by(user:, auth_source: token.auth_source, integration:)
     end
 
     def call
-      if @model.new_record?
+      if @model.new_record? || @force_update
         user_id = @integration.extract_origin_user_id(@token)
         return user_id if user_id.failure?
 
         @model.origin_user_id = user_id.result
-        if @model.save
-          OpenProject::Notifications.send(
-            OpenProject::Events::REMOTE_IDENTITY_CREATED,
-            integration: @integration
-          )
-          ServiceResult.success(result: @model)
-        else
-          ServiceResult.failure(result: @model, errors: @model.errors)
-        end
-      else
-        ServiceResult.success(result: @model)
+        return success unless @model.changed?
+        return failure unless @model.save
+
+        OpenProject::Notifications.send(
+          OpenProject::Events::REMOTE_IDENTITY_CREATED,
+          integration: @integration
+        )
       end
+
+      success
+    end
+
+    private
+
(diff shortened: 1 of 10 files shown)


app/services/remote_identities/create_service.rb
```diff
@@ -45,8 +45,11 @@ def initialize(user:, integration:, token:)
     end
 
     def call
-      @model.origin_user_id = @integration.extract_origin_user_id(@token)
-      if @model.changed?
+      if @model.new_record?
```
ba1ash: :red_circle: 
I am not sure about that change. At least the following case is possible which I consider incorrect:
1. OP user has OAuth connection to NC(OAuthClientToken and RemoteIdentity)
2. Then OP user removes OAuthClientToken through UI, RemoteIdentity remains.
3. User goes though OAuth grant flow, but logs in as a different user on NC side.
4. Exisiting RemoteIdentity is not updated and contains wrong origin_user_id.

NobodysNightmare: Agreed. That case is a problem... Let me think about it.

NobodysNightmare: Here's my idea, tell me what you think:

* `RemoteIdentities::CreateService` accepts a flag that forces an update of the remote identity
* ConnectionManager sets that flag, SSO doesn't

In general I think all of this is just a workaround. A really nice thing would be to offer the user management/transparency about remote identities, the same as we show for access tokens for other applications. But as long as we don't have that, this approach could solve most of the issues we see:

* the case you described does not occur anymore
* we do not flood the storage with queries about the user identity from SSO; the OAuth 2.0 case makes sure that we don't do that anyways, by not calling the `CreateService` every time

ba1ash: I'm fine with this idea.

NobodysNightmare: implemented it

ba1ash (approved):
LGTM.

app/services/remote_identities/create_service.rb
```diff
@@ -32,21 +32,25 @@ module RemoteIdentities
...
     end
 
     def call
-      @model.origin_user_id = @integration.extract_origin_user_id(@token)
-      if @model.changed?
+      if @model.new_record? || @force_update
+        user_id = @integration.extract_origin_user_id(@token)
+        return user_id if user_id.failure?
+
+        @model.origin_user_id = user_id.result
```
ba1ash, line 53: Would it make sense to check if model changed to avoid unnecessary event emission? 
```suggestion
        @model.origin_user_id = user_id.result
        return ServiceResult.success(result: @model) unless @model.changed?
```

NobodysNightmare: done

ba1ash: Thanks!

PR #18170, opened 2025-03-05
release notes: replace big GIF with smaller MP4

EinLama:
# Ticket
n/a

In the release notes for 15.0.0, there is a 50 MB GIF file. This PR replaces this with a ~18 MB MP4 video file. The advantage: this file is not only smaller, it is also much clearer, can be paused - and is only downloaded on demand. A huge relief for visitors from limited bandwidth plans.


as-op (changes requested):
👍 
But could you please point it to the `release/15.3` branch? So we don't have to wait until the next release to have the GIF removed from our website.

EinLama:
Happy to do so! But I had to create a new PR for this... See #18171

PR #18177, opened 2025-03-06
[#61936] Can't associate storage to project via storage admin view

ba1ash:
# Ticket

https://community.openproject.org/work_packages/61936

# Merge checklist

- [x] Added/updated tests


modules/storages/app/models/storages/storage.rb
```diff
@@ -119,6 +119,19 @@ def self.extract_part_from_piped_string(text, index)
...
+    def oauth_access_granted?(user)
+      selector = Peripherals::StorageInteraction::AuthenticationMethodSelector.new(
+        storage: self,
+        user:
+      )
+      case selector.authentication_method
+      when :sso
+        true
+      when :storage_oauth
+        OAuthClientToken.exists?(user:, oauth_client: oauth_client)
```
NobodysNightmare: :green_circle: We could use short syntax here

```suggestion
        OAuthClientToken.exists?(user:, oauth_client:)
```

modules/storages/app/models/storages/storage.rb
```diff
@@ -119,6 +119,19 @@ def self.extract_part_from_piped_string(text, index)
       end
     end
 
+    def oauth_access_granted?(user)
```
NobodysNightmare, line 122: :yellow_circle: I have the feeling that we should name this method differently now that it also includes SSO... But I can't come up with a good proposal myself, so I guess I am leaving this thought here, but consider it very optional to do anything about it.

I am thinking about something along the lines of `user_connected?` maybe, but not really satisfied with it.  
Thinking about that naming proposal it's also interesting that sometimes we check the presence of a remote identity (e.g. `authorizationState`) and in other cases (here) we check the presence of certain tokens to decide whether a user is hooked up to a storage.

**tl;dr:** Nothing really to do here, unless you take my feedback as inspiration for something :see_no_evil:

NobodysNightmare (approved):
All of my comments are optional and this PR seems to be doing what it's advertising :)

I found an issue while reviewing the UI unlocked by this PR, but as far as I can tell that's unrelated to this PR.

modules/storages/app/models/storages/storage.rb
ba1ash: I was thinking about the better name, but did not come up with something satisfying either...

PR #18180, opened 2025-03-06
[#61935] added first element to pattern input suggestions

Kharonus:
# Ticket
[OP#61935](https://community.openproject.org/wp/61935)

- first element in suggestion list should always show "Add as text: current_search"

- added new template with different click behavior

### DIFF
--- a/frontend/src/stimulus/controllers/pattern-input.controller.ts
+++ b/frontend/src/stimulus/controllers/pattern-input.controller.ts
@@ -46,6 +46,8 @@ export default class PatternInputController extends Controller {
     'suggestionsHeadingTemplate',
     'suggestionsDividerTemplate',
     'suggestionsItemTemplate',
+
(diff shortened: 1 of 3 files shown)


frontend/src/stimulus/controllers/pattern-input.controller.ts
```diff
@@ -98,7 +107,7 @@ export default class PatternInputController extends Controller {
     }
 
     // close the suggestions
-    if (event.key === 'Escape' || event.key === 'ArrowLeft' || event.key === 'ArrowRight') {
+    if (['Escape', 'ArrowLeft', 'ArrowRight', 'End', 'Home'].includes(event.key)) {
```
brunopagno, line 110: :heart:

brunopagno (approved):
Tested locally and it looks correct to me :tada: :tada:

PR #18188, opened 2025-03-06
dev db tasks to anonymize, and isolate data

machisuji:

**Anonymization**

Sometimes we would like to debug data of some OpenProject instance without exposing any data at all.
For that we have the `script/anonymize-sql-dump` script. The script makes some quite specific assumptions, which can make it tricky to use. 

This is why I've added a rake task that does the same thing but simply on the currently connected database.

```bash
bundle exec rake development:db:anonymize

Anonymizing data in currently connected database (opcore5_dev) is a destructive operation.

You sure about this? y/N
```

**Isolation**

Secondly, there are other times when someone would like to restore a backup or snapshot of their environment and host it to copy over deleted data into the current production environment.

For this it I have added another rake task that leaves the data broadly unchanged but removes everything that would cause the restored environment to communicate with the outside world, e.g. via email. It 'isolates' it, if you will.
If you can come up with a better name, I'm all ears.

```bash
bundle exec rake development:db:isolate

Isolating data in currently connected database (opcore5_dev) is a destructive operation.

You sure about this? y/N
```

### DIFF
--- a/lib/tasks/development.rake
+++ b/lib/tasks/development.rake
@@ -1,0 +1,82 @@
+# frozen_string_literal: true
+
+#-- copyright
+# OpenProject is an open source project management software.
+# Copyright (C) the OpenProject GmbH
+#
+# This program is free software; you can redistribute it and/or
+# modify it under the terms of the GNU General Public License version 3.
+#
+# OpenProject is a fork of ChiliProject, which is a fork of Redmine. The copyright follows:
+# Copyright (C) 2006-2013 Jean-Philippe Lang
+# Copyright (C) 2010-2013 the ChiliProject Team
+#
+# This program is free software; you can redistribute it and/or
+# modify it under the terms of the GNU General Public License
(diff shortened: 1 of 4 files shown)


machisuji:
Even after multiple restarts one spec keeps failing:

```
./modules/meeting/spec/features/meetings_index_spec.rb:240 # Meetings Index when visiting from a global context lists all upcoming meetings for all projects the user is invited to
```

Although this is thoroughly unrelated to the changes in this PR.

NobodysNightmare:
:see_no_evil: :see_no_evil: Just because we had the discussion today :see_no_evil: :see_no_evil: 

I think the SQL and script files, need a GPL header as well.

_ducks and runs away_

machisuji:
> 🙈 🙈 Just because we had the discussion today 🙈 🙈
> 
> I think the SQL and script files, need a GPL header as well.
> 
> _ducks and runs away_

Good point!

PR #18195, opened 2025-03-07
Fix sort direction translation in query component

top4ek:
Fix sort direction translation component

Translate  button popup 

## Screenshots
Screenshot_20250307_144435


oliverguenther (approved):
Good catch, thanks!

PR #18200, opened 2025-03-10
Fix production run after go Rails 7.2

Eric-Guo:

Make OpenProject can run in production mode.

Following Rails 7.2 [method rename](https://github.com/rails/rails/commit/02f66fe3e25cef2b0dec8bca26e777fe788af35c) and remove not available config.

# Merge checklist

### DIFF
--- a/config/environments/production.rb
+++ b/config/environments/production.rb
@@ -164,11 +164,6 @@ Rails.application.configure do
   # Do not dump schema after migrations.
   config.active_record.dump_schema_after_migration = false
 
-  # Silence the following warning
-  # "Rails couldn't infer whether you are using multiple databases from your database.yml"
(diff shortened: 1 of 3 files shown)


klaustopher:
According to the PR you linked, the replacement for `check_pending!` should be `check_all_pending!`. Why did you chose to not use that one and instead use `check_pending_migrations` which is not a publicly documented method?

Eric-Guo:
Nice catch! I do it wrong, not AI tools, I original check the source code and not found the `check_all_pending!`, we need using `check_all_pending!` here as it's public method.

crohr (approved):

klaustopher:
Will merge as soon as the test run is through

PR #18207, opened 2025-03-10
Bug/36559 custom fields on global work package list can be added as column even when user has no access to it

EinLama:
# Ticket
https://community.openproject.org/wp/36559

When listing work package columns, you should not see custom fields that you don't have access to.

With the changes in this PR, the same consistent behavior also applies to all other places, where custom fields are queried. For example, in the XLS export of work packages, in the cost report, and so on.

Sneaked in a couple of mini-corrections for things that RuboCop critized or that I stumbled upon while going through the code.

Apply the existing `WorkPackageCustomField.visible_by_user` scope to `User.current` when building the instances for `WorkPackageCustomFields`.

I am not _entirely_ sure that this is the correct place to apply the permission check. There might be another place that is more suitable. However, this is the best place that I could find.

# Merge checklist

- [x] Added/updated tests

### DIFF
--- a/app/controllers/concerns/accounts/current_user.rb
+++ b/app/controllers/concerns/accounts/current_user.rb
@@ -1,3 +1,5 @@
+# frozen_string_literal: true
+
 #-- copyright
 # OpenProject is an open source project management software.
 # Copyright (C) the OpenProject GmbH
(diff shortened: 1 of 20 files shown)


ulferts (approved):
👍

PR #18107, opened 2025-02-27
Feature/59540 highlight closest predecessor in relations tab component snapshot specs

myabc:
# Ticket

Relates to https://community.openproject.org/wp/60910

# What are you trying to accomplish?

Make component testing less painful.

## Screenshots

# What approach did you choose and why?

# Merge checklist

- [ ] Added/updated tests
- [ ] Added/updated documentation in Lookbook (patterns, previews, etc)
- [ ] Tested major browsers (Chrome, Firefox, Edge, ...)

PR #18208, opened 2025-03-10
[#61885] Sort relations in relations tab by created_at

myabc:
# Ticket

https://community.openproject.org/wp/61885
https://community.openproject.org/wp/61950

Update the sort order of relations in the relations tab and date picker by relation creation date, oldest first.
Mix visible and ghost relations together instead of having visible first and then ghost.

## Screenshots

image

This is the simplest possible implementation, with the fewest code changes. I did explore some refactoring to try to reduce the number of times we iterate over the relations, but didn't come up with a solution that I liked.

This PR also adds a bunch of component tests - these could certainly be DRY'ed up, but it's a start (see #18107 for an aborted attempt to use test these components with snapshot testing).

# Merge checklist

- [x] Added/updated tests

### DIFF
--- a/app/components/work_package_relations_tab/index_component.html.erb
+++ b/app/components/work_package_relations_tab/index_component.html.erb
@@ -57,6 +57,7 @@
           # Combine visible and invisible relations into a single list
           all_relations = relation_group.visible_relations.map { |r| [r, :visible] } +
             relation_group.ghost_relations.map { |r| [r, :ghost] }
+          all_relations.sort_by! { |r| r[0].id }
 
           flex.with_row(mb: 4) do
             render_relation_group(
@@ -84,6 +85,7 @@
           # Combine visible and invisible children into a single list
           all_children = visible_children.map { |r| [r, :visible] } +
             ghost_children.map { |r| [r, :ghost] }
+          all_children.sort_by! { |r| r[0].created_at }
 
           flex.with_row do
             render_relation_group(
--- a/spec/support/capybara/additional_accessible_selectors.rb
+++ b/spec/support/capybara/additional_accessible_selectors.rb
@@ -29,13 +29,29 @@
 #++
 
 Capybara.add_selector(:list) do
-  xpath { ".//ul | .//ol" }
+  xpath do |*|
+    XPath.descendant[[
+      XPath.self(:ul),
+      XPath.self(:ol)
+    ].reduce(:|)]
+  end
+
+  locator_filter skip_if: nil do |node, locator, exact:, **|
+    method = exact ? :eql? : :include?
+    if node[:"aria-labelledby"]
+      CapybaraAccessibleSelectors::Helpers.element_labelledby(node).public_send(method, locator)
+    elsif node[:"aria-label"]
+      node[:"aria-label"].public_send(method, locator.to_s)
+    end
+  end
 end
 
 Capybara.add_selector(:list_item) do
   label "list item"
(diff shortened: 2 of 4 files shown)


myabc:
@cbliard going ahead and putting this in review, even though I'm not 100% happy with the component specs as they are rn.

cbliard:
> @cbliard going ahead and putting this in review, even though I'm not 100% happy with the component specs as they are rn.

Thanks, I'll have a look.
What does "rn" mean?

myabc:
> What does "rn" mean?

Right now.

spec/support/capybara/additional_accessible_selectors.rb
```diff
@@ -29,13 +29,29 @@
 #++
 
 Capybara.add_selector(:list) do
-  xpath { ".//ul | .//ol" }
+  xpath do |*|
+    XPath.descendant[[
+      XPath.self(:ul),
+      XPath.self(:ol)
+    ].reduce(:|)]
+  end
```
cbliard, line 37: I'm not sure I see the difference here. Is it behaving the same or did you fix the xpath that was used.

To be honest, I'm not sure I find it more readable or maintainable. It feels like obfuscated code to me.
It's a bit like if there was a regular expression `/(a+|b+)/` that was modified into `Regexp.Capture[[Regexp.many("a"), Regexp.many("b")].reduce(:|)]`. That's another language to learn when one already exists, and the mapping is never perfect.

And [XPath library documentation](https://www.rubydoc.info/gems/xpath/3.2.0) is very terse.

spec/support/capybara/additional_accessible_selectors.rb
```diff
@@ -29,13 +29,29 @@
...
 end
 
 Capybara.add_selector(:list_item) do
   label "list item"
 
-  xpath { ".//li" }
+  xpath do |*|
+    XPath.descendant[XPath.self(:li)]
+  end
```
cbliard, line 54: Same here.

app/components/work_package_relations_tab/index_component.html.erb
```diff
@@ -57,6 +57,7 @@
           # Combine visible and invisible relations into a single list
           all_relations = relation_group.visible_relations.map { |r| [r, :visible] } +
             relation_group.ghost_relations.map { |r| [r, :ghost] }
+          all_relations.sort_by! { |r| r[0].to.created_at }
```
cbliard: I wonder if ordering by related work package creation date is actually correct. I would expect a newly added work package to be listed at the bottom of the list (like before the relations tab redesign).

This could be achieved if we sort by `r[0].id` instead of `r[0].to.created_at` (because Relation does not have any `created_at`).

I think that's worth asking at the next UI/UX meeting.

On the other hand, ordering by relation creation date is something we can't achieve with child relations, unless we start introducing a `parent_updated_at` field.

cbliard (commented):
It works as specified, and that's nice that you added component tests along it.

Regarding the order, I would prefer ordering by relation creation, oldest first, like it was before, rather than by related work package creation date. But maybe that's just me being used to how it worked before.

I'll add a point for dicussion at next UI/UX meeting.

app/components/work_package_relations_tab/index_component.html.erb
cbliard: I amended last commit to use `r[0].id` and adapted tests.

PR #18209, opened 2025-03-10
Ensure that sorting of locales always works

NobodysNightmare:
This intends to fix rarely occuring errors on that line according to AppSignal:

    comparison of String with :en failed
    Backtrace: app/seeders/basic_data/setting_seeder.rb:72:in 'Array#sort'

I don't know how the symbol ended up in the array, there is possibly a root cause elsewhere to be found. However, considering that this line of code already acknowledged that locales are sometimes represented as a symbol, it is probably sensible to convert the whole array to its string representation anyways.

# Ticket
https://appsignal.com/openproject-gmbh/sites/678fa146338a5d816f8d696c/exceptions/incidents/260

Prevent seeding errors that are repeatedly observed in the wild.

Ensure the error is prevented where it occurs, mostly because I have a hard time following the issue to its root.
If someone has an idea where the symbol leaks into the settings, that place might be the better location for a fix.

### DIFF
--- a/app/seeders/basic_data/setting_seeder.rb
+++ b/app/seeders/basic_data/setting_seeder.rb
@@ -69,7 +69,7 @@ module BasicData
         settings["commit_fix_status_id"] = status_closed.try(:id)
 
         # Add the current locale to the list of available languages
-        settings["available_languages"] = (Setting.available_languages + [I18n.locale.to_s]).uniq.sort
+        settings["available_languages"] = (Setting.available_languages + [I18n.locale]).map(&:to_s).uniq.sort
 
         settings.compact
       end


NobodysNightmare:
Re Rubocop:

> Assignment Branch Condition size for data is too high. [<10, 21, 3> 23.45/17]

I am inclined to not fix this issue in the scope of this PR. 23.45/17 means that the method was already too large before my change (for which I am not sure, whether it counts as an increase to ABC size or whether it's neutral).

Fixing this offense would require a proper refactoring of the `data` method, which I'd avoid to commit to a release branch.

machisuji (approved):
👍

PR #18215, opened 2025-03-10
Display SAML metadata at the end, not at the start

NobodysNightmare:
This is more consistent with the placement on the end in the horizontal layout.

# Ticket
https://community.openproject.org/projects/cross-application-user-integration-stream/work_packages/58241

# Notes

While fixing the reported issue was straight-forward, I still asked for feedback from @psatyal, whether the fix being asked for is really offering the best possible UX. At least the current sort order seems to have been explicitly chosen at some point.

# Screenshots

Before:

image

After:

image

### DIFF
--- a/modules/auth_saml/app/views/saml/providers/show.html.erb
+++ b/modules/auth_saml/app/views/saml/providers/show.html.erb
@@ -33,7 +33,7 @@
       render Saml::Providers::ViewComponent.new(@provider, view_mode: :show)
     end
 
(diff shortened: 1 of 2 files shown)


HDinger (commented):
The changes look fine, but I'd like to get agreement from the product team before the merge.

NobodysNightmare:
Closing, because we agreed that the current ordering is fine.

PR #18217, opened 2025-03-10
Feature flag for comments with restricted visibility

brunopagno:
# Ticket

https://community.openproject.org/projects/communicator-stream/work_packages/62009/activity

### DIFF
--- a/config/initializers/permissions.rb
+++ b/config/initializers/permissions.rb
@@ -313,25 +313,29 @@ Rails.application.reloader.to_prepare do
                      {},
                      permissible_on: %i[work_package project],
                      require: :loggedin,
-                     dependencies: :view_work_packages
+                     dependencies: :view_work_packages,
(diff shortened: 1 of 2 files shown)


config/initializers/permissions.rb
```diff
@@ -313,25 +313,29 @@
                      {},
                      permissible_on: %i[work_package project],
                      require: :loggedin,
-                     dependencies: :view_work_packages
+                     dependencies: :view_work_packages,
+                     visible: -> { OpenProject::FeatureDecisions.comments_with_restricted_visibility_active? }
```
akabiru, line 317: Nice, I didn't know this was a thing 👍🏾

akabiru (approved):

config/initializers/permissions.rb
brunopagno: I found out on this file when looking for other feature flags :sweat_smile:

PR #18218, opened 2025-03-10
Missing translation prevents dev. rendering of copy form

NobodysNightmare:
# Ticket
(I decided against creating a bug for this, because it purely seems to affect development)

Fix rendering of WP copy form that fails due to missing translation.

Adding the missing translation (though we probably never show this attribute to the user).

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -1146,6 +1146,7 @@ en:
           title: "Working days"
           false: "working days only"
           true: "include non-working days"
+        lock_version: Lock version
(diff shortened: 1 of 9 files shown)


klaustopher:
This looks like it's the same as #18219 but based on some different branch and bringing in extra commits?

NobodysNightmare:
> based on some different branch and bringing in extra commits?

meh, yeah I wanted to target 15.4, but didn't.

I am going to close mine in favor of the other. Thanks for cross-linking

PR #18219, opened 2025-03-10
fix lock_version i18n error

ulferts:
# Ticket

N/A

Avoid running into an 500 (in development) as the lock_version attribute of WorkPackages is not i18ned. This can be reproduced when trying to copy a work package.

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -1714,6 +1714,7 @@ en:
     journal_notes: "Comment"
     lastname: "Last name"
     login: "Username"
+    lock_version: "Lock version"
     mail: "Email"
     name: "Name"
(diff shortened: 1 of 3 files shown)


klaustopher (approved):

NobodysNightmare:
General feedback: I'd ask whether we should include this fix in the scope of 15.4, because I am pretty sure that the corresponding changes to attribute localization that caused this issue were introduced with 15.4 (before the branch-off of `release/15.4`)

NobodysNightmare (approved):

ulferts:
@NobodysNightmare good point. I rebased and will merge once the CI agrees

PR #18223, opened 2025-03-10
Code maintenance/61352 upgrade rails to 80

ulferts:
# Ticket

Bump ruby to 8.0 without functional changes.

### DIFF
--- a/config/initializers/filter_parameter_logging.rb
+++ b/config/initializers/filter_parameter_logging.rb
@@ -34,5 +34,5 @@
 # Use this to limit dissemination of sensitive information.
 # See the ActiveSupport::ParameterFilter documentation for supported notations and behaviors.
 Rails.application.config.filter_parameters += %i[
-  passw email secret token _key crypt salt certificate otp ssn cvv cvc
+  passw mail secret token _key crypt salt certificate otp ssn cvv cvc
 ]
--- a/config/initializers/filter_parameter_logging.rb
+++ b/config/initializers/filter_parameter_logging.rb
@@ -1,3 +1,5 @@
+# frozen_string_literal: true
+
 #-- copyright
 # OpenProject is an open source project management software.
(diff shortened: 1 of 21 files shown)


config/initializers/filter_parameter_logging.rb
```diff
@@ -32,5 +34,5 @@
 # Use this to limit dissemination of sensitive information.
 # See the ActiveSupport::ParameterFilter documentation for supported notations and behaviors.
 Rails.application.config.filter_parameters += %i[
-  passw secret token _key crypt salt certificate otp ssn
+  passw email secret token _key crypt salt certificate otp ssn cvv cvc
```
EinLama: I suggest we filter for `mail` here instead, as the attribute on the `User` model is called `mail`. Additionally, this might also future-proof us for some variants, such as `e_mail`.

```suggestion
  passw mail secret token _key crypt salt certificate otp ssn cvv cvc
```

EinLama (changes requested):
Looks and works great as far as I can tell 👏🏻 I dropped one remark about filtered log parameters for emails.

With Rails 8.0, [a Ruby version of at least 3.2.0 is required](https://guides.rubyonrails.org/upgrading_ruby_on_rails.html#ruby-versions). I wonder: do we need to specify this in our 15.5.0 version documentation explicitly? Do our release packages have to be reconfigured accordingly?

ulferts:
> With Rails 8.0, [a Ruby version of at least 3.2.0 is required](https://guides.rubyonrails.org/upgrading_ruby_on_rails.html#ruby-versions). I wonder: do we need to specify this in our 15.5.0 version documentation explicitly? Do our release packages have to be reconfigured accordingly?

Our application already requires ruby 3.4 and the packages/docker files are configured for that. So I don't see the need for further changes and documentation.

EinLama (approved):
🚀

PR #18233, opened 2025-03-11
Respect month locale in Gantt pdf export

top4ek:
# Ticket
TBD

While exporting Gantt to PDF month name is always short-english format.

Build month header inside PDF with I18n locale methods.


as-op (approved):
Thank you! 👍

PR #18239, opened 2025-03-11
[#62040] correct input of <,> characters

Kharonus:
# Ticket
[OP#62040](https://community.openproject.org/wp/62040)

- allow usage of `<>` for pattern input
- rendered token show localized text

- escape <> on rendering html from blueprint
- render token with localized text
  - property is stored in `data-prop` instead of the text

### DIFF
--- a/frontend/src/stimulus/controllers/pattern-input.controller.ts
+++ b/frontend/src/stimulus/controllers/pattern-input.controller.ts
@@ -71,15 +71,14 @@ export default class PatternInputController extends Controller {
   declare readonly suggestionsInitialValue:Record<string, Record<string, string>>;
   declare readonly insertAsTextTemplateValue:string;
 
-  validTokens:string[];
+  validTokenMap:Record<string, string> = {};
   currentRange:Range|undefined = undefined;
 
   connect() {
-    this.contentTarget.innerHTML = this.toHtml(this.patternInitialValue) || ' ';
-    this.validTokens = Object.values(this.suggestionsInitialValue)
-      .map((group) => (Object.keys(group)))
-      .reduce((acc, val) => acc.concat(val), []);
+    this.validTokenMap = Object.values(this.suggestionsInitialValue)
+      .reduce((acc, val) => ({ ...acc, ...val }), {});
 
+    this.contentTarget.innerHTML = this.toHtml(this.patternInitialValue) || ' ';
     this.tagInvalidTokens();
     this.clearSuggestionsFilter();
   }
@@ -255,11 +254,12 @@ export default class PatternInputController extends Controller {
   }
 
   private endsWithToken():boolean {
-    return this.contentTarget.innerHTML.endsWith('>');
+    const nodes = this.contentTarget.childNodes;
+    return this.isToken(nodes[nodes.length - 1]);
   }
 
   private startsWithToken():boolean {
-    return this.contentTarget.innerHTML.startsWith('<');
+    return this.isToken((this.contentTarget.childNodes)[0]);
   }
 
   private insertToken(tokenElement:HTMLElement) {
@@ -378,11 +378,11 @@ export default class PatternInputController extends Controller {
   }
 
   private tagInvalidTokens():void {
-    this.contentTarget.querySelectorAll('[data-role="token"]').forEach((element) => {
-      const token = element.textContent?.trim();
-
+    this.contentTarget.querySelectorAll('[data-role="token"]').forEach((element:HTMLElement) => {
       let exists = false;
-      this.validTokens.forEach((prop) => { if (prop === token) { exists = true; } });
+      Object.keys(this.validTokenMap).forEach((prop) => {
+        if (prop === element.dataset.prop) { exists = true; }
+      });
 
       if (exists) {
         element.classList.remove('Label--danger');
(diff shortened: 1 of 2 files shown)


Kharonus:
This is actually a bugfix, but while fixing it I had a nice idea of how to already rendering localized texts.

I added a pull preview, so that you can test it. Especially @dominic-braeunlein. The question is now, is it better then before. I know, that this is not the final state we want to reach.

brunopagno (approved):
I checked out on a local branch and tested it out. It looks pretty good!

And the code approach is what I had in mind. To me it looks like a win :clap:

frontend/src/stimulus/controllers/pattern-input.controller.ts
```diff
@@ -395,28 +395,36 @@ export default class PatternInputController extends Controller {
...
-        result += `{{${node.textContent?.trim()}}}`;
+        result += `{{${(node as HTMLElement).dataset.prop}}}`;
       }
     });
     return result.trim();
   }
 
+  private isToken(node:ChildNode):boolean {
+    return node.nodeType === Node.ELEMENT_NODE && (node as HTMLElement).dataset.role === 'token';
+  }
```
myabc:
```suggestion
  private isToken(node:ChildNode):node is TokenElement {
    return this.isElement(node) && node.dataset.role === 'token';
  }

  private isText(node:Node):node is Text {
    return node.nodeType === Node.TEXT_NODE; 
  }

  private isElement(node:Node):node is HTMLElement {
    return node.nodeType === Node.ELEMENT_NODE;
  }
```

frontend/src/stimulus/controllers/pattern-input.controller.ts
```diff
@@ -395,28 +395,36 @@ export default class PatternInputController extends Controller {
...
-    this.contentTarget.childNodes.forEach((node:Element) => {
+    this.contentTarget.childNodes.forEach((node:ChildNode) => {
       if (node.nodeType === Node.TEXT_NODE) {
         // Plain text node
         result += node.textContent;
-      } else if (node.nodeType === Node.ELEMENT_NODE && (node as HTMLElement).dataset.role === 'token') {
+      } else if (this.isToken(node)) {
         // Token element
-        result += `{{${node.textContent?.trim()}}}`;
+        result += `{{${(node as HTMLElement).dataset.prop}}}`;
```
myabc:
```suggestion
    this.contentTarget.childNodes.forEach((node:ChildNode) => {
      if (this.isText(node)) {
        // Plain text node
        result += node.textContent;
      } else if (this.isToken(node)) {
        // Token element
        result += `{{${node.dataset.prop}}}`;
```

frontend/src/stimulus/controllers/pattern-input.controller.ts
```diff
@@ -395,28 +395,36 @@ export default class PatternInputController extends Controller {
   private createToken(value:string):HTMLElement {
     const templateTarget = this.tokenTemplateTarget.content?.cloneNode(true) as HTMLElement;
     const contentElement = templateTarget.firstElementChild as HTMLElement;
-    contentElement.innerText = value;
+    contentElement.dataset.prop = value;
+    contentElement.innerText = this.validTokenMap[value];
     return contentElement;
```
myabc:
```suggestion
    const templateTarget = this.tokenTemplateTarget.content?.cloneNode(true) as TokenElement;
    const contentElement = templateTarget.firstElementChild as TokenElement;
    contentElement.dataset.prop = value;
    contentElement.innerText = this.validTokenMap[value];
    return contentElement;
```

frontend/src/stimulus/controllers/pattern-input.controller.ts
```diff
@@ -378,11 +378,11 @@ export default class PatternInputController extends Controller {
...
   private tagInvalidTokens():void {
-    this.contentTarget.querySelectorAll('[data-role="token"]').forEach((element) => {
-      const token = element.textContent?.trim();
-
+    this.contentTarget.querySelectorAll('[data-role="token"]').forEach((element:HTMLElement) => {
       let exists = false;
-      this.validTokens.forEach((prop) => { if (prop === token) { exists = true; } });
+      Object.keys(this.validTokenMap).forEach((prop) => {
+        if (prop === element.dataset.prop) { exists = true; }
+      });
```
myabc: It's not possible to break out of a `forEach` loop early:

```suggestion
let exists = Object.keys(this.validTokenMap).some((prop) => prop === element.dataset.prop);
```

frontend/src/stimulus/controllers/pattern-input.controller.ts
myabc: (`isText` and `isElement` might be better defined as utility functions somewhere - and for that matter, `isWhitespace` too)

myabc (commented):
You might consider defining a `TokenElement` type (locally) which would let you make use of user-defined type guards and reduce the number of `as` type assertions:

```ts
type TokenElement = HTMLElement & { dataset: { role: 'token', prop: string } };
```

frontend/src/stimulus/controllers/pattern-input.controller.ts
Kharonus: true, way better loop

frontend/src/stimulus/controllers/pattern-input.controller.ts
Kharonus: IMHO the functions are too small and too trivial yet, to be extracted into utility functions.

Extracted the other functions, though, as suggested.

Kharonus:
thanks @myabc for some nice hints for better typing.

PR #18241, opened 2025-03-11
Improve rendering of floating actions on focus

NobodysNightmare:
Floating actions will now be displayed when the wrapper is hovered OR has focus inside it. Attachment and file links views will hide their right-hand side content accordingly.

Previously the file links did not hide their content at all, leading to floating actions being drawn on top of that content.

On the other hand, attachments did hide their content, but without showing the floating actions. Those only became visible once moving the focus onto them.

# Ticket
https://community.openproject.org/projects/document-workflows-stream/work_packages/62113

## Screenshots
focus-navigation


Kharonus (approved):
LGTM, nice catch!

PR #18144, opened 2025-03-04
Implementation/60979 add capability to create work package comments with restricted visibility (UI)

akabiru:
# Ticket

https://community.openproject.org/work_packages/60979

#18091

# What are you trying to accomplish?

Introduce work package comments with restricted visibility- access controlled via dedicated permissions

> [!IMPORTANT]
> The UI is still undergoing refinements- this PR provides base functionality. OPEN POINTS exist for updating the UI and managing the odd [editor shrinking](https://community.openproject.org/projects/communicator-stream/work_packages/61953/github) behaviour

- [x] View restricted comments
   - [x] Color restricted comments
   - [ ] Distinguish associated comment restricted journals (such as attachments) https://community.openproject.org/wp/61874
   - [x] Filter out restricted comments via permissions
- [x] Edit restricted comments (cannot change restriction)
- [x] Add restricted comments
   - [x] Change background based on restriction
   - [ ] 🐛 primer color vars do not scale to dark mode (becomes invisible) https://community.openproject.org/wp/61949
   - [x] 🐛 on submit remove background color
   - [x] add border top color (restricted)
- [x] https://github.com/opf/openproject/pull/18245

## Screenshots

https://github.com/user-attachments/assets/629aac33-c0a9-4e48-98a5-8315ef379786

# Merge checklist

- [x] Added/updated tests
- [ ] Added/updated documentation in Lookbook (patterns, previews, etc)
- [x] Tested major browsers (Chrome, Firefox, Edge, ...)

PR #18245, opened 2025-03-11
Extend comments with restricted visibility feature flag

akabiru:
https://community.openproject.org/wp/62009

When feature flag is off- restricted comments are hidden away completely; they cannot be added nor viewed. @brunopagno I think this would be the correct route- rather than allowing view rights.

Follows #18131, #18144


brunopagno (approved):
Fair enough. I think this makes sense.

Thanks for adjusting it :bow:

PR #18255, opened 2025-03-12
Ensure to only store tokens for OIDC logins

NobodysNightmare:
The hook we are using is firing for all kinds of omniauth logins. By ensuring that we have a matching OpenIDConnect provider, we know that we are currently performing an OIDC login, which is the only case where we want to try storing OIDC tokens.

# Ticket
https://community.openproject.org/projects/cross-application-user-integration-stream/work_packages/62086

### DIFF
--- a/modules/openid_connect/lib/open_project/openid_connect/hooks/hook.rb
+++ b/modules/openid_connect/lib/open_project/openid_connect/hooks/hook.rb
@@ -1,3 +1,5 @@
+# frozen_string_literal: true
+
 #-- copyright
 # OpenProject is a project management system.
 # Copyright (C) the OpenProject GmbH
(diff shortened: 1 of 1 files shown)


NobodysNightmare:
> Error: Key Validation Error: runtime-logs-bug/62086-trying-to-create-oidc-user-tokens,-even-for-non-oidc-providers-2484c2e58b6266c84f4de94272ee944d58b12db0 cannot contain commas.

This is what happens if you copy the proposed branch names from OpenProject ^^

TODO: Bug report

ba1ash:
WDYT about adding a test for that?

ba1ash (commented):
WDYT about adding a test?
Not sure about the place for it... Maybe spec/services/authentication/omniauth_service_spec.rb or modules/openid_connect/spec/requests/openid_connect_spec.rb?

NobodysNightmare:
@ba1ash Will look into this. However, the tests can't succeed on this branch, because it contains a comma... So I am going to open a new PR :see_no_evil:

PR #18260, opened 2025-03-12
[62125] Avoid jumping to a different month when setting flatpickr dates

cbliard:
# Ticket

https://community.openproject.org/wp/62125

When setting flatpickr dates with `.setDates`, it jumps the calendar to the last date so that it's visible on the left side. This is annoying when clicking on a date on the right side because the calendar view then changes while it could stay the same.

This PR tries to avoid such jumping

With the previous fix of bug 61961 done in PR #18234 (commit 752c83b), there were issues and it was not jumping at all in some cases where it should have.

## Screenshots

Before fix of bug 61961, it would blindly jump to finish date to display it on the left side

[weird jumping to finish date.webm](https://github.com/user-attachments/assets/fa741322-d385-43b7-8e6e-e0000444f735)

[broken jump datepicker.webm](https://github.com/user-attachments/assets/653a9662-f582-472d-9fa0-c6401780c098)

After fix of bug 61961, it was not jumping anymore when setting dates through the input fields or "Today" buttons

[calendar not jumping to selected dates.webm](https://github.com/user-attachments/assets/04bf5fbf-4d47-4f24-bbd9-84717ba2e133)

Now, I should be correct and jump only to the selected date if it's not already visible.

[jumping correctly.webm](https://github.com/user-attachments/assets/4e446296-3fd9-4337-87ae-4df29f5b660c)

This commit fixes the issue in two steps:
- first it jumps to the date that has been changed, so that when setting the start date, the calendar shows the start date instead of showing the finish date
- then if the month previously displayed was on the right side, it jumps back to the previously displayed month so that the calendar view stays the same.

Also it does not jump to different dates if no dates have changed (for instance when focusing some other fields).

# Merge checklist

- [x] Added/updated tests
  - Reenabled 2 tests that were broken due to this bug 

### DIFF
--- a/frontend/src/app/shared/components/datepicker/wp-date-picker-modal/wp-date-picker-instance.component.ts
+++ b/frontend/src/app/shared/components/datepicker/wp-date-picker-modal/wp-date-picker-instance.component.ts
@@ -125,15 +125,12 @@ export class OpWpDatePickerInstanceComponent extends UntilDestroyedMixin impleme
   ) {
     const details = event.detail;
 
-    // try to set dates only if needed because if we set flatpickr dates,
-    // flatpickr jumps the calendar date to the due date, which is annoying.
-    if (this.isDifferentDates(details.dates, details.mode)) {
-      [this.startDateValue, this.dueDateValue] = details.dates;
-      const currentMonth = this.datePickerInstance.datepickerInstance.currentMonth;
-
-      this.datePickerInstance.setDates(details.dates);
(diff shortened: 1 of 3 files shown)


cbliard:
I thought it would also fix [Bug #62012 Clicking on 'Start date' resets the calendar view](https://community.openproject.org/wp/62012) but actually the date picker is reinitialized when clicking that button, and that resets the shown months. I could not find where the reinitialization comes from.

oliverguenther (approved):

HDinger (approved):

PR #18265, opened 2025-03-12
attempt to fix flickering activity spec

ulferts:
The specs fail with

```
Failure/Error: wait_for { page }.to have_test_selector("op-wp-activity-tab")
expected page to have test selector op-wp-activity-tab
```

The reason for this seems to be the usage of wait_for in combination with aggregate_failures. If the selector is not found, wait_for will correctly wait until it eventually is found. But the aggregate_failures will already note the error on the element not being found. So even if the test later on continues successfully, the error from the waiting is logged and returned as the test result.


akabiru (approved):
👏🏾 👏🏾 Amazing find @ulferts - thank you!

I think this might explain the "unexplainable" failures 😄 cc @brunopagno

PR #18270, opened 2025-03-12
Use settings based application name when generating costs report

top4ek:
Remove hardcoded "OpenProject" title from cover page in costs report PDF export. 

Take instance administrator defined name from Settings.app_title. It's "OpenProject" by default.


ulferts (approved):
🙇

PR #18271, opened 2025-03-12
Code maintenance/61352 rails 5 2 defaults

ulferts:
# Ticket

Bump default config used to rails 5.2 standards.

The algorithm for  ETag is already bumped to 7.0 as this avoids a back and forth.


config/application.rb
```diff
@@ -209,7 +209,7 @@ class Application < Rails::Application
     Rails.application.config.active_support.use_authenticated_message_encryption = true
 
     # Use SHA-1 instead of MD5 to generate non-sensitive digests, such as the ETag header.
```
EinLama: The comment is outdated now and should be updated to:
```ruby
# Use SHA-256 instead of MD5 to generate non-sensitive digests, such as the ETag header.
```

EinLama (approved):
Looks good! I double-checked possibly missing occurrences of `cache_key`-calls. But all that I could find were custom implementations by us that can be kept.

Added a nitpick about an outdated hashing-algorithm comment.

All else looks sane 👍🏻 Good to go :shipit: We might not want to merge this on a Friday afternoon, though 😁

config/application.rb
ulferts: Well spotted. I changed the comment.

PR #18276, opened 2025-03-12
Change generation of git branch names to only allow `[a-zA-Z\-_]`

klaustopher:
## Ticket
https://community.openproject.org/work_packages/62164

Change the git branch name generation to only allow latin ascii characters

### DIFF
--- a/modules/github_integration/frontend/module/git-actions/git-actions.service.ts
+++ b/modules/github_integration/frontend/module/git-actions/git-actions.service.ts
@@ -36,14 +36,12 @@ export class GitActionsService {
     // See https://stackoverflow.com/a/3651867 for how these rules came in.
     // This sanitization tries to be harsher than those rules
     return str
-      .replace(/&/g, 'and ') // & becomes and
-      .replace(/ +/g, '-') // Spaces become dashes
-      .replace(/[\x00-\x1F]/g, '') // ASCII control characters are out
-      .replace(/\x7F/g, '') // DEL is out
-      .replace(/[#\\\/\?\*\~\^\:\{\}@\.\[\]'"]/g, '') // Some other characters with special rules are out
-      .replace(/^[-]+/g, '') // Dashes at the start are removed
-      .replace(/[-]+$/g, '') // Dashes at the end are removed
-      .replace(/-+/g, '-') // Multiple dashes in a row are deduped
+      .replace(/&/g, "and ") // & becomes and
+      .replace(/ +/g, "-") // Spaces become dashes
+      .replace(/[^\w\-]+/g, "") // Remove any non ascii characters as they might make trouble in some tools
+	  .replace(/^[-]+/g, "") // Dashes at the start are removed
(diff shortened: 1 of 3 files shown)


modules/github_integration/frontend/module/git-actions/git-actions.service.ts
```diff
@@ -36,14 +36,12 @@ export class GitActionsService {
...
-      .replace(/[#\\\/\?\*\~\^\:\{\}@\.\[\]'"]/g, '') // Some other characters with special rules are out
-      .replace(/^[-]+/g, '') // Dashes at the start are removed
-      .replace(/[-]+$/g, '') // Dashes at the end are removed
-      .replace(/-+/g, '-') // Multiple dashes in a row are deduped
+      .replace(/&/g, "and ") // & becomes and
+      .replace(/ +/g, "-") // Spaces become dashes
+      .replace(/[^\w\-]+/g, "") // Remove any non ascii characters as they might make trouble in some tools
+	  .replace(/^[-]+/g, "") // Dashes at the start are removed
+      .replace(/[-]+$/g, "") // Dashes at the end are removed
+      .replace(/-+/g, "-") // Multiple dashes in a row are deduped
```
cbliard: Some steps could be combined and simplified.

```suggestion
      .replace(/\W+/g, "-") // Replace any consecutive non ascii characters by a single dash as they might make trouble in some tools.
      .replace(/^-/g, "") // Dash at the start is removed
      .replace(/-$/g, "") // Dash at the end is removed
```

klaustopher: This change will leave duplicate dashes in, won't it?

cbliard: No because it replaces all consecutive non ascii characters (`\W+`) by a single dash. `-` is included in `\W`, so in something like `hello - -- world`, `\W+` will match ` - -- ` so the result will be `hello-world`

```js
> "--  hello, this is some weird - fooo - underscore _ stuff  --".replace(/\W+/g, "-")
'-hello-this-is-some-weird-fooo-underscore-_-stuff-'
```

klaustopher: Yes, good one, overlooked the new `\W+` in your solution. Let's try and see if the tests are happy :)

cbliard (approved):
It looks good.

One regression we'll get is that any non-latin character will now be replaced by a `-`. For instance if subject name is "こんにちは" or "Всем привет" then the subject in the branch name will be "-". It will even be "" (empty) because we remove dashes at both ends of the string.

That's correct with regards to the [recommendation from GitHub to only use a limited set of characters](https://docs.github.com/en/get-started/using-git/dealing-with-special-characters-in-branch-and-tag-names#naming-branches-and-tags).

And the branch name will still be unique: it will be something like "user-story/1234-" so it's not that bad.

Maybe we could use a lib like [transliteration](https://www.npmjs.com/package/transliteration) in the future if some users complain about it.

PR #18277, opened 2025-03-12
[62012] Avoid resetting calendar when enabling a date field

cbliard:
# Ticket

https://community.openproject.org/wp/62012

When clicking on a date field in single date mode, it reinitializes the flatpickr instance. So if the user has navigated to a particular month, this is lost and reset to the default displayed month.

## Screenshots

For before, see the screen recording from the [bug report](https://community.openproject.org/wp/62012)

For after, here is how it loooks:

[calendar not reset.webm](https://github.com/user-attachments/assets/fbeb6b20-97d7-446d-921a-704a58900203)

This happens because it was using a turbo-stream to update the dialog. When doing so, the node where the flatpickr is attached is replaced.

We have some logic to avoid this replacement in the preview controller: it listens to turbo:before-frame-render events and avoids morphing for tags starting with "opce".

As the logic is already there, this commit uses a turbo-frame instead of a turbo-stream. This way the flatpickr instance is not replaced and the calendar view is kept.

The controller had to be updated to alter the work package attributes with the given params.

# Merge checklist

- [x] Added/updated tests

### DIFF
--- a/app/components/work_packages/date_picker/date_form_component.html.erb
+++ b/app/components/work_packages/date_picker/date_form_component.html.erb
@@ -7,7 +7,7 @@
             Primer::ButtonComponent.new(
               tag: :a,
               href: single_date_field_button_link("start_date"),
-              data: { turbo_stream: true },
+              data: { turbo_frame: "wp-datepicker-dialog--content", morph: true },
               classes: "wp-datepicker-dialog-date-form--add-button",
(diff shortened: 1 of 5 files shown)


HDinger (commented):
Behaviour-wise this looks fine, there are however some tests failing which might be related

PR #18278, opened 2025-03-12
Bug/59081 misleading caption shown on project list table header

ulferts:
# Ticket

https://community.openproject.org/wp/59081

Not have an html title attribute that makes no sense on the project list table headers as more options than just sorting exist there.

## Screenshots

Before:

image

After:
No tooltip on hover


as-op (approved):
👍

PR #18281, opened 2025-03-13
Translate color label in WorkPackages:Types::SettingsForm component

top4ek:
Translate color label in WorkPackages:Types::SettingsForm component.


HDinger (approved):
👍

PR #18286, opened 2025-03-13
Ensure to only store tokens for OIDC logins

NobodysNightmare:
The hook we are using is firing for all kinds of omniauth logins. By ensuring that we have a matching OpenIDConnect provider, we know that we are currently performing an OIDC login, which is the only case where we want to try storing OIDC tokens.

# Ticket
https://community.openproject.org/projects/cross-application-user-integration-stream/work_packages/62086

### DIFF
--- a/modules/openid_connect/lib/open_project/openid_connect/hooks/hook.rb
+++ b/modules/openid_connect/lib/open_project/openid_connect/hooks/hook.rb
@@ -1,3 +1,5 @@
+# frozen_string_literal: true
+
 #-- copyright
 # OpenProject is a project management system.
 # Copyright (C) the OpenProject GmbH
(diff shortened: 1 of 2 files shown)


NobodysNightmare:
@ba1ash [wrote](https://github.com/opf/openproject/pull/18255#pullrequestreview-2677893136):

> WDYT about adding a test?
> Not sure about the place for it... Maybe spec/services/authentication/omniauth_service_spec.rb or modules/openid_connect/spec/requests/openid_connect_spec.rb?

`openid_connect_spec.rb` is currently the only place that covers this hook with specs, so it seems like the right choice. However, that a very integrative spec. E.g. the login only works because a correct provider is present. This means for a clean test there, I'd need to create a different provider that could handle the login, but that would probably make an OIDC spec depend on a non-OIDC provider/module being present :see_no_evil: 

Maybe I can unit test this...

ba1ash (approved):
Looks nice! Thanks for taking care :heart:

PR #18293, opened 2025-03-13
[#62160] The Version custom field options are not grouped on project pages

dombesz:
# Ticket
https://community.openproject.org/work_packages/62160

Add version field project grouping on the Project General settings, Project Copy and Project Create pages.

Apply the same `groupByFn` function on the `select-input.component.ts`

# Merge checklist

- [X] Added/updated tests

### DIFF
--- a/frontend/src/app/shared/components/dynamic-forms/components/dynamic-inputs/select-input/select-input.component.ts
+++ b/frontend/src/app/shared/components/dynamic-forms/components/dynamic-inputs/select-input/select-input.component.ts
@@ -1,11 +1,30 @@
-import { Component, OnInit } from '@angular/core';
+import { ChangeDetectionStrategy, Component } from '@angular/core';
 import { FieldType } from '@ngx-formly/core';
-import idFromLink from 'core-app/features/hal/helpers/id-from-link';
+import { HalResource } from 'core-app/features/hal/resources/hal-resource';
+import { I18nService } from 'core-app/core/i18n/i18n.service';
 
 @Component({
   selector: 'op-select-input',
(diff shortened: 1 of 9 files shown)


frontend/src/app/shared/components/dynamic-forms/components/dynamic-inputs/select-input/select-input.component.ts
```diff
@@ -1,11 +1,30 @@
...
 })
 export class SelectInputComponent extends FieldType {
+  constructor(
+    readonly I18n:I18nService,
+  ) {
+    super();
+  }
+
+  groupByFn = (item:HalResource):string|null => {
+    if (!this.isVersionResource(item)) return null;
```
EinLama, line 20: This implementation is consistent with the existing other implementations of `groupByFn`, but if you had to implement this _from scratch,_ I would have suggested returning an empty string here. It also evaluates to false, but you could keep a clean method signature by strapping the `|null` from it, retaining meaning.

But as it is, this is consistent and well-working 👍

EinLama (approved):
Brilliant! 👏🏻

frontend/src/app/shared/components/dynamic-forms/components/dynamic-inputs/select-input/select-input.component.ts
dombesz: Unfortunately that is not a viable solution, because returning an empty string from this function, will group options under a grouping with an empty name. Returning null, will place the option as a root element, which is desirable.
<details><summary>This is how it would look like.</summary>
<p>

image


</p>
</details>

PR #18295, opened 2025-03-13
[#62047] fix creating folders from location picker

Kharonus:
# Ticket
[OP#62047](https://community.openproject.org/wp/62047)

- create folder should work
- created folder is shown when navigating breadcrumbs

- enable create folder from location picker shown in project storage mode selection
- also fix location picker state, when a folder was created, so that new folder shows, if user navigates breadcrumbs after creation

### DIFF
--- a/frontend/src/stimulus/controllers/dynamic/project-storage-form.controller.ts
+++ b/frontend/src/stimulus/controllers/dynamic/project-storage-form.controller.ts
@@ -98,6 +98,7 @@ export default class ProjectStorageFormController extends Controller {
   selectProjectFolder(_evt:Event):void {
     const locals = {
       projectFolderHref: this.projectFolderHref,
+      createFolderHref: `${this.storage._links.self.href}/folders`,
       storage: this.storage,
     };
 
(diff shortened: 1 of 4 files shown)


frontend/src/stimulus/controllers/dynamic/project-storage-form.controller.ts
```diff
@@ -98,6 +98,7 @@ export default class ProjectStorageFormController extends Controller {
   selectProjectFolder(_evt:Event):void {
     const locals = {
       projectFolderHref: this.projectFolderHref,
+      createFolderHref: `${this.storage._links.self.href}/folders`,
```
NobodysNightmare, line 101: What I understand: This local was added to the place where creating a folder worked, but it wasn't added here and therefore adding a folder from the picker didn't work here.

Correct?

NobodysNightmare (approved):
Seems to be working. Now that I looked very closely to the breadcrumb of the file picker I noticed an issue with moving text once navigating from the root to a subfolder. That was certainly not introduced in this PR, but depending on how keen you are to check some CSS, you might want to have a look:

moving-breadcrumb

frontend/src/stimulus/controllers/dynamic/project-storage-form.controller.ts
Kharonus: full points, correct deduction.

Kharonus:
probably some spacing was added in the links or removed from the "i'm not yet a link" element. will have a quick look

PR #18301, opened 2025-03-13
[#61967] Provide all the versions the user has access to on the Project and Global Work package list pages

dombesz:
# Ticket
https://community.openproject.org/work_packages/61967

Provide all the versions the user has access to:
- [x] On the Project list Version Custom field filter
- [x] On the Global WorkPackage list Version Custom field filter
- [x] On the Global WorkPackage list Version filter

# Merge checklist

- [x] Added/updated tests

### DIFF
--- a/lib/api/v3/queries/schemas/version_filter_dependency_representer.rb
+++ b/lib/api/v3/queries/schemas/version_filter_dependency_representer.rb
@@ -40,8 +40,6 @@ module API
             query_params = "sortBy=#{to_query [%i(semver_name asc)]}&pageSize=-1"
 
             if filter.project.nil?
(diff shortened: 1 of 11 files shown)


lib/api/v3/queries/schemas/version_filter_dependency_representer.rb
```diff
@@ -42,7 +42,7 @@ def href_callback
             if filter.project.nil?
               filter_params = [{ sharing: { operator: "=", values: ["system"] } }]
```
EinLama: This variable is no longer used and should be removed.

EinLama (approved):
Nice work, functions as desired 👍🏻 

I spotted a variable that is no longer needed. Apart from that, good to go! Approving so that you can merge as you see fit :)

PR #18316, opened 2025-03-14
Implements the BaseConfigurationValidator

mereghost:
#### ⚠️ Related WP: [OP#62098](https://community.openproject.org/work_packages/62098)

This PR introduces the first group of tests for Nextcloud Health Checks.

### DIFF
--- a/modules/storages/app/common/storages/peripherals/connection_validators/check_result.rb
+++ b/modules/storages/app/common/storages/peripherals/connection_validators/check_result.rb
@@ -31,25 +31,22 @@
 module Storages
   module Peripherals
     module ConnectionValidators
-      CheckResult = Struct.new("CheckResult", :key, :state, :message, :timestamp) do
-        def self.init_for(key)
+      CheckResult = Data.define(:key, :state, :message, :timestamp) do
+        private_class_method :new
+        def self.skipped(key)
           new(key:, state: :skipped, message: nil, timestamp: nil)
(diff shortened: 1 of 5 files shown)


modules/storages/app/common/storages/peripherals/connection_validators/check_result.rb
```diff
@@ -0,0 +1,53 @@
...
+        def self.failure(key, message)
+          new(key:, state: :failure, message: message, timestamp: Time.zone.now)
+        end
+
+        def self.success(key)
+          new(key:, state: :success, message: nil, timestamp: Time.zone.now)
+        end
+
+        def success? = state == :success
+        def failure? = state == :failure
```
Kharonus, line 49: ❓ is it intended, that the `:skipped` state is not exposed through any `?` method?

Kharonus (approved):
Looks very nice. As it is the first version of a validation group, I'd now be curious about how the business logic looks like to concatenate groups and skip others in action. :D But I think, that will be grant.

modules/storages/app/common/storages/peripherals/connection_validators/check_result.rb
mereghost: For now, yes. I'm adding methods on a needed to basis.

For the next group a bunch of patterns might emerge and a lot of things might change.

PR #18317, opened 2025-03-14
Fix week splitting in spec helper

NobodysNightmare:
Surprisingly next_occuring will forward the time to the next date where something occurs, but it will not change the time. So depending on when exactly a meeting is happening, the spec might have put it on the wrong "end" of the week-cutover.

This causes relatively stable CI failures today.


mrmir (approved):
Nothing broke :)

PR #18319, opened 2025-03-14
Feature/62150 add leading character to pattern input to initiate search

Kharonus:
# Ticket
[OP#62150](https://community.openproject.org/wp/62150)
[OP#62258](https://community.openproject.org/wp/62258)

- use `/` to indicate the search word and replacement range
- input changes within an existing token triggers token replacement

- amend pattern input component to behave accordingly to new specification

# Information
- this PR should give a first impression of the new interaction pattern, hence a pull preview is enabled
- please expect edge cases I might have missed
- Any implementation chunks inside the linked feature are supposedly NOT included yet.

### DIFF
--- a/frontend/src/stimulus/controllers/pattern-input.controller.ts
+++ b/frontend/src/stimulus/controllers/pattern-input.controller.ts
@@ -133,9 +133,9 @@ export default class PatternInputController extends Controller {
 
     this.tagInvalidTokens();
 
-    // this resets the cursor position without changing it
-    // it is necessary because chromium based browsers try to
-    // retain styling and adds a unwanted <font> tag,
+    // This resets the cursor position without changing it.
+    // It is necessary because chromium based browsers try to
+    // retain styling and adds an unwanted <font> tag,
     // breaking the behaviour of this component
     const selection = document.getSelection();
     if (selection && selection.rangeCount) {
@@ -338,8 +338,6 @@ export default class PatternInputController extends Controller {
     this.clearSuggestionsFilter();
     this.suggestionsTarget.classList.remove('d-none');
 
-    this.appendInsertAsTextElement(word);
-
     const filtered = this.getFilteredSuggestionsData(word.toLowerCase());
 
     // insert the HTML
@@ -369,6 +367,10 @@ export default class PatternInputController extends Controller {
         this.suggestionsTarget.appendChild(groupDivider);
       }
     });
+
+    if (this.suggestionsTarget.childNodes.length === 0) {
+      this.appendInsertAsTextElement(word);
+    }
   }
 
   private appendInsertAsTextElement(word:string):void {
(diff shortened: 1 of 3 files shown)


frontend/src/stimulus/controllers/pattern-input.controller.ts
```diff
@@ -166,13 +168,24 @@ export default class PatternInputController extends Controller {
...
+      this.contentTarget.appendChild(token);
       this.clearSuggestionsFilter();
+      return;
+    }
+
+    const parentNode = this.currentRange.startContainer.parentNode;
+    if (parentNode !== null && this.isToken(parentNode)) {
+      this.replaceToken(token, parentNode);
+    } else {
+      this.insertToken(token);
```
brunopagno: :green_circle: I like this behaviour better than the previous. Nice :clap:

frontend/src/stimulus/controllers/pattern-input.controller.ts
```diff
@@ -189,10 +202,7 @@ export default class PatternInputController extends Controller {
   private setRange():void {
     const selection = document.getSelection();
     if (selection?.rangeCount) {
-      const range = selection.getRangeAt(0);
-      if (range.startContainer.parentNode === this.contentTarget) {
-        this.currentRange = range;
-      }
+      this.currentRange = selection.getRangeAt(0);
```
brunopagno, line 217: The change here _could_ mean that we don't even need a `currentRange` local variable anymore. It was meant to keep track of a valid range (as in, a range that is not inside a token).

Worth checking if I'm not forgetting something else, though.

frontend/src/stimulus/controllers/pattern-input.controller.ts
```diff
@@ -263,44 +273,60 @@ export default class PatternInputController extends Controller {
...
 
-      const wordRange = document.createRange();
-      wordRange.setStart(targetNode, pos + 1);
-      wordRange.setEnd(targetNode, targetOffset);
+    if (textContent === null) { return; }
 
-      wordRange.deleteContents();
-      wordRange.insertNode(token);
+    let pos = targetOffset - 1;
+    while (pos > -1 && textContent.charAt(pos) !== COMPLETION_CHARACTER) { pos -= 1; }
```
brunopagno, line 305: :green_circle: Also much much better than before

brunopagno (commented):
The PR is still in draft, is that intended?

But the code changes look correct to me. There's a potential change with the `this.currentRange` if you want to investigate further.

I also tested locally and it looks good

frontend/src/stimulus/controllers/pattern-input.controller.ts
Kharonus: Hmm, I thought we need this variable to maintain the range, when I get out of the input with `ArrowDown` to select a completion element? Is my assumption wrong?

brunopagno: True that. I had forgotten about it :sweat_smile: 
Was just happy for a moment that we could remove some more core. but you're right.

psatyal (approved):
Did a visual check and I really like the new implementation. Great job, guys!

brunopagno (approved):
:ship:

PR #18320, opened 2025-03-14
[61990] On Gantt view, unidentified bar appears behind the date on hover, add margin right to the icon

bsatarnejad:
# Ticket
https://community.openproject.org/wp/61990

Add margin right to the scheduling icon, and align it with the text

## Screenshots
Before:
Screenshot 2025-03-14 at 16 34 42

After:
Screenshot 2025-03-14 at 16 11 52

### DIFF
--- a/frontend/src/global_styles/content/work_packages/timelines/elements/_labels.sass
+++ b/frontend/src/global_styles/content/work_packages/timelines/elements/_labels.sass
@@ -1,13 +1,19 @@
 @mixin timeline-label
   display: flex
   align-items: center
-  height: 16px
   min-width: 20px
   width: max-content
   font-size: 14px
   white-space: nowrap
   vertical-align: middle
   pointer-events: none
+
+@mixin scheduling-icon
+  span
+    display: flex
+    align-items: center
+    .display-field--scheduling-icon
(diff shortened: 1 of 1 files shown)


cbliard (changes requested):
For the background and margin, it now works as expected

Meanwhile, it seems this change has created a misalisgnment between left label and right label when they are _not_ hovered.

Before they are aligned:

image

After they're not anymore

image

Once that's fixed, it will be good.

Also unrelated but it seems that the left label on hover is better centered horizontally than the right label on hover: for the left one it looks like it's zoomed out, and for the right one it looks like it shifts and it is zoomed out.
- For the left one, it has `translateX(calc(-100% - 15px))`
- For the right one, it has `translateX(calc(100% + 10px))`

When harmonizing them by using `15px` in both cases, I find the transition to be smoother.

cbliard (approved):
Nice 👌

cbliard:
failing tests:
- ./spec/features/notifications/notification_center/notification_center_sidemenu_spec.rb:116
  - passing locally
- ./spec/features/work_packages/reminders_spec.rb:68
  - has been fixed in the meantime (cbe5c5d8957)

merging

PR #18327, opened 2025-03-17
[#62007] New scheduling mode  https://community.openproject.org/work_…

MayaBerd:
…packages/62007

[#62007] New scheduling mode

https://community.openproject.org/work_packages/62007


psatyal (approved):
Hi @MayaBerd. Just reviewed the docs and made some changes. 

I rewrote and restructured the Gantt chart docs because I thought we could further simply it. As for the main/big page about settings and changing docs, I mostly just made a few corrections and added some details. 

It now looks good to me but I think another review from you would surely help. Thanks!

Cheers
Parimal

PR #18329, opened 2025-03-17
[#62108] Meeting creation uses UTC offset, not timezone

mrmir:
# Ticket

https://community.openproject.org/work_packages/62108

- The timezone in the meeting/meeting series create and edit forms should change depending on the values of the date and start time fields


cbliard (changes requested):
This is my first review of meetings stuff, so maybe I missed things :)

Overall, I think it works well.

It took me some time to understand that the timezone being displayed is the one from the user settings and not the one from the browser/system. 

I chose Taipei (UTC+8) and the timezone name is CST. It's a bit imprecise because it can be either [Central Standard Time](https://www.timeanddate.com/time/zones/cst) (UTC-6) or [China Standard Time](https://www.timeanddate.com/time/zones/cst-china) (UTC+8).

I simply found an alignment issue that did not exist before and would need to be fixed.

Before:

image


After:

image

I did not look at code yet and will do it now.

modules/meeting/app/components/meetings/index/form_component.html.erb
```diff
@@ -8,7 +8,7 @@
         turbo: true,
         controller: [
           "show-when-value-selected",
-          @meeting.is_a?(RecurringMeeting) ? "recurring-meetings--form" : nil
+          @meeting.is_a?(RecurringMeeting) ? "recurring-meetings--form meetings--form" : "meetings--form"
```
cbliard: `"meetings-form"` could be added out of this condition, as the whole array is then compacted and joined.
```suggestion
          "meetings--form",
          @meeting.is_a?(RecurringMeeting) ? "recurring-meetings--form" : nil
```
It should even be put above so that the values in the array are alphabetically sorted, but the review suggestion is limited.

modules/meeting/app/components/meetings/index/form_component.html.erb
```diff
@@ -43,7 +43,31 @@
...
+              time_field.with_row(mt: 1) do
+                render(
+                  Primer::Beta::Text.new(
+                    id: "meeting-timezone",
+                    font_size: :small,
+                    color: :subtle
+                  )
+                ) { timezone_caption }
+              end
+            end
```
cbliard: Removing the `with_row` calls and `flex_layout: true`, and using `concat` to assemble the 2 `render`s together instead fixes the misalignement.

But I'm not very used to designing UI with our toolkit so there may be a better way.

modules/meeting/app/components/meetings/side_panel/details_form_component.html.erb
```diff
@@ -13,8 +17,32 @@
...
+                        font_size: :small,
+                        color: :subtle
+                      )
+                    ) { timezone_caption }
+                  end
+                end
+
+                time_row.with_column(flex: 1) do
+                  render(Meeting::Duration.new(f, meeting: @meeting))
+                end
```
cbliard: Looks like duplicated code.

can't it be extracted?

This `Meeting::TimeGroup` that was used before looked like a good extraction. Maybe this can be reused.
By the way, `Meeting::TimeGroup` is not used anywhere anymore, so it should have been deleted in the course of this PR, so the name is free to be picked up.

modules/meeting/app/components/meetings/index/form_component.html.erb
cbliard: Using `mt: 0` also fixes the misalignment issue.

modules/meeting/app/forms/meeting/start_date.rb
```diff
@@ -0,0 +1,53 @@
...
+    meeting_form.text_field(
+      name: :start_date,
+      type: "date",
+      value: @initial_date,
+      placeholder: @meeting.class.human_attribute_name(:start_date),
+      label: @meeting.class.human_attribute_name(:start_date),
+      required: true,
+      autofocus: false,
+      data: {
+        action: "input->recurring-meetings--form#updateFrequencyText input->meetings--form#updateTimezoneText"
```
cbliard: I like putting these on multiple lines, I find it easier to scan. But that's a preference. It's perfectly ok as is.

```suggestion
        action: "input->recurring-meetings--form#updateFrequencyText" \
                "input->meetings--form#updateTimezoneText"
```

modules/meeting/app/forms/meeting/start_date.rb
```diff
@@ -0,0 +1,53 @@
...
+      }
+    )
+  end
+
+  def initialize(meeting:)
+    super()
+
+    @meeting = meeting
+    @initial_date = meeting.start_date.presence || format_time_as_date(meeting.start_time, format: "%Y-%m-%d")
```
cbliard: I was curious to see what `format_time_as_date` would return so I added some debug logs, and got `undefined method 'format_time_as_date'`.

So I have the hunch that `meeting.start_date` is never `nil` and the part after `||` is dead code. If not, then there is a bug lying but I didn't manage to trigger it. You can probably judge better than me.

If it's really dead code, it should be removed.

cbliard (changes requested):
I have some code remarks, and I found an additional issue:

In create meeting modal, when deleting part of the date or time, the server is unable to calculate the timezone and it returns a 500 error (`fetch_timezone` action, "NoMethodError (undefined method 'utc' for nil)").

Same happens when editing date and/or time in meeting details.

Here is a reproduction video

[500 error with incomplete dates.webm](https://github.com/user-attachments/assets/8dca7331-5078-4017-8e9d-e45a1cab7790)

mrmir:
Thanks for the detailed review - not bad for your first time :D

I've made some changes to reuse the `TimeGroup` form component instead of splitting it up as I found a way to update the caption directly, instead of needing to treat it as a text field (which was the reason for the split in the first place). This also removes the unnecessary duplication you caught, as well as the alignment issue

I also incorporated your other feedback to improve on the code and fix the 500 error

modules/meeting/app/forms/meeting/time_group.rb
```diff
@@ -41,7 +41,8 @@ class Meeting::TimeGroup < ApplicationForm
         required: true,
         autofocus: false,
         data: {
-          action: "input->recurring-meetings--form#updateFrequencyText"
+          action: "input->recurring-meetings--form#updateFrequencyText \
+                   input->meetings--form#updateTimezoneText"
```
cbliard, line 45: Almost. The string you created is `"input->recurring-meetings--form#updateFrequencyText                    input->meetings--form#updateTimezoneText"`. It would be better to have `"input->recurring-meetings--form#updateFrequencyText input->meetings--form#updateTimezoneText"` without all the spaces

Ruby allows to concatenate string without any `+` between the strings:
```
irb
>> "hello"   "world"
=> "helloworld"
```

And `\` escape the following character, meaning that if it's used at the end of a line, it will ignore the newline character and consider that the 2 lines are on the same line.

So the above can be written like this:
```suggestion
          action: "input->recurring-meetings--form#updateFrequencyText " \
                   "input->meetings--form#updateTimezoneText"
```
This way it's only one string. This syntax is a bit strange but valid and recognized by rubocop like in [Layout/LineContinuationSpacing cop](https://docs.rubocop.org/rubocop/cops_layout.html#layoutlinecontinuationspacing) for instance.

modules/meeting/app/forms/meeting/time_group.rb
```diff
@@ -52,11 +53,13 @@ class Meeting::TimeGroup < ApplicationForm
         placeholder: Meeting.human_attribute_name(:start_time),
         label: Meeting.human_attribute_name(:start_time),
         required: true,
-        caption: formatted_time_zone_offset,
+        caption: timezone_caption,
         data: {
-          action: "input->recurring-meetings--form#updateFrequencyText"
+          action: "input->recurring-meetings--form#updateFrequencyText \
+                   input->meetings--form#updateTimezoneText"
```
cbliard, line 59: same

modules/meeting/app/forms/meeting/time_group.rb
```diff
@@ -81,8 +81,8 @@ def initialize(meeting:)
     super()
 
     @meeting = meeting
-    @initial_time = meeting.start_time_hour.presence || format_time(meeting.start_time, include_date: false, format: "%H:%M")
-    @initial_date = meeting.start_date.presence || format_time_as_date(meeting.start_time, format: "%Y-%m-%d")
+    @initial_time = meeting.start_time_hour.presence
```
mrmir, line 84: Just noticed the `.presence` that isn't needed anymore, will remove

modules/meeting/app/controllers/meetings_controller.rb
```diff
@@ -329,6 +329,23 @@ def notify
...
+    User.execute_as(User.current) do
+      meeting = Meeting.new(timezone_params)
+      @text = friendly_timezone_name(User.current.time_zone, period: meeting.start_time)
+    end
+
+    prefix = params[:structured_meeting] ? "structured_" : ""
+
+    add_caption_to_input_element_via_turbo_stream("input[name='#{prefix}meeting[start_time_hour]']",
+                                                  caption: @text,
+                                                  clean_other_captions: true)
```
cbliard, line 344: `@text` is not needed as it's only used in this method. You could use a local variable.

I suppose you had some issues with variable scope.
You can circumvent it this way :
```ruby
text = nil
User.execute_as(User.current) do
  text = ...
end
```

But I'm not sure I understand why `User.execute_as` is needed. In `User.execute_as` body, the save and restore of `User.current` is not needed because we're passing `User.current` as user, so only the `OpenProject::LocaleHelper.with_locale_for(user, &)` is useful.

Inside  `OpenProject::LocaleHelper.with_locale_for`, it saves the current locale and the current time zone, and sets the current time zone using the user's one. Then it also sets the language and executes the block and restores the values. I guess that's the part about setting `Time.zone` which is useful. But then it's using `friendly_timezone_name(User.current.time_zone, period: meeting.start_time)` where the time zone is explicitly passed to the function. In it everything is explicit so I'm not sure `User.execute_as` is actually needed.

But in doubt, I wouldn't touch it.

To fix the scope issue with `@text`, we can also use the fact that `User.execute_as` returns the last evaluated expression:

```ruby
text = User.execute_as(User.current) do
  meeting = Meeting.new(timezone_params)
  friendly_timezone_name(User.current.time_zone, period: meeting.start_time)
end
```

This could even be extracted into a `friendly_timezone` method.

cbliard (approved):
Great job 👍 

I could not find any issues. I only have minor code remarks so I'll merge the PR.

modules/meeting/app/forms/meeting/time_group.rb
mrmir: Aah, I missed the trailing space

I initially added it in the way you suggested, with two strings in quotes, but that broke the functionality. I assumed I got the syntax wrong and so made it one string, but now with the trailing space it works fine too, so I guess I missed that initially. Thanks for the info anyway! I hope I can remember to do it like this in the future instead of all in one line

PR #18330, opened 2025-03-17
Fix timezone parsing for reminder form and tests

klaustopher:
Today + 2 weeks is past the DST cutoff for central europe timezone. So the tests uncovered that we parsed the time wrongly. This PR fixes the parsing for the timestamp and the tests.

### DIFF
--- a/app/controllers/work_packages/reminders_controller.rb
+++ b/app/controllers/work_packages/reminders_controller.rb
@@ -184,7 +184,7 @@ class WorkPackages::RemindersController < ApplicationController
 
   def build_remind_at_from_params(remind_at_date, remind_at_time)
     if remind_at_date.present? && remind_at_time.present?
-      DateTime.parse("#{remind_at_date} #{User.current.time_zone.parse(remind_at_time)}")
+      User.current.time_zone.parse("#{remind_at_date} #{remind_at_time}")
     end
   end
 
--- a/spec/features/work_packages/reminders_spec.rb
+++ b/spec/features/work_packages/reminders_spec.rb
@@ -42,12 +42,9 @@ RSpec.describe "Work package reminder modal",
   end
 
   let!(:user_with_permissions) do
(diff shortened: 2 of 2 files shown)


spec/features/work_packages/reminders_spec.rb
```diff
@@ -79,6 +79,7 @@
           .to have_css(".spot-modal--subheader",
                        text: "You will receive a notification for this work package at the chosen time.")
         fill_in "Date", with: date
+        # capybara expets us to set a full timestamp here, eventhough the input is only for time and the date portion is irre
```
akabiru:
```suggestion
        # capybara expets us to set a full timestamp here, eventhough the input is only for time and the date portion is irrelevant 
```

app/controllers/work_packages/reminders_controller.rb
```diff
@@ -184,7 +184,7 @@ def reminder_params
 
   def build_remind_at_from_params(remind_at_date, remind_at_time)
     if remind_at_date.present? && remind_at_time.present?
-      DateTime.parse("#{remind_at_date} #{User.current.time_zone.parse(remind_at_time)}")
+      User.current.time_zone.parse("#{remind_at_date} #{remind_at_time}")
```
akabiru, line 187: Nice 👍🏾

akabiru (approved):
🥇

spec/features/work_packages/reminders_spec.rb
```diff
@@ -79,7 +76,9 @@
           .to have_css(".spot-modal--subheader",
                        text: "You will receive a notification for this work package at the chosen time.")
         fill_in "Date", with: date
-        fill_in "Time", with: time
+        # capybara expets us to set a full timestamp here, eventhough the
```
NobodysNightmare, line 79: Post review nit:

```suggestion
        # capybara expects us to set a full timestamp here, even though the
```

PR #18332, opened 2025-03-17
Remove specific work package permissions for comments with restricted visibility

brunopagno:
# Ticket

https://community.openproject.org/wp/60978

Adjust comments with restricted visibility permissions to project level only, not on individual work packages.

# Merge checklist

- [x] Added/updated tests

### DIFF
--- a/app/models/work_package.rb
+++ b/app/models/work_package.rb
@@ -382,17 +382,12 @@ class WorkPackage < ApplicationRecord
   # check if user is allowed to edit WorkPackage Journals.
   # see Acts::Journalized::Permissions#journal_editable_by
   def journal_editable_by?(journal, user)
-    can_edit_journal_note = ->(edit_others_notes_permission, edit_own_notes_permission) {
-      user.allowed_in_project?(edit_others_notes_permission, project) ||
-        (user.allowed_in_work_package?(edit_own_notes_permission, self) && journal.user_id == user.id)
-    }
-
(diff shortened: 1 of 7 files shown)


app/models/work_package.rb
```diff
@@ -382,17 +382,12 @@ def update_done_ratio_from_status
...
     if journal.restricted?
-      can_edit_journal_note.call(:edit_others_comments_with_restricted_visibility,
-                                 :edit_own_comments_with_restricted_visibility)
+      user.allowed_in_project?(:edit_others_comments_with_restricted_visibility, project) ||
+        (user.allowed_in_project?(:edit_own_comments_with_restricted_visibility, project) && journal.user_id == user.id)
     else
-      can_edit_journal_note.call(:edit_work_package_notes,
-                                 :edit_own_work_package_notes)
+      user.allowed_in_project?(:edit_work_package_notes, project) ||
+        (user.allowed_in_work_package?(:edit_own_work_package_notes, self) && journal.user_id == user.id)
```
akabiru, line 390: I'm slightly puzzled why the "old" permissions are applicable on the work package level whilst we're defining new ones on the project level. @ulferts talked about this in https://github.com/opf/openproject/pull/18091#discussion_r1993067307 but I'm wondering if we also need to amend the old permission "edit_own_work_package_notes" to be project level

akabiru (approved):

PR #18333, opened 2025-03-17
[61979] Switch successor to automatic mode even without rescheduling

cbliard:
# Ticket

https://community.openproject.org/wp/61979

Switch successor to automatic scheduling mode when adding a predecessor/successor link to a work package.
Sometimes it works, sometimes it does not. Actually it was working only when adding the predecessor would change the successor's dates. Without that condition, it would not switch (which is the bug).

## Screenshots

Switching to automatic mode was done in `ScheduleDependency` before. As the scheduling mode was changed outside of the `SetScheduleService`, it was not seen as a change and no dependent result was saved and eventually the scheduling mode change was not persisted.

By moving the scheduling mode change to the `SetScheduleService`, the dependent results are now correctly saved and the scheduling mode change is persisted.

# Merge checklist

- [x] Added/updated tests

### DIFF
--- a/app/services/work_packages/schedule_dependency.rb
+++ b/app/services/work_packages/schedule_dependency.rb
@@ -192,8 +192,6 @@ class WorkPackages::ScheduleDependency
 
     # rehydrate the predecessors and followers of follows relations
     rehydrate_follows_relations
-
-    fix_switching_to_automatic_mode_work_packages
   end
 
   # Returns all the descendants of moved and moving work packages that are not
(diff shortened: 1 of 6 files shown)


klaustopher (approved):

cbliard:
Two failing specs:
- './spec/features/work_packages/details/date_editor_spec.rb:479'
  - passing locally
- './spec/features/work_packages/reminders_spec.rb:68'
  - Already fixed in cbe5c5d

Merging

PR #18334, opened 2025-03-17
[61780] Regression: Rich text headlines show link icon on hover before the headline, while rendering after would be correct

bsatarnejad:
# Ticket
https://community.openproject.org/wp/61780

Shoe icon after after headline

## Screenshots
Before:
Screenshot 2025-03-17 at 12 19 28

After:
Screenshot 2025-03-17 at 12 19 02


dombesz (approved):
Works as expected.👍

Just a side note, the url for the link does not seem to be correct, perhaps we should log a bug for it, unless there is already one.

image

PR #18337, opened 2025-03-17
[#61959] Segmented control jumps when changing the date field

dombesz:
# Ticket
https://community.openproject.org/work_packages/61959

Prevent the segmented control jumping.

## Screenshots
<details><summary>Before (Please note the jumping on the "Automatic" segment control.)</summary>
<p>

before

</p>
</details> 

<details><summary>After</summary>
<p>

after

</p>
</details> 

~Enabling morphing on the datepicker seems to solve the flickering issue. Additionally, it solves another bug: When opening the datpicker and switching to manual, not the correct field is activated. It always activates the end date.~

There is an invisible text in bold over the segmented control buttons. This is to prevent jumping when the button gets clicked, because the active button is in bold and the width would shift without it.

Problem is: this invisble text is set from a `data-content` property which is set on the `onConnected` callback. This callback is not called when the content is morphed. When it's morphed, the `data-content` is removed, the invisible text is removed and the button text "jumps".

The fix is to hook into the morphing and call that callback manually for the segmented control so that the invisible text is there and there is no jumps.

# Merge checklist

- [ ] Added/updated tests

### DIFF
--- a/app/components/work_packages/date_picker/form_component.html.erb
+++ b/app/components/work_packages/date_picker/form_component.html.erb
@@ -23,7 +23,8 @@
                     icon: :pin,
                     href: work_package_datepicker_dialog_content_path(params.merge(schedule_manually: true).permit!),
                     data: {
-                      turbo_stream: true,
+                      turbo_frame: "wp-datepicker-dialog--content",
+                      morph: true,
                       qa_selected: schedule_manually
                     },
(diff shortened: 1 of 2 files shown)


cbliard (approved):
Nice 👏 

That was an interesting issue.

PR #18339, opened 2025-03-17
[#60122] Required custom fields prevent a WorkPackage from being added as a child

dombesz:
# Ticket
https://community.openproject.org/work_packages/60122

Allow saving work package relations when there is a custom field related validation error on the work package.

Apply context based validations while saving work packages, so custom fields won't get validated when we save work package related content, like relations. The custom field validation should only kick in when work package attributes are saved.

# Merge checklist

- [x] Added/updated tests


cbliard (approved):
Brilliant 👏

PR #18340, opened 2025-03-17
[62245] Fix date picker width to 600px

cbliard:
# Ticket

https://community.openproject.org/wp/62245

Text in the header banner can make the date picker too wide. The width is now fixed to 600px to avoid it and let the banner text wrap.

## Screenshots

Before

image

After

image

Before the limit was already 600px, but it was only applied to the body. Now it's applied to the whole content (header, body and footer).

# Merge checklist

- [ ] Added/updated tests

### DIFF
--- a/app/components/work_packages/date_picker/dialog_content_component.sass
+++ b/app/components/work_packages/date_picker/dialog_content_component.sass
@@ -7,12 +7,12 @@ $body-height: 460px
   .wp-datepicker-dialog
     &--content
       max-height: calc(var(--app-height) - 2rem)
+      width: 600px
       overflow: auto
 
     &--body
       // We set a fixed height for this dialog zo avoid that it jumps around when the tabs are switched or errors shown
       min-height: $body-height
-      width: 600px
 
     &--content-tab
       &--relation-tab


mrmir (approved):

app/components/work_packages/date_picker/dialog_content_component.sass
```diff
@@ -7,12 +7,12 @@ $body-height: 460px
   .wp-datepicker-dialog
     &--content
       max-height: calc(var(--app-height) - 2rem)
+      width: 600px
       overflow: auto
 
     &--body
       // We set a fixed height for this dialog zo avoid that it jumps around when the tabs are switched or errors shown
```
mrmir, line 14: Unrelated to this bug, but the height still jumps around when switching tabs, unlike what the comment says

cbliard: Thanks for the review.

Yeah. In the date picker, some months have 6 lines, and some have 5 lines. February can even have only 4 lines, like in 2026.
When there are 5 lines for both months, it does not jump. When one month has 6 lines, it jumps. 
I'll bring this to the UI/UX open door meeting.

I'll ignore it for this PR, then depending on the meeting outcome a bug may be created to fix it in a later release.

PR #18341, opened 2025-03-17
Translate units in time related filters

top4ek:
Use locale to render units in project filters of time type.

## Screenshots
Screenshot_20250317_184852

Use default rails locale definition of day unit.


ulferts (approved):
🙇 @top4ek

PR #18343, opened 2025-03-17
enable rails 6.0 settings

ulferts:
# Ticket

Bump default config used to rails 6.0 standards.


modules/bim/app/views/bim/ifc_models/ifc_models/_form.html.erb
```diff
@@ -114,22 +114,19 @@ See COPYRIGHT and LICENSE files for more details.
       setDirectUploadFileName();
 
       <%# rebuild form to post to S3 directly %>
-      if (jQuery("input[name=utf8]").length == 1) {
-        jQuery("input[name=utf8]").remove();
```
ulferts, line 118: @machisuji you wouldn't by any chance remember why you added this conditional 4 years ago? With the changed defaults, the utf8 hidden field is no longer part of the form. But I wonder about when it might ever haven't been part of the form and thus, why the behaviour should be different depending on the existence of the field.

... unless, you used it to achieve idempotence. So that the form is only changed once even if the user decides for a different file after all.

machisuji: @ulferts I don't quite remember. But it definitely had something to do with not changing the form. I think perhaps when a user had chosen a file before, then used the file picker again but cancelled. So that the last picked file remains instead of being reset?

ulferts: Thanks @machisuji, I readded a similar guard based on a different field that serves the same purpose. Cargo culting admittedly since I don't want to bother myself too much with the specifics.

EinLama (approved):
👍🏻 looks good, including the cargo cult 😀

PR #18348, opened 2025-03-18
Ensure SSO users do not fallback to ensure_connection

NobodysNightmare:
ensure_connection is intended for OAuth2 authentication at storage, but does not work, if we don't have an OAuth client, like for SSO authentication.

In the SSO case we can simply try to redirect the user to the correct URL, because they will most likely be able to login there and if in doubt, the storage can still refuse their request to visit the desired path.
Logging in to the storage, can in some cases lead to the user account being created on the storage, which should fix the main case of failed authentication in the first place. So hopefully this fallback helps ensuring users are eventually provisioned on the storage.

# Ticket
https://community.openproject.org/wp/62166

# Notes

I was surprised by the complexity involved in opening a storage. The happy path seems to go through:

* `ProjectStoragesController`
* `ProjectStorageOpenAPI`
* `ProjectStorage#open`

All of those have their own differentiation between happy path and sad path and in two of them we are performing authentication.

This PR just hooks into the existing decision tree, but we should probably clean that up/trim it down.

### DIFF
--- a/modules/storages/app/controllers/storages/project_storages_controller.rb
+++ b/modules/storages/app/controllers/storages/project_storages_controller.rb
@@ -98,18 +98,34 @@ class Storages::ProjectStoragesController < ApplicationController
         format.html do
           case result.code
           when :unauthorized
-            redirect_to(
-              oauth_clients_ensure_connection_url(
-                oauth_client_id: @storage.oauth_client.client_id,
-                storage_id: @storage.id,
-                destination_url: request.url
-              )
-            )
+            redirect_to(storage_fallback_url, allow_other_host: true)
           when :forbidden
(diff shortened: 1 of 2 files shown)


ba1ash (approved):
It makes sense to me. I agree that we should clean this code area up at some point.

ba1ash:
@NobodysNightmare I suppose `modules/storages/spec/requests/project_storages_open_spec.rb` could be extended to cover this case. Have you considered that?

NobodysNightmare:
@ba1ash good point, I added a spec there.

ba1ash:
> @ba1ash good point, I added a spec there.

Thanks!

PR #18039, opened 2025-02-21
Add magic comment for frozen string literals everywhere

NobodysNightmare:
This change was effectively achieved through

    rubocop -A --only Style/FrozenStringLiteralComment,Layout/EmptyLineAfterMagicComment,Style/RedundantFreeze

With this we ensure that all string literals allocated directly in OpenProject code are frozen, which allows the Ruby interpreter to deal with them more efficiently in memory (if duplicates occur) and will also ensure that our code is already compatible with an expectable future change to the Ruby language, where frozen string literals will be the default.

As a side effect we fix one of the major contributors to rubocop offenses in our codebase (by raw numbers).

PR #18349, opened 2025-03-18
Make Git SCM interactions frozen string compatible

NobodysNightmare:
Replaced `<<` with `+=`, so that strings are not changed in-place anymore.

This should fix more than half of the remaining spec failures in #18039.


NobodysNightmare:
I am ignoring existing Rubocop offenses here.

ba1ash (approved):

PR #18351, opened 2025-03-18
Feature/36743 filter work packages by an ancestor

EinLama:
# Ticket
https://community.openproject.org/wp/36743

Provide an ancestor filter that allows users to filter for any descendant of a given work package.

## Screenshots

https://github.com/user-attachments/assets/a77d0b38-0b24-4dc8-9bae-c68fb531075e

Copied the existing parent filter as a base and adjusted the query. Thanks to the excellent work package hierarchy table, I could use an existing method to filter for all descendants.

I decided to make the IS (OR) filter _inclusive_, which means that the selected ancestor _**will** show up_ in the result.
At the same time, the IS NOT filter will _exclude_ the ancestor in the result list, meaning it will _**not** show up._

This solution felt the most intuitive to me.

The selected translation is still a bit open to debate. Initially, I named the filter (label) "Ancestor", but later renamed it to "Descendants of", as stated in the work package description, because it seemed easier to understand for a non-technical person. Both are not ideal and I am open for suggestions.

# Merge checklist

- [x] Added/updated tests

### DIFF
--- a/app/models/queries/work_packages/filter/ancestor_filter.rb
+++ b/app/models/queries/work_packages/filter/ancestor_filter.rb
@@ -33,7 +33,11 @@ class Queries::WorkPackages::Filter::AncestorFilter <
   include ::Queries::WorkPackages::Filter::FilterForWpMixin
 
   def where
-    descendant_ids = WorkPackageHierarchy.where(ancestor_id: no_templated_values).pluck(:descendant_id)
+    descendant_ids = WorkPackageHierarchy
+                       .where(ancestor_id: no_templated_values)
+                       # exclude the selected ancestors:
+                       .where.not(descendant_id: no_templated_values)
+                       .pluck(:descendant_id)
 
     operator_strategy.sql_for_field(descendant_ids, self.class.model.table_name, :id)
   end
--- a/app/models/queries/work_packages/filter/ancestor_filter.rb
+++ b/app/models/queries/work_packages/filter/ancestor_filter.rb
@@ -39,9 +39,28 @@ class Queries::WorkPackages::Filter::AncestorFilter <
     Relation::TYPE_PARENT
   end
 
+  def apply_to(_query_scope)
+    cte = <<~SQL.squish
+      WITH RECURSIVE descendants AS (
+        SELECT id
+        FROM work_packages
+        WHERE id IN (:ids)
+        UNION ALL
+        SELECT wp.id
+        FROM work_packages wp
+        INNER JOIN descendants d ON wp.parent_id = d.id
+      )
+      SELECT * FROM descendants;
+    SQL
+
+    sql = ActiveRecord::Base.sanitize_sql([cte, { ids: no_templated_values }])
+
+    descendants = super.find_by_sql(sql)
+
+    WorkPackage.where(id: descendants.pluck(:id))
+  end
+
   def where
-    # The filter had been called parent before and it is stored in the database like that.
-    # The other association filters all have _id in their self.key.
-    operator_strategy.sql_for_field(no_templated_values, self.class.model.table_name, :parent_id)
+    "1 = 1"
   end
 end
--- a/app/models/queries/work_packages/filter/ancestor_filter.rb
+++ b/app/models/queries/work_packages/filter/ancestor_filter.rb
@@ -32,41 +32,10 @@ class Queries::WorkPackages::Filter::AncestorFilter <
   Queries::WorkPackages::Filter::WorkPackageFilter
   include ::Queries::WorkPackages::Filter::FilterForWpMixin
 
-  def apply_to(_query_scope)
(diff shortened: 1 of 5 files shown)


app/models/queries/work_packages/filter/ancestor_filter.rb
```diff
@@ -0,0 +1,41 @@
...
+# See COPYRIGHT and LICENSE files for more details.
+#++
+
+class Queries::WorkPackages::Filter::AncestorFilter <
+  Queries::WorkPackages::Filter::WorkPackageFilter
+  include ::Queries::WorkPackages::Filter::FilterForWpMixin
+
+  def where
+    descendant_ids = WorkPackage.where(id: no_templated_values).flat_map(&:descendants).pluck(:id)
+    descendant_ids.concat(no_templated_values)
```
ulferts: An easier and more performant statement would be:

```
descendant_ids = WorkPackageHierarchy.where(ancestor_id: no_templated_values).pluck(:descendant_id)
```

Since the entries in WorkPackageHierarchy also include a self reference (where ancestor_id and descendant_id are the same), the values of the filter themselves don't have to be added anew.  

I wonder though - would a user normally expect to also have the filtered for value returned as the result or should only work packages that are truly descendants be returned.

app/models/queries/work_packages/filter/ancestor_filter.rb
```diff
@@ -0,0 +1,41 @@
...
+
+    operator_strategy.sql_for_field(descendant_ids, self.class.model.table_name, :id)
```
ulferts, line 42: Even though this requires to first fetch the `descendant_ids` I think this approach is better than to write custom statements for every operator as would be the alternative. If ever we identify this as being too slow, we can still improve it.

spec/models/queries/work_packages/filter/ancestor_filter_spec.rb
```diff
@@ -0,0 +1,312 @@
...
+          expect(instance).not_to be_valid
+        end
+      end
+    end
+
+    describe "#where and #includes" do
+      let(:parent) { create(:work_package) }
+      let(:child) { create(:work_package, parent:) }
+      let(:grandchild) { create(:work_package, parent: child) }
+      let(:another_wp) { create(:work_package) }
```
ulferts, line 275: I think it would be good to also include a grandparent here which then also shouldn't be returned on `=`.

ulferts (changes requested):
Looks good @EinLama. I have smaller remarks which should be easy to address. 

The thing I am wondering about is whether the filter should also include the work packages that directly match the filtered for ids. In most queries I know there is a distinction between "descendants" and "self and descendants". I think I would exclude the filtered for work packages. What do you think?

app/models/queries/work_packages/filter/ancestor_filter.rb
EinLama: The PR currently implements it so that the selected ancestor itself is included. I did this because it is easy enough for users to _remove_ the ancestor by adding an ID filter for "IS NOT <id of the ancestor>", see image:
image

Note how the ancestor is still rendered (but greyed out), since we display the hierarchy in this view. That won't happen on a "flat" work package list.

In contrast, if we take the other route and **exclude** the selected ancestor in the result, a user that _wants_ to see the ancestor has a problem: since our filters are chained via `and`, there **is no way** (that I can think of) to show the ancestor.

I have attached a screenshot where I changed the filter to **exclude** the selected ancestor:
image

Comparing both approaches, I thought that including the selected ancestor is the more convenient and usable option, even if it's logically speaking not correct and might surprise some users 🤔 Then again, such a surprise could be nasty, especially in larger projects.

spec/models/queries/work_packages/filter/ancestor_filter_spec.rb
EinLama: Good idea, I added a grandparent to these specs 👍🏻

app/models/queries/work_packages/filter/ancestor_filter.rb
EinLama: After giving this another round of thought, it would be better to exclude the selected ancestor for now. If users request it, we can add another filter for "self and descendants" later.

EinLama: I have pushed a new commit in which selected ancestor is now excluded. Please have a look.

I am torn about the negative version of the filter (i.e. IS NOT a descendant of the selected WPs). Excluding the selected ancestor means that the selected ancestor will actually show up. This feels plain wrong, but is logically sound.

To me, the initial version of _including_ the selected ancestor feels more natural, especially when negating the filter 🤔

ulferts (approved):
Looks good 👍

PR #18353, opened 2025-03-18
[62261] Fix error displayed when switching parent to automatic

cbliard:
# Ticket

https://community.openproject.org/wp/62261

When switching a parent to automatic scheduling mode, the date picker displayed an error that the dates are read-only when the selected dates in manual mode are the ones that would be computed by automatic mode.

This error should not be displayed.

## Screenshots

Please see screen recordings in ticket.

This was due to system changes being the same as user changes, and as they were done after, it was not detected.

Fix is to track the attributes directly (maybe using private activemodel api) and check if the instance changed, in which case it can be considered a system change.

# Merge checklist

- [x] Added/updated tests

### DIFF
--- a/lib/open_project/changed_by_system.rb
+++ b/lib/open_project/changed_by_system.rb
@@ -83,12 +83,19 @@ module OpenProject
     # by the mass assignment the change in value will give that away.
     def change_by_system
       prior_changes = non_no_op_changes
+      prior_attributes_from_user = attributes_from_user(prior_changes.keys)
 
       ret = yield
 
-      changed_by_system(changes_compared_to(prior_changes))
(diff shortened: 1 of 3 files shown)


klaustopher (dismissed):

cbliard:
The code was faulty. I wrote `attributes` instead of `@attributes` in the condition.

So I added a unit test to cover it, fixed the issue, simplified the condition, amended my commit, and force-pushed.

It _should_ be better now.

Anyway now it will not make it for 15.4.0. It will be for 15.4.1.

cbliard:
I tried another implementation in #18392 which does not involve touching the `changed_by_system` mechanisms.

cbliard:
The fix in #18392 works and has been merged. Closing this one.

PR #18355, opened 2025-03-18
Update Ruby copyright headers using rake task

NobodysNightmare:
This effectively removes the year from a bunch of copyright headers. We stopped indicating a specific year at some point, but didn't propagate this change to every file yet.

This PR was effectively achieved through:

    rake copyright:update_rb


toy (approved):

NobodysNightmare:
Merging without waiting for specs, since we only touched comments.

PR #18356, opened 2025-03-18
Enable copyright task to add missing headers

NobodysNightmare:
The copyright task was so far only built to update existing (and properly marked) copyright headers.

We are now adding the capability to add a missing header for Ruby files. This allows us to properly indicate copyright where it was previously forgotten to be done.

# Dependencies

This PR should only be merged, after we normalized existing Ruby readers in #18355, as to keep the resulting diffs of future PRs smaller.

### DIFF
--- a/lib/tasks/copyright.rake
+++ b/lib/tasks/copyright.rake
@@ -1,3 +1,5 @@
+# frozen_string_literal: true
+
 #-- copyright
 # OpenProject is an open source project management software.
 # Copyright (C) the OpenProject GmbH
@@ -101,6 +103,15 @@ namespace :copyright do
     end
   end
 
+  def missing_copyright_regexp(format)
+    case format
+    when :ruby, :rb
+      /\A(?<shebang>#![^\n]+\n\n?)?(?<additional># frozen_string_literal: (?:true|false)\n\n?)?\n*/m
+    else
+      raise "Format #{format} is not yet supported for copyright creation"
+    end
+  end
+
   def rewrite_copyright(ending, additional_excluded_globs, format, path, options = {}) # rubocop:disable Metrics/AbcSize, Metrics/PerceivedComplexity
     regexp = options[:regex] || copyright_regexp(format)
     path = "." if path.nil?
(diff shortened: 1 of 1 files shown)


lib/tasks/copyright.rake
```diff
@@ -119,6 +130,12 @@ namespace :copyright do
       file_content = File.read(file_name)
       if file_content.match(regexp)
         file_content.gsub!(regexp, "\\k<shebang>\\k<additional>#{copyright}")
+      elsif options[:create]
+        if file_content.include?("OpenProject is a fork of ChiliProject")
+          next puts "#{file_name} does not match regexp, but seems to have a copyright header!"
```
toy: Rubocop is ok with this? `next` accepts an argument, but it is not needed here, so I suggest splitting into two lines

lib/tasks/copyright.rake
```diff
@@ -101,6 +103,15 @@ namespace :copyright do
     end
   end
 
+  def missing_copyright_regexp(format)
+    case format
+    when :ruby, :rb
+      /\A(?<shebang>#![^\n]+\n\n?)?(?<additional># frozen_string_literal: (?:true|false)\n\n?)?\n*/m
+    else
+      raise "Format #{format} is not yet supported for copyright creation"
```
toy, line 111: I suggest not failing, maybe returning `nil` and warning either here or where it is used, to be able to run it on whole codebase

lib/tasks/copyright.rake
```diff
@@ -143,10 +160,11 @@ namespace :copyright do
   desc "Update the copyright on .rb source files"
   task :update_rb, :path do |_task, args|
     excluded = %w[
-      lib_static/plugins/{acts_as_tree,rfpdf,verification}
+      lib_static/plugins/{acts_as_tree,rfpdf,verification}/**/*
```
toy, line 164: Nice

lib/tasks/copyright.rake
NobodysNightmare: I just copied the pattern that we use for all other case statements as well. It just makes sure that you (the rake programmer) don't call `rewrite_copyright` with `create: true` for a format that doesn't support create. The same way that the other case statements make sure that you don't call it with an entirely unsupported format.

Or am I overlooking an existing rake task that now doesn't work anymore due to this change?

lib/tasks/copyright.rake
NobodysNightmare: Rubocop didn't complain, but I can safely multiline this :D

I think Rubocop doesn't care, because `puts` is essentially just a method call and it returns nil.

lib/tasks/copyright.rake
toy: Please ignore, I misunderstood the logic

lib/tasks/copyright.rake
```diff
@@ -143,10 +161,11 @@ namespace :copyright do
...
+      lib/chronic_duration.rb
     ]
 
-    rewrite_copyright("rb", excluded, :rb, args[:path])
+    rewrite_copyright("rb", excluded, :rb, args[:path], create: true)
```
toy, line 168: The `create: true` flag can be copied to most other tasks handling `rb` files

NobodysNightmare: good point, added

toy (approved):

PR #18358, opened 2025-03-18
[#62257] display parent and project prefix in attribute tokens

Kharonus:
# Ticket
[OP#62257](https://community.openproject.org/wp/62257)

- show project or parent attributes tokens

- use localized group keys
- add prefixes in token display texts


brunopagno (approved):
At first I was going to say that it might be simpler to use `js-en.yml` but I think the result here looks great.

Looks great to me :tada:

PR #18366, opened 2025-03-19
implementation/60980 - Define restricted extension on work package journals for query safety

akabiru:
AR association extensions ensure the `restricted_visible` scope is only visible on the work package journals association.

https://github.com/opf/openproject/pull/18091#discussion_r1993034943

### DIFF
--- a/app/models/work_package/journalized.rb
+++ b/app/models/work_package/journalized.rb
@@ -30,7 +30,16 @@ module WorkPackage::Journalized
   extend ActiveSupport::Concern
 
   included do
-    acts_as_journalized
+    acts_as_journalized do
+      def restricted_visible
+        if OpenProject::FeatureDecisions.comments_with_restricted_visibility_active? &&
+            User.current.allowed_in_project?(:view_comments_with_restricted_visibility, proxy_association.owner.project)
+          all
+        else
+          where(restricted: false)
+        end
+      end
+    end
 
     # This one is here only to ease reading
     module JournalizedProcs
--- a/app/models/work_package/journalized.rb
+++ b/app/models/work_package/journalized.rb
@@ -30,7 +30,7 @@ module WorkPackage::Journalized
   extend ActiveSupport::Concern
 
   included do
-    acts_as_journalized do
+    acts_as_journalized journals_association_extension: proc {
       def restricted_visible
         if OpenProject::FeatureDecisions.comments_with_restricted_visibility_active? &&
             User.current.allowed_in_project?(:view_comments_with_restricted_visibility, proxy_association.owner.project)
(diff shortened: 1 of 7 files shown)


akabiru:
Oops- forgot to update the callers. 🏃🏾

brunopagno (commented):
The code changes look okay to me.

I do feel it's going to be more difficult for someone to find the `restricted_visible` method when someone wishes to debug it, but it's a bit of the nature of rails applications.

But it seems to have broken some specs, so we need to look into those before pushing forward

akabiru:
> The code changes look okay to me.
> 
> I do feel it's going to be more difficult for someone to find the `restricted_visible` method when someone wishes to debug it, but it's a bit of the nature of rails applications.

Cheers for the review- tbf, finding the association was already hard enough- and the journalization is encapsulated in one mixin, so hopefully that and the unit tests improve the odds of working it out. Definitely open to ideas on improving this 👍🏾 

> But it seems to have broken some specs, so we need to look into those before pushing forward

yeah, I blindly enabled frozen string literals - which broke some existing string mutations that I'd rather not mess with in this PR.

app/models/work_package/journalized.rb
```diff
@@ -30,7 +30,16 @@ module WorkPackage::Journalized
   extend ActiveSupport::Concern
 
   included do
-    acts_as_journalized
+    acts_as_journalized do
+      def restricted_visible
```
ulferts, line 34: Could we use a parameter for providing this block to the `journals` association instead? When we just use the contents of the block passed to aaj it is somewhat nondescript maybe something like:

```
acts_as_journalized journals_association_extension: -> {
  def restricted_visible ... end
}
```

ulferts (approved):
Overall, I like this change @akabiru . To me, it reads nicer and appears to be more solid. Thanks for picking up my review feedback from before.

I added on point that I think could improve readability but as is the nature of such things, it is opinionated. So feel free to ignore it if you disagree.

app/models/work_package/journalized.rb
akabiru: Good idea- 💡 I believe this also threw off @brunopagno slightly 😅

akabiru: Updated in https://github.com/opf/openproject/pull/18366/commits/1a46fc834eb99d5fdce5a39330e22fbac8d100a9 I ended up using a proc instead of the proposed lambda- as active record will define a dynamic module extension which expects a block. https://github.com/rails/rails/blob/main/activerecord/lib/active_record/associations/builder/collection_association.rb#L25

ulferts (approved):
👍

PR #18367, opened 2025-03-19
Improve fetcher query

brunopagno:
# Ticket
Requested changes at https://github.com/opf/openproject/pull/18091#discussion_r1993072780

Make the query more efficient and in the correct place.

The idea here is that by adjusting the query on the activity providers, we don't need to do a "janky" filter on the fetcher.

### DIFF
--- a/app/models/activities/base_activity_provider.rb
+++ b/app/models/activities/base_activity_provider.rb
@@ -167,9 +167,11 @@ class Activities::BaseActivityProvider
       return query.where(journals_table[:restricted].eq(false))
     end
 
-    permission = :view_comments_with_restricted_visibility
-    allowed_project_ids = Project.allowed_to(user, permission).pluck(:id)
-    query.where(projects_table[:id].in(allowed_project_ids).or(journals_table[:restricted].eq(false)))
+    query.where(
+      projects_table[:id]
+        .in(Project.allowed_to(user, :view_comments_with_restricted_visibility).select(:id).arel)
+        .or(journals_table[:restricted].eq(false))
+    )
   end
 
   def filter_for_event_datetime(query, from, to)
--- a/app/models/activities/base_activity_provider.rb
+++ b/app/models/activities/base_activity_provider.rb
@@ -141,6 +141,7 @@ class Activities::BaseActivityProvider
   def event_selection_query(user, from, to, options)
     query = journals_with_data_query
     query = extend_event_query(query)
+    query = filter_for_visibility(query, user)
     query = filter_for_event_datetime(query, from, to)
     query = restrict_user(query, options)
     restrict_projects(query, user, options)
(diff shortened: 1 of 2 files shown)


app/models/activities/base_activity_provider.rb
```diff
@@ -161,6 +162,16 @@ def apply_limit(query, options)
     end
   end
 
+  def filter_for_visibility(query, user)
+    unless OpenProject::FeatureDecisions.comments_with_restricted_visibility_active?
+      return query.where(journals_table[:restricted].eq(false))
+    end
+
+    permission = :view_comments_with_restricted_visibility
+    allowed_project_ids = Project.allowed_to(user, permission).pluck(:id)
```
akabiru: 🟢  Inline as it's not reused in any place?
```suggestion
    allowed_project_ids = Project.allowed_to(user, :view_comments_with_restricted_visibility).pluck(:id)
```

akabiru (approved):
🪄

app/models/activities/base_activity_provider.rb
ulferts: Calling `pluck` will immediately trigger the sql statement. Using `select(:id)` avoids that.

Using `select(:id)` however doesn't work with arel as done in the line below. But this can be circumvented by calling `.arel` on the scope.

app/models/activities/base_activity_provider.rb
```diff
@@ -161,6 +162,16 @@ def apply_limit(query, options)
...
+    query.where(projects_table[:id].in(allowed_project_ids).or(journals_table[:restricted].eq(false)))
```
ulferts: This line will have to be checked for performance (I didn't). Or statements over two separate columns oftentimes are a source for slow performance.

ulferts (approved):
Thanks for picking up my review feedback @brunopagno. I like this change and only have one small remark to optimize the performance. Otherwise, I think this is good to be merged.

app/models/activities/base_activity_provider.rb
brunopagno: I like my local variables. It makes it easier to debug :sweat_smile:

I'll review it, though.

app/models/activities/base_activity_provider.rb
brunopagno: Okay, for the documenting purposes, I created 10k comments and tested the query with the filtering and without the filtering.

The query was extracted by adding a breakpoint [before calling fill_events on the base_activity_provider.rb](https://github.com/opf/openproject/blob/dev/app/models/activities/base_activity_provider.rb#L75-L81) and doing a `events_query.to_sql`.

Diffing the queries looked correct enough to me. I, then, tried the raw queries on the database and got very similar execution times

-> Without restricted filtering
Sort  (cost=193.23..195.80 rows=1027 width=80) (actual time=1.170..1.200 rows=1048 loops=1)
...
Planning Time: 3.571 ms
Execution Time: 10.876 ms

-> With restricted filtering
Sort  (cost=1799.33..1822.11 rows=9111 width=83) (actual time=12.924..13.138 rows=7196 loops=1)
...
Planning Time: 3.909 ms
Execution Time: 13.423 ms

---

I might have overlooked something, but it looks like the query in the base_activity_provider might not be what makes the activities page so slow.

In any case I believe this is evidence enough that we are not making the query considerably worse than before and thus can move on.

---

(the diff for reference)
```sql
AND (
    "projects"."id" IN (
      SELECT "projects"."id"
      FROM "projects"
      WHERE
        "projects"."id" IN (
          SELECT "projects"."id" FROM "projects"
          WHERE "projects"."active" = TRUE
        )
    )
    OR "journals"."restricted" = FALSE
  )
```

PR #18368, opened 2025-03-19
Correctly handle when expires_in is a string

NobodysNightmare:
Azure/Entra ID seems to return stringly typed expiry dates under certain conditions. We'll now also handle properly readable numeric strings.

Missing expiration times or non-numeric strings, such as "banana" will be treated as not knowing the expiration time.

# Ticket
https://community.openproject.org/wp/62299

# Merge checklist

- [x] Added/updated tests

### DIFF
--- a/modules/openid_connect/spec/lib/open_project/openid_connect/hooks/hook_spec.rb
+++ b/modules/openid_connect/spec/lib/open_project/openid_connect/hooks/hook_spec.rb
@@ -66,6 +66,26 @@ RSpec.describe OpenProject::OpenIDConnect::Hooks::Hook do
         expect(session).to be_empty
       end
     end
+
+    context "when expires_in is missing" do
+      let(:expires_in) { nil }
+
+      it "does not store an expires_in" do
+        call_hook
+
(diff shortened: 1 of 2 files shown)


modules/openid_connect/spec/lib/open_project/openid_connect/hooks/hook_spec.rb
```diff
@@ -66,6 +66,26 @@
...
+        call_hook
+
+        expect(session["omniauth.oidc_expires_in"]).to be_nil
+      end
+    end
+
+    context "when expires_in is passed as a string" do
+      let(:expires_in) { "7200" }
+
+      it "does not store an expires_in" do
```
machisuji: This description seems to be inaccurate! @NobodysNightmare

NobodysNightmare: Well spotted. Apparently the test was written a bit too fast xD

machisuji (approved):
👍 🙏

PR #18371, opened 2025-03-19
Translate more button in meetings table

top4ek:
Use existing translation for button popup.

Simply remove hardcoded "More" string to translatable :label_more.


ulferts (approved):
Thanks @top4ek

PR #18372, opened 2025-03-19
[#62254] add content sanitation

Kharonus:
# Ticket
[OP#62254](https://community.openproject.org/wp/62254)

- allow targeting the position right behind any token

- sanitize on mouseup and on movement key up
- fix display text for existing tokens, if not containing current cursor
- add control spaces after tokens, if a cursor target is needed

### DIFF
--- a/frontend/src/stimulus/controllers/pattern-input.controller.ts
+++ b/frontend/src/stimulus/controllers/pattern-input.controller.ts
@@ -252,9 +252,7 @@ export default class PatternInputController extends Controller {
 
       // if the resulting range is empty it is at the start of the input
       if (testRange.toString() === '') {
-        // add a space
-        // const beforeToken = document.createTextNode(' ');
-        const beforeToken = new Text(CONTROL_SPACE);
+        const beforeToken = document.createTextNode(CONTROL_SPACE);
         const firstContent = this.contentTarget.firstChild as HTMLElement;
         this.contentTarget.insertBefore(beforeToken, firstContent);
 
@@ -276,9 +274,7 @@ export default class PatternInputController extends Controller {
 
       // if the resulting range is empty it is at the end of the input
       if (testRange.toString() === '') {
-        // add a space
-        // const afterToken = document.createTextNode(' ');
-        const afterToken = new Text(CONTROL_SPACE);
+        const afterToken = document.createTextNode(CONTROL_SPACE);
         this.contentTarget.appendChild(afterToken);
 
         this.setRealCaretPositionAtNode(afterToken);
@@ -472,14 +468,14 @@ export default class PatternInputController extends Controller {
 
         const follower = node.nextSibling;
         if (follower === null) {
-          node.after(new Text(CONTROL_SPACE));
+          node.after(document.createTextNode(CONTROL_SPACE));
         } else {
           if (this.isToken(follower)) {
-            node.after(new Text(CONTROL_SPACE));
+            node.after(document.createTextNode(CONTROL_SPACE));
           }
 
           if (this.isText(follower) && !this.isWhitespaceOrControlSpace(follower.wholeText[0])) {
-            node.after(new Text(CONTROL_SPACE));
+            node.after(document.createTextNode(CONTROL_SPACE));
           }
         }
       }
--- a/frontend/src/stimulus/controllers/pattern-input.controller.ts
+++ b/frontend/src/stimulus/controllers/pattern-input.controller.ts
@@ -121,7 +121,6 @@ export default class PatternInputController extends Controller {
       this.sanitizeContent();
     }
 
-    // update cursor
     this.setRange();
   }
 
@@ -139,6 +138,7 @@ export default class PatternInputController extends Controller {
     }
 
     this.tagInvalidTokens();
+    this.sanitizeContent();
 
     // This resets the cursor position without changing it.
     // It is necessary because chromium based browsers try to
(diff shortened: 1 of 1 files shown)


frontend/src/stimulus/controllers/pattern-input.controller.ts
```diff
@@ -452,11 +462,56 @@
...
+        if (!this.containsCursor(node) && node.innerText !== this.validTokenMap[key]) {
+          node.innerText = this.validTokenMap[key] || key;
+        }
+
+        const follower = node.nextSibling;
+        if (follower === null) {
+          node.after(new Text(CONTROL_SPACE));
+        } else {
+          if (this.isToken(follower)) {
+            node.after(new Text(CONTROL_SPACE));
```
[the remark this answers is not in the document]
Kharonus: no, Copilot, stop! that is not correct!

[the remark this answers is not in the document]
Kharonus: bad copilot!

frontend/src/stimulus/controllers/pattern-input.controller.ts
```diff
@@ -40,6 +40,11 @@ type TokenElement = HTMLElement&{ dataset:{ role:'token', prop:string } };
 type ListElement = HTMLElement&{ dataset:{ role:'list_item', prop:string } };
 
 const COMPLETION_CHARACTER = '/';
+const TOKEN_REGEX = /{{([0-9A-Za-z_]+)}}/g;
+
+// A zero-width space character, which is used
+// to have a caret position after tokens
+const CONTROL_SPACE = '\u200B';
```
brunopagno, line 47: Quite minor detail, but `CONTROL_SPACE` sounds a bit like ctrl + space. There might be another name that works.

brunopagno (commented):
I think this is the right path, but we need to be *very* through with inserting and cleaning up tokens.

This might be a Firefox issue, but right now the invisible spaces are interfering with text selection (I am double clicking to select strings in the video).

https://github.com/user-attachments/assets/05f84a1b-c9d4-4dda-b170-8e7f35923601

If I get one of the `CONTROL_SPACE`s between text it behaves a bit weird. That + I could not really set the caret between tokens with the mouse, but I don't really expect that to be possible.

frontend/src/stimulus/controllers/pattern-input.controller.ts
Kharonus: yeah, naming is hard. :D feel free to give me your thoughts on it.

regarding the double click: good find. I already had the idea of merging text together, and I skipped it, because I saw now necessity. Now, I see it :D Will add a logic, that merges text and eliminates the control characters in between.

the double click behaviour that reaches INTO the token is something, I find much less easier to fix. I also don't know, why this is happening. Because it actually IS another node even, why is the selection overlapping?

NobodysNightmare: `TOKEN_SEPARATOR`?

Kharonus: It does not separate tokens, but it enables a focusable character behind a token, independently, whether behind is a Text or a Token.

Kharonus:
Update from discussion with Dominic: 

The behaviour of FireFox is considered buggy, but on there part. There is no sane explanation, why they behave, as they do. If you want details, I can provide in a 1on1.

As a consequence, please test this behaviour on chromium or safari.

frontend/src/stimulus/controllers/pattern-input.controller.ts
brunopagno: I did not suggest anything because I did not have any better ideas. Also probably not worth spending too much effort into this :sweat_smile:. But for the sake of brainstorming in case you feel excited about it

`INVISIBLE_SPACE`
`SEPARATOR_CHARACTER`
`SPLITTY_THINGY` :grimacing:

brunopagno (approved):
I don't like the idea of not supporting firefox on a modern web application, but I understand your reasoning. I'm approving this then.

NobodysNightmare:
> I don't like the idea of not supporting firefox on a modern web application, but I understand your reasoning. I'm approving this then.

+1 on this. If we feel this is a bug in FF, we should _at least_ go the mile to report the bug to FF. But to me it doesn't seem too surprising in general that:

* It's possible to move the cursor past an invisible character
* That an invisible character _intended to separate words_, separated words (double-click behaviour)

I didn't click through this myself yet and probably will not be able to do, before someone presses the merge button.

NobodysNightmare:
> There is no sane explanation, why they behave, as they do

Okay, I had a quick look at the screencast and the code. What I think happens is that the zero-width-space (ZWS) is skipped during cursor navigation. So for this input (`_` is ZWS and `|` is cursor) `foo_bar|`, if I press left arrow three times, I will be at this state: `foo|_bar`. Now if I enter something ("banana"), my state is `foobanana|_bar`. If `foo` is a token as implemented in this PR, it will make sure that it is always followed by a ZWS, so after that the state is `foo_banana|_bar`.

I find the behaviour at least understandable. I guess it stems from the fact that they wanted to avoid no visible cursor movement when the user navigates the cursor. I guess this is why the cursor skips over the invisible whitespace.

No clue how chrome solves this. Do you have to explicitly navigate your cursor over the invisible character there?

NobodysNightmare (changes requested):
Okay, so I just tested things locally in FF and Chrome.

In FF it's even simpler than I thought. The cursor does not _skip_ the ZWS, so I have to tap one extra time when it's there. Chrome seems to skip it when it is behind a token, but otherwise behaves the same as FF. To me FF behaves more consistent than Chrome.

I also succeeded in "capturing" a NWS in Chrome: If you carefully delete a token, the ZWS remains and you can place text to its left and right:

[Capturing ZWS in Chromium](https://github.com/user-attachments/assets/c36fe520-b570-4626-a2b6-a14469742fb7)

Kharonus:
I had a lengthy talk with dominic and we came to the agreement, that we actually only adhere to specific browser behaviour. So, the usage of the input is different for FF and for chromium based browsers.

here are the known issues:
- on FF you can get "before" the ZWSP, and hence having it located in between text nodes
  - this leads to unexpected clicking double click selections
  - this leads to "double" arrow key navigation needed to overcome the ZWSP
- on chrome, you cannot get "before" any character (except it is the very first of the input), hence, you can use "DEL" on the first character of a token only from outside a token
  - this leads to the token being repaired directly

Both issues do not break the functionality of the component per se. The missing DEL is probably the thing, that impacts more. But at least both do not introduce any problems with actually persisting the data.

For that reason, I'd again raise the proposal: Let's merge this state, and list both issues as "known bugs". WDYT? @NobodysNightmare @brunopagno

brunopagno (approved):
@Kharonus yes, I'm good with the decision. Thanks for investigating it and dealing with the weirdness of _content_editable_ :sweat_smile: 

:ship: :ship: :ship:

NobodysNightmare:
Sounds good to me.

PR #18373, opened 2025-03-19
[#62319] PDF Export: Templates selection not translated

as-op:
# Ticket

https://community.openproject.org/work_packages/62319

Fix a bug with translations cached in a module constant

Don't use a constant for strings with translations


ulferts (approved):
So obvious in hindsight. 🤣 But as always, only in hindsight.

PR #18377, opened 2025-03-19
Implements the Authentication Configuration Checks

mereghost:
### Related Work Package: [OP#62100](https://community.openproject.org/work_packages/62100)

Implements the OIDC connection tests and some elementary OAuth2 tests.

### DIFF
--- a/modules/storages/app/common/storages/peripherals/connection_validators/nextcloud/authentication_validator.rb
+++ b/modules/storages/app/common/storages/peripherals/connection_validators/nextcloud/authentication_validator.rb
@@ -48,6 +48,36 @@ module Storages
 
           private
 
+          def validate_oauth
+            @results = {
+              existing_token: CheckResult.skipped(:existing_token),
+              user_bound_request: CheckResult.skipped(:user_bound_request)
+            }
+
+            oauth_token
+            user_bound_request
+          end
+
+          def oauth_token
+            if OAuthClientToken.where(user: @user, oauth_client: @storage.oauth_client).any?
+              pass_check(:existing_token)
+            else
+              warn_check(:existing_token, message(:oauth_token_missing))
+              throw :interrupted
+            end
+          end
+
+          def user_bound_request
+            Registry["nextcloud.queries.files"]
+              .call(storage: @storage, auth_strategy:, folder: ParentFolder.new("/")).on_failure do
+              fail_check(__method__, message("oauth_request_#{it.result}"))
+            end
+
+            pass_check(__method__)
+          end
+
+          def auth_strategy = Registry["nextcloud.authentication.user_bound"].call(storage: @storage, user: @user)
+
           def validate_sso
             @results = {
               non_provisioned_user: CheckResult.skipped(:non_provisioned_user),
@@ -98,8 +128,6 @@ module Storages
             fail_check(__method__, message(error_code))
             throw :interrupted
           end
-
-          def validate_oauth = fail
         end
       end
     end
--- a/modules/storages/app/common/storages/peripherals/connection_validators/nextcloud/authentication_validator.rb
+++ b/modules/storages/app/common/storages/peripherals/connection_validators/nextcloud/authentication_validator.rb
@@ -38,21 +38,14 @@ module Storages
             @user = User.current
           end
 
-          def call
-            catch :interrupted do
-              @storage.authenticate_via_idp? ? validate_sso : validate_oauth
-            end
-
-            @results
-          end
-
           private
 
+          def validate
+            @storage.authenticate_via_idp? ? validate_sso : validate_oauth
+          end
+
           def validate_oauth
-            @results = {
-              existing_token: CheckResult.skipped(:existing_token),
-              user_bound_request: CheckResult.skipped(:user_bound_request)
-            }
+            register_checks(:existing_token, :user_bound_request)
 
             oauth_token
             user_bound_request
@@ -62,28 +55,23 @@ module Storages
             if OAuthClientToken.where(user: @user, oauth_client: @storage.oauth_client).any?
               pass_check(:existing_token)
             else
-              warn_check(:existing_token, message(:oauth_token_missing))
-              throw :interrupted
+              warn_check(:existing_token, message(:oauth_token_missing), halt_validation: true)
             end
           end
 
           def user_bound_request
             Registry["nextcloud.queries.files"]
               .call(storage: @storage, auth_strategy:, folder: ParentFolder.new("/")).on_failure do
-              fail_check(__method__, message("oauth_request_#{it.result}"))
+              fail_check(:user_bound_request, message("oauth_request_#{it.result}"))
             end
 
-            pass_check(__method__)
+            pass_check(:user_bound_request)
           end
 
           def auth_strategy = Registry["nextcloud.authentication.user_bound"].call(storage: @storage, user: @user)
 
           def validate_sso
-            @results = {
-              non_provisioned_user: CheckResult.skipped(:non_provisioned_user),
-              non_oidc_provisioned_user: CheckResult.skipped(:non_oidc_provisioned_user),
-              token_usability: CheckResult.skipped(:token_usability)
-            }
+            register_checks(:non_provisioned_user, :provisioned_user_provider, :token_usability)
 
             non_provisioned_user
             non_oidc_provisioned_user
@@ -92,19 +80,17 @@ module Storages
 
           def non_provisioned_user
             if @user.identity_url.present?
-              pass_check(__method__)
+              pass_check(:non_provisioned_user)
             else
-              warn_check(__method__, message(:oidc_non_provisioned_user))
-              throw :interrupted
+              warn_check(:non_provisioned_user, message(:oidc_non_provisioned_user), halt_validation: true)
             end
           end
 
           def non_oidc_provisioned_user
             if @user.authentication_provider.is_a?(OpenIDConnect::Provider)
-              pass_check(__method__)
+              pass_check(:provisioned_user_provider)
             else
-              warn_check(__method__, message(:oidc_non_oidc_user))
-              throw :interrupted
+              warn_check(:provisioned_user_provider, message(:oidc_non_oidc_user), halt_validation: true)
             end
           end
 
@@ -112,7 +98,7 @@ module Storages
             service = OpenIDConnect::UserTokens::FetchService.new(user: @user)
 
             result = service.access_token_for(audience: @storage.audience)
-            return pass_check(__method__) if result.success?
+            return pass_check(:token_usability) if result.success?
 
             error_code = case result.failure
                          in { code: /token_exchange/ | :unable_to_exchange_token }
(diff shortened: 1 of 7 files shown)


modules/storages/app/common/storages/peripherals/connection_validators/nextcloud/authentication_validator.rb
```diff
@@ -0,0 +1,135 @@
...
+              @storage.authenticate_via_idp? ? validate_sso : validate_oauth
+            end
+
+            @results
+          end
+
+          private
+
+          def validate_oauth
+            @results = {
```
NobodysNightmare: :red_circle:  The usage of the `@results` instance variables seems to be all over the place to me. `BaseValidator` relies on the fact that it is defined when using `pass_check`, `warn_check` and `fail_check`. However, `BaseValidator didn't make sure that this instance variable exists.

`call` (in a subclass of `BaseValidator`) returns the instance variable, but it didn't fill in anything itself, relying on the variable being set by both methods called from it.

Setting the initial value happens in two different places for this subclass and in one method in another subclass.

That was the problem statement from my side, I think we should make sure that access to `@results` is better controlled.

---

Given what I currently see, my preferred solution would probably be to never use `@results` outside of  `BaseValidator`. We could add a single method to `BaseValidator` (name is a proposal):

```ruby
def register_check(name) # or prepare_check or register_result or ...
  name = name.to_sym
  @results[name] = CheckResult.skipped(name)
end

# the initializer would initialize an empty hash
```

Then subclasses would just call this method, e.g.

```ruby
register_check :non_provisioned_user
register_check :non_oidc_provisioned_user
register_check :token_usability

# if you are hungry for a nice DSL you could also allow many names
register_checks :a, :b, :c
```

As a last "problem" we'd need to figure out what to do with `call`.

Option A: `call` does not return results, you validate differently:

```ruby
validator.call
validator.results
```

Option B: `call` just returns `results`, which are preferably made accessible through an attribute reader.

Option C: `call` is actually implemented in `BaseValidator` and subclasses implement validations in a different method. This is my favorite approach. Example follows:

```ruby
class BaseValidator
  def call
    @results = {} #resetting results in built-into call
    catch :interrupted do
      validate
    end
    @results
  end

  def validate; raise SubclassResponsibility; end
end

class MyValidator < BaseValidator
  def validate
    register_check :foo
    fail_check :foo
  end
end
```

I just realize that Option C solves a problem I overlooked so far: Handling the throwable `interrupted` was **also** all over the place: The base class threw it and the subclasses needed to know to catch it. In Option C, only `BaseValidator is concerned with it and subclasses can focus on validation logic with a nice DSL.

modules/storages/app/common/storages/peripherals/connection_validators/nextcloud/authentication_validator.rb
```diff
@@ -0,0 +1,135 @@
...
+            oauth_token
+            user_bound_request
+          end
+
+          def oauth_token
+            if OAuthClientToken.where(user: @user, oauth_client: @storage.oauth_client).any?
+              pass_check(:existing_token)
+            else
+              warn_check(:existing_token, message(:oauth_token_missing))
+              throw :interrupted
```
NobodysNightmare: :yellow_circle: In accordance with my earlier stated opinion that we shouldn't throw (pun intended) `interrupted` around everywhere, what do you think about exposing this as an API, e.g. through `warn_check(:existing_token, message, critical: true)` or `critical_warn_check(...)`?

It would be great if subclasses wouldn't need to know the intricates of which symbols are thrown in a validator. Similarly to how I don't need to know how Sinatra's `halt` is implemented (probably also achievable by throwing just the right thing).

modules/storages/app/common/storages/peripherals/connection_validators/nextcloud/authentication_validator.rb
```diff
@@ -0,0 +1,135 @@
...
+            end
+          end
+
+          def user_bound_request
+            Registry["nextcloud.queries.files"]
+              .call(storage: @storage, auth_strategy:, folder: ParentFolder.new("/")).on_failure do
+              fail_check(__method__, message("oauth_request_#{it.result}"))
```
NobodysNightmare: :green_circle: A bit much meta-programming for my taste, since it doesn't seem to serve any other purpose than to avoid repeating the method name, which _happens to coincide_ with the name of a check we registered earlier.

Possible pitfalls that I see:
* searching for `:user_bound_request` does not show me this place (even though I saw that we registered this symbol as a test name earlier)
* Renaming the method now has the side effect of us not updating the correct check result anymore

IMO it's just not worth the meta.

modules/storages/app/common/storages/peripherals/connection_validators/nextcloud/authentication_validator.rb
```diff
@@ -0,0 +1,135 @@
...
+
+            oauth_token
+            user_bound_request
+          end
+
+          def oauth_token
+            if OAuthClientToken.where(user: @user, oauth_client: @storage.oauth_client).any?
+              pass_check(:existing_token)
+            else
+              warn_check(:existing_token, message(:oauth_token_missing))
```
Kharonus: ❓ why is it `warn_check` and not a `fail_check`?

modules/storages/app/common/storages/peripherals/connection_validators/nextcloud/authentication_validator.rb
```diff
@@ -0,0 +1,135 @@
...
+                         in { code: /token_refresh/ }
+                           :oidc_cant_refresh_token
+                         in { code: :no_token_for_audience }
+                           :oidc_cant_acquire_token
+                         else
+                           :unknown_error
+                         end
+
+            fail_check(__method__, message(error_code))
+            throw :interrupted
```
Kharonus: 🟡 no need to throw, if `fail_check` was called

modules/storages/app/common/storages/peripherals/connection_validators/nextcloud/authentication_validator.rb
Kharonus: same in `non_provisioned_user` and `non_oidc_provisioned_user`

modules/storages/app/common/storages/peripherals/connection_validators/nextcloud/authentication_validator.rb
NobodysNightmare: :yellow_circle: Unreachable code. `fail_check` already throws.

NobodysNightmare (changes requested):
In addition to the :red_circle: that I cleverly posted immediately, I added all other traffic light colors to round it off.

In general I like what those validators might look like soonish.

modules/storages/app/common/storages/peripherals/connection_validators/nextcloud/authentication_validator.rb
NobodysNightmare: I guess because the check doesn't quite mean that stuff is entirely broken, but rather "I can't check further, that makes me sad". So we need to abort, but we can't say for sure that it's broken.

modules/storages/app/common/storages/peripherals/connection_validators/nextcloud/authentication_validator.rb
NobodysNightmare: More naming ideas after writing in Eric's thread:

* `warn_check(..., abort: true`)
* `abort_with_warning(...)`

modules/storages/app/common/storages/peripherals/connection_validators/nextcloud/authentication_validator.rb
mereghost: Yep, it is not a failure as I can't simply get a token without going through the entire OAuth2 process.

This makes the context slightly different. =/

modules/storages/app/common/storages/peripherals/connection_validators/nextcloud/authentication_validator.rb
mereghost: Oh! I missed that.

modules/storages/app/common/storages/peripherals/connection_validators/nextcloud/authentication_validator.rb
mereghost: <del>I'm keeping it as so, as</del> I feel that I'll end up renaming all methods and keys once I gather all the results in one place.

Specially as the names represented negative outcomes as we where chaining calls and now we have proper checks with 2 relevant different outcomes.

modules/storages/app/common/storages/peripherals/connection_validators/nextcloud/authentication_validator.rb
mereghost: I fully agree with you on this. 🤣 

I'm playing with how to make it nicer overall, but I had to have some stuff in place so I can see where to take it and what would make sense.

My original idea was identical to yours: have a way to register the checks and never have to worry about the results. After a couple of tests, I concluded that I needed the actual implementations before I could extract this behaviour.

`pass_check` (etc) was born as the start of this and I'm slowly cleaning up the interface, as `BaseValidator` was just introduced to start to gobble those details.

I was already moving towards what you propose for the AMPF part of the validations.

modules/storages/app/common/storages/peripherals/connection_validators/nextcloud/authentication_validator.rb
mereghost: Removed the calls to `__method__`.

modules/storages/app/common/storages/peripherals/connection_validators/nextcloud/authentication_validator.rb
mereghost: I added the argument `halt_validation` and didn't add a specialized method for now.

modules/storages/app/common/storages/peripherals/connection_validators/nextcloud/base_validator.rb
```diff
@@ -0,0 +1,85 @@
...
+
+            @results
+          end
+
+          private
+
+          def validate = raise Errors::SubclassResponsibility
+
+          def register_checks(*keys)
+            @results = keys.to_h { [it, CheckResult.skipped(it)] }
```
NobodysNightmare, line 53: :green_circle: The current implementation of this implies that I can only call `register_checks` once, never twice.
 
I think that's enough, but it limits the ways in which I can write my validator a bit. E.g. this layout would not work:

```ruby
def validate
  register_checks(:common_check)
  if rand < 0.5
    register_checks(*a_checks)
    validate_a
  else
    register_checks(*b_checks)
    validate_b
  end
end
```

The other issue is that if I for whatever reason don't add any tests, we'd now return `nil` instead of returning `{}`. As indicated by the green traffic light, I think both of these problems are absolutely minor, but I wanted to explicitly write them down at least.

NobodysNightmare (approved):

PR #18384, opened 2025-03-20
[#62316] PDF Export Report: no attributes are exported if the default columns of a query is used

as-op:
# Ticket
https://community.openproject.org/work_packages/62316

OP clears out the columns in a query if the default columns are used. So we can't differentiate for an empty query columns array, if this should be empty for an export (PDF report only) or the default columns are used.

Make it explicit and use a dedicated parameter for the export job.

# Merge checklist

- [x] Added/updated tests


ulferts (approved):
It is unfortunate, that there isn't a better approach to it. At least, I couldn't find one. It would be good if the query could differentiate between wanting the default queries and no columns in a clear way.

As discussed, let's hold on merging until 15.4.1 is shipped.

PR #18388, opened 2025-03-20
implementation/60983 define quoting criteria for work package comments with restricted visibility

akabiru:
# Ticket

https://community.openproject.org/wp/60983

Quoting a restricted comment implicitly enables restriction if it did not exist before. When quoting both restricted and unrestricted comments within the same comment- the restricted state should triumph

This is to limit the risk of accidental/inadvertent leaking of the comment.

## Screenshots

https://github.com/user-attachments/assets/2a5d59ff-65b7-40e3-a6cb-355a2d41f948

Extracted quoting functionality from index.controller- leveraged [stimulus outlets](https://stimulus.hotwired.dev/reference/outlets) for inter-controller interfacing. This might be a good option to incrementally break down this controller @brunopagno 

# Merge checklist

- [x] Added/updated tests

### DIFF
--- a/app/components/concerns/work_packages/activities_tab/stimulus_controllers.rb
+++ b/app/components/concerns/work_packages/activities_tab/stimulus_controllers.rb
@@ -1,0 +1,45 @@
+# frozen_string_literal: true
+
+# -- copyright
+# OpenProject is an open source project management software.
+# Copyright (C) the OpenProject GmbH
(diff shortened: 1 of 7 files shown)


brunopagno (approved):
Double checked locally it looks good. Nice job extracting stuff from the controller. :tada:

PR #18389, opened 2025-03-20
[62341] Correctly convert dates when timezone is negative in datepicker

cbliard:
# Ticket

https://community.openproject.org/wp/62341

The dates are off by 1 or 2 days when in a negative timezone, both in the calendar view and the date fields.

## Screenshots

Video in to linked ticket.

flatpickr is using local dates at midnight internally and in its callback functions. In datepicker component and stimulus controller we are using UTC dates at midnight.

And the timezone service is assuming local time when converting to iso dates.

Usingtimezones with positive offset is ok, but with negative offset, the conversions and comparisons do not work properly.

Fixed it by explicitly converting from utc date to iso string, and using iso date strings when setting flatpickr dates, avoiding any issues.

It could be a good idea to use iso date strings everywhere as much as possible instead of always jumping back and forth from dates to iso string.

# Merge checklist

- [x] Added/updated tests

### DIFF
--- a/spec/support/cuprite_setup.rb
+++ b/spec/support/cuprite_setup.rb
@@ -108,7 +108,10 @@ register_better_cuprite "en"
 RSpec.configure do |config|
   config.around(:each, :js, type: :feature) do |example|
     # Skip if driver is explicitly requested
-    next if example.metadata[:driver]
+    if example.metadata[:driver]
+      example.run
+      next
+    end
 
     original_driver = Capybara.javascript_driver
 
--- a/frontend/src/app/core/datetime/timezone.service.ts
+++ b/frontend/src/app/core/datetime/timezone.service.ts
@@ -132,6 +132,14 @@ export class TimezoneService {
     return moment.duration(input, unit).toISOString();
   }
 
+  public utcDateToISODateString(date:Date):string {
+    return moment.utc(date).format('YYYY-MM-DD');
+  }
+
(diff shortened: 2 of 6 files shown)


spec/support/cuprite_setup.rb
```diff
@@ -108,7 +108,10 @@ def register_better_cuprite(language, name: :"better_cuprite_#{language}")
 RSpec.configure do |config|
   config.around(:each, :js, type: :feature) do |example|
     # Skip if driver is explicitly requested
-    next if example.metadata[:driver]
+    if example.metadata[:driver]
+      example.run
+      next
+    end
```
klaustopher, line 114: So, since before we did not call `example.run` in the around filter we did never run the test and now we are running it, but should we not set the `Capybara.javascript_driver` as well? Or is that automatically done by the `driver:` metadata?

klaustopher (approved):
approved in general, would be nice to clarify the one comment I added

spec/support/cuprite_setup.rb
klaustopher: Bildschirmfoto 2025-03-20 um 11 44 58

Setting the driver is part of capybara itself, so yes ... This does what it says on the tin. Good catch!

cbliard: It's automatically done by the `driver:` metadata. It's handled by Capybara.

PR #18390, opened 2025-03-20
Translate TimeEntry model name

top4ek:
Translate TimeEntry model name when(and not only) mass deleting work packages

## Screenshots
Screenshot_20250320_133309

Codebase's fine and use I18n already. Just add translation for model.

### DIFF
--- a/modules/costs/config/locales/en.yml
+++ b/modules/costs/config/locales/en.yml
@@ -65,8 +65,10 @@ en:
         spent_on: Date
         start_time: Start time
         end_time: Finish time
-
     models:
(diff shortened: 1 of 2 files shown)


klaustopher (approved):

klaustopher (changes requested):
As the model is defined in the `costs`module, the translation should be added to its locale file.

Also, we do not use `few` and `many`... We only need to provide `one` and `other` keys

klaustopher (approved):

klaustopher:
Thank you very much!

PR #18393, opened 2025-03-20
Validate scopes of JWTs

NobodysNightmare:
So far we are only accepting JWTs for requests to
APIv3 and our scopes are not more fine-grained than that.

However, warden strategies are generally responsible for validating
the scopes of a token (e.g. compare DoorkeeperOAuth) to the scope
of the currently accessed API.

# Ticket
https://community.openproject.org/wp/62360

# Merge checklist

- [x] Added/updated tests

### DIFF
--- a/lib_static/open_project/authentication/strategies/warden/jwt_oidc.rb
+++ b/lib_static/open_project/authentication/strategies/warden/jwt_oidc.rb
@@ -25,6 +25,11 @@ module OpenProject
             ::OpenIDConnect::JwtParser.new(required_claims: ["sub"]).parse(@access_token).either(
               ->(payload_and_provider) do
                 payload, provider = payload_and_provider
+                unless valid_scope?(payload)
+                  return fail_with_header! error: "insufficient_scope",
+                                           error_description: "Requires scope #{scope} to access this resource."
+                end
+
                 user = User.find_by(identity_url: "#{provider.slug}:#{payload['sub']}")
                 authentication_result(user)
               end,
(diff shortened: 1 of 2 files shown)


NobodysNightmare:
Keeping this PR as a draft until we decided whether we need to temporarily disable this check for APIv3 specifically or not.

lib_static/open_project/authentication/strategies/warden/jwt_oidc.rb
```diff
@@ -25,6 +25,11 @@ def authenticate!
             ::OpenIDConnect::JwtParser.new(required_claims: ["sub"]).parse(@access_token).either(
               ->(payload_and_provider) do
                 payload, provider = payload_and_provider
+                unless valid_scope?(payload)
+                  return fail_with_header! error: "insufficient_scope",
```
Kharonus, line 29: 🟡 I'd rather go with `invalid_scope` then insufficient. insufficient sounds a bit like "you are allowed to read, but not to write".

I'd like to be convinced otherwise :D but right now, I wouldn't try to interpret too much into anything the requester provided.

Edit: I just saw, that this error is the same as used OAuth strategy. So, we might as well reuse and not change it.

Kharonus (approved):

lib_static/open_project/authentication/strategies/warden/jwt_oidc.rb
NobodysNightmare: This is essentially not our decision. This exact value is specified as part of [RFC 6750](https://datatracker.ietf.org/doc/html/rfc6750#section-6.2.3), i.e. it's a registered error code.

PR #18396, opened 2025-03-20
Activity for deleted project life cycle steps

toy:
Noticed while implementing [OP#62330](https://community.openproject.org/wp/62330) that project life cycle step activity formatters will raise if life cycle steps were deleted


dombesz (approved):
Nice catch!👍

toy:
Forgot to merge, redone as db867779 in #18439

PR #18400, opened 2025-03-21
Those 7 method will be called by Journals::CreateService::Attachable and others

Eric-Guo:

Fix the #17595, wrongly mark those method private at d67c9771a9488de4dd5f93e7c070372243aec447

## Screenshots

Below is one of example error stack in production.

[NoMethodError] private method `journable_id' called for an instance of Journals::CreateService: app/services/journals/create_service/association.rb:49:in `journable_id';
app/services/journals/create_service/attachable.rb:65:in `changes_sql';
app/services/journals/create_service.rb:456:in `block in select_changed_sql';
app/services/journals/create_service.rb:453:in `each';
app/services/journals/create_service.rb:453:in `select_changed_sql';
app/services/journals/create_service.rb:213:in `journal_modification_sql';
app/services/journals/create_service.rb:194:in `create_journal_sql';
app/services/journals/create_service.rb:85:in `create_journal';
app/services/journals/create_service.rb:54:in `block in call';
app/services/journals/create_service.rb:53:in `call';
lib_static/plugins/acts_as_journalized/lib/acts/journalized/save_hooks.rb:64:in `block in save_journals';
lib_static/plugins/acts_as_journalized/lib/acts/journalized/save_hooks.rb:94:in `with_ensured_journal_attributes';
lib_static/plugins/acts_as_journalized/lib/acts/journalized/save_hooks.rb:61:in `save_journals';
openproject-labour_tracking (04fd6cb6dde6) app/services/save_project_time_entry_service.rb:68:in `create_or_update_time_entry';
openproject-labour_tracking (04fd6cb6dde6) app/services/save_project_time_entry_service.rb:21:in `block in call';
openproject-labour_tracking (04fd6cb6dde6) app/services/save_project_time_entry_service.rb:15:in `each';
openproject-labour_tracking (04fd6cb6dde6) app/services/save_project_time_entry_service.rb:15:in `each_with_index';
openproject-labour_tracking (04fd6cb6dde6) app/services/save_project_time_entry_service.rb:15:in `call';
openproject-labour_tracking (04fd6cb6dde6) app/controllers/th_keyins_controller.rb:85:in `create'

I guess no other choice...

# Merge checklist

### DIFF
--- a/app/services/journals/create_service.rb
+++ b/app/services/journals/create_service.rb
@@ -61,9 +61,40 @@ module Journals
       end
     end
 
-    private
-
     def journable_id = journable.id
+
+    def journable_class_name
+      journable.class.base_class.name
+    end
+
+    def normalize_newlines_sql(column)
+      "REGEXP_REPLACE(COALESCE(#{column},''), '\\r\\n', '\n', 'g')"
+    end
+
+    def cleanup_predecessor_for(predecessor, table_name, column, referenced_id)
+      return "SELECT 1" unless predecessor
+
+      sanitize(<<~SQL.squish, column => predecessor.send(referenced_id))
+        DELETE
+        FROM
+         #{table_name}
+        WHERE
+         #{column} = :#{column}
+      SQL
+    end
+
+    def id_from_inserted_journal_sql
+      "(SELECT id FROM inserted_journal)"
+    end
+
+    def only_if_created_sql
+      "EXISTS (SELECT * from inserted_journal)"
+    end
+
+    delegate :sanitize,
+             to: ::OpenProject::SqlSanitization
+
+    private
 
     # If the journalizing happens within the configured aggregation time, is carried out by the same user, has an
     # identical cause and only the predecessor or the journal to be created has notes, the changes are aggregated.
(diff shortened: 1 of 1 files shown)


dombesz:
Hi @Eric-Guo,
Thanks for the report! According to the call stack you provided, the problem is specific to your fork and it stems from `save_project_time_entry_service.rb:21`. Since this problem is specific to your fork, I suggest fixing it there.

Eric-Guo:
The code is only `time_entry.save` and it will trigger the TimeEntry `after_save :save_journals` in acts_as_journalized, both OP code, I guess you will see similar error for any time entry save action.

ulferts:
@Eric-Guo , we really tried to reproduce it by various means but couldn't reproduce the problem within our code base. Could you please provide a way of reproducing the problem?

toy:
@Eric-Guo Rails `delegate` generates method that uses `...` argument passing (https://github.com/rails/rails/blob/v8.0.2/activesupport/lib/active_support/delegation.rb#L134-L143, https://github.com/rails/rails/blob/v8.0.2/activesupport/lib/active_support/delegation.rb#L77-L105) and there seem to be a bug in ruby 3.4 that causes visibility of receiver to be ignored, see report https://bugs.ruby-lang.org/issues/21196.

Also making all those methods public on `Journals::CreateService` doesn't look like a good way to resolve it and I'm going to revert changes of d67c9771a9488de4dd5f93e7c070372243aec447

Eric-Guo:
@toy You are right, I'm still running in OP production in Rocky Linux 8.10, because nokogiri and many gems need new GLIB_C 2.29, so have to running in Ruby 3.3.7 and not 3.4.2, so I meet this problem. 🤦‍♂️

PR #18408, opened 2025-03-21
Revert "delegate to CreateService instead of extracting methods to Helpers module"

toy:
This reverts commit d67c9771a9488de4dd5f93e7c070372243aec447.

It shouldn't be possible to delegate to a private method, but it was hidden by a bug in ruby 3.4 (https://bugs.ruby-lang.org/issues/21196)

It is a better option than making all methods public instead (opf/openproject#18400)


dombesz (approved):
Thanks @toy ! 👍

PR #18409, opened 2025-03-22
Define user @mention criterion for work package comments with restricted visibility

akabiru:
# Ticket

https://community.openproject.org/wp/60988

- [x] It should not be possible to @mention users who will not be able to view the comment with restricted visibilty (i.e, the drop-down needs to exclude users without the requisite permissions)
- [x] If a user who does not have the permissions to view restricted-visibility comments is mentioned in a regular comment and the 'Restrict visibility' checkbox is then checked, the @mention is turned into plain text. Unchecking the "Restrict visibility" checkbox will not re-convert previously stripped @mentions back to links.
- [x] Cover edit case

## Screenshots

https://github.com/user-attachments/assets/44b1c6bb-2e93-4f9a-85d3-f399cf87a502

Introduce a sanitization endpoint on the server side that converts users mentions that do not have restricted visibility permissions- the functionality is further applied in create/update of comments.

# Merge checklist

- [x] Added/updated tests

### DIFF
--- a/app/controllers/work_packages/activities_tab_controller.rb
+++ b/app/controllers/work_packages/activities_tab_controller.rb
@@ -139,6 +139,13 @@ class WorkPackages::ActivitiesTabController < ApplicationController
     respond_with_turbo_streams
   end
 
+  def sanitize_restricted_mentions
+    sanitizer = WorkPackages::ActivitiesTab::RestrictedMentionsSanitizer.new(@work_package, journal_params[:notes])
+    sanitized_notes = sanitizer.call
+
+    render plain: sanitized_notes
+  end
+
   def toggle_reaction # rubocop:disable Metrics/AbcSize
     emoji_reaction_service =
       if @journal.emoji_reactions.exists?(user: User.current, reaction: params[:reaction])
--- a/app/controllers/work_packages/activities_tab_controller.rb
+++ b/app/controllers/work_packages/activities_tab_controller.rb
@@ -140,10 +140,7 @@ class WorkPackages::ActivitiesTabController < ApplicationController
   end
 
   def sanitize_restricted_mentions
-    sanitizer = WorkPackages::ActivitiesTab::RestrictedMentionsSanitizer.new(@work_package, journal_params[:notes])
-    sanitized_notes = sanitizer.call
-
-    render plain: sanitized_notes
+    render plain: sanitized_journal_notes
   end
 
   def toggle_reaction # rubocop:disable Metrics/AbcSize
@@ -228,6 +225,10 @@ class WorkPackages::ActivitiesTabController < ApplicationController
     User.current.preference&.comments_sorting || OpenProject::Configuration.default_comment_sort_order
   end
 
+  def sanitized_journal_notes
+    WorkPackages::ActivitiesTab::RestrictedMentionsSanitizer.new(@work_package, journal_params[:notes]).call
+  end
+
   def journal_params
     params.expect(journal: %i[notes restricted])
   end
@@ -302,12 +303,15 @@ class WorkPackages::ActivitiesTabController < ApplicationController
   end
 
   def create_journal_service_call
+    restricted = to_boolean(journal_params[:restricted], false)
+    notes = restricted ? sanitized_journal_notes : journal_params[:notes]
+
     AddWorkPackageNoteService
       .new(user: User.current,
            work_package: @work_package)
-      .call(journal_params[:notes],
+      .call(notes,
             send_notifications: to_boolean(params[:notify], true),
-            restricted: to_boolean(journal_params[:restricted], false))
+            restricted:)
   end
 
   def to_boolean(value, default)
@@ -315,9 +319,8 @@ class WorkPackages::ActivitiesTabController < ApplicationController
   end
 
   def update_journal_service_call
-    Journals::UpdateService.new(model: @journal, user: User.current).call(
-      notes: journal_params[:notes]
-    )
+    notes = @journal.restricted? ? sanitized_journal_notes : journal_params[:notes]
+    Journals::UpdateService.new(model: @journal, user: User.current).call(notes:)
   end
 
   def generate_time_based_update_streams(last_update_timestamp)
--- a/app/controllers/work_packages/activities_tab_controller.rb
+++ b/app/controllers/work_packages/activities_tab_controller.rb
@@ -141,6 +141,9 @@ class WorkPackages::ActivitiesTabController < ApplicationController
 
   def sanitize_restricted_mentions
     render plain: sanitized_journal_notes
+  rescue StandardError => e
+    handle_internal_server_error(e)
+    respond_with_turbo_streams
   end
 
   def toggle_reaction # rubocop:disable Metrics/AbcSize
@@ -226,7 +229,7 @@ class WorkPackages::ActivitiesTabController < ApplicationController
   end
 
   def sanitized_journal_notes
-    WorkPackages::ActivitiesTab::RestrictedMentionsSanitizer.new(@work_package, journal_params[:notes]).call
+    WorkPackages::ActivitiesTab::RestrictedMentionsSanitizer.sanitize(@work_package, journal_params[:notes])
   end
 
   def journal_params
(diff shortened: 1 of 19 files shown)


app/models/queries/principals/filters/restricted_mentionable_on_work_package_filter.rb
```diff
@@ -0,0 +1,81 @@
...
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program; if not, write to the Free Software
+# Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+#
+# See COPYRIGHT and LICENSE files for more details.
+# ++
+
+class Queries::Principals::Filters::RestrictedMentionableOnWorkPackageFilter <
```
akabiru, line 31: @ulferts I did not manage to easily extend the current `MentionableOnWorkPackageFilter`:

> https://github.com/opf/openproject/pull/18409/commits/d00cda7f652856144677f37f123682644c5d2888 _MentionableOnWorkPackageFilter does not depend solely on permissions, rather on project membership whilst RestrictedMentionableOnWorkPackageFilter depends primarily on a permission- hence we cannot easily reuse the `User.allowed_members_on_work_package(action, project)` method that would indeed be more concise- it's possible this can be revisited in the near future_

akabiru: AFAICT, we will not be able to use the `User.allowed_members_on_work_package(action, project)` on the restricted filter too after all in it's current shape, as it excludes Groups- which are valid mentions. The resulting query is also much simpler. This perhaps explains why the `MentionableOnWorkPackageFilter` did also not use it.

https://github.com/opf/openproject/pull/18409/commits/cc75dc117b766f3cc88f545b7652e71e104c1d92

app/models/queries/principals/filters/restricted_mentionable_on_work_package_filter.rb
Kharonus, line 1: ❓ What are those filters for? I run a collection query on principals and add  a filter `{"restricted_mentionable_on_work_package":{"operator":"=","values":["72"]}}` and that gives me principals that are available for mentions in restricted comments on work package of id 72?

app/controllers/work_packages/activities_tab_controller.rb
```diff
@@ -295,22 +303,24 @@ def update_index_component
...
             send_notifications: to_boolean(params[:notify], true),
-            restricted: to_boolean(journal_params[:restricted], false))
+            restricted:)
   end
 
   def to_boolean(value, default)
     ActiveRecord::Type::Boolean.new.cast(value.presence || default)
   end
 
   def update_journal_service_call
```
Kharonus, line 324: ❓ If a user gets the permission `:view_comments_with_restricted_visibility` retracted, but was before @-mentioned in a journal, the journal will only get updated, if there is any other update, right?

frontend/src/stimulus/controllers/dynamic/work-packages/activities-tab/restricted-comment.controller.ts
```diff
@@ -36,28 +36,68 @@ export default class RestrictedCommentController extends Controller {
...
+      const editorData = this.ckEditorInstance.getData({ trim: false });
+      if (editorData.length === 0) return;
+
+      const sanitizePath = `/work_packages/${this.workPackagesActivitiesTabIndexOutlet.workPackageIdValue}/activities/sanitize_restricted_mentions`;
+      const csrfToken = (document.querySelector('meta[name="csrf-token"]') as HTMLMetaElement).content;
+
+      try {
+        const response = await fetch(sanitizePath, {
+          method: 'POST',
+          body: JSON.stringify({ journal: { notes: editorData } }),
```
Kharonus, line 86: 🟡 is the editor data HTML? do we need a value sanitation before sending it inside a request body?

app/controllers/work_packages/activities_tab_controller.rb
akabiru: Good Q- yes, in that case the @mention will only be removed if the journal note is ever updated. What we ~do~ will also have in https://github.com/opf/openproject/pull/18451 is an additional check at the time notifications are sent out that validates the user permissions before sending out any new notifications.

~Even then, when a user receives a notification, it does not contain any comment data- just a notification reason.~ _(incorrect, mentions include  the comment)_

FWIW, if the user had the permission before and received the email notification with the restricted comment- the comment will be accessible in their email albeit only showing data that they had access to _at the time_. Which works out IMO.

app/models/queries/principals/filters/restricted_mentionable_on_work_package_filter.rb
akabiru: Yes- it filters for principals that are project members AND have the `view_comments_with_restricted_visibility` permission. i.e. permission to view a comment with restricted visibility.

The feature specifies that you can only mention people that will be able to see the comment. Hence the need for this new filter.

frontend/src/stimulus/controllers/dynamic/work-packages/activities-tab/restricted-comment.controller.ts
akabiru: Editor data is rich text; plain text with markup. There should be no need for additional sanitization. Ref: https://github.com/opf/openproject/pull/18409/files#diff-6abe6a57158be8818c2ff8e2e400ae167471e3f48df0c8ecf0e7429a3b93d9f7

akabiru:
Gentle nudge @Kharonus to have another look :)

Kharonus (approved):
I can approve from my perspective.

Question is, is this enough :D As I have few insight in most of the previous discussions around restricted visibility for comments and the underlying permissions.

akabiru:
> I can approve from my perspective.
> 
> Question is, is this enough :D As I have few insight in most of the previous discussions around restricted visibility for comments and the underlying permissions.

Thanks! Still a useful perspective, @brunopagno  / @ulferts can top up the review before merge. There's no huge rush on it albeit going stale.

brunopagno (commented):
The code changes look good, nothing called my attention as potentially problematic. So all good there.

But when I was testing locally I was able to get an error like this (likely caused by the state of my local environment):

image

It happened because I got a POST 500 on `/work_packages/41/activities/sanitize_restricted_mentions`.

I think we need to make sure the errors are not exposed to the client.

akabiru:
> The code changes look good, nothing called my attention as potentially problematic. So all good there.
> 
> But when I was testing locally I was able to get an error like this (likely caused by the state of my local environment):
> 
> image
> 
> It happened because I got a POST 500 on `/work_packages/41/activities/sanitize_restricted_mentions`.
> 
> I think we need to make sure the errors are not exposed to the client.

Really nice catch! It would seem the [try, catch](https://github.com/opf/openproject/pull/18409/files#diff-41836885f3787a5d1043c50b89052dc04ad74bfcdc98e670c76467b5e9497915R94) does not work as intended- will add a test case for this 👍🏾

frontend/src/stimulus/controllers/dynamic/work-packages/activities-tab/restricted-comment.controller.ts
```diff
@@ -36,28 +36,68 @@ export default class RestrictedCommentController extends Controller {
...
+          headers: {
+            'X-CSRF-Token': csrfToken,
+            'Content-Type': 'application/json',
+          },
+        });
+
+        const sanitizedNotes = await response.text();
+        this.ckEditorInstance.setData(sanitizedNotes);
+      } catch (error) {
+        console.error(`Failed to sanitize restricted mentions: ${error}`);
```
brunopagno: @akabiru now that you pointed out this is indeed wrong. The `try catch` block will capture exceptions, but not responses with errors. I think we need to explicitly handle with a [response.ok](https://developer.mozilla.org/en-US/docs/Web/API/Response/ok) check here.

akabiru: Thanks @brunopagno - handled in https://github.com/opf/openproject/pull/18409/commits/22cc675c2491e9699fe416dc40429c33f277b3f9 via primer banner. I would have wished to display the error closer to the comment box- and perhaps even transition the ckeditor input into an error state, but it doesn't seem like we have this convention atm (with ckeditor input)- perhaps an enhancement to discuss with UX and apply in the future.

_The screenshot is from manual testing by raising an StandardError_

Screenshot 2025-04-04 at 4 33 36 PM

brunopagno (approved):
Looking great :tada:

PR #18419, opened 2025-03-24
Work Package Subject Patterns - Adds Missing Fields

mereghost:
#### Related WP: [OP#62429](https://community.openproject.org/work_packages/62429)

This PR adds the missing fields reported and support for multi-value fields

### DIFF
--- a/app/models/types/pattern_resolver.rb
+++ b/app/models/types/pattern_resolver.rb
@@ -56,6 +56,8 @@ module Types
       case value
       when Date, Time, DateTime
         value.strftime("%Y-%m-%d")
+      when Array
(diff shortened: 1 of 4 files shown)


Kharonus (approved):
LGTM.

Small remark: one part of the bug was to reorder the groups. Where should we put this logic? TokenMapper or frontend?

mereghost:
> LGTM.
> 
> Small remark: one part of the bug was to reorder the groups. Where should we put this logic? TokenMapper or frontend?

That's the reason I didn't do it. XD I couldn't make a decision as both have pros and cons. WDYT?

PR #18435, opened 2025-03-25
Implemetns the AMPF group checks

mereghost:
Related WP: [OP#62099](https://community.openproject.org/work_packages/62099)

Implements the AMPF checks finishing all the blocks.

### DIFF
--- a/modules/storages/app/common/storages/peripherals/connection_validators/check_result.rb
+++ b/modules/storages/app/common/storages/peripherals/connection_validators/check_result.rb
@@ -38,7 +38,7 @@ module Storages
         end
 
         def self.failure(key, message)
-          new(key:, state: :failure, message: message, timestamp: Time.zone.now)
+          new(key:, state: :failure, message:, timestamp: Time.zone.now)
         end
 
         def self.success(key)
@@ -46,7 +46,7 @@ module Storages
         end
 
         def self.warning(key, message)
-          new(key:, state: :warning, message: message, timestamp: Time.zone.now)
+          new(key:, state: :warning, message:, timestamp: Time.zone.now)
         end
 
         def success? = state == :success
--- a/modules/storages/app/common/storages/peripherals/connection_validators/nextcloud/ampf_connection_validator.rb
+++ b/modules/storages/app/common/storages/peripherals/connection_validators/nextcloud/ampf_connection_validator.rb
@@ -35,39 +35,32 @@ module Storages
         class AmpfConnectionValidator < BaseValidator
           using ServiceResultRefinements
 
-          def call
-            @results = {
-              userless_access_denied: CheckResult.skipped(:userless_access_denied),
-              group_folder_not_found: CheckResult.skipped(:group_folder_not_found),
-              files_request_failed_with_unknown_error: CheckResult.skipped(:files_request_failed_with_unknown_error),
-              with_unexpected_content: CheckResult.skipped(:with_unexpected_content)
-            }
-
-            catch :interrupted do
-              userless_access_denied
-              group_folder_not_found
-              files_request_failed_with_unknown_error
-              with_unexpected_content
-            end
-
-            @results
-          end
-
           private
+
+          def validate
+            register_checks(
+              :userless_access, :group_folder_presence, :files_request, :group_folder_contents
+            )
+
+            userless_access_denied
+            group_folder_not_found
+            files_request_failed_with_unknown_error
+            with_unexpected_content
+          end
 
           def userless_access_denied
             if files.result == :unauthorized
-              fail_check(__method__, message(:userless_access_denied))
+              fail_check(:userless_access, message(:userless_access_denied))
             else
-              pass_check(__method__)
+              pass_check(:userless_access)
             end
           end
 
           def group_folder_not_found
             if files.result == :not_found
-              fail_check(__method__, message(:group_folder_not_found))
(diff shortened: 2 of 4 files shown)


modules/storages/app/common/storages/peripherals/connection_validators/check_result.rb
```diff
@@ -38,15 +38,15 @@ def self.skipped(key)
         end
 
         def self.failure(key, message)
-          new(key:, state: :failure, message: message, timestamp: Time.zone.now)
+          new(key:, state: :failure, message:, timestamp: Time.zone.now)
```
NobodysNightmare, line 41: :question: Surprised that Rubocop didn't cover this previously.

Anyways, well done :)

modules/storages/app/common/storages/peripherals/connection_validators/nextcloud/ampf_connection_validator.rb
```diff
@@ -0,0 +1,112 @@
...
+              fail_check(:group_folder_presence, message(:group_folder_not_found))
+            else
+              pass_check(:group_folder_presence)
+            end
+          end
+
+          def files_request_failed_with_unknown_error
+            if files.result == :error
+              error "Connection validation failed with unknown error:\n\t" \
+                    "storage: ##{@storage.id} #{@storage.name}\n\t" \
```
NobodysNightmare: :green_circle: WDYT about moving the `\t` to the beginning of the following line? That would make it clearer which lines are indented.

modules/storages/app/common/storages/peripherals/connection_validators/nextcloud/ampf_connection_validator.rb
```diff
@@ -0,0 +1,112 @@
...
+              pass_check(:files_request)
+            end
+          end
+
+          def with_unexpected_content
+            unexpected_files = files.result.files.reject { managed_project_folder_ids.include?(it.id) }
+            return pass_check(:group_folder_contents) if unexpected_files.empty?
+
+            log_extraneous_files(unexpected_files)
+            warn_check(:group_folder_contents, message(:unexpected_content))
```
NobodysNightmare, line 86: :information_source:  We noticed recently, that this error message is not very helpful. E.g. on QA this message is still rendered, but we don't know why. Then if you don't have access to the server logs (e.g. SaaS) you are out of luck.

Feel free to consider this out of scope, but it would be helpful if we included details like the number of unexpected files and maybe the name of the first unexpected files in the error message, giving some actionability to the error message (while not risking to accidentially render 2000 file names in a single error message).

modules/storages/app/common/storages/peripherals/connection_validators/nextcloud/ampf_connection_validator.rb
```diff
@@ -0,0 +1,112 @@
...
+              "Name: #{file.name}, ID: #{file.id}, Location: #{file.location}"
+            end
+
+            warn "Unexpected files/folder found in group folder:\n\t#{file_representation.join("\n\t")}"
+          end
+
+          def auth_strategy = Registry["nextcloud.authentication.userless"].call
+
+          def managed_project_folder_ids
+            @managed_project_folder_ids ||= ProjectStorage.automatic.where(storage: @storage).pluck(:project_folder_id)
```
NobodysNightmare: :green_circle: This is used in a loop where we call `include?`. That's a typical use case where I'd recommend storing this collection as a set, which is optimized for inclusion queries (complexity of O(1) instead of O(n)):

```suggestion
            @managed_project_folder_ids ||= ProjectStorage.automatic.where(storage: @storage).pluck(:project_folder_id).to_set
```

NobodysNightmare (approved):

modules/storages/app/common/storages/peripherals/connection_validators/nextcloud/ampf_connection_validator.rb
mereghost: Oh, I just copied over the previous line. Good catch.

modules/storages/app/common/storages/peripherals/connection_validators/nextcloud/ampf_connection_validator.rb
mereghost: Let's see how the UI component ends up looking like. I wanted to add the list of "unexpected files" to the error message, but I fear it will be too much for the current UI to handle.

modules/storages/app/common/storages/peripherals/connection_validators/nextcloud/ampf_connection_validator.rb
mereghost: True that! Will do it.

modules/storages/app/common/storages/peripherals/connection_validators/nextcloud/ampf_connection_validator.rb
NobodysNightmare: > I fear it will be too much for the current UI to handle.

Me too, that's why I thought of the "sample"/"first" file that's unexpected. In the easy cases where one or two files are unexpected, this is pretty much listing all you need to list and it gives you a feeling for WTF is unexpected here.

As soon as it's more, you probably don't need to see them all, but hopefully the example will still help to understand the problem.

PR #18436, opened 2025-03-25
JWT: Allow authentication through client credentials

NobodysNightmare:
# Ticket
https://community.openproject.org/wp/62301

Adding a basic framework of entities to allow authenticating requests with a JWT that represents not a single user, but another client in the realm of the identity provider.

Introducing a new class called `ServiceAccount` that is a special kind (i.e. subclass) of user. Being a `User` means that we can still rely on our regular authentication mechanism, which APIs already rely on. For many existing APIs it's essential to have a `current_user`, e.g. to perform permission checks.

`ServiceAccounts` are linked to a service, which is the entity that _owns_ the service account. The first service intended to be built is the `ScimClient`, which will be added in a later PR.

Since the number of service accounts will usually be very low compared to the number of users in a system, I extracted the association between the account and the service into a separate table, so that normal users don't receive "useless" columns related to service associations.

# Usage

A service implementation essentially needs to add relations to the `ServiceAccountAssociation` and `ServiceAccount` model, e.g.

```ruby
  has_one :service_account_association, as: :service
  has_one :service_account, through: :service_account_association
```

In general the service is responsible for the lifecycle of the service account. I.e. creating and deleting it.

To be usable for the `JwtOidc` strategy, the service account needs to have a prefilled `identity_url`, allowing the strategy to recognize which service account is associated to a given client credentials token. E.g. in the case of Keycloak, each client has an associated service account (on Keycloak's side) whose UUID will be passed into the token.

To be usable for the client credentials flow of an `oauth_application` built into OpenProject, the `client_credentials_user_id` of the given `Doorkeeper::Application` has to be set to the id of the service account.

# Merge checklist

- [x] Added/updated tests
- [x] Think about and implement proper default for deletion of dependent records

### DIFF
--- a/app/models/service_account.rb
+++ b/app/models/service_account.rb
@@ -1,0 +1,55 @@
+# frozen_string_literal: true
+
+#-- copyright
+# OpenProject is an open source project management software.
+# Copyright (C) the OpenProject GmbH
+#
+# This program is free software; you can redistribute it and/or
+# modify it under the terms of the GNU General Public License version 3.
+#
+# OpenProject is a fork of ChiliProject, which is a fork of Redmine. The copyright follows:
+# Copyright (C) 2006-2013 Jean-Philippe Lang
+# Copyright (C) 2010-2013 the ChiliProject Team
+#
+# This program is free software; you can redistribute it and/or
+# modify it under the terms of the GNU General Public License
+# as published by the Free Software Foundation; either version 2
+# of the License, or (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program; if not, write to the Free Software
+# Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+#
+# See COPYRIGHT and LICENSE files for more details.
+#++
+
+class ServiceAccount < User
+  alias_attribute(:name, :lastname)
+  validates :name, presence: true
+  validates :name, length: { maximum: 256 }
+
+  has_one :service_account_association, dependent: :destroy
+
+  def to_s
+    name
+  end
+
+  def available_custom_fields = []
+
+  def logged? = false
+
+  def builtin? = true
+
+  def mail = ""
+
+  def time_zone
+    ActiveSupport::TimeZone[Setting.user_default_timezone.presence || "Etc/UTC"]
+  end
+
+  def rss_key = nil
+end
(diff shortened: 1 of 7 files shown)


spec/requests/api/v3/authentication_spec.rb
```diff
@@ -421,10 +423,11 @@ def set_basic_auth_header(user, password)
         .to_return(status: 200, body: JWT::JWK::Set.new(jwk_response).export.to_json, headers: {})
     end
     let(:jwk_response) { jwk }
+    let(:user) { create(:user, identity_url: "keycloak:#{token_sub}") }
 
     before do
       create(:oidc_provider, slug: "keycloak")
-      create(:user, identity_url: "keycloak:#{token_sub}")
+      user.save!
```
ba1ash, line 430: Would the following be sufficient?
```suggestion
      user
```

app/models/service_account.rb
```diff
@@ -0,0 +1,57 @@
...
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program; if not, write to the Free Software
+# Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+#
+# See COPYRIGHT and LICENSE files for more details.
+#++
+
+class ServiceAccount < User
```
ba1ash, line 31: Should it have been inherited from `Principal? Could you elaborate why you choose User? I am just being curious.

spec/requests/api/v3/authentication_spec.rb
NobodysNightmare: That's technically sufficient, though I usually find it more intention-revealing to call `save!`. It says "Hey I put that here to make sure we actually perform the side effect of saving the user".

I think in the OP codebase it's also more accepted to just use `let!`, maybe I should've done that, though my personal preference is the way I wrote it here.

WDYT? Should I switch for `let!`?

app/models/service_account.rb
NobodysNightmare: Principal was my first goal, because principal is "less" stuff.

In a nutshell the main conflicts were:

* We assign the thing (on the warden level) to an accessor called `current_user`, so even if it had worked, it would invite misconceptions and wrong ideas
* The concept of _permissions_ (which is relied upon heavily in APIv3) is added on the user-level, not on the Principal-level

In the end I made my peace with it as a subclass of user, because now it's on the same level (as in: depth of the inheritance tree) as the `SystemUser`, which seems to make sense to me.

spec/requests/api/v3/authentication_spec.rb
ba1ash: I'd prefer your way or just `user` statement rather than `let!`. So, it is fine by me like it is.

app/models/service_account.rb
ba1ash: I see the reasoning. Thanks for clarifying!

ba1ash (approved):
LGTM!

PR #18438, opened 2025-03-25
[#62486] searching in tokens removes prefix from word

Kharonus:
# Ticket
[OP#62486](https://community.openproject.org/wp/62486)
[OP#62255](https://community.openproject.org/wp/62255)

- searching inside tokens now strips the prefix
- after inserting a token, caret position is placed behind the token

- slice prefixes from input words, if word is read from any token
- decide where to put caret dependent on the token to put the caret behind


frontend/src/stimulus/controllers/pattern-input.controller.ts
```diff
@@ -499,7 +515,7 @@ export default class PatternInputController extends Controller {
             node.after(document.createTextNode(CONTROL_SPACE));
           }
 
-          if (this.isText(follower) && !this.isWhitespaceOrControlSpace(follower.wholeText[0])) {
+          if (this.isText(follower) && !this.isWhitespaceOrControlSpace(follower.wholeText[0] || '')) {
```
NobodysNightmare: :question: I am a bit surprised about this change, considering that I am looking at TypeScript.

Apparently `[0]` can in some cases (e.g. empty string) return `undefined`. However, `isWhitespaceOrControlSpace` only accepts arguments of type string. What's surprising to me is that this didn't result in a TS error before, because we are clearly passing `string|undefined` into a method that accepts `string`.

In general I am wondering whether we should solve it as done here (`|| ""`) or whether we should simply accept undefined/null as a parameter as well and explicitly return false in that case. Something similar was done to many of the `is*` methods in this controller.

NobodysNightmare (commented):
I have one question and will now try the changes practically out in a browser.

NobodysNightmare:
In Firefox I can't delete tokens anymore. Neither through backspace, nor through delete:

[Screencast from 2025-03-27 11-01-39.webm](https://github.com/user-attachments/assets/a861298a-9593-4f61-a8f3-3156ded8ad1d)

Mh, I just double checked to be sure and as far as I can tell, I already have this problem on dev :/

frontend/src/stimulus/controllers/pattern-input.controller.ts
Kharonus: yes, I agree. Better use another signature.

As for, why is it no TS error: It is simply wrongly typed. es5 interface for string is written like

`readonly [index: number]: string;` while is should be `string|undefined`.

will change the signature a bit

NobodysNightmare:
I tried understanding the "strip prefix" part and I think I found an off-by-one error. See me inserting a space in random places, just to see from which point the autocompletion starts:

[Screencast from 2025-03-27 11-16-45.webm](https://github.com/user-attachments/assets/d338f248-c5b8-4845-8e44-33b2fbac8289)

In the first word (project) it's fine, but afterwards the first character is cut off.

Kharonus:
Nice catch, the off by one was introduce by changing one side, but not the other :D -> fixed

As for FF: I can't help with it. For some reason, FF jumps iniine elements with content editable, when hitting backspace. This is what chromium does better. Both have their issues with `DEL`, which I would indeed label as a bug, but not a critical one.

NobodysNightmare:
Off-by-one error is gone. I now tried understanding the "after inserting a token, caret position is placed behind the token" requirement and played around a bit. In the screencast you see me:

1. Copying a token and subsequently typing
2. Copying the middle of a token and subsequently typing
3. Copying a token plus some pieces left and right and subsequently typing

Of these tests, 1 has non-ideal UX, but feels acceptable, though it might be exactly what you wanted to fix. 2 is just plain weird and I have no clue what's going on. 3 works as expected.

[Screencast from 2025-03-27 13-12-48.webm](https://github.com/user-attachments/assets/c2a5281d-3513-4063-bfea-3ce2bbe13302)

Kharonus:
@NobodysNightmare 

I think, I found the reason for backspace not working in firefox and fixed it. Also the partial copy of tokens should work now as expected (at least, as I would expect it :P)

NobodysNightmare (approved):
Behaviourally LGTM.

Something I realize is that it would probably be great to have nitty-gritty browser-tests for this. E.g. pressing five times backspace or delete.

I understand if you don't want to add them now, but later we'd be happy to have them.

PR #18441, opened 2025-03-25
[62510] Change text of child items in relations tab menu

cbliard:
# Ticket

https://community.openproject.org/wp/62510

In relations tab:
* "Existing child" did not make sense because it would become a child only after the action is performed. "Add existing work package as child" is a bit too long, so simply "Child" is enough, similar to other items in the menu.

* "New child" was also renamed "Create new child".

In children work packages list:
* Change "Add existing child" to "Add child".

## Screenshots

image

image

- Rename i18n key `work_package_relations_tab.relations.existing_child` to `work_package_relations_tab.relations.child`
- Not rename i18n key `work_package_relations_tab.relations.new_child` and simply update its text

# Merge checklist

- [x] Added/updated tests


mrmir (approved):
👍🏽

PR #18447, opened 2025-03-26
[#62107] Add SCIM server API

ba1ash:
# Ticket

https://community.openproject.org/work_packages/62107

I am trying to implement SCIM Server API for OpenProject using https://github.com/pond/scimitar.
It involves:
- Map `User` and `Group` data to SCIM schema.
- Add appropriate routes.
- Implement route handlers.

# Merge checklist

- [x] Added/updated tests


Gemfile
```diff
@@ -59,6 +59,8 @@ gem "will_paginate", "~> 4.0.0"
 
 gem "friendly_id", "~> 5.5.0"
 
+gem "scimitar"
```
NobodysNightmare: > Scimitar uses [semantic versioning](https://semver.org/) so you can be confident that patch and minor version updates for features, bug fixes and/or security patches will not break your application.

:yellow_circle: Subsequently we should pin to the major version that we currently tested against:

```suggestion
gem "scimitar", "~> 2.11"
```

app/contracts/users/delete_contract.rb
```diff
@@ -41,7 +43,8 @@ def self.deletion_allowed?(user, actor)
       if actor == user
         Setting.users_deletable_by_self?
       else
-        actor.admin? && actor.active? && Setting.users_deletable_by_admins?
+        (actor.admin? && actor.active? && Setting.users_deletable_by_admins?) ||
+          User.system == actor
```
NobodysNightmare: :exclamation: This is something that we have to see how we want to solve it. With #18468 I want to change it so that the corresponding actions are performed in the name of a service account, not the system user. So we will have to see how exactly we want to handle this.

I could even imagine that we want to respect this setting and not allow SCIM clients to delete users, if this setting is set. However, I guess in that case we'd have to hide locked users from SCIM clients?

**tl;dr:** I think this is acceptable for a demo implementation, but a big question mark for production-readiness.

app/controllers/scim_v2/groups_controller.rb
```diff
@@ -0,0 +1,107 @@
...
+            .new(group, current_user: User.system)
+            .call(ids: scim_resource.members.map(&:value), send_notifications: false)
+            .on_failure { |call| raise call.message }
+
+          group.to_scim(location: url_for(action: :show, id: group.id))
+        end
+      end
+    end
+
+    def replace
```
NobodysNightmare, line 73: :question: What's the (SCIM) semantic difference between `replace` and `update`?

:yellow_circle: Also since we seem to implement them the same (I couldn't spot a difference), should we make that more clear?

E.g.

```ruby
def replace
  update
end
```

edit: seeing that we do the routing very manually anyways, we could also solve this in routing probably:

```ruby
    put    "Users/:id", to: "users#update"
    patch  "Users/:id", to: "users#update"
```

app/controllers/scim_v2/groups_controller.rb
```diff
@@ -0,0 +1,107 @@
...
+
+    def update
+      super do |group_id, patch_hash|
+        storage_class.transaction do
+          group = storage_scope.find(group_id)
+          group.from_scim_patch!(patch_hash: patch_hash)
+          Groups::UpdateService
+            .new(user: User.system, model: group)
+            .call
+            .on_failure { |call| raise call.message }
```
NobodysNightmare, line 81: :yellow_circle: Should we render more errors in a controlled way instead of raising them as exceptions?

app/controllers/scim_v2/groups_controller.rb
```diff
@@ -0,0 +1,107 @@
...
+                parser.parse(params[:filter])
+                parser.to_activerecord_query(storage_scope)
+              end
+
+      pagination_info = scim_pagination_info(query.count)
+      page_of_results = query
+                          .order(id: :asc)
+                          .offset(pagination_info.offset)
+                          .limit(pagination_info.limit)
+                          .to_a
```
NobodysNightmare: :yellow_circle: The start of the method until here seems to be mostly boilerplate that's not specific to our groups. I would say (though I didn't diff it), that the code is exactly the same as in the `UsersController`.

IMO this speaks for trying to extract this behaviour into a concern or something similar that both controllers could share:

```ruby
# concern internally relying on storage_class and storage_scope being defined
pagination_info, page_of_results = filtered_scim_list

# explicitly passing them in
pagination_info, page_of_results = filtered_scim_list(storage_class, storage_scope)
```

app/controllers/scim_v2/users_controller.rb
```diff
@@ -0,0 +1,103 @@
...
+    end
+
+    protected
+
+    def storage_class
+      User
+    end
+
+    def storage_scope
+      User.user.where.not(identity_url: [nil, ""])
```
NobodysNightmare: I am pretty sure that this can be simplified to

```suggestion
      User.where.not(identity_url: [nil, ""])
```

For comparision:

```ruby
[2] pry(main)> User.all.to_sql
=> "SELECT \"users\".* FROM \"users\" WHERE \"users\".\"type\" = 'User'"
[3] pry(main)> User.all.user.to_sql
=> "SELECT \"users\".* FROM \"users\" WHERE \"users\".\"type\" = 'User'"
[4] pry(main)> Principal.all.user.to_sql
=> "SELECT \"users\".* FROM \"users\" WHERE \"users\".\"type\" = 'User'"
[5] pry(main)> Principal.all.to_sql
=> "SELECT \"users\".* FROM \"users\""
```

NobodysNightmare: An unrelated thought (GitHub please, give me multiple threads per line of code!):

:red_circle: I am not sure that we should completely hide users from view that don't have an identity url. I agree that we can't match them through their external id, but now the SCIM client also can't match them through other factors such as an email address. Thus conflicts between existing "local" users and newly created "SCIM" users are unavoidable (e.g. duplicate email addresses in our validations).

That said, I am not sure whether any SCIM client is attempting to avoid such conflicts, but I would definitely check on the scim documentation on whether they say anything about returning users without an external id (e.g. whether we are _required_ to hide those users).

:red_circle: On the other hand we should probably try to filter out `builtin?` users. Though as far as I can tell, they are not database filterable... so we will have to come up with something there.

app/models/group.rb
```diff
@@ -67,6 +69,80 @@ def to_s
...
+    self.identity_url = "#{oidc_provider.slug}:#{external_id}"
+    external_id
+  end
+
+  def scim_users_and_groups
+    users
+  end
+
+  def scim_users_and_groups=(_array)
+    users
```
NobodysNightmare: :red_circle: Is that a pending TODO? We are not assigning anything here.

app/models/group.rb
```diff
@@ -67,6 +69,80 @@ def to_s
...
+          else
+            raise Scimitar::InvalidSyntaxError.new("Unrecognised type #{type.inspect}")
+          end
+        }
+      ]
+    }
+  end
+
+  def self.scim_mutable_attributes
+    nil
```
NobodysNightmare: :question: Nothing mutable in a group?

app/models/user.rb
```diff
@@ -610,7 +700,7 @@ def former_passwords_include?(password)
   def clean_up_former_passwords
     # minimum 1 to keep the actual user password
     keep_count = [1, Setting[:password_count_former_banned].to_i].max
-    (passwords[keep_count..-1] || []).each(&:destroy)
+    (passwords[keep_count..] || []).each(&:destroy)
```
NobodysNightmare, line 691: :green_circle: Probably that hint comes too late now, but this file contains a bunch of rubocop-induced code changes that would have probably belonged into a separate commit (maybe even PR, but with a separate commit that would've been a decision for later :) )

Anyways:

:heart: for tackling all of those rubocop offenses :)

config/routes.rb
```diff
@@ -868,6 +868,24 @@
     get "ensure_connection", controller: "oauth_clients", action: :ensure_connection, as: "oauth_clients_ensure_connection"
   end
 
+  namespace :scim_v2 do
```
NobodysNightmare, line 919: :yellow_circle: I realized that I just kinda accepted this whole namespace and naming of the scope so far, but now that I noticed I have to ask: Why do we call everything SCIM **v2**?

It's our first implementation of SCIM and I can't find a lot of official reference of the standard being called "version 2" in any way (some of the URNs contain an innocent `2.0` inside of them). That's setting it apart from OAuth 2.0, where the version number is even part of the name.

**tl;dr:** Is anything speaking against using `scim` as a name for our scopes and also as a namespace for our APIs?

spec/requests/scim_v2/groups_spec.rb
```diff
@@ -0,0 +1,303 @@
...
+                                                        "schemas" => ["urn:ietf:params:scim:schemas:core:2.0:Group"] }],
+                                      "itemsPerPage" => 100,
+                                      "schemas" => ["urn:ietf:params:scim:api:messages:2.0:ListResponse"],
+                                      "startIndex" => 1,
+                                      "totalResults" => 1 })
+      end
+
+      # it do
+      #   filter = ERB::Util.url_encode('displayName Eq "john"')
+      #   get "/scim_v2/Groups?filter=#{filter}", {}, headers
```
NobodysNightmare: :yellow_circle: Should this commented test be reactivated or removed?

NobodysNightmare (changes requested):
Something that we should make sure, since a lot of the implementation is an obvious proof of concept (e.g. lack of authentication): Let's make sure we have trackable tickets for all changes that will have to be done afterwards.

What I can see immediately:
* authenticate requests
* replace usage of system user
* choose the client's auth provider, not the first one (P.S.: I'd not limit it to OIDC)
* figure out what to do with global setting for delete prevention

In general also most of my code comments below could as well be turned into a ticket and be considered out of scope for this proof-of-concept. It's just important that we don't forget :)

app/contracts/users/delete_contract.rb
ba1ash: I will extract it as a follow-up ticket. And for now leave it as it is now.
There is a separate thing in SCIM for users being active/inactive. In this implementation there is the following mapping:
`scim_user.active` means `openproject_user.status == 'active'`
`scim_user.inactive` means `openproject_user.status == 'locked'`

app/controllers/scim_v2/groups_controller.rb
ba1ash: There is a difference. `replace` expects the whole object as `urn:ietf:params:scim:schemas:core:2.0:User` or `urn:ietf:params:scim:schemas:core:2.0:Group` respectively. `update` expects `urn:ietf:params:scim:api:messages:2.0:PatchOp`.
And controller action differs basically by this calls:
```ruby
         # update action
         # ....
          group.from_scim_patch!(patch_hash: patch_hash)
```
and
```ruby
         # replace action
         # ....
          group.from_scim!(scim_hash: scim_resource.as_json)
```

app/controllers/scim_v2/groups_controller.rb
ba1ash: Yes, we should. I'll do that.

app/controllers/scim_v2/groups_controller.rb
ba1ash: Yes, you're right. I postponed it till I'm sure they're the same. I'll do that.

app/models/group.rb
ba1ash: It means every mapped SCIM attribute that has write accessor will be assignable by create or update operations.
Source: https://www.rubydoc.info/github/pond/scimitar/Scimitar/Resources/Mixin search for `scim_mutable_attributes`

config/routes.rb
ba1ash: I believe it is due to protocol version we implement. And it is SCIM version 2.

spec/requests/scim_v2/groups_spec.rb
ba1ash: Yes, it should be reacrtivated.

app/controllers/scim_v2/users_controller.rb
ba1ash: > 🔴 I am not sure that we should completely hide users from view that don't have an identity url. I agree that we can't match them through their external id, but now the SCIM client also can't match them through other factors such as an email address. Thus conflicts between existing "local" users and newly created "SCIM" users are unavoidable (e.g. duplicate email addresses in our validations).
> 
> That said, I am not sure whether any SCIM client is attempting to avoid such conflicts, but I would definitely check on the scim documentation on whether they say anything about returning users without an external id (e.g. whether we are _required_ to hide those users).

Accroding to specs(https://www.rfc-editor.org/rfc/rfc7643.html#section-3.1) `externalId` is optional. So, the scope should not have **identity_url** constraint.

> 🔴 On the other hand we should probably try to filter out `builtin?` users. Though as far as I can tell, they are not database filterable... so we will have to come up with something there.

There is `not_builtin` scope. I will use it.

app/models/group.rb
NobodysNightmare: Understood, sounds good.

:green_circle: Maybe we want to document this for the purpose of avoiding confusion. E.g.

```ruby
# Allow mutation of everything with a write accessor
nil
```

app/models/group.rb
ba1ash: That is intentional. I added an explanational  comment. In short, group member modifications are handled by the service call in the controller.

app/models/user.rb
ba1ash: I used `rubocop -a` to format this file and decided to squeeze additional fixes in. True, they are irrelevant to the PR. Later I will separate them at least by having a dedicated commit. Accept my apologies for overwhelming :pray:

app/controllers/scim_v2/groups_controller.rb
ba1ash: Not done. Extracted to a follow-up feature.

modules/bim/lib/open_project/bim/engine.rb
```diff
@@ -214,10 +216,9 @@ class Engine < ::Rails::Engine
     end
 
     config.to_prepare do
-      Doorkeeper.configuration.scopes.add(:bcf_v2_1)
-
       unless defined? OpenProject::Authentication::Scope::BCF_V2_1
         OpenProject::Authentication::Scope::BCF_V2_1 = :bcf_v2_1
+        Doorkeeper.configuration.scopes.add(OpenProject::Authentication::Scope::BCF_V2_1)
```
NobodysNightmare: :yellow_circle: In which context did this change occur? I can't see a separate commit for it and I am wondering whether this is intended as a cleanup or whether something didn't work as intended.

I would have expected that adding the scope to the doorkeeper configuration was required on each reload, therefore it was included in the front of `config.to_prepare`.

On the other hand constants remain defined even when rails reloads its config, so I'd expect that the new code now effectively only adds the scopes once to doorkeeper, not on each reload.

Is this intentional?

ba1ash: As was discussed: I will extract this scope manipulations to a separate PR.

app/controllers/scim_v2/base_controller_actions.rb
```diff
@@ -0,0 +1,71 @@
...
+#
+# See COPYRIGHT and LICENSE files for more details.
+#++
+
+module ScimV2
+  module BaseControllerActions
+    extend ActiveSupport::Concern
+
+    included do
+      skip_before_action :verify_authenticity_token
```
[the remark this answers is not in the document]
ba1ash: The change has been done consciously.

.rubocop.yml
```diff
@@ -175,8 +175,7 @@ RSpec/IndexedLet:
 
 # The http verbs in Rack::Test do not accept named parameters (params: params)
 Rails/HttpPositionalArguments:
-  Exclude:
-    - 'spec/requests/api/v3/**/*.rb'
+  Enabled: false
```
NobodysNightmare, line 178: :yellow_circle:  I am currently assuming in good faith that this cop was useful outside of `spec/requests/api/v3` before this PR. Which would mean that it's only not working correctly in the APIv3 specs so far, and probably since this PR not working correctly in SCIM specs.

Should we just add a new path for our SCIM specs here instead of broadly disabling the cop?

app/contracts/groups/base_contract.rb
```diff
@@ -35,6 +37,7 @@ class BaseContract < ::ModelContract
     # hence we need to put "lastname" as an attribute here
     attribute :name
     attribute :lastname
+    attribute :identity_url
```
NobodysNightmare: :yellow_circle: Is that still necessary now that the identity url was deprecated in a previous PR? (I remember that our initial plan was to use the identity_url for both users and groups)

app/contracts/users/delete_contract.rb
NobodysNightmare: :red_circle: Considering our recent discussion that we do not want to give extra delete permissions to SCIM clients (i.e. still enforce the `users_deleteable_by_admins?` setting), shouldn't we remove this change here and keep the line as-is?

SCIM client service accounts have admin privileges.

app/controllers/scim_v2/scim_controller_mixins.rb
```diff
@@ -0,0 +1,53 @@
...
+    module Overwrites
+      # Completely overwriting authenticate method of Scimitar
+      def authenticate
+        return handle_scim_error(Scimitar::AuthenticationError.new) unless OpenProject::FeatureDecisions.scim_api_active?
+
+        warden = request.env["warden"]
+        User.current = warden.authenticate! scope: :scim_v2
+
+        # Only a ServiceAccount associated with a ScimClient can use SCIM Server API
+
```
NobodysNightmare: :green_circle: 

```suggestion
```

app/controllers/scim_v2/scim_controller_mixins.rb
```diff
@@ -0,0 +1,53 @@
...
+        unless User.current.is_a?(ServiceAccount) && User.current.service_account_association.service.is_a?(ScimClient)
```
NobodysNightmare: :green_circle: I am wondering whether we should "Rubify" this by only listening for the quack sounds of the service account here:

```suggestion
        unless User.current.respond_to?(:service_account_association) && User.current.service_account_association.service.is_a?(ScimClient)
```

If we added a `service` to the service account that delegates the call to `service_account_association.service`, we could even make this line more readable:

```suggestion
        unless User.current.respond_to?(:service) && User.current.service.is_a?(ScimClient)
```

app/controllers/scim_v2/users_controller.rb
```diff
@@ -0,0 +1,136 @@
...
+  class UsersController < Scimitar::ResourcesController
+    include BaseControllerActions
+
+    def create
+      super do |scim_resource|
+        storage_class.transaction do
+          user = storage_class.new
+          user.from_scim!(scim_hash: scim_resource.as_json)
+          user.firstname = "123" if user.firstname.blank?
+          user.lastname = "456" if user.lastname.blank?
```
NobodysNightmare: :red_circle: Those "123" and "456" look like test code to me. Do we want to remove it?

As for the correct behaviour for when those required values are missing: I don't know. We could return an error response to the SCIM client, if we don't allow these to be empty, I guess?

app/controllers/scim_v2/users_controller.rb
```diff
@@ -0,0 +1,136 @@
...
+          call = Users::CreateService
+                   .new(user: User.current, model: user)
+                   .call(user.attributes)
+                   .on_failure do |result|
+            uniqueness_error = result.errors.find { |e| e.type == :taken }
+            if uniqueness_error.present?
+              raise Scimitar::ErrorResponse.new(
```
NobodysNightmare, line 46: :yellow_circle: I saw the same code in the groups controller as well. Should we provide a common helper module for this? Or does that not make sense yet?

e.g.

```
def raise_result_errors_for_scim(result)
  result.on_failure do |result|
    # same as here... raise specific thing on uniqueness error, or raise something else
  end
end
```

app/controllers/scim_v2/users_controller.rb
```diff
@@ -0,0 +1,136 @@
...
+          if unauthorized_error.present?
+            raise Scimitar::ErrorResponse.new(
+              status: 403,
+              detail: "User can't be deleted due to permission absence."
+            )
+          else
+            raise result.message
+          end
+
+          raise call.message
```
NobodysNightmare: :red_circle: What rubocop said: This `raise` is unreachable.

I am also surprised about the formatting here. In GitHub it looks like `.on_failure do` is indented the same depth as the first line of the block below, where I would have expected the block to be indented further.

app/models/group.rb
```diff
@@ -67,6 +69,59 @@ def to_s
     lastname
   end
 
+  def scim_members
+    @scim_members ||= users
+  end
+
+  def scim_members=(array)
```
NobodysNightmare, line 76: :yellow_circle: I can't find any calls to `scim_members=` in the diff of this PR. Is this still used, maybe from a `Scimitar` method? Or does it stem from an intermediate implementation state?

I'd suggest to delete both of these methods if this writer method is unused or extend the comment to indicate _where_ this writer method is used, if it's purpose-built for an internal scimitar API.

app/models/group.rb
```diff
@@ -67,6 +69,59 @@ def to_s
...
+  end
+
+  def self.scim_queryable_attributes
+    {
+      displayName: { column: :lastname },
+      externalId: { column: UserAuthProviderLink.arel_table[:external_id] }
+    }
+  end
+
+  include Scimitar::Resources::Mixin
```
NobodysNightmare, line 123: It's surprising to me to find an `include` below all public method definitions. Is there a technical reason to do it this low or should we move the include to the top of the class definition?

app/models/principal.rb
```diff
@@ -192,6 +210,47 @@ def <=>(other)
...
+    if is_active
+      activate
+      true
+    else
+      lock if active?
+      false
+    end
+  end
+
+  def scim_active
```
NobodysNightmare: :green_circle: I guess also for these methods it would make sense to add a small intro-comment how/where they are used?

Also here it's not immediately obvious to me, how they will be called.

app/models/principal.rb
```diff
@@ -192,6 +210,47 @@ def <=>(other)
...
+    active_user_auth_provider_link&.external_id
+  end
+
+  def scim_external_id=(external_id)
+    oidc_provider = User.current.service_account_association.service.auth_provider
+
+    "::#{self.class}s::SetAttributesService"
+      .constantize
+      .new(user: User.current, model: self, contract_class: EmptyContract)
+      .call(identity_url: "#{oidc_provider.slug}:#{external_id}")
```
NobodysNightmare, line 226: :red_circle: It is surprising to me that we handle the setting of the `external_id` through the deprecated `identity_url` attribute via a `SetAttributesService` to be honest.

Seems like we are relying on code that's intended to ensure backwards compatibility here in newly written code.

A general remark for `scim_external_id=`: This is a method on `Principal` instances, but it seems to be specifically designed to be called in a specific manner during the SCIM process. For example we implicitly expect here that `User.current` is the service account of a SCIM client. At the same time this method does much more than a simple attribute assignment. Is there another way for us to hook into this? E.g. by overwriting the `from_scim!` method and calling a service there or something like that?

Maybe it's also fine, because all of these methods are prefixed with `scim_`, so maybe it's clear that they expose a "private API"...

app/models/principal.rb
```diff
@@ -192,6 +210,47 @@ def <=>(other)
...
+      .on_failure { |result| raise result.to_s }
+    external_id
+  end
+
+  def self.scim_mutable_attributes
```
NobodysNightmare: :green_circle: Just below these two `self.` method definitions there is a `class << self` block, which also has the effect of defining methods on the class side.

Should we move those two methods inside that block?

app/services/authentication/omniauth_service.rb
```diff
@@ -168,7 +168,7 @@ def find_existing_user
       if developer_provider?
         User.find_by(mail: auth_hash[:uid])
       else
-        UserAuthProviderLink.with_identity_url(user_attributes[:identity_url]).first&.user
+        UserAuthProviderLink.with_identity_url(user_attributes[:identity_url]).first&.principal
```
NobodysNightmare: :yellow_circle: This piece of the code is not prepared for or intended to handle authentication of non-users. I.e. it would be an error if a group tried to authenticate somewhere, because groups are not allowed to login anywhere.

Shouldn't we make sure that the principal we found is indeed a user and raise an error otherwise?

app/services/base_services/create.rb
```diff
@@ -28,14 +30,19 @@
 
 module BaseServices
   class Create < Write
+    def initialize(user:, model: nil, contract_class: nil, contract_options: {})
+      self.model = model
```
NobodysNightmare, line 34: :yellow_circle: Mh. At first I wanted to ask, why you didn't "just" override the `CreateService` of user and group to have the updated initializer and `instance` implementation.

But after looking into it, I start to smell an inconsistency in our base service classes again, because:
* `model` is added as an attribute accessor to `BaseContracted`, but there is no way to define it through the constructor
* two other prominent subclasses **do** define a constructor that includes `model`

So I guess your change here is generally fine, but I start to feel tension between different methods such as `model` and `instance` and the question whether they mean the same thing or something different. I feel the urge that this needs a cleanup, but maybe it's not in the scope of this long-lived PR to do that cleanup.

app/services/user_auth_provider_links_setter.rb
```diff
@@ -0,0 +1,54 @@
...
+          link = model.user_auth_provider_links
+                   .find_or_initialize_by(auth_provider_id:)
+          link.assign_attributes(external_id:, principal: model)
+          if link.changed? && link.persisted?
+            link.save!
+            model.user_auth_provider_links.reload
+            model.user_auth_provider_links.find { |l| l.id == link.id }.external_id_will_change!
+          end
+        else
+          raise ActiveRecord::RecordNotFound, "AuthProvider with slug: \"#{slug}\" has been not found"
```
NobodysNightmare: :yellow_circle: 

```suggestion
          raise ActiveRecord::RecordNotFound, "AuthProvider with slug: \"#{slug}\" has not been found"
```

config/initializers/feature_decisions.rb
```diff
@@ -47,5 +47,8 @@
 OpenProject::FeatureDecisions.add :calculated_value_project_attribute,
                                   description: "Allows the use of calculated values as a project attribute."
 
+OpenProject::FeatureDecisions.add :scim_api,
```
NobodysNightmare, line 50: :exclamation: Please check this again after a rebase. IIRC I added the feature decision _below_ the `block_note_editor`, which means there will not be a git conflict, but our changes effectively do conflict.

config/initializers/scimitar.rb
```diff
@@ -0,0 +1,56 @@
...
+    patch: Scimitar::Supportable.supported,
+    authenticationSchemes: [Scimitar::AuthenticationScheme.bearer]
+  )
+  Scimitar.engine_configuration = Scimitar::EngineConfiguration.new(
+    application_controller_mixin: ScimV2::ScimControllerMixins
+  )
+
+  module ScimitarSchemaExtension
+    def scim_attributes
+      super + [Scimitar::Schema::Attribute.new(name: "externalId",
```
NobodysNightmare: :question: I am surprised: `externalId` is not an attribute that scimitar supports out of the box?

config/initializers/warden.rb
```diff
@@ -26,6 +26,10 @@
      anonymous_fallback]
 end
 
+OpenProject::Authentication.update_strategies(OpenProject::Authentication::Scope::SCIM_V2, { store: false }) do |_|
+  %i[oauth jwt_oidc session]
```
NobodysNightmare: Actually do we need `session` auth for SCIM?

docker/dev/backend/Dockerfile
```diff
@@ -37,6 +37,9 @@ RUN echo "deb [signed-by=/etc/apt/keyrings/nodesource.gpg] https://deb.nodesourc
 RUN apt-get update
 RUN apt-get install -y nodejs
 
+# Setup java. Needed for https://github.com/NUARIG/ladle which is used in specs to test LDAP related behavior.
+RUN apt-get install -y openjdk-17-jdk
```
NobodysNightmare, line 41: :question: For me to understand: That's not a new dependency you introduced in this PR, but rather one that you discovered while running tests that no one else has run before in docker?

lib_static/open_project/authentication/strategies/warden/doorkeeper_oauth.rb
```diff
@@ -42,7 +42,7 @@ def access_token_invalid?(access_token)
           end
 
           def authenticate_user(id)
-            user = id && User.active.find_by(id:)
+            user = id && User.active.where(id:).or(ServiceAccount.where(id:)).first
```
NobodysNightmare: :yellow_circle: I think that should not be necessary after our fixes to lazy loading anymore, right?

A service account **is** a user and thus `User.active.where(:id)` should find it.

As a side note, if I read this code here correctly, it would not filter service accounts by the `active` scope, which would be a bug as far as I can tell.

lib_static/open_project/authentication/strategies/warden/jwt_oidc.rb
```diff
@@ -30,7 +30,7 @@ def authenticate!
                                            error_description: "Requires scope #{scope} to access this resource."
                 end
 
-                user = provider.user_auth_provider_links.find_by(external_id: payload["sub"])&.user
+                user = provider.user_auth_provider_links.find_by(external_id: payload["sub"])&.principal
```
NobodysNightmare: :yellow_circle: Similar to the comment in `OmniauthService`: This piece of code is not expected to authenticate groups ever. Groups are not expected/allowed to authenticate in open project, so I think we should make sure that they can't.

spec/factories/user_factory.rb
```diff
@@ -31,7 +31,7 @@
 FactoryBot.define do
   factory :user, parent: :principal, class: "User" do
     firstname { "Bob" }
-    lastname { "Bobbit" }
+    sequence(:lastname) { |n| "Bobbit#{n}" }
```
NobodysNightmare, line 34: :question: Do we have a uniqueness constraint on the last name of users or why was that change necessary?

NobodysNightmare (changes requested):
I left some comments for smaller and bigger issues. Let's discuss together how we want to move forward, maybe some of them are acceptable to be addressed at a later point.

Also note that there are still a lot of legitimately looking rubocop offenses in the code. I assume that you still want to address those and I would have second look at them later.

.rubocop.yml
ba1ash: I've been annoyed by this one too many times, so I finally decided to disable it completely.
Personally, I don't see much value in this cop (let me know if you see it useful pls).
Sure, we could add request specs here every time there's a new one, but honestly, I think it's much simpler to just disable it and not have to deal with it at all. Would you agree to that?

app/contracts/users/delete_contract.rb
ba1ash: You're absolutely right. It will be deleted.

app/contracts/groups/base_contract.rb
ba1ash: True, it is not necessary anymore.

app/controllers/scim_v2/users_controller.rb
ba1ash: It is an ugly way of dealing with a special case when SCIM Client does not send `firstname` and `lastname` to OpenProject. But for OP it is required to have these. Let me try to find more graceful placeholders for this scenario.

app/models/group.rb
ba1ash: It is used by `scimitar` itself. It is specified [here](https://github.com/opf/openproject/blob/feature/62107-add-scim-server-api/app/models/group.rb#L93). Then `scimitar` expects both getter and setter to be available.

app/models/group.rb
ba1ash: Yes, it looks a bit unusual. The explanation: https://github.com/pond/scimitar/blob/main/app/models/scimitar/resources/mixin.rb#L9
https://github.com/pond/scimitar/blob/main/app/models/scimitar/resources/mixin.rb#L290

app/models/principal.rb
ba1ash: Yes, it is the same as above with `scim_members` `scim_members=`. Let me check how to clarify this in the code.

app/services/authentication/omniauth_service.rb
ba1ash: I suggest to limit search to principals with type User. Wdyt?

config/initializers/scimitar.rb
ba1ash: It is supported, but not included in the schema returned by `/Schemas` endpoint.

config/initializers/warden.rb
ba1ash: I do not think that we need it. I will remove it.

docker/dev/backend/Dockerfile
ba1ash: Yes, it should have been a separate PR. I tried to run LDAP related specs (e.g. `spec/services/ldap/import_user_list_service_integration_spec.rb`) in docker and the absence of jdk was a bummer.

lib_static/open_project/authentication/strategies/warden/doorkeeper_oauth.rb
ba1ash: Not exactly. `User.active` does not return `builtin` entities. And `ServiceAccount` is a `builtin` one. 
The following should be correct:
```ruby
user = id && User.active.where(id:).or(ServiceAccount.where(id:, status: ServiceAccount.statuses[:active])).first
```
Or we could consider exclusion of `ServiceAccount` from `builtin` list. And then the following should be just fine.
```ruby
User.active.where(:id)
```
But I remember there were other consequences in making ServiceAccount `not_builtin`. Let me check.

spec/factories/user_factory.rb
ba1ash: We do not have a uniqueness constraint. To be honest I do not remember the reason quite well. 
I remember that at some point I had some extra users created in specs. And somehow having this sequence was my way to differentiate them. But looking at the code now I do not see real need in having this sequence. Would you like me to to try revert it?

app/controllers/scim_v2/users_controller.rb
ba1ash: Yes, that makes sense. I suggest to do it as a follow-up improvement.

.rubocop.yml
NobodysNightmare: I guess I'd usually take this at least as a FYI to one of the developer rounds, so that people can express opinions.

What surprises me is that this cop has no huge configuration block yet, i.e. there is exactly one like for all the APIv3 specs. Thus I assumed that for the APIv3 tests we use `Rack::Test` and therefore can't follow the guideline offered by the cop, but that in all other cases the cop would rightfully complain.

If this was something that we'd have to regularly repeat doing because of false positives, I'd agree that the cop should rather be disabled.

app/controllers/scim_v2/users_controller.rb
NobodysNightmare: Correct me if I am wrong, but isn't this a "normal" error case?

I'd expect that we indicate first and last name as "required" in our schema. Thus a client that's not sending this information is sending something that's invalid according to our schema, so we can respond with HTTP 400 and indicate the `invalidValue` error.

See https://www.rfc-editor.org/rfc/rfc7644#section-3.12

Am I missing something in that regard?

app/models/group.rb
NobodysNightmare: Okay. After thinking about it, I think it's what I wrote elsewhere: We can probably expect these to be "magically" used by scimitar, because you also prefixed them with `scim_`, which makes sense.

app/services/authentication/omniauth_service.rb
NobodysNightmare: I think we should not limit this during search, i.e. not do `where(type: "User")`, because that would most likely not take inheritance into account and thus refuse to login a `ServiceAccount`, which is a User and totally allowed to log in.

I guess I'd do a check on the returned object before returning it, i.e. when it's already Ruby.

lib_static/open_project/authentication/strategies/warden/doorkeeper_oauth.rb
NobodysNightmare: For some reason I have an extra hard time finding the implementation of `User.active`, but obviously this is surprising behaviour. I would not have expected that it makes a difference whether I do `User.active` or `User.all.select(&:active?)`, except for the former to be more efficient.

Meh, that's a bummer...

spec/factories/user_factory.rb
NobodysNightmare: > Would you like me to to try revert it?

I was just surprised about it, I am not sure if I have strong enough opinions on this, so probably not unless it is easy.

spec/requests/scim_v2/schemas_spec.rb
```diff
@@ -0,0 +1,79 @@
...
+  let(:headers) { { "CONTENT_TYPE" => "application/scim+json", "HTTP_AUTHORIZATION" => "Bearer #{token.plaintext_token}" } }
+  let(:token) { create(:oauth_access_token, resource_owner: service_account, scopes: ["scim_v2"]) }
+  let(:service_account) { create(:service_account, service: scim_client) }
+  let(:scim_client) { create(:scim_client, authentication_method: :oauth2_token, auth_provider_id: oidc_provider.id) }
+
+  before { token }
+
+  describe "GET /scim_v2/Schemas" do
+    context "with the feature flag enabled", with_flag: { scim_api: true } do
+      it "responds with supported schemas " do
```
NobodysNightmare:
```suggestion
      it "responds with supported schemas" do
```

NobodysNightmare (approved):

NobodysNightmare (changes requested):
Specs seem to be failing around the SCIM client UI. That's probably a real issue and not a flicker:

> Could not find the source association(s) "user" or :users in model UserAuthProviderLink. Try 'has_many :users, :through => :user_auth_provider_links, :source => <name>'. Is it one of principal or auth_provider?

NobodysNightmare (approved):

app/services/base_services/create.rb
ba1ash: As discussed I will create a code maintenance follow up ticket to deal with the above issue.

app/models/principal.rb
ba1ash: We discussed this topic and agreed that it is fine at the moment. I will make sure it is not forgotten when we get rid of `users.identity_url` in API v3.

app/controllers/scim_v2/users_controller.rb
ba1ash: Yes, your suggestion is absolutely valid and right. Let's go this way. Will be done in a follow-up.

PR #18459, opened 2025-03-27
[#62555] PDF: Insufficient space between table header and the border

as-op:
# Ticket

https://community.openproject.org/work_packages/62555

Add spacing between cell border and text content

Add padding to the cell


as-op:
Merged, there is nothing really to do for a code review.

PR #18464, opened 2025-03-27
[#62556] Long text CF not displayed when empty

as-op:
# Ticket

https://community.openproject.org/work_packages/62556

Display the label of a long text custom field even when the value is empty

Remove the check for emptiness

# Merge checklist

- [x] Added/updated tests


ulferts (approved):
👍

PR #18468, opened 2025-03-27
Add ScimClient model

NobodysNightmare:
Including basic services to manipulate them. The
service account is managed as part of the ScimClient's lifecycle.

We make sure to revoke non-chosen authentication methods with every update, so that it's not possible to keep authenticating through an outdated (and most likely UI-invisible) method anymore.

# Ticket
https://community.openproject.org/work_packages/62516/activity

To process SCIM requests we need to authenticate an entity during the corresponding API requests. This must be possible via three high-level methods:

* JWT from an identity provider
* access token issued by OpenProject `Doorkeeper::Application`
* static token issued "by OpenProject"

# Which approach did you choose and why?

The ScimClient is the central entity to configure settings around a configured SCIM client. It's related to a user (new subtype `ServiceAccount`) and optionally an oauth application. The service account is the authenticated entity. In this model we can cover all three authentication modes:

* JWT is natively supported by adding an `identity_url` to the service account
*  access token and static token can both be supported through a `Doorkeeper::Application` that uses the service account as a client credentials user
    * static token is a special case of this config, where we directly offer a token created for this oauth app to the user, without really performing an oauth token request

# Merge checklist

- [ ] Added/updated tests

### DIFF
--- a/app/services/scim_clients/create_service.rb
+++ b/app/services/scim_clients/create_service.rb
@@ -1,0 +1,78 @@
+# frozen_string_literal: true
+
+#-- copyright
+# OpenProject is an open source project management software.
+# Copyright (C) the OpenProject GmbH
+#
+# This program is free software; you can redistribute it and/or
+# modify it under the terms of the GNU General Public License version 3.
+#
+# OpenProject is a fork of ChiliProject, which is a fork of Redmine. The copyright follows:
+# Copyright (C) 2006-2013 Jean-Philippe Lang
+# Copyright (C) 2010-2013 the ChiliProject Team
+#
+# This program is free software; you can redistribute it and/or
+# modify it under the terms of the GNU General Public License
+# as published by the Free Software Foundation; either version 2
+# of the License, or (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program; if not, write to the Free Software
+# Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+#
+# See COPYRIGHT and LICENSE files for more details.
+#++
+
+class ScimClients::CreateService < BaseServices::Create
+  def after_perform(_)
+    super.tap do |service_result|
+      self.model = service_result.result
+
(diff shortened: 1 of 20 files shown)


app/services/scim_clients/create_service.rb
```diff
@@ -0,0 +1,78 @@
...
+      service_account.user_auth_provider_links.build(
+        auth_provider_id: params[:auth_provider_id],
+        external_id: params[:jwt_sub]
+      )
+    end
+    service_account.save!
+  end
+
+  def update_oauth_application(service_result)
+    return if !model.authentication_method_oauth2_client? && !model.authentication_method_oauth2_token?
```
ba1ash, line 55: Would the following code be a bit simpler? Of course, if we add new not oauth2 related authentication method later then it will be a bug here...
``` ruby
return if model.authentication_method_sso?
```

ba1ash (approved):
LGTM

db/migrate/20250605133700_create_scim_clients.rb
```diff
@@ -0,0 +1,41 @@
...
+# Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+#
+# See COPYRIGHT and LICENSE files for more details.
+#++
+
+class CreateScimClients < ActiveRecord::Migration[8.0]
+  def change
+    create_table :scim_clients do |t|
+      t.string :name, null: false
+      t.belongs_to :auth_provider, null: false
```
ba1ash: I suggest adding db constraint as well. WDYT?
```suggestion
      t.belongs_to :auth_provider, null: false, foreign_key: true,  on_delete: :restrict
```

app/services/scim_clients/create_service.rb
NobodysNightmare: > Of course, if we add new not oauth2 related authentication method later then it will be a bug here...

That's the reason why I wrote it like this. You are correct that the above suggestion would be simpler and it would currently behave as correctly. While I don't think that we will add a fourth authentication method anytime soon, it feels more correct to me, to keep the slightly more complex if-statement here.

Strong opinions from your side?

ba1ash: No, fine with it.

db/migrate/20250605133700_create_scim_clients.rb
NobodysNightmare: Excellent point. I was under the impression that the foreign key would be created implicitly. Having a foreign key without anything specified, usually means restrict then.

However, I was wrong: No foreign key was created, so no database guarantees. Fixed.

PR #18470, opened 2025-03-27
Test with i7ie

crohr:


crohr:
i7ie instances are not available in eu-west-1 😢

PR #18471, opened 2025-03-27
Re-Implements the NextcloudConnectionValidator using the multple status retun

mereghost:
### Related Work Package: [OP#62102](https://community.openproject.org/work_packages/62102)

Here I introduce a couple of minor refactors to the already in place validators, introducing the `ValidationGroupResult` so that we have a smarter bag of data. ;)

Also introduces a base idea of the main validator _per se_. I intend to move the class level details off of it once OneDrive is done, as requirements will be clear.

Also at that time, I'll probably be able to hook the Registry into it and the list goes on...

### DIFF
--- a/modules/storages/app/common/storages/peripherals/connection_validators/nextcloud_validator.rb
+++ b/modules/storages/app/common/storages/peripherals/connection_validators/nextcloud_validator.rb
@@ -1,0 +1,74 @@
+# frozen_string_literal: true
+
+#-- copyright
+# OpenProject is an open source project management software.
+# Copyright (C) the OpenProject GmbH
+#
+# This program is free software; you can redistribute it and/or
+# modify it under the terms of the GNU General Public License version 3.
+#
+# OpenProject is a fork of ChiliProject, which is a fork of Redmine. The copyright follows:
+# Copyright (C) 2006-2013 Jean-Philippe Lang
+# Copyright (C) 2010-2013 the ChiliProject Team
+#
+# This program is free software; you can redistribute it and/or
+# modify it under the terms of the GNU General Public License
+# as published by the Free Software Foundation; either version 2
+# of the License, or (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program; if not, write to the Free Software
+# Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+#
+# See COPYRIGHT and LICENSE files for more details.
+#++
+
+module Storages
+  module Peripherals
+    module ConnectionValidators
+      class NextcloudValidator
+        # Class Level interface will be moved to a superclass/mixin once we do the OneDrive port
+        class << self
+          def validation_groups
+            @validation_groups ||= {}
+          end
+
+          def register_group(group_name, klass, when: ->(storage, *) { storage.configured? })
+            validation_groups[group_name] = { klass:, when: }
+          end
+        end
+
+        register_group :base_configuration, Nextcloud::BaseConfigurationValidator
+        register_group :authentication, Nextcloud::AuthenticationValidator,
+                       when: ->(_, result) { result.group(:base_configuration).non_failure? }
+        register_group :ampf_configuration, Nextcloud::AmpfConfigurationValidator,
+                       when: ->(storage, result) {
+                         result.group(:base_configuration).non_failure? && storage.automatic_management_enabled?
+                       }
+
+        def initialize(storage:)
+          @storage = storage
+        end
+
+        def validate
+          validation_groups.each_with_object(ValidatorResult.new) do |(key, group_metadata), result|
+            if group_metadata[:when].call(@storage, result)
+              result.add_group_result(key, group_metadata[:klass].new(@storage).call)
+            end
+          end
+        rescue KeyError
+          ValidatorResult.new
+        end
+
+        private
+
+        def validation_groups = self.class.validation_groups
+      end
+    end
+  end
+end
--- a/modules/storages/app/common/storages/peripherals/connection_validators/nextcloud_validator.rb
+++ b/modules/storages/app/common/storages/peripherals/connection_validators/nextcloud_validator.rb
@@ -38,12 +38,12 @@ module Storages
             @validation_groups ||= {}
           end
 
-          def register_group(group_name, klass, when: ->(storage, *) { storage.configured? })
+          def register_group(group_name, klass, when: ->(*) { true })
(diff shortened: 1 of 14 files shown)


modules/storages/app/common/storages/peripherals/connection_validators/nextcloud_validator.rb
```diff
@@ -0,0 +1,74 @@
...
+          end
+
+          def register_group(group_name, klass, when: ->(storage, *) { storage.configured? })
+            validation_groups[group_name] = { klass:, when: }
+          end
+        end
+
+        register_group :base_configuration, Nextcloud::BaseConfigurationValidator
+        register_group :authentication, Nextcloud::AuthenticationValidator,
+                       when: ->(_, result) { result.group(:base_configuration).non_failure? }
```
NobodysNightmare, line 48: Reading the `when`, I am wondering, whether we would execute these validations when the storage is not configured, because in that case `base_configuration` is not tested. I'd expect that this means it's skipped, which is not a failure.

modules/storages/app/common/storages/peripherals/connection_validators/nextcloud_validator.rb
```diff
@@ -0,0 +1,74 @@
...
+          @storage = storage
+        end
+
+        def validate
+          validation_groups.each_with_object(ValidatorResult.new) do |(key, group_metadata), result|
+            if group_metadata[:when].call(@storage, result)
+              result.add_group_result(key, group_metadata[:klass].new(@storage).call)
+            end
+          end
+        rescue KeyError
```
NobodysNightmare: When would a `KeyError` occur in the given code?

NobodysNightmare (approved):
I have two questions, but it generally looks good to me.

modules/storages/app/common/storages/peripherals/connection_validators/nextcloud_validator.rb
mereghost: It could, if for some reason we try to access a result that isn't ready yet. It is more of a safeguard than anything. =)

modules/storages/app/common/storages/peripherals/connection_validators/nextcloud_validator.rb
mereghost: I've been debating this with myself. The base condition is that the storage is configured (the default argument for when).

But, should we even care? Should this be a check that fails on base configuration? Should we let the application take this decision of running validations against a a non-configured storage?

modules/storages/app/common/storages/peripherals/connection_validators/nextcloud_validator.rb
NobodysNightmare: > if for some reason we try to access a result that isn't ready yet

Wouldn't that have a skipped/pending status? Also is there any parallel execution involved?

> It is more of a safeguard than anything.

Not sure if I am a fan of this argumentation. If you want a safeguard, you should rescue from `StandardError` ;-P

mereghost: > Also is there any parallel execution involved?

Nope. At least not right now. We could run in parallel after a `BaseConfiguration`, but I think it is not work the hassle.

`StandardError` is way too broad. `KeyError` is raised when we try to access a result that isn't there yet, typos being the main reason XD

Actually that's why I added it there when I was coding. 🤣

NobodysNightmare: Pardon my insistance on this topic, but isn't an exception for typos during coding _good_ behaviour? So that you notice your typo quickly?

Also I think I still do not understand where exactly I have to make a typo to get a key error.

Anyways, you got my feedback on this, I trust you to come to the right conclusion. I am leaving it up to you to keep this `rescue` in or not. I have not yet been convinced ^^

modules/storages/app/common/storages/peripherals/connection_validators/nextcloud_validator.rb
```diff
@@ -0,0 +1,74 @@
...
+          def validation_groups
+            @validation_groups ||= {}
+          end
+
+          def register_group(group_name, klass, when: ->(storage, *) { storage.configured? })
+            validation_groups[group_name] = { klass:, when: }
+          end
+        end
+
+        register_group :base_configuration, Nextcloud::BaseConfigurationValidator
```
Kharonus: 🟡 I get now a light naming conflict. :D we have a class `BaseValidatorGroup` which is considered an abstract base class. And then we have `BaseConfigurationValidator`, but this is actually NOT a base class, but validates the basic configuration.

Maybe we can rename the later to `ConfigurationValidator`? Would this clash with any of our understandings?

Kharonus: I wouldn't even rename the group, to indicate the "basicness" of this validation group. But for a class name it is kind of ambigious.

modules/storages/app/common/storages/peripherals/connection_validators/nextcloud_validator.rb
Kharonus: I find it nice enough TBH, that we have then a single point in the UI where the user can get information about the correctness of his storage. So, I'm currently pro having missing configurational data expressed as a failure in the first validation group.

(ask me again in one month :D)

modules/storages/app/common/storages/peripherals/connection_validators/nextcloud_validator.rb
Kharonus: you can have a type in the key itself when calling `register_checks`. But I agree with Jan, having an exception when there is a typo is something I wouldn't want to hide.

modules/storages/app/common/storages/peripherals/connection_validators/validator_result.rb
```diff
@@ -0,0 +1,58 @@
...
+
+        def healthy? = group_results.values.all?(&:success?)
+        def unhealthy? = group_results.values.any?(&:failure?)
+        def warning? = group_results.values.any?(&:warning?)
+
+        def group(key) = group_results.fetch(key)
+        alias_method :fetch, :group
+
+        def add_group_result(key, result)
+          Kernel.warn "Overwriting #{key} results" if group_results.key?(key)
```
Kharonus, line 51: ❓ What is `Kernel.warn`? Is this supposed to be logging?

Kharonus (commented):
I have some comments, but none of it would be a real merge blocker.

modules/storages/app/common/storages/peripherals/connection_validators/validator_result.rb
mereghost: Yes. XD I 
was working with normal warnings rather than logs that I forgot to flag it back. Good catch.

modules/storages/app/common/storages/peripherals/connection_validators/nextcloud_validator.rb
mereghost: Ok, adding it to the StorageConfiguration checks.

PR #18475, opened 2025-03-28
[62580] Fix read-only error displayed when editing progress values

cbliard:
# Ticket

https://community.openproject.org/wp/62580

When editing progress values with a user having "Edit work packages" permission but missing "Add work packages" permission, an error message was displayed. This error message should not be displayed as the user is allowed to edit the work package.

## Screenshots

Before:

image

The contract class being used was always `WorkPackages::CreateContract`. It needs to be `WorkPackages::UpdateContract` when the user edits fields so that the correct permissions are checked.

# Merge checklist

- [x] Added/updated tests


bsatarnejad (approved):
LGTM 👍

PR #18478, opened 2025-03-28
[#61889] Primerize Project Settings > Information form

myabc:
# Ticket

https://community.openproject.org/wp/61889

Updates the form visuals and functionality according to new wireframes: the form on the Project Settings > Information page has been split into multiple discrete forms.

In detail, this PR:

* Introduces `Projects::Settings::General::ShowComponent`.
* Creates Primer forms with basic specs.
* Adds `Projects::Settings::GeneralController#update` action, routes and accompanying permissions.
* Updates "Edit Settings" feature specs for new forms, also rewriting some expectations that were reliant on `data-qa-name` attribute that Primer form components do not render by default.
* Removes the old Angular `ProjectsComponent` (`opce-project-settings`)

ℹ️ Work to primerize the Project Create and Copy forms, as well as creating a Project Status Component has been split out into separate PRs:

* #18649 (OP [63482](https://community.openproject.org/wp/63482))
* #18716 (OP [63593](https://community.openproject.org/wp/63593))
* #18715 (OP [63594](https://community.openproject.org/wp/63594))

## Screenshots

Since this is my first attempt at primerizing forms in OpenProject, I've taken a bit longer to try and understand best practices and patterns.

### When should we use Turbo? Frames vs Streams?

**This is the simplest possible form implementation, falling back to Turbo Drive's handling of Form Submissions.**

The pattern of decomposition offered by Turbo Frames would probably make most sense here, and would allow us to optimise page response size slightly (along with https://github.com/hotwired/turbo-rails?tab=readme-ov-file#a-note-on-custom-layouts).

That said, there is also an established pattern of using Turbo Streams + Component Rendering (`update_via_turbo_stream`) used in various newly-primerized pages in the application.

If we were to go with Turbo Streams, then we would probably want to split `Projects::Settings::General::IndexComponent` into separate components.

Frames vs. Streams is also pertinent to #18649 (WP #63482).

### Separate Controller actions for each form?

The PR currently adds a single `Controller#update` action - reused by the three forms on the Project Settings > Information page. Since the page previously consisted of one form, this was the easiest, DRY-est approach. That said, defining separate actions - with a permitted params subset - would be advantageous in terms of security, mitigating potential request parameter tampering.

### Error handling - best practice?

It doesn't look like we aren't handling errors (Service call failures) consistently in the application. See comment: https://github.com/opf/openproject/pull/18478/files#r2060229931

As currently implemented, updates can fail silently. This will happen if the Project is already in an inconsistent state (for example, if new project attributes are changed to required after project creation).

### Form testing - thoughts?

I have added some basic specs for Forms - making use of a subset of Capybara selectors. In the case of Form fields that wrap Angular components, we simply check that the correct attributes (inc. JSON payloads) are passed to Angular.

This work could be tidied up - possibly through addition of Capybara selectors (N.B. while I'm a fan of using [Capybara accessible selectors](https://github.com/citizensadvice/capybara_accessible_selectors), they wouldn't make sense here as we are testing server-rendered HTML. ARIA attributes and roles - which accessible selectors relies on - are typically set on Angular component initialization).

# Merge checklist

- [x] Added/updated tests


app/forms/projects/settings/relations_form.rb
```diff
@@ -0,0 +1,62 @@
...
+      delegate :parent, to: :model
+
+      form do |f|
+        f.project_autocompleter(
+          name: :parent_id,
+          label: attribute_name(:parent_id),
+          autocomplete_options: {
+            model: project_autocompleter_model,
+            focusDirectly: false,
+            dropdownPosition: "bottom",
```
myabc, line 42: Outside of dialogs, are we ok to use the default? (`dropdownPosition: "auto"`)

app/forms/projects/settings/status_form.rb
```diff
@@ -0,0 +1,65 @@
...
+module Projects
+  module Settings
+    class StatusForm < ApplicationForm
+      include ProjectStatusHelper
+
+      form do |f|
+        f.autocompleter(
+          name: :status_code,
+          label: attribute_name(:status_code),
+          include_blank: true,
```
myabc, line 39: `include_blank: true` doesn't appear to do anything for "decorated" autocompleters.

config/locales/en.yml
```diff
@@ -379,6 +379,12 @@ en:
         heading: "Delete this project list?"
         text: "This action will not delete any project the list contains. Are you sure you want to delete this project list?"
     settings:
+      header_details: Basic details
+      header_status: Project status
+      header_relations: Project relations
+      button_update_details: Update details
+      button_update_status: Update status
+      button_update_parent_project: Update parent project
```
myabc, line 392: Not sure what convention we're using for I18n key names?

Also, should we sort keys alphabetically?

app/controllers/projects/settings/general_controller.rb
```diff
@@ -46,4 +48,19 @@ def toggle_public
...
+      .new(model: @project, user: current_user)
+      .call(permitted_params.project)
+
+    @project = call.result
+
+    if call.success?
+      flash[:notice] = I18n.t(:notice_successful_update)
+      redirect_to project_settings_general_path(@project)
+    else
+      render action: :show, status: :unprocessable_entity
```
myabc, line 63: @marcalcobe @psatyal 

Is it enough to display validation errors inline (as currently implemented), or should we also display an additional flash message? (e.g. by adding `flash[:error] = call.message`)

image

image

myabc: It looks like `ErrorMessageHelper#error_messages_for` lets us accomplish the same thing.

app/forms/projects/settings/status_form.rb
klaustopher: I think it might be because you set `clearable: false`. That causes the empty option to be removed

lib/primer/open_project/forms/dsl/autocompleter_input.rb
```diff
@@ -19,12 +19,7 @@ def initialize(label:, value:, classes: nil, selected: false, group_by: nil)
             end
 
             def to_h
-              {
-                id: value,
-                name: label,
-                group_by:,
-                classes:
-              }.compact
+              { id: value, name: label }.merge({ group_by:, classes: }.compact)
```
klaustopher, line 22: What's the benefit of creating 2 hashes, compacting only the 2nd one, and then merging them into a third one, than constructing one and compacting it?

klaustopher (commented):
Great work. Cannot see anything wrong with the implementation per se. Not 100% sure if all of the `concat`s are needed inside the components, but you're better at primer stuff, so I don't doubt it.

I am personally not a huge fan of the multiple forms on one page without a clear distinction between them. We do have the tabbed view and are using this in i.e. the admin settings, etc. But this is not a question for the review of this PR.

klaustopher:
> The PR currently adds a single `Controller#update` action - reused by the three forms on the Project Settings > Information page. Since the page previously consisted of one form, this was the easiest, DRY-est approach. That said, defining separate actions - with a permitted params subset - would be advantageous in terms of security, mitigating potential request parameter tampering.

I think this is fine. We are using that pattern for quite a few places where a controller can update all sorts of things but gets only sparsely filled params so only parts of the model is udpated. The only case where we stray away from this is if updating something has many side effects (i.e. changing the status of a work package has so many side effects, that this should be its own action). But here I think it's completely fine.

klaustopher:
> I have added some basic specs for Forms - making use of a subset of Capybara selectors. In the case of Form fields that wrap Angular components, we simply check that the correct attributes (inc. JSON payloads) are passed to Angular.
> 
> This work could be tidied up - possibly through addition of Capybara selectors (N.B. while I'm a fan of using [Capybara accessible selectors](https://github.com/citizensadvice/capybara_accessible_selectors), they wouldn't make sense here as we are testing server-rendered HTML. ARIA attributes and roles - which accessible selectors relies on - are typically set on Angular component initialization).

Personally, I do not think that a component that sticks together other components really needs unit testing, I'd rather invest in some feature specs that do update things on the project and then test that it did and that the page updated as you intended to. But that's just my opinion :)

klaustopher (approved):

PR #18479, opened 2025-03-28
[#62217] allow custom fields that are not enabled in project

Kharonus:
# Ticket
[OP#62217](https://community.openproject.org/work_packages/62217)

- allow custom fields in generated subjects that are not enabled in project

- rescue when sending custom field methods

### DIFF
--- a/app/models/types/patterns/token_property_mapper.rb
+++ b/app/models/types/patterns/token_property_mapper.rb
@@ -33,6 +33,7 @@ module Types
     class TokenPropertyMapper
       DEFAULT_FUNCTION = ->(key, context) do
         return nil if context.nil?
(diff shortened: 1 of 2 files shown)


Kharonus:
@mereghost we need to talk about this one :D I know, that you will have an opinion on it.

mereghost:
Nice. Do you remember if the custom field code makes changes to `respond_to_missing?`? If so we could just return nil when `context.responds_to?(:method_name)` is false.

Cheaper than the exception.

mereghost (approved):
This works well enough. But see comment in PR. =)

Kharonus:
Changed to `respond_to?` and added a unit test

PR #18484, opened 2025-03-31
[62557][62607] date picker successor and children tabs must show the same correct chronological order

cbliard:
# Ticket

https://community.openproject.org/wp/62557
https://community.openproject.org/wp/62607

The relations order was not the same between date picker tabs and relation tab for the same relation group.

Also the order of relation groups was not the same from one work package to another in the relations tab.

This needs to be consistent.

## Screenshots

Move the ordering logic out of the view and put it in the `WorkPackageRelationsTab::RelationsMediator` object instead:
* `WorkPackageRelationsTab::RelationsMediator` returns all relation groups, child and relations, ordered as they should be
* `WorkPackageRelationsTab::RelationsMediator::RelationGroup` is ordering and returning `WorkPackageRelationsTab::RelationsMediator::RelationItem` objects as they should be
* `WorkPackageRelationsTab::RelationsMediator::RelationItem` have every information needed to do the rendering: visibility, relation, related work package, is the closest one, etc.

This is to alleviate most of the logic from the view, replacing conditionals with more specialized value objects.

# Merge checklist

- [x] Added/updated tests

### DIFF
--- a/app/components/work_package_relations_tab/relation_component.rb
+++ b/app/components/work_package_relations_tab/relation_component.rb
@@ -34,7 +34,7 @@ class WorkPackageRelationsTab::RelationComponent < ApplicationComponent
 
   attr_reader :relation_item, :editable
 
-  delegate :relation, :work_package, to: :relation_item
+  delegate :closest?, :relation, :visible?, :work_package, to: :relation_item
 
   # Checks if the relation or child work package is visible to the current user
   #
@@ -52,10 +52,10 @@ class WorkPackageRelationsTab::RelationComponent < ApplicationComponent
   end
 
   def child
-    if parent_child_relationship?
-      relation_item.related
-    end
+    relation_item.related if parent_child_relationship?
   end
+
+  def editable? = editable
 
   private
 
@@ -67,7 +67,7 @@ class WorkPackageRelationsTab::RelationComponent < ApplicationComponent
   end
 
   def should_render_action_menu?
-    return false unless editable
+    return false unless editable?
 
     if parent_child_relationship?
       allowed_to_manage_subtasks?
@@ -83,10 +83,6 @@ class WorkPackageRelationsTab::RelationComponent < ApplicationComponent
 
   def allowed_to_manage_relations?
     helpers.current_user.allowed_in_project?(:manage_work_package_relations, work_package.project)
-  end
-
-  def visible?
-    relation_item.visible?
   end
 
   def should_display_description?
@@ -113,10 +109,6 @@ class WorkPackageRelationsTab::RelationComponent < ApplicationComponent
     return false if parent_child_relationship?
 
     relation.relation_type_for(work_package) == Relation::TYPE_PRECEDES
-  end
-
-  def closest?
-    relation_item.closest?
   end
 
   def edit_path
--- a/app/components/work_package_relations_tab/relation_component.rb
+++ b/app/components/work_package_relations_tab/relation_component.rb
@@ -32,27 +32,21 @@ class WorkPackageRelationsTab::RelationComponent < ApplicationComponent
   include ApplicationHelper
   include OpPrimer::ComponentHelpers
 
-  attr_reader :work_package, :relation_item, :relation, :child, :editable
+  attr_reader :relation_item, :editable, :closest
+
+  delegate :relation, :work_package, to: :relation_item
 
   # Checks if the relation or child work package is visible to the current user
   #
-  # @param work_package [WorkPackage] The work package whose relations are being displayed
-  # @param relation [Relation, nil] The relation between work packages, if any
-  # @param child [WorkPackage, nil] The child work package, if this is a parent-child relationship
+  # @param relation_item [WorkPackageRelationsTab::RelationsMediator::RelationItem] The relation item to display
   # @param editable [Boolean] Whether the relation can be edited
   # @param closest [Boolean] Whether the follows relation is the closest
-  def initialize(work_package:,
-                 relation_item:,
-                 relation:,
-                 child: nil,
+  def initialize(relation_item:,
                  editable: true,
                  closest: false)
     super()
 
-    @work_package = work_package
     @relation_item = relation_item
-    @relation = relation
-    @child = child
     @editable = editable
     @closest = closest
   end
@@ -61,9 +55,15 @@ class WorkPackageRelationsTab::RelationComponent < ApplicationComponent
     relation_item.related
   end
 
+  def child
+    if parent_child_relationship?
+      relation_item.related
+    end
+  end
+
   private
 
-  def parent_child_relationship? = @relation.nil?
+  def parent_child_relationship? = relation.nil?
 
   def should_render_edit_option?
     # Children have nothing to edit as it's not a relation.
@@ -81,12 +81,12 @@ class WorkPackageRelationsTab::RelationComponent < ApplicationComponent
   end
 
   def allowed_to_manage_subtasks?
-    helpers.current_user.allowed_in_project?(:manage_subtasks, @work_package.project) &&
-      helpers.current_user.allowed_in_project?(:manage_subtasks, @child.project)
+    helpers.current_user.allowed_in_project?(:manage_subtasks, work_package.project) &&
+      helpers.current_user.allowed_in_project?(:manage_subtasks, child.project)
   end
 
   def allowed_to_manage_relations?
-    helpers.current_user.allowed_in_project?(:manage_work_package_relations, @work_package.project)
+    helpers.current_user.allowed_in_project?(:manage_work_package_relations, work_package.project)
   end
 
   def visible?
@@ -120,22 +120,22 @@ class WorkPackageRelationsTab::RelationComponent < ApplicationComponent
   end
 
   def closest?
-    @closest
+    closest
   end
 
   def edit_path
     if parent_child_relationship?
       raise NotImplementedError, "Children relationships are not editable"
     else
-      edit_work_package_relation_path(@work_package, @relation)
+      edit_work_package_relation_path(work_package, relation)
     end
   end
 
   def destroy_path
     if parent_child_relationship?
-      work_package_children_relation_path(@work_package, @child)
+      work_package_children_relation_path(work_package, child)
     else
-      work_package_relation_path(@work_package, @relation)
+      work_package_relation_path(work_package, relation)
     end
   end
 
--- a/app/components/work_package_relations_tab/relation_component.rb
+++ b/app/components/work_package_relations_tab/relation_component.rb
@@ -1,44 +1,69 @@
+# frozen_string_literal: true
+
+#-- copyright
+# OpenProject is an open source project management software.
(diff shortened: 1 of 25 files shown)


app/components/work_package_relations_tab/relation_component.rb
```diff
@@ -1,44 +1,69 @@
...
   include ApplicationHelper
   include OpPrimer::ComponentHelpers
 
-  attr_reader :work_package, :relation, :visibility, :child, :editable
+  attr_reader :work_package, :relation_item, :relation, :child, :editable
 
   # Checks if the relation or child work package is visible to the current user
   #
   # @param work_package [WorkPackage] The work package whose relations are being displayed
   # @param relation [Relation, nil] The relation between work packages, if any
```
myabc: Can you document the `@param relation_item` here too?

cbliard: Will do.
It's still a draft at the moment. I just want the test suite to run.

app/components/work_package_relations_tab/relations_mediator.rb
```diff
@@ -46,16 +58,59 @@ def all_relations
...
-        .max
-        &.relation
+  # Represents a relation item to be displayed in the relations tab or the date
+  # picker tabs.
+  #
+  # @param type [String] The type of relation
+  # @param work_package [WorkPackage] The work package for which the relation
+  #   is displayed
+  # @param related [WorkPackage] The related work package: the other work
+  #   package of therelation, or the child for child relations.
```
myabc: missing space: `therelation`

app/models/relation.rb
```diff
@@ -49,6 +49,9 @@ class Relation < ApplicationRecord
   TYPE_PARENT       = "parent"
   TYPE_CHILD        = "child"
 
+  # The order of the types is important. It's used to build up `ORDERED_TYPES`
+  # which is used to order relations of different kind like in the "Add relation"
+  # menu or the relations tab.
```
myabc, line 54: 👍🏻

app/components/work_package_relations_tab/relation_component.rb
```diff
@@ -1,44 +1,65 @@
...
+#++
+
 class WorkPackageRelationsTab::RelationComponent < ApplicationComponent
   include ApplicationHelper
   include OpPrimer::ComponentHelpers
 
-  attr_reader :work_package, :relation, :visibility, :child, :editable
+  attr_reader :relation_item, :editable
+
+  delegate :relation, :work_package, to: :relation_item
```
myabc:
```suggestion
  delegate :relation, :work_package, :visible?, :closest?, to: :relation_item
```

app/components/work_package_relations_tab/relation_component.rb
```diff
@@ -1,44 +1,65 @@
...
-                              else
-                                relation.from == work_package ? relation.to : relation.from
-                              end
+    relation_item.related
+  end
+
+  def child
+    if parent_child_relationship?
+      relation_item.related
+    end
```
myabc:
```suggestion
   relation_item.related if parent_child_relationship?
```

reads a bit better to me.

app/components/work_package_relations_tab/relation_component.rb
```diff
@@ -1,44 +1,65 @@
...
+#
+# See COPYRIGHT and LICENSE files for more details.
+#++
+
 class WorkPackageRelationsTab::RelationComponent < ApplicationComponent
   include ApplicationHelper
   include OpPrimer::ComponentHelpers
 
-  attr_reader :work_package, :relation, :visibility, :child, :editable
+  attr_reader :relation_item, :editable
```
myabc, line 35: While working on this part of the codebase a few weeks ago, I hesitated about getting rid of the `editable` accessor. For the sake of consistency, I would prefer to define an `editable?` predicate method. That said, `editable` appears to only be called once - from `should_render_action_menu?` - so we could forgo the method completely.

app/components/work_package_relations_tab/relations_mediator.rb
```diff
@@ -46,16 +58,59 @@ def all_relations
...
+
+    def ghost_relation_items
+      to_relation_items(ghost_relations, :ghost)
+    end
+
     def closest_relation?(relation) = closest_relation == relation
 
-    def closest_relation
-      return nil unless type.follows?
+    def to_relation_items(relations, visibility)
```
myabc, line 65: this method could be made private.

app/components/work_package_relations_tab/relations_mediator.rb
```diff
@@ -29,9 +29,21 @@
 #++
 
 class WorkPackageRelationsTab::RelationsMediator
-  RelationGroup = Data.define(:type, :visible_relations, :ghost_relations) do
-    def initialize(type:, visible_relations:, ghost_relations:)
-      super(type: ActiveSupport::StringInquirer.new(type), visible_relations:, ghost_relations:)
+  RelationGroup = Data.define(:type, :work_package, :visible_relations, :ghost_relations, :closest_relation) do
+    def initialize(type:, work_package:, visible_relations:, ghost_relations:)
+      type = ActiveSupport::StringInquirer.new(type.to_s)
+      closest_relation = self.class.closest_relation(type, visible_relations + ghost_relations)
```
myabc: 👍🏻 good that we're only calculating the closest relation once on instantiation (that was my mistake).

We could remove the type param from `self.closest_relation` and check the type here instead.

```suggestion
      closest_relation = self.class.closest_relation(visible_relations + ghost_relations) if type.follows?
```

app/components/work_package_relations_tab/relations_mediator.rb
```diff
@@ -29,9 +29,21 @@
...
+      super(type:, work_package:, visible_relations:, ghost_relations:, closest_relation:)
+    end
+
+    def self.closest_relation(type, relations)
+      return nil unless type.follows?
+
```
myabc:
```suggestion
   def self.closest_relation(relations)
```

We could also make this a `private_class_method`.

myabc: That said, it might be even more sense to move the closest logic to `ClosestRelation`, e.g.

```ruby
    def self.of(relations)
      relations
        .map { new(it) }
        .select(&:soonest_start)
        .max
        &.relation
    end
```

spec/components/work_package_relations_tab/relation_component_spec.rb
```diff
@@ -43,92 +43,119 @@
...
+          component = render_component(relation_item:, editable: true)
           expect(component).to have_menu # FIXME: aria-labelledby does not resolve here "Relation actions"
           expect(component).to have_selector :menuitem, "Delete relation"
         end
       end
+
+      context "when not editable" do
+        it "does not render an action menu" do
+          component = render_component(relation_item:, editable: false)
+          expect(component).not_to have_menu
```
myabc, line 74: 👍🏻

spec/components/work_package_relations_tab/relations_mediator_spec.rb
```diff
@@ -94,5 +129,33 @@
...
       end
     end
+
+    describe "#all_relation_items" do
+      let(:work_package) { build_stubbed(:work_package) }
+
+      it "returns all relations of the group as RelationItem instances, " \
+         "ordered by oldest first (lowest id first), mixing visible and ghost relations" do
+        relation_group = described_class::RelationGroup.new(
+          type: "relates",
```
myabc: would it be more consistent to use the constant here?

```suggestion
          type: Relation::TYPE_RELATES,
```

myabc (commented):
I haven't had a chance to look at the table helper code just yet, but thought I would go ahead and give feedback on the substantive part of the PR:

Overall this is a significant improvement to `RelationsComponent`, `RelationsMediator` and associated classes. It makes the code a lot easier to reason about.

I have to admit that I don't particularly like `RelationItem` as a name (it's generic and not immediately clear how it differs from `Relation`). Having said that, I don't have a better suggestion for the moment. The fact that it is a nested class within the `WorkPackageRelationsTab` module does make it clear that it is UI code.

myabc:
Ideally there would also be some basic component spec coverage for the `DialogContentComponent` relations tabs. `dialog_content_component_spec.rb` really only tests `DatePicker::FormComponent`.

app/components/work_package_relations_tab/relation_component.rb
cbliard: That's funny. Actually it does not read better for me.

When I scan it, I tend to omit the `if` and I think that it always returns a value. It's only once I read it that I see the `if`. That's why most of the time I reserve one-line if for guard clauses.

What would read better for me would be something like this:

```
return unless parent_child_relationship?

relation_item.related
```

Also, I could inline the `child` and replace its call with `relation_item.related` directly but I found `child` more intention-revealing.

What do you think?

app/components/work_package_relations_tab/relations_mediator.rb
cbliard: I like the idea of moving it to `ClosestRelation`. I'll try that.

spec/components/work_package_relations_tab/relations_mediator_spec.rb
cbliard: yes, definitely.
Actually, I changed "children" to "child", then I looked into `Relation` to see how to implement the order and saw that constant `TYPE_CHILD`, so I thought I should use it, and then the editor suggested to replace more of them which seemed like a good idea, but I did not replace them all.

That's good that you spotted this one.

app/components/work_package_relations_tab/relation_component.rb
cbliard: I prefer `editable?` too. I added the method and kept the `attr_reader`.

app/components/work_package_relations_tab/relation_component.rb
cbliard: Adopted.

I have a love-hate relation with `delegate`. I find it make the code harder to scan, and may also hide some unwanted chained calls (for instance `relation.description.present?` is actually `relation_item.relation.description.present?`) but on the other hand, it's simply convenient and conveys the idea that some methods are delegated.

Not sure why I have that feeling though.

app/components/work_package_relations_tab/relations_mediator.rb
cbliard: Actually most of the class could be made private. The methods are used only to build the relation items.

I have an experiment on another branch that I just pushed where I introduced a `RelationGroupBuilder` where all these methods are grouped, and it builds up a `RelationGroup` which only has `type` and `relation_items` left.

Could you review it and tell me if I'm pushing it too far or if it's worth adding it? The commit is here: https://github.com/opf/openproject/commit/b5b64f7cce69ec93393fe7005c423cb5c60e3491

myabc (approved):

app/components/work_package_relations_tab/relation_component.rb
cbliard: I used the code you suggested.

PR #18490, opened 2025-03-31
Feature/57388 save export configuration for next export of a view

EinLama:
# Ticket
https://community.openproject.org/wp/57388

Offer a checkbox that users can use to save the settings for a work package export.

Important caveats:

* this feature is only offered for saved queries, unsaved queries cannot have their export settings saved
* each export type (csv, xls, pdf report, pdf gantt, pdf table) has their settings saved individually
* only the changes in the currently displayed tab are saved

## Screenshots
https://github.com/user-attachments/assets/ba7fe721-cc03-4cb4-be96-a77a15f028ff

Created a new table for saving export settings, tied to a query. Each format has its own row. This allows us to check for uniqueness so that each query can only have one setting per format. The actual settings are saved via JSONB, so that we can add more settings here without having to perform migrations.

Please note that this is my first Open Project migration. If there is a process I might have missed it. So take care while reviewing.

# Merge checklist

- [x] Added/updated tests

### DIFF
--- a/app/components/work_packages/exports/base_export_settings_component.rb
+++ b/app/components/work_packages/exports/base_export_settings_component.rb
@@ -35,16 +35,16 @@ module WorkPackages
       include OpTurbo::Streamable
       include WorkPackagesHelper
 
-      attr_reader :query
+      def query
+        model
+      end
 
-      def initialize(query)
-        super
-
-        @query = query
+      def format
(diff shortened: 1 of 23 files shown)


as-op (approved):
👍  Well done, nicely readable and understandable code! No notes.

PR #18491, opened 2025-03-31
:unacceptable is not a recognized rails status code

crohr:
It is `:not_acceptable`.


oliverguenther (approved):

crohr:
@oliverguenther I now wonder if the status was meant to be `422` (i.e. `:unprocessable_entity`) instead? Would be more in line with the static asset returned.

PR #18502, opened 2025-04-01
[#62657] PDF Export of a single work package in query view does not open dialog

as-op:
# Ticket

https://community.openproject.org/work_packages/62657

There are two context menus from where the generate PDF dialog can be invoked. One of them did not work properly.

Send a turbo request in both menus.


as-op:
Closing because targeting the wrong branch

PR #18513, opened 2025-04-02
Update remote identity indices

NobodysNightmare:
### Fixing unique index

Since changing the table layout, uniqueness should be ensured across a wider range of columns.

In general it's necessary to include *_type and *_id columns of any polymorphic association.

Since the introduction of the integration association, it's also required to include this in the index. It's very expected to see remote identities of the same user in multiple integrations, but it's also (in some edge cases) possible to even have the same user twice per integration with different authentication providers.

### Optimizing foreign key indices

There were a few useless indices and some were too scattered to be optimal for queries. E.g. only indexing by integration_id is inefficient when there are multiple integrations and since integration is a polymorphic association, there is no useful query where anyone would just filter data by the integration_id, it's always a composite of integration type and id. Just filtering by integration type would at least semantically make sense (hence type is the first in the composite).

# Ticket
Discovered while working on https://community.openproject.org/wp/62328

### DIFF
--- a/db/migrate/20250402083709_change_remote_identities_foreign_key_indices.rb
+++ b/db/migrate/20250402083709_change_remote_identities_foreign_key_indices.rb
@@ -1,0 +1,39 @@
+# frozen_string_literal: true
+
+#-- copyright
+# OpenProject is an open source project management software.
+# Copyright (C) the OpenProject GmbH
+#
+# This program is free software; you can redistribute it and/or
+# modify it under the terms of the GNU General Public License version 3.
+#
+# OpenProject is a fork of ChiliProject, which is a fork of Redmine. The copyright follows:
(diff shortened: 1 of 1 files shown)


NobodysNightmare:
@machisuji asking for your feedback additionally, because this PR is touching a few indices. Any additional advice or requirements from your side? E.g. should the migration be split out into more steps?

NobodysNightmare:
Nvm, I have to rework this PR.

Lots of the changes were apparently due to me reviewing an intermediate state of the PR that introduced the new columns. The actual database state is better than what I had locally...

NobodysNightmare:
Okay, after fixing my local errors the main problems were already solved :see_no_evil: 

However, I could keep two optimizations, so I am going to leave this PR up for review anyways.

machisuji (approved):
👍

PR #18516, opened 2025-04-02
Early remote identity

NobodysNightmare:
For SSO users there is no need to wait with fetching of a remote identity until they visit the files tab (or whatever place they'd find an integration) before we can establish the remote identity. This is usually possible right after login.

Fetching the remote identities during login has some advantages:

* Users navigating to the files tab for the first time have a higher chance of already having their permissions configured (more time for background job to do this)
* We avoid checking the presence of remote identities on every API call (a consequence of the previous event-based approach)

# Review note

~~This PR is a draft right now.~~ _Update: The PR is now ready for general review._ I want to have feedback on the **general idea** to fetch remote identities once and the interface between the three involved modules (everything around `RemoteIdentities::AutoCreate`):

* `RemoteIdentities` provides central APIs used by other modules
* `storages` (the integration) registers itself in its engine
* OIDC (the auth provider) calls the login hook

The implementation is so far still very OIDC-specific. Doing it more generic would be hard, because I have no clue how the same integration would need to look like for SAML. Though I at least have ways to imagine on how the interface could be generalized in the future. E.g. the `token_fetcher` might be provided largely by the OIDC module, then the integration would just have to provide an identifier, which is the `audience` for OIDC and maybe something else for SAML. Though I'd stall such a refactoring to once we need it.

# Ticket
https://community.openproject.org/wp/62328

# Merge checklist

- [x] Added/updated tests
- [x] Refactoring of the code to be non-drafty


app/services/remote_identities/auto_create.rb
```diff
@@ -0,0 +1,71 @@
...
+  class AutoCreate
+    IntegrationConfig = Data.define(:name, :integration_fetcher, :token_fetcher)
+
+    class << self
+      def register(name, integration_fetcher:, token_fetcher:)
+        configs << IntegrationConfig.new(name:, integration_fetcher:, token_fetcher:)
+      end
+
+      def handle_login(user)
+        configs.each do |config|
```
mereghost, line 41: 🟡 Will we run through all registered configs every time `.handle_login` is called or am I holding it wrong?

app/services/remote_identities/auto_create.rb
```diff
@@ -0,0 +1,71 @@
...
+        RemoteIdentities::CreateService
+          .new(user:, integration:, auth_source: user.authentication_provider)
+          .call(token:)
+          .on_failure do |e|
+            error("RemoteIdentity creation for user #{user.id} failed: #{e.message}")
+          end
+      end
+
+      def error(message)
+        Rails.logger.error(message)
```
Kharonus, line 67: 🟡 can we use tagged logging here?

app/services/remote_identities/auto_create.rb
```diff
@@ -0,0 +1,71 @@
...
+#
+# See COPYRIGHT and LICENSE files for more details.
+#++
+
+module RemoteIdentities
+  class AutoCreate
+    IntegrationConfig = Data.define(:name, :integration_fetcher, :token_fetcher)
+
+    class << self
+      def register(name, integration_fetcher:, token_fetcher:)
```
Kharonus, line 36: 🟡 It did take me some time and hoops to understand, that those two `fetchers` are stateless callbacks. Could we show this more, maybe directly in the name?

`fetch_integration_callback` and `fetch_token_callback`

this is clearly highly subjective ^^ just wanted to express, why it took me so long to understand the functionality.

modules/storages/lib/open_project/storages/engine.rb
```diff
@@ -135,21 +135,26 @@ def self.external_file_permissions
...
-              .call(user: token.user, integration: storage, token:)
-              .on_failure do |failure|
-                Rails.logger.error("RemoteIdentity creation for user #{token.user.id} failed: #{failure.message}")
-              end
+    initializer "openproject_storages.remote_identity_setup" do
+      Rails.application.config.to_prepare do
+        RemoteIdentities::AutoCreate.register(
+          :storages,
+          integration_fetcher: ->(user) do
+            Storages::NextcloudStorage.all.select do |storage|
```
Kharonus, line 146: 🟡 can we add a named scope for this? looks complex enough ;)

app/services/remote_identities/auto_create.rb
```diff
@@ -0,0 +1,71 @@
...
+        configs << IntegrationConfig.new(name:, integration_fetcher:, token_fetcher:)
+      end
+
+      def handle_login(user)
+        configs.each do |config|
+          config.integration_fetcher.call(user).each do |integration|
+            token = config.token_fetcher.call(user, integration)
+            create_remote_identity(user:, integration:, token:)
+          end
+        rescue StandardError => e
```
Kharonus, line 46: 🟡 error handling is a bit clumsy me thinks... This probably comes from this reason:

- fetching integrations and fetching tokens have fully different interfaces
- fetching integrations is an active relation that is evaluated on `each`
- fetching tokens is a "hidden" result monad, which is evaluated with .value! internally and hence is throwing an exception

I think, we could achieve more readability here, if we agree on an interface for both callbacks, and handle them similar. Of course, this could be a result monad for both. :D But I'd say, if you find another interface (exception handling etc.) 

Ideally the resulting code would read more like this:

```ruby
integrations = yield config.integration_fetcher.call(user) # the yield is just the example for the do notation, but it is placeholder for any error handling)

integrations.each do |integration|
  token = yield config.token_fetcher.call(user, integration)
  yield create_remote_identity(user:, integration:, token:)
end
```

My goal with this piece of code is to get rid of the "omnipotent" rescue around the multiple layers.

But this also reveals another flaw, we currently have: Failing to fetch a token interrupts the whole execution, while failing to create a remote identity "only" leads to a error log, while the code continues. So, I'd say we either say we create remote identities ONLY if we got tokens for every integration, and we do so in a transaction to roll back. OR: we actually want partial creations and simply log failures for single integrations. But this should include failing to retrieve tokens.

uff, long text... but worthy of discussion, imho ^^

modules/storages/lib/open_project/storages/engine.rb
```diff
@@ -135,21 +135,26 @@ def self.external_file_permissions
...
-          audience = payload[:audience]
-          token = payload[:token]
-          storage = Storages::Storage.with_audience(audience).first
-          if storage
-            RemoteIdentities::CreateService
-              .call(user: token.user, integration: storage, token:)
-              .on_failure do |failure|
-                Rails.logger.error("RemoteIdentity creation for user #{token.user.id} failed: #{failure.message}")
-              end
+    initializer "openproject_storages.remote_identity_setup" do
```
Kharonus, line 141: 🔴 it took me a while to think, if we break something by removing the event subscription. And here is my take:

- user logs into OP where no SSO storage is configured
- user is surprisingly an admin and configures an SSO storage with AMPF
- user enables the file storage in any project
  - the sync job is run

when the job is run at that time, there is no remote identy, is there?  so, even if the user now goes to the files tab, which triggers a token fetch (and previously would have given the user the remote identity), he still is without, as now only the login to OP triggers the creation of a remote identity.

Or did I miss an important part you added?

Kharonus (changes requested):

app/services/remote_identities/auto_create.rb
NobodysNightmare: > uff, long text... but worthy of discussion, imho ^^

Probably asks for an in-person talk... tomorrow xD

But quick instant reaction:

The purpose of the "omnipotent rescue" was exactly to ensure that one integration can't ruin the entire login procedure. I have to double check the code myself again to see whether this was achieved as intended. In general I'd agree that errors duing token fetching should be covered with the same certainty as errors creating the remote identity. As far as I can tell, both potentially include interactions with third party systems, which bears some risk of failures.

modules/storages/lib/open_project/storages/engine.rb
NobodysNightmare: That's a fair point. It's even true for all users that logged in before the admin. Effectively the new implementation means that everyone has to relog once a storage was set-up. That sucks.

I mean... At least we should show a nice error message in that case, hinting at the fact that logging in again should help. But I didn't want to optimize for showing this error message to more people.

On the other hand, I also was happy to get rid of this event, because I was not a huge fan of the abstraction we built here (it feels weird to have a subscribable event for fetching a user token). And having both events feels wrong as well.

app/services/remote_identities/auto_create.rb
```diff
@@ -0,0 +1,71 @@
...
+          error("Unhandled exception while creating #{config.name} RemoteIdentity for user #{user.id}: #{e.message}")
```
mereghost, line 47: 🔴  By rescuing here and only logging the error, we are potentially hiding issues. At least we should notify AppSignal if available.

app/services/remote_identities/auto_create.rb
```diff
@@ -0,0 +1,71 @@
...
+# See COPYRIGHT and LICENSE files for more details.
+#++
+
+module RemoteIdentities
+  class AutoCreate
+    IntegrationConfig = Data.define(:name, :integration_fetcher, :token_fetcher)
+
+    class << self
+      def register(name, integration_fetcher:, token_fetcher:)
+        configs << IntegrationConfig.new(name:, integration_fetcher:, token_fetcher:)
```
mereghost, line 37: 🟢 Should we make config a `Concurrent::Array` for thread safety?

modules/openid_connect/lib/open_project/openid_connect/hooks/hook.rb
```diff
@@ -38,22 +38,22 @@ def user_logged_in(context)
...
+          refresh_token: session["omniauth.oidc_refresh_token"],
+          expires_in: session["omniauth.oidc_expires_in"],
+          known_audiences: [OpenIDConnect::UserToken::IDP_AUDIENCE],
+          # We clear previous tokens while adding this one to avoid keeping
+          # stale tokens around (and to avoid piling up duplicate IDP tokens)
+          # -> Fresh login causes fresh set of tokens
+          clear_previous: true
+        )
+
+        RemoteIdentities::AutoCreate.handle_login(user)
```
mereghost, line 56: 🟢 should this call be part of the `UserTokens::CreateService`?

modules/openid_connect/spec/lib/open_project/openid_connect/hooks/hook_spec.rb
```diff
@@ -110,6 +110,7 @@
 
     before do
       allow(OpenIDConnect::UserTokens::CreateService).to receive(:new).and_return(create_service)
+      allow(RemoteIdentities::AutoCreate).to receive(:handle_login)
```
mereghost, line 113: 🟡 Those can hide wrong args being passed to the methods.

Also could they be set as `class_double` or `class_spy` (for the latter)?

modules/storages/lib/open_project/storages/engine.rb
```diff
@@ -135,21 +135,26 @@ def self.external_file_permissions
...
+      Rails.application.config.to_prepare do
+        RemoteIdentities::AutoCreate.register(
+          :storages,
+          integration_fetcher: ->(user) do
+            Storages::NextcloudStorage.all.select do |storage|
+              storage.configured? &&
+                Storages::Peripherals::StorageInteraction::AuthenticationMethodSelector.new(user:, storage:).sso?
+            end
+          end,
```
mereghost, line 150: 🟡 We load all `Nextcloud` storages in memory and then cycle through all of them here.

Couldn't we just load the ones that have the SSO authentication scheme active?

modules/storages/lib/open_project/storages/engine.rb
```diff
@@ -135,21 +135,26 @@ def self.external_file_permissions
...
+          token_fetcher: ->(user, storage) do
+            OpenIDConnect::UserTokens::FetchService
+              .new(user:)
+              .access_token_for(audience: storage.audience)
+              .value! # should we do something else than raising on error?
```
mereghost, line 155: 🟢 To address the question: maybe? The user should at least be informed what has happened wrong, but since it is logging in, like for the first time, it could be very confusing.

app/services/remote_identities/auto_create.rb
mereghost: Yes, this would help others to understand the code better. We are using the term `fetcher` a lot recently, there must have a better word to represent it:

`integration_lookup_callback`... idk I suck at naming. xD

app/services/remote_identities/auto_create.rb
mereghost: `Rails.logger.tagged('tag name').error(message)` iirc.

modules/storages/lib/open_project/storages/engine.rb
mereghost: What bothers me the most is that we now have 2 different points in time where the user is logged against the storage.

OAuth users log when they first interact with it. SSO users, maybe, do it at OP login. They could do it by interacting with a storage... probably (not sure since the event handling was removed).

app/services/remote_identities/auto_create.rb
```diff
@@ -0,0 +1,71 @@
...
+#
+# You should have received a copy of the GNU General Public License
+# along with this program; if not, write to the Free Software
+# Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+#
+# See COPYRIGHT and LICENSE files for more details.
+#++
+
+module RemoteIdentities
+  class AutoCreate
```
mereghost, line 32: ❓ I have a wider question: What is the advantage here compared to emitting an event when the user logs in?

mereghost (changes requested):

modules/storages/lib/open_project/storages/engine.rb
NobodysNightmare: > What bothers me the most is that we now have 2 different points in time where the user is logged against the storage.

I am not sure how I feel about this. First of all "logged" is a very vague term. There is no "login" required to access the storage. A token is required and in both cases we aquire the token as late as possible.

On the other hand (and that's the process changed in this PR) we need to know the users ID in the remote system at some point and we can only do that as soon as we are able to authenticate requests for the user. This moment in time is inherently different for new users in both approaches. A new "SSO user" can immediately after login make authenticated requests, so we _can_ immediately start obtaining the remote identity. A new "OAuth2 user" can't authenticate until they actively interact with a specific "login UI". On the other hand, there is no such UI required or even desirable for the SSO case. This means it's virtually impossible to obtain the remote identity for both kinds of users at exactly the same time.

app/services/remote_identities/auto_create.rb
NobodysNightmare: Yes, that's correct. It effectively means that each module can add its own configuration and we run through the config for each module, so that remote identities for all the different integrations would be created.

modules/openid_connect/spec/lib/open_project/openid_connect/hooks/hook_spec.rb
NobodysNightmare: Note that down below there is an expectation that expects the exact arguments being passed.

That's the usual way I try to write stubbing specs:

* Allow _anything_ while stubbing
* Expect specifically during expectation

The reason is that otherwise a broken implementation (which is exactly what specs should be able to detect) can lead to a bypass of the stubbed behaviour, which can have varying consequences (side effects of real implementation). The worst offender would be that real HTTP calls happen when they were intended to be stubbed away, though to be fair this is not possible in our test suite due to webmock's strict settings.

modules/storages/lib/open_project/storages/engine.rb
NobodysNightmare: The current implementation makes sure that **only** the `AuthenticationMethodSelector` is responsible to figure out which storages are doing SSO authentication. Your proposal would be more performant at the cost of duplicating "code knowledge" of which storages qualify for SSO. One example is that we currently have one dormant authentication method for SSO+fallback that would need to be considered here as well.

Now, I can accept code duplication if the performance gains are worth it. But when we talk about loading **all** Nextcloud storages into memory, we talk about one? Two? **Five**? I don't see a risk of this number scaling very high to be honest.

app/services/remote_identities/auto_create.rb
NobodysNightmare: I see two lines of argumentation:

1. Code reuse: _Once_ there are other integrations that also need to create remote identities during login (our expectation is that all of them would need it), they'd otherwise have to repeat the same code for remote identity creation
2. Event clarity: There are already two event "when the user logs in" and the OIDC module hooks into both of them (`omniauth_user_authorized` and `user_logged_in`). I already have trouble remembering at which exact point of the login flow both happen, but they occur in the order I listed them. The code for remote identity creation has to run after we obtained the access token, so a third event would be necessary and it should receive a clear name.

That said: Both approaches can work well.

mereghost: I ask because the current implementation departs heavily from the use of events that the rest of the code base relies heavily (mildly?) on. It is not inherently wrong or bad, it is just different. 

Now that you mention the "login" events, maybe we should have a quick event storming session about the process of user <=> storage login. Why? We have this change, the ensure connection thingy, a couple of extra checks here and there etc. I'm getting the feeling that this entire process isn't very clear to us all.

app/services/remote_identities/auto_create.rb
mereghost: Could you think of a way or case were we just want to run one of them?

NobodysNightmare: Not really TBH. Hypothetically if one of them failed and we wanted to retry at a later time, but none of that exists yet.

app/services/remote_identities/auto_create.rb
NobodysNightmare: Sounds good to me.

mereghost: Oh, and that shouldn't block this PR at all. We can always go back to the drawing board with the knowledge we gain from the session.

app/services/remote_identities/auto_create.rb
Kharonus: I meant for the `TaggedLogging`  helper, but now I realized this one is only under storages module. SO, you probably can disregard this comment

mereghost: The effect is the same. It is always worth tagging the logs, as it makes easier to figure out things.

NobodysNightmare:
Closing this PR. More infos can be found in the linked work package.

The **tl;dr** is that not enough of the original problem was solved by this approach, while too many genuinely new problems were introduced.

PR #18521, opened 2025-04-02
avoid long list of ids in filter

ulferts:
# Ticket

https://community.openproject.org/wp/62541

Avoid having a long list of ids in the filter that lead to the query url becoming too long.

The endpoint filtered against, /api/v3/principals, already has the [visible scope attached](https://github.com/opf/openproject/blob/dev/app/models/queries/principals/principal_query.rb#L38) to it. Therefore, no non visible principals can be retrieved anyway. The visible scope will result in filtering for those principals, that are in any visible project the current user is also in. Unless the user is admin in which case all principals are available.


dombesz (approved):
Thanks Jens, it works as expected.👍

PR #18522, opened 2025-04-02
Use component for global and project definition administration

ulferts:
# Ticket

https://community.openproject.org/wp/62709

Use a common component on both the global as well as the project specific administration to render the information about the project phase definition:

image

image

There shouldn't be visible changes.

### DIFF
--- a/app/components/projects/phase_definition_component.rb
+++ b/app/components/projects/phase_definition_component.rb
@@ -31,10 +31,6 @@ module Projects
   class PhaseDefinitionComponent < ApplicationComponent
     include OpPrimer::ComponentHelpers
 
-    # TODO: This component is currently not in use! It should be a shared component
-    # between the Projects::Settings::LifeCycleSteps::StepComponent and the
-    # Settings::ProjectLifeCycleStepDefinitions::RowComponent.
-    # It should hold the icon, definition name and gate text information.
     def phase_text
       model.name
     end
--- a/app/components/projects/phase_definition_component.rb
+++ b/app/components/projects/phase_definition_component.rb
@@ -28,14 +28,22 @@
 # See COPYRIGHT and LICENSE files for more details.
 # ++
 module Projects
-  class LifeCycleComponent < ApplicationComponent
+  class PhaseDefinitionComponent < ApplicationComponent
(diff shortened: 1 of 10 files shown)


app/components/projects/phase_definition_component.rb
```diff
@@ -28,14 +28,22 @@
...
 # ++
 module Projects
-  class LifeCycleComponent < ApplicationComponent
+  class PhaseDefinitionComponent < ApplicationComponent
     include OpPrimer::ComponentHelpers
 
     # TODO: This component is currently not in use! It should be a shared component
     # between the Projects::Settings::LifeCycleSteps::StepComponent and the
     # Settings::ProjectLifeCycleStepDefinitions::RowComponent.
     # It should hold the icon, definition name and gate text information.
```
dombesz: I think this comment is not relevant anymore.
```suggestion
```

app/components/projects/settings/life_cycle_steps/step_component.html.erb
```diff
@@ -3,16 +3,14 @@
...
-      end
-      title_container.with_column(pt: 1, classes: "no-wrap") do
-        render(Primer::Beta::Text.new(**gate_text_options)) { gate_info }
-      end
+    step_container.with_column(mr: 2, classes: "min-width-0") do
+      render(
+        Projects::PhaseDefinitionComponent.new(
+          definition,
+          edit_link: User.current.admin?,
+          phase_text_options: { classes: "filter-target-visible-text" }
```
dombesz, line 11: I like how the `"filter-target-visible-text"` is logically separated to be context aware, but I'm unsure if it is required, because we seem to use in all the instances. Do you think we will use this component without the `"filter-target-visible-text"` in the future?

dombesz (approved):
Looks and works nicely, thank you Jens.👍
Just left two comments, regarding an outdated comment and one question.
Feel free to merge.

app/components/projects/settings/life_cycle_steps/step_component.html.erb
ulferts: @dombesz I don't know. I just felt that it is not the responsibility of that component to be ready to be filterable with that controller.

PR #18523, opened 2025-04-02
Feature/59180 automatic scheduling for phases and phase gates

toy:
# Ticket
[OP#59180](https://community.openproject.org/wp/59180)

Part of automatic scheduling, currently only handles updating following phases duration using UI from https://github.com/opf/openproject/pull/18605
<del>It also handles activating/deactivating phase, but it is not used and should be deleted if will not be used</del>

## Screenshots

Note: updating duration was fixed after recording

https://github.com/user-attachments/assets/0e9fe662-6515-41bf-a871-1892ef2f6790

# Merge checklist

- [x] Added/updated tests


app/models/project.rb
```diff
@@ -90,8 +90,7 @@ class Project < ApplicationRecord
   has_many :available_phases,
            -> { visible.eager_load(:definition).order(position: :asc) },
            class_name: "Project::Phase",
-           inverse_of: :project,
-           dependent: :destroy
```
dombesz, line 94: I wonder why is the `dependent: :destroy` is removed here?

app/workers/work_packages/apply_working_days_change_job.rb
```diff
@@ -34,11 +34,11 @@ class WorkPackages::ApplyWorkingDaysChangeJob < ApplicationJob
...
   )
 
-  attr_accessor :previous_working_days, :previous_non_working_days
+  attr_reader :previous_working_days, :previous_non_working_days
 
   def perform(user_id:, previous_working_days:, previous_non_working_days:)
-    self.previous_working_days = previous_working_days
-    self.previous_non_working_days = previous_non_working_days
+    @previous_working_days = previous_working_days
+    @previous_non_working_days = previous_non_working_days
```
dombesz, line 41: Seems like project phase rescheduling is not implemented when changing non working days settings.

db/migrate/20250411124501_add_duration_to_project_phase_journals.rb
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
+
+class AddDurationToProjectPhaseJournals < ActiveRecord::Migration[8.0]
+  def change
+    add_column :project_phase_journals, :duration, :integer, null: true
```
dombesz: I think this can be done in the same migration with adding the project phase duration.

spec/contracts/project_life_cycle_steps/update_contract_spec.rb
```diff
@@ -0,0 +1,147 @@
...
+      let(:date_range) { date - 1..date + 1 }
+      let(:preceding_date_range) { date - 6..date - 5 }
+      let(:following_date_range) { date + 5..date + 6 }
+
+      before do
+        allow(project).to receive(:available_phases).and_return(phases)
+      end
+
+      context "with successive non overlapping dates" do
+        it_behaves_like "contract is valid"
```
dombesz, line 89: From the specification:
> - Start date in only active for the first phase. For phases that have a predecessor the start date is inactive.
>     - If the predecessor has no finish date set yet from which the start date of the currently edited phase can be deduced, the start date is editable.

I believe this would also mean to have a validation that no gaps are allowed between the previous phase's finish date and the current phase's start date, if both dates are set.
Should we validate that start date is the next day of the previous phase's finish date, if the previous phase has a finish date? (Perhaps also validate that the start date is after the previous phase's start date? Not sure about this one.)

spec/services/project_life_cycle_steps/update_service_spec.rb
```diff
@@ -0,0 +1,234 @@
...
+        expect(following).to have_attributes(start_date: date + 2, finish_date: date + 4201, duration: 3000)
+      end
+    end
+
+    context "when activating" do
+      let(:phases) { [phase, following] }
+      let(:phase) { create(:project_phase, project:, date_range: date - 1..date + 1, active: false) }
+      let(:following) { create(:project_phase, project:, date_range: date + 6..date + 8, duration: 3) }
+
+      it "reschedules starting with date after phase finish date" do
```
dombesz: Would be nice to have a test case where the current phase also overlaps the previous phase, probably the current phase's start date should be rescheduled too.

spec/services/project_life_cycle_steps/update_service_spec.rb
```diff
@@ -0,0 +1,234 @@
...
+        expect(service.call(active: true)).to be_success
+
+        expect(following).to have_attributes(start_date: date + 2, finish_date: date + 6, duration: 3)
+      end
+    end
+
+    context "when deactivating" do
+      let(:phases) { [phase, following] }
+      let(:phase) { create(:project_phase, project:, date_range: date - 1..date + 1, active: true) }
+      let(:following) { create(:project_phase, project:, date_range: date + 6..date + 8, duration: 3) }
```
dombesz: Just for the sake of testing I would have a different duration for the following phase to make sure it does not get altered during the rescheduling.

spec/services/project_life_cycle_steps/update_service_spec.rb
```diff
@@ -0,0 +1,234 @@
...
+        expect(following3).to have_attributes(start_date: date + 9, finish_date: date + 9, duration: 1)
+      end
+    end
+
+    context "for following phase without duration" do
+      let(:phases) { [phase, following] }
+      let(:phase) { create(:project_phase, project:, date_range: date..date) }
+      let(:following) { build(:project_phase, project:, date_range: date + 6..date + 8, duration: nil) }
+
+      it "doesn't get rescheduled" do
```
dombesz, line 155: Since the date range is set, it feels inconsistent to not have the duration set. While this is not really possible to achieve via the UI (...and we don't care about data migration at this stage), I would still suggest adding a presence validation on the duration field when the start and finish date are set.

dombesz (commented):
Overall looks good, I left a few questions and suggestions.

spec/models/project/phase_spec.rb
```diff
@@ -107,42 +125,39 @@
...
-    it "returns the correct number of days if start_date and finish_date are the same" do
-      subject.start_date = Time.zone.today
-      subject.finish_date = Time.zone.today
-      expect(subject.working_days_count).to eq(1)
-    end
+        expect { subject.set_calculated_duration }.to change(subject, :duration).from(0).to(20)
+      end
+
+      it "sets duration to nil nil if date range is incomplete" do
+        subject.duration = 0
```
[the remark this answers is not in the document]
toy: Correct nitpick, but wrong suggestion

app/models/project.rb
toy: There is already `dependent: :destroy` on `has_many :phases`, so it is not needed for `project.destroy` to work and will be an extra query. Documentation mentions that it is still relevant for calls like `project.available_phases.delete(phase)`, `project.available_phases = phases` or `project.available_phases.clear`, but I doubt that we will use those and I just added disallowed NULL for references to be certain

app/workers/work_packages/apply_working_days_change_job.rb
toy: Yes it will be done separately

db/migrate/20250411124501_add_duration_to_project_phase_journals.rb
toy: True

spec/services/project_life_cycle_steps/update_service_spec.rb
toy: That would make the contract invalid, so service will not do anything

spec/services/project_life_cycle_steps/update_service_spec.rb
toy: I did originally add validation on duration, but it lead to a problem with mechanics in contract and service:

- if I just add validation to current state, trying to ask the service to set dates on a phase that has no dates will lead to contract being invalid, as model is validation before duration is set

- if I move assigning duration before contract validation, it will be considered an attribute that is provided by the user

- if I add `duration` to `writable_attributes` it will allow overriding attribute, or require double assigning duration (before validation and after assigning provided attributes)

spec/services/project_life_cycle_steps/update_service_spec.rb
toy: Different from the duration of `phase`?

spec/contracts/project_life_cycle_steps/update_contract_spec.rb
toy: It may be easier to skip until finalising interface, as when start date of phase is automatically set to next working day after previous phase and it is possible to modify only finish date, this check will not be needed.
For the second validation - there is already a check that this phase should start after all previous phases and phase can't have finish date be before start date, so only directly adding entries to database allows the condition for validation to be needed

spec/services/project_life_cycle_steps/update_service_spec.rb
dombesz: Not sure why would it make the contract invalid, the `:phase` is inactive before the activation, AFAIK it wouldn't trigger any validation error on the contract.

spec/services/project_life_cycle_steps/update_service_spec.rb
dombesz: Yes.

spec/services/project_life_cycle_steps/update_service_spec.rb
dombesz: I see, at some point in the future we will allow editing the duration field as well, so I think having it in the writable attributes is the way to go. We can still restrict the duration field assignment via the [permitted params](https://github.com/opf/openproject/pull/18523/files#diff-eb5213babf99c1c0cb3130ccaf08ab620683c4a07b261a00c9295f5287f4af4aR294).

toy: I think it is better to allow duration as input when it will be properly handled

spec/services/project_life_cycle_steps/update_service_spec.rb
toy: Contract validation is happening after setting attributes, so phase validate as active, and contract besides checking model validity is checking that this phase dates are after preceding phase dates. This check makes sense only while it is possible to specify both dates, so will be gone after allowing only finish date to be specified for phases that have preceding phase or phases.

dombesz (approved):
As per our synchronous discussion regarding [this thread](https://github.com/opf/openproject/pull/18523#discussion_r2058718430), activation via the update contract is not used anywhere at the moment.

Approving the PR for now, and feel free to address the phase activation rescheduling issue here, or in another PR.

toy:
I've removed ability to activate/deactivate through update service for now and was already working on a service to handle it, but will better do it in a separate PR

PR #18535, opened 2025-04-03
[62595] Fix date buttons not opening input in Safari in date picker

cbliard:
# Ticket

https://community.openproject.org/wp/62595

When no dates are set on a work package and the date field is clicked, the date picker opens in single date mode: one date field (generally finish date) is a focused field and the other date field is a button. Once clicked, it turns into an input field too.

But on Safari, clicking that button has no effect.

## Screenshots

Idiomorph is used to morph the date picker dialog content, except the active element. When clicking the date button in Safari, it does not become the active element because it's actually a link, and for some reason links cannot be active in Safari. What becomes active instead is the closest focusable parent element. It is the whole `panel-wp-datepicker-dialog--content-tab--dates` as it has `tabindex: 0`. As idiomorph does not change the active element, that's why the date fields stayed the same.

To fix this, we make pass `ignoreActiveValue = true` to idiomorph only if the active element is an input. This way the whole date picker form is always updated when no input is focused.

It also fixes similar issues that existed on "Today" links and "Working days only" checkbox too.

# Merge checklist

- [ ] Added/updated tests
  - The ideal would be to run the whole date picker feature test suite using Safari.

### DIFF
--- a/app/components/work_packages/date_picker/date_form_component.html.erb
+++ b/app/components/work_packages/date_picker/date_form_component.html.erb
@@ -29,7 +29,7 @@
     unless @is_milestone
       body.with_column(classes: "wp-datepicker-dialog-date-form--due-date") do
         flex_layout do |due_date|
-          due_date.with_row(classes: container_classes(:due_date), data: { skip_morphing: true }) do
+          due_date.with_row(classes: container_classes(:due_date)) do
             render(
               Primer::ButtonComponent.new(
                 tag: :a,
--- a/frontend/src/stimulus/controllers/dynamic/work-packages/date-picker/preview.controller.ts
+++ b/frontend/src/stimulus/controllers/dynamic/work-packages/date-picker/preview.controller.ts
@@ -124,21 +124,21 @@ export default class PreviewController extends DialogPreviewController {
     }
   }
 
-  private get dueDateField():HTMLInputElement {
-    return document.getElementsByName('work_package[due_date]')[0] as HTMLInputElement;
+  private get dueDateField():HTMLInputElement|undefined {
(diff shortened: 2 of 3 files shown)


bsatarnejad:
Hi @cbliard 
The issue for setting start date is fixed, but the issue about changing Duration is still visible:

> toggling "Non working days" checkbox was not updating duration


https://github.com/user-attachments/assets/22b790cd-0197-4e86-a19f-761166aa17bd

cbliard:
Hi @bsatarnejad,

Thanks for testing. Actually, the problem before was that none of the fields were updated when checking and unchecking "Non-working days". Only the mini calendar was updated.

The fact that duration is kept and due date is recomputed is the expected behavior: https://github.com/opf/openproject/blob/0aa89171071140abb4f1d800ee9ef52661cd3e11/app/services/work_packages/set_attributes_service.rb#L89-L93

I wrote that duration should be updated, but it's because I was testing it while editing the finish date of an automatically scheduled work package so duration should have changed. I have updated the description to be less specific about which field should update.

We'll probably review this behavior so that we can fix [Bug: Toggling non-working days resets the start date in some cases (#62641)](https://community.openproject.org/projects/stream-planning-and-reporting/work_packages/62641/activity?query_id=5758). Anyhow that's part of another ticket.

Could review this bugfix again please?

bsatarnejad:
Hi @cbliard thanks for clarifying. I'll review it 👍

bsatarnejad (approved):
LGTM 👍

cbliard:
Failing test `./spec/features/activities/work_package/activities_spec.rb:539` is passing locally.

Merging.

PR #18543, opened 2025-04-04
Implementation/62608 add hovercard to gates

EinLama:
# Ticket
https://community.openproject.org/wp/62608

Show hover cards on phase gates.

## Screenshots

https://github.com/user-attachments/assets/343d7ca9-f8df-45e4-959d-75b1daaf1d7b

I have created a new hover card component for phase gates. This component is responsible for rendering the view within the hover card. I chose the best directory to place these new files in that I could find. Since we don't have a dedicated folder for phases yet, I made something up.

To make the hover card actually show up, you need to place a trigger in a `data` attribute within the DOM, so that the hover card trigger service can start listening for mouse events, making a turbo request, and finally rendering the hover card component view within a popover. Luckily, all of this was already done, I only had to attach some data attributes to our existing phase component ✌🏻 With that, the hover cards will trigger both on the project overview as well as in the project list table (and all further future places where we display them).

I sneaked a small bugfix into the hover card trigger service, since the event bubbling behaves a bit weird for the svg icon. With the fix in place, displaying the hover card upon hovering over the gate icon works reliably 👏🏻 

As for specs, I unit tested the component itself and added a small feature spec that triggers a hover card by hovering. I tried to keep the feature spec as short as possible, since I think that hover cards are a nice-to-have in most cases.

# Merge checklist

- [x] Added/updated tests


app/components/projects/phases/hover_card_component.html.erb
```diff
@@ -0,0 +1,66 @@
...
+        end
+      end
+    end
+  else
+    render Primer::Beta::Blankslate.new(border: false, narrow: true) do |component|
+      component.with_visual_icon(icon: "x-circle")
+      # Show a generic error message to avoid leaking information
+      component.with_heading(tag: :h3).with_content(I18n.t("http.response.unexpected"))
+    end
+  end
```
dombesz: Since the controller already restricts the scope to active phases, do you think this is still necessary? As I checked, it is impossible to access this branch of the condition via the UI.

app/components/projects/phases/hover_card_component.rb
```diff
@@ -0,0 +1,69 @@
...
+        super
+
+        @phase = phase
+        @gate = gate.to_sym
+      end
+
+      def phase_gate_name
+        case @gate
+        when :start
+          @phase.definition.start_gate? ? @phase.start_gate_name : nil
```
dombesz: Just a minor detail, `@phase.start_gate?` would also work here due to the delegation.

dombesz (approved):
Looks and works well!👍
Just noted a few minor things, feel free to merge.

app/components/projects/phases/hover_card_component.html.erb
toy: I was not sure about keeping it, but after changes to controller I can't currently come up with a reason to do it

PR #18545, opened 2025-04-04
improve performance of WorkPackageCustomField.visible_by_user

ulferts:
# Ticket

N/A

This is a drive by fix.

Improve the performance when fetching the visible work work package custom fields. This is done in various places throughout the application, most notably, when the selects for the work package query are determined. This is done again in various places, e.g. on the work package list page but also other screens as the work package administration and the type form configuration.

When joining, the cross product of projects, types and custom fields was returned from the SQL query. AR then only uses the distinct values to create the WorkPackageCustomField model instances. On my instance with ~500 CustomFields and ~2000 Projects and 12 types, ~130k records are returned which is then reduced to 50 WorkPackageCustomField records. The database therefore does a lot of unnecessary work. In this case, a subquery is more performant.

The query time is reduced form 2-3 seconds to ~100ms.

### DIFF
--- a/app/models/work_package_custom_field.rb
+++ b/app/models/work_package_custom_field.rb
@@ -47,10 +47,15 @@ class WorkPackageCustomField < CustomField
   }
 
   scope :visible_by_user, ->(user) {
-    where(projects: { id: Project.visible(user) })
-      .where(types: { id: Type.enabled_in(Project.visible(user)) })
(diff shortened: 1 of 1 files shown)


klaustopher (approved):

klaustopher:
Would it make sense to even put this in 15.5?

ulferts:
Good idea, @klaustopher.

PR #18547, opened 2025-04-04
Fix remaining unit test failures due to frozen string literals

NobodysNightmare:
This should fix the last known unit test failures occuring due to the usage of
frozen string literal comments. Those were tested in a separate
commit. Broadly enabling frozen string literals will happen in multiple
separate commits after this one and only after feature tests have been fixed
as well.

# References

* https://github.com/opf/openproject/pull/18039

### DIFF
--- a/app/helpers/repositories_helper.rb
+++ b/app/helpers/repositories_helper.rb
@@ -1,3 +1,5 @@
+# frozen_string_literal: true
+
 #-- copyright
 # OpenProject is an open source project management software.
 # Copyright (C) the OpenProject GmbH
@@ -47,7 +49,7 @@ module RepositoriesHelper
 
   def render_properties(properties)
     unless properties.nil? || properties.empty?
-      content = ""
+      content = +""
       properties.keys.sort.each do |property|
         content << content_tag("li", raw("<b>#{h property}</b>: <span>#{h properties[property]}</span>"))
       end
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -1,3 +1,5 @@
+# frozen_string_literal: true
+
 #-- copyright
 # OpenProject is an open source project management software.
 # Copyright (C) the OpenProject GmbH
@@ -185,7 +187,7 @@ class ApplicationController < ActionController::Base
   # Create CSRF issue
   def log_csrf_failure
     message = "CSRF validation error"
-    message << " (No session cookie present)" if openproject_cookie_missing?
+    message += " (No session cookie present)" if openproject_cookie_missing?
 
     op_handle_error message, reference: :csrf_validation_failed
   end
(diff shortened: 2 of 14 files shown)


NobodysNightmare:
All Rubocop offenses affect previously existing code, which I am just touching to fix another rubocop offense.

As far as I can tell, I didn't introduce any new offenses, so that would be good enough for me right now.

app/helpers/repositories_helper.rb
```diff
@@ -47,7 +49,7 @@ def truncate_at_line_break(text, length = 255)
 
   def render_properties(properties)
     unless properties.nil? || properties.empty?
-      content = ""
+      content = +""
```
Kharonus, line 52: 🟡 can't we use immutable strings here, too? 

like `content = ""` and later `content += ...`

Kharonus: same in other examples below.

Kharonus: this seems like a general question 🫡 

- use immutable strings (enforced due to frozen string literals) and use `+=` when amending them to create new strings
- use mutable duplicates of strings to amend them

Kharonus (approved):
Overall I see no blocker of not merging this, but I have this VEEEERY important question about mutability of strings :P

app/helpers/repositories_helper.rb
NobodysNightmare: To be honest: I mostly went through this by gut feeling, not by hard numbers, but I can at least explain the general theory between using `<<` vs `+=`.

The advantage of frozen strings is that the Ruby runtime can know that they are not changing, so it doesn't need to hold them in memory twice, if it has one representation it can safely be used everywhere. Immutability is also great as a programming paradigm, because when you have code like this:

```ruby
my_string = "I am a unicorn! 🦄"
mysterious_method_you_dont_know(my_string)
puts my_string
```

you can still know that you will print a unicorn in the end, because no one could have changed the string to not contain a unicode unicorn.

The disadvantage of frozen strings is that you can't change them. This means if you have code like this:

```ruby
a = "my string"
10.times { |i| a += "|is long" }
```

then Ruby will have to allocate some memory to hold the original string, then on each `+=` operation it will allocate memory to hold the new (concatenated) string and at the same time lose a reference to the old string, which will be a cleanup job for the garbage collector later. Thus in cases where a string is changed _often™_, it's better to change it in-place.

### Conclusion

And now I took that knowledge and tried applying it in useful ways. If anything looked like it would only change the same string once or twice, I didn't bother to make the string mutable, because immutability is great.

I was (hopefully) also careful to not make a string mutable, because some random method that accepted the string would need it to be mutable (the unicorn case above).

On the other hand if there was a clear case of a string being used as a buffer to be extended in many (usually looped) calls, and the mutator was the same code that created the string, then I opted for the mutable option, because it's less GC stress. A good sign for such a buffer-string is that it starts its life as `""` and then expands to something useful afterwards.

NobodysNightmare: I am going ahead with merging this PR. Feel free to raise additional or follow-up questions or concerns if you have some.

PR #18548, opened 2025-04-04
[62641] toggling non working days keeps dates and updates duration

cbliard:
# Ticket

https://community.openproject.org/wp/62641

Previously we tried to keep duration over the dates when updating "Working days only", also known as `ignore_non_working_days`. The behavior was a bit surprising and was inconsistent with the fact that setting a new start date would keep the the finish date and update the duration.

Now it's consistent for any attribute changed in the date picker modal: the dates are kept and the duration is updated, (unless the duration is set by the user).

## Screenshots

Before:

[before.webm](https://github.com/user-attachments/assets/04f4e361-39a2-4bb3-8a6a-934f50aa9a05)

After:

[after.webm](https://github.com/user-attachments/assets/1c1fa894-cc8b-43de-9170-4042d1a8e56c)

There was a condition to explicitly check if "ignore_non_working_days" was changed and change the derived attributes priority. I simply removed the special case.

# Merge checklist

- [x] Added/updated tests

### DIFF
--- a/app/services/work_packages/set_attributes_service.rb
+++ b/app/services/work_packages/set_attributes_service.rb
@@ -92,12 +92,7 @@ class WorkPackages::SetAttributesService < BaseServices::SetAttributes
   # If +ignore_non_working_days+ has been changed, try deriving +due_date+ and
   # +start_date+ before +duration+.
   def derivable_date_attribute_by_others_presence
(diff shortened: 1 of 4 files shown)


bsatarnejad (approved):
👍

PR #18562, opened 2025-04-07
Reorganise components in overviews module

toy:
# Ticket
As part of [OP#59180](https://community.openproject.org/wp/59180)

Flatten structure, rename LifeCycles to Phases, remove duplication in nested modules


modules/overviews/app/components/project_custom_fields/sections/project_custom_fields/show_component.rb
dombesz, line 1: The folders `modules/overviews/app/components/project_life_cycles/sections` are also empty and can be removed.

modules/overviews/app/components/overviews/project_phases/side_panel_component.rb
```diff
@@ -0,0 +1,45 @@
...
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program; if not, write to the Free Software
+# Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+#
+# See COPYRIGHT and LICENSE files for more details.
+#++
+
+module Overviews
```
dombesz, line 31: I kinda like having the module name here, however from the path's perspective, I find it confusing to have `overviews` present twice `modules/overviews/app/components/overviews/project_phases/side_panel_component.rb`.

dombesz (commented):
Just left 2 comments. Overall, I like the idea of namespacing the overview module's components, however the file path naming is a bit confusing, and I'm not entirely convinced if it's better than how it was before.

dombesz (approved):

modules/overviews/app/components/project_custom_fields/sections/project_custom_fields/show_component.rb
toy: If I understood your comment correctly, directories without files don't exist for git, but in some cases it can probably not be deleted automatically in a local checkout

modules/overviews/app/components/overviews/project_phases/side_panel_component.rb
toy: As agreed synchronously, let us try to remove double `overviews` later

PR #18566, opened 2025-04-07
Rework "open storage" action

NobodysNightmare:
The previous implementation was a deeply nested decision tree for different error cases or other unexpected states.

Most of this was necessary, because setting up correct permissions is partially happening asynchronously. Some of the redirects would eventually have rendered a modal that would periodically check whether user permissions appeared in the background.

All of this was replaced by an approach where we immediately (synchronously) ensure that the setup is correct.

# Ticket
https://community.openproject.org/wp/62758

# Merge checklist

- [x] Added/updated tests


NobodysNightmare:
Opening this PR for review already. Relevant tests are passing. The only thing left is that I need to split tests for code that I separated (currently the original unit tests cover the extracted concerns).

modules/storages/app/services/storages/nextcloud_managed_folder_create_service.rb
```diff
@@ -309,27 +210,18 @@ def build_set_permissions_input_data(file_id, user_permissions)
...
     end
 
-    def remote_group_users
-      info "Retrieving users that a part of the #{@storage.group} group"
-      group_users.call(storage: @storage, auth_strategy:, group: @storage.group)
-    end
+    def create_folder = Peripherals::Registry.resolve("nextcloud.commands.create_folder")
 
-    ### Model Scopes
+    def rename_file = Peripherals::Registry.resolve("nextcloud.commands.rename_file")
```
Kharonus, line 215: ❓ why removing the injector and instead defining a method for every command/query used?

modules/storages/app/services/storages/managed_folder_sync_service.rb
Kharonus, line 1: ❓ you extracted logic from the previous managed sync service into two sub services (folder create and folder permissions).

Is there any change in the logic? Did you do anything else but copying the method blocks into the new service and rename some variables? It is hard to check this inside the PR :D

modules/storages/app/workers/storages/automatically_managed_storage_sync_job.rb
```diff
@@ -57,14 +57,7 @@ def self.key(storage) = "sync-#{storage}-#{storage.id}"
...
 
-      sync_result = case storage.short_provider_type
-                    when "nextcloud"
-                      NextcloudManagedFolderSyncService.call(storage)
-                    when "one_drive"
-                      OneDriveManagedFolderSyncService.call(storage)
-                    else
-                      raise "Unknown Storage Type"
-                    end
+      sync_result = ManagedFolderSyncService.call(storage)
```
Kharonus, line 60: 🟢 nice

modules/storages/app/models/storages/project_storage.rb
```diff
@@ -97,7 +97,7 @@ def open(user)
 
     def open_with_connection_ensured
```
Kharonus, line 101: 🟡 you removed this from the menu. is this still used elsewhere?

modules/storages/app/models/storages/project_storage.rb
```diff
@@ -106,17 +106,12 @@ def open_with_connection_ensured
       )
     end
 
-    private
-
     def open_project_storage_url
-      OpenProject::StaticRouting::StaticRouter
-        .new
-        .url_helpers.open_project_storage_url(host: Setting.host_name,
-                                              protocol: "https",
```
Kharonus, line 115: 🟡 why is protocol and host no longer needed?

modules/storages/app/services/storages/nextcloud_managed_folder_permissions_service.rb
```diff
@@ -0,0 +1,187 @@
...
+
+    def self.i18n_key = "nextcloud_sync_service"
+
+    class << self
+      def call(storage:, project_storages: nil)
+        new(storage:, project_storages:).call
+      end
+    end
+
+    def initialize(storage:, project_storages: nil)
```
Kharonus: 🟡 I just realized, that this is actually a relation, not an array. makes me wonder, if we should show this somehow, maybe in the name as we do not have method signatures?

I was stumbling across this, as the controller, who already has the fetched object, does this stunt:

```ruby
  def project_storage_scope
    Storages::ProjectStorage.where(id: @project_storage.id)
  end
```

so, in case of the controller action it leads to the project storage being loaded twice.

Kharonus (commented):
I have some open points I'd like to handle before giving this a green light

modules/storages/app/services/storages/nextcloud_managed_folder_create_service.rb
NobodysNightmare: I kept it at first, but in the end I had a circular dependency that @mereghost spotted as such. I.e. it was a problem that to use the injector you needed to require the registry, but to load the registry you needed to load the create service, which needed to load the injector, which needed to load the registry, which needed to load the create service, which needed to load the injector...

@mereghost shrugged the injector off as an idea/experiment that we could also remove if it caused problems.

modules/storages/app/services/storages/managed_folder_sync_service.rb
NobodysNightmare: **There are no intentional changes.**

In the current state of the PR you can still see that intention, because the existing specs (e.g. `modules/storages/spec/services/storages/nextcloud_managed_folder_sync_service_spec.rb`) are essentially unchanged. I.e. I was purposefully reusing the old specs to make sure that my changes were not changing anything.

That said, right now I am splitting the tests to become unit tests again and I already realized that the coverage of the existing tests was mostly there for the `create` service I extracted, but it was near zero for the `permissions` service I extracted :shrug:

modules/storages/app/models/storages/project_storage.rb
NobodysNightmare: The API exposes a link `openWithConnectionEnsured`. I am not sure if **we** use that API link, but since it's in the API, I didn't want to remove it in any case.

modules/storages/app/models/storages/project_storage.rb
NobodysNightmare: `OpenProject::StaticRouting` takes care of these already, so links to "us" are possible by default.

While I would not recommend anyone to run a production OP instance without HTTPS, it's weird that this is then the only link in the application where we wouldn't _detect_ the used protocol, but instead enforce/assume that it must be HTTPS.

**tl;dr:** Those values were mostly not needed or overly specific.

modules/storages/app/services/storages/nextcloud_managed_folder_permissions_service.rb
NobodysNightmare: Yeah, that's a good point. The thing was called `something_something_scope` beforehand, when it was still an instance method. Let me check whether I want to clarify the name or whether I want to allow passing this as an array.

modules/storages/app/models/storages/project_storage.rb
Kharonus: I'd introduce a deprecation round for that link in the project storage resource.

meaning, adding the deprecation warning to the API spec, waiting n days, then removing it from code and spec

modules/storages/app/services/storages/nextcloud_managed_folder_create_service.rb
Kharonus: I see, that was the discussion about circular dependencies. ack.

modules/storages/app/services/storages/managed_folder_sync_service.rb
Kharonus: fine for me with that explanation

modules/storages/app/models/storages/project_storage.rb
NobodysNightmare: I think I just found one more place using this:

* https://github.com/opf/openproject/blob/dev/frontend/src/app/shared/components/storages/storage/storage.component.ts#L219
* (used in) https://github.com/opf/openproject/blob/dev/frontend/src/app/shared/components/storages/storage/storage.component.html#L13

Let me check which part of the UI this exactly belongs to...

NobodysNightmare: After some time of confusing myself, I can confirm that right now this endpoint is still in use to render at least one link in our UI.

I'd not yet deprecate it, even though we could discuss whether the current API structure still makes sense.

modules/storages/app/services/storages/nextcloud_managed_folder_permissions_service.rb
NobodysNightmare: I renamed it back to `project_storages_scope`. A lot of places depend on this being an active record relation.

modules/storages/app/services/storages/nextcloud_managed_folder_permissions_service.rb
```diff
@@ -0,0 +1,189 @@
...
+    end
+
+    def apply_permissions_to_folders
+      info "Setting permissions to project folders"
+      remote_admins = admin_remote_identities.pluck(:origin_user_id)
+
+      @project_storages
+        .active
+        .automatic
+        .where.not(project_folder_id: nil)
```
mereghost: WDYT about making this into a `ProjectStorage` scope?

modules/storages/app/services/storages/one_drive_managed_folder_permissions_service.rb
```diff
@@ -0,0 +1,128 @@
...
+      end
+    end
+
+    private
+
+    # rubocop:disable Metrics/AbcSize
+    def apply_permission_to_folders
+      info "Setting permissions to project folders"
+      @project_storages.includes(:project)
+                                   .where.not(project_folder_id: nil)
```
mereghost: Here we miss the automatic and enabled checks.

mereghost (commented):
Thanks for taking care of a thing I planned in early 2024. XD

modules/storages/app/services/storages/one_drive_managed_folder_permissions_service.rb
NobodysNightmare: We don't, because I sneaked this into the initializer:

```ruby
@project_storages = (project_storages_scope || @storage.project_storages).active.automatic
```

Note how the parenthesis do the work for us here. I mostly switched to this pattern after realizing that some of the services use the scope more than once. Though I think my original implementation still applied this base-scope lazily.

modules/storages/app/services/storages/nextcloud_managed_folder_permissions_service.rb
NobodysNightmare: Yeah, I think we use this in a few places don't we?

Let me find a suitable name, I'll be back in five hours!

NobodysNightmare: I am starting to wonder whether these checks are working as intended. I remember from a discussion on Matrix a while ago that missing project folders are indicated by an _empty_ string, not by nil.

I can also confirm that experimentally in my local database. So should I change the condition to exclude `""`as well?

NobodysNightmare: P.S.: See my latest commit on how I approached the new scope

modules/storages/app/models/storages/project_storage.rb
```diff
@@ -97,7 +98,7 @@ def open(user)
 
     def open_with_connection_ensured
       return unless storage.configured?
-      return open_project_storage_url if storage.authenticate_via_idp?
+      return open_project_storage_url if storage.authenticate_via_idp? # TODO: use auth method selector
```
Kharonus: 🟡 there is still a new `TODO` in here. Is this fine or just forgotten?

modules/storages/app/models/storages/project_storage.rb
Kharonus: ⚠️ can you please add a small code maintenance? In it we should deprecate the link, and move the mentioned link you mentioned above to your new `open` method

Kharonus (approved):
LGTM, but please check the remaining two comments ;)

modules/storages/app/models/storages/project_storage.rb
NobodysNightmare: It's a new TODO, but the problem is old. We introduced the auth method selector before this PR, but after the code I annotated here. It would be the better tool to use to decide whether or not we want to skip connection ensuring.

I guess I am going to address this TODO in this PR here (before we will later delete the method as suggested by you ^^)

NobodysNightmare: I retract that proposal of mine... the auth method selector requires that we pass in a user (because it's making a user-specific decision). That's not possible here. I am going to remove the TODO comment and we will simply delete this method at some point...

NobodysNightmare:
> ⚠️ can you please add a small code maintenance? In it we should deprecate the link, and move the mentioned link you mentioned above to your new open method

https://community.openproject.org/wp/63691

PR #18568, opened 2025-04-07
[#63358] Wiki ToC links are not working on certain browsers 

top4ek:
# Ticket
[#63358](https://community.openproject.org/work_packages/63358) 

Somehow(I think it's something related to turbo), links for heading started with non ascii symbols aren't working. Tried different browsers with japanese and russian "runes".

Setting link with target="_self" fixes problem. ToC links are still on the same page.


HDinger (changes requested):
Hi @top4ek 

thanks for looking into this! 🙇 For me, this does not fix the linked issue. I think this is not (only) a turbo issue. When you hover over the headline you already see that the generated link is not correct, so I assume, that the reason is rahter in the link generation.

Bildschirmfoto 2025-08-08 um 09 52 44

top4ek:
The other way is to sluggerize these ids, but I've poked it some time but wasn't successful. Will take another look.

PR #18574, opened 2025-04-08
Implementation/62981 show and edit linked phases on work packages

ulferts:
# Ticket

https://community.openproject.org/wp/62981

## Screenshots

image

* [x] Work pages receive a new attribute: The phase they belong to.
    
    * The name of the attribute is &quot;Project phase&quot;.
        
    * A work package can be assigned to a single phase. There is no multi assignment.
    * In case no project phase is active in the project, the "Project phase" field is hidden.
        
* [x] The attribute can be administrated same as every other attribute (e.g. status, assignee, version).
    
    *   It can be moved in and out of an attribute group.
        
    *   Existing installations are not migrated to have the attribute activated for their existing types.
        
* [x]  The attribute&#39;s value is displayed including the colored type indicator prefix in the split/full.
  * [x] The autocompleter displays the colored icon as well
  * This is done using the work around of relying on the project phase's name which is the same as the definition's name which is guaranteed to be unique.
        
* [x]  The attribute can be changed via an ng-select.
            
* [x]  In case the associated phase is later disabled in the project administration, the value is no longer displayed
    
* [x] The API v3 has a dedicated resource for phases
    
    *   GET endpoint for referencing a selected phase
        
    *   GET endpoint for selecting all available (active) phases of the project when selecting the value of the attribute.
        
* [x]  The new functionality is hidden behind the feature flag

Not part of the original scope

* [x] Fix permissions on copying work packages. Before, the user had to have the move_work_packages permission as well in order to access the form for duplicating work packages to a different project. The autocompleter for selecting target projects also relied on the move permission. Both are changed without checking if there are any other places the code relies only on the move permission.

To consider:
* [x] Moving a work package between projects where
  * [x] the currently linked phase is not active 
  * [x] the currently linked phase is active 
* [x] Copying a work package to another project where 
  * [x] the currently linked phase is not active 
  * [x] the currently linked phase is active 
* [x] Performance of the work package list - eager loading


app/helpers/colors_helper.rb
```diff
@@ -67,6 +67,26 @@ def resources_scope_color_css(name, scope, inline_foreground: false)
...
+  #
+  # That way the frontend does not have to load the definitions to get the color.
+  # The = signs at the end of the base64 string are replaced with _ to make it a valid class name.
+  # This needs to be kept in sync with the ProjectPhaseDisplayField#phaseIcon method in the front end.
+  def project_phase_color_css
+    Project::PhaseDefinition.includes(:color).find_each do |definition|
+      resource_color_css("project_phase_definition", definition, inline_foreground: true)
+
+      set_foreground_colors_for(
+        class_name: ".#{hl_inline_class('project_phase_definition', Base64.strict_encode64(definition.name).tr('=', '_'))}",
```
EinLama, line 84: That's a really smart solution for providing a unique color per phase definition name, without having to load a definition from the frontend. Very nice!

EinLama (approved):
Very nice, well done! 👏🏻

PR #18575, opened 2025-04-08
[62834] Display error when clicking date field of a deleted work package

cbliard:
# Ticket

https://community.openproject.org/wp/62834

When a work package is deleted but still visible in a page (because not refreshed yet), clicking on the date field to open the date picker fails badly with a 500.

Displaying a message instead and cancelling the edition like if the field was readonly or something similar

## Screenshots

First try to fix some eslint issues on the incriminating files.
Then there is a mechanism to cancel edition if there are some errors when loading the update form, but this mechanism is triggered too early in the process: it's ok when the update form is not loaded yet because the network request adds some delay and the edit fields have some time to be displayed, but when the update form is already loaded, it fails immediately and tries to close the edit fields but they are not open yet.

I try to add a synchronization so that the edit fields are closed only when they are displayed. Best would be to prevent displaying if there is an error.

# Merge checklist

- [ ] Added/updated tests

### DIFF
--- a/frontend/src/app/features/hal/schemas/schema-proxy.ts
+++ b/frontend/src/app/features/hal/schemas/schema-proxy.ts
@@ -39,8 +39,10 @@ export interface ISchemaProxy extends SchemaResource {
 }
 
 export class SchemaProxy implements ProxyHandler<SchemaResource> {
-  constructor(protected schema:SchemaResource,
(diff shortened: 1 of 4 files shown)


cbliard:
Closing in favor of #20099.

PR #18580, opened 2025-04-08
[#62103] health status report for storages

Kharonus:
# Ticket
[OP#62103](https://community.openproject.org/work_packages/62103)

- introduce sophisticated health status report

- show health status report on separate page
- moved ampf sync status in sidepanel to next section (WIP)
- show summary in sidepanel of storage page
- regenerate health status reports on user request
- cache health status reports for storages


config/locales/en.yml
```diff
@@ -1015,7 +1015,6 @@ en:
       project/phase:
         date_range: "Date range"
         definition: "Definition"
-        date_range: "Date range"
```
NobodysNightmare, line 1018: :heart: Nice catch!

NobodysNightmare:
Not sure what I am doing wrong, but the UI is unusable for me right now. It never shows me a report:

[Screencast from 2025-04-10 09-04-46.webm](https://github.com/user-attachments/assets/9903de83-3fa0-48fb-9580-a92f43d7bd21)

I don't see obvious errors in the server logs or the JS console.

edit: I just checked one more thing. The turbo response I get when clicking the button is the "no reports" page, so the turbo refresh is probably working, but no report is coming in.

edit 2: We noticed that when this was caused by the cache being disabled. We seem to agree that **at least clicking the button on the page should work without a cache**.

modules/storages/app/components/storages/admin/health/check_result_component.html.erb
```diff
@@ -0,0 +1,62 @@
...
+          end
+          text.with_column(ml: 2) do
+            render(Primer::Beta::Text.new(font_size: :small, color: data[:status_color])) { data[:status_text] }
+          end
+        end
+
+        if data[:error_text].present?
+          row.with_column do
+            render(Primer::Beta::Link.new(href: data[:docs_href])) do |link|
+              link.with_trailing_visual_icon(icon: :"link-external")
```
NobodysNightmare, line 46: :question: To be confirmed: Does this link open in a new tab?

As a user I'd expect that, because of the `link-external` icon and because I am opening docs that I need alongside the current page.

modules/storages/app/components/storages/admin/health/check_result_component.rb
```diff
@@ -0,0 +1,82 @@
...
+
+        private
+
+        def data
+          @data ||= {
+            text: I18n.t("storages.health.checks.#{@group}.#{model.key}"),
+            status_color:,
+            status_text:,
+            error_text: model.message,
+            docs_href: ::OpenProject::Static::Links[:storage_docs][:health_status][:href]
```
NobodysNightmare: Does it make sense to improve the `Static::Links.url_for` method to accept `*args`?

```ruby
        def url_for(*items)
          links.dig(*items, :href)
        end
```

My main motivation would be to keep as many accesses as possible guarded by a common accessor method. E.g. if we decided to add additional checks to docs (say: raising an error on missing links during specs), there would be a single place to achieve this.

modules/storages/app/components/storages/admin/health/health_report_component.html.erb
```diff
@@ -0,0 +1,92 @@
...
+
+<%=
+  component_wrapper do
+    render(Primer::Alpha::Layout.new(stacking_breakpoint: :lg)) do |page|
+      page.with_main do
+        if @report.nil?
+          render(Primer::Beta::Blankslate.new(border: true)) do |placeholder|
+            placeholder.with_visual_icon(icon: :meter)
+            placeholder.with_heading(tag: :h3) { data[:heading] }
+            placeholder.with_description { data[:description] }
```
NobodysNightmare: Wouldn't it be simpler to just insert the static "no data" translations here instead of putting them into the `data` hash?

```
            placeholder.with_heading(tag: :h3) { I18n.t("storages.health.no_report") }
            placeholder.with_description { I18n.t("storages.health.no_report_description") }
```

This would also mean that `data` would always have the same keys and not two different layouts depending on whether a report is present or not.

modules/storages/app/components/storages/admin/health/health_report_component.rb
```diff
@@ -0,0 +1,107 @@
...
+            }
+          else
+            description = if @report.unhealthy?
+                            I18n.t("storages.health.summary.failure")
+                          else
+                            I18n.t("storages.health.summary.warning")
+                          end
+            {
+              summary: summary_with_icon(checks_tally),
+              description:
```
NobodysNightmare: :yellow_circle: Would you agree that the only thing changing in the big `if-else` around is the `description` key of the returned hash, while the `summary` is always the same? (because we extracted that into the `summary_with_icon` method)

I.e. I believe that the code could be simplied like

```ruby
description = if @report.healthy?
                I18n.t("storages.health.summary.success")
              elsif @report.unhealthy?
                I18n.t("storages.health.summary.failure")
              else
                I18n.t("storages.health.summary.warning")
              end

{
  summary: summary_with_icon(checks_tally),
  description:
}
```

modules/storages/app/components/storages/admin/side_panel/health_notifications_component.rb
```diff
@@ -55,6 +57,31 @@ def notification_status
...
+      def formatted_health_reason
+        identifier = @storage.health_reason_identifier.tr("_", " ").strip
+        description = @storage.health_reason_description
+
+        if description.present?
+          identifier.capitalize!
+          identifier << ": #{description}"
+        end
+
+        "#{identifier} #{I18n.t('storages.health.since', datetime: helpers.format_time(@storage.health_changed_at))}"
```
NobodysNightmare, line 83: Two notes:

:red_circle: We should definitely make the `identifier` part of the translation, e.g.

```suggestion
        I18n.t("storages.health.since", identifier:, datetime: helpers.format_time(@storage.health_changed_at))
```

Otherwise we enforce the "english ordering" onto all other languages, regardless of whether it makes sense there. E.g. imagine in a language it must be written as "Since DATE ERROR_MESSAGE", this is only possible by moving the error message into the translation.

:yellow_circle: To avoid confusion, I'd actually change the sentence.

```diff
- Not found: Outbound request destination not found since 12/07/2023 03:45 PM
+ Not found: Outbound request destination not found. Storage unhealthy since 12/07/2023 03:45 PM
```

The reason for that is, that we only know that the storage is unhealthy since a given date, not that it's the same error occuring all the time.

modules/storages/app/components/storages/admin/side_panel/validation_result_component.rb
```diff
@@ -31,29 +31,42 @@
...
-          when :healthy
-            { scheme: :success, label: I18n.t("storages.health.label_healthy") }
-          when :warning
-            { scheme: :attention, label: I18n.t("storages.health.label_warning") }
-          when :error
-            { scheme: :danger, label: I18n.t("storages.health.label_error") }
+        def summary_header
+          tally = @result.tally
+          failure = tally[:failure] || 0
+          warning = tally[:warning] || 0
```
NobodysNightmare: :green_circle: I think this is the second time I see us building these `|| 0` fallbacks for a `tally` result. WDYT about changing our (custom-built anyways) `tally` implementations to initialize their hashes with `Hash.new(0)`?

modules/storages/app/components/storages/admin/side_panel/validation_result_component.rb
```diff
@@ -31,29 +31,42 @@
...
         end
+
+        def summary_description
+          text = if @result.healthy?
+                   I18n.t("storages.health.summary.success")
+                 elsif @result.unhealthy?
+                   I18n.t("storages.health.summary.failure")
+                 else
+                   I18n.t("storages.health.summary.warning")
+                 end
```
NobodysNightmare, line 72: :green_circle: Counting a code suggestion I made earlier, this is the second time we use exactly this `if-else` chain, including their bodies.

Makes me wonder whether it would be useful to extract this into a `HealthHelper` or similar.

modules/storages/app/components/storages/admin/side_panel/validation_result_component.rb
```diff
@@ -31,29 +31,42 @@
...
+
+          "#{text} #{I18n.t('storages.health.checked', datetime: helpers.format_time(@result.latest_timestamp))}"
```
NobodysNightmare, line 74: :red_circle: Same as for the other translation composed this way: Languages have different needs on ordering their sentences. Additionally I'd rename the input `text` to help translators understand the context better:

```suggestion
          I18n.t("storages.health.checked", health_check_text:, datetime: helpers.format_time(@result.latest_timestamp))
```

modules/storages/app/controllers/storages/admin/health_status_controller.rb
```diff
@@ -62,6 +69,24 @@ def find_model_object(object_id = :storage_id)
...
+
+        report
+      end
+
+      def validator
+        @validator ||= case @storage.provider_type
+                       when ::Storages::Storage::PROVIDER_TYPE_NEXTCLOUD
+                         Peripherals::ConnectionValidators::NextcloudValidator.new(storage: @storage)
+                       when ::Storages::Storage::PROVIDER_TYPE_ONE_DRIVE
+                         raise "Unsupported provider type: #{@storage.provider_type}"
```
NobodysNightmare: :question: This PR is supposed to be merged after the one from @mereghost and with OneDrive support from the start, right?

The way it currently looks would mean that we completely drop support for OneDrive until a future PR, which I'd avoid. Otherwise we'd be merging something that **must be changed** before another release is possible.

modules/storages/config/locales/en.yml
```diff
@@ -231,17 +231,44 @@ en:
       storage_oauth: Nextcloud OAuth
       storage_provider: Storage provider
     health:
+      actions:
+        run_checks: Run tests now
+        rerun_checks: Re-run all checks
```
NobodysNightmare, line 235: :yellow_circle: Translation inconsistency: Do we run "tests" or "checks"?

Currently we run tests, but we rerun checks.

edit: Looking further down, I think it should be "checks" consistently, which also happens to match the translation key :tada:

modules/storages/config/locales/en.yml
```diff
@@ -251,22 +278,31 @@ en:
...
         userless_access_denied: The configured app password is invalid.
       label_error: Error
+      label_failed: Failed
       label_healthy: Healthy
+      label_passed: Passed
       label_pending: Pending
+      label_skipped: Skipped
       label_warning: Warning
+      no_report: No report available
+      no_report_description: "Run the tests now for a full health status report for this file storage."
```
NobodysNightmare: Translation consistency:

```suggestion
      no_report_description: "Run the checks now for a full health status report for this file storage."
```

NobodysNightmare (changes requested):
Manual testing was not possible for me. Thus my UI feedback is still limited at this point.

NobodysNightmare:
:red_circle: Translations for SSO-configured storage seem to be missing:

image

mereghost:
> We noticed that when this was caused by the cache being disabled. We seem to agree that at least clicking the button on the page should work without a cache.

Caching is enabled by default in production, that's only an issue in development.

modules/storages/app/controllers/storages/admin/health_status_controller.rb
mereghost: We will probably have another one hooking both together, or update this one to join both. We usually go this way, so we don't block each other.

modules/storages/config/locales/en.yml
```diff
@@ -231,17 +231,44 @@ en:
...
+        authentication:
+          existing_token: "User token"
+          header: "Authentication"
+          user_bound_request: "Sample request for user"
+        ampf_configuration:
+          files_request: "Check for file storage content"
+          group_folder_contents: "Group folder content"
+          group_folder_presence: "Group folder"
+          header: "Automatically managed project folder configuration"
+          userless_access: "Sample request without a user"
```
NobodysNightmare: Without knowing what the checks exactly do, it's pretty hard for me to know what these checks were testing:

image

E:g. what's the difference between "file storage content" and "group folder content"? And what does "group folder" do?

After double checking the implementation, these are my translation proposals:

```yml
        ampf_configuration:
          files_request: "Fetching group folder files"
          group_folder_contents: "Group folder has expected content"
          group_folder_presence: "Group folder exists"
          header: "Automatically managed project folder configuration"
          userless_access: "Files request authentication"
```

One challenge in proper wording is definitely, that we run only one check and then derive all our results from there. E.g. it's confusing that "fetching group folder files" can be successful, when we were not able to fetch the group folder files...

modules/storages/app/components/storages/admin/side_panel/validation_result_component.rb
mereghost: We could also pattern match for this use case.

```ruby
case @result.tally
in { failure: 1.. } then failed_stuff
in { warning: 1.. } then achtung_dinge
else 
  sucesso
end
```

modules/storages/app/components/storages/admin/side_panel/validation_result_component.rb
mereghost: In this case, I think it is fine to leave as is. This "label" will go at the top of the page listing the last status and when it was run. ("Success. Last checked on: 1920-02-30 25:61")

It isn't as much of a sentence than 2 values together.

modules/storages/config/locales/en.yml
mereghost: That's on the headers, which will be challenging to get right, but I think the level of granularity is correct.

1. files_request -> Communication with NC
2. userless_access -> OpenProject/NC Server-to-Server authorization
3. group_folder_presence -> OpenProject Group Folder exists
4. group_folder_contents -> Group Folder Contents

modules/storages/config/locales/en.yml
```diff
@@ -251,22 +278,31 @@ en:
         oidc_cant_refresh_token: There was an error while trying to check your access to the storage. Please check the server logs for further information.
```
NobodysNightmare, line 289: Proposal for line 275 (obviously out of original scope for this PR, but nice to have):

```diff
-oauth_token_missing: OpenProject cannot test the user level communication with Nextcloud as there's no token for the current user.
+oauth_token_missing: OpenProject cannot test the user level communication with Nextcloud as the current user did not yet link their Nextcloud account.
```

The idea is, that we formulate the message in a way that's understandable/actionable for the user performing the check.

I am a bit sad that we just call it "Login to Nextcloud" in the rest of the UI, because it's not a good hint to tell a user to "Login to Nextcloud" (they can do that by opening a new Browser tab). So I just made the term "link" up here.

modules/storages/app/components/storages/admin/health/check_result_component.html.erb
NobodysNightmare: We confirmed that it does. But @Kharonus was surprised that this is the case.

Feel free to find out why everything behaves as it should, but otherwise resolve :)

modules/storages/app/components/storages/admin/side_panel/validation_result_component.rb
NobodysNightmare: True. It does even translate to two sentences. In that case it's fine, though maybe arguable whether the component should then compose the two things together in a string, rather than the erb rendering it like that.

But yeah, that comment is clearly not a :red_circle: anymore :)

modules/storages/app/components/storages/admin/health/check_result_component.html.erb
Kharonus: hmm, you dig some really strange up. I actually have no explanation, why it does open the link in a new tab. 

If I'd change the href to some internal link, like `/admin`, it opens in the same tab. I have no idea and need to investigate.

modules/storages/app/controllers/storages/admin/health_status_controller.rb
NobodysNightmare: > or update this one to join both

I briefly discussed with @Kharonus that I'd prefer if we updated this PR to join then. My reasoning was that we should never merge a known-broken state to `dev` and without the change, we'd know that we are breaking OneDrive.

modules/storages/app/components/storages/admin/health/check_result_component.rb
Kharonus: Probably yes, it would make sense. I did not investigate into it, do you think it is worthwile to take a shot at that within this PR?

modules/storages/app/components/storages/admin/health/health_report_component.html.erb
Kharonus: Yes, I agree, let me change it.

modules/storages/app/components/storages/admin/health/health_report_component.rb
Kharonus: true, changed

modules/storages/app/components/storages/admin/side_panel/validation_result_component.rb
Kharonus: I like those compact hashes, so I'd go with Marcellos suggestion of improving the caller in how it handles stuff. lemme have another look.

edit: yeah, pattern matching makes this code actually very nice

NobodysNightmare:
:red_circle: Open question from a discussion with Eric: What kind of cache do we have in (all the different) production?

If we in a relevant amount of cases have a simple in-process cache, it would mean that deployments with multiple workers (different machines or simply different puma processes) might show "no results" even after the user performed a test. And then reloading multiple times would give different results again.

modules/storages/app/components/storages/admin/side_panel/validation_result_component.rb
Kharonus: on those two occurrences I see not yet the need to deduplicate anything. Also, the display texts are only "loosly" coupled, not by logic. Product people could anytime decide to resolve different texts in each component.

NobodysNightmare:
> What kind of cache do we have in (all the different) production?

I quickly confirmed with @machisuji that we should have a memcached - i.e. a shared cache - in all the relevant places. This means the implementation in this PR should at least not make customers unhappy :)

modules/storages/app/components/storages/admin/health/check_result_component.rb
NobodysNightmare: I'd include it. The change should be minimal and it's backwards compatible.

I was short before introducing it myself, but then we put the corresponding WP on hold.

modules/storages/app/components/storages/admin/side_panel/validation_result_component.rb
Kharonus: I agree that it might be weird to have those concatenated. But they ARE two different locales, just put into the same \<p\>.

modules/storages/config/locales/en.yml
Kharonus: yeah, I stupidly took over the display texts from the wireframes . Will change to `checks`.

modules/storages/config/locales/en.yml
Kharonus: no big deal, I changed the text

modules/storages/config/locales/en.yml
Kharonus: I tried to combine some of those proposals and come up with some "good enough" results

Kharonus:
> 🔴 Translations for SSO-configured storage seem to be missing:

haha, true :D fully overlooked by me

modules/storages/app/components/storages/admin/health/check_result_component.rb
Kharonus: done

modules/storages/app/components/storages/admin/health/check_result_component.html.erb
Kharonus: there is indeed a middleware, a global click listener doing that magic:

https://github.com/opf/openproject/blob/dev/frontend/src/app/core/setup/globals/global-listeners/link-hijacking.ts

mereghost:
I18n failure is probably related to the dynamic resolution of keys. You may need to edit the `config/i18n-tasks.yml` and update the pattern I added.

Kharonus:
okay, so now there is only a flaky feature spec left :D

mereghost (approved):

PR #18582, opened 2025-04-08
Implements the multi-state validators for OneDrive

mereghost:
### Related Work Packages: [OP#62749](https://community.openproject.org/work_packages/62749), [OP#62750](https://community.openproject.org/work_packages/62750), [OP#62751](https://community.openproject.org/work_packages/62751)

Gives OneDrive the same treatment we gave Nextcloud

- [x] Base Group
- [x] Authentication Group
- [x] AMPF Group
- [x] Main Validator
- [x] Hook into the Registry

### DIFF
--- a/config/i18n-tasks.yml
+++ b/config/i18n-tasks.yml
@@ -154,6 +154,7 @@ ignore_unused:
   - '*.permission_header_explanation'
   - 'storages.upsale.*'
   - 'services.*'
+  - 'storages.health.connection_validation.one_drive.*'
 
 ## Exclude these keys from the `i18n-tasks eq-base' report:
 # ignore_eq_base:
--- a/modules/storages/app/common/storages/peripherals/connection_validators/one_drive/ampf_configuration_validator.rb
+++ b/modules/storages/app/common/storages/peripherals/connection_validators/one_drive/ampf_configuration_validator.rb
@@ -74,8 +74,12 @@ module Storages
           def create_folder
             Registry["one_drive.commands.create_folder"]
               .call(storage: @storage, auth_strategy:, folder_name: TEST_FOLDER_NAME, parent_location: ParentFolder.root)
-              .on_failure { fail_check(:client_folder_creation, message("one_drive.client_write_permission_missing")) }
               .on_success { pass_check(:client_folder_creation) }
+              .on_failure do
+              reason = it.result == :already_exists ? :existing_test_folder : :client_write_permission_missing
+
+              fail_check(:client_folder_creation, message("one_drive.#{reason}", folder_name: TEST_FOLDER_NAME))
+            end
           end
 
           def log_extraneous_files(unexpected_files)
--- a/modules/storages/app/common/storages/peripherals/connection_validators/one_drive/ampf_configuration_validator.rb
+++ b/modules/storages/app/common/storages/peripherals/connection_validators/one_drive/ampf_configuration_validator.rb
@@ -1,0 +1,100 @@
+# frozen_string_literal: true
+
+#-- copyright
+# OpenProject is an open source project management software.
+# Copyright (C) the OpenProject GmbH
+#
+# This program is free software; you can redistribute it and/or
+# modify it under the terms of the GNU General Public License version 3.
+#
+# OpenProject is a fork of ChiliProject, which is a fork of Redmine. The copyright follows:
+# Copyright (C) 2006-2013 Jean-Philippe Lang
+# Copyright (C) 2010-2013 the ChiliProject Team
+#
+# This program is free software; you can redistribute it and/or
+# modify it under the terms of the GNU General Public License
+# as published by the Free Software Foundation; either version 2
+# of the License, or (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program; if not, write to the Free Software
+# Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+#
+# See COPYRIGHT and LICENSE files for more details.
+#++
+
+module Storages
+  module Peripherals
+    module ConnectionValidators
+      module OneDrive
+        class AmpfConfigurationValidator < BaseValidatorGroup
+          TEST_FOLDER_NAME = "ConnectionValidatorFolder"
+
+          private
+
+          def validate
+            register_checks :client_folder_creation, :client_folder_removal, :drive_contents
+
+            client_permissions
+            unexpected_content
+          end
+
+          def unexpected_content
+            unexpected_files = files_query.on_failure { fail_check(:drive_content, message(:unknown_error)) }
+                                          .result.files.reject { managed_project_folder_ids.include?(it.id) }
+
+            if unexpected_files.empty?
+              pass_check(:drive_contents)
+            else
+              log_extraneous_files(unexpected_files)
+              warn_check(:drive_contents, message("one_drive.unexpected_content"))
+            end
+          end
+
+          # Testing setting permissions and checking permission inheritance would be great
+          # but there are some challenges to it. We need to figure out a good way to go about this
+          # 2025-04-08 @mereghost
+          def client_permissions
+            folder = create_folder.result
+            delete_folder(folder)
+          end
+
+          def delete_folder(folder)
+            Registry["one_drive.commands.delete_folder"]
+              .call(storage: @storage, auth_strategy:, location: folder.id)
+              .on_failure { fail_check(:client_folder_removal, message("one_drive.client_cant_delete_folder")) }
+              .on_success { pass_check(:client_folder_removal) }
+          end
+
+          def create_folder
+            Registry["one_drive.commands.create_folder"]
+              .call(storage: @storage, auth_strategy:, folder_name: TEST_FOLDER_NAME, parent_location: ParentFolder.root)
+              .on_failure { fail_check(:client_folder_creation, message("one_drive.client_write_permission_missing")) }
+              .on_success { pass_check(:client_folder_creation) }
+          end
+
+          def log_extraneous_files(unexpected_files)
+            file_representation = unexpected_files.map do |file|
+              "Name: #{file.name}, ID: #{file.id}, Location: #{file.location}"
+            end
+
+            warn "Unexpected files/folder found in group folder:\n\t#{file_representation.join("\n\t")}"
+          end
+
+          def managed_project_folder_ids
+            @managed_project_folder_ids ||= ProjectStorage.automatic.where(storage: @storage)
+                                                          .pluck(:project_folder_id).to_set
+          end
+
+          def files_query = Registry["one_drive.queries.files"].call(storage: @storage, auth_strategy:, folder: ParentFolder.root)
+          def auth_strategy = Registry["one_drive.authentication.userless"].call
+        end
+      end
+    end
+  end
+end
--- a/modules/storages/app/common/storages/peripherals/connection_validators/one_drive/authentication_validator.rb
+++ b/modules/storages/app/common/storages/peripherals/connection_validators/one_drive/authentication_validator.rb
@@ -48,7 +48,7 @@ module Storages
           end
 
           def oauth_token
-            if OAuthClientToken.where(user: @user, oauth_client: @storage.oauth_client).any?
+            if OAuthClientToken.for_user_and_client(@user, @storage.oauth_client).exists?
               pass_check(:existing_token)
             else
               warn_check(:existing_token, message(:oauth_token_missing), halt_validation: true)
--- a/modules/storages/app/common/storages/peripherals/connection_validators/one_drive/authentication_validator.rb
+++ b/modules/storages/app/common/storages/peripherals/connection_validators/one_drive/authentication_validator.rb
@@ -1,0 +1,71 @@
+# frozen_string_literal: true
+
+#-- copyright
+# OpenProject is an open source project management software.
+# Copyright (C) the OpenProject GmbH
+#
+# This program is free software; you can redistribute it and/or
+# modify it under the terms of the GNU General Public License version 3.
+#
+# OpenProject is a fork of ChiliProject, which is a fork of Redmine. The copyright follows:
+# Copyright (C) 2006-2013 Jean-Philippe Lang
+# Copyright (C) 2010-2013 the ChiliProject Team
+#
+# This program is free software; you can redistribute it and/or
+# modify it under the terms of the GNU General Public License
+# as published by the Free Software Foundation; either version 2
+# of the License, or (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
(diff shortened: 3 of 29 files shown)


config/i18n-tasks.yml
```diff
@@ -154,6 +154,7 @@ ignore_unused:
   - '*.permission_header_explanation'
   - 'storages.upsale.*'
   - 'services.*'
+  - 'storages.health.connection_validation.one_drive.*'
```
NobodysNightmare, line 157: :yellow_circle: Why do we need this here, but it wasn't needed for Nextcloud?

modules/storages/app/common/storages/peripherals/connection_validators/one_drive/ampf_configuration_validator.rb
```diff
@@ -0,0 +1,100 @@
...
+              log_extraneous_files(unexpected_files)
+              warn_check(:drive_contents, message("one_drive.unexpected_content"))
+            end
+          end
+
+          # Testing setting permissions and checking permission inheritance would be great
+          # but there are some challenges to it. We need to figure out a good way to go about this
+          # 2025-04-08 @mereghost
+          def client_permissions
+            folder = create_folder.result
```
NobodysNightmare, line 63: :question: Will `create_folder` return the folder also if it was created previously? I am wondering whether we might get into trouble if deleting fails in one test run. Worst case would be that this taints all subsequent test runs.

modules/storages/app/common/storages/peripherals/connection_validators/one_drive/ampf_configuration_validator.rb
```diff
@@ -0,0 +1,100 @@
...
+            delete_folder(folder)
+          end
+
+          def delete_folder(folder)
```
NobodysNightmare, line 67: :green_circle: This is probably a "me problem", but I notice it in this file a lot and I also had it when working on the code I am currently refactoring: The method order is consistently surprising me here.

I know that it's not always possible to order methods "perfectly", but on a base level for methods that are called exactly once, I'd usually expect that:

* The caller comes first
* the callees appear in order of appearance at the caller

E.g.

```ruby
def a
  b
  c
end

def b; end
def c; end
```

Here `delete_folder` is defined before `create_folder`, even though it was called last. The callees of `validate` are also inverted the same way.

This comment is probably too long for a :green_circle: already xD But I wanted to at least voice my expectation once. Now you know :D

modules/storages/app/common/storages/peripherals/connection_validators/one_drive/storage_configuration_validator.rb
```diff
@@ -0,0 +1,141 @@
...
+              fail_check(:drive_id_not_found, message("one_drive.drive_id_not_found"))
+            else
+              pass_check(:drive_id_not_found)
+            end
+          end
+
+          def check_tenant_id
+            return pass_check(:tenant_id) if query_result.success?
+
+            tenant_id_regex = /Tenant '#{@storage.tenant_id}' not found|is neither/
```
NobodysNightmare: :yellow_circle: Is this regex well formed? From the looks I'd expect that you want to match either of these:

* Tenant 'abc' not found
* Tenant 'abc' is neither

However, there is no grouping in this regex, so the following string would also match for a tenant `abc`: `Dragon 'Karl' is neither chromatic nor metallic.` It's enough for `is neither` to be present in the string.

modules/storages/app/common/storages/peripherals/connection_validators/one_drive/storage_configuration_validator.rb
```diff
@@ -0,0 +1,141 @@
...
+
+            if error_payload[:error] == "invalid_client"
+              fail_check(:client_secret, message(:client_secret_wrong))
+            else
+              pass_check(:client_secret)
+            end
+          end
+
+          def diagnostic_request
+            if query_result.result == :error
```
NobodysNightmare, line 104: :yellow_circle: I was a bit surprised when I saw this check. Probably because I was wondering whether any of the other cases that are covered in later checks would be considered `query_result.result == :error`.

However, the more I think about it, the more I believe that maybe we are overusing the UI that we built here to the point that we are misusing it.

So I believe that `diagnostic_request` was strategically put as one of the first checks, so that if it fails, we don't claim that anything else has succeeded. For example just reordering the tests to check `client_id` first, would make it look successful if another error was to shadow it.

How sure can we be about the check order here? E.g. `tenant_id` is now checked before `check_client_id`. But if Microsoft decided to authorize requests first before telling you whether a given tenant exists, we'd errorneously show the tenant id check as successful, if the client id was wrong.

Long story short: Shouldn't we strip many of these checks down to `:diagnostic_request` and then make that test pass if the request was successful and if the request was not successful we should select the error message depending on the error we can see?

Then we'd always show an accurate error message and we would never indicate something as healthy, that might just look healthy because another error is shadowing it. After all: All checks that are based on `query_result` are different perspectives on the same, **single** check that was performed.

modules/storages/app/common/storages/peripherals/storage_interaction/authentication_strategies/strategy.rb
```diff
@@ -56,6 +56,10 @@ def with_token(token)
             @token = token
             self
           end
+
+          def ==(other)
```
NobodysNightmare, line 60: :yellow_circle: Aren't there "rules"/best practices that one should also overwrite the `hash` method when overwriting the `==` method? Right now what could happen (as far as I can tell) is this:

```ruby
a = AuthenticationStrategy.new
b = a.dup
puts a == b # Output: true
h = {}
h[a] = 1
h[b] = 2
puts h.size # Output: 2
```

Of course it should be a surprise that the hash stores two keys, because `a` and `b` are the same, but because rubies maps are hash maps, they rely on the hash first and foremost, to check object identity (code above is untested, this is from memory of why not overwriting `hash` can be a problem).

modules/storages/app/common/storages/peripherals/connection_validators/one_drive/storage_configuration_validator.rb
Kharonus: The current order is the order we get from `login.microsoft.com`. I.e., if tenant and client id is wrong, tenant failure has precedence before client id. This can change, of course, but we should try to be very specific about WHY the request failed.

this leads me to point 2:
stripping down everything to `diagnostic_request ` and pass error message was a very early approach. And then we got the request to please, please interpret that error. This is, what Marcello here did - parsing the error message to to specific checks.

Also having two checks "Client ID" and "Client Secret", where one failed, is much easier to parse for a human, then having to read that detail from the error description below.

The shadowing "should" not be a problem, as long as the microsoft api is stable in their error behaviour.

modules/storages/app/common/storages/peripherals/connection_validators/one_drive/storage_configuration_validator.rb
Kharonus: this should be `(Tenant '#{@storage.tenant_id}' )(not found|is neither)`, right?

modules/storages/app/common/storages/peripherals/connection_validators/one_drive/authentication_validator.rb
```diff
@@ -0,0 +1,71 @@
...
+
+          def validate
+            register_checks(:existing_token, :user_bound_request)
+
+            oauth_token
+            user_bound_request
+          end
+
+          def oauth_token
+            if OAuthClientToken.where(user: @user, oauth_client: @storage.oauth_client).any?
```
Kharonus: 🟡 didn't you wrote a scope exactly for this `where`?

modules/storages/app/common/storages/peripherals/connection_validators/one_drive/ampf_configuration_validator.rb
Kharonus: haha :D I also noticed it, but as you said, method order is fully subjective ^^

modules/storages/app/common/storages/peripherals/connection_validators/one_drive/ampf_configuration_validator.rb
Kharonus: Creating a folder that already exists leads to an error in our commands. It should result in a 409. Hence, a second run would already fail at create folder then, until the admin runs a manual cleanup.

I did talk to marcello about this. The failure case is rare, but if it happens, then the admin REALLY must look into it.

Kharonus (approved):
I do not yet see any blocker.

Well, except to the cleanup of the "old" validators.

config/i18n-tasks.yml
mereghost: The old cold uses the old keys, it will probably be needed once we delete the old code.

modules/storages/app/common/storages/peripherals/connection_validators/one_drive/ampf_configuration_validator.rb
mereghost: It will fail to create and therefore fail the test.

modules/storages/app/common/storages/peripherals/connection_validators/one_drive/ampf_configuration_validator.rb
mereghost: Funnily enough I normally expected the other way around. XD

modules/storages/app/common/storages/peripherals/connection_validators/one_drive/storage_configuration_validator.rb
mereghost: Oh... I forgot the parenthesis there. I added this when I notice we could get 2 different responses depending on the case. It wouldn't explode or "bad match", but hey... better safe than sorry. ;)

modules/storages/app/common/storages/peripherals/connection_validators/one_drive/storage_configuration_validator.rb
mereghost: > any of the other cases that are covered in later checks would be considered query_result.result == :error.

No, they won't. `:error` will only happen on cases where we don't return a more specific error, usually meaning we don't understand the response from the server.

If this fails, there's no point in continuing (same goes for Nextcloud `capabilties` request).

> Shouldn't we strip many of these checks down to :diagnostic_request 

More granular steps is the whole purpose of this, so if we get an `:unauthorized` response, we check the response and return what is wrong with it.

The order here is, in fact, tied to the order that the MS Graph API returns errors for those cases. (Tenant -> Secret -> Id)

In a way, what we are doing is exactly what you want, we are just explicit of the checks we are running that are based on a single **request**. (again, same as NC)

modules/storages/app/common/storages/peripherals/storage_interaction/authentication_strategies/strategy.rb
mereghost: You are right, it is indeed good practice and I totally forgot. At the time I just wanted my test to pass. XD

modules/storages/app/common/storages/peripherals/connection_validators/one_drive/authentication_validator.rb
mereghost: Yes. And I copied over code and never really read it. 🤦🏼

modules/storages/app/common/storages/peripherals/connection_validators/one_drive/ampf_configuration_validator.rb
NobodysNightmare: Is there any way for us to improve the user experience here?

E.g. "skip" the create test on a 409 and then delete the folder anyways?

I dislike it a little bit that we create real persistent folders for a health check, but I really see a problem when this leads to health checks failing (probably with irritating error messages in the wrong places).

> The failure case is rare

I want to disagree. Simple example scenario: Delete permissions are broken.

1. Admin checks health, sees that there is an error, but
    * forgets to take a screenshot / is bad at remembering error messages
    * is forced to re-run tests because cache empties between showing aggregated result and full report
    * any other reason you want to come up with
2. Admin looks at health check that was generated from _any test run after the first_

Expected behaviour: The admin sees that deletion is broken.

Actual behaviour: The UI tells the admin to check the create permissions.

So the scenario is rare for a well-configured storage. But you don't run health checks on a well-configured storage :D

modules/storages/app/common/storages/peripherals/connection_validators/one_drive/ampf_configuration_validator.rb
NobodysNightmare: > I normally expected the other way around

You are a monster! I don't judge, but you are ^^

mereghost:
> I do not yet see any blocker.
> 
> Well, except to the cleanup of the "old" validators.

We should remove then on the PR that replaces the UI, otherwise tests will explode. =/

modules/storages/app/common/storages/peripherals/connection_validators/one_drive/storage_configuration_validator.rb
NobodysNightmare: @Kharonus I think the second group is enough, i.e.

```
Tenant '#{@storage.tenant_id}' (not found|is neither)
```

Also since we don't care for the content of the group we can make it non-capturing:

```
Tenant '#{@storage.tenant_id}' (?:not found|is neither)
```

This will not change whether the regex matches, but it avoids that the content of the group is accessible under `match[1]`.

modules/storages/app/common/storages/peripherals/connection_validators/one_drive/storage_configuration_validator.rb
NobodysNightmare: > stripping down everything to diagnostic_request and pass error message was a very early approach. And then we got the request to please, please interpret that error. 

Just to be very clear: I still suggest to _interpret_ the error/error message. I am just suggesting to boil it down to one check with different messages (that **we** control).

> The shadowing "should" not be a problem, as long as the microsoft api is stable in their error behaviour.

That "should" is exactly the concern that I wanted to express here. It's a brittle assumption and one that you can't even properly put into a VCR casette. Our customers will be first to notice once our assumption breaks.

If we want to keep it like that, I'd probably wish for a comment that highlights how this depends on M$ APIs to evaluate errors in the exact same order that we assume here.

modules/storages/app/common/storages/peripherals/connection_validators/one_drive/ampf_configuration_validator.rb
mereghost: If push comes to shove, we can handle the conflict result and return a message telling the user to delete the folder on the remote.

If the deletion failed, an error would already have happened. We can instruct the user to check the permissions **and** delete the folder, but that is it.

Also a couple of examples can be rendered moot point if we persist the results which has already been talked about.

modules/storages/app/common/storages/peripherals/connection_validators/one_drive/ampf_configuration_validator.rb
mereghost: I've been called worse XD

modules/storages/app/common/storages/peripherals/connection_validators/one_drive/ampf_configuration_validator.rb
mereghost: I've added handling for when folder creation fails. We still fail, but with a more appropriate error message.

PR #18583, opened 2025-04-08
[#61896] Overview of time logged per day per user in PDF timesheet

as-op:
# Ticket

https://community.openproject.org/work_packages/61896

Modify the overview table to show more sums (per user per day, total per user, total)
Changing the details per user tables a bit.

## Screenshots
image

# Dev checklist

- [x] "Total sum: HH:MM h" for a sum of all users in the first line of the overview page
- [x] The summary table includes row per day.
- [x] The first column in the summary table shows the date of day, the abbreviated weekday
- [x] For each user there is a column.
- [x] The summary table columns are sorted by user name
- [x] In case there are many users, e.g. more than 6 users the table is split across multiple tables (but balanced amount, 
         e.g. 7 users => 2 table (4 users, 3 users)
         e.g. 12 users => 2 tables (6 users, 6 users)
- [x] Empty (no logged time for that day) rows are not shown
- [x] The hours are always shown in hours (hh:mm) not in days and hours (2 d 2 h).
- [x] The values are aligned at the separator (right aligned values)
- [x] The values in the cells are linked to the corresponding heading in the details table below
   - [x] User A
   - [x] Hours User B for Monday 03.03.2025
- [x] In the details section 
   - [x] the sum of hours per day is listed as the row if there is more than one entry per day
   - [x] the sum of all days is listed in a row if there is more than on day
   - [x] The IDs of the work packages are listed. The IDs include a link to the work package itself
   - [x] The first column shows the date of day, a comma and the abbreviated weekday

# Merge checklist

- [x] Added/updated tests


lib/chronic_duration.rb
```diff
@@ -184,6 +189,25 @@ def output(seconds, opts = {})
...
+        divider = ":"
+        result = str.split(divider).each_with_index.map do |n, index|
+          if index > 0 && n.length == 1
+            "0#{n}"
+          elsif index > 0 && n.empty?
+            "00"
+          else
+            n
+          end
+        end.join(divider)
```
ulferts, line 204: Or
```
result = str.split(divider).each_with_index.map do |n, index|
          n.rjust(index > 0 ? 2 : 1, '0')
        end.join(divider)
```

ulferts (approved):
The PR is solid - I didn't find a way to break the functionality. Plus, I only have a single code remark, a tiny suggestion. 

Functionality-wise, I have to say, though, that the start of the document to me has lost its structure. That might be due to the test data I have which leads to having more entries and a lot of empty cells. I would expect the usual usage scenario to include the time span of a month so in that case, the initial table would fit on the first content page (unless there are more than 6 users).

But if the time span is longer, the user is presented with:
* First a total which looks lost since it does not sum up anything but is just that single number.
* Then a table on which I have to spent some time to understand it. There is no heading for it so I have to parse it until I understand that it represents the entries per day and user. 
* Then potentially another such table which I again have to understand as it might look completely different from the one before due to the changed names and potentially different dates.

image

In my opinion, and this is opinionated, I would show the overview table again as before. A per user total and the complete sum. After that, the tables per day and users can be added but I would do so with a heading to help the user make sense of the tables.

PR #18590, opened 2025-04-09
[#62444] Webhook embedded resource custom fields are null

dombesz:
# Ticket
https://community.openproject.org/work_packages/62444

Provide all the custom fields for a work package embedded project.

The previous fix calling `to_hash` on the representer wouldn't transform the deeply nested structures into a hash, thus no embedded objects were transformed during the call.
The fix consists of calling `to_json` inside the authenticated user block, this way all the deeply nested structures are converted to json while having the current user set.

# Merge checklist

- [X] Added/updated tests


oliverguenther (approved):
Great catch 👍 I would vote for merging this into 15.5., scope is limited and can be tested by L2, so not blocking Cécile

PR #18594, opened 2025-04-09
[62891][63152] Recalculate start date correctly when toggling "Working days only"

cbliard:
# Ticket

https://community.openproject.org/wp/62891
https://community.openproject.org/wp/63152

There were weird recalculations of finish date when saving date picker, a bit hard to reproduce. It was due to the start date recalculation when toggling "Working days only" which was not reflected correctly in the date picker. On top of that, the calculation of the start date could be incorrect if lag was involved.

There was also a bug in date picker enabled dates when the timezone offset is negative.

For an automatically scheduled successor, when "Working days only" is turned off and the day before the start date is a non-working day, the start will shift to an earlier date. Multiple things need to work correctly:

- The start date must be recalculated and displayed in the date picker.
- The mini calendar of the date picker need to have enabled dates updated, as the start date has changed.
- If the due date was not set manually, duration must be kept and due date will be recalculated.
- If the due date was set manually, it must not be changed and duration will be recalculated.
- If the due date was unset manually, it must remain unset and duration must be unset too.

Also the lag calculation was partly incorrect: it was adding non-working days when "Working days only" was on, and days when "Working days only" was off. This has been fixed: lag is always about working days regardless of the successor "Working days only" setting.

For this reason, when the work package is automatically scheduled, the start date is always recalculated in the `SetAttributesService` to ensure we never open the date picker with an error message saying "Start date can only be set to 2025-04-16 or later so as not to violate the work package's relationships.".

## Screenshots

Before:

due date saved is not the chosen one, and can select date before start date as finish date (using New York time zone)

[bug when setting due date and toggling nwd.webm](https://github.com/user-attachments/assets/e0e50d54-bd1a-44e3-abe4-2af90739ea1c)

After:

due date saved is the chosen one, and dates before start date cannot be selected (using New York time zone)

[ok when setting due date and toggling nwd.webm](https://github.com/user-attachments/assets/fba63833-f96a-42e2-b230-83193925501d)

- Change the way the soonest_start is calculated: first calculate the date with the lag, which is always working days only, then move to next working day or not depending on "ignore_non_working_days" value of the work package.
- Change the algorithm in `WorkPackages::SetAttributesService` to prioritize keeping `due_date` if it has been set by the user
- Always calculate and set start date for automatically scheduled work packages, to fix any wrong start date
  - This has the side effect to recompute the start date when "Working days only" is toggled. Before it was recalculated only if scheduling mode was changed.
- In date picker, correctly convert start date to have the same time shift as the date from flatpickr for easier comparisons. And remove the minimalStartDate parameter as it's not needed: it's always the start date.

# Merge checklist

- [x] Added/updated tests
  - a lot
  - the test checking the enabled dates in the date picker is now run also in the New York time zone

### DIFF
--- a/app/components/work_package_relations_tab/closest_relation.rb
+++ b/app/components/work_package_relations_tab/closest_relation.rb
@@ -57,7 +57,7 @@ module WorkPackageRelationsTab
     def soonest_start
       return @soonest_start if defined?(@soonest_start)
 
-      @soonest_start = relation.successor_soonest_start(gap: 0.days)
+      @soonest_start = relation.successor_soonest_start
     end
 
     def inspect
--- a/app/components/work_packages/date_picker/form_component.html.erb
+++ b/app/components/work_packages/date_picker/form_component.html.erb
@@ -105,7 +105,6 @@
                                           ignore_non_working_days: work_package.ignore_non_working_days,
                                           schedule_manually:,
                                           is_schedulable: !disabled?,
-                                          minimal_scheduling_date:,
                                           is_milestone: milestone?,
                                           date_mode:,
                                           start_date_field_id: "work_package_start_date",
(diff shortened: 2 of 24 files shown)


bsatarnejad (approved):
Works perfect 👍

PR #18600, opened 2025-04-11
[59915] Update PageHeaders & SubHeaders in the (rails) project pages (Part 2)

bsatarnejad:
# Ticket
https://community.openproject.org/wp/59915

Update page header in remaining rails pages


modules/backlogs/app/views/version_settings/edit.html.erb
```diff
@@ -1,36 +0,0 @@
...
-
-++#%>
-
-<%= toolbar title: t(:label_version) %>
-<% html_title t(:label_version) %>
-
-<%= labelled_tabular_form_with model: @version, url: project_version_path(@project, @version), html: { method: :put } do |f| %>
-  <%= render partial: "form", locals: { f: } %>
-  <%= styled_button_tag t(:button_save), class: "-primary -with-icon icon-checkmark" %>
-<% end %>
```
HDinger, line 36: Are you sure, that this can simply be deleted? Further, you deleted these files, but not the route or the controller action. So please tidy that up as well.

lookbook/docs/components/sub-header.md.erb
```diff
@@ -26,6 +26,7 @@ This component is an own implementation for OpenProject requirements of contextu
 
 - Restrict SubHeader actions to the page's content-specific actions (e.g. filtering or creating a new object).
 - Use IconButtons instead of Buttons for actions only when the icon is clear enough to describe the action.
+- Include all the actions to the SubHeader that adds or remove elements to the content bellow it (eg. New work package), and/or modifies the display of the content bellow (eg. Filter)
```
HDinger: How does that differ from the point above? 

`Restrict SubHeader actions to the page's content-specific actions (e.g. filtering or creating a new object).`

bsatarnejad: Hi @HDinger sometimes ago @marcalcobe asks me to add this text here to the lookbook to clarify the sub header actions better.

marcalcobe: I think what I intended there was to specify something different between filtering and creating elements, but I would be okay with only keeping the first sentence.

HDinger (changes requested):
Hi @bsatarnejad 

I did not manage to go through all of it yet, but would like to give you some initial feedback already:

* Please extract the more complex PageHeaders into separate components. Especially the one for the wiki/show page is so complex, it has to be extracted. I think a general rule of thumb would be that as soon as you have logic inside, extract it into a component (e.g when having `if` clauses or permission checks)
* Please do not use classes for testing but provide test_selectors to the elements which you then access in the class.
* I am wondering whether in this example, the toolbar should read "Seeded project / Wiki / Coole version" . What do you think? 

Bildschirmfoto 2025-04-15 um 15 21 07

app/components/wiki_pages/wiki_header_component.html.erb
```diff
@@ -0,0 +1,220 @@
+<%#-- copyright
```
HDinger: I guess the folder should be named `components/wiki/`.  
In other places, we name the components after the action they represent. So in this case it would be `show_page_header_component`

app/components/wiki_pages/wiki_header_component.html.erb
```diff
@@ -0,0 +1,220 @@
...
+++#%>
+
+<%=
+  render Primer::OpenProject::PageHeader.new do |header|
+    header.with_title { h(@page.title) }
+    header.with_breadcrumbs(
+      breadcrumb_for_page(@project, @page)
+    )
+    if @editable
+      if User.current.allowed_in_project?(:edit_wiki_pages, @project) && @page.current_version?
```
HDinger: This `html` contains a lot of logical expressions like this one. Please move all of that logic to the `rb` file into dedicated methods and just call these methods in the html, eg. `if show_edit?`

app/components/wiki_pages/wiki_header_component.html.erb
```diff
@@ -0,0 +1,220 @@
...
+        ) do |button|
+          button.with_leading_visual_icon(icon: "book")
+          t(:label_history)
+        end
+      end
+    end
+    header.with_action_menu(
+      menu_arguments: { anchor_align: :end },
+      button_arguments: {
+        icon: "op-kebab-vertical",
```
HDinger:
```suggestion
        icon: "kebab-horizontal",
```

app/components/wiki_pages/wiki_header_component.html.erb
```diff
@@ -0,0 +1,220 @@
...
+            menu.with_item(
+              label: t(:button_unlock),
+              tag: :a,
+              size: :medium,
+              content_arguments: { data: { method: :post } },
+              href: url_for(controller: "wiki", action: "protect", id: @page, protected: 0)
+            ) do |item|
+              item.with_leading_visual_icon(icon: :unlock)
+            end
+          end
```
HDinger: There is a lot of duplication in here with only minor differences as the icon, the label and the protected value. You could let a method define these values and just call one menu item here:

for example:
```

 menu.with_item(
              label: lock_data[:label],
              tag: :a,
              size: :medium,
              content_arguments: { data: { method: :post } },
              href: url_for(controller: "wiki", action: "protect", id: @page, protected:  lock_data[:protected])
            ) do |item|
              item.with_leading_visual_icon(icon:  lock_data[:icon])
            end
       end
```

app/components/wiki_pages/wiki_header_component.html.erb
```diff
@@ -0,0 +1,220 @@
...
+        tag: :a,
+        test_selector: "wiki-new-child-button",
+        href: url_for({ controller: "/wiki", action: "new_child", project_id: @project.identifier, id: @page })
+      ) do |button|
+        button.with_leading_visual_icon(icon: :plus)
+        t(:create_new_page)
+      end
+    end
+  end
+%>
```
HDinger: Please move the SubHeader to a different component.

app/controllers/wiki_controller.rb
```diff
@@ -415,7 +415,7 @@ def default_breadcrumb
   end
 
   def show_local_breadcrumb
-    @page&.ancestors&.any?
+    false
```
HDinger, line 422: We probably can get rid of this whole method and the belonging logic 🤔  But this is out of scope for this ticket but rather a following clean-up.

app/views/wiki/history.html.erb
```diff
@@ -26,9 +26,15 @@ Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
 See COPYRIGHT and LICENSE files for more details.
 
 ++#%>
-<%= toolbar title: @page.title %>
 <% html_title t("activerecord.models.wiki"), t(:label_history) %>
-<% breadcrumb_paths(*(@page.ancestors.reverse.collect { |parent| link_to h(parent.breadcrumb_title), { id: parent, project_id: parent.project } } + [h(@page.breadcrumb_title)])) %>
+<%=
+  render Primer::OpenProject::PageHeader.new do |header|
+    header.with_title { h(@page.title) }
```
HDinger: This feels weird.. I'd expect something like "History of {title}" and get rid of that second headline. Or just "History" and represent the title in the breadcrumb

spec/features/menu_items/wiki_menu_item_spec.rb
```diff
@@ -92,8 +92,8 @@
     visit project_wiki_path(project, wiki_page)
 
     # creating the menu item with the pages name for the menu item
-    click_link_or_button "More"
-    click_link_or_button "Configure menu item"
+    page.find('[data-test-selector="wiki-more-dropdown-menu"]').click
```
HDinger: You can use our test helpers for things like that:
`page.find_test_selector("foo-bar").click`

app/views/wiki/new.html.erb
```diff
@@ -29,6 +29,14 @@ See COPYRIGHT and LICENSE files for more details.
 
 <% html_title t("create_new_page") %>
 
+<%=
+  render Primer::OpenProject::PageHeader.new do |header|
+    header.with_title { t("create_new_page") }
```
HDinger: That string tranlsates to "Wiki page", maybe we should change it to "Create new wiki page"?

app/views/wiki/new.html.erb
```diff
@@ -29,6 +29,14 @@ See COPYRIGHT and LICENSE files for more details.
...
+    header.with_breadcrumbs(
+      [{ href: project_overview_path(@project.id), text: @project.name }, t("create_new_page")]
```
HDinger: Here again, we have the question, whether the module "Wiki" should be present in the toolbar. It feels like it is missing:

Bildschirmfoto 2025-04-16 um 09 09 06

bsatarnejad:
> Hi @bsatarnejad
> 
> I did not manage to go through all of it yet, but would like to give you some initial feedback already:
> 
> * Please extract the more complex PageHeaders into separate components. Especially the one for the wiki/show page is so complex, it has to be extracted. I think a general rule of thumb would be that as soon as you have logic inside, extract it into a component (e.g when having `if` clauses or permission checks)
> * Please do not use classes for testing but provide test_selectors to the elements which you then access in the class.
> * I am wondering whether in this example, the toolbar should read "Seeded project / Wiki / Coole version" . What do you think?
> 
> Bildschirmfoto 2025-04-15 um 15 21 07

Hey @HDinger thanks for the hints.
Regarding the breadcrumb in the image, I think it is correct now based on what we discussed before in UX meeting:

> Use breadcrumb fromat:
> 
> {Project name} → {Wiki name}

app/helpers/wiki_helper.rb
```diff
@@ -40,22 +40,30 @@ def wiki_page_options_for_select(pages,
                                               ids:)
   end
 
-  def breadcrumb_for_page(page, action = nil)
-    related_pages = page.ancestors.reverse
+  def breadcrumb_for_page(project, page, action = nil)
+    breadcrumbs = []
+    breadcrumbs << project_breadcrumb(project)
```
HDinger, line 45: Maybe we should add the module "Wiki" in between here. I stumbled multiple times over the fact that it is missing (see other comments I made). That would mean however, if you have a default Wiki page called "Wiki", there would be two duplicated entries.. I think we have to discuss that with the product team.

HDinger (changes requested):

app/controllers/wiki_controller.rb
HDinger: I created a follow-up ticket for this: https://community.openproject.org/projects/design-system/work_packages/63485/activity

app/components/wiki_pages/wiki_header_component.html.erb
bsatarnejad: The Module name for wiki is WikiPages, so the folder name should be wiki_pages.

app/components/wiki_pages/show_sub_header_component.html.erb
```diff
@@ -27,8 +27,20 @@ See COPYRIGHT and LICENSE files for more details.
...
 ++#%>
 
-<%= error_messages_for "version" %>
-
-<% if @project.enabled_modules.map(&:name).include?("backlogs") %>
-  <%= version_settings_fields @version, @project %>
-<% end %>
+<%=
+  render(Primer::OpenProject::SubHeader.new) do |subheader|
+    if show_create?
```
HDinger: This should probably be before calling the component to avoid empty space:

Bildschirmfoto 2025-04-17 um 14 12 35

app/views/wiki/history.html.erb
```diff
@@ -26,10 +26,16 @@ Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
...
-<% breadcrumb_paths(*(@page.ancestors.reverse.collect { |parent| link_to h(parent.breadcrumb_title), { id: parent, project_id: parent.project } } + [h(@page.breadcrumb_title)])) %>
-<h3><%= t(:label_history) %></h3>
+<%=
+  render Primer::OpenProject::PageHeader.new do |header|
+    header.with_title { t(:label_history) }
+    header.with_breadcrumbs(
+      breadcrumb_for_page(@project, @page)
+    )
+  end
+%>
```
HDinger, line 37: The breadcrumb and the title does not match
Bildschirmfoto 2025-04-17 um 14 17 21

spec/features/wiki/child_pages_spec.rb
```diff
@@ -62,7 +62,10 @@
     click_button "Save"
 
     # hierarchy displayed in the breadcrumb
-    expect(page).to have_css("#breadcrumb #{test_selector('op-breadcrumb')}",
+    within('[data-test-selector="wiki-page-header-breadcrumbs"]') do
+      expect(page).to have_text(parent_page.title.to_s)
+    end
+    expect(page).to have_css(".breadcrumb-item",
```
HDinger: What is the point of testing this? You are already checking the for the title to be part of the breadcrumb above.

HDinger (changes requested):
Looks good, I only have minor remarks left 👍

app/views/wiki/history.html.erb
bsatarnejad: I take it from your comment here: https://github.com/opf/openproject/pull/18600#discussion_r2046225203,
'History' for title and page.title in breadcrumb, did I misunderstand?

HDinger: Yeah, this is not what I meant.. Normally, the breadcrumb and the title should be the same. The only excpetion is when using the `selected_item_font` option. So you have two options:

* Both show "History"
* The title shows "History", the breadcrumb shows "Blup: **History**"

spec/features/wiki/child_pages_spec.rb
bsatarnejad: Sorry, It was duplicated, fixed.

oliverguenther (approved):

PR #18607, opened 2025-04-11
hide internal comments for users without permission on atom export

brunopagno:
# Ticket
https://community.openproject.org/wp/63371

Don't _atom export_ internal comments when the user requesting does not have access to it.

Just included the scope on the atom export

# Merge checklist
- [ ] Added/updated tests


akabiru (approved):
🔒

PR #18613, opened 2025-04-11
[62992] Show single date work packages in the calendar

cbliard:
# Ticket

https://community.openproject.org/wp/62992

In calendar module, show work packages even if they only have 1 date set.

For work package with end date only: use end date as start date if start date is absent.
For work package with start date only: they were already visible so nothing particular.

# Merge checklist

- [X] Added/updated tests

### DIFF
--- a/frontend/src/app/features/calendar/wp-calendar/wp-calendar.component.ts
+++ b/frontend/src/app/features/calendar/wp-calendar/wp-calendar.component.ts
@@ -400,11 +400,11 @@ export class WorkPackagesCalendarComponent extends UntilDestroyedMixin implement
       const startDate = this.workPackagesCalendar.eventDate(workPackage, 'start');
       const endDate = this.workPackagesCalendar.eventDate(workPackage, 'due');
 
-      const exclusiveEnd = moment(endDate).add(1, 'days').format('YYYY-MM-DD');
+      const exclusiveEnd = endDate && moment(endDate).add(1, 'days').format('YYYY-MM-DD');
 
       return {
         title: workPackage.subject,
-        start: startDate,
+        start: startDate || endDate,
         editable: this.workPackagesCalendar.dateEditable(workPackage),
         durationEditable: this.workPackagesCalendar.eventDurationEditable(workPackage),
         end: exclusiveEnd,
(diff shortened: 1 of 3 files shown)


frontend/src/app/features/calendar/wp-calendar/wp-calendar.component.ts
```diff
@@ -400,11 +400,11 @@ export class WorkPackagesCalendarComponent extends UntilDestroyedMixin implement
       const startDate = this.workPackagesCalendar.eventDate(workPackage, 'start');
       const endDate = this.workPackagesCalendar.eventDate(workPackage, 'due');
 
-      const exclusiveEnd = moment(endDate).add(1, 'days').format('YYYY-MM-DD');
+      const exclusiveEnd = endDate && moment(endDate).add(1, 'days').format('YYYY-MM-DD');
 
       return {
         title: workPackage.subject,
-        start: startDate,
+        start: startDate || endDate,
```
HDinger: This feels really weird.. why should the start date be either start date or end date. Any future developer without the backgound of this ticket will fail to understand this. Please leave at least a comment.

HDinger (commented):
Actually, the issue only happens when there is no start date set. I will update the ticket accordingly. 

With regards to the solution: It works as "specified" altough the AC are rather slim.. Currently there is no visual difference between a WP that has start and end at the same date vs a WP that only has one of them set. In the Gantt chart, we solve that with a gradient:

Bildschirmfoto 2025-04-14 um 10 59 57

Imho this would make sense here as well. Currently in the Gantt chart, this is an inline style, but we could easily extract that into a class and apply it in both cases. Given however, that WP with only a start date have always been shown in the calendar and nobody complained about it before, is might not be such an issue after all 🤷

frontend/src/app/features/calendar/wp-calendar/wp-calendar.component.ts
cbliard: Agreed, I have added a proper comment that it's the start date of the calendar event.

cbliard:
> Imho this would make sense here as well. Currently in the Gantt chart, this is an inline style, but we could easily extract that into a class and apply it in both cases. **Given however, that WP with only a start date have always been shown in the calendar and nobody complained about it before, is might not be such an issue after all** 🤷

Yes, that was my conclusion too. That's probably good enough.
Maybe there could be another feature to add an indication that the start or finish date is not set. Gradient could be a solution.

cbliard:
@HDinger, I have added a comment as requested. Could you review it again please?

HDinger (approved):
I left a comment in the ticket for the gradient thing. I am fine with extracting that into a separate ticket if the product team agrees with the proposition.

cbliard:
The failing spec is `spec/features/work_packages/scheduling/automatic_scheduling_logic_spec.rb:340`.
It looks to be flickering because it passes locally.

PR #18630, opened 2025-04-14
[#62293] PDF export does not export all embedded screenshots

as-op:
# Ticket

https://community.openproject.org/work_packages/62293

If an image attachment link in a description (wp A) is copied from another work package (wp B) description, the image file is not copied with it. The PDF export searches in the work package (wp A) but does not find it, the web app does, as it requests just the image src, regardless of the work package origin. PDF export should do the same.

Get the attachment id from the src and request the attachment directly instead of iterating through the attachments of the work package. An additional check is needed if the exporting user is allowed to see the attachment.

# Merge checklist

- [x] Added/updated tests

### DIFF
--- a/app/models/work_package/pdf_export/common/attachments.rb
+++ b/app/models/work_package/pdf_export/common/attachments.rb
@@ -71,9 +71,11 @@ module WorkPackage::PDFExport::Common::Attachments
   end
 
   def attachment_by_api_content_src(_work_package, src)
-    # /attachments/#{id}/content
-    parts = src.split("/")
-    attachments_id = parts[-2]
+    attachment_regex = %r{/attachments/(\d+)/content}
+    return nil unless src&.match?(attachment_regex)
+
+    attachments_id = src.scan(attachment_regex).first.first
+
     # return nil if the src is not an attachment url
     return nil unless api_url_helpers.attachment_content(attachments_id) == src
 
--- a/app/models/work_package/pdf_export/common/attachments.rb
+++ b/app/models/work_package/pdf_export/common/attachments.rb
@@ -70,8 +70,21 @@ module WorkPackage::PDFExport::Common::Attachments
     resize_image(local_file.path)
   end
 
-  def attachment_by_api_content_src(work_package, src)
-    # find attachment by api-path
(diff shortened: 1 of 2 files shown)


app/models/work_package/pdf_export/common/attachments.rb
```diff
@@ -70,8 +70,21 @@ def attachment_image_filepath(work_package, src)
...
 
-  def attachment_by_api_content_src(work_package, src)
-    # find attachment by api-path
-    work_package.attachments.find { |a| api_url_helpers.attachment_content(a.id) == src }
+  def attachment_by_api_content_src(_work_package, src)
+    # /attachments/#{id}/content
+    parts = src.split("/")
+    attachments_id = parts[-2]
+    # return nil if the src is not an attachment url
+    return nil unless api_url_helpers.attachment_content(attachments_id) == src
```
klaustopher: WDYT about instead of splitting here, we use a regex:

```ruby
attachment_regex = %r{/attachments/(\d+)/content}

return nil if src !~ attachment_regex
attachments_id = src.scan(attachment_regex).first.first
```

app/models/work_package/pdf_export/common/attachments.rb
```diff
@@ -70,8 +70,21 @@ def attachment_image_filepath(work_package, src)
...
+
+    attachment = Attachment.find_by(id: attachments_id.to_i)
```
klaustopher, line 82: Interesting that the Attachment does not have a visible scope. Maybe we should add that at some point :)

app/models/work_package/pdf_export/common/attachments.rb
as-op: Yes, a regular expression is the better way. Thanks! 💙

app/models/work_package/pdf_export/common/attachments.rb
as-op: Communicator stream will probably change the usage of Attachments for rich text, so let's see what is left of the code after that.

klaustopher (approved):

as-op:
Thanks @klaustopher!

PR #18646, opened 2025-04-15
Fix custom action title flaky spec by fetching titles on initialization.

dombesz:
# Ticket
N/A

Fix flaky spec

```
1) Custom actions viewing workflow buttons
     Failure/Error:
       expect(page)
         .to have_button("Unassign", title: "Removes the assignee")

       expected to find button "Unassign" that is not disabled with title Removes the assignee within #

rspec ./spec/features/work_packages/custom_actions/custom_actions_spec.rb:140 # Custom actions viewing workflow buttons
```

## Screenshots
<details><summary>The bug</summary>
<p>

button-bug

</p>
</details> 

<details><summary>The fix</summary>
<p>

button-fix

</p>
</details> 

The issue is, the Custom action resource holding the title is not being fetched unless [the custom action button is hovered](https://github.com/opf/openproject/blob/584f6768b0ad8cc3b90d24f64f094b74fe1dad34/frontend/src/app/features/work-packages/components/wp-custom-actions/wp-custom-actions/wp-custom-action.component.ts#L132-L134). The fix is to fetch the resource earlier in the initialize method of the `wp-custom-action` component.

# Merge checklist

- [ ] Added/updated tests

### DIFF
--- a/frontend/src/app/features/work-packages/components/wp-custom-actions/wp-custom-actions/wp-custom-action.component.ts
+++ b/frontend/src/app/features/work-packages/components/wp-custom-actions/wp-custom-actions/wp-custom-action.component.ts
@@ -74,6 +74,7 @@ export class WpCustomActionComponent extends UntilDestroyedMixin implements OnIn
         this.untilDestroyed(),
       )
       .subscribe(() => this.cdRef.detectChanges());
+    this.fetchAction();
   }
 
   private fetchAction() {


NobodysNightmare:
Coming here from https://github.com/opf/openproject/pull/18643 and now wondering when the hovering starts in that spec and whether the increase of wait-time was necessary in the first place.

If I understand your fix here correctly, you now ensure that we fetch the necessary information eagerly, thus I'd expect the default wait timeout to be enough.

akabiru:
> Coming here from #18643 and now wondering when the hovering starts in that spec and whether the increase of wait-time was necessary in the first place.
> 
> If I understand your fix here correctly, you now ensure that we fetch the necessary information eagerly, thus I'd expect the default wait timeout to be enough.

Agree- we should be good to revert the wait time increase. It was a naive bandaid admittedly.

NobodysNightmare:
One more thing I noticed on Community (I have to check behaviour of this PR locally): each time the mouse enters a custom action buttons area, we perform a web request, even after the initial and successful load. Is that something we want to avoid?

[Screencast from 2025-04-15 16-36-02.webm](https://github.com/user-attachments/assets/c296e405-4f0a-4273-bcbe-d7e18c728a65)

akabiru:
> One more thing I noticed on Community (I have to check behaviour of this PR locally): each time the mouse enters a custom action buttons area, we perform a web request, even after the initial and successful load. Is that something we want to avoid?
> Screencast.from.2025-04-15.16-36-02.webm

yeah, this does seem rather odd. I confirm the same experience locally.

akabiru (approved):
Really nice one @dombesz 🏅 

Shall we revert the timeout increase https://github.com/opf/openproject/commit/4901de9a4005e6fbe022ed31ada1de696ac5beb1 for good measure?

dombesz:
> Coming here from #18643 and now wondering when the hovering starts in that spec and whether the increase of wait-time was necessary in the first place.
> 
> If I understand your fix here correctly, you now ensure that we fetch the necessary information eagerly, thus I'd expect the default wait timeout to be enough.

Yes, the wait time should be sufficient now. ~~I'm not even sure how the request was passing sometimes, because according to my understanding, the request should not have been fetched unless a button hover happened.~~ My hunch is, there is a race condition between fetching the action and displaying the title label. If the request is faster than displaying the label, then it passes, otherwise don't.

dombesz:
> Really nice one @dombesz 🏅
> 
> Shall we revert the timeout increase [4901de9](https://github.com/opf/openproject/commit/4901de9a4005e6fbe022ed31ada1de696ac5beb1) for good measure?

@akabiru Yes, I think we can revert the extra wait time.

dombesz:
> One more thing I noticed on Community (I have to check behaviour of this PR locally): each time the mouse enters a custom action buttons area, we perform a web request, even after the initial and successful load. Is that something we want to avoid?
> 
>  Screencast.from.2025-04-15.16-36-02.webm

Not entirely sure why it is there in the first place, my assumption is that it was meant to fetch the resource on demand to make the UI more asynchronous and the resource fetching is in a race condition with displaying the title. If that is the case, then fetching on hover should not be required. However I wouldn't remove it, because there could be a race condition from the `wp-custom-actions` fetching the custom actions.

https://github.com/opf/openproject/blob/584f6768b0ad8cc3b90d24f64f094b74fe1dad34/frontend/src/app/features/work-packages/components/wp-custom-actions/wp-custom-actions.component.ts#L62-L73

akabiru:
> > One more thing I noticed on Community (I have to check behaviour of this PR locally): each time the mouse enters a custom action buttons area, we perform a web request, even after the initial and successful load. Is that something we want to avoid?
> > Screencast.from.2025-04-15.16-36-02.webm
> 
> Not entirely sure why it is there in the first place, my assumption is that it was meant to fetch the resource on demand to make the UI more asynchronous and the resource fetching is in a race condition with displaying the title. If that is the case, then fetching on hover should not be required. However I wouldn't remove it, because there could be a race condition from the `wp-custom-actions` fetching the custom actions.
> 
> https://github.com/opf/openproject/blob/584f6768b0ad8cc3b90d24f64f094b74fe1dad34/frontend/src/app/features/work-packages/components/wp-custom-actions/wp-custom-actions.component.ts#L62-L73

Good topic for Dev Open Door today? @NobodysNightmare

NobodysNightmare:
I'll not be there today.

akabiru:
> > Really nice one @dombesz 🏅
> > Shall we revert the timeout increase [4901de9](https://github.com/opf/openproject/commit/4901de9a4005e6fbe022ed31ada1de696ac5beb1) for good measure?
> 
> @akabiru Yes, I think we can revert the extra wait time.

Seems we still need the extended wait time- lots of flaky occurrences since reverting. Will reinstate, seems low enough risk to me. https://github.com/opf/openproject/actions/runs/14592814913/job/40931717462

PR #18648, opened 2025-04-15
implementation/63384 On work package comment edit set attachment replacements

akabiru:
When a comment is updated, replace attachments.

https://community.openproject.org/wp/63384

### DIFF
--- a/app/controllers/work_packages/activities_tab_controller.rb
+++ b/app/controllers/work_packages/activities_tab_controller.rb
@@ -129,6 +129,7 @@ class WorkPackages::ActivitiesTabController < ApplicationController
       call = update_journal_service_call
 
       if call.success? && call.result
+        claim_journal_attachments_for(call.result)
         update_item_show_component(journal: call.result, grouped_emoji_reactions: grouped_emoji_reactions_for_journal)
       else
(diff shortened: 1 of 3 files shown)


brunopagno (approved):
Looks great. And so straightforward :clap::clap::clap:

One thing to think about, though, is if these checks should be part of the journal contracts instead of living in the controller.

These should become more clear and easier to handle when we extract comments from journals. But at this point the line looks a bit blurry on where the responsibility of each action lies.

In any case, just food for thought, no changes necessary. The PR looks good :+1:

akabiru:
> Looks great. And so straightforward 👏👏👏
> 
> One thing to think about, though, is if these checks should be part of the journal contracts instead of living in the controller.
> 
> These should become more clear and easier to handle when we extract comments from journals. But at this point the line looks a bit blurry on where the responsibility of each action lies.
> 
> In any case, just food for thought, no changes necessary. The PR looks good 👍

Good eye, I considered baking the claims service as a dependent service to Journals{Create,Update}Service but unfortunately those services don't extend the BaseService implementation that allows for this behaviour, and since this should be soon moved to comments- perhaps it's good to keep it decoupled from journals.

PR #18649, opened 2025-04-15
[#63482] Create Project Status Button Component

myabc:
⚠️ **~~This PR is based on #18478. Please review/merge first.~~**

# Ticket

https://community.openproject.org/wp/63482

Implements a Project Status Button Component based on the existing OP Primer Status Component. Integrates the Status Button Component on the Project Settings > Information page.

In detail:

* Introduces `Projects::StatusButtonComponent` and its Lookbook Preview.
* Defines colours, icons as (Ruby) constants nested under `Project::Statuses` (c.f. `Meetings::Statuses`).
* Adds `Projects::StatusController` with `#update` action, routes and accompanying permissions.

## Screenshots

Screenshot 2025-04-28 at 21 12 03
Screenshot 2025-04-28 at 21 11 11
Screenshot 2025-04-28 at 21 11 02
Screenshot 2025-04-28 at 21 10 53

The implementation is based on the Meeting Status Button.

# Merge checklist

- [x] Added/updated tests

### DIFF
--- a/app/components/projects/status_button_component.rb
+++ b/app/components/projects/status_button_component.rb
@@ -87,7 +87,7 @@ class Projects::StatusButtonComponent < ApplicationComponent
       tag: :a,
       href: project_status_path(project, status_code: status.value),
       content_arguments: {
-        data: { turbo_method: :put },
+        data: { turbo_method: status.value ? :put : :delete },
         aria: { current: (true if status == @status) }
       }
     )
--- a/app/components/projects/status_button_component.rb
+++ b/app/components/projects/status_button_component.rb
@@ -1,0 +1,95 @@
+# frozen_string_literal: true
+
+#-- copyright
+# OpenProject is an open source project management software.
+# Copyright (C) the OpenProject GmbH
+#
+# This program is free software; you can redistribute it and/or
+# modify it under the terms of the GNU General Public License version 3.
+#
+# OpenProject is a fork of ChiliProject, which is a fork of Redmine. The copyright follows:
+# Copyright (C) 2006-2013 Jean-Philippe Lang
+# Copyright (C) 2010-2013 the ChiliProject Team
+#
+# This program is free software; you can redistribute it and/or
+# modify it under the terms of the GNU General Public License
+# as published by the Free Software Foundation; either version 2
+# of the License, or (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program; if not, write to the Free Software
+# Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+#
+# See COPYRIGHT and LICENSE files for more details.
+#++
+
+class Projects::StatusButtonComponent < ApplicationComponent
+  include ApplicationHelper
+  include OpTurbo::Streamable
+  include OpPrimer::ComponentHelpers
+  include ProjectStatusHelper
+
+  attr_reader :project, :user
+
+  def initialize(project:, user:, size: :medium)
+    super
+
+    @project = project
+    @user = user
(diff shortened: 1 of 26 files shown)


config/locales/js-en.yml
```diff
@@ -474,6 +474,7 @@ en:
     label_created_on: "created on"
     label_edit_comment: "Edit this comment"
     label_edit_status: "Edit the status of the work package"
+    label_edit_project_status: "Edit the status of the project"
```
HDinger: This should be part of the normal `en.yml` file. The `js-en.yml` should ideally only contain strings that need to be accessed from Angular.

app/views/highlighting/styles.css.erb
```diff
@@ -9,6 +9,10 @@
 <%= resource_color_css("meeting_status", meeting_status) %>
 <% end %>
 
+<% Projects::Statuses::AVAILABLE.each do |project_status| %>
+<%= resource_color_css("project_status", project_status) %>
```
HDinger, line 13: Missing indentation
```suggestion
  <%= resource_color_css("project_status", project_status) %>
```

app/components/projects/status_button_component.rb
```diff
@@ -0,0 +1,101 @@
...
+          items: build_items,
+          readonly: !edit_enabled?,
+          disabled: !edit_enabled?,
+          button_arguments: {
+            title: I18n.t("js.label_edit_project_status"),
+            size: @size
+          },
+          menu_arguments: {
+            select_variant: :single
+          }
```
HDinger: I think you can remove that and fallback to the default selection strategy which is `:none`. Then you can get rid of the additional CSS file and class you introduced.

app/components/op_primer/status_button_component.sass
```diff
@@ -0,0 +1,2 @@
+.op-status-button .ActionListItem-action--leading
```
HDinger: I really don't like that, and I think it is not needed if you remove the `select_variant` from the menu_arguments of the Button (see my other comment).

HDinger (changes requested):
Works nicely 👍  I only have minor remarks.

app/views/highlighting/styles.css.erb
myabc: @HDinger I'll keep this as-is, as I think want to avoid extra indentation in the generated CSS.

myabc:
I've changed the base to `release/16.0` - this PR failed its first review before the release cut-off, but I am hoping we can still get this in to 16.0? I have addressed @HDinger's feedback.

In addition to making the changes requested by @HDinger, I have also improved/fixed error handling (the previous iteration of this PR ignored the unhappy path 🫨):

- ensuring the user is notified of both successful and failed updates via flash message.
- only rendering the Status Button with updated status if the update is successful.
- adding negative Controller specs.

image

image

app/components/projects/status_button_component.rb
```diff
@@ -0,0 +1,95 @@
...
+      name: project_status_name(status.code),
+      color_namespace: :project_status,
+      color_ref: status.id,
+      icon: status.icon,
+      item_id: status.id,
+      tag: :a,
+      href: project_status_path(project, status_code: status.value),
+      content_arguments: {
+        data: { turbo_method: status.value ? :put : :delete },
+        aria: { current: (true if status == @status) }
```
myabc, line 91: @bsatarnejad FYI. ActionMenu options use the `menuitem` role, so I think `aria-current` is appropriate here (vs. `aria-selected` or `aria-checked`) (we need to manually apply this ARIA attribute because we are using the [default `select_variant: :none`](https://primer.style/product/getting-started/rails/components/action_menu/#arguments))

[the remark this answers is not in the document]
myabc: since we are only interested in rendering `aria-current="true"`, we want this to evaluate to `true || nil`, not `true || false`.

HDinger (approved):
Lgtm 👍  I changed the base branch however back to back, since the feature freeze is long through.

HDinger:
mhh.. okay, I talked with QA about this again, and especially the part of the error handling is considered crutial to have in 16.0. In order, to avoid duplicate work, we will make an exception for this and merge it into 16.0

PR #18658, opened 2025-04-16
[63434] Fix user with a particular set of permissions seeing errors when updating work packages dates

cbliard:
# Ticket

https://community.openproject.org/wp/63434

The wrong permissions are checked in the date picker. It always checks for `add_work_packages`, but it should also check for `edit_work_packages` when it's an update.

## Screenshots

image

Use correct contract when calling `SetAttributesService`: either `UpdateContract` when editing an existing work package or `CreateContract` when being with a new work package.

I also revised the routing as there was something odd going on when trying to add a controller spec for `DatePickerController`. I kept having some "no route exists" for `new` action, so I made it a bit more sane by:
* changing route path from `.../datepicker_dialog_content` to `.../date_picker` to match controller name and make things easier to find when navigating code.
* moving `new` action for `date_picker` to the collection level of work_packages scope.
* moving `new` action for `progress` to the collection level of work_packages scope.

# Merge checklist

- [x] Added/updated tests

### DIFF
--- a/app/components/work_packages/date_picker/date_form_component.rb
+++ b/app/components/work_packages/date_picker/date_form_component.rb
@@ -218,9 +218,9 @@ module WorkPackages
         permitted_params = params.merge(date_mode: "range", focused_field:).permit!
 
         if params[:action] == "new"
-          new_work_package_datepicker_dialog_content_path(permitted_params)
+          new_date_picker_path(permitted_params)
         else
-          work_package_datepicker_dialog_content_path(permitted_params)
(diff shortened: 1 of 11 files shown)


cbliard:
Fix the flaky spec in another PR.

oliverguenther (approved):

cbliard:
The failing spec `./modules/boards/spec/features/board_enterprise_spec.rb:63` is a flaky one which has been fixed in another PR.
Merging.

PR #18664, opened 2025-04-16
Avoid SystemStackError on project eager loading wrapper

ulferts:
# Ticket

https://community.openproject.org/wp/63498

Avoid SystemStackError when fetching a larger list of projects via the API v3. The problem arises when having 1000 projects in the system (could also happen sooner) and fetching those via the API by specifying `pageSize=1000`.

For all of the projects, the ancestors are prefetched for efficient rendering in the representer. 

Before, this was done by constructing an AREL object that had a structure like this:

```
Grouping
  OR
    Grouping
      OR
```

So a grouping and an OR was added for every project to the statement.

This crashed but using a similar approach with AR does not seem to be affected by the same problem. This was tested for 2000 projects which ran through without problems.

### DIFF
--- a/lib/api/v3/projects/project_eager_loading_wrapper.rb
+++ b/lib/api/v3/projects/project_eager_loading_wrapper.rb
@@ -51,16 +51,15 @@ module API
           end
 
           def ancestor_projects(projects)
-            ancestor_selector = projects[1..].inject(ancestor_project_select(projects[0])) do |select, project|
-              select.or(ancestor_project_select(project))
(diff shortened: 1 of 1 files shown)


as-op (approved):
👍  Looks good. 
If I understand the large request Project.where is split into several Project.where requests. 
I tried `http://localhost:3000/api/v3/projects?pageSize=1000` successfully with 2300 projects in the my local DB which previously did throw the stack error, but no longer does with this change.

PR #18665, opened 2025-04-16
Extend data model to allow logging time on other resources

klaustopher:
# Ticket
https://community.openproject.org/projects/stream-time-and-costs/work_packages/61533

Extend the data model of the `TimeEntry` model (and journal) to allow being logged on ohter models than work packages. For now Meetings will be the only other allowed resource. So we add a validation for this as well.

## Screenshots

# Merge checklist

- [ ] Added/updated tests


app/models/work_package.rb
```diff
@@ -59,7 +59,7 @@ class WorkPackage < ApplicationRecord
   belongs_to :priority, class_name: "IssuePriority"
   belongs_to :category, class_name: "Category", optional: true
 
-  has_many :time_entries, dependent: :delete_all
+  has_many :time_entries, dependent: :delete_all, as: :entity
```
myabc: Active Record Doctor reports "use `dependent: :destroy` or similar on WorkPackage.time_entries - associated model TimeEntry has callbacks that are currently skipped"

```suggestion
  has_many :time_entries, dependent: :destroy, as: :entity
```

app/models/work_packages/costs.rb
```diff
@@ -31,7 +31,7 @@ module WorkPackages::Costs
 
   included do
     belongs_to :budget, inverse_of: :work_packages, optional: true
-    has_many :cost_entries, dependent: :delete_all
+    has_many :cost_entries, dependent: :delete_all, inverse_of: :entity
```
myabc: "use `dependent: :destroy` or similar on WorkPackage.cost_entries - associated model CostEntry has callbacks that are currently skipped"

```suggestion
    has_many :cost_entries, dependent: :destroy, inverse_of: :entity
```

app/models/work_package/time_entries_cleaner.rb
```diff
@@ -56,7 +56,7 @@ def cleanup_time_entries_before_destruction_of(work_packages,
     end
 
     def update_time_entries(work_packages, action)
-      TimeEntry.where(["work_package_id IN (?)", work_packages.map(&:id)]).update_all(action)
+      TimeEntry.where(["entity_type = ? AND entity_id IN (?)", "WorkPackage", work_packages.map(&:id)]).update_all(action)
```
myabc, line 59: I think you can just do:

```suggestion
      TimeEntry.where(entity: work_packages).update_all(action)
```

app/models/version.rb
```diff
@@ -81,8 +81,8 @@ def estimated_hours
   def spent_hours
     @spent_hours ||= TimeEntry
       .not_ongoing
-      .includes(:work_package)
-      .where(work_packages: { version_id: id })
+      .joins("INNER JOIN work_packages ON entity_type = 'WorkPackage' AND work_packages.id = entity_id")
+      .where(work_packages: { version_id: id }, entity_type: "WorkPackage")
```
myabc, line 97: You shouldn't need the additional filtering in `.where` since you're already joining on `entity_type = 'WorkPackage`, Additionally, table aliasing might things a bit clearer.

```suggestion
      .joins("INNER JOIN work_packages ON work_packages.id = time_entries.entity_id AND time_entries.entity_type = 'WorkPackage'")
      .where(work_packages: { version_id: id })
```

app/models/work_packages/costs.rb
```diff
@@ -125,13 +125,13 @@ def reassign_cost_entries_before_destruction(work_packages, user, ids)
 
         false
       else
-        condition = "work_package_id = #{reassign_to.id}, project_id = #{reassign_to.project_id}"
-        ::WorkPackage.update_cost_entries(work_packages.map(&:id), condition)
+        condition = "entity_type='WorkPackage', entity_id = #{reassign_to.id}, project_id = #{reassign_to.project_id}"
+        ::WorkPackage.update_cost_entries(work_packages, condition)
```
myabc, line 129: TI realise this wasn't your choice of naming, but `condition` is a poor choice - the parameter is named `action`.

Passing a Hash rather than raw SQL string would seem more elegant.

```suggestion
        condition = "entity_type='WorkPackage', entity_id = #{reassign_to.id}, project_id = #{reassign_to.project_id}"
        ::WorkPackage.update_cost_entries(work_packages, { entity_type: "WorkPackage", entity_id: reassign_to.id, project_id: reassign_to.project_id })
```

app/models/work_package/time_entries_cleaner.rb
```diff
@@ -73,7 +73,7 @@ def reassign_time_entries_before_destruction_of(work_packages, user, ids)
 
         false
       else
-        condition = "work_package_id = #{reassign_to.id}, project_id = #{reassign_to.project_id}"
+        condition = "entity_type = 'WorkPackage', entity_id = #{reassign_to.id}, project_id = #{reassign_to.project_id}"
         WorkPackage.update_time_entries(work_packages, condition)
```
myabc, line 77:
```suggestion
        WorkPackage.update_time_entries(work_packages, { entity_type: "WorkPackage", entity_id: reassign_to.id, project_id: reassign_to.project_id })
```

frontend/src/app/features/hal/resources/time-entry-resource.ts
```diff
@@ -39,7 +40,7 @@ export class TimeEntryResource extends HalResource {
 
   comment:Formattable;
 
-  workPackage:WorkPackageResource;
+  entity:WorkPackageResource|MeetingResource;
```
myabc, line 43: Can't we just specify the general type?

```suggestion
  entity:HalResource;
```

frontend/src/app/shared/components/grids/widgets/time-entries/list/time-entries-list.component.ts
```diff
@@ -125,8 +127,14 @@ export abstract class WidgetTimeEntriesListComponent extends AbstractWidgetCompo
...
   }
 
-  public workPackagePath(entry:TimeEntryResource):string {
-    return this.pathHelper.workPackagePath(idFromLink(entry.workPackage.href));
+  public entityPath(entry:TimeEntryResource):string {
+    if (entry.entity instanceof WorkPackageResource) {
+      return this.pathHelper.workPackagePath(idFromLink(entry.entity.href));
+    } if (entry.entity instanceof MeetingResource) {
+      return this.pathHelper.meetingPath(idFromLink(entry.entity.href));
+    }
```
myabc, line 135: nitpick:

```suggestion
    const id = idFromLink(entry.entity.href);
    if (entry.entity instanceof WorkPackageResource) {
      return this.pathHelper.workPackagePath(id);
    } if (entry.entity instanceof MeetingResource) {
      return this.pathHelper.meetingPath(idFromLink(id);
    }
```

modules/costs/app/contracts/time_entries/base_contract.rb
```diff
@@ -54,7 +54,9 @@ def self.model
               unless: Proc.new { spent_on.blank? }
 
     attribute :project_id
-    attribute :work_package_id
+    attribute :entity
```
myabc: Is this necessary? We're already defining `entity` (reader) above in `delegate :entity`.

modules/costs/app/contracts/time_entries/base_contract.rb
```diff
@@ -121,11 +124,11 @@ def validate_activity_active
     end
 
     def work_package_invisible?
-      model.work_package.nil? || !model.work_package.visible?(user)
+      model.entity.nil? || !model.entity.visible?(user)
```
myabc, line 126: `model.entity.nil?` will always be false.

myabc: Since we're delegating `entity`:

```suggestion
      !entity.visible?(user)
```

(this method could probably be inlined)

modules/costs/app/contracts/time_entries/base_contract.rb
```diff
@@ -121,11 +124,11 @@ def validate_activity_active
...
     end
 
     def work_package_not_in_project?
-      model.work_package && model.project != model.work_package.project
+      model.entity && model.project != model.entity.project
```
myabc, line 130:
```suggestion
     project != entity.project
```

(this could also be inlined)

modules/costs/db/migrate/20250416095154_add_entity_to_time_entry.rb
```diff
@@ -0,0 +1,26 @@
+# frozen_string_literal: true
+
+class AddEntityToTimeEntry < ActiveRecord::Migration[8.0]
+  def up
+    add_reference :time_entries, :entity, polymorphic: true, index: true
```
myabc: I'm not sure what our policy on migrations is, but should we be adding indexes concurrently? (doing so will require us to `disable_ddl_transaction!`)

modules/costs/app/models/work_package/abstract_costs.rb
```diff
@@ -34,9 +34,11 @@ def costs_of(work_packages:)
       #   where(work_package: work_packages)
       # However, AR (Rails 4.2) will not expand :includes + :references inside a subquery,
       # which will render the query invalid. Therefore we manually extract the IDs in a separate (pluck) query.
-      wp_ids = work_package_ids work_packages
+      wp_ids = work_package_ids(work_packages)
 
-      filter_authorized(costs_model.where(work_package_id: wp_ids).joins(work_package: :project))
+      scope = costs_model.where(entity_type: "WorkPackage", entity_id: wp_ids).joins(:project)
```
myabc, line 39:
```suggestion
      scope = costs_model.where(entity: work_packages).joins(:project)
```

modules/costs/app/models/cost_entry.rb
```diff
@@ -27,8 +29,12 @@
 #++
 
 class CostEntry < ApplicationRecord
+  ALLOWED_ENTITY_TYPES = %w[WorkPackage].freeze
+
+  self.ignored_columns += [:work_package_id]
```
myabc: `Entry::DeprecatedAssociation` already adds this column.

modules/costs/app/controllers/costlog_controller.rb
```diff
@@ -57,7 +57,7 @@ def create
     elsif @cost_entry.save
 
       flash[:notice] = t(:notice_cost_logged_successfully)
-      redirect_back_or_default work_package_path(@cost_entry.work_package)
+      redirect_back_or_default work_package_path(@cost_entry.entity)
```
myabc: How about using `polymorphic_path` here?

modules/costs/app/controllers/costlog_controller.rb
```diff
@@ -161,7 +162,14 @@ def update_cost_entry_from_params
...
-    params.require(:cost_entry).permit(:work_package_id, :spent_on, :user_id,
-                                       :cost_type_id, :units, :comments)
+    params
+      .require(:cost_entry)
+      .permit(:entity_id, :entity_type, :spent_on, :user_id,
+              :cost_type_id, :units, :comments).tap do |permitted_params|
+      if params[:cost_entry][:work_package_id]
+        permitted_params[:entity_id] = params[:cost_entry][:work_package_id]
+        permitted_params[:entity_type] = "WorkPackage"
+      end
```
myabc: AFAICT only `modules/costs/app/views/costlog/edit.html.erb` is sending the `work_package_id` param. Can we update the form params instead?

modules/costs/app/views/costlog/edit.html.erb
```diff
@@ -54,9 +54,9 @@ See COPYRIGHT and LICENSE files for more details.
                      autofocus: true,
                      container_class: "-xslim" %>
 
-    <% if @cost_entry.work_package %>
+    <% if @cost_entry.entity %>
       <div class="form--field-instructions">
-        <%= h(@cost_entry.work_package.to_s) %>
+        <%= h(@cost_entry.entity.to_s) %>
```
myabc, line 60: We define `WorkPackage#to_s` but this probably isn't what we want for other entity types.

modules/costs/config/locales/en.yml
```diff
@@ -99,6 +99,9 @@ en:
     units: "Units"
     valid_from: "Valid from"
     fixed_date: "Fixed date"
+    entity: Entity
+    entity_id: Entity
```
myabc: nitpick: I don't think it's necessary to define `entity_id` as well

modules/costs/spec/services/time_entries/set_attributes_service_spec.rb
```diff
@@ -227,11 +229,11 @@
   context "with project not specified" do
     let(:params) do
       {
-        work_package:
+        entity: work_package
       }
     end
 
-    it "sets the project to the work_package's project" do
+    it "sets the project to the entities project" do
```
myabc:
```suggestion
    it "sets the project to the entity's project" do
```

modules/costs/spec/requests/api/time_entries/create_resource_spec.rb
```diff
@@ -87,6 +87,37 @@
           expect(time_entry.start_time).to eq(13 * 60) # 12:00 UTC = 13:00 Berlin on December 24th, 2024
           expect(time_entry.time_zone).to eq("Europe/Berlin")
           expect(time_entry.user).to eq(german_user)
+          expect(time_entry.entity).to eq(work_package)
+        end
+
+        context "with the deprecated workPackge field" do
```
myabc: (typo)
```suggestion
        context "with the deprecated workPackage field" do
```

modules/costs/spec/requests/api/cost_entries/cost_entry_resource_spec.rb
```diff
@@ -52,44 +56,44 @@
...
+      context "when ost entry is not his own" do
         it_behaves_like "error response",
                         403,
                         "MissingPermission",
                         I18n.t("api_v3.errors.code_403")
       end
 
-      context "cost entry is his own" do
-        let(:cost_entry) { create(:cost_entry, project:, user: current_user) }
+      context "when cost entry is his own" do
```
myabc: we should probably use gender-neutral language here:

```suggestion
      context "when cost entry is their own" do
```

modules/costs/spec/models/queries/time_entries/time_entry_query_spec.rb
```diff
@@ -133,19 +133,22 @@
...
+      instance.where("entity_id", "=", ["1"])
     end
 
     describe "#results" do
       it "is the same as handwriting the query" do
+        # we need to build the exact same query as the filter, so we cannot optimize for the cop
         expected = base_scope
-                   .where(["time_entries.work_package_id IN (?)", ["1"]])
+         .where("time_entries.entity_type IN (?)", "WorkPackage") # rubocop:disable Rails/WhereEquals
+          .where("time_entries.entity_id IN (?)", ["1"]) # rubocop:disable Rails/WhereEquals
```
myabc, line 151: nitpick: indenting is a bit wonky

modules/costs/app/models/cost_entry.rb
```diff
@@ -38,20 +44,24 @@ class CostEntry < ApplicationRecord
...
+  before_save :before_save
   validate :validate
 
-  scope :on_work_packages, ->(work_packages) { where(work_package_id: work_packages) }
+  validates :entity, :project_id, :user_id, :logged_by_id, :cost_type_id, :units, :spent_on, presence: true
+  validates :units, numericality: { allow_nil: false, message: :invalid }
+  validates :comments, length: { maximum: 255, allow_nil: true }
+  validates :entity_type,
+            inclusion: { in: ALLOWED_ENTITY_TYPES },
+            allow_blank: true
```
myabc, line 55: I suspect there are some duplicate validations here.

modules/costs/app/models/time_entries/scopes/ongoing.rb
```diff
@@ -37,7 +37,12 @@ def ongoing
 
       def visible_ongoing(user = User.current)
         TimeEntry
-          .where(work_package_id: visible_work_packages(user).select(:id), user:, ongoing: true)
+          .where(
+            entity_type: "WorkPackage",
+            entity_id: visible_work_packages(user).select(:id),
+            user:,
+            ongoing: true
+          )
```
myabc, line 45:
```suggestion
          .where(
            entity: visible_work_packages(user),
            user:,
            ongoing: true
          )
```

myabc (changes requested):
Hi @klaustopher,

Sorry for taking so long to review this PR. I realise this has been a considerable amount of work for you! Although I still want to do some more testing from a user's perspective and take a closer look at the GlobalID handling, I'll go ahead and submit this first round review.

## UI/UX stuff

possibly a question for  @psatyal or the UI/UX Open Door, but do we want to use the term "Entity" in the UI? Something like "Time logged for" might be better.

## Spec coverage

  - it would be good to add some specs for `CostEntry` and `TimeEntry`. In particular examples for:
    - ~~the accessor overrides~~ ✅ added for the overriden setters
    - ~~validations (from cursory glance it looks like there might be some duplicate validations - Shoulda Matchers can help untangle these).~~ ✅ added for the things I added
    - ~~deprecated `work_package` association/accessors.~~ ✅ added
  - ~~also some representer specs for the deprecated `work_package` associated resource, until we get rid of it.~~ ✅ Those specs exist: 405a31067731f3b8da324900e15952ca1e63e887 and 9e7baa1b9ce8ae881940ec6a3cff3171bfb96639
  - ~~and a spec for `JournalFormatter::PolymorphicAssociation`.~~ ✅ added
  - there are some pending filter specs - I don't know how much work it would be to flesh these out.


## Misc

 - there are a few Rubocop violations that should be fixed.

modules/costs/app/views/costlog/edit.html.erb
myabc: I'm not sure if we're even still using this bit of UI?

(/cost_entries/10/edit is broken for me locally)

app/models/work_package/time_entries_cleaner.rb
klaustopher: For a polymorphic association it is actually better to keep it that way, as your suggestion would generate a query like `(entity_type="WorkPackage" AND entity_id=1) OR (entity_type = "WorkPackage" AND entit_id=2)`. This way the query is more efficient.

app/models/version.rb
klaustopher: Again, an optimization. Yes, as we are inner joining, entries with a type != WorkPackage would be filtered out, but that would only happen after loading all t ime entries. So we can pre filter here.

app/models/work_packages/costs.rb
klaustopher: I opted to just update the code to make it work, and not cause more changes by updating the style. If you feel this should be updated, we can create a code maintenance ticket to fix this.

app/models/work_package/time_entries_cleaner.rb
klaustopher: Same as above, did not want to change the style here.

frontend/src/app/features/hal/resources/time-entry-resource.ts
klaustopher: Why should we do that? There is a limited set of resources it can be, why should we make it more broadly in the JS world?

frontend/src/app/shared/components/grids/widgets/time-entries/list/time-entries-list.component.ts
klaustopher: Is there a benefit from doing it this way or is this just personal preference?

modules/costs/app/contracts/time_entries/base_contract.rb
klaustopher: removed

modules/costs/app/contracts/time_entries/base_contract.rb
klaustopher: Why do you assume `entity.nil?` will always be false? You can call the contract with a time entry that has no entity set.

modules/costs/db/migrate/20250416095154_add_entity_to_time_entry.rb
klaustopher: We are adding a new column that does not have any data in it. So adding the index will not be needed to be done concurrently as there is no data to index.

modules/costs/app/models/work_package/abstract_costs.rb
klaustopher: Again, optimzing the query here.

modules/costs/app/models/cost_entry.rb
klaustopher: You are correct, good catch.

modules/costs/app/controllers/costlog_controller.rb
klaustopher: 👍

modules/costs/app/controllers/costlog_controller.rb
klaustopher: Fixed in 1a7c847ec01

modules/costs/app/views/costlog/edit.html.erb
klaustopher: Yeah, we are using this form to create and update cost entries

Bildschirmfoto 2025-07-04 um 08 47 13

For meetings it is showing the meeting title, so I am okay with this for now. The right thing would be to add an autocompleter there anyways. We will probably do this once we primerize this form

modules/costs/config/locales/en.yml
klaustopher: In some instances we have to reference the ID field. It's good practice to add the translation for both

modules/costs/app/models/cost_entry.rb
klaustopher: That's a tough comment to take work out of. What do you feel is duplicate as I don't see anything

modules/costs/app/models/time_entries/scopes/ongoing.rb
klaustopher: Again, the way I am querying for it is more efficient

app/models/work_package.rb
klaustopher: I did mot want to change anything regarding this logic, but you are correct. Let's see if this causes any weird side effects.

klaustopher:
> there are a few Rubocop violations that should be fixed.

those are pre existing rubocop issues in files that I touched, some of them would be a bigger refactoring, that I do not want to include in this PR

klaustopher:
> possibly a question for @psatyal or the UI/UX Open Door, but do we want to use the term "Entity" in the UI? Something like "Time logged for" might be better.

By not using a "global" translation for the attribute but individually translating it on the 2 models, we can do hat. Renamed to `Logged for` analogue to other attributes

klaustopher:
I have reverted the `delete_all -> destroy` changes. Delete all is correct here, as there are no callbacks as AR doctor has falsely determined.

klaustopher:
@myabc adressed all of your feedback. See inline edits in your comment https://github.com/opf/openproject/pull/18665#pullrequestreview-2982827713

modules/costs/app/contracts/time_entries/base_contract.rb
myabc: See above: https://github.com/opf/openproject/pull/18665/files#diff-0f591a25204ca2455d89cbacf9fbe4fdb37a56bcf490809f125277e1addeef63R95

if `model.entity.is_a?(WorkPackage)` then `model.entity.nil? => false`

modules/costs/app/models/cost_entry.rb
myabc: If you compare the `validates` validations above with the `validate` method below, e.g.

```ruby
validates :entity, :project_id, :user_id, :logged_by_id, :cost_type_id, :units, :spent_on, presence: true

def validate # rubocop:disable Metrics/AbcSize, Metrics/PerceivedComplexity
    # ...
    errors.add :project_id, :invalid if project.nil?
    # ...
    errors.add :entity, :invalid if entity.nil? || (project != entity.project)
    # ..etc
end
```

It's not necessarily something that needs to be addressed in this PR (I realise this is just a change to use the new validations syntax).

modules/costs/app/models/time_entries/scopes/ongoing.rb
myabc: ok, my bad!

modules/costs/db/migrate/20250416095154_add_entity_to_time_entry.rb
myabc: It was understanding that adding an index to a table with a large number of rows will still block writes - regardless of whether the column in question has data/is all `NULL`s or not.

But I'm Not a DBA (IANADBA), so take with a pinch of salt! I imagine @machisuji or @ulferts would know better.

myabc: Can these columns be made `NOT NULL` in a follow up migration?

modules/reporting/app/models/cost_query/result.rb
```diff
@@ -68,6 +68,10 @@ def end_timestamp
       start + self["units"].to_f.hours
     end
 
+    def entity_gid
+      "gid://#{GlobalID.app}/#{self['entity_type']}/#{self['entity_id']}"
```
myabc, line 72: GlobalID provides a low-level method to construct `URI::GID`:

```suggestion
      URI::GID.build([GlobalID.app, *self.values_at('entity_type', 'entity_id'), nil])
```

app/components/activities/item_component.rb
```diff
@@ -125,18 +125,45 @@ def filter_journal_details
...
+      details["#{field}_id"]&.at(index) || @event.journal.journable.public_send("#{field}_id")
+    ]
+  end
+
+  def remove_detail_when_changing_from_empty(details, field)
     details.delete(field) if details[field] && details[field].first.nil?
   end
 
+  def build_gid(entity_type, entity_id)
+    "gid://#{GlobalID.app}/#{entity_type}/#{entity_id}"
```
myabc, line 164:
```suggestion
    URI::GID.build([GlobalID.app, entity_type, entity_id, nil])
```

myabc (approved):
A few minor comments, but looks good overall!

modules/costs/db/migrate/20250416095154_add_entity_to_time_entry.rb
klaustopher: Yes, part of the follow up WP: https://community.openproject.org/work_packages/64975/activity

modules/reporting/app/models/cost_query/result.rb
klaustopher: I'll use the `URI::Gid.create` method: https://community.openproject.org/work_packages/64975/activity

klaustopher: Personally, I like the string construction better here

PR #18668, opened 2025-04-16
avoid error on empty workflow params

ulferts:
# Ticket

https://community.openproject.org/wp/63514

Be robust when sending an empty work flow form.

image

In such a case, the params["status"] is not present so the call to `to_unsafe_h` fails.


ulferts:
The failing feature spec 

```
rspec ./spec/features/notifications/notification_center/split_screen_spec.rb:55
```
cannot have been caused by the change within this PR.

as-op (approved):

PR #18670, opened 2025-04-16
[#63512] Custom User filter registration breaks PROD image build

top4ek:
# Ticket
[OP#63512](https://community.openproject.org/projects/openproject/work_packages/details/63512/overview)

Fix custom filters registration(while modules development)

Use model itself, as it seems works now.


top4ek:
@ulferts can you take a look as I found [your comment](https://github.com/opf/openproject/pull/8973/files#r573521298) here? :). 
May be bumped rails o smthing now plays well and it works.

PR #18672, opened 2025-04-16
Feature/61610 date picker to edit phases and phase gates of a project lifecycle (2/2)

dombesz:
# Ticket
https://community.openproject.org/wp/61610

Change the project phases datepicker to be displayed in an editing modal, disable non working days and make it function more like the work package datepicker modal.

Todo:
- [X] Style the editing link to look like an input field when hovering it.
- [x] Split the date range input field into Start, Finish dates and Duration.
- [x] Display datepicker inside the the edit dialog instead of its own pop up.
- [x] Disable setting the Start date if a previous phase has a Finish date.
- [x] Implement highlighted field logic in order to allow choosing dates for the focused input field.
- [x] Disable selecting dates that are before the previous phase's finish date.
- [x] Add validation that no start and end date are on a Non-Working day.
- [x] Add/update specs.
- [x] Update mobile view to render system datepicker
- [x] Fix bug when sometimes the start date gets filled if no dates are selected and the finish date is focused.
        Have no start date, set finish date, clear the finish date with the button, choose a new finish date. The start date will be set instead of the finish date. Reason is, turbo preview kills the focused field because it re renders on the clearing.
- [x] When typing with a certain speed, characters get dismissed from the date fields due to turbo preview updates

# Merge checklist

- [x] Added/updated tests


app/models/project/phase.rb
```diff
@@ -84,7 +86,14 @@ def range_incomplete?
...
 
   def validate_date_range
-    errors.add(:date_range, :start_date_must_be_before_finish_date) if range_set? && (start_date > finish_date)
+    if range_set? && (start_date > finish_date)
+      date_field, error_message = if finish_date_changed?
+                                    %i[finish_date must_be_after_start_date]
+                                  else
+                                    %i[start_date must_be_before_finish_date]
+                                  end
+      errors.add(date_field, error_message)
```
toy: I think it is simpler to do more directly:
```suggestion
      if finish_date_changed?
        errors.add(:finish_date, :must_be_after_start_date)
      else
        errors.add(:start_date, :must_be_before_finish_date)
      end
```

lookbook/previews/open_project/projects/phase_component_preview.rb
```diff
@@ -35,7 +35,7 @@ def phase_with_both_gates
       phase = Project::Phase.new(id: 1,
                                  definition: definition(start_gate: true, finish_gate: true),
                                  start_date: Date.current - 2.days,
-                                 end_date: Date.current + 2.days)
+                                 finish_date: Date.current + 2.days)
```
ulferts, line 38: Thanks for fixing those here

app/services/project_life_cycle_steps/set_attributes_service.rb
```diff
@@ -29,6 +29,7 @@
 module ProjectLifeCycleSteps
   class SetAttributesService < ::BaseServices::SetAttributes
     def perform(*)
+      model.set_default_start_date
```
ulferts: This would normally be placed in

```
def set_default_attributes(_params)
  self.start_date = model.default_start_date
end
```

which would allow to remove the method from the model.

Ideally, the `default_start_date` method would also be moved to the contract since it is the contract's objective to enforce, and by that also to know about, constraints. But that might be harder to do since it is also referred to by the form.

app/models/project/phase.rb
```diff
@@ -88,7 +90,13 @@ def range_incomplete?
   end
 
   def validate_date_range
-    errors.add(:date_range, :start_date_must_be_before_finish_date) if range_set? && (start_date > finish_date)
+    if range_set? && (start_date > finish_date)
```
ulferts: Not directly related to the code here but rather to the other methods like `#date_range=` and `#range_set?`. I am wondering if we still need them at all or if they should at least be reworked. The UI is now no longer range based (except for showing). The form sends start and finish date except so it feels weird having the range based methods in here.

frontend/src/stimulus/controllers/form-preview.controller.ts
```diff
@@ -41,8 +41,10 @@ export default class FormPreviewController extends Controller<HTMLFormElement> {
     }
 
     const form = this.element;
+    const submitUrl = form.action;
     form.action = this.urlValue;
```
ulferts, line 45: I guess this line can now be deleted as the value is set anew a few lines below.

spec/contracts/project_life_cycle_steps/update_contract_spec.rb
```diff
@@ -137,6 +137,71 @@ def create_phase(**) = create(:project_phase, project:, **)
         it_behaves_like "contract is valid"
       end
     end
+
+    describe "#validate_dates_must_be_on_working_days" do
```
ulferts: Since the method is not actually called, I find that this breaks the black box character of the tests. I would therefore change this to either a flat listing without a describe  block to wrap it in or to use a context block in order to group the elements semantically, but not by method name. So, e.g. 

```
context "(non) working day start and finish dates" do
  ...tests...
end
```

frontend/src/stimulus/controllers/dynamic/overview/project-life-cycles-form.controller.ts
```diff
@@ -28,19 +28,170 @@
...
+
+  private dateToIso(date:Date|null):string {
+    if (date) {
+      return this.timezoneService.utcDateToISODateString(date);
+    }
+    return '';
+  }
+
+  private toDate(date:string|null):Date|null {
+    if (date && /^\d{4}-\d{2}-\d{2}$/.test(date)) {
```
ulferts, line 201: You could use the timezoneService here via:
```
const date = this.timezoneService.parseISODate(date)
if (date.isValid()) {
  return date.toDate();
} else {
  return null;
}
```

Admittedly, it does not safe any lines. The only benefit I see is that you don't repeat the date format here. Not sure if that is worth it.

frontend/src/stimulus/controllers/dynamic/overview/project-life-cycles-form.controller.ts
```diff
@@ -28,19 +28,170 @@
...
+      if ((this.highlightedField === this.finishDateTarget) || this.startDateTarget.disabled) {
+        this.finishDateTarget.value = this.dateToIso(dates[0]);
+        if (!this.startDateTarget.value) {
+          this.highlightField(this.startDateTarget);
+        } else if (this.startDateTarget.value > this.finishDateTarget.value) {
+          this.startDateTarget.value = '';
+        }
+      } else {
+        this.startDateTarget.value = this.dateToIso(dates[0]);
+        this.finishDateTarget.value = '';
```
ulferts: From trying out the feature for a couple of hours, I'd prefer the removal of the finish date not to happen here. This behaviour also differs from the work package date picker. There, the finish date is kept when setting a new start date. I would only remove it if the start date is after the finish date.

frontend/src/stimulus/controllers/dynamic/overview/project-life-cycles-form.controller.ts
```diff
@@ -28,19 +28,170 @@
...
+        this.highlightField(this.finishDateTarget);
+      }
+    } else {
+      this.dateInputFields
```
ulferts: How does this case happen? I tried to have the execution stop at a breakpoint I set in here but did not manage to run into it by clicking around in the UI.

frontend/src/stimulus/controllers/dynamic/overview/project-life-cycles-form.controller.ts
```diff
@@ -28,19 +28,170 @@
...
+      // For mobile, we have to make sure that the active field is scrolled into view after the keyboard is opened
+      field.scrollIntoView(true);
+    }, 300);
   }
 
-  datePickerVisible(element:HTMLElement) {
-    return element.classList.contains('active');
+  private clearHighLight() {
+    this.dateInputFields
+        .forEach((el) => el.classList.remove('op-datepicker-modal--date-field_current'));
```
ulferts: Since the 'op-datepicker-modal--date-field_current' class is references in three methods, I think it safer to move it into a class constant.

frontend/src/stimulus/controllers/dynamic/overview/project-life-cycles-form.controller.ts
```diff
@@ -28,19 +28,170 @@
...
+        },
+      }),
+    );
+  }
+
+  handleFlatpickrDatesChanged(event:CustomEvent<{ dates:Date[] }>) {
+    const dates = event.detail.dates;
+
+    if (dates.length === 1) {
+      if ((this.highlightedField === this.finishDateTarget) || this.startDateTarget.disabled) {
```
ulferts, line 112: When is the second part of the condition (`this.startDateTarget.disabled`) necessary? From clicking around, I couldn't get to a case where the first part is false and only the second part would be true.

frontend/src/stimulus/controllers/dynamic/overview/project-life-cycles-form.controller.ts
```diff
@@ -28,19 +28,170 @@
...
+        .forEach((field, index) => {
+          if (!field.disabled) {
+            field.value = this.dateToIso(dates[index]);
+          }
+        });
+      this.clearHighLight();
+    }
+
+    this.updateFlatpickrCalendar();
+    this.previewForm();
```
ulferts, line 126: This method is complicated to read. But rather than trying to come up with a proposal on how to improve the readability I first have a number of questions regarding it.

frontend/src/stimulus/controllers/dynamic/overview/project-life-cycles-form.controller.ts
```diff
@@ -28,19 +28,170 @@
...
+      return;
     }
 
-    void this.submit();
+    field.classList.add('op-datepicker-modal--date-field_current');
+    this.updateFlatpickrCalendar();
+    window.setTimeout(() => {
+      // For mobile, we have to make sure that the active field is scrolled into view after the keyboard is opened
+      field.scrollIntoView(true);
+    }, 300);
```
ulferts: This leads to an irritating behaviour on non mobil resolutions where the modal is scrolled unnecessarily:

Kapture 2025-05-21 at 09 31 32

which also sometimes happens without the user clicking into the field:

Kapture 2025-05-21 at 09 43 03

Maybe this can be limited to only happen on mobile. Or, we figure out how to prevent scrolling in this element by setting the height of the modal to be the same as the height of its content. Because even when not clicking in the input, the behaviour can be triggered which is irritating:

Kapture 2025-05-21 at 09 37 17

frontend/src/stimulus/controllers/dynamic/overview/project-life-cycles-form.controller.ts
```diff
@@ -28,19 +28,170 @@
...
+
+  preventValueMorphingActiveElement(event:CustomEvent<{ attributeName:string }>) {
+    const target = event.target as HTMLInputElement;
+    const { attributeName } = event.detail;
+    const isActiveElement = this.highlightedField && this.highlightedField.id === target?.id;
+
+    if (isActiveElement && ['value', 'class'].includes(attributeName)) {
+      event.preventDefault();
+    }
+  }
```
ulferts, line 137: Can we extend this to also prevent the morphing of the 'Save' button which flickers with every click on a date:

Kapture 2025-05-21 at 09 38 28

app/forms/projects/life_cycles/form.rb
```diff
@@ -37,42 +43,110 @@ def qa_field_name
...
+    end
 
-      input_attributes = { name: :date_range, value: }
-      if model.duration
-        input_attributes[:caption] =
-          I18n.t("project_phase.duration", count: model.duration)
+    def value(field_name)
+      case field_name
+      when :start_date
+        model.default_start_date || model.start_date_before_type_cast
```
ulferts: This leads to a sometimes weird behaviour. The scenario is as follows (maybe edge case):
* Have three phases with dates set (A, B, C)
* Remove the finish date of the phase in the middle (B)
* => Both dates are no longer displayed in the overview page for the phase whose date was just removed. The dates of the phase after that (C) remain unchanged.
* When clicking on the dates of the last phase (C), to open the date picker for it
* => the start date is not the one shown on the overview page but rather the finish date of the first phase (A) + 1

Kapture 2025-05-21 at 09 54 28

I think keeping the date shown in the overview page, which is the one stored in the database, also for the date shown as the start date in the date picker would be the better approach here. 

While we are on the subject of the overview page. What do you think about always showing the value stored in the database also on the overview page. I mean also in the case where there is only one date set (e.g. by removing the finish date). That would change the behaviour somehow of only showing a complete range but I don't see the harm of it. Maybe also a topic for @marcalcobe to chime in.

frontend/src/app/shared/components/datepicker/wp-date-picker-modal/wp-date-picker-instance.component.ts
```diff
@@ -220,31 +221,38 @@ export class OpWpDatePickerInstanceComponent extends UntilDestroyedMixin impleme
...
+      // eslint-disable-next-line @typescript-eslint/no-misused-promises
+      onDayCreate: async (dObj:Date[], dStr:string, fp:flatpickr.Instance, dayElem:DayElement) => {
+        onDayCreate(
+          dayElem,
+          this.ignoreNonWorkingDays,
+          await this.datePickerInstance?.isNonWorkingDay(dayElem.dateObj),
+          this.isDayDisabled(dayElem),
+        );
+      },
+      minDate: this.minDate,
```
ulferts, line 250: Do you know if it is possible to have a more explicit visualization for the minDate?. What we currently have left me wondering a couple of times as to why I couldn't click on a certain date.

ulferts (changes requested):
Hi @dombesz, solid PR. As is the nature with something like this (and as the work package date picker has shown us), there are quite a number of smaller things we should at least consider to amend. 

I did check iOS as I have a simulator for that running but I have no such thing for android so I did not test that.

On quite a number of the reported issues I don't claim to know the truth so feel free to discuss.

Apart from what I have noted down right at the code there are the following issues:

### Date validation

When inserting invalid dates (or just a string), no error is displayed on saving but rather, the date picker is closed and the value is not applied:

Kapture 2025-05-21 at 10 15 02

### Non working days

You already mentioned that this is an issue so I am just adding it to have a complete checklist when retesting this: 

Subsequent phases have their start date set to the next day when it should be the next non working day.

frontend/src/stimulus/controllers/dynamic/overview/project-life-cycles-form.controller.ts
dombesz: Disabling the save button was added intentionally by providing the submitter button for the form submission. Removing it, since it is confusing.
https://github.com/opf/openproject/blob/32dc0990418725b5f369ef491dcbd9bd523d6c10/frontend/src/stimulus/controllers/form-preview.controller.ts#L46-L47

frontend/src/stimulus/controllers/form-preview.controller.ts
dombesz: This supposed to restore the original submit url after the preview is submitted. Ideally this is not necessary because the turbo response replaces the url automatically. However, it is required for solving an edge case when the preview response is loaded slowly and meanwhile submit button still can be pressed.

Not replacing the url would lead to submitting the form to the preview url again instead of saving it.

frontend/src/stimulus/controllers/dynamic/overview/project-life-cycles-form.controller.ts
dombesz: The problem is, moment.js is too permissive and it would parse even the following dates, and that is not desired in this case. (The work package form datepicker uses the same logic implemented here)

```js
moment("2021-01-213", "YYYY-MM-DD").isValid()
true
moment.utc("2021-01-213", "YYYY-MM-DD").isValid()
true
moment.utc("2021-01-213a", "YYYY-MM-DD").isValid()
true
moment.utc("2021-01a-213a", "YYYY-MM-DD").isValid()
true
moment.utc("2021-01a-213a", "YYYY-MM-DD").toDate()
Date Thu Jan 21 2021 02:00:00 GMT+0200 (Eastern European Standard Time)
```

frontend/src/stimulus/controllers/dynamic/overview/project-life-cycles-form.controller.ts
dombesz: It happens on the first phase that has both start and finish dates editable, admittedly the `if (!field.disabled) {` can be removed.
image

app/models/project/phase.rb
dombesz: Good point, I removed the `date_range=` as it is not being used anymore, however the `range_set?` is in use, I renamed it to `date_range_set?` as it seems like a better method name.

frontend/src/stimulus/controllers/dynamic/overview/project-life-cycles-form.controller.ts
dombesz: It is being activated when clicking the calendar on a freshly opened dialog of a successor phase with a finish date already saved.

frontend/src/stimulus/controllers/dynamic/overview/project-life-cycles-form.controller.ts
dombesz: It got a bit simpler having applied all the modifications.

frontend/src/app/shared/components/datepicker/wp-date-picker-modal/wp-date-picker-instance.component.ts
dombesz: We have a predefined style for disabled days, in order to change it I need to consult the UI team first:

https://github.com/opf/openproject/blob/5dde7f8e02b34ee0ae4fb73ec7f15a02d0e2a6e9/frontend/src/global_styles/content/_datepicker.sass#L31-L35

dombesz: I applied the `--fgColor-disabled` to the disabled days, and it makes them more easy to spot.
image

Dark mode
image

app/forms/projects/life_cycles/form.rb
dombesz: I discussed both issues with Marc:
1. > What do you think about always showing the value stored in the database also on the overview page. I mean also in the case where there is only one date set (e.g. by removing the finish date).

    We agreed to not show the start date alone, because we may even remove the start date at some point in the future.
2. > The start date is not the one shown on the overview page but rather the finish date of the first phase (A) + 1

    This is indeed problematic, since there is no rescheduling happening with a missing previous phase, I came up with the solution of displaying the start date saved in the database. Additionally, since the previous phase does not have a finish date, we cannot align the start date to it, so instead of setting it "automatically" and disabling it, I decided to enable the start field, unless the previous phase has a complete range.

    <details><summary>See the Demo here</summary>
    <p>

    May-27-2025 11-39-17

    </p>
    </details>

app/services/project_life_cycle_steps/set_attributes_service.rb
dombesz: Unfortunately using `set_default_attributes` attributes is not ideal, because it is executed for new objects only.
I created a new callback [`ensure_default_attributes`](https://github.com/opf/openproject/blob/64712f0ee03341161b74396458801e2bf8049455/app/services/base_services/set_attributes.rb#L59-L68) that is executed for both existing and new objects.

frontend/src/stimulus/controllers/dynamic/overview/project-life-cycles-form.controller.ts
dombesz: Done. https://github.com/opf/openproject/blob/64712f0ee03341161b74396458801e2bf8049455/frontend/src/stimulus/controllers/dynamic/overview/project-life-cycles-form.controller.ts#L53

frontend/src/stimulus/controllers/dynamic/overview/project-life-cycles-form.controller.ts
dombesz: Fixed, the main issue was the presence of the scrollbar.

frontend/src/app/shared/components/datepicker/wp-date-picker-modal/wp-date-picker-instance.component.ts
dombesz: For now the decision was made to not make any changes, because the contrast points between the enabled and disabled dates are not sufficient. The UX team will further explore possible colours.

app/forms/projects/life_cycles/form.rb
dombesz: During fixing this issue, I found another similar issue, but that seems to be caused by not rescheduling following phases while the current phase without dates is deactivated.

https://community.openproject.org/wp/64292

ulferts (approved):
Very good @dombesz. Looks and feels really nice now.

I found one [bug related to highlighting](https://community.openproject.org/wp/64322) that can be fixed later on.

frontend/src/stimulus/controllers/dynamic/overview/project-life-cycles-form.controller.ts
toy: moment.js has [strict mode](https://momentjs.com/guides/#/parsing/strict-mode/)

PR #18689, opened 2025-04-22
Fix hours cannot see in Dark theme in time tracking calendar page.

Eric-Guo:

Fix the text color in dark mode

## Screenshots

Before:
image

After:
image

The color `--fgColor-default` suggested in [Primer Guide](https://primer.style/product/primitives/color/#foreground).

# Merge checklist


klaustopher:
Hey @Eric-Guo,

thanks for the hint and the contribution. We do sparsely use inline CSS, instead I have fixed it in the proper `calendar_component.sass` file. I also noticed another bug in dark mode. So I have opened https://github.com/opf/openproject/pull/18690 that will fix this issue.

PR #18699, opened 2025-04-22
[63467] Adds download report action to the health report

mereghost:
### Related Work Package: [OP#63467](https://community.openproject.org/work_packages/63467)

This introduces a text version of the report for ease of download and sharing.

### DIFF
--- a/modules/storages/app/common/storages/peripherals/connection_validators/check_result.rb
+++ b/modules/storages/app/common/storages/peripherals/connection_validators/check_result.rb
@@ -53,6 +53,10 @@ module Storages
         def failure? = state == :failure
         def warning? = state == :warning
         def skipped? = state == :skipped
+
+        def to_h
+          { state: state.to_s, message:, timestamp: timestamp&.iso8601 }.compact
+        end
       end
     end
   end
--- a/modules/storages/app/common/storages/peripherals/connection_validators/check_result.rb
+++ b/modules/storages/app/common/storages/peripherals/connection_validators/check_result.rb
@@ -55,7 +55,7 @@ module Storages
         def skipped? = state == :skipped
 
         def to_h
-          { state: state.to_s, message:, timestamp: timestamp&.iso8601 }.compact
+          { state: state.to_s, message:, timestamp: timestamp&.iso8601 }
         end
       end
     end
--- a/modules/storages/app/common/storages/peripherals/connection_validators/nextcloud/ampf_configuration_validator.rb
+++ b/modules/storages/app/common/storages/peripherals/connection_validators/nextcloud/ampf_configuration_validator.rb
@@ -39,9 +39,10 @@ module Storages
 
           def validate
             register_checks(
-              :files_request, :userless_access, :group_folder_presence, :group_folder_contents
+              :group_folder_app, :files_request, :userless_access, :group_folder_presence, :group_folder_contents
             )
 
+            group_folder_app_checks
             files_request_failed_with_unknown_error
             userless_access_denied
             group_folder_not_found
@@ -53,6 +54,23 @@ module Storages
               fail_check(:userless_access, message(:userless_access_denied))
             else
               pass_check(:userless_access)
+            end
+          end
+
+          def group_folder_app_checks
+            required_version = SemanticVersion.parse(
+              nextcloud_dependencies.dig("dependencies", "group_folders_app", "min_version")
+            )
+
+            capabilities = Registry["nextcloud.queries.capabilities"].call(storage: @storage, auth_strategy: noop).result
+            dependency = I18n.t("storages.dependencies.nextcloud.group_folders_app")
+
+            if capabilities.group_folder_disabled?
+              fail_check(:group_folder_app, message(:missing_dependencies, dependency:))
+            elsif capabilities.group_folder_version < required_version
+              fail_check(:group_folder_app, message(:dependency_version_mismatch, dependency:))
+            else
+              pass_check(:group_folder_app)
             end
           end
 
@@ -103,9 +121,17 @@ module Storages
 
           def files
             @files ||= Peripherals::Registry
-              .resolve("#{@storage}.queries.files")
-              .call(storage: @storage, auth_strategy:, folder: ParentFolder.new(@storage.group_folder))
+                       .resolve("#{@storage}.queries.files")
+                       .call(storage: @storage, auth_strategy:, folder: ParentFolder.new(@storage.group_folder))
           end
+
+          def noop = StorageInteraction::AuthenticationStrategies::Noop.strategy
+
+          def nextcloud_dependencies
+            @nextcloud_dependencies ||= YAML.load_file(path_to_config).deep_stringify_keys
+          end
+
+          def path_to_config = Rails.root.join("modules/storages/config/nextcloud_dependencies.yml")
         end
       end
     end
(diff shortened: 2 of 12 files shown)


modules/storages/app/common/storages/peripherals/connection_validators/check_result.rb
```diff
@@ -53,6 +53,10 @@ def success? = state == :success
         def failure? = state == :failure
         def warning? = state == :warning
         def skipped? = state == :skipped
+
+        def to_h
+          { state: state.to_s, message:, timestamp: timestamp&.iso8601 }.compact
```
NobodysNightmare: :question:  The timestamp is missing for... skipped results?

:green_circle: Structure-wise I usually prefer to have a stable structure and allow `null` values. I.e. I'd be inclined to leave the `compact` call out of here. The result is the same, but you will always see the same keys per entry.

modules/storages/app/common/storages/peripherals/connection_validators/nextcloud/ampf_configuration_validator.rb
```diff
@@ -103,9 +121,17 @@ def managed_project_folder_ids
...
-              .resolve("#{@storage}.queries.files")
-              .call(storage: @storage, auth_strategy:, folder: ParentFolder.new(@storage.group_folder))
+                       .resolve("#{@storage}.queries.files")
+                       .call(storage: @storage, auth_strategy:, folder: ParentFolder.new(@storage.group_folder))
           end
+
+          def noop = StorageInteraction::AuthenticationStrategies::Noop.strategy
+
+          def nextcloud_dependencies
+            @nextcloud_dependencies ||= YAML.load_file(path_to_config).deep_stringify_keys!
```
NobodysNightmare: :yellow_circle: Why do we need `deep_stringify_keys` here? Looking at the YAML file the keys are all encoded as strings already (Yaml does actually support symbol keys, even if I will never get used to their syntax).

Also: :yellow_circle: Since we use it in an assignment, `deep_stringify_keys` (no `!`) should be fine.

modules/storages/app/common/storages/peripherals/connection_validators/nextcloud/ampf_configuration_validator.rb
```diff
@@ -56,6 +57,23 @@ def userless_access_denied
...
+              nextcloud_dependencies.dig("dependencies", "group_folders_app", "min_version")
+            )
+
+            capabilities = Registry["nextcloud.queries.capabilities"].call(storage: @storage, auth_strategy: noop).result
+            app_name = I18n.t("storages.dependencies.nextcloud.group_folders_app")
+
+            if capabilities.group_folder_disabled?
+              fail_check(:group_folder_app, message(:missing_dependencies, dependency: app_name))
+            elsif capabilities.group_folder_version < required_version
+              fail_check(:group_folder_app, message(:app_version_mismatch))
```
NobodysNightmare: :red_circle: This seems to be the wrong translation. As far as I can tell you want to use `group_folder_version_mismatch`. The translation currently used would indicate that the OP integration app has the wrong version.

modules/storages/app/controllers/storages/admin/health_status_controller.rb
```diff
@@ -48,6 +48,14 @@ def admin_or_frame_layout
 
       def show
         @report = Rails.cache.read(validator.report_cache_key)
+
+        respond_to do |format|
+          format.html
+          format.text do
+            filename = "#{@storage.name.underscore}_health_report_#{@report.latest_timestamp.iso8601}.txt"
```
NobodysNightmare: :red_circle: This will fail if the `cache.read` does not return a value (`@report` is nil), which is possible. The HTML version of the page would then render a nice fallback text about regenerating the report, but here we would already fail at generating the file name.

NobodysNightmare (changes requested):

modules/storages/app/common/storages/peripherals/connection_validators/check_result.rb
mereghost: @NobodysNightmare Yeah. We could change it and have skipped values have a timestamp, but for some reason that I can't put into words... it feels weird.

modules/storages/app/common/storages/peripherals/connection_validators/nextcloud/ampf_configuration_validator.rb
mereghost: I copied over the code from the legacy implementations and actually never paid attention to the file itself. 🤣 Good catch.

modules/storages/app/controllers/storages/admin/health_status_controller.rb
mereghost: Good catch. I pretended that this shouldn't happen as there's no button to call the format as text when `@report` is `nil` but this should be handled. =)

modules/storages/app/common/storages/peripherals/connection_validators/nextcloud/ampf_configuration_validator.rb
mereghost: I'll keep the stringify keys just in case.... :P

NobodysNightmare: I am not going to die on that hill, but let me at least write down a short monologue on "just in case programming":

I've seen this a couple of times where an unnecessary nil check or an extra `compact` call are left in the code "just in case", even though there was no plausible path for anything becoming nil ever. To me this puts a burden on future readers of the code, because they will come with the assumption that the person before them thought well about the code they wrote and if they added a nil check, then you - the reader - must be overlooking a way for the thing to become nil. That's terror!

I think the same argument can be made here. A reader can now spend a meaningful amount of time trying to figure out which edge case they are overlooking for the stringification to make sense, when there is none.

Or they spend a few minutes writing a wall of text on a previously resolved discussion xD

Anyways, feel free to resolve this discussion again ;)

modules/storages/app/controllers/storages/admin/health_status_controller.rb
NobodysNightmare: FYI: Unresolving the discussion, since I can't see any change/fix to the problem yet.

The most plausible scenario for me that this could occur in the wild is someone having a tab open for many hours, then receiving a response from OP support that they should download a report and then clicking the button for the report that is long vanished in the background.

mereghost: What do you propose then? Rerun the report and send it? This would be the least intrusive way of solving this, meaning that even if you go directly to the URL you get the "freshest" version of the report.

NobodysNightmare: That'd be the safest thing to do. If we do it for the entire action, this would mean that you can never see a "no report" page.

The only alternative that I can come up with is to make a fallback for the filename generation that works with a `@report == nil`. The generation of the file content already seems to be fit for it. My preferred way is the automatic regeneration, though.

mereghost: I've amended the filename for now.

Once we figure out the best way to persist the report, this will solve itself.

NobodysNightmare (approved):

PR #18701, opened 2025-04-22
[63568] Fix crowdin issue manually until we have a better solution

cbliard:
# Ticket
https://community.openproject.org/wp/63568

Fix crowdin translation file for Arabic and Latvian languages.

Good old `sed` to replace the bad strings with the correct strings.

# Merge checklist

- [ ] Added/updated tests

Please run crowdin action on GitHub once merged.

### DIFF
--- a/.github/workflows/crowdin.yml
+++ b/.github/workflows/crowdin.yml
@@ -81,6 +81,9 @@ jobs:
       - name: "Fix root key in Portuguese crowdin translation files"
         run: |
           script/i18n/fix_crowdin_pt_language_root_key
+      - name: "Fix storages.health.checks.failures.zero key in Arabic and Latvian crowdin translation files"
+        run: |
+          script/i18n/fix_crowdin_storages_health_checks_failures_zero_key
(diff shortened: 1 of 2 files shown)


NobodysNightmare (approved):
This should fix the issue we are seeing and make our CI green again. It also allows translators (assuming a working Crowdin UI) to properly translate this key and us not break their translation.

The only thing I am wondering is whether we should include a way to make the script fail when `sed` has nothing to replace anymore, which would effectively mean that the script is not needed any longer. Self-destructing workarounds :D

Kharonus (approved):

cbliard:
> The only thing I am wondering is whether we should include a way to make the script fail when `sed` has nothing to replace anymore, which would effectively mean that the script is not needed any longer. Self-destructing workarounds :D

Good idea. Feel free to implement it. I've spent enough time on this.

The failing spec `./spec/features/users/index_spec.rb:73` is passing locally. I'll fix it in another PR tomorrow.
Merging now.

PR #18703, opened 2025-04-22
Use `ActiveRecord::Normalization` API

myabc:
# Ticket

## Screenshots

# Merge checklist

- [ ] Added/updated tests

### DIFF
--- a/app/models/ldap_auth_source.rb
+++ b/app/models/ldap_auth_source.rb
@@ -62,8 +62,9 @@ class LdapAuthSource < ApplicationRecord
   validate :validate_filter_string
   validate :validate_tls_certificate_string, if: -> { tls_certificate_string.present? }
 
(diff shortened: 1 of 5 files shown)


app/models/ldap_auth_source.rb
```diff
@@ -54,16 +54,17 @@ def self.unique_attribute
   validates :tls_mode, inclusion: { in: tls_modes.keys }
 
   validates :host, :port, :attr_login, presence: true
-  validates :name, :host, length: { maximum: 60, allow_nil: true }
+  validates :host, length: { maximum: 60, allow_nil: true }
```
klaustopher, line 57: removal of the length validation for name was intended here?

myabc: There is already a length validation defined for `name`.

myabc: I think there is another validation that is also declared twice (I need to double-check still, hence still WIP!)

PR #18707, opened 2025-04-23
[#63566] Order of users is not alphabetical in the details tables of PDF time sheet

as-op:
# Ticket

https://community.openproject.org/work_packages/63566

Sort the details table per user by user name


klaustopher (approved):

as-op:
Thanks again, @klaustopher 💙⭐

PR #18710, opened 2025-04-23
Fix flaky custom_action spec

cbliard:
Failing spec is spec/features/work_packages/custom_actions/custom_actions_spec.rb:140 on job run https://github.com/opf/openproject/actions/runs/14613189364/job/40995333431?pr=18684

It occurs because the custom actions stored in the cache are overwritten with some data after fetching the work package: when the work pacakge is fetched, it has a `links` property with multiple things, including a `customActions` array. This array of custom actions only has a name property, so when the cache is updated, the custom actions are overwritten with only the name property.

As there is only the name property, the description is lost and the title of the button is set to the name instead of being set to the description.

Sometimes it works because the work package is fetched before the action are fetched, and sometimes it doesn't.

The fix is to keep custom actions in the cache if they exist. This probably introduces bugs because if the custom actions array is very detailed, we want to override it, but that's a try.

So the fix applied here is absolutely not clean and should be revisited. It's a basis for a discussion about how we want to handle this.

cc @akabiru and @dombesz as you have already tried to fix this flaky spec in #18646.

I've reproduced it by moving the title property to the component so that we can set a debugger breakpoint. When reloading, you can see the title is correct at first, then the work package is fetched, and the title becomes incorrect. It becomes correct again later if the action is fetched again, like when hovering.

I suspect hovering was a workaround for this issue and that it will not be needed anymore once we find a proper way to preserve custom actions.

### DIFF
--- a/frontend/src/app/core/apiv3/endpoints/work_packages/work-package.cache.ts
+++ b/frontend/src/app/core/apiv3/endpoints/work_packages/work-package.cache.ts
@@ -34,13 +34,16 @@ import { StateCacheService } from 'core-app/core/apiv3/cache/state-cache.service
 import { InjectField } from 'core-app/shared/helpers/angular/inject-field.decorator';
 import { SchemaCacheService } from 'core-app/core/schemas/schema-cache.service';
 import isNewResource from 'core-app/features/hal/helpers/is-new-resource';
+import { CustomActionResource } from 'core-app/features/hal/resources/custom-action-resource';
 
 @Injectable()
 export class WorkPackageCache extends StateCacheService<WorkPackageResource> {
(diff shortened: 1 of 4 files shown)


dombesz (approved):
Thanks @cbliard for taking a look, it works as expected! 👍

cbliard:
Yes, but I'm not happy with the implementation. It looks too specific.

cbliard:
Discussed it with Oliver. We could also have a service in between that loads all `CustomAction`s and that would be immune to being overwritten, but we agreed that it may also not be worth it as this fix is working.

Let's try keeping this fix and see how it goes.

cbliard:
Thansk for the review @dombesz

PR #18711, opened 2025-04-23
[#63518] added error code to check result

Kharonus:
# Ticket
[OP#63518](https://community.openproject.org/work_packages/63518)

- show error codes on health status reports

- check result contains error code
- view component renders error code for failures and warnings


modules/storages/app/common/storages/peripherals/connection_validators/check_result.rb
```diff
@@ -31,27 +31,31 @@
 module Storages
   module Peripherals
     module ConnectionValidators
-      CheckResult = Data.define(:key, :state, :message, :timestamp) do
+      CheckResult = Data.define(:key, :state, :code, :message, :timestamp) do
```
Kharonus: ❓ I'm still working on this, but catching early feedback.

`code`  and `message` is kind of duplicate (as we already know from the specification phase of this feature). 

> [!NOTE]
> Quick reminder, why we want to have both, a message here and an error code for lookup in the docs:
> We want a "quick feedback" on the report itself, and a more sophisticated explanation of the 
> possible reasons and solution options in the external docs.

When thinking about this, I came to the idea that we should reflect that knowledge in the code of the domain. Meaning, WDYT about removing the message from here, and basically pinning a check result to an error code - for failures and warnings. Then, the view component does the lookup for the correct message in i18n with the code and the storage provider type (because messages can slightly differ for storages).

we can even go one step further and state, that we have unique codes only. Means, no code exists for NC and in a different wording for OD. 

What are your takes on that?

NobodysNightmare: Looking up human error messages by a "machine code" that we can also find in docs, sounds good.

I am torn between having different or the same keys for OneDrive and Nextcloud. I actually faced a similar problem in a recent refactoring, where error messages for something like "forbidden" differentiated between "check your drive settings" vs. "check your group folder settings".

...

On the one hand: It doesn't seem to make sense to distinguish things like "forbidden" or "not found" by a provider, which means IMO the codes should be the same.

On the other hand: The specific hints we can give will probably be different per provider, so our docs need to accomodate for that. Here it could be helpful if we ensured that error codes are not duplicated across providers, so that users can't accidentially read the error code up in the wrong section of the docs...

Kharonus: Yeah, that were my thoughts, too. The error codes and related messages, explanations and solutions seem pretty specific for each provider.

modules/storages/config/locales/en.yml
```diff
@@ -269,40 +269,36 @@ en:
...
+        od_drive_id_not_found: The configured drive id could not be found. Please check the configuration.
+        od_existing_test_folder: The folder name needed for testing already exists. Please delete it and try again.
+        od_oauth_request_not_found: OpenProject couldn't find the root folder in the drive using the current logged-in user. Please check the server logs for further information.
+        od_oauth_request_unauthorized: The current user isn't authorized to see the root folder of OneDrive/SharePoint or we couldn't acquire a token. Please check the server logs for further information.
+        od_oauth_token_missing: OpenProject cannot test the user level communication with OneDrive/SharePoint as the user did not yet link their Microsoft account.
+        od_tenant_id_wrong: The configured directory (tenant) id is invalid. Please check the configuration.
+        od_unexpected_content: Unexpected content found in the drive.
         unknown_error: The connection could not be validated. An unknown error occurred. Please check the server logs for further information.
```
Kharonus, line 305: ❓ so, next question:

are those error codes we are fine with? Those codes translate to `ERR_NC_OAUTH_TOKEN_MISSING`  and `WRN_OD_UNEXPECTED_CONTENT` and `ERR_NOT_CONFIGURED`. The prefix is derived from the `state`  obviously. ;)

if we think them nice and cool, I'd amend the docs.

\/cc @NobodysNightmare @ba1ash @mereghost

modules/storages/app/common/storages/peripherals/connection_validators/check_result.rb
mereghost: I'm soft against removing the error messages completely. Why? We have a place where we execute tests and can give immediate feedback on what is wrong to the user with no need to jump to another place to figure out what a `wrn_dendency_version` error means. It is not that we are an appliance that can only give error codes on a 7 segment display.

If anything the message should give enough information to the user that it can figure out what's wrong and where. Jumping to docs should be the last thing they would need to do (given they have no clue what to do).

mereghost: Now about the error codes, let's set some prerequisites here: 
1. Error / Warning messages are good enough to help the user to figure out what's wrong.
2. They link to a relevant area of setup procedure of the docs (which would be available for an external contribution, mind you)

What benefit would they bring in this case? I saw it in the past as an excuse for the bad docs (and reporting space) we had at the time.

NobodysNightmare: > What benefit would they bring in this case?

I think the main use case for an error code is that it's translation independent. So if you are running an arabic OpenProject instance and see an error, you still can map that to the correct troubleshooting steps in the (english) docs. And **we** can still recognize the error message on a screenshot. But I agree that copy&pasting the error code should not be generally required for a good product experience.

Kharonus: I'd second Jan's point.

Error codes are not internationalized, making them stable - and thus also not prone to translation mistakes.

Messages will suffer from being incorrect through to people translating it, who does not have full knowledge about our code and its structure.

But there is an even more important point. I see errors and the reporting of those as a thing of three steps.

1. Tell me what the error is.
2. Tell me why the error happens.
3. Tell me how I can fix this dump shit.

Our error messages are only covering 1 - maybe they can cover 2 or 3 in a very short way, but usually we do not have the space - nor the desire - to go into much detail. And maybe - getting those numbers out of thin air - 80% of admins looking at 1 will directly know how to fix it and they never have to look at 2 or 3. But there might be cases, where the admin can't figure out the problem and he wants more leads. Hence he looks at 2 and 3.

My example? The very early check for "is NC host reachable". In the error message we tell 1: We can't reach the host under the defined URL. But alone the 2 is incredibly diverse. It could be a badly configured proxy, it could be a simple typo.

So, the admin can go on a witchhunt just after seeing 1. But maybe he just can't find the reason. And that is the point in time where he opens the trouble shooting guide. There he sees a (hopefully) elaborate section with codes, descriptions, and different ideas how he could solve this.

Kharonus: > I'm soft against removing the error messages completely.

Maybe there is a misunderstanding. We do not remove the error messages. We simply move the localization to the view component. This has one disadvantage: We loose any context we do not transport in the error code. Right now, I found only one point, where this actually had an effect: We provided the context of the test folder name, if said folder can't get created. So instead of reporting "We can't create folder 'OpenProjectTestFolder'." my proposal reports "We can't create a test folder." But as the message then says: "Check the drive" I deemed the missing context as not critical, because the folder has a name and if there is a conflicting folder, the user would see it.

mereghost (approved):

PR #18714, opened 2025-04-23
[#60730] PDF export of meetings

as-op:
# Ticket

https://community.openproject.org/work_packages/60730

## Screenshots

# Dev checklist

- [x] Background job
- [x] Menu entry 
- [x] Support GIFS (first frame) in PDF Exports
- [x] Support work package links in PDF Exports 
- [x] Export 
   - [x] Page header
   - [x] Participants list
   - [x] Section headers
   - [x] Agenda items
   - [x] Attachment list
   - [x] Outcomes
   - [x] Backlog
- [x] Styling
   - [x] Use paddings & margins from Figma
   - [x] Extract hardcoded style values to yml
   - [x] Outcomes background (not supported for now)
   - [x] Icon in status badge (not supported for now)
- [x] Check for page breaks (e.g. dangling headers)
- [x] Export dialog

# Merge checklist

- [x] Added/updated tests


modules/meeting/app/controllers/meetings_controller.rb
```diff
@@ -538,4 +546,19 @@ def redirect_to_project
...
+      mime_type: :pdf,
+      query: @meeting,
+      options: params.to_unsafe_h
+    )
+    if request.headers["Accept"]&.include?("application/json")
+      render json: { job_id: job.job_id }
+    else
+      redirect_to job_status_path(job.job_id)
+    end
+  end
```
ulferts, line 563: Just an idea: In order to prevent the controller from growing any larger, it might be possible to extract the pdf generation bits into a separate controller.

app/models/work_package/pdf_export/common/attachments.rb
```diff
@@ -59,18 +55,36 @@ def attachment_image_local_file(attachment)
...
 
-    resize_image(local_file.path)
+    filename = local_file.path
+    filename = convert_gif_to_png(filename) if attachment.content_type == "image/gif"
+
+    resize_image(filename)
+  end
+
+  def temp_image_file(extension)
+    tmp_file = Tempfile.new(["temp_image", extension])
```
ulferts, line 73: Not sure if it is actually an issue but this smells like it could lead to collisions on the static name.

app/models/work_package/pdf_export/export/markdown.rb
```diff
@@ -68,19 +69,61 @@ def hyphenate(text)
...
+      count = content.count("#")
+      if count > 1
+        work_package = WorkPackage.find_by(id: id)
+        unless work_package.nil? || !work_package.visible?
+          # ##1234: {Type} #{ID}: {Subject}
+          content = "#{work_package.type} ##{work_package.id}: #{work_package.subject}"
+          if count == 3
+            # ###1234: {Status} {Type} #{ID}: {Subject}
+            content = "#{work_package.status.name} #{content}"
+          end
```
ulferts, line 110: The dates are not included in the export but are included in the web UI. Don't know if that is intentional.

modules/meeting/app/workers/meetings/pdf/page_head.rb
```diff
@@ -0,0 +1,109 @@
...
+
+    def badge_text
+      case meeting.state
+      when "open"
+        I18n.t("label_meeting_state_open")
+      when "in_progress"
+        I18n.t("label_meeting_state_in_progress")
+      when "closed"
+        I18n.t("label_meeting_state_closed")
+      else
```
ulferts, line 81: One could be daring and write:

```
I18n.t("label_meeting_state_#{meeting_state}")
```

modules/meeting/app/workers/meetings/pdf/page_head.rb
```diff
@@ -0,0 +1,109 @@
...
+    def meetings_state_color
+      case meeting.state
+      when "in_progress"
+        Meetings::Statuses::IN_PROGRESS.color
+      when "closed"
+        Meetings::Statuses::CLOSED.color
+      else
+        Meetings::Statuses::OPEN.color
+      end
+    end
```
ulferts, line 99: How about:

```
Meetings::Statuses::AVAILABLE.find { it.id == meeting.state }.color
```

Doesn't have the fallback, admittedly. 

The status discovery could also be turned into a class method of Meeting::Statuses.

ulferts (approved):
The export is working nicely for the most parts @as-op . Admittedly, I did not do a fully thorough review given the volume of the PR and the time available. 

I think this PR can be merged but there are some bugs in here. The information leakage and also the section handling I would consider to be necessary to fix before the release.


### Inconsistent date and time formats

The date and time formats between the cover page and the header on the next page are different:

image

### Section header displayed unnecessarily

There is a section header displayed unnecessarily in case the filter has no sections. It is titled "Untitled section"

image

### Information leak on linked work package

If the meeting has a work package agenda item:

image

and if a user exports the meeting lacking the permission to see the work package, then the meeting displays the agenda point like this:

image

But when exporting the meeting, the title of the work package is revealed:

image
 
### No dots at the end of the sentence

Not a bug as such but given that the captions are full sentences, it is weird that they are not concluded by a '.'

image

app/models/work_package/pdf_export/common/attachments.rb
as-op: I expect no problems.  A unique filename will be created by `Tempfile.new` using "temp_image" as a base.

modules/meeting/app/workers/meetings/pdf/page_head.rb
as-op: ✅

app/models/work_package/pdf_export/export/markdown.rb
as-op: That was an oversight. Fixed ✅

modules/meeting/app/workers/meetings/pdf/page_head.rb
as-op: ✅

PR #18715, opened 2025-04-23
[#63594] Primerize Project copy form

myabc:
⚠️ **~~This PR is based on #18716, #18478. Please review/merge those PRs first.~~**

# Ticket

https://community.openproject.org/wp/63594

Updates the form visuals based on new mockups: the Project copy form now only includes required (built-in and custom) fields, along with the "Subproject of" field.

In detail:

* Introduces `Projects::CopyFormComponent`.
* Adds `Projects::CopyForm` with basic specs.
* Adds `ProjectsController#copy_form` action, permissions and route mappings.
* Updates copy project feature specs to reflect form changes.
* Removes the old Angular `CopyProjectComponent` and remainders of `OpenprojectProjectsModule`.
* Ensures both the create and copy forms correctly handle the scenario where no Copy Options dependencies or associations are selected.

## Screenshots

Screenshot 2025-06-25 at 11 42 23
Screenshot 2025-06-25 at 11 42 40
Screenshot 2025-06-25 at 11 42 55

This work is similar in approach to #18716.

N.B. This PR does not fully implement the design specified in the latest iteration of the mockups - follow up work will take place in [OP 65026](https://community.openproject.org/wp/65026).

# Merge checklist

- [x] Added/updated tests


spec/controllers/projects_controller_spec.rb
```diff
@@ -305,25 +306,90 @@
...
-  describe "#copy" do
+  describe "#copy_form" do
     let(:project) { create(:project, identifier: "blog") }
 
-    it "renders 'copy'" do
-      get "copy", params: { id: project.id }
-      expect(response).to be_successful
-      expect(response).to render_template "copy"
+    shared_examples_for "successful request" do
+      it "renders 'copy_form'", :aggregate_faulures do
```
cbliard: typo: `:aggregate_faulures` => `aggregate_failures`

```suggestion
      it "renders 'copy_form'", :aggregate_failures do
```

spec/controllers/projects_controller_spec.rb
```diff
@@ -305,25 +306,90 @@
...
+      get "copy_form", params: { id: project.identifier }
+    end
+
+    context "as an admin" do
+      it_behaves_like "successful request"
+    end
+
+    context "as a non-admin with copy projects permission" do
+      let(:user) { create(:user, member_with_roles: { project => project_role }) }
+      let(:project_role) { build(:project_role, permissions: [:copy_projects]) }
```
cbliard: This works, but just as a FYI, there is also a `member_with_permissions` parameter available which takes care of creating the `ProjectRole`. I find it more readable as there is less boilerplate code.

It would then look like this:

```ruby
let(:user) { create(:user, member_with_permissions: { project => [:copy_projects] }) }
```

spec/features/projects/copy_spec.rb
```diff
@@ -178,16 +178,11 @@
       end
 
       it "separates optional and required custom fields for new" do
```
cbliard:
```suggestion
      it "renders only required project attributes" do
```

cbliard (changes requested):
Nice 🍍 
I only found one issue:
- if there is an error at submitting the copy like for instance an empty project name, then an error is displayed next screen but the user menu becomes unusable: clicking on the avatar icon in the top-right corner does not display the user menu anymore. I can only reproduce the issue on the feature branch. It works correctly on `dev`.

This is the only reason I'm marking the PR as "Request changes".

For the rest, they are only minor remarks which can be ignored or addressed in a later PR.

Also `js.forms.advanced_settings` I18n key can be removed from `js-en.yml`, and for `js.project.copy.copy_options`, I think it should migrate to `en.yml` even if that means it will have to be retranslated.

spec/controllers/projects_controller_spec.rb
myabc: good catch! thanks

myabc:
@cbliard Thanks again for your feedback! I will have to address the issue with the top menu bar in a separate BUG WP/PR as I realise it also affects other project pages.

> For the rest, they are only minor remarks which can be ignored or addressed in a later PR.

I've addressed these remarks. 

> Also `js.forms.advanced_settings I18n key can be removed from `js-en.yml`, and for js.project.copy.copy_options`, I think it should migrate to `en.yml` even if that means it will have to be retranslated.

Good point - I've noticed that we (as a team) tend to avoid renaming translation strings, but it's good to have strings in the right place. (with translation memories and AI, retranslating shouldn't be a huge issue). I'll make a note to do this as part of #18715 ([OP 65026](https://community.openproject.org/wp/65026)).

cbliard (approved):
Thanks for incorporating my suggestions

cbliard:
Failing feature tests `modules/gantt/spec/features/timeline/timeline_dates_spec.rb:205` and `spec/features/work_packages/display_representations/switch_display_representations_on_mobile_spec.rb:60` are passing locally.

PR #18716, opened 2025-04-23
[#63593] Primerize Project create form

myabc:
⚠️ **~~This PR is based on #18933. Please review/merge first.~~**

# Ticket

https://community.openproject.org/wp/63593
https://community.openproject.org/wp/61534

Updates the form visuals according to new wireframes: the Project create form now only includes required (built-in and custom) fields, along with the "Subproject of" field.

In detail:

* Introduces `Projects::NewComponent` and `TemplateSelectComponent`.
* Adds, extends necessary Primer forms with basic specs.
* Adds `ProjectsController#create` action, routes and permissions.
* Adds `highlight-when-value-selected` Stimulus controller.
* Updates project feature specs for new form, including rewriting some expectations that were reliant on `data-qa-name` attribute that Primer form components do not render by default.
* Removes now obsolete Project Status administation feature spec.
* Removes old ng `NewProjectComponent`.
* Adds `FlashMessagesHelper#render_flash_messages_as_turbo_streams`, including specs (not used).
* Evaluated (but decided against) above solution for rendering flash messages.

ℹ️ Work to primerize the Project Copy form has been split into a separate PR:

* #18715 (OP [63594](https://community.openproject.org/wp/63594))

## Screenshots

### Turbo Frames

I listed some of the trade-offs of the various approaches under "Open Points" on #18478, and have discussed briefly with @klaustopher.

In this PR Turbo Frames seem to get us the most functionality "for free" and with the least amount of custom JavaScript. By applying `data-turbo-frame: "projects-new-component"` to the template selector form, we are able to enable master-detail relationship and refresh the project form fields. Additionally, the setting of `aria-busy` on the frame is automatically taken care of - which would make it relatively straightforward for us to show some sort of loading indicator (e.g. Skeleton Box) in the future.

### Flash messages

~The pattern for displaying flash messages from a Turbo Frame (via `render_flash_messages_as_turbo_streams`) is new to the codebase, although it's similar to [the approach we recommend for triggering an ARIA live update](https://github.com/opf/openproject/blob/dev/lookbook/docs/patterns/accessibility/18-aria-live.md.erb#L63). See the commit message for [8858f66](https://github.com/opf/openproject/pull/18716/commits/8858f66e47b794e4ea45f8ca5ac2edf971b7aa01) for more details.~

~Alternate approaches would have been to forgo Turbo for submission of the new project form (i.e. by setting `:"data-turbo": false`) OR to switch to a Turbo Stream-based approach, calling `render_flash_message_via_turbo_stream` in `ProjectsController`.~

I ended up going with a simpler approach instead: [8bd9620](https://github.com/opf/openproject/pull/18716/commits/8bd9620534ea5f1c942bf2f765ae1b121137275b)

### Known Issues

* https://community.openproject.org/wp/64594

# Merge checklist

Specific issues:

- [X] flash messages on create failure.
- [x] flash messages on create success.
- [x] various broken feature specs (mostly to do with custom field activation - to discuss with @ulferts)
- [x] broken behaviour when project is created from template (backgrounded) and Feedback Dialog is closed before creation is complete.
- [x] custom field help texts not displaying.
- [x] code cleanup: commented code, etc.

Generic items:

- [x] Added/updated tests


app/controllers/projects_controller.rb
```diff
@@ -89,7 +89,73 @@ def index # rubocop:disable Metrics/AbcSize
...
+      end
+
+      # format.turbo_stream do
+      # update_via_turbo_stream(
+      #   component: Projects::NewComponent.new(project: @project, template: @template, copy_options: @copy_options)
+      # )
+      # current_url = url_for(params.permit(:parent_id, :template_id).compact_blank)
+      # turbo_streams << turbo_stream.push_state(current_url)
+      # render turbo_stream: turbo_streams
+      # end
```
HDinger: Can be removed?

app/controllers/projects_controller.rb
```diff
@@ -89,7 +89,73 @@ def index # rubocop:disable Metrics/AbcSize
   end
 
   def new
```
HDinger, line 93: There seem to some issue with the project selector in the top bar on the creation page. It does not load the any text and only shows the arrow:

Bildschirmfoto 2025-06-03 um 14 32 20

app/controllers/projects_controller.rb
```diff
@@ -89,7 +89,73 @@ def index # rubocop:disable Metrics/AbcSize
...
-    render layout: "no_menu"
+    @project = if params[:template_id].present?
```
HDinger: Currently, it looks like I can have either the `parent_id` or `template_id` param. When I have a  param `parent_id` and select a template afterwards, the parent is deselected from the field. Previously, this information was preserved.

Jun-03-2025 14-36-50

HDinger (commented):
As discussed, I focused more on the usage point of the review.

app/controllers/projects_controller.rb
```diff
@@ -89,7 +89,73 @@ def index # rubocop:disable Metrics/AbcSize
...
+                 @template = Project.find_by(id: params[:template_id])
```
klaustopher:
```suggestion
                 @template = Project.visible.find_by(id: params[:template_id])
```

As we are copying data from that project, not using the visible scope here might leak data from the existing project that a user does not have access to. And we should also probably only allow this for templated projects, shouldn't we? But I am not 100% sure about this

```suggestion
                 @template = Project.visible.find_by(id: params[:template_id], templated: true)
```

app/controllers/projects_controller.rb
```diff
@@ -89,7 +89,73 @@ def index # rubocop:disable Metrics/AbcSize
...
+                 @copy_options = Projects::CopyOptions.new
+                 Projects::CopyService
+                   .new(user: current_user, source: @template, contract_options: { validate_model: false })
+                   .call(target_project_params: {}, attributes_only: true)
+                   .result
+               elsif params[:parent_id]
+                 Project.find(params[:parent_id]).children.build
```
klaustopher: same
```suggestion
                 Project.visible.find(params[:parent_id]).children.build
```

app/controllers/projects_controller.rb
```diff
@@ -89,7 +89,73 @@ def index # rubocop:disable Metrics/AbcSize
...
+        flash[:notice] = I18n.t(:notice_successful_create)
+        redirect_to project_path(@project)
+      else
+        render action: :new, status: :unprocessable_entity
+      end
+    end
+  end
+
+  def create_from_template
+    @template = Project.find_by(id: params[:template_id])
```
klaustopher:
```suggestion
    @template = Project.visible.find_by(id: params[:template_id], templated: true)
```

app/controllers/projects_controller.rb
```diff
@@ -162,5 +228,9 @@ def supported_export_formats
     ::Exports::Register.list_formats(Project).map(&:to_s)
   end
 
+  def copy_options_params
+    params.expect(copy_options: [[dependencies: []], :send_notifications])
+  end
+
```
klaustopher: Should probably be moved to `app/models/permitted_params.rb`

app/forms/projects/submit_or_cancel.rb
```diff
@@ -0,0 +1,67 @@
...
+#
+# You should have received a copy of the GNU General Public License
+# along with this program; if not, write to the Free Software
+# Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+#
+# See COPYRIGHT and LICENSE files for more details.
+#++
+
+module Projects
+  class SubmitOrCancel < ApplicationForm
```
klaustopher, line 32: This feels like it should be a general component that we use everywhere where we have a form

klaustopher (commented):
LGTM. When loading projects we should make sure to only load visible ones. People with the `add_project` permission do not have automatic read access to all projects, so this could potentially leak project names since you can pass in any project ID as the template. We should also make sure to only load projects that have been marked as templated. Otherwise this looks good.

app/controllers/projects_controller.rb
myabc: Removed.

app/controllers/projects_controller.rb
myabc: This is fixed (by using a different ivar - `@new_project` in the Controller, ensuring `@project` is always `nil`).

app/controllers/projects_controller.rb
```diff
@@ -135,6 +149,68 @@ def export_list_modal
...
+      redirect_to job_status_path(job.job_id)
+    else
+      @new_project = service_call.result
+      flash[:error] = I18n.t(:notice_unsuccessful_create_with_reason, reason: service_call.message)
+      render action: :new, status: :unprocessable_entity
+    end
+  end
+
+  def find_optional_template
+    @template = Project.visible(current_user).find(params[:template_id]) if params[:template_id].present?
```
klaustopher: This still allows me to send in a project id that is not marked as being a template. The autocompleter does not show it, but we should not rely on the frontend being the only entity controlling this.
```suggestion
    @template = Project.visible(current_user).find_by(id: params[:template_id], templated: true) if params[:template_id].present?
```

myabc: @klaustopher Yes, you're absolutely right. This is now fixed, albeit with a newly-introduced `templated` scope and `find`, rather than `find_by`.

N.B. `find`/`find_by` is probably preferable in this case (or at least for the `create` action) since passing a `template_id` param for a template the current user does not have permissions to view will result in a 404 being rendered - as opposed to attempting to create a blank project.

app/forms/projects/submit_or_cancel.rb
myabc: I agree. However, this may be something I try to do in a follow-up code maintenance ticket.

app/controllers/projects_controller.rb
myabc: fixed

spec/helpers/flash_messages_helper_spec.rb
```diff
@@ -0,0 +1,186 @@
...
+          expect(element).to have_css "template", visible: :all
+        end
+      end
+
+      # N.B. Capybara does not consider <template> contents as part of the document.
+      # As such, the following is not possible:
+      #
+      # it_behaves_like "rendering a banner", "flash", :bell, "zu deiner Information"
+      #
+      # See https://github.com/teamcapybara/capybara/issues/2510
```
myabc, line 149: https://github.com/teamcapybara/capybara/issues/2510

config/locales/en.yml
```diff
@@ -3452,6 +3457,8 @@ en:
   notice_successful_delete: "Successful deletion."
   notice_successful_cancel: "Successful cancellation."
   notice_successful_update: "Successful update."
+  notice_unsuccessful_create: "Creation failed."
+  notice_unsuccessful_create_with_reason: "Creation failed: %{reason}"
   notice_unsuccessful_update: "Update failed."
   notice_unsuccessful_update_with_reason: "Update failed: %{reason}"
```
myabc, line 3475: Not sure about the naming here - these could also be `error_unsuccessful_*` to reflect that they will usually be used with the `error` flash key.

app/forms/projects/copy_options_form.rb
```diff
@@ -0,0 +1,45 @@
...
+# Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+#
+# See COPYRIGHT and LICENSE files for more details.
+#++
+
+module Projects
+  class CopyOptionsForm < ApplicationForm
+    form do |f|
+      f.check_box_group(name: :dependencies, label: I18n.t("js.project.copy.copy_options")) do |group|
+        CopyOptions.all_dependencies.each do |value, label|
```
myabc, line 35: Should we sort these checkboxes by label?

app/helpers/flash_messages_helper.rb
```diff
@@ -34,23 +36,19 @@ module FlashMessagesHelper
...
-      render_flash_message(key, **content)
-    else
-      render_flash_message(key, message: content)
-    end
+  # Renders flash messages wrapped in `<turbo-stream>` tags, suitable for
+  # inclusion inline (e.g. as part of a turbo-frame response).
+  #
+  # @return [String] an HTML-safe string.
+  def render_flash_messages_as_turbo_streams
+    safe_join(build_flash_components.map { it.render_as_turbo_stream(view_context: self, action: :flash) })
```
myabc, line 51: As noted in the PR description, this pattern for displaying flash messages from a Turbo Frame (via `render_flash_messages_as_turbo_streams`) is new. Feedback from @klaustopher @bsatarnejad would be very much appreciated!

app/components/projects/new_component.html.erb
```diff
@@ -0,0 +1,61 @@
...
+
+You should have received a copy of the GNU General Public License
+along with this program; if not, write to the Free Software
+Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+
+See COPYRIGHT and LICENSE files for more details.
+
+++#%>
+
+<%= component_wrapper(tag: "turbo-frame", refresh: :morph, autoscroll: true, data: { turbo_action: :replace, autoscroll_block: :start, autoscroll_behavior: :smooth }) do %>
```
myabc: @HDinger I'm not sure if the autoscroll behaviour is desirable here? 

https://github.com/user-attachments/assets/5bb7a134-1434-403d-976d-bae38e8b6f50

My goal was actually to refocus the form on the first field in the new project form (i.e. Name), but I wasn't able to find a solution that doesn't involve additional, custom JavaScript.

(possibly something for the UX/UI open-door - /cc @marcalcobe )

app/controllers/projects_controller.rb
myabc: @klaustopher we're only calling `copy_options_params` in one place currently. If it's ok, I'll hold off on consolidating parameter permissions until #18715.

app/controllers/projects_controller.rb
```diff
@@ -135,6 +149,68 @@ def export_list_modal
...
+    if service_call.success?
+      redirect_to project_path(@new_project), notice: I18n.t(:notice_successful_create)
+    else
+      flash.now[:error] = I18n.t(:notice_unsuccessful_create_with_reason, reason: service_call.message)
+      render action: :new, status: :unprocessable_entity
+    end
+  end
+
+  def create_from_template # rubocop:disable Metrics/AbcSize
+    @copy_options = Projects::CopyOptions.new(copy_options_params)
```
myabc: FWIW it might have been better to introduce a Form Object here (composed of `CopyOptions` and project params). / @klaustopher

app/forms/projects/settings/custom_fields_form.rb
```diff
@@ -0,0 +1,52 @@
...
+#
+# You should have received a copy of the GNU General Public License
+# along with this program; if not, write to the Free Software
+# Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+#
+# See COPYRIGHT and LICENSE files for more details.
+#++
+module Projects
+  module Settings
+    class CustomFieldsForm < ApplicationForm
```
myabc, line 32: Filing as a TODO for later: This can probably be consolidated with `Projects::CustomFields::Form`.

app/helpers/flash_messages_helper.rb
myabc: ended up going with the much simpler approach of breaking out of frames (see PR description), so this is no longer needed.

app/forms/projects/copy_options_form.rb
oliverguenther: I stumbled upon this during testing the feature. I guess they are consistent with how it was done before, but sorting by label would make more sense to me too. The Project settings > Modules are also sorted alphabetically.

app/helpers/flash_messages_helper.rb
```diff
@@ -34,23 +36,19 @@ module FlashMessagesHelper
...
-    when Hash
-      render_flash_message(key, **content)
-    else
-      render_flash_message(key, message: content)
-    end
+  # Renders flash messages wrapped in `<turbo-stream>` tags, suitable for
+  # inclusion inline (e.g. as part of a turbo-frame response).
+  #
+  # @return [String] an HTML-safe string.
+  def render_flash_messages_as_turbo_streams
```
oliverguenther, line 50: I think this would be an interesting pattern to document when to use this vs the FlashStreamHelper (although, probably a bigger discussion on when to use frames vs streams that is behind this)

config/locales/en.yml
oliverguenther: Leaning towards them being errors, but as we already have unsuccessful_update afterwards, it seems to be inconsistent already. If we do that, we could do it separately for both sets of keys

frontend/package.json
```diff
@@ -108,6 +108,7 @@
     "@primer/live-region-element": "^0.8.0",
     "@primer/primitives": "^9.1.2",
     "@primer/view-components": "npm:@openproject/primer-view-components@^0.68.0",
+    "@stimulus-components/auto-submit": "^6.0.0",
```
oliverguenther, line 111: Nice to re-use the existing helper 👍

oliverguenther (approved):
Functionality-wise and code LGTM. Great to see this huge step towards getting rid of the angular formly implementation 👏

PR #18719, opened 2025-04-24
[#63600] PDF timesheet: Error for TimeEntry with hours set to nil

as-op:
# Ticket

https://community.openproject.org/work_packages/63600

Avoid an error in a sum function with nil values

remove the nil values


klaustopher:
Again, interesting where this is coming from, as hours has been a mandatory field for VERY VERY LONG

klaustopher (approved):

as-op:
> Again, interesting where this is coming from, as hours has been a mandatory field for VERY VERY LONG

Yes, that's why I didn't create test cases for it, but probably should have. Thanks for the review!

PR #18729, opened 2025-04-25
Storage SSO enterprise banner

NobodysNightmare:
This PR removes the feature flag for the token exchanging and replaces it with a proper enterprise check, including a banner.

# Ticket
https://community.openproject.org/wp/60612


modules/storages/spec/features/storages/admin/create_storage_spec.rb
```diff
@@ -189,7 +189,7 @@
       end
     end
 
-    it "renders a Nextcloud specific multi-step form when using OAuth 2.0 SSO", :webmock, with_flag: :oidc_token_exchange do
+    it "renders a Nextcloud specific multi-step form when using OAuth 2.0 SSO", :webmock do
```
klaustopher: you'll need to enable the enterprise feature for the tests

```suggestion
    it "renders a Nextcloud specific multi-step form when using OAuth 2.0 SSO", :webmock, with_ee: [:nextcloud_sso] do
```

modules/storages/spec/features/storages/admin/create_storage_spec.rb
```diff
@@ -91,6 +91,9 @@
 
       aggregate_failures "General information" do
         within_test_selector("storage-general-info-form") do
+          expect(page).to have_enterprise_banner(:corporate)
+          expect(page).to have_selector('option[disabled=disabled][value=oauth2_sso]') # expect SSO option to be disabled
```
NobodysNightmare: :exclamation: Help wanted, this expectation looks pretty ugly :exclamation: 

What I'd prefer to do is something along the lines of (made up code):

```ruby
expect(page).to have_select("Authentication Method").with_disabled_option("Single-Sign-On through OpenID Connect Identity Provider")
```

I.e. something where I can check that the right option in the right select is disabled, but only using the user visible (accessible) names.

This check here works, but that's all the nice things I can say about it.

klaustopher: `option:disabled` also works

NobodysNightmare: That's fair. It at least makes the CSS selector nicer. But I read that you would also keep using a CSS selector for this expectation? (vs. knowing of a much nicer accessible matcher that I am not aware of)

klaustopher: I think the CSS is fine ...If it were a simple form field, we `expect(page).to have_field("something", disabled: true)`. I don't expect it works with options 🤔

ba1ash (approved):
LGTM

NobodysNightmare:
Merging with red pipeline. The only failing spec seems to be a failure unrelated to this PR:

```
rspec ./spec/features/work_packages/table/queries/query_menu_refresh_spec.rb:44
```

PR #18739, opened 2025-04-25
Fix extra relations visible when user adds any other type of relations

cbliard:
# Ticket

https://community.openproject.org/wp/62587

Some relations unrelated to the work package are visible in the relations tab.

# Explanation of the bug

It's related to the way the sql query generated by `Relation.visible` works: it gathers all relations of the work package, take the ids of all work packages of those relations, filter them by user's visibility, and then get all relations where both `to` and `from` are from that list of work packages.

That means that if we have the following work packages:
- work package "Red", id 1
- work package "Green", id 2
- work package "Blue", id 3

And each of them has a relation to the other, we have those relations:
- [1, 2]
- [2, 3]
- [3, 1]

When wanting to display all relations of work package "Red":
- it takes all relations related to "1": [1, 2] and [3, 1]
- it takes all work package ids: [1, 2, 3]
- it checks that they are all visible, and then takes all relations where `from_id` and `to_id` are in [1, 2, 3]
=> that includes [2, 3], even if it's not originally related to work package 1.

That's the bug.

## Screenshots

# Merge checklist

- [ ] Added/updated tests

### DIFF
--- a/app/models/work_packages/relations.rb
+++ b/app/models/work_packages/relations.rb
@@ -42,8 +42,16 @@ module WorkPackages::Relations
         # Since the method looks for relations on the current work package, only work packages
         # that are on either end of a relation with the current work package need to be considered.
         # The number should be quite small compared to the total number of work packages.
-        merge(Relation.visible(user,
-                               work_package_focus_scope: Arel::Nodes::UnionAll.new(select(:from_id).arel, select(:to_id).arel)))
+
+        # merge() method that we will be using below will be overriding all where conditions with the
+        # ones coming in from the other relation. This is not what we want. We want to keep the old constraints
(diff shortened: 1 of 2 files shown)


klaustopher:
Cause here is that we use [#merge](https://api.rubyonrails.org/v7.2.2/classes/ActiveRecord/SpawnMethods.html#method-i-merge)

> For conditions that exist in both relations, those from other will take precedence. To find the intersection of two relations, use [QueryMethods#and](https://api.rubyonrails.org/v7.2.2/classes/ActiveRecord/QueryMethods.html#method-i-and).

This overrides the limitation to only select those for `to` and `from`.

Working on a fix

oliverguenther (approved):

PR #18757, opened 2025-04-28
fix sql alias when ordering&sorting by same table

ulferts:
# Ticket

https://community.openproject.org/wp/63671

When grouping by version (or assignee/responsible) and sorting by a custom field of type version (or user) the generated SQL fails:

```SQL
SELECT
	"work_packages"."id"
FROM
	"work_packages"
	INNER JOIN "statuses" ON "statuses"."id" = "work_packages"."status_id"
	LEFT OUTER JOIN "projects" ON "projects"."id" = "work_packages"."project_id"
	LEFT OUTER JOIN "versions" "versions_work_packages" ON "versions_work_packages"."id" = "work_packages"."version_id"
	LEFT OUTER JOIN (
		SELECT DISTINCT
			ON (CV.CUSTOMIZED_ID) CV.CUSTOMIZED_ID,
			VERSIONS.NAME "value"
		FROM
			"custom_values" CV
			INNER JOIN "versions" VERSIONS ON VERSIONS.ID = CV.VALUE::BIGINT
		WHERE
			CV.CUSTOMIZED_TYPE = 'WorkPackage'
			AND CV.CUSTOM_FIELD_ID = 7
			AND CV.VALUE IS NOT NULL
			AND CV.VALUE != ''
		ORDER BY
			CV.CUSTOMIZED_ID,
			CV.ID
	) CF_ORDER_7 ON CF_ORDER_7.CUSTOMIZED_ID = "work_packages".ID
WHERE
	(
		(STATUSES.IS_CLOSED = FALSE)
		AND (PROJECTS.ID IN (4))
	)
	AND "work_packages"."id" IN (
		SELECT
			"work_packages"."id"
		FROM
			"work_packages"
			INNER JOIN "projects" ON "projects"."id" = "work_packages"."project_id"
		WHERE
			"projects"."active" = TRUE
	)
ORDER BY
	LOWER(VERSIONS.NAME),
	CF_ORDER_7.VALUE DESC NULLS LAST,
	WORK_PACKAGES.ID
LIMIT
	200
OFFSET
	0
```

The failure lies in `ORDER BY LOWER(VERSIONS.NAME) ...` when the alias for the joined versions table is `"versions_work_packages"`. In this case, an alias would not be necessary at all, as the versions table is only joined once. But because of the `INNER JOIN "versions" VERSIONS` within the joined subselect for the ordering on the custom field (id 7 in this case), the somewhat loose regexp in the [`AliasTracker`](https://github.com/rails/rails/blob/v8.0.2/activerecord/lib/active_record/associations/alias_tracker.rb#L39-L40), that inner join is found and wrongfully interpreted as being a join on the main work packages table. Because of this, `AliasTracker` counts to already have a `versions` table joined leading to the 'next' join on `versions` to receive the unnecessary alias. 

By providing a different alias within the inner join, this problem is avoided.

A similar problem existed also for users (assigned_to/responsible).

The other custom field formats should be good since there is no way of causing a conflict with them. 

# Merge checklist

- [x] Added/updated tests

### DIFF
--- a/app/models/custom_field/order_statements.rb
+++ b/app/models/custom_field/order_statements.rb
@@ -75,7 +75,7 @@ module CustomField::OrderStatements
   # Returns the expression to use in SELECT clause if it differs from one used
   # to group by
   def group_by_select_statement
-    return unless field_format == "list" || field_format == "hierarchy"
+    return unless ["list", "hierarchy"].include?(field_format)
 
     # MIN needed to not add this column to group by, ANY_VALUE can be used when
     # minimum required PostgreSQL becomes 16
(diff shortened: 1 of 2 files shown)


ulferts:
@mereghost this is the bugfix you asked for.

mereghost (approved):

PR #18770, opened 2025-04-29
Link to storage SSO docs from storage wizard

NobodysNightmare:
Additionally adding a previously forgotten health check.

# Ticket
* https://community.openproject.org/wp/61839
* where the check should have initially been introduced: https://community.openproject.org/wp/60161

### DIFF
--- a/modules/storages/config/locales/en.yml
+++ b/modules/storages/config/locales/en.yml
@@ -476,6 +476,7 @@ en:
       description: Deactivating this option will hide the attachments list on the work packages files tab. The files attached in the description of a work package will still be uploaded in the internal attachments storage.
       label: Show attachments in the work packages files tab
     storage_audience:
+      documentation_intro: Details on the following options and what needs to be configured in the identity provider can be found in our documentation.
       idp:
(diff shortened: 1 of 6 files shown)


modules/storages/config/locales/en.yml
```diff
@@ -288,6 +289,7 @@ en:
         nc_oauth_token_missing: OpenProject cannot test the user level communication with Nextcloud as the user did not yet link their Nextcloud account.
         nc_unexpected_content: Unexpected content found in the managed group folder.
         nc_userless_access_denied: The configured app password is invalid.
+        no_offline_access_scope: It is recommended to configure the OpenID Connect provider to request the offline_access scope. The integration may still work anyways, but make sure that refresh tokens do not expire.
```
Kharonus: 🟡 minor: I'd suggest renaming the key to `offline_access_scope_missing`, I'd like to make those codes quite similar fashioned, with a adjective in the end, so they have a common pattern and reading experience.

Kharonus (approved):
Only a small plea ;) LGTM otherwise

modules/storages/config/locales/en.yml
NobodysNightmare: done

PR #18772, opened 2025-04-29
Allow more than 2 pg versions to be installed

crohr:
Currently we have:

- pg13 as default
- pg15 on saas

We want to test pg17 on some saas shards, without disturbing pg15 shards. Adding one more version is a small price to pay for the flexibility.


oliverguenther (approved):
Nice idea 👍

PR #18779, opened 2025-04-30
bug/61544 Activity Tab renders the same turbo frame multiple times inside of itself

akabiru:
https://community.openproject.org/work_packages/61544

Move the `component_wrapper` to the top level so that the turbo (stream) replace target is applied directly on the top-level turbo frame element. The target is derived from the component wrapper_key

(see [OpTurbo::Streamable#render_as_turbo_stream](https://github.com/opf/openproject/blob/03c8db36bf7aaa379663ef439d3127441ef33886/app/components/concerns/op_turbo/streamable.rb#L48))

https://github.com/user-attachments/assets/19a47068-a7f6-4122-89f0-3c93f28e1c29

### DIFF
--- a/app/components/concerns/work_packages/activities_tab/stimulus_controllers.rb
+++ b/app/components/concerns/work_packages/activities_tab/stimulus_controllers.rb
@@ -31,11 +31,13 @@
 module WorkPackages
   module ActivitiesTab
     module StimulusControllers
+      module_function
+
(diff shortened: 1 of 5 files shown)


akabiru:
Thanks @brunopagno - fwiw, was waiting for CI to pass before requesting (your) review 😄

HDinger (approved):
Lgtm 👍 Thanks for taking care of this 🙇

PR #18786, opened 2025-04-30
[#63660] show attribute name in resolved subject

Kharonus:
# Ticket
[OP#63660](https://community.openproject.org/work_packages/63660)

- show attribute name in resolved subject

- return array of token resolver objects from property mapper
- resolver reacts on absent context
- different reaction to nil values and not resolvable attributes

### More information

- the first commit contains the change regarding the pattern resolver and the underlying property mapper utility class. Those changes need to be reflected in the pattern input component, which will be part of a next commit.

### DIFF
--- a/app/components/work_packages/types/subject_configuration_component.rb
+++ b/app/components/work_packages/types/subject_configuration_component.rb
@@ -63,9 +63,38 @@ module WorkPackages
         ::Types::Forms::SubjectConfigurationFormModel.new(
           subject_configuration: values[:subject_configuration],
           pattern: values[:pattern],
-          suggestions: ::Types::Patterns::TokenPropertyMapper.new.tokens_for_type(model),
+          suggestions: sort_attributes(supported_attributes),
           validation_errors: model.errors
         )
+      end
+
+      def supported_attributes
+        attribute_tokens = ::Types::Patterns::TokenPropertyMapper.new.tokens_for_type(model)
+
+        result = {
+          work_package: {
+            title: I18n.t("types.edit.subject_configuration.token.context.work_package"),
+            tokens: []
+          },
+          parent: {
+            title: I18n.t("types.edit.subject_configuration.token.context.parent"),
+            tokens: []
+          },
+          project: {
+            title: I18n.t("types.edit.subject_configuration.token.context.project"),
+            tokens: []
+          }
+        }
+
+        attribute_tokens.each_with_object(result) do |token, obj|
+          token_hash = { key: token.key, label: token.label, label_with_context: token.label_with_context }
+          obj[token.context][:tokens] << token_hash
+        end
+      end
+
+      def sort_attributes(attributes)
+        attributes.each_value { |group| group[:tokens] = group[:tokens].sort_by { |a| a[:label] } }
+        attributes
       end
 
       def subject_configuration_form_values
--- a/app/models/types/pattern_resolver.rb
+++ b/app/models/types/pattern_resolver.rb
@@ -40,6 +40,8 @@ module Types
     end
 
     def resolve(work_package)
+      @token_resolver = @mapper.tokens_for_type(work_package.type)
+
       @tokens.inject(@pattern) do |pattern, token|
         pattern.gsub(token.pattern, get_value(work_package, token))
       end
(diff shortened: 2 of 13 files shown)


app/models/types/pattern_resolver.rb
```diff
@@ -49,17 +51,32 @@ def resolve(work_package)
 
     def get_value(work_package, token)
       context = token.context == :work_package ? work_package : work_package.public_send(token.context)
+      return ATTRIBUTE_PLACEHOLDER if context.nil?
+
+      attribute_resolver = @token_resolver.detect { |t| t.key == token.key }
```
NobodysNightmare: :yellow_circle: Not sure how happy I am about the naming of `@token_resolver`. It seems to be simply an array of... things... Apparently an array of attribute resolvers? Couldn't we call it `@attribute_resolvers`? But it is returned by a method called `tokens_for_type`, so maybe it should be called `@tokens`.

app/models/types/patterns/token_property_mapper.rb
```diff
@@ -31,88 +31,82 @@
...
-        }
-      )
-
-      def fetch(key)
-        TOKEN_PROPERTY_MAP.dig(key, :fn) || DEFAULT_FUNCTION.call(key)
-      end
-
-      alias :[] :fetch
+      # rubocop:disable Layout/LineLength
+      BASE_ATTRIBUTE_TOKENS = [
```
NobodysNightmare, line 35: :yellow_circle: Adding to my naming confusion mentioned earlier: Here we have an array called `*_TOKENS` and it is filled with classes of type `AttributeResolver`...

app/models/types/patterns/token.rb
```diff
@@ -37,21 +37,15 @@ def self.build(pattern)
...
-
-        context.to_sym
+        case key.to_s
+        when /^project_/
+          :project
+        when /^parent_/
+          :parent
+        else
+          :work_package
+        end
```
NobodysNightmare, line 48: Where is this `context` method being called? All the usages I found so far seem to call the (similar) implementation on the `AttributeMapper`. Where do we use `Token`?

app/models/types/patterns/token_property_mapper.rb
```diff
@@ -31,88 +31,82 @@
...
-        base = default_tokens
-        base[:work_package].merge!(tokenize(work_package_cfs_for(type)))
-        base[:project].merge!(tokenize(project_attributes, "project_"))
-        base[:parent].merge!(tokenize(all_work_package_cfs, "parent_"))
-
-        base.transform_values { _1.sort_by(&:last).to_h }
+        [
+          *BASE_ATTRIBUTE_TOKENS,
+          *tokenize(work_package_cfs_for(type)),
+          *tokenize(project_attributes, "project_"),
```
NobodysNightmare: :green_circle: What do you think about renaming `project_attributes` to `project_cfs`? Then I would have understood earlier, that this method is (like the ones above and below) only adding custom fields to the static list of attributes, which already included _project attributes_  (not CFs) beforehand.

NobodysNightmare:
Generally looks like an improvement of the code we had before. I believe most of my questions only tough on pieces that come from working and reworking code until it fit.

:+1:

app/models/types/pattern_resolver.rb
Kharonus: yep, agreed.

app/models/types/patterns/token_property_mapper.rb
Kharonus: naming is hard ... :D Will improve with your suggestions

app/models/types/patterns/token.rb
Kharonus: it was used before... maybe it is now obsolete? thanks for the hint, will check.

app/models/types/patterns/token_property_mapper.rb
Kharonus: sadly, project attributes is a synonym for cfs on project level.

But I agree, in here we can name them cfs consitently.

app/models/types/patterns/token_property_mapper.rb
Kharonus: I now differentiate mainly between pattern tokens and attribtue tokens.

The pattern tokens are the simple string representations persisted in a subject pattern bleuprint. The attribute tokens are object sharing the same keys, but containing logic of resolving values and localising labels.

app/models/types/patterns/token.rb
Kharonus: Still used, hopefully with the rename it is now better visible.

app/components/work_packages/types/subject_configuration_component.rb
Kharonus, line 1: ℹ️ preparation of hierarchical tree and sorting of fields now happen in form component.

mereghost (approved):

PR #18788, opened 2025-04-30
why was rb_sys added as dependency of pry 0.15.2?

toy:
Added in d099ed1efa6b (325424a5a0dc of opf/openproject#18528)


machisuji (approved):
👍 🙏

PR #18791, opened 2025-05-01
`ServiceResult` minor code improvements and documentation

myabc:
# Ticket

N/A

* Add comprehensive docs for `ServiceResult` using YARD.
* Some small code improvements to the `ServiceResult` class, including fixing a `Lint/DuplicateMethods` offence.

## Screenshots

# Merge checklist

- [ ] Added/updated tests

### DIFF
--- a/app/services/service_result.rb
+++ b/app/services/service_result.rb
@@ -35,10 +35,10 @@ class ServiceResult
   attr_accessor :success,
                 :result,
                 :errors,
-                :message,
                 :dependent_results
 
-  attr_writer :state
+  attr_writer :message,
+              :state
 
   # Creates a successful ServiceResult.
   def self.success(errors: nil,
--- a/app/services/service_result.rb
+++ b/app/services/service_result.rb
@@ -41,38 +41,16 @@ class ServiceResult
               :state
 
   # Creates a successful ServiceResult.
-  def self.success(errors: nil,
-                   message: nil,
-                   message_type: nil,
-                   state: nil,
-                   dependent_results: [],
-                   result: nil)
-    new(success: SUCCESS,
-        errors:,
-        message:,
-        message_type:,
-        state:,
-        dependent_results:,
-        result:)
+  def self.success(**)
+    new(**, success: SUCCESS)
   end
 
   # Creates a failed ServiceResult.
-  def self.failure(errors: nil,
-                   message: nil,
-                   message_type: nil,
-                   state: nil,
-                   dependent_results: [],
-                   result: nil)
-    new(success: FAILURE,
-        errors:,
-        message:,
-        message_type:,
-        state:,
-        dependent_results:,
-        result:)
+  def self.failure(**)
+    new(**, success: FAILURE)
   end
 
-  def initialize(success: false,
+  def initialize(success: FAILURE,
                  errors: nil,
                  message: nil,
                  message_type: nil,
--- a/app/services/service_result.rb
+++ b/app/services/service_result.rb
@@ -68,7 +68,7 @@ class ServiceResult
     self.dependent_results = dependent_results
   end
 
-  alias success? :success
+  alias success? success
 
   def failure?
     !success?
--- a/app/services/service_result.rb
+++ b/app/services/service_result.rb
@@ -28,28 +28,75 @@
 # See COPYRIGHT and LICENSE files for more details.
 #++
 
+##
+# Represents the result of a call to a service.
+#
+# @example
+#   result = Projects::UpdateService
+#     .new(user: current_user, model: @project)
+#     .call(permitted_params.project)
+#   result.success? # => true if the service call was successful.
+#   result.result # => #<Project id: 1011>
+#   result.errors # => #<ActiveModel::Errors []>
+#
 class ServiceResult
   SUCCESS = true
   FAILURE = false
+  private_constant :SUCCESS, :FAILURE
 
-  attr_accessor :success,
-                :result,
-                :errors,
-                :dependent_results
+  # @return [Boolean] whether the service call was successful.
+  attr_accessor :success
+
+  ##
+  # Returns the result of the service call. In typical usage this will be a
+  # model, i.e. an {ActiveRecord::Base} descendant.
+  #
+  # @return [Object, nil] the service call result object. This can also be nil.
+  attr_accessor :result
+
+  # @return [ActiveModel::Errors] errors resulting from the service call.
+  attr_accessor :errors
+
+  # @return [Array<ServiceResult>] all dependent ServiceResults - by virtue of
+  #   the service calling other services.
+  attr_accessor :dependent_results
 
   attr_writer :message,
               :state
 
   # Creates a successful ServiceResult.
+  #
+  # @param (see #initialize)
+  # @return (see #initialize)
   def self.success(**)
     new(**, success: SUCCESS)
   end
 
+  ##
   # Creates a failed ServiceResult.
+  #
+  # @param (see #initialize)
+  # @return (see #initialize)
   def self.failure(**)
     new(**, success: FAILURE)
   end
 
+  ##
+  # @note
+  #   Prefer using {.success} or {.failure} factory methods to calling
+  #   `ServiceResult.new(success: true)` or `ServiceResult.new(success: false)`.
+  #
+  # @param errors [ActiveModel::Errors, nil] errors resulting from the service
+  #   call.
+  # @param message [String, nil] an error message associated with the service
+  #   call.
+  # @param message_type [#to_sym, nil] the type of error message when displayed
+  #   as a Controller flash message.
+  # @param state [Shared::ServiceState, nil] the Service State object.
+  # @param dependent_results [Array<ServiceResult>] any dependent
+  #   ServiceResults.
+  # @param result [Object, nil] the result of the service call.
+  # @return [ServiceResult] a new, successful instance of ServiceResult.
   def initialize(success: FAILURE,
                  errors: nil,
                  message: nil,
@@ -68,15 +115,24 @@ class ServiceResult
     self.dependent_results = dependent_results
   end
 
+  # @see failure?
+  # @return [Boolean] whether the service call succeeded.
   alias success? success
 
+  # @see success?
+  # @return [Boolean] whether the service call failed.
   def failure?
     !success?
   end
 
   ##
-  # Merge another service result into this instance
-  # allowing optionally to skip updating its service
+  # Merges another ServiceResult into this instance, optionally allowing its
+  # {#success} to be ignored.
+  #
+  # @param other [ServiceResult] the other ServiceResult.
+  # @param without_success [Boolean] whether to ignore the {#success} of the
+  #   other ServiceResult.
+  # @return [void]
   def merge!(other, without_success: false)
     merge_success!(other) unless without_success
     merge_errors!(other)
@@ -84,26 +140,41 @@ class ServiceResult
   end
 
   ##
-  # Print messages to flash
+  # Prints messages to the Controller's flash.
+  #
+  # @param flash [ActionDispatch::Flash::FlashHash]
+  # @return [void]
   def apply_flash_message!(flash)
     if message
       flash[message_type] = message
     end
   end
 
+  ##
+  # Returns all {#result}s, including from dependent ServiceResults.
+  #
+  # @return [Array<Object, nil>] all results.
   def all_results
     dependent_results.map(&:result).tap do |results|
       results.unshift result unless result.nil?
     end
   end
 
+  ##
+  # Returns all {#errors}, including from dependent ServiceResults.
+  #
+  # @return [Array<ActiveModel::Errors>] all errors.
   def all_errors
     [errors] + dependent_results.map(&:errors)
   end
 
   ##
-  # Test whether the returned errors respond
-  # to the search key
+  # Tests whether the returned errors, including from dependent ServiceResults,
+  # include the error key.
+  #
+  # @param attribute [:base, Symbol] the attribute.
+  # @param error_key [Symbol] the type of the error.
+  # @return [Boolean] whether the returned errors include the error key.
   def includes_error?(attribute, error_key)
     all_errors.any? do |error|
       error.symbols_for(attribute).include?(error_key)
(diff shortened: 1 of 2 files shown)


app/services/service_result.rb
```diff
@@ -28,51 +28,68 @@
...
+  # @return [Boolean] whether the Service call was successful.
+  attr_accessor :success
+
+  # Returns the result of the Service call. In typical usage this will be a model, i.e. an {ActiveRecord::Base}
+  # descendant.
+  #
+  # @return [Object, nil] the Service call result object. This can also be nil.
+  attr_accessor :result
+
+  # @return [ActiveModel::Errors] errors resulting from the Service call.
```
NobodysNightmare: I think there are also cases where `errors` is not an `ActiveModel::Errors`, but something simpler like a string or an array of strings<sup>[citation needed]</sup>

NobodysNightmare: Additionally I'd suggest to add `nil` to the possible responses. I am very sure that we don't always add an empty errors representation. Most simple case is `ServiceResult.success(...)`

app/services/service_result.rb
```diff
@@ -28,51 +28,68 @@
...
+  def self.failure(**)
+    new(**, success: FAILURE)
   end
 
-  def initialize(success: false,
+  ##
+  # @api private
+  # @note Prefer using {.success} or {.failure} factory methods to calling {ServiceResult.new}.
+  #
+  # @param errors [ActiveModel::Errors, nil] errors resulting from the Service call.
```
NobodysNightmare: Missing: `@param success`

app/services/service_result.rb
mereghost: In general, the errors should always be an `ActiveModel::Errors` to ensure the same behaviour all around (and facilitate translations).

app/services/service_result.rb
```diff
@@ -90,53 +107,74 @@ def initialize(success: false,
...
   def apply_flash_message!(flash)
     if message
       flash[message_type] = message
     end
   end
 
+  ##
+  # Returns all {#result}s, including from dependent ServiceResults.
+  #
+  # @return [Array] all results.
```
mereghost: I think you can note this as `Array<Object, nil>` to imply we have no idea what the results can be.

app/services/service_result.rb
```diff
@@ -90,53 +107,74 @@ def initialize(success: false,
...
+  #
+  # @return [Array<ActiveModel::Errors>] all errors.
   def all_errors
     [errors] + dependent_results.map(&:errors)
   end
 
   ##
-  # Test whether the returned errors respond
-  # to the search key
+  # Test whether the returned errors include the error key.
```
NobodysNightmare:
```suggestion
  # Test whether the returned errors, including from dependent ServiceResults, include the error key.
```

app/services/service_result.rb
```diff
@@ -162,21 +205,45 @@ def add_dependent!(dependent)
     self.dependent_results += inner_results
   end
 
+  ##
+  # Callback to be executed if the Service call succeeds.
```
NobodysNightmare: Tense of the sentence, as well as the word "Callback" imply that `ServiceResult` is some kind of promise, where the result is yet unknown, but at some point in the future the result will be known and then the callback will be executed.

But actually these methods are intended to replace `if` based control-flow statements.

My take:

```suggestion
  # Execute block argument if the Service call succeeded.
```

Similarly I'd also change the docs for `on_failure` and other following methods.

app/services/service_result.rb
NobodysNightmare: "should" is a stretchable term. A quick search from my side revealed [this beauty](https://github.com/opf/openproject/blob/741e8ab707010017ab0cd99b8125c676778a4cc4/spec/services/bulk_services/project_mappings/behaves_like_bulk_project_mapping_create_service.rb#L154), but I am sure there are more:

```ruby
      service_result = instance.call
      expect(service_result).to be_failure
      expect(service_result.errors).to eq("not found") # "not found" is neither array, nor ActiveModel::Errors
```

app/services/service_result.rb
myabc: @NobodysNightmare that was intentional. Yard copies these input parameter docs for the `.success` and `.failure` factory methods (with the `@param (see #initialize)` syntax), but in the case of those methods `@param success` is not a valid input parameter.

(I couldn't find a way of excluding a param from being copied).

app/services/service_result.rb
```diff
@@ -28,51 +28,68 @@
...
+  # @param (see #initialize)
+  # @return (see #initialize)
+  def self.failure(**)
+    new(**, success: FAILURE)
   end
 
-  def initialize(success: false,
+  ##
+  # @api private
+  # @note Prefer using {.success} or {.failure} factory methods to calling {ServiceResult.new}.
```
myabc: The `@api private` is not correct - after looking at `UseServiceResultFactoryMethods` cop, it looks like we are only discouraging use of `success: true/false` **keywords** in the constructor.

```
      #   # good
      #   ServiceResult.new(success: some_value)
```

is acceptable.

app/services/service_result.rb
myabc: In this case, `all_results` filters out any `nil` results, so this will be `Array<Object>`.

app/services/service_result.rb
myabc: You're right - "callback" implies asynchronous behaviour. Will correct both methods.

app/services/service_result.rb
NobodysNightmare: That's overall sad. I might mainly consume this documentation as inline code comment (not as parsed HTML yard doc) and now the limitations of the parser cause the inline docs to not properly document.

In the end I don't care too much about this, since this PR already provides a huge improvement over the missing docs we had beforehand. But it reminds me of a thing that the OpenStreetMap community refers to as ["tagging for the renderer"](https://wiki.openstreetmap.org/wiki/Tagging_for_the_renderer).

app/services/service_result.rb
mereghost: BTW, I wonder if we could change the implementation to `filter_map` 🤔

app/services/service_result.rb
mereghost: I think that was caused exactly by lack of documentation. `Storages` had exact this issue as the `ServiceResult`s bubbled to the UI, we had a patch up a ton.

I consider it being a `ActiveModel::Errors` the interface, everything else a bug. ;)

app/services/service_result.rb
myabc: I've updated the PR to define the subset of input parameters on `.success` and then copy to `.failure` and the constructor. Admittedly, it's a little less intuitive for readers of code, but better for consumers of the generated docs.

(I'm pretty sure we could accomplish the above with [Yard macros](https://www.rubydoc.info/gems/yard/file/docs/Tags.md#macro) - they certainly allow you to document dynamic methods defined with meta-programming. However my gut feeling is that this would be overkill for this use case.)

app/services/service_result.rb
myabc: @NobodysNightmare @mereghost agree! ~~that's a bug~~ that should be a bug :)

myabc: `service_result_spec.rb` suggests we support arbitrary objects:

```rb
    let(:errors) { ["errors"] }

    it "is what has been provided" do
      instance.errors = errors

      expect(instance.errors).to eql errors
    end
```

mereghost: Git knows how much time I have spent chasing down bugs caused exactly because of this. =/

NobodysNightmare: For an unrelated reason I searched for the implementation of `message`. Turns out that this method is also written in a way that shows `ActiveModel::Errors` is just one possibility for `errors`, but we accept much more.

app/services/service_result.rb
```diff
@@ -162,21 +207,45 @@ def add_dependent!(dependent)
...
+  # @yield block to be called on success.
+  # @yieldparam result [Object, nil] the result of the Service call.
+  # @return [self]
   def each
     yield result if success?
     self
   end
 
+  ##
+  # Returns a new ServiceResult whose {#result} is the return value from the block.
```
cbliard: It's new, but it's also a copy, so the current values of message, state, dependent_results, etc are copied along. And the copy happens only if the service result succeeded.
This should be made clearer
```suggestion
  # If the service call succeeded, returns a copy of the ServiceResult whose
  # {#result} is the return value from the block. Returns self otherwise.
```

I tried this in a `rails console`:
```
>> sr = ServiceResult.success
=>
#<ServiceResult:0x000000013fcf84c0
...
>> sr.add_dependent!(ServiceResult.success)
=>
[#<ServiceResult:0x000000015f591090
  @dependent_results=[],
  @errors=#<ActiveModel::Errors []>,
  @message=nil,
  @message_type=nil,
  @result=nil,
  @state=nil,
  @success=true>]
>> sr
=>
#<ServiceResult:0x000000013fcf84c0
 @dependent_results=
  [#<ServiceResult:0x000000015f591090
    @dependent_results=[],
    @errors=#<ActiveModel::Errors []>,
    @message=nil,
    @message_type=nil,
    @result=nil,
    @state=nil,
    @success=true>],
 @errors=#<ActiveModel::Errors []>,
 @message=nil,
 @message_type=nil,
 @result=nil,
 @state=nil,
 @success=true>
>> new_sr = sr.map { 42 }
=>
#<ServiceResult:0x000000015fdb2530
...
>> new_sr
=>
#<ServiceResult:0x000000015fdb2530
 @dependent_results=
  [#<ServiceResult:0x000000015f591090
    @dependent_results=[],
    @errors=#<ActiveModel::Errors []>,
    @message=nil,
    @message_type=nil,
    @result=nil,
    @state=nil,
    @success=true>],
 @errors=#<ActiveModel::Errors []>,
 @message=nil,
 @message_type=nil,
 @result=42,
 @state=nil,
 @success=true>
>>
```

app/services/service_result.rb
```diff
@@ -185,6 +254,8 @@ def map
     end
   end
 
+  # @return [Array<Object, nil>] the {#result} wrapped in an Array if the Service call succeeds, or an empty Array if
+  #   the Service call fails.
```
cbliard: I find the `nil` disturbing. Sadly there is no way to indicate the size in yard notation.

```suggestion
  # @return [Array<Object>] the {#result} wrapped in an Array if the Service call succeeds, or an empty Array if
  #   the Service call fails.
```

app/services/service_result.rb
```diff
@@ -162,21 +207,45 @@ def add_dependent!(dependent)
...
+  #
+  # @yield block to be called on failure.
+  # @return [self]
   def on_failure(&)
     tap(&) if failure?
     self
   end
 
+  ##
+  # Iterates exactly once, passing the result to the block, if the Service call succeeds.
```
cbliard:
```suggestion
  # Iterates exactly once, passing the result to the block, if the Service call succeeded.
```

Past tense is already used in the above methods, so let's keep it consistent.

cbliard (approved):
Nice 👏 

I like the way `.success` and `.failure` are written now.

I have left some minor remarks.
I would also prefer if documentation lines were no longer than 80 characters. I find it easier to read wen it's shorter.

Also about style, some doc have verbs with a `s` at the end (`# Returns the ...`) and some don't (`Print messages ...`). That could be more consistent, but honestly I don't really care. That's still better than before anyway.

app/services/service_result.rb
myabc: @cbliard `#to_a` wraps `#result`, which can return `Object` or `nil` (we don't currently enforce the presence of a result for a successful Service call[^1], although ofc in practice, result is usually non-nil on success).

After doing a bit of Googling, it looks like think there is a syntax for fixed-sized arrays/tuples, etc. BUT it looks like it's not possible to delimit multiple types ([see discussion here](https://github.com/lsegal/yard/discussions/1579)).

(I am guessing that `Array()` can be used to denote an empty Array, but haven't seen it used in the wild.)

```rb
  # @return [Array(Object), Array()] the {#result} wrapped in an Array if the Service call succeeds, or an empty Array if
```

[^1]: something to add to a list of possible enhancements?

app/services/service_result.rb
myabc: Good point!

app/services/service_result.rb
myabc: I've fixed the tense for all of the param docs.

myabc:
@cbliard thanks for the helpful feedback.

> I would also prefer if documentation lines were no longer than 80 characters. I find it easier to read wen it's shorter.

Fixed (semi-automatically - I'd prefer if if we could set up Rubocop and/or code formatting to enforce this) 

> Also about style, some doc have verbs with a s at the end (# Returns the ...) and some don't (Print messages ...).

Fixed. I've changed all docs to the third-person for consistency.

app/services/service_result.rb
myabc: The previous solution I had pushed (defining input params on `.success` and then adding the `@param (see .success)` tag on `#initialize`) ended up spitting out warnings:

```
[warn]: @param tag has unknown parameter name: errors 
    in file `app/services/service_result.rb' near line 79
[warn]: @param tag has unknown parameter name: message 
    in file `app/services/service_result.rb' near line 79
[warn]: @param tag has unknown parameter name: message_type 
    in file `app/services/service_result.rb' near line 79
[warn]: @param tag has unknown parameter name: state 
    in file `app/services/service_result.rb' near line 79
[warn]: @param tag has unknown parameter name: dependent_results 
    in file `app/services/service_result.rb' near line 79
[warn]: @param tag has unknown parameter name: result 
    in file `app/services/service_result.rb' near line 79
```

It turns out that, in order to define `@param`s for a method that accepts arbitrary kwargs (i.e. `**`), you need to first define an `@overload`.

After trying a couple different combinations of `@overload`, `@param (see ..) and macros, I ended up going for a macro-based solution. 

For doc consumers who read source code, it's not great, but not unintelligible either.

For those who like their docs in HTML, signatures and input parameters are now documented correctly for `.success`, `.failure`, as well as `#initialize`.

image

app/services/service_result.rb
myabc: @mereghost I was thinking the same thing:

```rb
    self_and_dependent.flat_map(&:result)
```

However it would introduce a behaviour change, as the current implementation can contain nil results from dependent ServiceResults. (it might be that that wasn't the intention, but since I am not the original author that is only me speculating)

cbliard:
> > I would also prefer if documentation lines were no longer than 80 characters. I find it easier to read wen it's shorter.
> 
> Fixed (semi-automatically - I'd prefer if if we could set up Rubocop and/or code formatting to enforce this)

Please no.
We should aim for 80 characters because it's more convenient to read, but having a rubocop complaining because of that is really a PITA. That's why we set 120 characters.

The editor can help. On VSCode, I use [rewrap](https://marketplace.visualstudio.com/items?itemName=stkb.rewrap) extension which wraps at my configured editor rulers. I have rulers at 80, 90, 100, 120, and 130 chars. Each time I ask it to wrap, it wraps to 80, then 90, then 100, etc.
I use it all the time.

PR #18806, opened 2025-05-05
Freeze string literals in specs

NobodysNightmare:
Rolling out frozen string literals further by freezing all string literals in core specs.

This follows seeing no further test failures in #18039. The risk of this PR is extra low, because it can't break any live code, just tests (which we would notice by running them).

### DIFF
--- a/spec/components/op_turbo/frame_component_spec.rb
+++ b/spec/components/op_turbo/frame_component_spec.rb
@@ -1,3 +1,5 @@
+# frozen_string_literal: true
+
 #-- copyright
 # OpenProject is an open source project management software.
 # Copyright (C) the OpenProject GmbH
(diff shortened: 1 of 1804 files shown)


klaustopher:
Suggestions, WDYT about adding your commit sha to `.git-blame-ignore-revs` so we don't end up here in a `bisect`and you're blamed for everything :D

klaustopher (approved):

NobodysNightmare:
I still don't know what I think about this file. E.g. freezing strings outside of specs could actually break stuff and such a commit should show up in a bisect.

On the other hand my changes are unlikely to hinder any blame, because you'd have to blame the first line of the file.

**tl;dr:** I'll leave it up to future generations to decide whether my commits for freezing strings should be added to this file. They can use git blame to find my commits ;-P

PR #18810, opened 2025-05-05
[#63776] resolve language for subject patterns

Kharonus:
# Ticket
[OP#63776](https://community.openproject.org/work_packages/63776)

- use user language in subject configuration form
- use system language for nil replacements in wp subjects

- label for an attribute token is now a callback (again)
- use default locale in pattern resolver for nil replacements


ba1ash (approved):
LGTM

PR #18820, opened 2025-05-05
[63739] do not format work and remaining work for xls and csv exports

cbliard:
# Ticket

https://community.openproject.org/wp/63739

In XLS and CSV exports, formatting the work, remaining work, and % complete fields prevents doing calculations on them in the spreadsheet.

This commit removes this formatting ("1d 3.5h" -> "11.5") and adds the totals as separate columns ("1d 2h · Σ 1d 7h" -> "10", "15").

## Screenshots

Before:

image

After:

image

For `estimated_hour`and `remaining_hours` attributes, use the formatter only if the format is pdf, and use the default formatter otherwise.

For `done_ratio` attribute, use a compound percent formatter for pdf ("xx% · Σ yy%") and simple percent formatter for xls and csv ("xx%").

Then override `get_columns` to add related total column when wor, remaining work, or % complete columns are present.

# Merge checklist

- [x] Added/updated tests


klaustopher (approved):
Looks good to me. Should we maybe now include the unit in the column name? Instead of having "Total remaining work" we could call it "remaining work hours"?

But feel fee to merge

cbliard:
> Looks good to me. Should we maybe now include the unit in the column name? Instead of having "Total remaining work" we could call it "remaining work hours"?

Thanks. That would indeed be a sensible thing to do but probably not worth it. And actually I already find the column widths to be larger than needed. Let's keep it like this for now.

cbliard:
Ran failing spec `./spec/features/custom_fields/multi_version_custom_field_spec.rb:97` locally and it passes.
Merging.

PR #18823, opened 2025-05-05
Fix behavior of wrong entries in edit modal when switching timezones

klaustopher:
# Ticket
https://community.openproject.org/projects/stream-time-and-costs/work_packages/63812/activity
https://community.openproject.org/projects/stream-time-and-costs/work_packages/63817/activity

Cecile identified another weird behavior.

- Set user TZ to Europe/Berlin
- Create a time entry for 10:00-13:00
- Switch timezone to something like America/New_York
- Go to my time tracking
- Calendar entry is correctly shown at 4:00-7:00 (since we are displaying everything in the user's new TZ)
- Clicking on the time entry opens the modal and this still shows 10:00-13:00

To be consistent, we also need to adjust the times in the modal. But only when editing your own entries.

## Screenshots

# Merge checklist

- [ ] Added/updated tests

### DIFF
--- a/modules/costs/app/components/time_entries/days_and_hours_form.rb
+++ b/modules/costs/app/components/time_entries/days_and_hours_form.rb
@@ -49,7 +49,7 @@ module TimeEntries
                        type: "time",
                        required: start_and_end_time_required?,
                        label: TimeEntry.human_attribute_name(:start_time),
-                       value: model.start_timestamp&.strftime("%H:%M"),
+                       value: start_time_in_local_time,
                        show_clear_button: true,
                        data: {
                          "time-entry-target" => "startTimeInput",
@@ -60,7 +60,7 @@ module TimeEntries
                        type: "time",
                        required: start_and_end_time_required?,
                        label: TimeEntry.human_attribute_name(:end_time),
-                       value: model.end_timestamp&.strftime("%H:%M"),
+                       value: end_time_in_local_time,
                        show_clear_button: true,
                        caption: end_time_caption,
                        data: {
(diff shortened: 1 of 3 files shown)


modules/costs/app/components/time_entries/days_and_hours_form.rb
```diff
@@ -80,6 +80,26 @@ class DaysAndHoursForm < ApplicationForm
...
     private
 
+    def start_time_in_local_time
+      return if model.start_timestamp.blank?
+
+      if model.user == User.current
+        model.start_timestamp.in_time_zone(User.current.time_zone)
+      else
+        model.start_timestamp
+      end.strftime("%H:%M")
```
cbliard: Shouldn't it always be `model.start_timestamp.in_time_zone(model.user.time_zone)`?

That may be the reason why the times are not correctly shown when editing the time entries for someone else.

cbliard (approved):
The reported bug is fixed, but I think there are still issues with modal displayed for time entries for another user, if that user is in a different time zone than when time was initially logged.

Also there is no unit tests about it.

modules/costs/app/components/time_entries/days_and_hours_form.rb
klaustopher: Yes, I fixed it to exactly this 👍

klaustopher:
Fixed what you mentioned and also added the display of the timezone hint, that you mentioned.

Also fixed [#63817](https://community.openproject.org/projects/stream-time-and-costs/work_packages/63817/activity) here

modules/costs/app/components/time_entries/user_form.rb
```diff
@@ -84,5 +87,13 @@ def user_completer_filters
 
       filters
     end
+
+    def caption
+      if model.user.time_zone == User.current.time_zone
```
cbliard, line 92: So if I have Europe/Paris and you have Europe/Berlin and you log time for me, it will show the message "This user has a different time zone (CEST). Time will be logged using their time zone." even though we are both in CEST?

cbliard (approved):
Nice 👏

modules/costs/app/components/time_entries/user_form.rb
klaustopher: Yeah, ... The problem is ... If for whatever reason you want to log time for before 1940, `Europe/Paris` is equal to GMT, since 1940 it is the same as `Europe/Berlin`... Personally, not a big fan of using the short display of the timezone names (`CET`) ... But well, this is where we are ...

We're also talking about the edgiest of edge cases...

cbliard: Yes, that's good enough I suppose. I too would prefer a more precise time zone indication.

PR #18824, opened 2025-05-05
Feature/59180 activation service that reschedules instead of plain upsert

toy:
# Ticket
[OP#63893](https://community.openproject.org/wp/63893) (part of [OP#59180](https://community.openproject.org/wp/59180), follows #18523)

Automatic rescheduling after activation/deactivation of one or multiple phases

# Merge checklist

- [ ] Added/updated tests

### DIFF
--- a/app/models/project.rb
+++ b/app/models/project.rb
@@ -88,7 +88,7 @@ class Project < ApplicationRecord
   has_many :storages, through: :project_storages
   has_many :phases, class_name: "Project::Phase", dependent: :destroy
   has_many :available_phases,
-           -> { visible.eager_load(:definition).order(position: :asc) },
+           -> { visible.order_by_position },
            class_name: "Project::Phase",
            inverse_of: :project
 
--- a/config/initializers/permissions.rb
+++ b/config/initializers/permissions.rb
@@ -158,6 +158,7 @@ Rails.application.reloader.to_prepare do
                      },
                      permissible_on: :project,
                      require: :member,
+                     dependencies: :edit_project_phases,
                      visible: -> { OpenProject::FeatureDecisions.stages_and_gates_active? }
 
       map.permission :manage_members,
--- a/spec/services/project_life_cycle_steps/activation_service_spec.rb
+++ b/spec/services/project_life_cycle_steps/activation_service_spec.rb
@@ -1,0 +1,388 @@
+# frozen_string_literal: true
+
+# -- copyright
+# OpenProject is an open source project management software.
+# Copyright (C) the OpenProject GmbH
+#
+# This program is free software; you can redistribute it and/or
+# modify it under the terms of the GNU General Public License version 3.
+#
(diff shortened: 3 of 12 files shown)


app/models/project.rb
```diff
@@ -88,7 +88,7 @@ class Project < ApplicationRecord
   has_many :storages, through: :project_storages
   has_many :phases, class_name: "Project::Phase", dependent: :destroy
   has_many :available_phases,
-           -> { visible.eager_load(:definition).order(position: :asc) },
```
dombesz, line 91: Available phases are used in other places like the [`Overviews::ProjectPhases::ShowComponent`](https://github.com/opf/openproject/blob/5f9ef3f44fa7103bac37dc93043c82e38e75b7e4/modules/overviews/app/components/overviews/project_phases/show_component.rb#L42).
I would prefer having the eager loading on the association.

config/initializers/permissions.rb
```diff
@@ -158,6 +158,7 @@
                      },
                      permissible_on: :project,
                      require: :member,
+                     dependencies: :edit_project_phases,
```
dombesz, line 161: Nice catch! 👍

spec/services/project_life_cycle_steps/activation_service_spec.rb
```diff
@@ -0,0 +1,450 @@
...
+
+      context "with date range" do
+        let!(:phase1) { create_phase(definition: definitions[1], active: true, date_range: date + 2..date + 3) }
+
+        it "reschedules following phases using dates of that phase" do
+          service.call(active: false)
+
+          expect(phase0.reload).to have_attributes(active: true, start_date: date - 1, finish_date: date - 1)
+          expect(phase1.reload).to have_attributes(active: false, start_date: date + 2, finish_date: date + 3)
+          expect(phase2.reload).to have_attributes(active: true, start_date: date + 2, finish_date: date + 4)
```
dombesz, line 407: I'm wondering if the expected behaviour is to reschedule the last phase to the deactivated previous phase, or rather reschedule it to the previous active phase? In this case should it be `start_date: date` instead ?

dombesz (commented):
Nice job overall @toy, I just have one question to clarify.

app/models/project.rb
dombesz: Added back in ca7e4bb266e1

dombesz (approved):
Since my [question](https://github.com/opf/openproject/pull/18824#discussion_r2087456830) is not impeding the PR being merge, I would move forward with it.

spec/services/project_life_cycle_steps/activation_service_spec.rb
toy: I did it following the spec and normally phases will already be properly following each other, so there should be no difference between using deactivated phase start date or following previous active phase. If you think this is important, we can adapt logic

PR #18847, opened 2025-05-07
[#62318] submit inline create work packages conditionally

Kharonus:
# Ticket
[OP#62318](https://community.openproject.org/wp/62318)

- when inline creating a wp of a type with generated subjects, the creation must be instantaneous, if no field is missing

- inline create component now submits the form after creating the row, if the form has no validation errors - i.e. no fields are active anymore.


NobodysNightmare (approved):
Elegant solution to the problem.

PR #18848, opened 2025-05-07
Implementation/63015 pdf xls cvs exports of linked phases for work packages

ulferts:
# Ticket

https://community.openproject.org/wp/63015

To get the full functionality, including exporting to xls, csv and list pdfs, #18851 will need to be merged first. But checking on a branch rebased on it looked solid.

Enable exporting the project_phase column for work packages for the formats:
* [x] pdf
* [x] xls
* [x] csv

# Merge checklist

- [x] Added/updated tests

### DIFF
--- a/app/models/work_package.rb
+++ b/app/models/work_package.rb
@@ -338,6 +338,15 @@ class WorkPackage < ApplicationRecord
     duration ? duration * 24 : nil
   end
 
+  def project_phase
+    # This might look less efficient than using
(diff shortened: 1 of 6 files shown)


as-op (approved):
Looks good to me. 
I investigated where the unimplemented function `format_value` might be called.  It appears it's only used for numeric sums and custom fields, so raising an exception there should be safe.

PR #18851, opened 2025-05-08
Implementation/62982 project phase column on work package list

EinLama:
# Ticket
https://community.openproject.org/wp/62982

Make the new project phase attribute on work packages viewable, sortable and groupable within the work package table.

Since the property has a different name than the association, I had to do some group name conversions. Additionally the project phase icon is also shown within group headers. Admittedly, my solution for this is a bit hacky. But as long as we only have one WP property that makes use of this, I tried to keep it rather simple (and stupid).

AC from the ticket (all done in this PR):

* The attribute can be added as a column to the work package list.
* The table can be ordered and grouped by that column.
* The order is not by name of the phase but by the order the phase are in.
* The group header has the colored icon prefixed (if possible).
* Within the table column for project phases, the selected phase is displayed and can be edited same as in the split & full screen view: The selected value has a colored icon prefixed same as the possible values in the autocompleter.

## Screenshots

https://github.com/user-attachments/assets/da1bfcd8-ebe6-4605-85e3-28ef87330c27

# Merge checklist

- [x] Added/updated tests


spec/features/work_packages/table/project_phase_field_spec.rb
```diff
@@ -0,0 +1,172 @@
...
+            expect(page).to have_test_selector("project-phase-icon #{project_phase.name}")
+          end
+        end
+      end
+    end
+
+    context "with one phase being inactive" do
+      let(:other_project_phase) { create(:project_phase, active: false) }
+
+      it "does not show the inactive phase" do
```
dombesz, line 153: The grouping of an inactive phase still shows up, I'm not sure what would be the right behaviour.
In the following picture, the "Planning" phase is made inactive:
image

dombesz (approved):
Nice work Tobi 👏 I left one note only.

spec/features/work_packages/table/project_phase_field_spec.rb
EinLama: Thanks for bringing this up, very good point! We discussed this in the weekly and we decided that a group with no active phases for the listed work packages should be hidden/shown as "-" in the group header.

I will adjust this PR accordingly :)

EinLama: It was a bit more complicated than I anticipated, but this feedback has been addressed now. I also added a test case for it.

dombesz: Unfortunately in the dev environment I still don't get the correct results, however the specs pass locally.

image

EinLama: I was not able to reproduce _this_ exact issue, but _another_ one 🤪 Thank you for reporting it. Jens and I dug deeper and found and fixed the cause 💪

dombesz: Thanks Tobi, I checked now and it displays correctly. However the sorting still takes the omitted definition's position into account. In the screenshot below, "Fifth dev" should be at the bottom of the page, as the second page continues with non phase associated work packages.
image
image

EinLama: I could not reproduce this exact issue 🤔 In your example, is "fifth dev" associated to an deactivated phase?

With a similar setup, I get correctly sorted groups.
image

While investigating, I found another issue with the sorting that is fixed as of the latest commit. I doubt that it has any effect on the erroneous behavior from your screenshot, but if you have time I'd appreciate it if you took a look. 🙇🏻‍♂️

dombesz: I re-seeded my data, and I cannot reproduce the issue anymore, it is possible that I had some inconsistent data.

frontend/src/app/features/hal/resources/wp-collection-resource.ts
```diff
@@ -60,6 +60,7 @@ export interface WorkPackageCollectionResource extends WorkPackageCollectionReso
  */
 export interface GroupObject {
   value:any;
+  leadingIcon?:string;
```
ulferts: I am wondering about the necessity of this property. Our UI needs it to know that an icon is to be displayed. But the fact that an icon is displayed is a decision taken on the UI. Other UIs might not want to have an icon displayed. Of course, they can then just ignore the property. What I am getting at though is that this is not an intrinsic part of the GroupObject. 

For our use case, we also don't differentiate between the phases to display different icons. We do differentiate whether to display an phase icon or not but all the phase groups have the same icon. So we could also rely on the information present in `_links/groupBy/href` to know that we group by `/api/v3/queries/group_bys/projectPhaseDefinition` in this case.

docs/api/apiv3/tags/work_packages.yml
```diff
@@ -20,30 +20,31 @@ description: |-
...
+  | type                   | The type of the work package                                                                                                                          | Type                   | not null                                  | READ / WRITE          |                                                                                             |
+  | version                | The version associated to the work package                                                                                                            | Version                |                                           | READ / WRITE          |                                                                                             |
+  | watchers               | All users that are currently watching this work package                                                                                               | Collection             |                                           | READ                  | **Permission** view work package watchers                                                   |
```
ulferts, line 47: Adding the projectPhaseDefinition here is only part of what need to be done in order to document the added link. It will also be needed as a description in the [model](https://github.com/opf/openproject/blob/dev/docs/api/apiv3/components/schemas/work_package_model.yml#L429-L435)

spec/features/work_packages/table/project_phase_field_spec.rb
```diff
@@ -0,0 +1,191 @@
...
+        wp_table.expect_work_package_with_attributes(work_package, { projectPhase: project_phase.name })
+        wp_table.expect_work_package_with_attributes(other_wp, { projectPhase: "-" })
+        wp_table.expect_work_package_with_attributes(wp_without_phase, { projectPhase: "-" })
+      end
+
+      context "when sorting by project phase ASC" do
+        let(:sort_criteria) { [%w[project_phase asc]] }
+
+        it "sorts work packages with an inactive project phase like work packages without a project phase" do
+          wp_table.expect_work_package_order(wp_without_phase, other_wp, work_package)
```
ulferts: The expected order between this test case and the one where the phase is active is no different. Since based on my tests, the active state is currently not correctly handled, I would recommend disabling the project phase of `work_package` as that would lead to a different expectation.

spec/features/work_packages/table/project_phase_field_spec.rb
```diff
@@ -0,0 +1,191 @@
...
+      let!(:query) { build(:global_query, user: current_user) }
+
+      it "does not render project phases you don't have permission for" do
+        # permission given, phase visible:
+        wp_table.expect_work_package_with_attributes(work_package, { projectPhase: project_phase_definition.name })
+
+        # permission missing, phase invisible:
+        wp_table.expect_work_package_with_attributes(wp_from_another_project, { projectPhase: "" })
+      end
+    end
```
ulferts, line 255: The cases for sorting and grouping are not speced here.

app/models/queries/work_packages/selects/project_phase_select.rb
```diff
@@ -0,0 +1,106 @@
...
+    <<~SQL.squish
+      LEFT OUTER JOIN (
+        SELECT d.id, d.position
+        FROM project_phase_definitions d
+        JOIN project_phases ph ON ph.definition_id = d.id
+        JOIN projects p ON ph.project_id = p.id
+        WHERE ph.active = true
+        GROUP BY d.id
+      ) pd
+      ON project_phase_definition_id = pd.id
```
ulferts: There is a flaw in here that will lead to the following results in the UI. It only manifests in case there are multiple projects in the installation where at least one has the phase active. Then, even if the phase is disabled in the project for which a query is run, work packages having a disabled phase associated, will get sorted and grouped as if the phase would still be active:

image

image

The reason for this is that the ON clause lacks a condition to select the phase active in the project. That can only be done, though if the subquery projects the project_id as well. The resulting query will have to look something like this:

```
SELECT
			D.ID,
			D.POSITION,
			PH.project_id
		FROM
			PROJECT_PHASE_DEFINITIONS D
			JOIN PROJECT_PHASES PH ON PH.DEFINITION_ID = D.ID
			JOIN PROJECTS P ON PH.PROJECT_ID = P.ID
		WHERE
			PH.ACTIVE = TRUE
		GROUP BY
			D.ID,
			PH.project_id
	) PD ON PROJECT_PHASE_DEFINITION_ID = PD.ID AND PD.project_id = work_packages.project_id

```

This could be simplified to just joining
```
LEFT OUTER JOIN "project_phases" ON "project_phases".project_id = work_packages.project_id AND project_phases.definition_id = work_packages.project_phase_definition_id
```

and changing the sortable_statement to:
```
CASE WHEN project_phases.active THEN project_phase_definitions.position ELSE -1
```

However, I am not sure if this simplification can be kept up since the implementation does not take heed of the permissions. While it is checked that the column is not available if the permission is lacking, the permission is not enforced on each record. So when two projects are filtered on (parent and child) and the user has the permission to  see project phases in the parent project, they can sort and group by project phases. That is true even if the user lacks the permission to see the project phases in the child project. While the phase column on work packages from the child project are empty, they are still sorted/grouped as if they would be visible. In the example below, the work packages called "ABC in Large childer" is in the child work package and the user lacks the permission to see project phases in there. 

When sorting:
image

When grouping 
image

Therefore I would probably keep the subselect and add a check for whether the project_id is in 
```
Project.allowed_to(User.current, :view_project_phases).select(:id).to_sql
```
or something similar

ulferts (changes requested):
Almost there, @EinLama . Most of the implementation looks solid and it took me a while to find a bug. But unfortunately I did, sorry. It is related to having multiple projects and permission checks. So not that easy to stumble over.

docs/api/apiv3/tags/work_packages.yml
EinLama: Thank you, added. I also added an entry to the [schema model](https://github.com/opf/openproject/blob/dev/docs/api/apiv3/components/schemas/work_package_schema_model.yml) since I think it belongs there as well.

frontend/src/app/features/hal/resources/wp-collection-resource.ts
EinLama: I wondered about this, too. Since we are already two now, I have moved the logic to the UI as proposed. 👍

spec/features/work_packages/table/project_phase_field_spec.rb
EinLama: Good point, specs have been added!

app/models/queries/work_packages/selects/project_phase_select.rb
EinLama: Thank you, Jens! I missed this case completely. I had to tinker with your proposed SQL a bit since the count for the different groups did not treat inactive phases correctly in the above cases.

I have now added the updated code (also taking permissions on a per-project basis into consideration) along with test cases for these situations - so we don't break this accidentally in the future.

frontend/src/app/features/hal/resources/wp-collection-resource.ts
EinLama: Note: I added a very simple implementation for now. At first, I wanted to add a map between an href-identifier and an iconFetcher-callback. But it turned out ESLint was very unhappy with my attempt. After refactoring everything to comply with the rules, the generic approach turned into quite a complex endeavor. I did not deem the added complexity worthy for our current use case, where we show an icon in exactly one case - so I reverted it back to the KISS version for now.

ulferts (approved):
All solid and good, @EinLama . 

Couldn't find any cases where the grouping & sorting misbehaves and there are quite a few.

While taking a look at the produced SQL I created #19095 which reduces the joining needed. But the functionality was already good before that PR.

PR #18852, opened 2025-05-08
have project phase changes show up on wp activity

ulferts:
# Ticket

https://community.openproject.org/wp/62983

This PR introduces functionality to journalize changes to project phase definitions associated with work packages. It uses the default state changes of journals (e.g. 'Project phase set to abc'). The difference is just in the way the values are displayed. Since project phases can both be inactive in the project as well not being configured in it at all, the formatter checks for the active state of the phase in the project. If it is inactive or not configured, "(Inactive)" is appended to the name of the definition.

## Screenshots

image

# Merge checklist

- [x] Added/updated tests

### DIFF
--- a/lib/open_project/journal_formatter/project_phase_definition.rb
+++ b/lib/open_project/journal_formatter/project_phase_definition.rb
@@ -32,7 +32,9 @@ class OpenProject::JournalFormatter::ProjectPhaseDefinition < JournalFormatter::
   private
 
   def associated_object_name(object)
-    if phase_active?(object)
+    if object.nil?
+      I18n.t(:"activity.project_phase.deleted_project_phase")
+    elsif phase_active?(object)
       super
     else
       "#{super} (#{I18n.t(:label_inactive)})"
@@ -40,8 +42,11 @@ class OpenProject::JournalFormatter::ProjectPhaseDefinition < JournalFormatter::
   end
 
   def phase_active?(definition)
-    phase = @journal.journable.project.phases.detect { |phase| phase.definition_id == definition.id } if definition
-
-    phase&.active
+    @journal
+      .journable
+      .project
+      .phases
+      .detect { |phase| phase.definition_id == definition.id }
+      &.active
   end
 end
--- a/lib/open_project/journal_formatter/project_phase_definition.rb
+++ b/lib/open_project/journal_formatter/project_phase_definition.rb
@@ -1,0 +1,47 @@
+# frozen_string_literal: true
+
+# -- copyright
+# OpenProject is an open source project management software.
+# Copyright (C) the OpenProject GmbH
+#
+# This program is free software; you can redistribute it and/or
+# modify it under the terms of the GNU General Public License version 3.
+#
+# OpenProject is a fork of ChiliProject, which is a fork of Redmine. The copyright follows:
+# Copyright (C) 2006-2013 Jean-Philippe Lang
+# Copyright (C) 2010-2013 the ChiliProject Team
+#
+# This program is free software; you can redistribute it and/or
+# modify it under the terms of the GNU General Public License
+# as published by the Free Software Foundation; either version 2
+# of the License, or (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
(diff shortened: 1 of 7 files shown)


lib_static/plugins/acts_as_journalized/lib/journal_formatter/named_association.rb
```diff
@@ -53,19 +53,17 @@ def format_values(values, key, cache:)
       klass = class_from_field(key)
 
       values.map do |value|
-        if klass && value
-          record = associated_object(klass, value.to_i, cache:)
-          if record
-            if record.respond_to? :name
-              record.name
-            else
-              record.subject
```
ulferts, line 62: This extra path should no longer be necessary as [work packages have a name method](https://github.com/opf/openproject/blob/dev/app/models/work_package.rb#L343-L345) implemented by now

toy:
Not a problem of this PR, but it feels odd that if a work package is assigned to inactive phase, then from looking at activity it is assigned, but when looking at the work package, it is not

lib/open_project/journal_formatter/project_phase_definition.rb
```diff
@@ -0,0 +1,47 @@
...
+  def associated_object_name(object)
+    if phase_active?(object)
+      super
+    else
+      "#{super} (#{I18n.t(:label_inactive)})"
+    end
+  end
+
+  def phase_active?(definition)
+    phase = @journal.journable.project.phases.detect { |phase| phase.definition_id == definition.id } if definition
```
toy: Why load all phases and not find needed phase by `definition_id`?

spec/lib/open_project/journal_formatter/project_phase_definition_spec.rb
toy, line 1: Maybe better to add a test that checks changing from active to inactive and/or vice versa

ulferts: I would trust the unit specs to test that. The feature specs ensure that the journal formatter is hooked up and the details can then be enforced by the unit spec.

lib/open_project/journal_formatter/project_phase_definition.rb
ulferts: It is a bet. While fetching that one specific phase via a find_by directly would definitely be faster for a single journal, I wager that the performance will increase if the phase is changed more than once. Since the project is shared across journals, so are the phases. By that, journals can access them again without taking a performance hit.

spec/lib/open_project/journal_formatter/project_phase_definition_spec.rb
toy: I am suggesting to add a unit test, there are already 9 cases covered, but none for phases being in different active state.

lib/open_project/journal_formatter/project_phase_definition.rb
toy: Ok, let it be

lib/open_project/journal_formatter/project_phase_definition.rb
```diff
@@ -0,0 +1,47 @@
...
+
+    phase&.active
```
toy: Maybe then:
```suggestion
    return unless definition
    
    @journal.journable.project.phases
      .find { it.definition_id == definition.id }
      &.active?
```

or

```suggestion
    return unless definition
    
    @journal.journable.project.phases
      .find { |phase| phase.definition_id == definition.id }
      &.active?
```

toy:
If definition gets deleted, it will be displayed as just `(Inactive)`, probably better if it says that it was a deleted phase

ulferts:
> If definition gets deleted, it will be displayed as just `(Inactive)`, probably better if it says that it was a deleted phase

Well spotted. I added a special return value for this case in the formatter. That change also made the handling of a nil object in `#phase_active?` superfluous.

spec/lib/open_project/journal_formatter/project_phase_definition_spec.rb
```diff
@@ -0,0 +1,264 @@
...
+               label: "<strong>Project phase</strong>",
+               old: "<strike><i>#{project_phase.name} (Inactive)</i></strike>")
+      end
+
+      it_behaves_like "renders project phase definition change"
+    end
+
+    context "when setting a phase whose definition is deleted" do
+      let(:old_value) { nil }
+      let(:new_value) { -1.to_s }
```
toy: I'd expect rubocop to annoy about this:
```suggestion
      let(:new_value) { "-1" }
```

toy (approved):

PR #18856, opened 2025-05-08
bug/63897 Missing target element "internalCheckbox" error when internal comments feature is disabled

akabiru:
https://community.openproject.org/wp/63897

When the Internal comments feature is disabled, the internalCommentCheckboxTarget does not exist, causing a warning from Stimulus; the InternalCommentController listens to form submit end events from the activities IndexController which manages comment form submission- it is then able to update the UI state on form close. E.g. Ensuring that the background color is reset https://github.com/opf/openproject/commit/e4270942f935c127bf322955933c7bb81b93192b

Screenshot 2025-05-08 at 4 48 47 PM


brunopagno (approved):
Nice catch!

:ship: :ship:

PR #18872, opened 2025-05-09
Correctly handle users unable to authenticate at a storage

NobodysNightmare:
The previous code assumed that if the authentication method selector doesn't choose SSO, it must have chosen OAuth 2.0, which results in rendering an ensure connection link. However, it's also possible that the method selector chooses NO authentication method, because the user simply can't authenticate at the storage. In this case rendering an ensure connection link would fail.

This was not detected earlier due to missing tests.

# Ticket
https://community.openproject.org/wp/63909


ba1ash (approved):
LGTM

NobodysNightmare:
Merging now even though flickering specs are present... Four test executions and the failing tests are always different ones and never related to the code touched in this PR.

PR #18883, opened 2025-05-09
Fix parsing of multiple certificates from metadata

NobodysNightmare:
Once the XML contains multiple certificates, the metadata hash has a nil idp_cert. The idp_cert_multi key is always populated. However, we never tried parsing data from there, so it was lost.

The way that the Saml::Provider right now represents multiple certificates is to concatenate them inside the idp_cert, so the metadata parsing has been adapted to use that.

# Ticket
https://community.openproject.org/wp/63610

### DIFF
--- a/modules/auth_saml/app/services/saml/update_metadata_service.rb
+++ b/modules/auth_saml/app/services/saml/update_metadata_service.rb
@@ -1,3 +1,5 @@
+# frozen_string_literal: true
+
 #-- copyright
 # OpenProject is an open source project management software.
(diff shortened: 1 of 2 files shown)


ba1ash (commented):
What do you think about extending the spec to include an encryption certificate as well?
I'd vote for adding it if it is not much trouble.

NobodysNightmare:
Specs have been extended.

ba1ash (approved):
:heart:

PR #18884, opened 2025-05-09
Feature/59180 rescheduling on changing week days and holidays 

toy:
# Ticket

[OP#63901](https://community.openproject.org/wp/63901) (part of [OP#59180](https://community.openproject.org/wp/59180), follows #18523 and #18824)

Automatic rescheduling of phases when working days and holidays are change in admin interface

## Screenshots

# Merge checklist

- [ ] Added/updated tests

### DIFF
--- a/app/models/project/phases/scopes/covering_dates_and_days_of_week.rb
+++ b/app/models/project/phases/scopes/covering_dates_and_days_of_week.rb
@@ -1,0 +1,88 @@
+# frozen_string_literal: true
+
+#-- copyright
+# OpenProject is an open source project management software.
+# Copyright (C) the OpenProject GmbH
+#
+# This program is free software; you can redistribute it and/or
+# modify it under the terms of the GNU General Public License version 3.
+#
+# OpenProject is a fork of ChiliProject, which is a fork of Redmine. The copyright follows:
+# Copyright (C) 2006-2013 Jean-Philippe Lang
+# Copyright (C) 2010-2013 the ChiliProject Team
+#
+# This program is free software; you can redistribute it and/or
+# modify it under the terms of the GNU General Public License
+# as published by the Free Software Foundation; either version 2
+# of the License, or (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program; if not, write to the Free Software
+# Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+#
+# See COPYRIGHT and LICENSE files for more details.
+#++
+
+module Project::Phases::Scopes::CoveringDatesAndDaysOfWeek
+  extend ActiveSupport::Concern
+  using CoreExtensions::SquishSql
+
+  class_methods do
+    # Fetches all phases that cover specific days of the week, and/or specific dates.
+    #
+    # The period considered is from the phase start date to the due date.
+    #
+    # @param dates Date[] An array of the Date objects.
+    # @param days_of_week number[] An array of the ISO days of the week to
+    #   consider. 1 is Monday, 7 is Sunday.
+    def covering_dates_and_days_of_week(days_of_week: [], dates: [])
+      days_of_week = Array(days_of_week)
+      dates = Array(dates)
+      return none if days_of_week.empty? && dates.empty?
+
+      ids_sql = sanitize_sql([<<~SQL.squish, { days_of_week:, dates: }])
+        WITH
+          -- select phases with at least one date
+          phases_with_dates AS (
+            SELECT
+              id,
+              COALESCE(start_date, finish_date) AS start_date,
+              COALESCE(finish_date, start_date) AS finish_date
+            FROM project_phases
+            WHERE
+              start_date IS NOT NULL
+              OR finish_date IS NOT NULL
+          ),
+
+          -- generate a row for every date between start_date and finish_date
+          covered_dates AS (
+            SELECT
+              id,
+              generate_series(start_date, finish_date, '1 day') AS date
+            FROM phases_with_dates
+          ),
+
+          -- add day of the week column
+          covered_dates_and_wday AS (
+            SELECT
+              id,
+              date,
+              EXTRACT(isodow FROM date) AS dow
+            FROM covered_dates
+          )
+
+        -- select id of phases covering the given days
+        SELECT id
+        FROM covered_dates_and_wday
+        WHERE dow IN (:days_of_week) OR date IN (:dates)
+      SQL
+
+      where("id IN (#{ids_sql})")
+    end
+  end
+end
--- a/app/models/project/phases/scopes/covering_dates_and_days_of_week.rb
+++ b/app/models/project/phases/scopes/covering_dates_and_days_of_week.rb
@@ -70,16 +70,14 @@ module Project::Phases::Scopes::CoveringDatesAndDaysOfWeek
           -- eqal, then the range covers the target day of week
           EXISTS (
             SELECT 1
-            FROM unnest(Array[:days_of_week]::INT4[]) AS target_dow
-            WHERE (
-              MOD((target_dow + 7 - EXTRACT(isodow FROM start_date)) % 7, 7) <= (finish_date - start_date)
-            )
+            FROM UNNEST(ARRAY[:days_of_week]::INT[]) AS target_dow
+            WHERE (target_dow + 7 - EXTRACT(ISODOW FROM start_date)::INT) % 7 <= finish_date - start_date
           )
           OR
           -- Check if the range covers any of the provided dates
           EXISTS (
             SELECT 1
-            FROM unnest(Array[:dates]::DATE[]) AS target_date
+            FROM UNNEST(ARRAY[:dates]::DATE[]) AS target_date
             WHERE target_date BETWEEN start_date AND finish_date
           )
       SQL
--- a/app/models/project/phases/scopes/covering_dates_and_days_of_week.rb
+++ b/app/models/project/phases/scopes/covering_dates_and_days_of_week.rb
@@ -57,29 +57,31 @@ module Project::Phases::Scopes::CoveringDatesAndDaysOfWeek
             WHERE
               start_date IS NOT NULL
               OR finish_date IS NOT NULL
-          ),
-
-          -- generate a row for every date between start_date and finish_date
-          covered_dates AS (
-            SELECT
-              id,
-              generate_series(start_date, finish_date, '1 day') AS date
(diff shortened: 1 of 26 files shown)


app/models/project/phases/scopes/covering_dates_and_days_of_week.rb
```diff
@@ -0,0 +1,88 @@
...
+
+  class_methods do
+    # Fetches all phases that cover specific days of the week, and/or specific dates.
+    #
+    # The period considered is from the phase start date to the due date.
+    #
+    # @param dates Date[] An array of the Date objects.
+    # @param days_of_week number[] An array of the ISO days of the week to
+    #   consider. 1 is Monday, 7 is Sunday.
+    def covering_dates_and_days_of_week(days_of_week: [], dates: [])
```
ulferts: This method seems to return all the phases which:
* span a certain weekday OR
* span a certain date

So I think the method should rather be named `covering_dates_or_days_of_week`

app/models/work_packages/scopes/covering_dates_and_days_of_week.rb
```diff
@@ -59,30 +59,25 @@ def where_covers_periods(work_packages_periods_cte, days_of_week, dates)
...
+            SELECT 1
+            FROM UNNEST(ARRAY[:days_of_week]::INT[]) AS target_dow
+            WHERE (target_dow + 7 - EXTRACT(ISODOW FROM start_date)::INT) % 7 <= (end_date - start_date)
+          )
+          OR
+          -- Check if the range covers any of the provided dates
+          EXISTS (
+            SELECT 1
+            FROM unnest(Array[:dates]::DATE[]) AS target_date
+            WHERE target_date BETWEEN start_date AND end_date
```
ulferts, line 82: 👍 for cleaning this up in lockstep with the sibling scope.

app/workers/apply_working_days_change_job_base.rb
```diff
@@ -0,0 +1,82 @@
...
+
+    User.execute_as user do
+      apply_working_days_change
+    end
+  end
+
+  private
+
+  def apply_working_days_change
+    fail NoMethodError, "Must be overridden in subclass"
```
ulferts, line 56: Not overly important but why is a NoMethodError used here instead of a NotImplementedError?

app/workers/projects/phases/apply_working_days_change_job.rb
```diff
@@ -0,0 +1,62 @@
...
+# Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+#
+# See COPYRIGHT and LICENSE files for more details.
+#++
+
+class Projects::Phases::ApplyWorkingDaysChangeJob < ApplyWorkingDaysChangeJobBase
+  private
+
+  def apply_working_days_change
+    user = User.current
```
ulferts: I think this is here uselessly.

app/workers/projects/phases/apply_working_days_change_job.rb
```diff
@@ -0,0 +1,62 @@
...
+    end
+  end
+
+  def applicable_phases
+    days_of_week = changed_days.keys
+    dates = changed_non_working_dates.keys
+
+    Project::Phase
+      .active # TODO: should visible be used?
+      .covering_dates_and_days_of_week(days_of_week:, dates:)
```
ulferts: Since a user calling this job is an admin, using the `visible` scope would only change the scope to only include active projects. Archived projects would be included as permissions are not active in there, even for admins. 

That might be a worthwhile approach, though, since it would then not change the scheduling of already completed projects. However, if we follow along this path, we would have to reschedule the phases when reactivating a project. That is a use case, that while being a valid one, I currently judge to be of lesser importance so I would currently not support it. I also wouldn't know where to make the cut. Because one could also argue that phases that are scheduled in the past should not be rescheduled. To me however, that argument only is worth to follow up on when we support validity periods for the non working days settings.

TL;DR: Using the `active` scope seems fine to me.

app/workers/projects/phases/apply_working_days_change_job.rb
```diff
@@ -0,0 +1,62 @@
...
+# See COPYRIGHT and LICENSE files for more details.
+#++
+
+class Projects::Phases::ApplyWorkingDaysChangeJob < ApplyWorkingDaysChangeJobBase
+  private
+
+  def apply_working_days_change
+    user = User.current
+
+    applicable_phases.distinct.pluck(:project_id).each do |project_id|
```
ulferts: The `distinct` IMO is unnecessary since the scope should not return duplicates. On the other hand, adding a `.uniq` after the `pluck(:project_id)` would save unnecessary iterations. 

Here, however, it might be better to use something like 

```
Project.where(id: applicable_phases.select(:project_id)).each do |project|
```

That way, the check for the existence of the project can be be omitted as well.

app/workers/projects/phases/apply_working_days_change_job.rb
```diff
@@ -0,0 +1,62 @@
...
+      project = Project.find_by(id: project_id)
+      next unless project
+
+      phases = project.available_phases
+      next if phases.empty?
+
+      from = phases.first.start_date
```
ulferts: The first phase might have its start date not set but other phases in the same project can still have dates. Because of that, I think it safer to use 

```
from = phases.pluck(:start_date, :finish_date).flatten.compact.min
```

app/services/project_life_cycle_steps/reschedule_service.rb
```diff
@@ -0,0 +1,81 @@
...
+    def persist(service_call)
+      reschedule_phases(**params)
+
+      service_call
+    end
+
+    def reschedule_phases(phases:, from:)
+      phases.each do |phase|
+        next unless phase.date_range_set?
+        next unless phase.duration&.positive? # not updated using service that sets duration
```
ulferts, line 52: Currently the caller is responsible to filter out phases that should not be rescheduled. On the other hand, some selection is taking place in here. Why is `active` not filtered for here?

ulferts (changes requested):
The implementation looks solid and well structured. 

I did not manage to run through a complete set of test cases yet but found one edge case that is worth to take a look at.

### Having project phases with a phase in the middle not having a date set

image

When adding a non working day within the time frame spanned by the other phases, the rescheduling will break.

app/workers/apply_working_days_change_job_base.rb
toy: It is incorrect to use `NotImplementedError` in such case, at it is «Raised when a feature is not implemented on the current platform». Defining method and raising same `NoMethodError` error as when it would not be defined is for documenting and giving more info in the raised exception. So it would be better to use that or introduce an `AbstractMethodError` and use it in all other places, but given that this looks like the only place where `NoMethodError` is used instead of `NotImplementedError`, it can be changed later across the codebase if other developers agree

app/workers/projects/phases/apply_working_days_change_job.rb
toy: It is needed for `RescheduleService`

app/workers/projects/phases/apply_working_days_change_job.rb
toy: `applicable_phases` returns a scope for phases, so can return multiple phases for same project and `distinct` applies to the select from `pluck`. Compare SQLs for simpler queries without and with distinct:
```ruby
Project::Phase.pluck(:project_id)
# SELECT "project_phases"."project_id" FROM "project_phases"
```
```ruby
Project::Phase.distinct.pluck(:project_id)
# SELECT DISTINCT "project_phases"."project_id" FROM "project_phases"
```

I went with iterating ids of projects and loading projects later to not load all projects in memory and because slight decrease of performance loading projects one by one should not be very important here. Alternative is probably to use `Project.where(id: applicable_phases.select(:project_id)).find_each`

app/workers/projects/phases/apply_working_days_change_job.rb
toy: Good point. First I though that `covering_dates_and_days_of_week` should not allow that, but it returns also phases with only one date set if it is on the affected day of week or date. Still it is better to force querying `phases`, as they are needed afterwards anyway

app/services/project_life_cycle_steps/reschedule_service.rb
toy: Good point, while now I rely on input to provide only active phases, it is better to add the check to be less confusing in the future. Also it is a reminder that phases are also not checked for order, so if a scope is used that would return them not in definition order, they will be rescheduled in incorrect order, but it felt odd to load definitions here only to ensure order.

app/workers/projects/phases/apply_working_days_change_job.rb
toy: I'm also not sure if it makes sense to consider `finish_date` if `start_date` is not set

toy:
@ulferts Thank you for thorough review, I either applied changes or replied, so please have a look

app/workers/projects/phases/apply_working_days_change_job.rb
ulferts: True, I didn't see the connection. I guess it was moved outside of the loop for performance reasons. However, I think the overhead of calling User.current on a call to 

```
ProjectLifeCycleSteps::RescheduleService.new(user: User.current, project:)
```

would be minimal and would increase readability.

app/workers/apply_working_days_change_job_base.rb
ulferts: Thanks for the clarification. I wasn't aware of that definition so I misused the `NotImplementedError` in the past it seems.

app/workers/projects/phases/apply_working_days_change_job.rb
ulferts:
```
Project.where(id: applicable_phases.select(:project_id)).find_each
```

Is what I would work with, then. To me, the extra loading and checking of the project produces noise.

app/workers/projects/phases/apply_working_days_change_job.rb
ulferts: When not considering the finish_date, something like this can happen (edge case admittedly):

Have a set of phases where one only has its finish date set. 

image

Add a non working days to the finish date set for the phase. 

Now add a start date to the phase which only had a finish date:

image

I am not sure, if this should be prevented by the automatic rescheduling, though. It could set the date to be on the next non working day. But since it requires the user to interact with the phase again to raise the error it might be acceptable to have the invalid date kept. What do you think, @toy ?

ulferts (approved):
Good work @toy. From my point of view, this is good to merge. 

There is this one edge case still open which I am undecided about but this can either decided to be ignored or be fixed later on. 

So once the CI is green, go ahead and merge 🎉

app/workers/projects/phases/apply_working_days_change_job.rb
toy: I would keep it as is for now, I think updating finish date when phase has no start date may be more unexpected than allowing it to become invalid

PR #18885, opened 2025-05-09
[60580] Fix error 500 when bulk editing work packages

cbliard:
# Ticket

https://community.openproject.org/wp/60580

Fix 500 error happening randomly when bulk editing some work packages having long text custom fields.

Note: on development environment, it produces a 409, not a 500. Visible from Network tab of browser DevTools.

## Screenshots

Not applicable.

When multiple CKEditor instances are on the same form, each of them intercepts the submit event and submit the form, leading to multiple identical parallel submit requests.

This happens for instance on the bulk edit page when the work packages have long text custom fields because this adds a ckeditor in the form. The form is submitted multiple times and only one request will succeed, the other will produce a 500. That's why a 500 error is sometimes visible on bulk edit.

The fix is to track the "in-flight" state per form in a variable shared with all instances of the component instead of storing it in an instance variable.

# Merge checklist

- [ ] Added/updated tests
  - I'm not sure how to test that. I'll happily add tests if someone has a good idea of reliable unit tests for that.

### DIFF
--- a/frontend/src/app/shared/components/editor/components/ckeditor-augmented-textarea/ckeditor-augmented-textarea.component.ts
+++ b/frontend/src/app/shared/components/editor/components/ckeditor-augmented-textarea/ckeditor-augmented-textarea.component.ts
@@ -60,12 +60,15 @@ import { populateInputsFromDataset } from 'core-app/shared/components/dataset-in
 import { navigator } from '@hotwired/turbo';
 import { uniqueId } from 'lodash';
 
-
 @Component({
   templateUrl: './ckeditor-augmented-textarea.html',
   changeDetection: ChangeDetectionStrategy.OnPush,
 })
 export class CkeditorAugmentedTextareaComponent extends UntilDestroyedMixin implements OnInit {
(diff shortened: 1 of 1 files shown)


myabc (approved):
I was able to reproduce the bug locally (albeit with `409 Conflict` response), as well as verify the fix.

LGTM 👍🏻

myabc:
> I'm not sure how to test that. I'll happily add tests if someone has a good idea of reliable unit tests for that.

I'm not sure. A system spec testing bulk edit with multiple rich text fields might make sense in this scenario.

cbliard:
> > I'm not sure how to test that. I'll happily add tests if someone has a good idea of reliable unit tests for that.
> 
> I'm not sure. A system spec testing bulk edit with multiple rich text fields might make sense in this scenario.

Yes, but that would not be a reliable unit test. It could pass while the bug is still there. Unless it could count how many times the controller was called.

Well, not everything has to have a test I suppose. It's a balance between maintainability and confidence.

PR #18887, opened 2025-05-09
[chore] fix pattern attribute matcher

Kharonus:
- searching for words in custom field names must work

- label must be cast to lowercase for search matches

### DIFF
--- a/frontend/src/stimulus/controllers/pattern-input.controller.ts
+++ b/frontend/src/stimulus/controllers/pattern-input.controller.ts
@@ -444,7 +444,7 @@ export default class PatternInputController extends Controller {
         key,
         label: group.title,
         values: group.tokens
-          .filter((token) => token.key.includes(word) || token.label.includes(word) || word === '*')
+          .filter((token) => token.key.includes(word) || token.label.toLowerCase().includes(word) || word === '*')
           .map((token) => ({ prop: token.key, value: token.label })),
       };
     }).filter((group) => group.values.length > 0);


frontend/src/stimulus/controllers/pattern-input.controller.ts
```diff
@@ -444,7 +444,7 @@ export default class PatternInputController extends Controller {
         key,
         label: group.title,
         values: group.tokens
-          .filter((token) => token.key.includes(word) || token.label.includes(word) || word === '*')
+          .filter((token) => token.key.includes(word) || token.label.toLowerCase().includes(word) || word === '*')
```
NobodysNightmare, line 447: Do we know that `word` is already lower-cased? Otherwise it seems to me that searching for `Assignee` would never work.

Kharonus: the word is always lower case, it is converted when ~~parsed from the caret position~~ passed into the filter function - as we only want to filter on a predictable letter set.

NobodysNightmare: :green_circle: I think preferably we'd do the lower-casing of `word` inside `getFilteredSuggestionsData`, so that this method is responsible for ensuring matching characters alone and isn't dependent on callers knowing how to pass in a word.

NobodysNightmare (approved):

frontend/src/stimulus/controllers/pattern-input.controller.ts
Kharonus: true enough, with a near future patch I will move that.

PR #18888, opened 2025-05-09
[63831] Fix error 500 when bulk deleting related work packages

cbliard:
# Ticket

https://community.openproject.org/wp/63831

Prevent 500 error which can occur when bulk deleting work packages.

It occurs for instance in this situation: 3 work packages. One parent and two children following each other.

## Screenshots

Not applicable

When bulk deleting work packages, successors of descendants are rescheduled, because as they have one less predecessor, they may need to be switched to manual scheduling mode.

Problem: if one of these successor to reschedule is also a child of the deleted work package, then it cannot be rescheduled: it's deleted. When trying to rescheduled and save it, the error
`ActiveRecord::StaleObjectError` is raised.

Solution: only reschedule successors that are not descendants of the deleted work package.

# Merge checklist

- [x] Added/updated tests

### DIFF
--- a/spec/controllers/work_packages/bulk_controller_spec.rb
+++ b/spec/controllers/work_packages/bulk_controller_spec.rb
@@ -81,7 +81,7 @@ RSpec.describe WorkPackages::BulkController, with_settings: { journal_aggregatio
            principal: user,
            roles: [role])
   end
-  shared_let(:work_package1) do
+  shared_let(:work_package1, refind: true) do
     create(:work_package,
            author: user,
            assigned_to: user,
@@ -91,7 +91,7 @@ RSpec.describe WorkPackages::BulkController, with_settings: { journal_aggregatio
            custom_field_values: { custom_field1.id => custom_field_value },
            project: project1)
   end
-  shared_let(:work_package2) do
+  shared_let(:work_package2, refind: true) do
     create(:work_package,
            author: user,
            assigned_to: user,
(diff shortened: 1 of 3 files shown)


spec/controllers/work_packages/bulk_controller_spec.rb
```diff
@@ -643,5 +643,23 @@ def send_destroy_request
...
+        work_package1.update(subject: "parent", schedule_manually: false)
+        work_package2.update(subject: "predecessor child", parent: work_package1, schedule_manually: true)
+        work_package3.update(subject: "successor child", parent: work_package1, schedule_manually: false)
+        create(:follows_relation, predecessor: work_package2, successor: work_package3)
+      end
+
+      let(:params) { { "ids" => [work_package1.id, work_package2.id, work_package3.id] } }
+
+      it "deletes them all without errors" do
+        send_destroy_request
```
klaustopher: Might be a bit redundant, but since the error states that we are explicitly testing for not raising an error, WDYT about adding this to explicitly check that we do not raise an error:

```suggestion
        expect { send_destroy_request }.not_to raise_error
```

Of course, the result would be the same as the test would also fail with the raising, but it might make it a bit more obvious what we are explicitly expecting from the test

klaustopher (approved):

spec/controllers/work_packages/bulk_controller_spec.rb
cbliard: Good point. That indeed reads better.

PR #18895, opened 2025-05-12
Build a minimalistic Keycloak theme

NobodysNightmare:
This is intended to be used on community.openproject.org and it doesn't (yet) seem to be enough to warrant being moved to its own repository, though at some point we might want to think about moving the entire keycloak docker deployment out.

Having it in the repository means that users of a local Keycloak can use and test our theme locally as well.

# Ticket
https://community.openproject.org/wp/63771


docker/dev/keycloak/themes/openproject/account/resources/css/style.css
```diff
@@ -0,0 +1,32 @@
...
+:root body .pf-v5-c-masthead .pf-v5-c-toolbar {
+  --pf-v5-c-masthead--c-toolbar--BackgroundColor: #1A67A3;
+}
+
+body {
+  background-color: rgb(235, 243, 247);
+}
+
+.pf-v5-c-brand {
+  font-size: 2em;
```
Kharonus, line 29: ❓ any reason to use `em` instead of `rem`?

Do you want to override root fontsize?

docker/dev/keycloak/themes/openproject/account/resources/css/style.css
```diff
@@ -0,0 +1,32 @@
+:root body {
```
Kharonus, line 1: 🟡 if there is the possibility to use scss or sass, this file would look much slimmer :D any chance?

Kharonus (approved):

docker/dev/keycloak/themes/openproject/account/resources/css/style.css
NobodysNightmare: Totally makes sense. Mistake of someone not using CSS regularly enough.

docker/dev/keycloak/themes/openproject/account/resources/css/style.css
NobodysNightmare: I'd like to avoid that for now, because so far my goal is to provide a minimalistic setup. Something that's copy-pastable easily. Once I add a CSS precompiler I need setup instructions and stuff. That day might come, but I'd prefer it to not be today :)

PR #18897, opened 2025-05-12
Deprecate outdated API link

NobodysNightmare:
Having a separate "ensure connection" link does not add any value, because even the original "open" link does redirect unconnected users properly. On the other hand, immediately returning an ensure_connection URL is only possible if the storage was configured for mutual OAuth 2 authentication, not for authentication through an SSO provider.

# Ticket
https://community.openproject.org/wp/63691

### DIFF
--- a/docs/api/apiv3/components/schemas/project_storage_model.yml
+++ b/docs/api/apiv3/components/schemas/project_storage_model.yml
@@ -81,15 +81,18 @@ properties:
           - $ref: './link.yml'
           - description: |-
               A link to OpenProject strorage.
+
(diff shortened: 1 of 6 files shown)


ba1ash (approved):
LGTM
Only tests are red at the moment.

NobodysNightmare:
> Only tests are red at the moment.

I understood one of the conditions tested by the spec too late. I thought an early return from the representer code was being tested (but it's not). Instead an early return in the method I removed was being tested. Thus I removed the failing test as well. (because it tested the method that was removed)

PR #18898, opened 2025-05-12
Freeze string literals in the meetings module

NobodysNightmare:
Achieved through

    rubocop -A --only Style/FrozenStringLiteralComment,Layout/EmptyLineAfterMagicComment,Style/RedundantFreeze modules/meeting

Part of freezing string literals across the application step-by-step to be well prepared for future Ruby releases.

### DIFF
--- a/modules/meeting/app/components/authors/autocomplete_item_component.rb
+++ b/modules/meeting/app/components/authors/autocomplete_item_component.rb
@@ -1,4 +1,5 @@
 # frozen_string_literal: true
+
 #-- copyright
 # OpenProject is an open source project management software.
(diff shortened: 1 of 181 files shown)


myabc:
I only see whitespace changes (i.e. `Layout/EmptyLineAfterMagicComment` fixes)

NobodysNightmare:
You are right. Apparently there is no missing frozen string literals left. The only thing that was fixed is the whitespace after the magic comment **and** a few double freezes (`RedundantFreeze`).

So it's not _only_ whitespace, but much less than I expected when I copy-pasted the PR description ^^

PR #18901, opened 2025-05-12
fix upsell page

machisuji:
WP [#64014](https://community.openproject.org/projects/openproject/work_packages/64014/activity)

### DIFF
--- a/modules/team_planner/spec/controllers/team_planner/team_planner_controller_spec.rb
+++ b/modules/team_planner/spec/controllers/team_planner/team_planner_controller_spec.rb
@@ -38,8 +38,10 @@ RSpec.describe TeamPlanner::TeamPlannerController do
   before do
     login_as user
 
+    # We need to set this as the team planner controller needs to be used in the context of a project.
+    # Without a project, any action will result in a 403 forbidden response.
     controller.instance_variable_set(:@project, project)
   end
 
-  it_behaves_like "the upsell banner is shown"
+  it_behaves_like "it has an upsell action"
 end
--- a/modules/team_planner/spec/controllers/team_planner/team_planner_controller_spec.rb
+++ b/modules/team_planner/spec/controllers/team_planner/team_planner_controller_spec.rb
@@ -1,0 +1,45 @@
+# frozen_string_literal: true
+
+#-- copyright
+# OpenProject is an open source project management software.
+# Copyright (C) the OpenProject GmbH
+#
+# This program is free software; you can redistribute it and/or
+# modify it under the terms of the GNU General Public License version 3.
+#
+# OpenProject is a fork of ChiliProject, which is a fork of Redmine. The copyright follows:
+# Copyright (C) 2006-2013 Jean-Philippe Lang
+# Copyright (C) 2010-2013 the ChiliProject Team
+#
+# This program is free software; you can redistribute it and/or
+# modify it under the terms of the GNU General Public License
+# as published by the Free Software Foundation; either version 2
+# of the License, or (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program; if not, write to the Free Software
+# Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+#
+# See COPYRIGHT and LICENSE files for more details.
+#++
+
+require "spec_helper"
+require_relative "../../../../../spec/controllers/shared_upsell_examples"
+
+RSpec.describe TeamPlanner::TeamPlannerController do
+  let(:user) { create(:admin) }
+  let(:project) { create(:project) }
+
+  before do
+    login_as user
+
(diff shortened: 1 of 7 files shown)


spec/controllers/shared_upsell_examples.rb
```diff
@@ -0,0 +1,48 @@
...
+# along with this program; if not, write to the Free Software
+# Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+#
+# See COPYRIGHT and LICENSE files for more details.
+#++
+
+require "spec_helper"
+
+RSpec.shared_examples_for "the upsell banner is shown" do
+  subject { get :upsell }
```
NobodysNightmare, line 34: I am pretty unhappy about the naming of this shared example. Reading the specs I was surprised that there is a generic helper to check for the presence of an upsell banner, then I realized that the tests don't even specify which _action_ would show the banner (I was thinking about showing a banner on an edit page or something). Now I see that the action is something that we expect to be present on the tested controllers.

So I guess `it_behaves_like "having an upsell action"` would be my proposal.

I am also wondering whether there is a nicer way to include shared examples (i.e. avoiding `../../../`), but at least it's somewhat clear where the example comes from.

modules/team_planner/spec/controllers/team_planner/team_planner_controller_spec.rb
```diff
@@ -0,0 +1,45 @@
...
+require_relative "../../../../../spec/controllers/shared_upsell_examples"
+
+RSpec.describe TeamPlanner::TeamPlannerController do
+  let(:user) { create(:admin) }
+  let(:project) { create(:project) }
+
+  before do
+    login_as user
+
+    controller.instance_variable_set(:@project, project)
```
NobodysNightmare, line 43: It would at least be nice to explain (edit: **In a code comment**) why this is necessary. To me it's not _obvious_ why it's necessary to set a `@project` here. I could probably find out by seeing how exactly it fails when I remove this line, but by just scanning the code, it's not apparent to me why a `@project` might be required.

spec/controllers/shared_upsell_examples.rb
machisuji: Not sure about a nicer way to include these. For now this will have to do. I've addressed the other comment.

NobodysNightmare (approved):
Change LGTM. Though I have to admit that I don't quite understand the key part of what it's fixing, so let me write one last comment:

From what I see the Trial button was broken on the upsell pages, because we didn't call a helper that set two variables in `gon`. For starters: I have no clue what `gon` is and whether values set there have a longer lifetime than "per-request".

However, now I am wondering how upsell banners outside of `upsell` actions will fare. E.g. I added an enterprise banner to a storage edit page and it will surely also show a "Free Trial" button. Have I ever clicked that button to see if it works? Of course not! Have I prepared anything aside from `render(Enterprise::BannerComponent)`? Nope!

So I am now wondering why my enterprise banner should work, when the ones here do require extra work. And if I did the mistake, then others have a nice chance of having repeated said mistake. So what's your take? Am I overlooking something why this is only a problem on `upsell` actions or do we have a wider issue? What is gon and how is it making the trial buttons work?

machisuji:
> Change LGTM. Though I have to admit that I don't quite understand the key part of what it's fixing, so let me write one last comment:
> 
> From what I see the Trial button was broken on the upsell pages, because we didn't call a helper that set two variables in `gon`. For starters: I have no clue what `gon` is and whether values set there have a longer lifetime than "per-request".
> 
> However, now I am wondering how upsell banners outside of `upsell` actions will fare. E.g. I added an enterprise banner to a storage edit page and it will surely also show a "Free Trial" button. Have I ever clicked that button to see if it works? Of course not! Have I prepared anything aside from `render(Enterprise::BannerComponent)`? Nope!
> 
> So I am now wondering why my enterprise banner should work, when the ones here do require extra work. And if I did the mistake, then others have a nice chance of having repeated said mistake. So what's your take? Am I overlooking something why this is only a problem on `upsell` actions or do we have a wider issue? What is gon and how is it making the trial buttons work?

That's a good point. I'm not super familiar with the issue. But I think banners and the actual upsell pages are separate things. Does the banner you added have a "Start free trial" button, or does it just redirect to the admin/enterprise page? If it's the latter then you have nothing to worry about. As far as I can tell, we only really have those 3 places where the upsell page (it's a whole page, not just banner) is used.

As for gon, it's a gem that helps us share data from the backend with the frontend. It's basically just a hash that gets written into `window.gon` where it then can be accessed by the frontend code. In this case to get the necessary URL (and version) for requesting a trial.

NobodysNightmare:
My banner can look like this (it's a standard component, so all banners should behave the same):

image

I can't see what clicking the button does without clicking it (thank you Angular!)... _click_ ...

image

It's a modal on top of my page, so not a simple redirect... Apparently it's an `<enterprise-trial-modal>` that contains an `<enterprise-trial-form>`, but only Angular knows where it would be submitting to. I am not going to do that now, because I am deeply afraid of terms and conditions...

But maybe Angular has no clue where it would post that form:

image

machisuji:
@NobodysNightmare Feel free to try it, I assume this will result in an error too. So you are right. There are more places with banners that are probably all broken. I have no overview over where we have banners. But I imagine all those need fixing too then as they rely on the gon data.

NobodysNightmare:
Yeah, it fails as expected... I am going to ask for input on this...

PR #18902, opened 2025-05-12
Make feature tests more stable

cbliard:
Following advices on https://github.com/teamcapybara/capybara/issues/2796

This occurs sometimes that some tests are failing in headless mode: the DOM is not correctly reflecting the state of the page until a screenshot is taken.

This happened in ./spec/features/work_packages/table/queries/query_menu_refresh_spec.rb:44 in job https://github.com/opf/openproject/actions/runs/14965572620/job/42034828494

The screenshot shows that the query menu item is here, but before the screenshot is taken, capybara cannot find it.

This is a shot in the dark. I have no idea if that will really fix the issue mentioned above.

### DIFF
--- a/modules/meeting/spec/support/pages/meetings/show.rb
+++ b/modules/meeting/spec/support/pages/meetings/show.rb
@@ -226,9 +226,8 @@ module Pages::Meetings
           page.find_test_selector("op-meeting-agenda-actions").click
         end
         page.find(".Overlay")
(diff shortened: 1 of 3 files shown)


NobodysNightmare (approved):
I don't know whether this will fix the issue it wants to fix, but the settings added are well-explained and it's plausible to set them.

Let's try :+1:

klaustopher (approved):

PR #18903, opened 2025-05-12
Bug/63898 missing form label in work package comments rich text editor raises browser console error

akabiru:
https://community.openproject.org/work_packages/63898

Amend the text area label requirement introduced in https://github.com/opf/openproject/commit/6bf31400ed07e0ad1141160e6c2f8362a9683d28 to allow for visually hidden label as not all use cases fit this requirement- but **_there should always be a reference for screen-readers._**

As CKEditor re-renders the text area aria-label in every focus/blur event, we rely on the focusTracker to keep the aria-labelled-by up to date.

### DIFF
--- a/frontend/src/app/shared/components/editor/components/ckeditor-augmented-textarea/ckeditor-augmented-textarea.component.ts
+++ b/frontend/src/app/shared/components/editor/components/ckeditor-augmented-textarea/ckeditor-augmented-textarea.component.ts
@@ -67,8 +67,6 @@ import { uniqueId } from 'lodash';
 export class CkeditorAugmentedTextareaComponent extends UntilDestroyedMixin implements OnInit {
   @Input() public textAreaId:string;
 
-  @Input() public textAreaAriaLabel:string;
-
   @Input() public previewContext:string;
 
   @Input() public macros:ICKEditorMacroType;
@@ -221,12 +219,10 @@ export class CkeditorAugmentedTextareaComponent extends UntilDestroyedMixin impl
     // Set initial label
     this.setLabel();
 
-    // Use focusTracker to maintain aria-label as CKEditor re-renders ARIA-label every time on focus/blur
+    // Use focusTracker to maintain aria-labelledby as CKEditor re-renders aria-label on every focus/blur event
     // eslint-disable-next-line @typescript-eslint/no-unsafe-call,@typescript-eslint/no-unsafe-member-access
-    editor.ui.focusTracker.on('change:isFocused', (_evt:unknown, _name:string, isFocused:boolean) => {
-      if (isFocused) {
-        this.setLabel();
-      }
+    editor.ui.focusTracker.on('change:isFocused', (_evt:unknown, _name:string, _isFocused:boolean) => {
+      this.setLabel();
     });
 
     return editor;
@@ -321,10 +317,8 @@ export class CkeditorAugmentedTextareaComponent extends UntilDestroyedMixin impl
           .pipe(this.untilDestroyed())
           .subscribe(() => ckContent.focus());
       }
-    } else if (this.textAreaAriaLabel) {
-      ckContent.setAttribute('aria-label', this.textAreaAriaLabel);
     } else {
-      throw new Error(`Please provide a label for the textarea with id ${this.textAreaId} or set the textAreaAriaLabel property.`);
+      console.error(`Please provide a label for the textarea with id ${this.textAreaId}`);
     }
   }
 }
--- a/frontend/src/app/shared/components/editor/components/ckeditor-augmented-textarea/ckeditor-augmented-textarea.component.ts
+++ b/frontend/src/app/shared/components/editor/components/ckeditor-augmented-textarea/ckeditor-augmented-textarea.component.ts
@@ -297,28 +297,29 @@ export class CkeditorAugmentedTextareaComponent extends UntilDestroyedMixin impl
   }
 
   private setLabel() {
-    const ckContent = this.element.querySelector<HTMLElement>('.ck-content')!;
     const label = document.querySelector<HTMLLabelElement>(`label[for=${this.textAreaId}]`);
-
-    if (label) {
-      let labelId;
-      if (label.hasAttribute('id')) {
-        labelId = label.getAttribute('id')!;
-      } else {
-        labelId = uniqueId('label-');
-        label.setAttribute('id', labelId);
-      }
-
-      ckContent.removeAttribute('aria-label');
-      ckContent.setAttribute('aria-labelledby', labelId);
-
-      if (!this.labelClickSubscription) {
-        this.labelClickSubscription = fromEvent(label, 'click')
-          .pipe(this.untilDestroyed())
-          .subscribe(() => ckContent.focus());
-      }
-    } else {
+    if (!label) {
       console.error(`Please provide a label for the textarea with id ${this.textAreaId}`);
+      return;
+    }
+
+    const ckContent = this.element.querySelector<HTMLElement>('.ck-content')!;
+
+    let labelId;
+    if (label.hasAttribute('id')) {
+      labelId = label.getAttribute('id')!;
+    } else {
+      labelId = uniqueId('label-');
+      label.setAttribute('id', labelId);
+    }
+
+    ckContent.removeAttribute('aria-label');
+    ckContent.setAttribute('aria-labelledby', labelId);
+
+    if (!this.labelClickSubscription) {
+      this.labelClickSubscription = fromEvent(label, 'click')
+        .pipe(this.untilDestroyed())
+        .subscribe(() => ckContent.focus());
     }
   }
 }
--- a/frontend/src/app/shared/components/editor/components/ckeditor-augmented-textarea/ckeditor-augmented-textarea.component.ts
+++ b/frontend/src/app/shared/components/editor/components/ckeditor-augmented-textarea/ckeditor-augmented-textarea.component.ts
@@ -54,12 +54,11 @@ import {
   ICKEditorContext,
   ICKEditorInstance,
 } from 'core-app/shared/components/editor/components/ckeditor/ckeditor.types';
-import { fromEvent } from 'rxjs';
+import { fromEvent, Subscription } from 'rxjs';
 import { AttachmentCollectionResource } from 'core-app/features/hal/resources/attachment-collection-resource';
 import { populateInputsFromDataset } from 'core-app/shared/components/dataset-inputs';
 import { navigator } from '@hotwired/turbo';
 import { uniqueId } from 'lodash';
-
 
 @Component({
   templateUrl: './ckeditor-augmented-textarea.html',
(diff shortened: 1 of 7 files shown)


frontend/src/app/shared/components/editor/components/ckeditor-augmented-textarea/ckeditor-augmented-textarea.component.ts
```diff
@@ -288,23 +301,30 @@ export class CkeditorAugmentedTextareaComponent extends UntilDestroyedMixin impl
...
-      .subscribe(() => ckContent.focus());
+      if (!this.labelClickSubscription) {
+        this.labelClickSubscription = fromEvent(label, 'click')
+          .pipe(this.untilDestroyed())
+          .subscribe(() => ckContent.focus());
+      }
+    } else if (this.textAreaAriaLabel) {
+      ckContent.setAttribute('aria-label', this.textAreaAriaLabel);
+    } else {
+      throw new Error(`Please provide a label for the textarea with id ${this.textAreaId} or set the textAreaAriaLabel property.`);
```
myabc: I think we should reserve throwing Errors for critical failures, and instead go for `console.error` or `console.warn` (thus, not halting execution):

```suggestion
      console.error(`Please provide a label for the textarea with id ${this.textAreaId} or set the textAreaAriaLabel property.`);
```

(although I'd love a second opinion on this comment - @HDinger or @oliverguenther?)

lib/primer/open_project/forms/rich_text_area.html.erb
```diff
@@ -1,11 +1,13 @@
 <%= render(FormControl.new(input: @input)) do %>
   <%= content_tag(:div, hidden: true) do %>
-    <%= builder.text_area(@input.name,
-                          id: @text_area_id,
-                          **@input.input_arguments) %>
+    <%= builder.text_area(
+          @input.name,
+          id: @text_area_id,
+          **@input.input_arguments
+        ) %>
```
myabc: nitpick: It'd be good to have whitespace and formatting (erb_lint, rubocop) fixes as separate commits - to make it easier to focus on the salient changes.

frontend/src/app/shared/components/editor/components/ckeditor-augmented-textarea/ckeditor-augmented-textarea.component.ts
```diff
@@ -215,7 +218,17 @@ export class CkeditorAugmentedTextareaComponent extends UntilDestroyedMixin impl
...
+    // Set initial label
     this.setLabel();
+
+    // Use focusTracker to maintain aria-label as CKEditor re-renders ARIA-label every time on focus/blur
+    // eslint-disable-next-line @typescript-eslint/no-unsafe-call,@typescript-eslint/no-unsafe-member-access
+    editor.ui.focusTracker.on('change:isFocused', (_evt:unknown, _name:string, isFocused:boolean) => {
+      if (isFocused) {
+        this.setLabel();
+      }
+    });
```
myabc: 👍🏻 💯 for looking into this + documenting!

spec/features/activities/work_package/activity_tab_comment_editor_spec.rb
```diff
@@ -118,6 +118,25 @@ def expect_editor_to_be_dismissed_with_confirmation(&)
...
+    end
+
+    it "has an aria-label on the editor" do
+      activity_tab.add_comment(text: "Sample text", save: false)
+
+      activity_tab.expect_focus_on_editor
+
+      within_test_selector("augmented-text-area-notes") do
+        expect(page).to have_css(".ck-content[aria-label='Add a comment. Type @ to notify people.']")
+      end
```
myabc: You should be able to use [Capybara accessible selectors](https://github.com/citizensadvice/capybara_accessible_selectors) here. In fact, my original motivation for the change in introduced in 6bf3140 (PR #18478) was to fix labels so that they would work with accessible selectors.

Either using the locator (see [`rich_text` locator](https://github.com/citizensadvice/capybara_accessible_selectors?tab=readme-ov-file#rich_text)):

```suggestion
      expect(page).to have_selector :rich_text, "Add a comment. Type @ to notify people."
```
OR without locator, but filtered by `aria-label` attribute (see [`aria` [Hash] docs](https://github.com/citizensadvice/capybara_accessible_selectors?tab=readme-ov-file#aria-hash)) - if you want to be explicit about the need for `aria-label` being set

```suggestion
      expect(page).to have_selector :rich_text, aria: { label: "Add a comment. Type @ to notify people." }
```

I'm personally not a big fan of using test selectors in Capybara feature specs - I prefer to drive specs using semantic elements and labels. If you do want to use test selectors in this scenario though, you could also write:

```suggestion
      within_test_selector("augmented-text-area-notes") do
        expect(page).to have_css ".ck-content", aria: { label: "Add a comment. Type @ to notify people." }
      end
```

lib/primer/open_project/forms/rich_text_area.rb
```diff
@@ -17,6 +17,14 @@ def initialize(input:, rich_text_options:)
...
           @text_area_id = rich_text_options.delete(:text_area_id) || builder.field_id(@input.name)
         end
+
+        private
+
+        def rich_text_options
+          @rich_text_options.tap do |options|
+            options[:textAreaAriaLabel] = options.delete(:aria_label) if options.key?(:aria_label)
+          end
+        end
```
myabc: we need to process various `rich_text_options` before being passed as `inputs:` to `angular_component_tag`. Let's either do the Hash munging in the ERb template OR in `rich_text_area.rb`, but not in both.

e.g. 

```rb
          @rich_text_options = rich_text_options
          @text_area_id = rich_text_options.delete(:text_area_id) || builder.field_id(@input.name)
          @text_area_aria_label = rich_text_options.delete(:aria_label) if rich_text_options.key?(:aria_label)
```

myabc (changes requested):
Hi @akabiru 

The TS/JS code looks good! Thanks for documenting why we need the `focusTracker`.

I've commented on the Ruby code in various places. 

As a general point, rather than passing an `aria_label` option to `rich_text_options`, I would prefer it if we instead update `RichTextAreaInput` to properly support the `visually_hide_label` option - thereby making it consistent with other form inputs, e.g.:

```ruby
      notes_form.rich_text_area(
        classes: "ck-editor-primer-adjusted",
        name: :notes,
        label: I18n.t("activities.work_packages.activity_tab.label_type_to_comment"),
        visually_hide_label: true,
        rich_text_options: {
          showAttachments: false,
          resource:,
          editor_type: "constrained"
        }
      )
```

In addition to the feature specs, it would be great to see some specs for `RichTextArea`/`RichTextAreaInput` that test the correct options/inputs are passed to the Angular component. While we would only be testing server HTML rendering (no JS!), there are enough permutations of form input options to make this worth testing.

* I've started adding form specs under `spec/forms` but they're very much a work in progress - and my goal would be to test the form components under `lib/primer/open_project` themselves.

Get in touch if you have any questions about my comments. Cheers A!

lib/primer/open_project/forms/rich_text_area.html.erb
akabiru: Sure, this erb lint fix is a dedicated commit: https://github.com/opf/openproject/pull/18903/commits/0c44eb6562eee7ed5133dd67b270979247e93046 unless you mean as a separate PR?

lib/primer/open_project/forms/rich_text_area.rb
akabiru: Agree, considered this but didn't have better ideas at the time :)

spec/features/activities/work_package/activity_tab_comment_editor_spec.rb
akabiru: Nice one, I'll try this out! 👍🏾

akabiru (commented):
Thanks for the review @myabc great points, happy to address 💯

lib/primer/open_project/forms/rich_text_area.html.erb
akabiru: I'll split this out- as I'm no longer "touching" this file + will look into component unit tests in the same PR!

akabiru:
> 
> In addition to the feature specs, it would be great to see some specs for `RichTextArea`/`RichTextAreaInput` that test the correct options/inputs are passed to the Angular component. While we would only be testing server HTML rendering (no JS!), there are enough permutations of form input options to make this worth testing.
> 
>     * I've started adding form specs under `spec/forms` but they're very much a work in progress - and my goal would be to test the form components under `lib/primer/open_project` themselves.

Do you have a PR up for this, happy to add it on there. I took a stab and got stumped at how to stub out the `:builder` and `:form`, and I wonder if you've already solved this. I think we can aim for that to go into dev.

frontend/src/app/shared/components/editor/components/ckeditor-augmented-textarea/ckeditor-augmented-textarea.component.ts
```diff
@@ -288,23 +297,28 @@ export class CkeditorAugmentedTextareaComponent extends UntilDestroyedMixin impl
...
 
-    let labelId;
-    if (label.hasAttribute('id')) {
-      labelId = label.getAttribute('id')!;
-    } else {
-      labelId = uniqueId('label-');
-      label.setAttribute('id', labelId);
-    }
+    if (label) {
+      let labelId;
```
cbliard: I may be nitpicking, but I prefer an early return in a guard clause rather than a long if.

Something like this:

```js
const label = document.querySelector<HTMLLabelElement>(`label[for=${this.textAreaId}]`);
if (!label) {
  console.error(`Please provide a label for the textarea with id ${this.textAreaId}`);
  return
}

...
```

That makes things clearer for me.

akabiru: Sure, I like guard clauses too. Will update 👍🏾 fwiw, I had an else if case before which is now removed, which is why the "original" structure is this way.

spec/features/activities/work_package/activity_tab_comment_editor_spec.rb
```diff
@@ -118,6 +118,34 @@ def expect_editor_to_be_dismissed_with_confirmation(&)
...
+
+      activity_tab.expect_blur_on_editor
+      expect_rich_text_to_be_aria_labelled
+    end
+
+    def expect_rich_text_to_be_aria_labelled
+      expect(page).to have_selector(:label, for: "journal_notes", class: "sr-only",
+                                            text: "Add a comment. Type @ to notify people.")
+      label_id = page.find('label[for="journal_notes"]')["id"]
+      expect(page).to have_selector(:rich_text, aria: { labelledby: label_id })
```
myabc:
```suggestion
      expect(page).to have_selector(:rich_text, "Add a comment. Type @ to notify people.")
```

myabc (approved):

PR #18910, opened 2025-05-12
Stable seeding on deleted project admin role and projects

ulferts:
# Ticket

https://community.openproject.org/wp/62582
https://community.openproject.org/wp/63826

Aims to increase the process of reseeding an instance in case of:
* all projects being deleted
* roles being deleted AND/OR types being deleted 

`db:seed` is executed on all distributions whenever a new version of OP is installed. Instances where the above was true failed their seeding run. 

The problem was that references subsequent seeders relied upon were not set in previous seeders and that the `applicable?` checks of the `ProjectsSeeder` did not check for the existence of types and roles. Now, if roles and/or types are missing, the projects are not seeded as that would fail hard later on. 

# Merge checklist

- [x] Added/updated tests


ulferts:
Let's only merge this after 16.0 is released

app/seeders/basic_data/workflow_seeder.rb
```diff
@@ -26,40 +26,22 @@
...
 
     def seed_data!
-      if any_types_or_statuses_or_workflows_already_configured?
-        print_status "   *** Skipping types, statuses and workflows as there are already some configured"
-      else
-        seed_statuses
-        seed_types
-        seed_workflows
-      end
+      seed_workflows
```
cbliard, line 40: Good point to actually move the responsibility of seeding statuses and types one step higher to `BasicDataSeeder`s.

app/seeders/demo_data/projects_seeder.rb
```diff
@@ -74,5 +69,11 @@ def seedable_welcome_settings
         welcome_on_homescreen: 1
       }
     end
+
+    private
+
+    def project_seeders
+      @project_seeders ||= seed_data.each_data("projects").map { |project_data| ProjectSeeder.new(project_data) }
```
cbliard: minor but I like shorter lines
```suggestion
      @project_seeders ||= seed_data.each_data("projects")
                                    .map { |project_data| ProjectSeeder.new(project_data) }
```

modules/meeting/spec/seeders/demo_data/projects_seeder_spec.rb
```diff
@@ -0,0 +1,97 @@
...
+# You should have received a copy of the GNU General Public License
+# along with this program; if not, write to the Free Software
+# Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+#
+# See COPYRIGHT and LICENSE files for more details.
+# ++
+
+require "spec_helper"
+
+RSpec.describe DemoData::ProjectsSeeder do
```
cbliard: This file is located at `modules/meeting/spec/seeders/demo_data/projects_seeder_spec.rb`.
Shouldn't it rather be `spec/seeders/demo_data/projects_seeder_spec.rb` to match the location of `DemoData::ProjectsSeeder` in `app/seeders/demo_data/projects_seeder.rb`?

modules/meeting/spec/seeders/development_data/projects_seeder_spec.rb
```diff
@@ -0,0 +1,119 @@
...
+# You should have received a copy of the GNU General Public License
+# along with this program; if not, write to the Free Software
+# Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+#
+# See COPYRIGHT and LICENSE files for more details.
+# ++
+
+require "spec_helper"
+
+RSpec.describe DevelopmentData::ProjectsSeeder do
```
cbliard: Same, this file should rather be located at `spec/seeders/development_data/projects_seeder_spec.rb` to match the location of `DevelopmentData::ProjectsSeeder` in `app/seeders/development_data/projects_seeder.rb`.

app/seeders/demo_data/project_seeder.rb
```diff
@@ -60,6 +60,13 @@ def project_content_seeder_classes
       ]
     end
 
+    def applicable?
+      types = seed_data.lookup("types")
+
+      seed_data.reference_exists?(:default_role_project_admin) &&
+        (types.nil? || types.all? { |type_reference| seed_data.reference_exists?(type_reference) })
+    end
```
cbliard: I find the `types.nil? || types.all { |ref| ... }` part a bit confusing to read.

I think it would be better to change `types = seed_data.lookup("types")` into `types = seed_data.lookup("types") || []` to get rid of the `types.nil?` part.

And even better, as the same code is used line 104 to get the types from `project_data`, a method can be extracted:

Code becomes something like this:

```suggestion
    def applicable?
      seed_data.reference_exists?(:default_role_project_admin) \
        && project_data_types.all? { |type_reference| seed_data.reference_exists?(type_reference) }
    end

    private

    def set_types
      print_status "   -Assigning types."

      project.types = seed_data.find_references(project_data_types)
    end

    def project_data_types
      project_data.lookup("types") || []
    end
```

I'm not too sure about the method name. Could also well be `type_references_data`, `types_data`, or something else.

app/seeders/source/seed_data.rb
```diff
@@ -72,6 +72,10 @@ def find_references(references, default: :__unset__)
     Array(references).map { |reference| find_reference(reference, default:) }
   end
 
+  def reference_exists?(reference)
+    registry.key?(reference)
+  end
+
```
cbliard, line 78: Nice addition

There is a remaining call to `registry.key?` on line 39 which could be replaced by `reference_exists?` like you did line 56.

modules/meeting/spec/seeders/demo_data/projects_seeder_spec.rb
cbliard: I see, you probably did it because there was an existing `modules/meeting/spec/seeders/demo_data/project_seeder_spec.rb` there. But this file is actually testing the additions added by the meetings module so it's at the right place.

modules/meeting/spec/seeders/development_data/projects_seeder_spec.rb
```diff
@@ -0,0 +1,119 @@
...
+      end
+    end
+
+    context "with a project having identifier 'dev-custom-fields'" do
+      before { create(:project, identifier: "dev-custom-fields") }
+
+      it "returns false" do
+        expect(project_seeder).not_to be_applicable
+      end
+    end
```
cbliard:
```suggestion
    %w[dev-empty dev-work-package-sharing dev-large dev-large-child dev-custom-fields].each do |identifier|
      context "with a project having identifier '#{identifier}'" do
        before { create(:project, identifier: identifier) }

        it "returns false" do
          expect(project_seeder).not_to be_applicable
        end
      end
    end
```

cbliard (approved):
Nice work 👏 

I have some small improvement suggestions which can be ignored as you see fit.

I tried seeding multiple times after renaming types and "Project admin" roles. Each time it skipped project seeding or it did it if the data was correct.

Also, even if I think it fixes the reported issues, there are still possibilities for "ArgumentError: Nothing registered with reference" errors to occur in the seeders used by project.

For instance `WorkPackageBoardSeeder` tries to find references for `:default_status_new`, `:default_status_in_progress`, `:default_status_closed`, and `:default_status_rejected`. If one of them is not found, like renaming status "In progress" to "WIP", the `WorkPackageBoardSeeder` will fail to seed.

Not sure how far we want to fix this. Probably not worth it as we see no occurrence of such issue in AppSignal, so I'll happily approve this PR.

modules/meeting/spec/seeders/demo_data/projects_seeder_spec.rb
ulferts: Thanks for clearing that up. Lazy oversight on my end.

modules/meeting/spec/seeders/development_data/projects_seeder_spec.rb
ulferts: I had that first but didn't like it since I prefer verbose tests.

PR #18924, opened 2025-05-16
Move success translation to separate key

NobodysNightmare:
We do this mainly because we faced an issue with Crowdin, where the translation didn't work correctly in languages that have a separate translation for the zero plural case.

Looking at the way we use those translations though, it also makes sense in general to change the translations like this. Having success as the special case of zero failures is not quite right, because we also have zero failures if there is a single warning, but we don't show success in that case. Success is always a separate branch in a case statement and thus it makes sense to treat it as a separate translation.

# Ticket
https://community.openproject.org/wp/63568


cbliard (approved):
Thanks for fixing this i18n issue.
Even if that's a small change, the code definitely reads better like this.

PR #18933, opened 2025-05-16
[#63738, #63737] Primerize attribute help texts; Render in Settings > Information form

myabc:
# Ticket

https://community.openproject.org/wp/63738
https://community.openproject.org/wp/63737

This PR creates the following Primerized components:

* `AttributeHelpTextComponent` to render an icon that will invoke the
  attribute help text dialog on click.
* `AttributeLabelComponent` to render an attribute label and help text
  icon together with the correct spacing.
* `AttributeHelpTexts::ShowDialogComponent` to render the attribute help
  text dialog itself.

Also creates accompanying Controller actions, routes and specs.

It renders `AttributeHelpTextComponent` in Primerized forms automatically if the form model supports attribute help texts and the `include_help_text: true` option is passed to the form DSL method.

Additionally the ng `attribute-help-text-component` has been updated to use the question Octicon (due to time constraints, colours and other styling have not been adjusted).

## Screenshots

# Merge checklist

- [x] Added/updated tests


app/components/open_project/common/attribute_help_text_component.sass
```diff
@@ -0,0 +1,30 @@
...
+//
+// You should have received a copy of the GNU General Public License
+// along with this program; if not, write to the Free Software
+// Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+//
+// See COPYRIGHT and LICENSE files for more details.
+//++
+
+.op-attribute-help-text
+  line-height: 1em
```
HDinger:
```suggestion
  line-height: 1rem
```

**Before**
Bildschirmfoto 2025-06-03 um 09 00 17

**After**
Bildschirmfoto 2025-06-03 um 09 01 07

app/components/attribute_help_texts/show_dialog_component.html.erb
```diff
@@ -0,0 +1,33 @@
...
+        concat helpers.list_attachments(resource_representer, inputs: { allowUploading: false })
+      end
+    end
+
+    dialog.with_footer(show_divider: true) do
+      concat(
+        render(Primer::Beta::Button.new(data: { close_dialog_id: dialog_id })) { I18n.t(:button_close) }
+      )
+
+      if allowed_to_edit?
```
HDinger, line 19: Can we please change the order of the buttons? I know, that this was like that in the Angular component as well, but it makes absolutely no sense. Also from an accessibility point of view it is confusing.

app/components/attribute_help_texts/show_dialog_component.html.erb
```diff
@@ -0,0 +1,33 @@
...
+        end
+      )
+
+      if has_attachments?
+        concat render(Primer::Beta::Heading.new(tag: :h4)) { I18n.t(:label_attachment_plural) }
+        concat helpers.list_attachments(resource_representer, inputs: { allowUploading: false })
+      end
+    end
+
+    dialog.with_footer(show_divider: true) do
```
HDinger, line 18: I am not a fan of the divider but I see the benefit it brings for longer help texts. As discussed, I am fine with keeping it for now and review once the UX team is back.

HDinger (commented):
Looks very good, great job @myabc 👏  I only have some very minor remarks.

I noticed, when using the keyboard to close the Attribute help texts, the focus is not set back to the trigger but to the next element. This might however be an issue with the async dialogs themselves. We can leave that out of scope for this PR.

As discussed, there are a number of follow-up task we should track. Please create tickets for those things:

* [Bug] Missing help text for Subproject field (version 16.1)
* [Bug] Missing help text for Project status (version 16.1)
* [Code maintenance] Document this in the lookbook (version 16.1)
* [Code maintenance] Replace `opce-attribute-help-text` component with new helper (version 16.2 (?))
* [Bug] Focus should come back to trigger action when closing the Attribute help text (version 16.2(?))

HDinger (approved):
Good to merge once the CI is green.

myabc:
@HDinger Great! thanks. 

I was able to implement all of your suggestions. Thanks in particular for the hint about line height (I had been struggling with getting the baseline alignment to look right).

I also tidied up the commit messages, filled out the PR description and went ahead and referenced OP 63738 as well, since this PR basically implements that WP too.

myabc:
@HDinger One more FYI - I removed the changes that update the projects overview page from this PR (i.e. switching `Overviews::ProjectCustomFields::ItemComponent` from ng `opce-attribute-help-text` to `AttributeLabelComponent`). Although this work was helpful in making sure I got the implementation of `AttributeLabelComponent` (and in turn, `AttributeHelpTextComponent`) right, it is technically out-of-scope for this WP.

We can deal with it in the WP you suggested I create:
> [Code maintenance] Replace opce-attribute-help-text component with new helper (version 16.2 (?))

app/components/attribute_help_texts/show_dialog_component.rb
```diff
@@ -0,0 +1,57 @@
...
+  def title = @attribute_help_text.attribute_caption
+
+  def has_attachments? = @attribute_help_text.attachments.any?
+
+  def allowed_to_edit? = @current_user.allowed_globally?(:edit_attribute_help_texts)
+
+  def edit_button_href = url_helpers.edit_attribute_help_text_path(@attribute_help_text)
+
+  def resource_representer
+    ::API::V3::HelpTexts::HelpTextRepresenter.new(@attribute_help_text, current_user: @current_user, embed_links: false)
```
myabc, line 55: @HDinger I did notice that `embed_links: true` seems to have no effect on the number of API requests made when the ng `attachment-list-component` is rendered (via `AttachmentsHelper#list_attachment`)- I would have expected that by including the attachments in the serialized resource we wouldn't need to hit the API again.

This code was initially copied from the meetings module. @mrmir is this why you opted for `embed_links: false`?

(Not going to fix before merging - more of an FYI)

mrmir: @myabc I don't follow. What was copied over from meetings?

myabc: @mrmir I was referring to these lines:

```ruby
              resource = ::API::V3::Meetings::MeetingRepresenter.new(@meeting, current_user: User.current, embed_links: false)
              concat render(Primer::Beta::Heading.new(tag: :h4)) { I18n.t(:label_attachment_plural) }
              concat helpers.list_attachments(
                resource,
                inputs: { allowUploading: true }
              )
```

https://github.com/opf/openproject/blob/dev/modules/meeting/app/components/meetings/show_component.html.erb

However, I just did a git blame and it was @toy that worked on the lines in question, not you. Sorry for the noise.

toy: Only applying erb lint and on lines after `embed_links: false`, while that code was added in https://github.com/opf/openproject/commit/f58ac75cfec2e4be4be1698a575d408ef6fb82f9

